PR #7, opened 2025-08-30
fix: use the correct go test file name.

yafengio:
**What type of PR is this?**

/kind bug

Use the correct go test file name.


rootfs (approved):

Xunzhuo:
@rootfs i think the notify is not working for non-write access user, can u take a deeper look on this issue?

rootfs:
> @rootfs i think the notify is not working for non-write access user, can u take a deeper look on this issue?

sure or @yafengio do you like to fix [it](https://github.com/vllm-project/semantic-router/blob/main/.github/workflows/owner-notification.yml)?

yafengio:
> @rootfs i think the notify is not working for non-write access user, can u take a deeper look on this issue?

ok, I'm glad to try to fix it.

/assign

PR #26, opened 2025-09-03
docs: add vsr star history diagram

Xunzhuo:
**What type of PR is this?**

docs: add star history diagram, checkout at [preview](https://github.com/vllm-project/semantic-router/tree/dos-add-star)


rootfs (approved):
This is very impressive!

PR #74, opened 2025-09-07
chore: remove docker compose version, because verion is obsolete

yuluo-yx:
https://docs.docker.com/reference/compose-file/version-and-name/

`version` field is onsolete


Xunzhuo (approved):
lgtm。

PR #77, opened 2025-09-07
Use (*OpenAIRouter)(nil) for interface compliance check

cryo-zd:
**What type of PR is this?**

chore: Use (*OpenAIRouter)(nil) for interface compliance check

This avoids unnecessary allocation at package init and makes the intent clearer, following common Go idioms.

### DIFF
--- a/src/semantic-router/pkg/extproc/router.go
+++ b/src/semantic-router/pkg/extproc/router.go
@@ -37,7 +37,7 @@ type OpenAIRouter struct {
 }
 
 // Ensure OpenAIRouter implements the ext_proc calls
-var _ ext_proc.ExternalProcessorServer = &OpenAIRouter{}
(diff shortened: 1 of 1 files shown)


rootfs (approved):

rootfs:
@cryo-zd nice catch. Would you like to take one step further to add [go vet](https://pkg.go.dev/cmd/vet) in the CI

cryo-zd:
> @cryo-zd nice catch. Would you like to take one step further to add [go vet](https://pkg.go.dev/cmd/vet) in the CI

My pleasure. I plan to introduce a separate `vet` target in the Makefile, and then make `make test` depend on it. 

Does this approach sound reasonable to you?

rootfs:
sounds good!

PR #78, opened 2025-09-07
docs: add python install setups in install-local

yuluo-yx:


yuluo-yx:
I found that the code file in the project does not have a declaration of license.

such as: 
https://github.com/vllm-project/vllm/blob/0661cb9df390a47dda0a5c201cf712dd2d943666/setup.py#L1
https://github.com/apache/shenyu/blob/e624a19aae027361f309cb3e236d6345962ac811/pom.xml#L3

website/docs/getting-started/installation.md
```diff
@@ -44,6 +45,16 @@ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 source ~/.cargo/env
 ```
 
+#### Install Python (if not already installed)
+```bash
+# Check if Python is installed
+python --version
+
+# If not installed:
+# macOS: brew install go
```
rootfs: this looks a `golang` install. Does brew come with built-in python?

yuluo-yx: yep, my bad. I'll push it later today.

yuluo-yx: fixed. pls

PR #84, opened 2025-09-08
feat(config): watch config file and hot-reload router without restart

### LINKED ISSUE #82: Router should watch config yaml and load it when the config is updated
**Is your feature request related to a problem? Please describe.**
The router should be re-configured online without having to restart if the config yaml is updated.

**Describe the solution you'd like**
In a kubernetes environment, the config yaml can be refreshed as a ConfigMap, the router should watch the config yaml and load it if it is modified.

tao12345666333:
**What type of PR is this?**

Fixes #82 

Release Notes: Yes/No

### DIFF
--- a/src/semantic-router/pkg/config/parse_configfile_test.go
+++ b/src/semantic-router/pkg/config/parse_configfile_test.go
@@ -1,0 +1,62 @@
+package config_test
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/vllm-project/semantic-router/semantic-router/pkg/config"
+)
+
+var _ = Describe("ParseConfigFile and ReplaceGlobalConfig", func() {
+	var tempDir string
+
+	BeforeEach(func() {
+		var err error
+		tempDir, err = os.MkdirTemp("", "config_parse_test")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		os.RemoveAll(tempDir)
+		config.ResetConfig()
+	})
+
+	It("should parse configuration via symlink path", func() {
+		if runtime.GOOS == "windows" {
+			Skip("symlink test is skipped on Windows")
+		}
+
+		// Create real config target
+		target := filepath.Join(tempDir, "real-config.yaml")
+		content := []byte("default_model: test-model\n")
+		Expect(os.WriteFile(target, content, 0o644)).To(Succeed())
+
+		// Create symlink pointing to target
+		link := filepath.Join(tempDir, "link-config.yaml")
+		Expect(os.Symlink(target, link)).To(Succeed())
+
+		cfg, err := config.ParseConfigFile(link)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(cfg).NotTo(BeNil())
+		Expect(cfg.DefaultModel).To(Equal("test-model"))
+	})
+
+	It("should return error when file does not exist", func() {
+		_, err := config.ParseConfigFile(filepath.Join(tempDir, "no-such.yaml"))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("failed to read config file"))
+	})
+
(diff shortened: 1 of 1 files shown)


src/semantic-router/pkg/config/config.go
```diff
@@ -291,30 +292,56 @@ var (
...
-
 	if configErr != nil {
 		return nil, configErr
 	}
+	configMu.RLock()
+	defer configMu.RUnlock()
 	return config, nil
 }
 
+// ParseConfigFile parses the YAML config file without touching the global cache.
```
rootfs, line 318: can you add a unit test for this too?

tao12345666333: Sure

src/semantic-router/pkg/config/config.go
```diff
@@ -291,30 +292,56 @@ var (
...
 	configErr  error
+	configMu   sync.RWMutex
 )
 
-// LoadConfig loads the configuration from the specified YAML file
+// LoadConfig loads the configuration from the specified YAML file once and caches it globally.
 func LoadConfig(configPath string) (*RouterConfig, error) {
 	configOnce.Do(func() {
-		data, err := os.ReadFile(configPath)
+		cfg, err := ParseConfigFile(configPath)
```
rootfs, line 301: Would you please consider emitting an k8s event if there is an err (if running in k8s)

src/semantic-router/pkg/extproc/router.go
```diff
@@ -41,10 +41,13 @@ var _ ext_proc.ExternalProcessorServer = (*OpenAIRouter)(nil)
 
 // NewOpenAIRouter creates a new OpenAI API router instance
 func NewOpenAIRouter(configPath string) (*OpenAIRouter, error) {
-	cfg, err := config.LoadConfig(configPath)
+	// Always parse fresh config for router construction (supports live reload)
+	cfg, err := config.ParseConfigFile(configPath)
```
rootfs, line 45: nit: emit an event if running in k8s and also log this so we can track a new config is applied

src/semantic-router/pkg/extproc/server.go
```diff
@@ -78,6 +91,105 @@ func (s *Server) Start() error {
...
+			})
+			return
+		}
+		s.service.Swap(newRouter)
+		observability.LogEvent("config_reloaded", map[string]interface{}{
+			"file": cfgFile,
+		})
+	}
+
+	for {
```
rootfs, line 166: can you add a followup pr to test config file update under different scenarios?

tao12345666333: OK

src/semantic-router/pkg/config/config.go
tao12345666333: I have created an issue for tracking https://github.com/vllm-project/semantic-router/issues/91
Because I think adding events is a relatively independent part

rootfs: sgtm

src/semantic-router/pkg/extproc/server.go
tao12345666333: https://github.com/vllm-project/semantic-router/issues/92 for tracking

src/semantic-router/pkg/extproc/router.go
rootfs: resolving this one as it is related to #84 too

src/semantic-router/pkg/config/config.go
tao12345666333: Added here https://github.com/vllm-project/semantic-router/pull/84/commits/784d67bc2eecc5f11f3515e125c78bb208bd228d

yuluo-yx:
hi, @tao12345666333 @rootfs 
 
for this regard, I have an idea, why don't we use viper directly? 🤔

https://github.com/spf13/viper

tao12345666333:
> for this regard, I have an idea, why don't we use viper directly? 🤔
> 
> https://github.com/spf13/viper

I don't want to introduce viper because it's over-engineered and has some irrelevant dependencies.

Here we need more precise control, and we don't need the complex templates and environment variable binding features in viper.

PR #96, opened 2025-09-08
fix: don't set reasoning effort for non-reasoning models

rootfs:
**What type of PR is this?**

If the model doesn't support reasoning, don't set the reasoning effort in API call.

This is a regression when reasoning mode support is added

Fixes #

Release Notes: Yes/No


rootfs:
This hardcoded approach is quite fragile and not scalable. Let me try another way.

PR #97, opened 2025-09-08
fix: don't set reasoning effort for non-reasoning models

rootfs:
**What type of PR is this?**

If the model doesn't support reasoning, don't set the reasoning effort in API call.

This is a regression when reasoning mode support is added

Fixes #

Release Notes: Yes/No

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -253,4 +253,32 @@ api:
       sample_rate: 1.0                   # Collect metrics for all requests (1.0 = 100%, 0.5 = 50%)
       # Histogram buckets for metrics (directly configure what you need)
       duration_buckets: [0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30]
-      size_buckets: [1, 2, 5, 10, 20, 50, 100, 200] 
+      size_buckets: [1, 2, 5, 10, 20, 50, 100, 200]
+
+# Model reasoning configurations - define how different models handle reasoning syntax
+model_reasoning_configs:
+  - name: "deepseek"
+    patterns: ["deepseek", "ds-", "ds_", "ds:", "ds "]
+    reasoning_syntax:
+      type: "chat_template_kwargs"
+      parameter: "thinking"
+  
+  - name: "qwen3"
+    patterns: ["qwen3"]
+    reasoning_syntax:
+      type: "chat_template_kwargs"
+      parameter: "enable_thinking"
+  
+  - name: "gpt-oss"
+    patterns: ["gpt-oss", "gpt_oss"]
+    reasoning_syntax:
+      type: "reasoning_effort"
+      parameter: "reasoning_effort"
+  
+  - name: "gpt"
+    patterns: ["gpt"]
+    reasoning_syntax:
+      type: "reasoning_effort"
+      parameter: "reasoning_effort"
+  
+# No default fallback - unknown models should have no reasoning syntax applied 
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -57,6 +57,28 @@ model_config:
       pii_types_allowed: ["EMAIL_ADDRESS", "PERSON", "GPE", "PHONE_NUMBER"]  # Only allow these specific PII types
     # Specify which endpoints can serve this model (optional - if not specified, uses all endpoints that list this model)
     preferred_endpoints: ["endpoint1", "endpoint3"]
+    # Reasoning family - phi4 doesn't support reasoning, so omit this field
+  
+  # Example: DeepSeek model with custom name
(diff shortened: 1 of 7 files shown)


rootfs:
cc @tao12345666333

src/semantic-router/pkg/config/config.go
```diff
@@ -264,6 +268,73 @@ type Category struct {
...
+		for _, pattern := range config.Patterns {
+			if matchesPattern(modelLower, strings.ToLower(pattern)) {
+				return &config
+			}
+		}
+	}
+
+	// Look for a default/fallback configuration
+	for _, config := range rc.ModelReasoningConfigs {
+		if config.Name == "default" || contains(config.Patterns, "*") {
```
tao12345666333: I think it's better not to hardcode config.Name and just keep the judgment logic for config.Patterns. 

Otherwise, we will always need to consider the default name separately in our subsequent processing logic.

WDYT?

src/semantic-router/pkg/config/config.go
```diff
@@ -264,6 +268,73 @@ type Category struct {
...
+
+	// Handle prefix patterns like "ds-", "ds_", "ds:", "ds "
+	if strings.HasSuffix(pattern, "-") || strings.HasSuffix(pattern, "_") ||
+		strings.HasSuffix(pattern, ":") || strings.HasSuffix(pattern, " ") {
+		return strings.HasPrefix(modelName, pattern) ||
+			(len(modelName) >= len(pattern)-1 && modelName == pattern[:len(pattern)-1])
+	}
+
+	// Handle exact matches and contains
+	return strings.Contains(modelName, pattern)
```
tao12345666333: Perhaps it would be more convenient for us to use regular expressions?

src/semantic-router/pkg/extproc/reason_mode_selector.go
```diff
@@ -47,61 +48,42 @@ func (r *OpenAIRouter) getReasoningModeAndCategory(query string) (bool, string)
...
-	}
-	if strings.Contains(lower, "deepseek") || hasDeepSeekAlias(lower) {
-		return "deepseek", "thinking"
+// buildReasoningRequestFields returns the appropriate fields to add to the request based on model config
+func (r *OpenAIRouter) buildReasoningRequestFields(model string, useReasoning bool, categoryName string) (map[string]interface{}, string) {
+	modelConfig := r.getModelReasoningConfig(model)
+	if modelConfig == nil {
+		// No configuration found for this model - don't apply any reasoning syntax
+		// Unknown models should not have reasoning fields added
+		return nil, "N/A"
```
tao12345666333: We can set N/A as a constant, I see it used in many places

config/config.yaml
```diff
@@ -253,4 +253,32 @@ api:
...
+      type: "reasoning_effort"
+      parameter: "reasoning_effort"
+  
+  - name: "gpt"
+    patterns: ["gpt"]
+    reasoning_syntax:
+      type: "reasoning_effort"
+      parameter: "reasoning_effort"
+  
+# No default fallback - unknown models should have no reasoning syntax applied
```
liangyuanpeng:
```suggestion
# No default fallback - unknown models should have no reasoning syntax applied 

# Global default reasoning effort level
default_reasoning_effort: "medium"
```

Nit:  How about moving the `default_reasoning_effort` configuration here to keep it consistent with the website doc?

Xunzhuo:
let us stop the cycles round reasoning string match、pattern etc works for a better choice. I think we should have better approach to take care of the auto reasoning logics, the modelname can be very dynamic since that is set in vllm startup options, like i can set dsv31 as the modelname for deepseek-v3.1.

we should have the fixed groups/family names for models like we have supported:
1. deepseek
2. qwen
3. gpt-oss
theses three vars should be fixed and the model name just is an alias around it. so maybe we should have an extra params for modelconfig like reasoning-familiy:

deepseek:
```
model_config:
  ds-v31-balabala:
    reasoning-family: deepseek
......
    preferred_endpoints:
    - endpoint1
```

gpt-oss:
```
model_config:
  gptoss-balabala:
    reasoning-family: gpt-oss
......
    preferred_endpoints:
    - endpoint1
```

qwen
```
model_config:
  qwen3-balabala:
    reasoning-family: qwen3
......
    preferred_endpoints:
    - endpoint1
```

the modelname can be very flexible, like for deepseek family, we can set (ds-v3,self-hosted/ds,official/deepseek, anything we want, since that is used in modelname in request body) we just care about what the model reasoning family is. if it is empty, we just think it is not supported hybrid reasoning.

rootfs:
@Xunzhuo model family is a good idea. It is added to the new config now.

rootfs:
review comments addressed, merging it so I can add other PRs using this new config.

PR #113, opened 2025-09-11
refactor: move classifier model init to classifier.go and unify the classifier model init logic

### LINKED ISSUE #107: refactor: move classifier model init to classifier.go and unify the classifier model init logic
1. `candle_binding.InitModernBertJailbreakClassifier` or `candle_binding.InitJailbreakClassifier` was called twice in `NewOpenAIRouter`. 
    1. Location 1: https://github.com/vllm-project/semantic-router/blob/fe0b5b58a09185762a8e2a4217150306fc514884/src/semantic-router/pkg/extproc/router.go#L232-L250  
    2. Location 2: https://github.com/vllm-project/semantic-router/blob/fe0b5b58a09185762a8e2a4217150306fc514884/src/semantic-router/pkg/extproc/router.go#L140-L145
2. The init logic has some differences:
    1. In Location 1, when `cfg.PromptGuard.ModelID` is empty, it will use the value of `cfg.BertModel.ModelID`. In Location 2, it skips init when `cfg.PromptGuard.ModelID` is empty.
    2. In Location 1, when `numClasses < 2`, it doesn't return error. In Location 2, it returns error.
3. Category and PII classifier model init logic currently are outside of `classifier.go`

My proposal:
1. Move Category and PII classifier model init logic to `classifier.go`.
2. Call all 3 classifier model inits in `NewClassifier`.

Benefit:
1. Unify the initialization logic, improve maintainability.
2. Improve testability.

I want to confirm the init logic: 
1. When `cfg.Classifier.CategoryModel.ModelID` or `cfg.Classifier.PIIModel.ModelID` or `cfg.PromptGuard.ModelID` are empty, do we use `cfg.BertModel.ModelID` or skip init? I don't understand why it used `cfg.BertModel.ModelID`, because in the comments it said `cfg.BertModel.ModelID` is the model for similarity comparison. I prefer skip inits, and follow `IsJailbreakEnabled()` to take these cases as model disabled.
2. There is a `JailbreakInitialized` and `c.IsJailbreakEnabled()` check before Jailbreak classification (if false, return nil quickly). But for Category and PII classification, they return errors when mapping is nil. How about we unify the logic here, return error for Jailbreak classification? The caller should check `c.IsJailbreakEnabled()` before call this method. `JailbreakInitialized` can be removed since the `NewClassifier` should guarantee the model init successful when `c.IsJailbreakEnabled()`.
3. When `numClasses < 2`, do we return error or not in init? (I think should return error, silently fail is not good in my opinion)

Do you have some suggestions?

Thanks! I am willing to submit a PR when the above questions are clear.

aeft:
**What type of PR is this?**

refactor: move classifier model init to classifier.go and unify the classifier model init logic

The main change:
1. move classifier model init logic (`candle_binding.Init*` ) to `classifier.go`. 

Other minor changes:
1. provide `Is*Enabled()` method and check for all category/jailbreak/pii. 
2. remove `initMutex` in `router.NewOpenAIRouter()`. I thought it was used to guard model init, and actually we have `sync.Once.Do` for all model init in `candle-binding`. I think we don't need this mutex anymore.
3. The main logic of `ClassifyCategory` doesn't change (except for the guard if condition). I just change the position of the code.

Other thoughts:
1. I don't change `test_utils_test.go` or `security_test.go` much currently, since finally it should go to the individual integration test package.

Fixes #107 

Release Notes: No


rootfs:
(from Copilot)

This pull request refactors and improves the classifier initialization logic in the semantic-router codebase. Here’s a summary and review of the key changes:

## Summary of Changes

### 1. **Classifier Initialization Refactor**
- **Old approach:**  
  Initialization was managed via global variables (`initialized`, `initMutex`) and a large `initializeModels` function.
- **New approach:**  
  Initialization is now encapsulated using interfaces (`CategoryInitializer`, `JailbreakInitializer`, `PIIInitializer`) and injected via options into the Classifier struct.  
  - Each model type (category, jailbreak, PII) now has a dedicated initializer and inference implementation.
  - Model initialization errors are propagated and handled explicitly.

### 2. **Error Handling**
- Initialization functions now return errors when configuration is missing or invalid (e.g., not enough classes, missing mapping, or model ID).
- The code now checks for proper configuration before running classification and returns meaningful errors rather than silent failures or nil results.

### 3. **Removal of Global State**
- Removed global initialization state and mutexes, reducing risks of concurrency bugs and simplifying the code.

### 4. **Unit Test Improvements**
- Tests have been updated to reflect the new initialization logic and error handling.
- Mock initializers and inference implementations are used for precise, controlled testing of error scenarios.
- Tests now assert on error cases for missing config, not enough classes, or failed initialization.

### 5. **Code Organization**
- Initialization logic is broken down into smaller, more maintainable functions.
- Option pattern is used for dependency injection, improving testability and flexibility.

---

## Review

### **Strengths**
- **Robustness:** Error handling is greatly improved. Misconfiguration (missing model ID, mappings, etc.) is now surfaced clearly and early.
- **Test Coverage:** Unit tests are more comprehensive and check edge cases, including error scenarios.
- **Design:** The use of interfaces and the option pattern enables easier extension and testing.
- **Safety:** Global mutable state is eliminated, reducing side effects and making the code more thread-safe.

### **Potential Considerations**
- **Migration Complexity:** If other parts of the codebase depend on the previous global state or initialization sequence, they may need updates.
- **Performance:** Initializing models multiple times (if not managed carefully) could be expensive. However, this PR appears to maintain one-time initialization per Classifier instance.
- **Integration Testing:** Some TODOs (e.g., using a real model path in tests) suggest some integration tests may still need refinement.

### **Code Style and Clarity**
- The code is significantly cleaner and easier to follow.
- Interface-based design provides clear contracts for initialization and inference.

### **Backward Compatibility**
- Any external consumers of the Classifier API need to update usage to handle errors (now returned by constructors).
- Global initialization logic is removed, so consumers need to ensure correct model setup per instance.

---

## Recommendation

**This PR is a strong improvement for code quality, reliability, and testability.**  
Before merging, ensure:
- All integration tests (using real models) are passing.
- Downstream code (if any) is updated to use the new constructor and error handling.
- Documentation is updated to reflect the new initialization pattern.

**Overall, I recommend merging after validating integration tests and any downstream dependencies.**  
Let me know if you want comments on specific files or a deeper dive!

PR #115, opened 2025-09-11
Refactor: use worker pool for batch classification concurrency

cryo-zd:
**What type of PR is this?**

Refactor: use worker pool for batch classification concurrency

This PR replaced per-task goroutines with a fixed-size worker pool (<= `max_concurrency`). This ensures accurate concurrency metrics, aligns with API semantics, and avoids overhead when batch size is large.

This PR refactors `processConcurrently` to address two issues:
- The current implementation spawns one goroutine per text and increments `ConcurrentGoroutines` at creation time. This makes this metric does not accurately reflect the number of simultaneously running goroutines, which is limited by the `semaphore` to `maxConcurrency`. 
  As a result, monitoring may show thousands of goroutines even when `max_concurrency` is set to 8, which diverges from the API contract that defines `max_concurrency` as the maximum concurrent goroutines.
- Large batches could create excessive goroutines, adding unnecessary scheduling overhead.

[link](https://vllm-semantic-router.com/docs/api/classification#configuration)
api

### DIFF
--- a/src/semantic-router/pkg/api/server.go
+++ b/src/semantic-router/pkg/api/server.go
@@ -563,45 +563,53 @@ func (s *ClassificationAPIServer) processConcurrently(texts []string, options *C
 	if s.config != nil && s.config.API.BatchClassification.MaxConcurrency > 0 {
 		maxConcurrency = s.config.API.BatchClassification.MaxConcurrency
 	}
+	// Get the actual number of workers to start
+	numWorkers := min(len(texts), maxConcurrency)
 
 	results := make([]services.Classification, len(texts))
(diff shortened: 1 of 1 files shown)


rootfs:
@cryo-zd this is interesting! do you happen to have a test to see the relationship between concurrency and CPU/GPU usage?

rootfs:
cc @aeft for recent semaphore refactor

aeft:
> cc @aeft for recent semaphore refactor

looks good to me

rootfs:
@cryo-zd thanks for this work! If possible, can you follow up with a PR to share the concurrency benchmark?

PR #127, opened 2025-09-13
perf: enable concurrent classification via Arc+clone

cryo-zd:
**What type of PR is this?**

perf: enable concurrent classification via Arc+clone

This PR updates the `MODERNBERT_CLASSIFIER` definition to use `Arc<ModernBertClassifier>` so that the global lock is only held for initialization. Subsequent calls now clone the `Arc` and release the lock immediately, allowing concurrent inference without serialization.
[concurrency-safe guarantee]
sendsync
For now, only `MODERNBERT_CLASSIFIER` is changed; `MODERNBERT_PII_CLASSIFIER` and `MODERNBERT_JAILBREAK_CLASSIFIER` remain unchanged, so that we can focus discussion on this design and trade-offs before extending it further.

Fixes #

Release Notes: Yes/No

### DIFF
--- a/candle-binding/src/modernbert.rs
+++ b/candle-binding/src/modernbert.rs
@@ -268,7 +268,7 @@ pub struct ModernBertClassifier {
 }
 
 lazy_static::lazy_static! {
-    static ref MODERNBERT_CLASSIFIER: Arc<Mutex<Option<ModernBertClassifier>>> = Arc::new(Mutex::new(None));
+   static ref MODERNBERT_CLASSIFIER: Arc<Mutex<Option<Arc<ModernBertClassifier>>>> = Arc::new(Mutex::new(None));
     static ref MODERNBERT_PII_CLASSIFIER: Arc<Mutex<Option<ModernBertClassifier>>> = Arc::new(Mutex::new(None));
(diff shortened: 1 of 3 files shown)


candle-binding/src/modernbert.rs
```diff
@@ -930,18 +930,23 @@ pub extern "C" fn classify_modernbert_text(text: *const c_char) -> ModernBertCla
...
-            Err(e) => {
-                eprintln!("Error classifying text with ModernBERT: {e}");
-                default_result
+    let classifier_arc_opt = MODERNBERT_CLASSIFIER.lock().unwrap();
+    match &*classifier_arc_opt {
+        Some(classifier_arc) => {
+            let classifier = classifier_arc.clone();
+            drop(classifier_arc_opt);
+
+            match classifier.classify_text(text) {
```
rootfs: this still under the lock

cryo-zd: Aha, let me try to fix it

rootfs:
@cryo-zd thanks for making this happen! If you can have a followup PR for performance test suite to eval the latency with respect to concurrency, that'll be great!

PR #134, opened 2025-09-14
logging: unify stdlib log usage to pkg/observability (zap)

tao12345666333:
**What type of PR is this?**

logging: unify stdlib log usage to pkg/observability (zap)

Since we previously chose zap as the unified logging library, in this PR I will standardize and normalize all logging operations., in this PR I will standardize and normalize all logging operations.

Fixes #

Release Notes: Yes/No


src/semantic-router/cmd/main.go
```diff
@@ -26,43 +25,44 @@ func main() {
 
 	// Initialize logging (zap) from environment.
 	if _, err := observability.InitLoggerFromEnv(); err != nil {
-		log.Printf("failed to initialize logger, falling back to stdlib: %v", err)
```
rootfs, line 29: is there a reason `InitLoggerFromEnv` could fail?

src/semantic-router/pkg/extproc/processor.go
```diff
@@ -26,62 +26,62 @@ func (r *OpenAIRouter) Process(stream ext_proc.ExternalProcessor_ProcessServer)
...
 				return err
 			}
 			if err := sendResponse(stream, response, "request header"); err != nil {
-				log.Printf("ERROR: sendResponse for headers failed: %v", err)
+				observability.Errorf("sendResponse for headers failed: %v", err)
 				return err
 			}
 
 		case *ext_proc.ProcessingRequest_RequestBody:
-			log.Printf("DEBUG: Processing Request Body - THIS IS WHERE ROUTING HAPPENS")
```
rootfs, line 76: i think this message could be removed

src/semantic-router/cmd/main.go
tao12345666333: For example, it will fail when the value set by SR_LOG_ENCODING is not supported by zap.

rootfs (approved):

src/semantic-router/pkg/extproc/processor.go
tao12345666333: Sure, removed.

PR #105, opened 2025-09-09
feat: add milvus persistent storage support

rootfs:
**What type of PR is this?**

This is a WIP to support more persistent storage for semantic cache

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #94 #95 

Release Notes: Yes/No

PR #139, opened 2025-09-15
docs: add semantic cache doc to explain how to use in-memory and milvus in the config

rootfs:
**What type of PR is this?**

This is a followup PR of #105, adding missing doc about how to use the semantic cache for development and production

Fixes #

Release Notes: Yes/No


Xunzhuo (changes requested):
plz add the docs under website/sidebars.js

PR #154, opened 2025-09-17
fix: use request id to locate the correct cache entry to update

### LINKED ISSUE #144: Bug: The cache might have been incorrectly updated
In `UpdateWithResponse`, we only use `query` to find the cache entry and then fill the response. However, it's possible for two queries with the same value to occur at the same time. For example, if they use different models, the cached result may be incorrect in this case.

https://github.com/vllm-project/semantic-router/blob/31069a9caeb9335eb8c50160d0d0ceb9c2c42194/src/semantic-router/pkg/extproc/response_handler.go#L157-L165

(here `cacheID` is `query`)

I propose using `ctx.RequestID` to locate the correct cache because it's unique. That is, I will add `RequestID` to the `CacheEntry` struct. Also, we don't need to return `cacheID` anymore.

---

By the way, why do we use this two-step design? It seems like if the second step (filling the LLM response) fails, the first step is useless. Why don't we add the cache entry only when we get the full response?

aeft:
**What type of PR is this?**

fix: use request id to locate the correct cache entry to update

The major change:
1. Update the cache interface to use request id to locate the cache entry

Other changes:
-   Added `request_id` to Milvus storage.
-   Replaced the `insert` operation with `upsert` in `MilvusCache.addEntry`. Previously, it always inserted two entries into storage—one for `AddPendingRequest` and another for `UpdateWithResponse`. Now, it updates the entry if a cache entry with the same `request_id` already exists.
-   Removed the logic that added an unknown cache entry when no cache entry was found in `UpdateWithResponse`. I think it's not useful because we this unknown cache entry doesn't have `model` field.
-   Removed `pendingRequests` and `pendingRequestsLock` since they are no longer in use.
-   Added a Docker command to `milvus.mk` to enable starting Attu (the Milvus UI), which is helpful for browsing data.

Fixes #144

Release Notes: Yes/No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -205,6 +205,7 @@ The test suite includes:
 - **Envoy logs:** Check the terminal running `make run-envoy` for detailed request/response logs
 - **Router logs:** Check the terminal running `make run-router` for classification and routing decisions
 - **Rust library:** Use `RUST_LOG=debug` environment variable for detailed Rust logs
+- **Go library:** Use `SR_LOG_LEVEL=debug` environment variable for detailed Go logs
 
 ## Code Style and Standards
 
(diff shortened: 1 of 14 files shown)


Xunzhuo:
@aeft can you DM me in slack, thanks

PR #156, opened 2025-09-17
docs(router.md): add error metrics and example queries for llm_request_errors_total

### LINKED ISSUE #133: docs:  Add description for RequestErrorsTotal
can you add to the doc too in a followup PR? thanks

_Originally posted by @rootfs in https://github.com/vllm-project/semantic-router/pull/132#discussion_r2347390731_

samzong:
**What type of PR is this?**

docs: add error metrics and example queries for llm_request_errors_total

Fixes #133

Release Notes: Yes/No


website/docs/api/router.md
```diff
@@ -247,6 +253,30 @@ sum by (model) (increase(llm_model_cost_total{currency="USD"}[1h]))
 sum by (reason_code) (increase(llm_routing_reason_codes_total[15m]))
 ```
 
+### Error Metrics
```
tao12345666333:
```suggestion
### Request Error Metrics
```

website/docs/api/router.md
```diff
@@ -247,6 +253,30 @@ sum by (model) (increase(llm_model_cost_total{currency="USD"}[1h]))
...
+
+Example PromQL queries:
+
+```prometheus
+# Total errors by reason over the last hour
+sum by (reason) (increase(llm_request_errors_total[1h]))
+
+# Error rate by model over the last 15 minutes
+sum by (model) (increase(llm_request_errors_total[15m])) /
+sum by (model) (increase(llm_model_requests_total[15m]))
```
tao12345666333:
```suggestion
sum by (model) (rate(llm_request_errors_total[15m]))
```

I think we should use this for error rate calculation.

tao12345666333 (changes requested):

website/docs/api/router.md
samzong: Thanks for u suggestion! @tao12345666333 

But I think alone only gives the rate of errors per second. Without the total request volume as a baseline, it doesn’t really reflect the error rate.

```
sum by (model) (rate(llm_request_errors_total[15m]))
```

If we want to capture is: “In the past 15 minutes, what percentage of requests ended up as errors?” For that, we still need the total requests (or their rate) in the same window as the denominator.

maybe use it is a better choice?
```
sum by (model) (rate(llm_request_errors_total[15m])) /
sum by (model) (rate(llm_model_requests_total[15m]))
```

rootfs:
@tao12345666333 Thanks for the review, is it ok to merge?

tao12345666333:
@rootfs  please wait. We haven’t reached an agreement yet. I will write a detailed explanation for this tomorrow.

https://github.com/vllm-project/semantic-router/pull/156#discussion_r2354511005

website/docs/api/router.md
tao12345666333: Thanks for the thoughtful reply, @samzong 


```prometheus
sum by (model) (rate(llm_request_errors_total[15m]))
```

About this one:
- It catches absolute error spikes even when traffic grows (a ratio can look flat while real user impact increases).
- It avoids noisy 100% spikes when traffic is very low (ratios blow up with tiny denominators).
- It’s often what on-call cares about first during incidents: “how many errors/sec are we serving?”

For an SLO-style percentage over the last 15m, I agree we should also show a normalized view. I’d present it alongside using increase() to make the window semantics explicit:

```prometheus
100 * sum by (model) (increase(llm_request_errors_total[15m])) /
      sum by (model) (increase(llm_model_requests_total[15m]))
```

Happy to include both under two headings, e.g. “Error throughput (errors/sec)” and “Error ratio (% of requests failing)”. WDYT?

samzong: @tao12345666333 Thanks for u suggestion.

I’ve included both the errors/sec and percentage queries in the docs, so teams can monitor spikes and ratios side by side. 

Could you take another look and let me know if this matches what you had in mind?

rootfs:
@samzong thanks! 
@tao12345666333 wdyt?

tao12345666333:
We can merge this, thank you

PR #177, opened 2025-09-19
fix: avoid double counting cache hits

cryo-zd:
**What type of PR is this?**

fix: avoid double counting cache hits

Currently, `handleCaching` calls `metrics.RecordCacheHit()` when a cache hit is found, while both `InMemoryCache.FindSimilar` and `MilvusCache.FindSimilar` already record cache hits internally. This causes a single cache hit to be counted twice.
This PR removes the duplicate call in `handleCaching` so that each hit is counted only once.
[InMemoryCache]
https://github.com/vllm-project/semantic-router/blob/fac50b1eb93361e012dcc30b120b1e1614b5a54a/src/semantic-router/pkg/cache/inmemory_cache.go#L294-L304
[Milvus]
https://github.com/vllm-project/semantic-router/blob/fac50b1eb93361e012dcc30b120b1e1614b5a54a/src/semantic-router/pkg/cache/milvus_cache.go#L598-L609

Accurate cache hit metrics are important for monitoring and analysis.

Fixes #

Release Notes: Yes/No

### DIFF
--- a/src/semantic-router/pkg/extproc/request_handler.go
+++ b/src/semantic-router/pkg/extproc/request_handler.go
@@ -263,8 +263,7 @@ func (r *OpenAIRouter) handleCaching(ctx *RequestContext) (*ext_proc.ProcessingR
 		if err != nil {
 			observability.Errorf("Error searching cache: %v", err)
 		} else if found {
-			// Record and log cache hit
-			metrics.RecordCacheHit()
+			// Log cache hit
 			observability.LogEvent("cache_hit", map[string]interface{}{
 				"request_id": ctx.RequestID,
 				"model":      requestModel,


rootfs:
@cryo-zd would it ok if we keep the request handler code but remove the cache hit from inmemory and milvus? So when yet another storage is added, the cache hit is already counted.

cryo-zd:
> @cryo-zd would it ok if we keep the request handler code but remove the cache hit from inmemory and milvus? So when yet another storage is added, the cache hit is already counted.

Ack, keeping the cache hit metric in the request handler and letting each backend only record its own `RecordCacheOperation` makes the design cleaner and more extensible.

One small thought: right now only hits would be counted in the handler, while misses are still counted in each backend. To keep the logic symmetric and easier to maintain, would it make sense to also move `RecordCacheMiss` into the handler (e.g., in the `else` branch)? That way:
- handler is the single place for global hit/miss counters,
- backends focus only on `RecordCacheOperation(...)` for per-backend stats.

https://github.com/vllm-project/semantic-router/blob/aa6c22fd334e6d09311dfacf6a2927c06cb5cdaa/src/semantic-router/pkg/cache/inmemory_cache.go#L276-L278

https://github.com/vllm-project/semantic-router/blob/aa6c22fd334e6d09311dfacf6a2927c06cb5cdaa/src/semantic-router/pkg/cache/inmemory_cache.go#L302-L304

I can update the PR in either direction — do you think this adjustment would be preferable?

rootfs:
I see, that makes sense. I'll merge this one then.

rootfs:
@cryo-zd do you like to come up with a doc PR about how to add yet another cache storage (e.g. chroma, redis)?

Also cc @aeft

PR #178, opened 2025-09-19
refactor: move use_reasoning to the model level from the category level to support non-reasoning models

rootfs:
**What type of PR is this?**

This allows fine grained reasoning control for each model

Fixes #

Release Notes: Yes/No


rootfs:
@yuluo-yx can you review this? thanks

src/semantic-router/pkg/config/config.go
```diff
@@ -336,6 +336,36 @@ func LoadConfig(configPath string) (*RouterConfig, error) {
 	return config, nil
 }
 
+// BoolPtr returns a pointer to a bool value (helper for tests and config)
```
yuluo-yx, line 339: I noticed that this function was only used in the tests. Should it be moved into the tests section instead? The functional code doesn't use it. However, I don't think it's a major issue. It can be improved later.

Xunzhuo (approved):

src/semantic-router/pkg/config/config.go
```diff
@@ -573,6 +610,33 @@ func (c *RouterConfig) SelectBestEndpointAddressForModel(modelName string) (stri
...
+	}
+	return false // Default to false if category or model not found
+}
+
+// GetBestModelForCategory returns the best scoring model for a given category
+func (c *RouterConfig) GetBestModelForCategory(categoryName string) (string, bool) {
+	for _, category := range c.Categories {
+		if category.Name == categoryName {
+			if len(category.ModelScores) > 0 {
+				useReasoning := category.ModelScores[0].UseReasoning != nil && *category.ModelScores[0].UseReasoning
```
Xunzhuo, line 632: why just read the index0?

src/semantic-router/pkg/extproc/reason_mode_selector.go
```diff
@@ -32,16 +32,24 @@ func (r *OpenAIRouter) getReasoningModeAndCategory(query string) (bool, string)
...
 	normalizedCategory := strings.ToLower(strings.TrimSpace(categoryName))
 
-	// Look up the category in the configuration
+	// Look up the category in the configuration and get the best model for it
 	for _, category := range r.Config.Categories {
 		if strings.EqualFold(category.Name, normalizedCategory) {
-			reasoningStatus := "DISABLED"
-			if category.UseReasoning {
-				reasoningStatus = "ENABLED"
+			// Get the best model for this category (first in the list)
```
Xunzhuo, line 38: how to define a best model in category? if it is based on score, maybe we should sort the list and get the biggest one? instead of reading the first index

src/semantic-router/pkg/config/config.go
```diff
@@ -261,14 +261,14 @@ func (c *RouterConfig) GetCacheSimilarityThreshold() float32 {
 
 // Category represents a category for routing queries
 type ModelScore struct {
-	Model string  `yaml:"model"`
-	Score float64 `yaml:"score"`
+	Model        string  `yaml:"model"`
```
Xunzhuo, line 264: use_reasoning: true|false
    reasoning_description: "Why reasoning is needed"
    reasoning_effort: "low|medium|high"

should we move all these three reasoning related vars into model score?

rootfs:
@Xunzhuo thank you for the review. Would it be ok if we have this first and you help refactor the API to be more consistent next? Thanks.

Xunzhuo:
Sure @rootfs

rootfs:
@Xunzhuo great!

PR #188, opened 2025-09-21
fix: fix the timing of precommit image build

yuluo-yx:


yuluo-yx:
hi @rootfs pls review

.github/workflows/precommit-publish.yml
```diff
@@ -2,7 +2,8 @@ name: Create and publish Precommit Image
 
 on:
   push:
-    branches: [ "main" ]
+    paths:
```
rootfs, line 5: this seems to be based on the doc [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore)

yuluo-yx: yep. changed Dockerfile.precommit file, rebuild image.

rootfs (approved):

rootfs:
Thanks for the prompt action @yuluo-yx !

.github/workflows/precommit-publish.yml
yuluo-yx: it is effective

image

PR #204, opened 2025-09-23
feat: add config validation to NewCacheBackend

### LINKED ISSUE #199: feat: Add config validation in `NewCacheBackend`
### Description / Change would like to make:
`NewCacheBackend` currently instantiates backends without validating the given `CacheConfig`, which is error-prone.
https://github.com/vllm-project/semantic-router/blob/7a0221beb090bf4e054fa42a01cafb374e7654cb/src/semantic-router/pkg/cache/cache_factory.go#L10-L23

This issue proposes adding a `ValidateCacheConfig(Config)` call at the start of `NewCacheBackend` . This ensures invalid configs fail fast and keeps backend creation safe and consistent.
https://github.com/vllm-project/semantic-router/blob/7a0221beb090bf4e054fa42a01cafb374e7654cb/src/semantic-router/pkg/cache/cache_factory.go#L62-L79

Would the team agree with introducing this validation step?

cryo-zd:
**What type of PR is this?**

feat: add config validation to NewCacheBackend

This CL added a `ValidateCacheConfig(Config)` call at the start of `NewCacheBackend` . This ensures invalid configs fail fast and keeps backend creation safe and consistent.

Fixes #199 

Release Notes: Yes/No

### DIFF
--- a/src/semantic-router/pkg/cache/cache_factory.go
+++ b/src/semantic-router/pkg/cache/cache_factory.go
@@ -9,6 +9,10 @@ import (
 
 // NewCacheBackend creates a cache backend instance from the provided configuration
 func NewCacheBackend(config CacheConfig) (CacheBackend, error) {
+	if err := ValidateCacheConfig(config); err != nil {
+		return nil, fmt.Errorf("invalid cache config: %w", err)
+	}
+
 	if !config.Enabled {
 		// Create a disabled cache backend
 		observability.Debugf("Cache disabled - creating disabled in-memory cache backend")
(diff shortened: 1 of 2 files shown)


rootfs:
@aeft do you have anything to add?

aeft:
@rootfs currently no, it looks good to me

cryo-zd:
> @aeft do you have anything to add?

I am wondering if we should also validate the `EvictionPolicy` in `CacheConfig` .

Currently, `NewInMemoryCache` treats any unknown value as FIFO.  This is tolerant, but may silently hide misconfigurations. Would it make sense to add a check in `ValidateCacheConfig` that only accepts `fifo/lru/lfu` or an empty string when `config.BackendType==InMemoryCacheType/""`? 
https://github.com/vllm-project/semantic-router/blob/c92a6a595d91d133e3afd77adebf2b5235a0d610/src/semantic-router/pkg/cache/inmemory_cache.go#L42-L54

But is this validation worth adding, 🤔 ?

aeft:
> > @aeft do you have anything to add?
> 
> 
> 
> I am wondering if we should also validate the `EvictionPolicy` in `CacheConfig` .
> 
> 
> 
> Currently, `NewInMemoryCache` treats any unknown value as FIFO.  This is tolerant, but may silently hide misconfigurations. Would it make sense to add a check in `ValidateCacheConfig` that only accepts `fifo/lru/lfu` or an empty string when `config.BackendType==InMemoryCacheType/""`? 
> 
> https://github.com/vllm-project/semantic-router/blob/c92a6a595d91d133e3afd77adebf2b5235a0d610/src/semantic-router/pkg/cache/inmemory_cache.go#L42-L54
> 
> 
> 
> But is this validation worth adding, 🤔 ?

Yes. I think this check makes sense.

PR #223, opened 2025-09-25
feat: validate eviction policy in cache config

cryo-zd:
**What type of PR is this?**

feat: validate eviction policy in cache config
This PR adds validation for `EvictionPolicy` field in `CacheConfig` when `config.BackendType==InMemoryCacheType/""`

[discussions] (https://github.com/vllm-project/semantic-router/pull/204#issuecomment-3329624301)
Currently, any unsupported value is silently treated as FIFO. By explicitly validating accepted value(`lru/lfu/fifo` or empty for FIFO), it ensures misconfigurations fail fast and improves clarity.

Also, cc @aeft

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -107,6 +107,10 @@ const config = {
                 to: '/community/work-groups',
               },
               {
+                label: 'Promotion',
+                to: '/community/promotion',
+              },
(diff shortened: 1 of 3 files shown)


aeft:
Thanks! looks good to me

PR #238, opened 2025-09-26
infra: fix models cache GHA

yuluo-yx:

PR #244, opened 2025-09-27
infra: add GHA restore key

yuluo-yx:
- For https://github.com/vllm-project/semantic-router/pull/238#issuecomment-3339474105


rootfs:
@yuluo-yx is the huggingface model cache key also restored here?

yuluo-yx:
> @yuluo-yx is the huggingface model cache key also restored here?

image

no, the pr submitted yesterday included the restore key for the models.

rootfs:
@yuluo-yx got it, thanks!

Issue #90, opened 2025-09-08
[v0.1] Envoy ExtProc integration for Envoy AI Gateway


rootfs:
# Acceptance: 
- [ ] ExtProc header/body mutation consistent with Envoy AI Gateway filter chains (documented setup for each)  
- [ ] Example Envoy configs for common patterns (e.g., A/B test, canary routing)

rootfs:
@Xunzhuo is it fixed by #245 ?

Xunzhuo:
yes

Issue #89, opened 2025-09-08
[v0.1] Envoy ExtProc integration for GIE


rootfs:
# Acceptance
- [ ] ExtProc header/body mutation consistent with [GIE](https://github.com/kubernetes-sigs/gateway-api-inference-extension) filter chains
- [ ] Example Envoy configs for common patterns (e.g., A/B test, canary routing)

rootfs:
@Xunzhuo is it fixed by #245 ?

Xunzhuo:
yep

srampal:
@Xunzhuo @rootfs I submitted a PR https://github.com/vllm-project/semantic-router/pull/229 
for this on Friday which @Xunzhuo reviewed and now he has submitted a PR for the same issue after that ? That PR is on GIE and LLM-D integration using Istio as the GIE gateway.

srampal:
There is an additional piece to this for Istio based GIE and that is getting addressed by PR https://github.com/vllm-project/semantic-router/pull/229

PR #245, opened 2025-09-27
feat: support kubernetes environment

Xunzhuo:
**What type of PR is this?**

feat: support kubernetes environment

This PR added kubernetes support, as well as support in Kind.

Release Notes: Yes


deploy/kubernetes/ai-gateway/configuration/config.yaml
```diff
@@ -0,0 +1,72 @@
+# Copyright Envoy AI Gateway Authors
```
rootfs: do we need to keep this license or we have dual license with envoy and vllm?

Xunzhuo: no I will remove this : )

rootfs:
ref: #89 #90

Xunzhuo:
Blocked by https://github.com/envoyproxy/ai-gateway/pull/1239

deploy/kubernetes/ai-gateway/inference-pool/inference-pool.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: inference.networking.x-k8s.io/v1alpha2
+kind: InferencePool
```
srampal, line 2: This use of InferencePool does not seem to match with the Gateway API Inference SIG's design intent for InferencePool. I do see some reasoning why you chose to do this.  However this can potentially cause issues down the road. Is there a design document where you provide some background for this design choice ?

PR #248, opened 2025-09-27
perf: optimize FindSimilarTools by early pruning

cryo-zd:
**What type of PR is this?**

perf: optimize FindSimilarTools by early pruning

Instead of calculating similarity for all `ToolEntry` items and then sorting the entire list, we now apply the similarity threshold filter before appending results to the temporary slice, which can reduce the in-memory sorting overhead.


rootfs (approved):
@cryo-zd thanks, do you mind following up with a unit test too?

cryo-zd:
> @cryo-zd thanks, do you mind following up with a unit test too?

My pleasure. I will work on it as soon as possible in the following days.

Issue #253, opened 2025-09-27
Envoy ext_proc 500 error when both `value` and `raw_value` are set in HeaderValue


ztang2370:
**Description:**
Envoy immediately returns HTTP 500 when an external processor sends a `HeaderMutation` with both `value` and `raw_value` set for the same header. Logs show:

```
Only one of value or raw_value in the HeaderValue message should be set
```

**Steps to Reproduce:**

1. Configure Envoy with `ext_proc` filter.
2. External processor sets a header like:

```go
&core.HeaderValue{
    Key:      "x-selected-model",
    Value:    actualModel,
    RawValue: []byte(actualModel),
}
```

3. Send any HTTP request through Envoy.
4. Observe a 500 response.

**Expected Behavior:**
Envoy should not fail the request; the header should be applied correctly using either `value` or `raw_value`.

**Fix / Workaround:**
Set only one of `value` or `raw_value`. For example, remove the `Value` field:

```go
&core.HeaderValue{
    Key:      "x-selected-model",
    RawValue: []byte(actualModel),
}
```

This resolves the immediate 500 error. Reference: [request_handler.go#L441](https://github.com/vllm-project/semantic-router/blob/9f90230047e53492ff024b96fe8e697a34d45be1/src/semantic-router/pkg/extproc/request_handler.go#L441)

**Additional Context:**

* Happens with routing headers in semantic-router external processor.
* Envoy version: v1.31.7.
* Running under Docker Compose.

rootfs:
@ztang2370 PR is welcome!

PR #255, opened 2025-09-28
Fix: Envoy ext_proc 500 error when both value and raw_value are set in HeaderValue

ztang2370:
**What type of PR is this?**

Bug fix

Fix envoy ext_proc 500 error when both value and raw_value are set in HeaderValue

Fixes https://github.com/vllm-project/semantic-router/issues/253

Release Notes: No


rootfs:
@ztang2370 thank you for contributing! Can you follow up with a unit test too? That'll help for future refactoring.

PR #270, opened 2025-09-28
Bench: Add more dataset in router evaluation

rootfs:
**What type of PR is this?**

Add new dataset implementations: AQUA-RAT, DROP, GSM8K, MATH, OpenBookQA, SciQ, StrategyQA

Fixes #

Release Notes: Yes/No


yuezhu1 (commented):
Do we want to let users to input which dataset subset (test? valid?) they want to use? As some of the datasets may not have test subsets. Also instead of giving sample count per dataset, do we want to support the percentage base sampling over the datasets?

rootfs:
> Do we want to let users to input which dataset subset (test? valid?) they want to use? As some of the datasets may not have test subsets. Also instead of giving sample count per dataset, do we want to support the percentage base sampling over the datasets?

you are right, at the moment, not all the datasets have test/train/validate splits. I will follow up with such split next.

rootfs:
merging it for now so we can use these dataset for testing

PR #274, opened 2025-09-29
chore: move all dockerfile into tools

Xunzhuo:
**What type of PR is this?**

chore: move all dockerfile into tools

Release Notes: Yes


JaredforReal:
There are some configs for Grafana and Prometheus for Docker Compose Path in `config/prometheus.yaml` and `config/grafana/`, since Observability config will be put into `deploy/kubernetes/observability/`(said in #279 ), maybe you can pack the configs in `config/` into `tools/docker/` too. @Xunzhuo

PR #186, opened 2025-09-21
feat: add v1/models endpoint

JaredforReal:
**What type of PR is this?**
Feat: add OpenAI Compatible API endpoint /v1/models

**What this PR does / why we need it**:

Add v1/models endpoint in src/semantic-router/pkg/server.go
Add unit test in src/semantic-router/pkg/server_test.go
Add Model Endpoint section in OpenAI API Compatibility of docs/api/router.md
Add Endpoints-to-Ports Mappint in docs/api/router.md and classification.md

**Which issue(s) this PR fixes**:

Fixes #185

PR #283, opened 2025-09-29
feat: support /v1/models in direct response

Xunzhuo:
**What type of PR is this?**

feat: support /v1/models in direct response

### DIFF
--- a/src/semantic-router/pkg/extproc/models_endpoint_test.go
+++ b/src/semantic-router/pkg/extproc/models_endpoint_test.go
@@ -1,0 +1,236 @@
+package extproc
+
+import (
+	"encoding/json"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


rootfs:
@Xunzhuo can you review this solution too?
https://github.com/vllm-project/semantic-router/pull/186#issuecomment-3344471731

Xunzhuo:
This need additional xDS config generation which is not standard for InferencePool implementation, so that using direct response. Otherwise we need every Gateway implementation to inject one more cluster to support our models api.

rootfs:
@Xunzhuo sounds good. Can you run precommit? It is ready to go.
cc @JaredforReal

JaredforReal:
Sounds good to me

Xunzhuo:
tested and move this forward

Issue #285, opened 2025-09-29
[CI] Broken Test Pipeline


rootfs:
**Describe the bug**
The CI is currently broken due to missing HF pkg import
https://github.com/vllm-project/semantic-router/actions/runs/18093944987/job/51480739334?pr=284

cooktheryan:
can this be closed @rootfs ?

PR #287, opened 2025-09-29
fix: force install hf_transfer to avoid missing pkg

rootfs:
**What type of PR is this?**

For some reason, `hf_transfer` is not installed in `huggingface_hub`. This fix force installing this pkg.

Fixes #

Release Notes: Yes/No


rootfs:
fix #285

rootfs:
merging it now, the broken CI step is fixed.

PR #288, opened 2025-09-29
docs: add mermaid modal

### LINKED ISSUE #263: docs: optimize the display effect of mermaid image
Such as: https://vllm-semantic-router.com/docs/overview/architecture/system-architecture#high-level-architecture-overview

When I want to look at the mermaid image carefully, the font is very small, which affects the reading experience. When I want to zoom in to see it more clearly, I can only zoom in the browser page. emmm~

When clicking on the mermaid image, a modal box is displayed to zoom in. This is the method I have thought of so far.

@rootfs @Xunzhuo WDYT?

yuluo-yx:
- Fixed #263

### DIFF
--- a/website/src/components/ZoomableMermaid/index.js
+++ b/website/src/components/ZoomableMermaid/index.js
@@ -187,11 +187,11 @@ const ZoomableMermaid = ({ children, title, defaultZoom = 1.2 }) => {
         >
           <div
             className={styles.diagramContainer}
-            style={{ 
-              transform: `scale(${zoomLevel})`, 
+            style={{
+              transform: `scale(${zoomLevel})`,
               // Ensure scaling is from the center of the diagram.
               // Fix the issue where the top scroll bar is not visible when the chart is enlarged.
-              transformOrigin: 'center top'
+              transformOrigin: 'center top',
             }}
           >
             <Mermaid value={children} />


website/docs/overview/architecture/system-architecture.md
```diff
@@ -4,8 +4,10 @@ The Semantic Router implements a sophisticated Mixture-of-Models (MoM) architect
 
 ## High-Level Architecture Overview
 
-```mermaid
-graph TB
+import ZoomableMermaid from '@site/src/components/ZoomableMermaid';
```
rootfs, line 7: how does it work? i don't see any difference between the preview and existing diagrams

website/src/components/ZoomableMermaid/index.js
```diff
@@ -0,0 +1,230 @@
...
+          <div className={styles.modalControls}>
+            <span className={styles.zoomIndicator}>
+              {Math.round(zoomLevel * 100)}
+              %
+            </span>
+            <button
+              className={styles.zoomButton}
+              onClick={zoomOut}
+              disabled={zoomLevel <= 0.5}
+              aria-label="缩小图表"
```
rootfs: can you switch to English?

website/docs/overview/architecture/system-architecture.md
yuluo-yx: You can place the mouse cursor on "mermaid", like this

before:

image

now:

image

then: 
image

website/src/components/ZoomableMermaid/index.js
yuluo-yx: got it, I am currently working on this.

website/docs/overview/architecture/system-architecture.md
rootfs: oooh, i was on the wrong page :D 
Now i see how this works, this is cool!

rootfs: btw, the zoom in (+) button greyed out 
image

website/src/components/ZoomableMermaid/index.js
yuluo-yx: > got it, I am currently working on this.

done

rootfs (dismissed):

yuluo-yx:
@rootfs To be honest, this feature is a bit complicated. The principle is to use React Portal along with some CSS. At present, I have only adjusted mermaid in the system-architecture. I want to recreate a PR and replace all the Mermaid components in the document website with the ZoomableMermaid component. Currently, pr only provides component implementation. WDYT?

rootfs:
@yuluo-yx good to me, @Xunzhuo what do you think?

website/docs/overview/architecture/system-architecture.md
yuluo-yx: The current maximum zoom limit for the ZoomableMermaid component is 300. So when it is turned to the maximum level, it turns grey. I carefully examined the mermaid in Docusaurus. It transformed mermaid into SVG format for display on the web page. Therefore, it can be magnified without any loss. Because it is an SVG that has been rendered, it is impossible to determine exactly what size is the most suitable. Some complex mermaid 300 images are already very clear, while some 300 images are still blurry. I think 500 should be a good magnification factor. I think the best magnification factor should be specified when writing the article for better readability. By the way, the ZoomableMermaid component provides a default magnification factor attribute.

`<ZoomableMermaid title="System Architecture Overview" defaultZoom={5.5}>`

This is also the reason why I want to uniformly adjust them in the next PR. They need to be adjusted one by one to find the most suitable magnification. We cannot use a single standard to measure everything.

rootfs:
Let's have this one. @Xunzhuo if this is a good start, @yuluo-yx can you follow up with more enhancements?

yuluo-yx:
> Let's have this one. @Xunzhuo if this is a good start, @yuluo-yx can you follow up with more enhancements?

no problem.

PR #292, opened 2025-09-29
feat: implement comprehensive ExtProc testing with cache bypass

yossiovadia:
01-envoy-extproc-test.py with ExtProc-specific functionality tests and implements cache bypass using unique UUIDs.

Test Coverage:

1. test_request_headers_propagation

What: Tests that custom headers flow correctly through the ExtProc How:
- Sends request with custom headers: X-Test-Trace-ID, X-Original-Model
- Verifies ExtProc doesn't break header handling
- Checks response contains proper Content-Type and model fields ExtProc Value: Ensures headers aren't corrupted during ExtProc processing

2. test_extproc_body_modification

What: Tests that ExtProc can inspect/modify request and response bodies How:
- Sends request with custom field: "test_field": "should_be_preserved"
- Uses header X-Test-Body-Modification: true to signal ExtProc
- Verifies response is valid and processing succeeded ExtProc Value: Confirms ExtProc can access and potentially transform request/response data

3. test_extproc_error_handling

What: Tests ExtProc resilience against malformed/unusual requests How:
- Sends problematic headers: very long headers (1000 chars), special characters
- Uses headers like X-Test-Error-Recovery: true
- Expects graceful handling (no crashes/hangs)
- Accepts either success OR protective disconnection ExtProc Value: Ensures ExtProc acts as protective filter, doesn't crash on bad input

4. test_extproc_performance_impact

What: Tests that ExtProc doesn't add excessive latency How:
- Measures end-to-end response time with ExtProc processing
- Uses performance-specific headers: X-Test-Performance: true
- Validates response time < 30 seconds (reasonable threshold)
- Checks request succeeds without timeout ExtProc Value: Confirms ExtProc doesn't bottleneck the request pipeline

Release Notes: No

### DIFF
--- a/e2e-tests/01-envoy-extproc-test.py
+++ b/e2e-tests/01-envoy-extproc-test.py
@@ -10,18 +10,18 @@ These tests use custom headers to trace request processing.
 import json
 import os
 import sys
+import unittest
 import uuid
 
 import requests
 
-# Add parent directory to path to allow importing common test utilities
-sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-from tests.test_base import SemanticRouterTestBase
+# Import test base from same directory
+from test_base import SemanticRouterTestBase
 
 # Constants
 ENVOY_URL = "http://localhost:8801"
 OPENAI_ENDPOINT = "/v1/chat/completions"
-DEFAULT_MODEL = "qwen2.5:32b"  # Changed from gemma3:27b to match make test-prompt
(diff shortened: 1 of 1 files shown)


rootfs:
@yossiovadia can you add a test case to test semantic cache?

yossiovadia:
> @yossiovadia can you add a test case to test semantic cache?

on my list ( along with PII, Jail break etc ) about 10 in total.

PR #299, opened 2025-10-01
docs: use ts replace js in docs website

### LINKED ISSUE #268: docs: use ts replace js
The document website uses JavaScript. Let's replace it with TypeScript. More modern and type-safe

WDYT? @Xunzhuo

yuluo-yx:
- Fixed #268 

- Use typeScript replace javaScript
- Replace the function definition with the arrow function of ES6
- Translate zh comments


Xunzhuo:
Thanks, I wonder what is the direct benefit to move js to ts? Like any performance or UX improvements?

yuluo-yx:
> Thanks, I wonder what is the direct benefit to move js to ts? Like any performance or UX improvements?

ts is suitable for object-oriented programming. Compared to js, it offers type safety, static type checking and compilation optimization, which improves performance. 
And nowadays, more and more projects are being developed using TypeScript rather than JavaScript. For example, https://github.com/facebook/docusaurus

PR #342, opened 2025-10-04
update doc: install in local

FeiDaLI:
## What type of PR is this?
This PR updates the installation documentation to include instructions for running the application locally without vLLM, using a built-in mock server. It adds a __main__ entry point to app.py, enabling the FastAPI application to be launched directly via python app.py. 

## What this PR does / why we need it:
This change simplifies local development and testing by eliminating the need for Docker or a full vLLM environment.


website/docs/installation/installation.md
```diff
@@ -193,6 +193,23 @@ curl -X POST http://localhost:8801/v1/chat/completions \
   }'
 ```
 
+> **Tip:** Optional: Run Without vLLM (Using Mock Server)
+>
+> No vLLM? Use the built-in mock server for testing:
+> 
+> pip install fastapi uvicorn
+> python tools/mock-vllm/app.py
```
rootfs, line 201: there is also a lightweight llm-katan, can you add another instruction for this one too?

ABC12345anouys: OK,I will do add that soon.

PR #346, opened 2025-10-04
docs: add container connectivity docs & add vsr-headers to sidebar

JaredforReal:
**What type of PR is this?**
docs: add container connectivity docs & add vsr-headers to sidebar

Add documentation: Container Connectivity Troubleshooting based on practices.
add vsr-headers to `sidebar.ts`

### DIFF
--- a/website/docs/troubleshooting/container-connectivity.md
+++ b/website/docs/troubleshooting/container-connectivity.md
@@ -1,0 +1,190 @@
+---
+title: Container Connectivity Troubleshooting
+sidebar_label: Container Connectivity
+---
+
+This guide summarizes common connectivity issues we hit when running the router with Docker Compose or Kubernetes and how we fixed them. It also covers the “No data” problem in Grafana and how to validate the full metrics chain.
+
+## 1. Use IPv4 addresses for backend endpoints
+
+Symptoms
+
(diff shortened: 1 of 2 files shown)


website/docs/troubleshooting/container-connectivity.md
```diff
@@ -0,0 +1,190 @@
+---
+title: Container Connectivity Troubleshooting
+sidebar_label: Container Connectivity
+---
+
+This guide summarizes common connectivity issues we hit when running the router with Docker Compose or Kubernetes and how we fixed them. It also covers the “No data” problem in Grafana and how to validate the full metrics chain.
+
+## 1. Use IPv4 addresses for backend endpoints
```
rootfs, line 8: can you check if similar kubernetes networking diagnostics help?
https://goteleport.com/blog/troubleshooting-kubernetes-networking/

JaredforReal: Thanks for sharing the link! I’ve checked it, and while it might be helpful, it doesn’t fully cover the issues I ran into on my lab servers. This PR is based on those practical troubleshooting experiences. Do you think it would make sense to keep both references, so the docs can complement each other?

rootfs: Sure, let's keep your inline diagnostics

Issue #338, opened 2025-10-04
Feature Request: Dynamic CPU/GPU Switching


FeiDaLI:
## Is your feature request related to a problem? Please describe.

Currently,  BERT processing typically runs on a fixed device (CPU or GPU), without dynamic adaptation based on query complexity. This leads to suboptimal resource utilization:
- **Simple queries** processed on GPU waste valuable GPU resources.
- **Complex queries** processed on CPU suffer from slow inference times.
- There is no mechanism to **dynamically switch** between CPU and GPU based on actual computational demands.

###  Automatic CPU/GPU Switching
Implement a **dynamic resource manager** that:
- Estimates the computational complexity of incoming queries.
- Automatically routes simple queries to CPU and complex ones to GPU.
- Balances latency, throughput, and hardware utilization.

This s could leverage profiling metrics such as token length, syntactic complexity, or historical processing times to make real-time decisions. 

To reduce GPU transfer overhead and improve throughput, introduce a **queue-based batching** inspired by **continuous batching** in vLLM:
- GPU-bound queries are batched efficiently.
- Data transfers and computation are overlapped where possible.
- CPU and GPU workloads are decoupled via a shared batch queue.

Example design:

```go
type ResourceManager struct {
    CPUProcessor  *CPUProcessor
    GPUProcessor  *GPUProcessor
    Profiler      *ComputeProfiler
    Queue         *BatchQueue
}

func (rm *ResourceManager) ProcessQuery(query string) (*ClassificationResult, error) {
    // 1. Estimate query complexity
    complexity := rm.Profiler.EstimateComplexity(query)
    
    // 2. Route based on compute bounds
    if complexity.IsCPUBound() {
        return rm.CPUProcessor.Process(query)
    } else if complexity.IsGPUBound() {
        return rm.Queue.AddToBatch(query) // Enqueue for GPU batch processing
    }
    
    // Fallback to CPU
    return rm.CPUProcessor.Process(query)
}
```

#### Static Profiling at Startup: 
Run a benchmark with sample data during initialization to calibrate CPU compute-bound vs. I/O-bound thresholds based on system configuration.

## Additional context
If this feature aligns with the project's roadmap, would it be possible to assign this issue to me? I’d appreciate the opportunity to start working on it.

ABC12345anouys:
I ran tests on a standard laptop CPU (AMD Ryzen 5 5500U) to see how the input length affects inference time. The test covered input lengths from 50 to 10,000 tokens.

[...]

PR #350, opened 2025-10-05
bench: add script to measure query length impact on response time

FeiDaLI:
**What type of PR is this?**
add script to measure query length impact on response time

Adds a test script that sends HTTP requests with incrementally increasing content lengths to the Semantic Router and measures response times. This helps analyze how query length affects inference performance and latency, and can be integrated with metrics collection systems for deeper performance analysis https://github.com/vllm-project/semantic-router/issues/338.

### DIFF
--- a/bench/bench_query_length.sh
+++ b/bench/bench_query_length.sh
@@ -1,0 +1,230 @@
+#!/bin/bash
+
+
+# Purpose:
+#   This script sends HTTP requests with different content lengths to the 
+#   Semantic Router (Envoy proxy) and measures the response times. It's designed
+#   to help analyze how query length affects inference performance and latency.
+#
+# Usage:
+#   The script sends serial curl requests with incrementally increasing content
+#   lengths (configurable start, end, and step size) and records:
+#   - Request character count
+#   - Response time (client send to response received)
+#   - HTTP status codes
+#   - Full response content
+#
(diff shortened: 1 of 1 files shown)


bench/bench_query_length.sh
```diff
@@ -0,0 +1,230 @@
...
+            fi
+            break
+        fi
+    done
+    
+    echo "$content"
+}
+
+# Create result log file
+log_file="test_results_$(date +%Y%m%d_%H%M%S).log"
```
rootfs: if saving to the current directory, the file can be added to gitignore

ABC12345anouys: To avoid saving the file locally, I directly output the results to the shell for easy viewing.

ABC12345anouys: sorry i forget delete the original code

bench/bench_query_length.sh
```diff
@@ -0,0 +1,392 @@
...
+            if [ $remaining -gt 0 ]; then
+                content="$content${word:0:$remaining}"
+            fi
+            break
+        fi
+    done
+    
+    echo "$content"
+}
+#!/bin/bash
```
rootfs: this is a dup

rootfs (approved):

PR #372, opened 2025-10-08
fix: keep memory cache metrics accurate

cryo-zd:
### What type of PR is this?

fix: keep memory cache metrics accurate

### What this PR does / why we need it:
**Context:**
`func (c *InMemoryCache) UpdateWithResponse` runs `c.cleanupExpiredEntries()`, which can silently drop TTL-expired cache entries, but doesn't call `metrics.UpdateCacheEntries("memory", len(c.entries))`) to refresh `CacheEntriesTotal` afterword - leaving the gauge stale until another write(`func (c *InMemoryCache) AddEntry`/`func (c *InMemoryCache) AddPendingRequest`)  happened.
**Fix:**
- Centralized the cleanup so it now logs TTL removals and updates the gauge immediately, keeping metrics in sync whenever expired entries are purged.
- Made `Close` explicitly zero the gauge after cleaning memory so shutdown reports the correct size.

https://github.com/vllm-project/semantic-router/blob/565c7abd3c3a0cd326e124266c752fe13fbe2df3/src/semantic-router/pkg/cache/inmemory_cache.go#L120-L153
https://github.com/vllm-project/semantic-router/blob/565c7abd3c3a0cd326e124266c752fe13fbe2df3/src/semantic-router/pkg/metrics/metrics.go#L511-L514

Fixes #

Release Notes: Yes/No

### DIFF
--- a/website/docs/proposals/nvidia-dynamo-integration.md
+++ b/website/docs/proposals/nvidia-dynamo-integration.md
@@ -1,0 +1,1694 @@
+# Semantic Intelligence Layer for NVIDIA Dynamo
+
+## 1. Executive Summary
+
+This proposal outlines a comprehensive integration strategy between **vLLM Semantic Router** and **NVIDIA Dynamo**, combining semantic intelligence with high-performance distributed inference. The integration creates a unified inference stack that leverages:
+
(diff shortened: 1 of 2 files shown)


rootfs:
@cryo-zd can you add a unit test for this case?

cryo-zd:
> @cryo-zd can you add a unit test for this case?

My oversight, I will add a test for it.

PR #407, opened 2025-10-12
feat: k8s support and some fixes

JaredforReal:
- add k8s support
- update README of dashboard
- delete network config 
- add JaredforReal to Owner

### DIFF
--- a/dashboard/frontend/src/App.tsx
+++ b/dashboard/frontend/src/App.tsx
@@ -5,9 +5,11 @@ import LandingPage from './pages/LandingPage'
 import MonitoringPage from './pages/MonitoringPage'
 import ConfigPage from './pages/ConfigPage'
 import PlaygroundPage from './pages/PlaygroundPage'
+import { ConfigSection } from './components/ConfigNav'
 
 const App: React.FC = () => {
   const [isInIframe, setIsInIframe] = useState(false)
+  const [configSection, setConfigSection] = useState<ConfigSection>('models')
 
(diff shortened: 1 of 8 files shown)


dashboard/OWNER
```diff
@@ -0,0 +1,2 @@
+# Dashboard directory Owners
+@JaredforReal
```
Xunzhuo: plz add me to this module as well

JaredforReal: done!

Dockerfile.extproc
```diff
@@ -57,17 +57,11 @@ FROM golang:1.24 AS go-builder
...
-ENV GOPROXY=https://goproxy.cn,direct
-ENV GOSUMDB=sum.golang.google.cn
-
 # Copy Go module files first for better layer caching
 RUN mkdir -p src/semantic-router
 COPY src/semantic-router/go.mod src/semantic-router/go.sum src/semantic-router/
 COPY candle-binding/go.mod candle-binding/semantic-router.go candle-binding/
 
-RUN cd src/semantic-router && go mod download && \
-    cd /app/candle-binding && go mod download
```
Xunzhuo, line 69: why remove this

JaredforReal: I added these two lines to make it fail faster when mirrors are unreachable, saving time. We can keep it to make go build fail faster whenever network crush, and make download a bit faster in some scenarios.

PR #410, opened 2025-10-13
fix(make): mark model downloads with .downloaded sentinel (#309)

### LINKED ISSUE #309: make download: Download the incomplete local model file
**Describe the bug**
When I started the semantic-router service, an error occurred saying that sentence-transformers/all-MiniLM-L12-v2 was missing. Actually, I have already finished making download-models. After the "make download" is completed, the model files in models indeed do not report the model files that were required to be initialized.
**To Reproduce**
1.make download
2.docker compose up -d

**Expected behavior**

Image

Image

**Screenshots**
make download should fully download all the model files necessary for starting the container and should not be missing them

**Additional context**

samzong:
**What type of PR is this?**

fix(make): mark model downloads with .downloaded sentinel

ensure model downloads check for completion marker file to avoid redundant downloads

Fixes #309

Release Notes: Yes


tools/make/models.mk
```diff
@@ -24,59 +24,59 @@ download-models:
 
 download-models-minimal:
 	@mkdir -p models
-	@if [ ! -d "models/category_classifier_modernbert-base_model" ]; then \
-		hf download LLM-Semantic-Router/category_classifier_modernbert-base_model --local-dir models/category_classifier_modernbert-base_model; \
+	@if [ ! -f "models/category_classifier_modernbert-base_model/.ok" ] || [ ! -d "models/category_classifier_modernbert-base_model" ]; then \
+		hf download LLM-Semantic-Router/category_classifier_modernbert-base_model --local-dir models/category_classifier_modernbert-base_model && touch models/category_classifier_modernbert-base_model/.ok; \
```
rootfs: can you add something more meaningful e.g. the HF repo revision or timestamp of download?

samzong: great! I will modify it after #309

samzong: @rootfs It's changed.

PR #413, opened 2025-10-13
refactor(config): remove models field from vLLM endpoints

Xunzhuo:
Remove the redundant models field from VLLMEndpoint configuration. Model-to-endpoint mapping is now solely determined by the preferred_endpoints field in model_config, eliminating the need for bidirectional association.

Changes:
- Remove Models field from VLLMEndpoint struct
- Update GetEndpointsForModel to use only preferred_endpoints
- Update GetAllModels to retrieve models from model_config keys
- Update all configuration files to remove models field
- Update all tests to reflect the new configuration structure
- Update TypeScript interface in dashboard frontend

This simplifies the configuration and removes potential inconsistencies between models and preferred_endpoints.

### DIFF
--- a/website/docs/installation/configuration.md
+++ b/website/docs/installation/configuration.md
@@ -159,10 +159,13 @@ Configure your LLM servers:
 vllm_endpoints:
   - name: "my_endpoint"
     address: "127.0.0.1"  # Your server IP - MUST be IP address format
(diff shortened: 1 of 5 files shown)


rootfs:
cc @yossiovadia for e2e tests

PR #415, opened 2025-10-13
ut: add test for tls file

yuluo-yx:


rootfs:
@yuluo-yx this looks a generic tls unit test? is there any specific function or pkg we are using that needs to be tested?

yuluo-yx:
> @yuluo-yx this looks a generic tls unit test? is there any specific function or pkg we are using that needs to be tested?

There are no functions that require this test. This is merely a unit test to ensure the normal operation of the TLS code.

rootfs:
can you customize the network tests to be more focused on the functions created and used by semantic router?

yuluo-yx:
> can you customize the network tests to be more focused on the functions created and used by semantic router?

Well, this requires more comprehensive e2e cases. I will improve this function in the new PR.

PR #416, opened 2025-10-13
add ALERT dataset for jailbreak and filtering wiki training

joyful-ii-V-I:
**What type of PR is this?**
improve training data will ALERT training set and prevent wiki api from recursing more than 4 deep, this fixes the more obvious jailbreak criminal type activities

Needs better training data for model

Release Notes: No


rootfs:
@joyful-ii-V-I thanks for the contribution. can you move these new files to a folder of their own?

rootfs:
@joyful-ii-V-I can you fix the pre-commit errors?

PR #421, opened 2025-10-14
chore(e2e): remove legacy mock/real vLLM test modes and Makefile targets

### LINKED ISSUE #420: Drop Legacy vLLM E2E Targets and Flags
The project still advertises legacy mock/real vLLM workflows (Make targets and run_all_tests.py flags), but the supporting scripts no longer exist. Following those instructions fails because only the LLM Katan path is maintained.

**To Reproduce**
Run make help and note the `test-e2e-mock/test-e2e-real` targets.
Execute make test-e2e-mock (or make test-e2e-real).
Observe the command fails: run_all_tests.py no longer accepts `--mock/--real`, and the referenced start/stop scripts are missing.
The developer workflow stalls because the advertised modes cannot be used.

**Expected behavior**
Either the Make targets and CLI flags should work end to end, or they should be removed so only the supported LLM Katan workflow is documented.

**Screenshots**
N/A

**Additional context**
Cleaning up these stale targets/flags would avoid confusion for new contributors and align the tools with current documentation.

samzong:
**What type of PR is this?**

chore: remove legacy mock/real vLLM test modes and Makefile 

Drop the `--mock / --real` flags from e2e-tests/run_all_tests.py and remove
legacy start/stop and test targets for mock/real vLLM servers from the
tools/make/build-run-test.mk. Simplifies the e2e workflow to rely on the
LLM Katan flow and the manual start/run pattern.

Fixes #420

Release Notes: Yes

### DIFF
--- a/e2e-tests/run_all_tests.py
+++ b/e2e-tests/run_all_tests.py
@@ -89,12 +89,6 @@ def main():
     )
     parser.add_argument("--pattern", default="*.py", help="Test file pattern to run")
     parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
-    parser.add_argument(
-        "--mock", action="store_true", help="Running with mock vLLM servers"
-    )
-    parser.add_argument(
-        "--real", action="store_true", help="Running with real vLLM servers"
-    )
     args = parser.parse_args()
 
     # Get the directory where this script is located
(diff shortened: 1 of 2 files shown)


rootfs:
@yossiovadia PTAL, thanks

rootfs (approved):

PR #423, opened 2025-10-14
fix: stop returning expired in-memory cache hits

### LINKED ISSUE #404: TTL expiry ignored when in-memory cache hits
### Summary/Change would like to make:
When `ttl_seconds` is set, entries whose `LassAccessAt` has already crossed the TTL can still win `FindSimilar` and be returned if they remain the top match and no purge happens. It violates the spec in [documentation](https://vllm-semantic-router.com/docs/tutorials/semantic-cache/in-memory-cache#configuration-options).

- steps:
 1. Enable the in-memory cache with `ttl_seconds` > 0
 2. Add an entry, access it once, then wait past the TTL without mutating the cache
 3. Call `FinaSimilar` with a query that best matches that expired entry ---> it's returned  and logged as a hit
```Go
It("should skip expired entries during similarity search", func() {
			ttlCache := cache.NewInMemoryCache(cache.InMemoryCacheOptions{
				Enabled:             true,
				SimilarityThreshold: 0.1,
				MaxEntries:          10,
				TTLSeconds:          1,
			})
			defer ttlCache.Close()

			err := ttlCache.AddEntry("ttl-request-id", "ttl-model", "time-sensitive query", []byte("request"), []byte("response"))
			Expect(err).NotTo(HaveOccurred())

			time.Sleep(1100 * time.Millisecond)

			response, found, err := ttlCache.FindSimilar("ttl-model", "time-sensitive query")
			Expect(err).NotTo(HaveOccurred())
			Expect(found).To(BeFalse())
			Expect(response).To(BeNil())
		})
```
- Actual: The stale response is served, `HitCount` increments, and `LastAccessAt` gets refreshed via `updateAccessInfo()`, effectively reviving "brought back to life"
- suggested direction: Fold the TTL checking into the main similarity loop so we skip expired candidates, this eliminates `cleanupExpiredEntriesReadOnly()` to keeps the read path single-pass but trades off some readability. Alternatively, keep the helper for clarity and add an extra TTL guard inside the loop (slightly slower).

Would the team agree with introducing such checking step ? If it is, guidance from maintainers on the preferred trade-off would be helpful before submitting a PR.

cryo-zd:
**What type of PR is this?**

fix: stop returning expired in-memory cache hits

This PR skips TTL-expired entries while scanning for the best match so `FindSimilar` doesn't serves stale responses.

Fixes #404 

Release Notes: Yes/No


rootfs:
@aeft PTAL, thanks

aeft:
Thanks! It looks good to me. cc @rootfs

PR #426, opened 2025-10-14
feat: make llm-katan as default in docker compose up

JaredforReal:
Adds a lightweight local OpenAI-compatible model (llm-katan) to the default Docker Compose stack for a fully self-contained out‑of‑the‑box experience.

Key Changes:

- New llm-katan service (Qwen/Qwen3-0.6B served as qwen3) with static IP `172.28.0.20:8002`.
- Router config: set `qwen3` as the default model.
- Added pre-download of Qwen3-0.6B to model bootstrap targets.
- New make target: `docker-compose-up-core` (stack without local model).
- Updated `README` with usage.
- Lint config: exclude `models/` from markdown lint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,20 @@ To view the ***Chain-Of-Thought*** of the vLLM-SR's decision-making process, we 
 
 ## Quick Start 🚀
 
+Get up and running in seconds with our interactive setup script:
+
 ```bash
-make docker-compose-up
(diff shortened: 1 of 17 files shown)


rootfs:
lgtm, once the markdown link is fixed, it is good to go

cc @yossiovadia

Xunzhuo (approved):

PR #431, opened 2025-10-14
fix: update response model field to match routing decision

yossiovadia:
Ran tests with and without the fix using the same configuration to prove effectiveness: 

From testing session on 2025-10-14 at 14:32:
{"msg":"routing_decision","selected_model":"Model-A","category":"economics","selected_endpoint":"127.0.0.1:8000"}
{"msg":"routing_decision","selected_model":"Model-B","category":"psychology","selected_endpoint":"127.0.0.1:8001"}
Router correctly routes economics to Model-A (endpoint 127.0.0.1:8000)
Router correctly routes psychology to Model-B (endpoint 127.0.0.1:8001)
BUT no log line showing "Updating response model field" (fix not present)
API Response Evidence:
Economics query returned: "model": "Model-B" ❌ (WRONG - should be Model-A)
Psychology query returned: "model": "Model-A" ❌ (WRONG - should be Model-B)

With the fix:

{"msg":"routing_decision","selected_model":"Model-A","category":"economics","selected_endpoint":"127.0.0.1:8000"}
{"msg":"Updating response model field from 'Model-B' to 'Model-A' (routing decision)"}

{"msg":"routing_decision","selected_model":"Model-B","category":"psychology","selected_endpoint":"127.0.0.1:8001"}
{"msg":"Updating response model field from 'Model-A' to 'Model-B' (routing decision)"}
API Response Evidence:
Economics query returned: "model": "Model-A" ✅ (CORRECT)
Psychology query returned: "model": "Model-B" ✅ (CORRECT)

### DIFF
--- a/src/semantic-router/pkg/cache/cache_test.go
+++ b/src/semantic-router/pkg/cache/cache_test.go
@@ -576,6 +576,30 @@ development:
 			Expect(stats.HitRatio).To(Equal(0.5))
 		})
 
+		It("should skip expired entries during similarity search", func() {
+			ttlCache := cache.NewInMemoryCache(cache.InMemoryCacheOptions{
+				Enabled:             true,
+				SimilarityThreshold: 0.1,
+				MaxEntries:          10,
+				TTLSeconds:          1,
+			})
+			defer ttlCache.Close()
+
+			err := ttlCache.AddEntry("ttl-request-id", "ttl-model", "time-sensitive query", []byte("request"), []byte("response"))
+			Expect(err).NotTo(HaveOccurred())
+
+			time.Sleep(1100 * time.Millisecond)
+
+			response, found, err := ttlCache.FindSimilar("ttl-model", "time-sensitive query")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(found).To(BeFalse())
+			Expect(response).To(BeNil())
+
+			stats := ttlCache.GetStats()
+			Expect(stats.HitCount).To(Equal(int64(0)))
+			Expect(stats.MissCount).To(Equal(int64(1)))
+		})
+
 		It("should handle error when updating non-existent pending request", func() {
 			err := inMemoryCache.UpdateWithResponse("non-existent-query", []byte("response"))
 			Expect(err).To(HaveOccurred())
--- a/src/semantic-router/pkg/cache/inmemory_cache.go
+++ b/src/semantic-router/pkg/cache/inmemory_cache.go
@@ -230,25 +230,31 @@ func (c *InMemoryCache) FindSimilar(model string, query string) ([]byte, bool, e
 	}
 
 	c.mu.RLock()
-
-	// Check for expired entries during search
-	c.cleanupExpiredEntriesReadOnly()
-
 	var (
 		bestIndex      = -1
 		bestEntry      CacheEntry
 		bestSimilarity float32
 		entriesChecked int
+		expiredCount   int
 	)
+	// Capture the lookup time after acquiring the read lock so TTL checks aren’t skewed by embedding work or lock wait
+	now := time.Now()
 
 	// Compare with completed entries for the same model, tracking only the best match
 	for entryIndex, entry := range c.entries {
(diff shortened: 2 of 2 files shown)


Xunzhuo:
this is weird, why does this happen? the response model should be the same with the real called ones?

src/semantic-router/pkg/extproc/response_handler.go
```diff
@@ -217,6 +217,25 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 		observability.Errorf("Error parsing tokens from response: %v", err)
 		metrics.RecordRequestError(ctx.RequestModel, "parse_error")
 	}
+
```
rootfs, line 220: i don't think we should update the model field, it is returned by vLLM.

yossiovadia:
I will validate if there was a configuration issue in my openshift deployment ( maybe i had mismatch between model-a/b ) 
I'll update.

yossiovadia:
OK! Good news. the issue was incorrect envoy configuration. I will document it here for the future. 

The OpenShift Envoy config I used (deploy/openshift/envoy-openshift.yaml) was using static clusters with header-based routing:
routes:
  - match: { prefix: "/", headers: [{ name: "x-gateway-destination-endpoint", exact_match: "127.0.0.1:8000" }] }
    route: { cluster: model_a_cluster }
  - match: { prefix: "/", headers: [{ name: "x-gateway-destination-endpoint", exact_match: "127.0.0.1:8001" }] }
    route: { cluster: model_b_cluster }
  - match: { prefix: "/" }
    route: { cluster: model_b_cluster }  # DEFAULT - always hit!
Problem: In Envoy's processing pipeline:
Route selection happens FIRST (based on headers)
ExtProc filter runs SECOND (sets the x-gateway-destination-endpoint header)
So the header was never set when routing was evaluated → always fell to default route (model_b_cluster) → everything went to Model-B!


The Correct Fix: Use ORIGINAL_DST Cluster
The local development config (config/envoy.yaml) already had the correct approach:
routes:
  - match: { prefix: "/" }
    route: { cluster: vllm_dynamic_cluster }

clusters:
  - name: vllm_dynamic_cluster
    type: ORIGINAL_DST
    lb_policy: CLUSTER_PROVIDED
    original_dst_lb_config:
      use_http_header: true
      http_header_name: "x-gateway-destination-endpoint"
With ORIGINAL_DST, the cluster reads the header value after ExtProc sets it and routes to the correct endpoint dynamically.

yossiovadia:
OK! Good news. the issue was incorrect envoy configuration. I will document it here for the future. 

The OpenShift Envoy config I used (deploy/openshift/envoy-openshift.yaml) was using static clusters with header-based routing:
routes:
  - match: { prefix: "/", headers: [{ name: "x-gateway-destination-endpoint", exact_match: "127.0.0.1:8000" }] }
    route: { cluster: model_a_cluster }
  - match: { prefix: "/", headers: [{ name: "x-gateway-destination-endpoint", exact_match: "127.0.0.1:8001" }] }
    route: { cluster: model_b_cluster }
  - match: { prefix: "/" }
    route: { cluster: model_b_cluster }  # DEFAULT - always hit!
Problem: In Envoy's processing pipeline:
Route selection happens FIRST (based on headers)
ExtProc filter runs SECOND (sets the x-gateway-destination-endpoint header)
So the header was never set when routing was evaluated → always fell to default route (model_b_cluster) → everything went to Model-B!


The Correct Fix: Use ORIGINAL_DST Cluster
The local development config (config/envoy.yaml) already had the correct approach:
routes:
  - match: { prefix: "/" }
    route: { cluster: vllm_dynamic_cluster }

clusters:
  - name: vllm_dynamic_cluster
    type: ORIGINAL_DST
    lb_policy: CLUSTER_PROVIDED
    original_dst_lb_config:
      use_http_header: true
      http_header_name: "x-gateway-destination-endpoint"
With ORIGINAL_DST, the cluster reads the header value after ExtProc sets it and routes to the correct endpoint dynamically.

PR #469, opened 2025-10-17
Openshift dashboard clean

yossiovadia:
Integrating the dashboard into openshift env.

1. Creates the `dashboard-custom` imagestream if needed
2. Builds the patched dashboard image with OpenWebUI integration
3. Pushes the image to the OpenShift internal registry
4. Applies the deployment YAML if the dashboard doesn't exist, or updates the image if it does
5. Waits for the rollout to complete

### Access the Dashboard

```bash
# Get the dashboard URL
oc get route dashboard -n vllm-semantic-router-system -o jsonpath='https://{.spec.host}'
```
### Why a Custom Build?

The upstream dashboard uses `localhost:3001` for local OpenWebUI development. In OpenShift:

- Services are accessed via routes with unique hostnames
- The OpenWebUI URL must be dynamically constructed based on the deployment environment
- The patch is applied during build time to inject this logic

### DIFF
--- a/bench/vllm_semantic_router_bench/dataset_implementations/openmathrreasoning_dataset.py
+++ b/bench/vllm_semantic_router_bench/dataset_implementations/openmathrreasoning_dataset.py
@@ -1,0 +1,244 @@
+"""
+OpenMathReasoning Dataset Implementation
+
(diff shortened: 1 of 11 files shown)


yossiovadia:
Sample ( on openshift )
image

PR #470, opened 2025-10-18
docs: ddd error prompts when installing VSR using Docker Compose.

yuluo-yx:
AS TITLE


website/docs/installation/docker-compose.md
```diff
@@ -20,6 +20,7 @@ This guide provides step-by-step instructions for deploying the vLLM Semantic Ro
 - **Download classification models (≈1.5GB, first run only):**
 
   ```bash
+  # Tips: If you encounter this error 'hf: command not found', run 'pip install huggingface_hub'.
```
rootfs: also add  `hf_transfer`

yuluo-yx: done

PR #472, opened 2025-10-18
chore: limit make test to minimal model download

cryo-zd:
**What type of PR is this?**

chore: limit make test to minimal model download

[Context]
`make test` always depends on `download-models`, which triggers the full model set whenever `CI_MINIMAL_MODELS` isn't set. That makes PR CI runs re-download LoRA bundle and slows local iteration
https://github.com/vllm-project/semantic-router/blob/f13157cfb9db44a2ac076700c108a089ee4444d6/tools/make/models.mk#L10-L20

[Change]
Update the test aggregate in build-run-test.mk to depend on `download-models-minimal`  instead. CI still runs `make download-models`  earlier, so that prerequisite becomes a no-op there, while local `make test` now reuses the lightweight bundle unless a developer explicitly fetches the full models.

take [PR#451](https://github.com/vllm-project/semantic-router/actions/runs/18558282891/job/52900917457?pr=451#logs) as an example:
modals1
modals2

Fixes #

Release Notes: Yes/No

### DIFF
--- a/deploy/openshift/dashboard/PlaygroundPage.tsx.patch
+++ b/deploy/openshift/dashboard/PlaygroundPage.tsx.patch
@@ -1,0 +1,57 @@
+import React, { useState, useEffect } from 'react'
+import styles from './PlaygroundPage.module.css'
+
+const PlaygroundPage: React.FC = () => {
+  // Use the OpenWebUI route for OpenShift deployments, fallback to embedded proxy
+  const getOpenWebUIUrl = () => {
+    const hostname = window.location.hostname
+    // In OpenShift, use the direct OpenWebUI route
+    if (hostname.includes('dashboard-vllm-semantic-router-system')) {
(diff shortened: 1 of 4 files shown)


tools/make/build-run-test.mk
```diff
@@ -41,7 +41,7 @@ test-semantic-router: build-router
 
 # Test the Rust library and the Go binding
 test: ## Run all tests (Go, Rust, binding)
-test: vet go-lint check-go-mod-tidy download-models test-binding test-semantic-router
+test: vet go-lint check-go-mod-tidy download-models-minimal test-binding test-semantic-router
```
rootfs, line 44: where is the `minimal` target defined?

cryo-zd: Currently, this target is defined in:
https://github.com/vllm-project/semantic-router/blob/886c99bfb5cc218dd3717c7ee32cf88d2f23dc12/tools/make/models.mk#L28-L49

PR #483, opened 2025-10-20
chore: Update test description from Math to General

carlory:

`"Who are you?"` is not a math question.


Xunzhuo (approved):
good to go, dockerhub is down, CI wont pass

PR #491, opened 2025-10-20
feat: add HNSW index to inmemory semantic cache

rootfs:
**What type of PR is this?**

The inmemory cache search is not optimized, adding HNSW index will accelerate the search

This is the benchmark result:
image
image

cc @aeft  

Fixes #

Release Notes: Yes/No


rootfs:
closing and replacing it with #504

PR #518, opened 2025-10-23
Fix: The caller information points to the wrapper function instead of the actual call location

carlory:

image

Fixes #

Release Notes: Yes/No


rootfs (approved):

rootfs:
i was about to remove that function too :D

PR #266, opened 2025-09-28
refactor: Implement modular candle-binding architecture (#254)

rootfs:
﻿
- Restructure codebase into modular layers (core/, ffi/, model_architectures/, classifiers/)
- Add unified error handling and configuration loading systems
- Implement dual-path architecture for traditional and LoRA models
- Add comprehensive FFI layer with memory safety

Maintains backward compatibility while enabling future model integrations.

refactor: Implement modular candle-binding architecture ﻿
- Restructure codebase into modular layers (core/, ffi/, model_architectures/, classifiers/)
- Add unified error handling and configuration loading systems
- Implement dual-path architecture for traditional and LoRA models
- Add comprehensive FFI layer with memory safety

Maintains backward compatibility while enabling future model integrations.

**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes https://github.com/vllm-project/semantic-router/issues/485

Release Notes: Yes/No

PR #528, opened 2025-10-23
refactor: Replace lazy_static with OnceLock for zero-cost concurrent reads based on review 

rootfs:
**What type of PR is this?**

based on review https://github.com/vllm-project/semantic-router/pull/266#discussion_r2448751956

Fixes #

Release Notes: Yes/No

### DIFF
--- a/candle-binding/src/ffi/init.rs
+++ b/candle-binding/src/ffi/init.rs
@@ -136,14 +136,8 @@ pub extern "C" fn init_similarity_model(model_id: *const c_char, use_cpu: bool) 
 
     match BertSimilarity::new(model_id, use_cpu) {
         Ok(model) => {
-            // Set using OnceLock - fails if already initialized
-            match BERT_SIMILARITY.set(Arc::new(model)) {
-                Ok(_) => true,
-                Err(_) => {
-                    eprintln!("BERT similarity model already initialized");
-                    false
-                }
-            }
+            // Set using OnceLock - returns false if already initialized (safe to re-call)
+            BERT_SIMILARITY.set(Arc::new(model)).is_ok()
         }
         Err(e) => {
             eprintln!("Failed to initialize BERT: {e}");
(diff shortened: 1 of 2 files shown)


rootfs:
@OneZero-Y @ivarflakstad PTAL, thanks!

candle-binding/src/ffi/init.rs
```diff
@@ -128,9 +136,14 @@ pub extern "C" fn init_similarity_model(model_id: *const c_char, use_cpu: bool)
...
-            *bert_opt = Some(model);
-            true
+            // Set using OnceLock - fails if already initialized
+            match BERT_SIMILARITY.set(Arc::new(model)) {
+                Ok(_) => true,
+                Err(_) => {
+                    eprintln!("BERT similarity model already initialized");
+                    false
+                }
+            }
```
ivarflakstad: This way of handling `Result`s is fine, but 
1. there are better logging options that `eprintln!`, and
2. if there are cases where you don't need logging you could simply call `.is_ok()` to get a boolean from the result

candle-binding/src/ffi/memory_safety.rs
```diff
@@ -48,17 +48,16 @@ pub struct MemorySafetyResult {
...
-        Arc::new(Mutex::new(HashMap::new()));
-    static ref LORA_MEMORY_POOL: Arc<Mutex<LoRAMemoryPool>> =
-        Arc::new(Mutex::new(LoRAMemoryPool::new()));
-    static ref PATH_SWITCH_GUARD: Arc<RwLock<PathSwitchState>> =
-        Arc::new(RwLock::new(PathSwitchState::new()));
-}
+// Global memory tracker for dual-path safety using once_cell::Lazy
+// These are lazily initialized mutable global state for runtime memory tracking
+static MEMORY_TRACKER: Lazy<Arc<RwLock<HashMap<usize, AllocationTracker>>>> =
+    Lazy::new(|| Arc::new(RwLock::new(HashMap::new())));
```
ivarflakstad, line 54: Nice 👌

candle-binding/src/ffi/init.rs
rootfs: thank you!

Issue #547, opened 2025-10-28
quickstart.sh script failed on local host running


yehuditkerido:
**Description of the bug**
Following the quickstart guideline to run semantic-router on local host are failing. 

**To Reproduce**
1. Go to the quickstart documentation.
2. Try to install semantic-router on local host - run the script. 

**Expected behavior**
Installed  successfully without any issues.

PR #548, opened 2025-10-28
fix: Resolve quickstart script failures and add automated testing

yehuditkerido:
The quickstart script failed to execute successfully in local environment.
  After fixing local issues, added CI testing which revealed additional
  environment-specific failures. This commit includes all fixes plus
  comprehensive testing infrastructure to prevent regressions.

  **Local Environment Fixes:**
  - Fix critical services health check logic
  - Add proper error handling and timeout protection
  - Add helper functions for consistent messaging (success_msg, error_msg, section_header)
  - Fix duplicate YAML keys in config.yaml causing parsing failures
  - Improve service readiness detection

  **CI Environment Fixes:**
  - Add CI detection to skip terminal-only operations (clear, browser prompts)
  - Optimize for CI constraints (disk cleanup, CI_MINIMAL_MODELS)

  **Testing Infrastructure:**
  - Add quickstart-integration-test.yml workflow for E2E validation
  - Test actual routing functionality with real queries (not just health checks)
  - Add shell-syntax-check Makefile target
  - Integrate shell syntax validation into pre-commit workflow

  Tested in both local and CI environments. Workflow passing successfully.
  [https://github.com/yehudit1987/semantic-router/actions/runs/18870299710/job/53846820124](url)
  
  Fixed issue: #547


rootfs:
@JaredforReal @yuluo-yx PTAL, thanks

deploy/docker-compose/docker-compose.yml
```diff
@@ -173,11 +175,13 @@ services:
     container_name: pipelines
     environment:
       - PYTHONUNBUFFERED=1
+      # Load pipeline from URL on startup
```
JaredforReal: Can we revert this? Keep the `vllm_semantic_router_pipe.py` locally for the `pipelines` service is easier for developing. And when we loaded pipeline file from online,  no more addons needed

config/config.yaml
JaredforReal, line 1:
```yaml
default_model: openai/gpt-oss-20b
```
Can u change the `openai/gpt-oss-2b` to `qwen3` in line 227? Thanks!

JaredforReal:
Left a few comments, PTAL @yehudit1987  Thanks!

.github/workflows/quickstart-integration-test.yml
```diff
@@ -0,0 +1,110 @@
...
+          docker logs semantic-router || true
+          echo "=== Envoy Logs ==="
+          docker logs envoy-proxy || true
+          echo "=== Dashboard Logs ==="
+          docker logs semantic-router-dashboard || true
+
+      - name: Clean up
+        if: always()
+        run: |
+          docker compose -f deploy/docker-compose/docker-compose.yml down -v --remove-orphans || true
```
JaredforReal: `make docker-compose-up` in `quickstart.sh` will bring up llm-katan service, which `docker compose -f deploy/docker-compose/docker-compose.yml down -v --remove-orphans || true` may not be able to get down.
Use `make docker-compose-down` should be better. Double check, Thanks!

JaredforReal: `make docker-compose-up` in `quickstart.sh` will bring up llm-katan service, which `docker compose -f deploy/docker-compose/docker-compose.yml down -v --remove-orphans || true` may not be able to get down.
Double check, Thanks!
```suggestion
          make docker-compose-down
```

tools/make/linter.mk
```diff
@@ -41,3 +41,14 @@ codespell: ## Check for common misspellings in code and docs
 shellcheck: ## Lint all shell scripts in the project
 	@$(LOG_TARGET)
 	shellcheck --rcfile=tools/linter/shellcheck/.shellcheckrc $(shell find . -type f -name "*.sh" -not -path "./node_modules/*" -not -path "./website/node_modules/*" -not -path "./dashboard/frontend/node_modules/*" -not -path "./models/*" -not -path "./.venv/*")
+
+shell-syntax-check: ## Validate shell script syntax without execution
```
yuluo-yx: This is a repetition of `make shellcheck`. 

They should be performing the same function. Could we just keep one of them?

JaredforReal:
@yehudit1987 sorry my bad

deploy/docker-compose/docker-compose.yml
```diff
@@ -7,9 +7,9 @@ services:
     ports:
       - "50051:50051"
     volumes:
-      - ../../config:/app/config:ro
-      - ../../models:/app/models:ro
-      - ~/.cache/huggingface:/root/.cache/huggingface
+      - ../../config:/app/config:ro,z
```
yuluo-yx, line 10: Why was the container permission changed from ro to z?
The key point here should be that the data being mounted should not be modified.

.github/workflows/pre-commit.yml
```diff
@@ -95,6 +95,9 @@ jobs:
     - name: Run Code Spell Check
       run: make codespell
 
+    - name: Run Shell Syntax Check
```
yuluo-yx: I'm not sure if we really need this file. I think adding it to "make precommit-check" would be a better option.

refer: https://github.com/vllm-project/semantic-router/blob/860c4c8bd2adc349c59211ac4c42920fbdb0b15a/tools/make/pre-commit.mk#L11

deploy/docker-compose/docker-compose.yml
yehuditkerido: I'm not sure a full revert would be OK, but please correct me if I'm missing something.
In the previous version, I encountered the following issue:
> PermissionError: [Errno 13] Permission denied: './pipelines/vllm_semantic_router_pipe.py'
> OSError: [Errno 16] Device or resource busy: './pipelines/vllm_semantic_router_pipe.py' -> './pipelines/failed/...'

From what I've read, it's considered a bad practice to mount a writable Docker volume and a read-only bind mount to the same path, since it can cause PermissionError or Device or resource busy issues.
I agree that the original fix wasn’t correct — but what do you think about the current solution, where we mount the local addons directory outside /app/pipelines to avoid this conflict?

tools/make/linter.mk
yehuditkerido: Done. My bad.

deploy/docker-compose/docker-compose.yml
yehuditkerido: The permission hasn't changed - it's still ro.
The :z is an additional SELinux label for compatibility with SELinux-enabled systems (Fedora/RHEL/CentOS). The format ro,z means "read-only AND apply SELinux private labeling".
Without the :z label, containers may fail to access mounted files on SELinux-enabled systems due to security policy blocking, even with correct file permissions. The data remains unmodifiable by the container (still ro).

config/config.yaml
yehuditkerido: Done.

.github/workflows/quickstart-integration-test.yml
yehuditkerido: Done.

rootfs:
the quickstart failure was due to the GLIBC dependency, retesting with latest main

deploy/docker-compose/docker-compose.yml
```diff
@@ -176,8 +178,8 @@ services:
     volumes:
       # Persistent pipelines storage (auto-loaded on start)
```
JaredforReal, line 179: Current change would cause `openwebui-pipeline` `target` to be lost.
``` suggestion
      # Persistent pipelines storage (auto-loaded on start)
        - type: volume
        source: openwebui-pipelines
        target: /app/pipelines
        volume:
          nocopy: true
      # Mount our vLLM Semantic Router pipeline (read-only) into the persistent dir
      - type: bind
        source: ./addons/vllm_semantic_router_pipe.py
        target: /app/pipelines/vllm_semantic_router_pipe.py
        read_only: true
```
- The pipeline file is clearly enforced as read-only, reducing accidental edits in the container.
- Using nocopy avoids unintended initial file copies into the named volume.

JaredforReal:
@yehudit1987 Take another look, thanks! And others look good to me.

deploy/docker-compose/docker-compose.yml
yehuditkerido: Done.

PR #550, opened 2025-10-28
fix:hnsw heap polarity

cryo-zd:
**What type of PR is this?**

fix: hnsw heap polarity

- This PR replaces the HNSW  layer-search queues with the correct heap types(min-heap for candidate frontier, max-heap for result set), re-aligning the algorithm with the **negative-dot-product** distance use in `func (h *HNSWIndex) searchLayer`
https://github.com/vllm-project/semantic-router/blob/860c4c8bd2adc349c59211ac4c42920fbdb0b15a/src/semantic-router/pkg/cache/inmemory_cache.go#L837-L849
- Now, nearest-neighbor search once again explores and retains the closest points, restoring HNSW correctness and preventing the cache from poor or empty matches

[Efficient and robust approximate nearest neighbor search using Hierarchical Navigable Small World graphs](https://arxiv.org/abs/1603.09320)
hnsw

Fixes #

Release Notes: Yes/No

### DIFF
--- a/src/semantic-router/pkg/cache/inmemory_cache.go
+++ b/src/semantic-router/pkg/cache/inmemory_cache.go
@@ -879,23 +879,8 @@ func (h *minHeap) pop() (int, float32) {
 	return result.index, result.dist
 }
 
-func (h *minHeap) peekDist() float32 {
-	if len(h.data) == 0 {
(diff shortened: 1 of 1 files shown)


rootfs:
thank you @cryo-zd for spotting this!

PR #499, opened 2025-10-21
chore: upgrade rust version to 1.90 in all related Dockerfiles

carlory:
**What this PR does / why we need it**:

```
## Prerequisites

- Go Version 1.24.1 or higher (matches the module requirements)
- Rust Version 1.90.0 or higher (for Candle bindings, supports 2024 edition)
```
related-to https://github.com/vllm-project/semantic-router/pull/176

**Which issue(s) this PR fixes**:

Fixes #

Release Notes: Yes/No

PR #553, opened 2025-10-29
fix(k8s ci): extend wait windows in the workflow

JaredforReal:
Fixes #499


JaredforReal:
cc @rootfs

rootfs:
@JaredforReal you probably also need a CI specific readiness probe too
https://github.com/vllm-project/semantic-router/blob/main/deploy/kubernetes/deployment.yaml#L144

JaredforReal:
@rootfs PTAL, thanks!

PR #564, opened 2025-10-31
feat: add embedding model continuous batching scheduler

rootfs:
**What type of PR is this?**

Adding 

Also inspired by vLLM, the embedding continuous batching fully utilizes the GPU compute and reduces the embedding latency by 11x.

Fixes #

Release Notes: Yes/No


rootfs:
@OneZero-Y @cryo-zd @aeft can you review it? thanks!

candle-binding/src/model_architectures/embedding/continuous_batch_scheduler.rs
```diff
@@ -0,0 +1,590 @@
...
+                    .fetch_add(actual_batch_size as u64, Ordering::Relaxed)
+                    + actual_batch_size as u64;
+
+                // Update average latency (minimal lock for float arithmetic)
+                let total_latency_ms: f64 = latencies.iter().sum();
+                let mut avg_latency = stats.avg_latency_ms.lock().unwrap();
+                *avg_latency = (*avg_latency * (total_requests - actual_batch_size as u64) as f64
+                    + total_latency_ms)
+                    / total_requests as f64;
+                drop(avg_latency);
```
OneZero-Y: The average latency calculation uses values that might change between reads?
```rust
let total_requests = stats
    .total_requests
    .fetch_add(actual_batch_size as u64, Ordering::Relaxed)  // ← Thread A
    + actual_batch_size as u64;

// Update average latency
let mut avg_latency = stats.avg_latency_ms.lock().unwrap();
*avg_latency = (*avg_latency * (total_requests - actual_batch_size as u64) as f64  
    // ↑ Thread B might have modified total_requests between fetch_add and here
    + total_latency_ms)
    / total_requests as f64;  // ← Using stale total_requests value
```
 would be to combine the related fields into a single mutex to ensure atomic updates.
```rust
// In continuous_batch_scheduler.rs, update Stats struct
struct Stats {
    batches_processed: AtomicU64,
    cached_hits: AtomicU64,
    // Combine these two into a single lock:
    inner: Mutex<StatsInner>,
}

struct StatsInner {
    total_requests: u64,
    avg_latency_ms: f64,
}

//lines 506-517
let mut stats_inner = stats.inner.lock().unwrap();
stats_inner.total_requests += actual_batch_size as u64;
stats_inner.avg_latency_ms = (
    stats_inner.avg_latency_ms * (stats_inner.total_requests - actual_batch_size as u64) as f64
    + total_latency_ms
) / stats_inner.total_requests as f64;
```

rootfs: thanks!

rootfs:
CI error is due to dashboard container image, will be investigated

PR #587, opened 2025-11-04
fix: correct HNSW frontier comparisons in hybrid cache

cryo-zd:
**What type of PR is this?**

fix: correct HNSW frontier comparisons in hybrid cache

&emsp; This PR fixes the **sign bug** in both `func (h *HybridCache) searchLayerHybrid` and `func (h *HybridCache) searchLayerHybridWithEarlyStop`: the code compared `currentDist` against  `-results.data[0].dist` even though distances were already stored as negatives, so saturated frontiers kept enqueuing worse neighbors instead of pruning them.  
&emsp; ( the stopping condition `(currentDist > -results.data[0].dist)` almost never fires, and the “is this better than the frontier?” check `(dist < -results.data[0].dist)` is nearly always true.)
&emsp; This prevents needless graph traversal, keeps lookup latency under control, and avoids returning responses routed through poor matches.
https://github.com/vllm-project/semantic-router/blob/f8f042d4b5c5c13d8d380dd589fecba5e43d51bf/src/semantic-router/pkg/cache/hybrid_cache.go#L927-L951

Fixes #

Release Notes: Yes/No

### DIFF
--- a/src/semantic-router/pkg/cache/cache_test.go
+++ b/src/semantic-router/pkg/cache/cache_test.go
@@ -1514,6 +1514,95 @@ milvus:
 	}
 }
 
+// Ensures hybrid layer search skips candidates that are already worse than the frontier.
+func TestHybridCacheSearchLayerPrunesWeakerBranch(t *testing.T) {
+	// Regression fixture: the buggy comparison let the frontier accept a much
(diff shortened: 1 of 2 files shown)


rootfs:
@cryo-zd thank you for testing this!

Issue #88, opened 2025-09-08
[v0.1] Envoy ExtProc integration for llm-d gateway


rootfs:
# Acceptance
- [ ] ExtProc header/body mutation consistent with LLM-d filter chains
- [ ] Example Envoy configs for common patterns (e.g., A/B test, canary routing)

srampal:
Hey @Gregory-Pereira I am already working on this as part of the parent issue 39 which is assigned to me (and 2 others). Let's co-ordinate

Gregory-Pereira:
Sanjeev and I discussed this, he will be taking over the issue

srampal:
Currently waiting for https://github.com/vllm-project/semantic-router/pull/229 to merge. That includes most of the functionality to also address this issue. Once that PR merges, this issue will also get addressed but I will add additional docs to test and confirm this issue as well.

Xunzhuo:
close this as it is already done.

Issue #39, opened 2025-09-04
[v0.1] Envoy ExtProc integration for AI gateways


rootfs:
# Acceptance: 

- [ ] ExtProc header/body mutation consistent with GIE/LLM-d/Envoy AI Gateway filter chains (documented setup for each)  
- [ ] Example Envoy configs for common patterns (e.g., A/B test, canary routing)

LYZJU2019:
I can take a look at this!

rootfs:
Thanks! If possible, can you outline a plan and we can loop in other folks too? Since this is very community collaborative, it would be great if your plan can be shared with GIE/llm-d/AI GW/etc communities.

LYZJU2019:
Sure. I plan to proceed with llm-d first with two stages:

Stage 1: Documentation
- Show the typical chain for LLM-d (auth → rate-limit → **ext_proc** → router).
- Emphasize ext_proc’s role: normalize request/response, set routing metadata, and ensure consistency before endpoint selection.
- Document which headers/metadata are expected downstream (e.g., x-ig-model, x-ig-criticality, dynamic metadata ig.model.id, ig.session.affinity_key).

Stage 2: Implement llm-d ext_proc
- Provide a minimal Go gRPC server using Envoy’s ext_proc API [link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto)
- Normalization request into a canonical schema (e.g., OpenAI chat style).
- Compute lightweight token estimates, set metadata (model.id, criticality, lora_ids, session.affinity_key, cost.estimated_tokens).
- Mirror key fields into x-ig-* headers for observability/debugging.

Let me know how this plan sounds!

Xunzhuo:
I think our ext proc should be firstly integrated with an AI GW, not integrated with an existing extproc, I suggest having a try with envoy ai gw, then integrate with inference framework with ai gw

rootfs:
@LYZJU2019 thanks! looking forward to your progress.

rootfs:
@srampal @Xunzhuo @LYZJU2019 Thanks for contributing to this task. For clarity, I've created sub issues for each of these integration targets. If you are working on these gateway targets, please share your plans on the related sub issues. Thanks

srampal:
[...]

PR #589, opened 2025-11-04
feat(llm-d): integrate vsr with llm-d 

srampal:
**What type of PR is this?**
feat(llm-d): Test vLLM semantic Router + LLM-D solution on a single combined gateway. Add a deployment guide and deployment manifests.

Validates a solution that combines vLLM Semantic Router and adds deployment guide and example manifests to allow this to be tested.

https://github.com/vllm-project/semantic-router/issues/39

Fixes #
Sub-issue #88 within #39

Release Notes: Yes/No


deploy/kubernetes/llmd-base/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Send Test Requests
+
+Try the following cases with and without model "auto" selection to confirm that Istio + vsr together are able to route queries to the appropriate model. The query responses will include information about which model was used to serve that request.
+
+Example queries to try include the following
+
+```bash
+# Model name llama3-8b provided explicitly, should route to this backend 
+curl http://192.168.49.2:32293/v1/chat/completions   -H "Content-Type: application/json"   -d '{
```
Xunzhuo, line 187: Can u change the IP?

Xunzhuo:
Can you describe the architecture you are adding now? Like where VSR is and where InferencePool is, and how they work together

deploy/kubernetes/llmd-base/README.md
srampal: So to be clear the <IP, port> on line 186 is the <LoadBalancer IP, port> exported by minikube to which external clients can send http requests/ LLM queries. It is not the internal IP of the backend pod or EPP or anything like that. This is simply the external IP on which the Gateway receives original client requests. I will also push a small update to make the text in the comment line above a bit clearer.

srampal:
> Can you describe the architecture you are adding now? Like where VSR is and where InferencePool is, and how they work together

This design is fully aligned and compatible with the Istio gateway integration feature (before LLM-D) that I pushed in the past. In both cases, I neither use original_dst cluster of Envoy nor InferencePools. InferencePools are only used by the LLM-D/ EPP but not by VSR. Take a look at the httproute manifests. The user adds routes for header match of "x-selected-model" returned by vsr which point to the InferencePool for the corresponding base model.  Also note that endpoint1 IP configured in the VSR config does not matter and is not used (ideally it should be removed from vsr config in future). When I added the Istio gateway integration, I also added the "clear_route_cache" to the vsr config previously for the Istio integration (see deploy/kubernetes/istio/config.yaml). This is used to make the gateway logic work for both basic Istio and Istio + LLM-D cases. Hope that helps. I have tested it out to confirm these steps work correctly.

Issue #486, opened 2025-10-20
Routing Core Optimization: Model-Based Routing


Xunzhuo:
### Is your feature request related to a problem? Please describe.

The current routing implementation is based on the original destination cluster approach, which has several significant limitations:

- Limited support for different LLM connection types (restricted to specific protocols)
- Cannot handle mixed routing scenarios (in-cluster services + external LLM providers)
- Lacks advanced traffic management capabilities (failover, weighted distribution, circuit breaking)
- No built-in support for LLM provider authentication
- Missing retry and health check mechanisms
- Cannot implement token-based rate limiting
- Inflexible architecture that tightly couples routing logic with cluster management

These limitations prevent the system from scaling effectively and supporting diverse LLM deployment scenarios required by modern AI applications.

### Describe the solution you'd like

**Architecture Overview:**

Migrate to a modular architecture where semantic router focuses solely on model switching, while cluster and endpoint management is delegated to the AI Gateway layer.

**Implementation Approach:**

1. **Resource Rendering**: User-configured models and IPs are rendered as AI Gateway resources, allowing the AI Gateway to act as the Envoy control plane and pre-distribute routing rules, clusters, and endpoints.

2. **Global External Processor Integration**: Semantic router operates as a global external processor (extproc) injected into the Gateway. When requests reach the AI Gateway's Envoy:
   - Transform the model identifier from MoM (Mixture of Models) to the actual model value
   - Render additional headers to support complex routing scenarios
   - This behavior is transparent to the AI Gateway

**Key Benefits:**

- ✅ Support for multiple LLM connection types: HTTPS, HTTP, IP-based, DNS-based
- ✅ Hybrid routing support: in-cluster services + external LLM providers (Claude, Gemini, DeepSeek, etc.)
- ✅ Model-level failover capabilities
- ✅ Weighted distribution between models
- ✅ LLM provider authentication support
- ✅ Built-in retry, health check, and circuit breaking mechanisms
- ✅ Token-based rate limiting

**Trade-offs:**

- ⚠️ Load balancing operates at the model level rather than IP level

**Technical Components:**

- Semantic router as global extproc filter
- AI Gateway as Envoy control plane
- Dynamic resource rendering for models and endpoints
- Header-based routing enhancement
- Integration with existing Envoy traffic management features

### Additional context

[...]

PR #593, opened 2025-11-05
feat(openshift): Split vllm-katan-a and vllm-katan-b to run on separate pods rather than the same semantic router pod.

szedan-rh:
Auto-discovers vLLM service ClusterIPs at deployment time, enabling portable OpenShift deployments across different clusters.

Tested on OpenShift with split architecture (3 pods: vllm-model-a, vllm-model-b, semantic-router+envoy).

### DIFF
--- a/deploy/openshift/deploy-to-openshift.sh
+++ b/deploy/openshift/deploy-to-openshift.sh
@@ -170,9 +170,9 @@ sed -e "s|address: \".*\" # model-a-ip|address: \"$MODEL_A_IP\"|g" \
 if ! grep -q "$MODEL_A_IP" "$TEMP_CONFIG" || ! grep -q "$MODEL_B_IP" "$TEMP_CONFIG"; then
     warn "IP substitution may have failed. Using template config instead..."
     # Fallback: create config with sed on known patterns
-    cat "$SCRIPT_DIR/config-openshift.yaml" | \
-        sed "s/172\.30\.64\.134/$MODEL_A_IP/g" | \
-        sed "s/172\.30\.116\.177/$MODEL_B_IP/g" > "$TEMP_CONFIG"
+    sed -e "s/172\.30\.64\.134/$MODEL_A_IP/g" \
+        -e "s/172\.30\.116\.177/$MODEL_B_IP/g" \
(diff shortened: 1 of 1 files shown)


Xunzhuo:
@szedan-rh i have a better idea to solve this in short term, in mid term, we need to wait for https://github.com/vllm-project/semantic-router/issues/486

szedan-rh:
Hi @Xunzhuo , Thanks for your comment, could you please share your idea?
This is existing script that was developed by Yossi for the demo, I did just the changes for running it in split pods rather than one pod with all containers inside it.

Xunzhuo:
i guess this is to support to configure k8s service as backend, right? like configuring service namespace/name

szedan-rh:
Nope. actually it's just changing how we was deploying the semantic routers pods, it was creating just for the demo that Yossi did in PyTorch conference.
The previous implementation was to create one pod with 4 containers, semantic router, envoy, vllm-katan-a and vllm-katan-b
The current implementation is that vllm-a and vllm-b will be in separated pods and one pod for semantic router and envoy to demonstrate that the semantic router is routing to model-a and model-b according to the classification

szedan-rh:
@Xunzhuo - could you please take another look?

Xunzhuo (commented):
vllm_endpoint is for testing/PoC in local/docker-compose, i suggest looking and testing https://vllm-semantic-router.com/docs/installation/k8s/ai-gateway

With this we dont need to care about how to manage backend ip or service name, i just select the model, the model can be an in-cluster and also can be an external LLM Provider.

not a blocker for this PR, plz fix the lint and i will get this in. thanks

szedan-rh:
Thank you for sharing @Xunzhuo , will check that for sure.

yossiovadia:
Hi @szedan-rh , @Xunzhuo , i looked into it ( the lint issues ) , it seems we had a bug in the 
.pre-commit-config.yaml  this should resolve it - 
https://github.com/vllm-project/semantic-router/pull/609

the chicken and egg would be resolved once this PR will be merged ( then the #609) pre-commit test will pass.

rootfs (approved):

PR #595, opened 2025-11-05
fix: Add DNS hostname support to vLLM endpoint validator [WIP]

### LINKED ISSUE #594: validateIPAddress Rejects DNS Names for Kubernetes Service Discovery
Current Behavior

  The validateIPAddress function (lines 22-51) uses net.ParseIP() to validate endpoint addresses, which only accepts IP addresses in the format x.x.x.x or IPv6.

  Problem

  In Kubernetes/OpenShift environments, service discovery uses DNS names instead of static IP addresses:
  - Services are referenced as <service-name>.<namespace>.svc.cluster.local
  - Example: vllm-model-a.vllm-semantic-router-system.svc.cluster.local
  - ClusterIP services are dynamically assigned and may change across clusters

  When using DNS names in configuration, the validator incorrectly rejects valid Kubernetes service addresses, forcing users to:
  1. Manually discover ClusterIPs after deployment
  2. Update configuration with hardcoded IPs
  3. Lose portability across clusters

  Expected Behavior

  The validation function should accept both:
  - IP addresses: 172.30.89.145
  - DNS names: vllm-model-a.vllm-semantic-router-system.svc.cluster.local
  - Short DNS names: vllm-model-a (within same namespace)

  Impact

  This limitation prevents using standard Kubernetes service discovery patterns and breaks deployment portability across different clusters.

szedan-rh:
## Summary

Fixes #594

This PR adds support for DNS hostnames in vLLM endpoint address validation, enabling Kubernetes service discovery patterns like `vllm-service.namespace.svc.cluster.local`.

## Problem

The current `validateIPAddress` function in `validator.go` only accepts IP addresses (IPv4/IPv6) and rejects DNS hostnames. This prevents using Kubernetes DNS-based service discovery, which is the recommended pattern for service communication in Kubernetes/OpenShift environments.

## Solution

Enhanced `validateIPAddress` to accept both IP addresses and RFC-compliant DNS hostnames:
- Added `validateDNSHostname()` function with proper DNS validation
- Validates hostname length (1-253 chars total)
- Validates label length (1-63 chars per label)
- Ensures labels start/end with alphanumeric characters
- Allows hyphens within labels
- Rejects invalid characters (underscores, @, etc.)

### Modified Files
- `src/semantic-router/pkg/config/validator.go`: Added DNS validation support
- `src/semantic-router/pkg/config/config_test.go`: Updated tests to reflect DNS support

### Validation Still Rejects
- Protocol prefixes (`http://`, `https://`)
- Paths (`/api/v1`, `/health`)
- Port numbers in address field (use `port` field instead)

## Supported Address Formats

✅ **IPv4**: `192.168.1.1`, `127.0.0.1`
✅ **IPv6**: `::1`, `2001:db8::1`
✅ **DNS names**: `example.com`, `vllm-service.namespace.svc.cluster.local`
✅ **Short names**: `vllm-service`, `localhost`

## Testing

- All 141 existing tests pass
- Updated 4 tests to validate DNS hostname acceptance
- Test And Build workflow: ✅ PASSED
- Pre-commit workflow: ✅ PASSED

## Example Configuration

```yaml
vllm_endpoints:
  - name: "model-a"
    address: "vllm-model-a.vllm-semantic-router-system.svc.cluster.local"
    port: 8000
    weight: 1
  - name: "model-b"
    address: "vllm-model-b"  # Short name also works
    port: 8001
    weight: 1
```

### DIFF
--- a/src/semantic-router/pkg/config/config_test.go
+++ b/src/semantic-router/pkg/config/config_test.go
@@ -1048,7 +1048,7 @@ default_model: "test-model"
 			})
 
 			Context("with invalid address formats", func() {
-				It("should reject domain names", func() {
+				It("should accept valid DNS names", func() {
 					configContent := `
 vllm_endpoints:
   - name: "endpoint1"
@@ -1072,11 +1072,9 @@ default_model: "test-model"
 					err := os.WriteFile(configFile, []byte(configContent), 0o644)
 					Expect(err).NotTo(HaveOccurred())
 
-					_, err = Load(configFile)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(ContainSubstring("endpoint1"))
-					Expect(err.Error()).To(ContainSubstring("address validation failed"))
-					Expect(err.Error()).To(ContainSubstring("invalid IP address format"))
+					cfg, err := Load(configFile)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(cfg.VLLMEndpoints[0].Address).To(Equal("example.com"))
 				})
 
 				It("should reject protocol prefixes", func() {
@@ -1200,8 +1198,8 @@ default_model: "test-model"
 					Expect(errorMsg).To(ContainSubstring("Supported formats"))
 					Expect(errorMsg).To(ContainSubstring("IPv4: 192.168.1.1"))
 					Expect(errorMsg).To(ContainSubstring("IPv6: ::1"))
+					Expect(errorMsg).To(ContainSubstring("DNS names: example.com"))
 					Expect(errorMsg).To(ContainSubstring("Unsupported formats"))
-					Expect(errorMsg).To(ContainSubstring("Domain names: example.com"))
 					Expect(errorMsg).To(ContainSubstring("Protocol prefixes: http://"))
 				})
 			})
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Have u tested this? I guess this won't work

szedan-rh:
Yes it was tested, let me double check that again

szedan-rh:
Hi @Xunzhuo - I have checked that again with deploy to Openshift script and it working.
do you have another way to test that more carefully to ensure the functionality is not broken?

Xunzhuo:
so you are configuring dns name in endpoint address? are you using envoyproxy with original dst?

Xunzhuo:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/original_dst

szedan-rh:
@Xunzhuo 
We are not using Envoy's original destination load balancing. The semantic router operates as an Envoy External Processing (ExtProc) filter, but it makes direct HTTP  connections to vLLM backends, not routing through Envoy clusters.

szedan-rh:
@Xunzhuo ,
After reading the documentation, I got your point.
closing the ticket and the PR.
Thank you for looking into this

Issue #573, opened 2025-11-02
[Testing] improve and enable the skipped testing cases


Xunzhuo:
### 1. BERT Similarity Model Tests (8 tests)
**Root Cause:** Failed to initialize BERT similarity model (`sentence-transformers/all-MiniLM-L6-v2`)

- `TestInitModel/InitWithSpecificModel` - Model initialization with specific model
- `TestTokenization` - Tokenization functionality tests
- `TestEmbeddings` - Embedding generation tests
- `TestSimilarity` - Similarity calculation tests
- `TestFindMostSimilar` - Finding most similar items tests
- `TestUtilityFunctions/IsModelInitialized` - Model initialization status check
- `TestErrorHandling/EmptyStringHandling` - Empty string error handling
- `TestConcurrency` - Concurrent operations tests

### 2. Milvus Vector Database Tests (5 tests)
**Root Cause:** `MILVUS_URI` environment variable not set

- `TestHybridCacheBasicOperations` - Basic hybrid cache operations
- `TestHybridCachePendingRequest` - Pending request handling in hybrid cache
- `TestHybridCacheEviction` - Cache eviction logic
- `TestHybridCacheLocalCacheHit` - Local cache hit scenarios
- `TestHybridVsMilvusSmoke` - Smoke test comparing hybrid vs Milvus (also skipped in short mode)

### 3. Model Directory Tests (3 tests)
**Root Cause:** Models directory does not exist at `../../../../../models`

- `TestAutoDiscoverModels_RealModels` - Auto-discovery of available models
- `TestAutoInitializeUnifiedClassifier` - Unified classifier initialization
- `TestUnifiedClassifier_Integration` - Integration tests for unified classifier

### 4. LoRA Model Tests (1 test)
**Root Cause:** Candle BERT token classifier not available at `../models/lora_pii_detector_bert-base-uncased_model`

- `TestCandleBertTokensWithLabels` - Token classification with labels using LoRA model

### 5. Embedding Model Tests (4 tests)
**Root Cause:** Failed to initialize embedding models (Gemma model config not found at `../models/embeddinggemma-300m/config.json`)

- `TestGetEmbeddingSmart` - Smart embedding retrieval
- `TestGetEmbeddingWithDim` - Embedding with specific dimensions
- `TestEmbeddingConsistency` - Embedding consistency validation
- `TestEmbeddingPriorityRouting` - Priority-based routing with embeddings

yehuditkerido:
[...]

PR #632, opened 2025-11-11
fix(tests): resolve 3 skipped model directory tests (Section 3/5)

yehuditkerido:
- Fix incorrect relative path to models directory
- Fix Rust double initialization bug in init_lora_unified_classifier
- Add LoRA models to CI minimal downloads
- Bump model cache version to force fresh download
- Change tests to fail visibly instead of silently skipping

Tests TestAutoDiscoverModels_RealModels, TestAutoInitializeUnifiedClassifier,
and TestUnifiedClassifier_Integration now pass.

Fix Partially issue #573

### DIFF
--- a/src/semantic-router/pkg/classification/classifier_test.go
+++ b/src/semantic-router/pkg/classification/classifier_test.go
@@ -21,6 +21,8 @@ import (
 	mcpclient "github.com/vllm-project/semantic-router/src/semantic-router/pkg/mcp"
 )
 
+const testModelsDir = "../../../../models"
+
 func TestClassifier(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Classifier Suite")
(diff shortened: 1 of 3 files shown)


rootfs:
@yuluo-yx PTAL, thanks

.github/workflows/test-and-build.yml
```diff
@@ -62,9 +62,9 @@ jobs:
         with:
           path: |
             models/
-          key: ${{ runner.os }}-models-v1-${{ hashFiles('tools/make/models.mk') }}
+          key: ${{ runner.os }}-models-v2-${{ hashFiles('tools/make/models.mk') }}
```
yuluo-yx: Why is v2 being used here? Is it really necessary for us to upgrade to v2? The model files are very large and will occupy a lot of cache space on GitHub.

src/semantic-router/pkg/classification/classifier_test.go
```diff
@@ -2909,13 +2909,11 @@ func createMockModelFile(t *testing.T, dir, filename string) {
 
 func TestAutoDiscoverModels_RealModels(t *testing.T) {
 	// Test with real models directory
-	modelsDir := "../../../../../models"
+	modelsDir := "../../../../models"
```
yuluo-yx: Can we define it as a constant? This is used in many places.

.github/workflows/test-and-build.yml
yehuditkerido: The only reason I introduced the v2 cache key was that, at the time, the v1 cache was restoring an outdated/partial models folder, which caused CI failures after we added the LoRA models. Using a new key prevented GitHub Actions from picking up that old cache and forced a clean download, which solved the issue.

Now that the old v1 cache has expired, reverting back to v1 works again and doesn’t seem to cause problems. I also understand the concern about cache size, so keeping a single version definitely makes sense.

src/semantic-router/pkg/classification/classifier_test.go
yehuditkerido: Done.

yuluo-yx (approved):
/lgtm

PR #651, opened 2025-11-14
fix(helm): remove namespace template to resolve installation conflicts

Xunzhuo:
**What type of PR is this?**

Bug fix - This PR fixes a Helm chart installation issue related to namespace creation.

This PR removes the `namespace.yaml` template from the Helm chart to resolve installation conflicts when the namespace already exists.

**Problem:**
- The `namespace.yaml` template uses the `lookup` function to check if a namespace exists
- This causes installation failures with the error: `namespaces "vllm-semantic-router-system" already exists`
- The `lookup` function doesn't work reliably in all scenarios (OCI registries, certain Helm versions)

**Solution:**
- Remove the `namespace.yaml` template entirely
- Let Helm's built-in `--create-namespace` flag handle namespace creation
- This approach is more reliable and follows Helm best practices

**Testing:**
After this change, users can install the chart using:
```bash
helm install semantic-router oci://ghcr.io/xunzhuo/charts/semantic-router \
  --version v0.0.0-latest \
  --namespace vllm-semantic-router-system \
  --create-namespace
```

The installation will work whether the namespace exists or not.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

### DIFF
--- a/deploy/helm/validate-chart.sh
+++ b/deploy/helm/validate-chart.sh
@@ -119,7 +119,6 @@ required_files=(
     "templates/deployment.yaml"
     "templates/service.yaml"
     "templates/configmap.yaml"
-    "templates/namespace.yaml"
     "templates/pvc.yaml"
     "templates/serviceaccount.yaml"
     "templates/ingress.yaml"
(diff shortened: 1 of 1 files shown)


Xunzhuo:
done @yuluo-yx thanks

Issue #725, opened 2025-11-23
Enable LoRA auto-detection for Jailbreak Detection


yossiovadia:
## Summary

Following the request from @Xunzhuo in [PR #718](https://github.com/vllm-project/semantic-router/pull/718#issuecomment-2493848361), this issue tracks the migration of Jailbreak Detection to support LoRA auto-detection.

## Background

Currently, PII detection has LoRA auto-detection (PR #709), and Intent Classification will have it (#724), but Jailbreak Detection does not. This creates an inconsistency where some classification features can leverage LoRA models while others cannot.

**Note**: This work depends on #724 (Intent Classification LoRA) being merged first, as it will establish the pattern to follow.

## Current Behavior (BEFORE)

**Problem**: Jailbreak detection cannot automatically use LoRA models.

**How it works**:
1. Configuration has a `use_modernbert` flag that determines which initializer to use
2. System makes a hardcoded choice between two paths:
   - `use_modernbert: false` → Uses `LinearJailbreakInitializer` (Traditional BERT only)
   - `use_modernbert: true` → Uses `ModernBertJailbreakInitializer` (ModernBERT only)
3. Neither path can detect or use LoRA models automatically
4. Even if you point `model_id` to a LoRA jailbreak model, it will fail

**Current config**:
```yaml
prompt_guard:
  use_modernbert: true
  model_id: "models/jailbreak_classifier_modernbert-base_model"
  # Cannot use: models/lora_jailbreak_classifier_bert-base-uncased_model
```

## Expected Behavior (AFTER)

**Solution**: Jailbreak detection should auto-detect LoRA models (just like PII and Intent).

**How it should work**:
1. Single auto-detecting initializer that intelligently routes based on model type
2. Detection happens automatically by checking:
   - LoRA weights in model.safetensors file
   - Presence of lora_config.json
3. Smart fallback chain: LoRA → Traditional BERT → ModernBERT
4. The `use_modernbert` config flag becomes optional/ignored (backward compatible)
5. Zero configuration needed - just point to model path and system figures it out

**Example** (can now use LoRA models):
```yaml
prompt_guard:
  model_id: "models/lora_jailbreak_classifier_bert-base-uncased_model"
  use_modernbert: false  # Ignored - auto-detection finds LoRA and uses it
```

## Implementation Notes

- **Depends on**: Issue #724 being merged first (provides the pattern)
- Follow the same pattern as Intent Classification and PII detection
[...]

PR #718, opened 2025-11-22
fix(api): expose actual PII confidence scores instead of hardcoded 0.9

yossiovadia:
## Summary

This PR fixes the PII API endpoint to return actual confidence scores from the LoRA model instead of hardcoded `0.9` values.

Fixes #717

## Changes

1. **Added `ClassifyPIIWithDetails()` method** - New classifier method that returns full entity details including actual confidence scores
2. **Updated `DetectPII()` service** - Modified to use `ClassifyPIIWithDetails()` instead of `ClassifyPII()` to access real confidence values
3. **Updated `config.e2e.yaml`** - Configured to use LoRA PII model and added default catch-all decision for E2E testing

### Before (Hardcoded)
```json
{
  "entities": [
    {"type": "B-EMAIL_ADDRESS", "confidence": 0.9},
    {"type": "B-PERSON", "confidence": 0.9}
  ]
}
```

### After (Actual Scores)
```json
{
  "entities": [
    {"type": "B-EMAIL_ADDRESS", "confidence": 0.9869},
    {"type": "B-PERSON", "confidence": 0.9991}
  ]
}
```

## Test Results

Local testing shows the API now correctly returns actual model confidence scores:

| Test Case | Before | After |
|-----------|--------|-------|
| Email detection | 0.9 (hardcoded) | **0.9869** (actual) |
| SSN detection | 0.9 (hardcoded) | **0.9916, 0.8326** (actual) |
| Multiple PII | 0.9 (hardcoded) | **0.7734-0.9991 range** (actual) |

## Technical Details

The underlying LoRA PII model was already producing accurate confidence scores (fixed in #709), but the API layer was discarding them. This change passes through the actual scores to API consumers.

The new `ClassifyPIIWithDetails()` method returns `[]PIIDetection` with full entity information, while maintaining backward compatibility by keeping the existing `ClassifyPII()` method for callers that only need entity types.

## Testing

- [x] Local testing with `make run-router-e2e`
- [x] Manual API testing with curl
- [x] Pre-commit checks passed
- [x] DCO sign-off added

PR #709, opened 2025-11-20
fix(647): enable LoRA PII auto-detection with minimal changes

yossiovadia:
Switch PII classification from hardcoded ModernBERT to auto-detecting Candle BERT classifier. The Rust layer already has built-in auto-detection that checks for lora_config.json and routes to LoRA or Traditional models.

Changes:
1. Init: Use InitCandleBertTokenClassifier (has auto-detect built-in)
2. Inference: Use ClassifyCandleBertTokens (auto-routes to initialized classifier)

This enables LoRA PII models to work automatically without config changes, providing higher confidence scores for PII entity detection.

Fixes #647

Issue #647, opened 2025-11-12
PII Model Produces Insufficient Confidence Scores for Critical PII Types


yossiovadia:
# PII Model Produces Insufficient Confidence Scores for Critical PII Types

## Summary

The PII classification model (`pii_classifier_modernbert-base_presidio_token_model`) produces confidence scores below the 0.7 threshold for critical PII types (SSN, Credit Card, Email), preventing policy enforcement from triggering.

## Problem: Model Quality Issue

The PII model detects entities but assigns **insufficient confidence scores** to critical PII types:

| PII Type | Example Input | Model Confidence | Threshold | Result |
|----------|---------------|------------------|-----------|---------|
| **SSN** | `My SSN is 123-45-6789` | **0.6478** | 0.7 | ❌ Not enforced |
| **Email** | `john@example.com` | **0.6368** | 0.7 | ❌ Not enforced |
| **Email (dots)** | `john.doe@example.com` | **0.5742** | 0.7 | ❌ Not enforced |
| **Credit Card** | `4111-1111-1111-1111` | **0.3679** | 0.7 | ❌ Not enforced |
| **Credit Card (sentence)** | `Card number 4111-1111-1111-1111` | **0.4742** | 0.7 | ❌ Not enforced |

Because scores fall below the threshold, entities are filtered out and policy enforcement never triggers.

## What Works (Proof Model Quality is the Issue)

When the model produces high confidence, enforcement **does** work correctly:

| PII Type | Example Input | Model Confidence | Result |
|----------|---------------|------------------|---------|
| **Person Names** | `John Smith` | **0.9988** | ✅ Policy enforced |
| **Person (sentence)** | `My name is John Smith` | **0.9995** | ✅ Policy enforced |
| **Email (prefixed)** | `Email: john@example.com` | **0.8262** | ✅ Policy enforced |
| **Phone (sentence)** | `Call me at 555-123-4567` | **0.9435** | ✅ Policy enforced |
| **Credit Card (no dashes)** | `4111111111111111` | **0.7549** | ✅ Policy enforced |

**This proves the enforcement logic is working correctly** - the issue is purely the model's inability to confidently detect critical PII patterns.

### 1. Misclassification

The model incorrectly classifies some PII types:

- `4111-1111-1111-1111` → Detected as `PHONE_NUMBER` instead of `CREDIT_CARD` (conf: 0.37)
- `john@example.com` → Sometimes detected as `PERSON` instead of `EMAIL_ADDRESS`
[...]

PR #726, opened 2025-11-23
feat(classifier): enable LoRA auto-detection for intent classification

### LINKED ISSUE #724: Enable LoRA auto-detection for Intent/Category Classification
## Summary

Following the request from @Xunzhuo in [PR #718](https://github.com/vllm-project/semantic-router/pull/718#issuecomment-2493848361), this issue tracks the migration of Intent/Category Classification to support LoRA auto-detection.

## Background

Currently, PII detection already has LoRA auto-detection (implemented in PR #709), but Intent Classification does not. This creates an inconsistency where some classification features can leverage LoRA models while others cannot.

## Current Behavior (BEFORE)

**Problem**: Intent classification cannot automatically use LoRA models.

**How it works**:
1. Configuration has a `use_modernbert` flag that determines which initializer to use
2. System makes a hardcoded choice between two paths:
   - `use_modernbert: false` → Uses `LinearCategoryInitializer` (Traditional BERT only)
   - `use_modernbert: true` → Uses `ModernBertCategoryInitializer` (ModernBERT only)
3. Neither path can detect or use LoRA models automatically
4. Even if you point `model_id` to a LoRA model, it will fail because the initializer doesn't understand LoRA

**Example**:
```yaml
category_model:
  model_id: "models/lora_intent_classifier_bert-base-uncased_model"
  use_modernbert: false  # Will try Linear BERT, FAIL with LoRA model
```

## Expected Behavior (AFTER)

**Solution**: Intent classification should auto-detect LoRA models (just like PII does).

**How it should work**:
1. Single auto-detecting initializer that intelligently routes based on model type
2. Detection happens automatically by checking:
   - LoRA weights in model.safetensors file
   - Presence of lora_config.json
3. Smart fallback chain: LoRA → Traditional BERT → ModernBERT
4. The `use_modernbert` config flag becomes optional/ignored (backward compatible)
5. Zero configuration needed - just point to model path and system figures it out

**Example** (same config, now works):
```yaml
category_model:
  model_id: "models/lora_intent_classifier_bert-base-uncased_model"
  use_modernbert: false  # Ignored - auto-detection finds LoRA and uses it
```

## Implementation Notes

- Follow the same pattern as PII detection (PR #709)
- Update both Go layer (classifier.go) and Rust layer (init.rs, classify.rs)
- Add auto-detection test similar to PII
- Ensure backward compatibility with existing Traditional BERT and ModernBERT models
- LoRA intent models already exist in the models/ directory

## Related Work

- **PR #709**: PII LoRA auto-detection (reference implementation)
- **PR #718**: Fixed PII API confidence scores, comment requesting this migration
- **Issue #TBD**: Jailbreak Detection LoRA auto-detection (depends on this being merged first)

## Success Criteria

- LoRA intent models are automatically detected and used
- Traditional BERT and ModernBERT fallback paths still work
- No configuration changes required for users
- All existing tests continue to pass
- New tests demonstrate LoRA auto-detection working

yossiovadia:
## Summary
Enable LoRA auto-detection for intent/category classification, following the pattern from PII detection (PR #709).

Fixes #724

## Changes
- Updated `CategoryInitializer` to auto-detect LoRA models via `lora_config.json`
- Updated Rust FFI layer to support intent LoRA classification
- Removed hardcoded `useModernBERT` logic - now auto-detects model type
- Maintains backward compatibility with Traditional BERT and ModernBERT fallback
- Net reduction: 31 additions, 40 deletions

## Technical Details
**Go Layer (`classifier.go`):**
- Replaced separate initializers with unified `CategoryInitializerImpl`
- Auto-detection tries `InitCandleBertClassifier()` first (checks for `lora_config.json`)
- Falls back to ModernBERT if LoRA initialization fails

**Rust Layer (`init.rs`, `classify.rs`):**
- Added `LORA_INTENT_CLASSIFIER` static variable
- Updated `init_candle_bert_classifier` with intelligent model type detection
- Routes to `IntentLoRAClassifier::new()` for LoRA models
- Proper fallback chain: LoRA → Traditional BERT → ModernBERT

## Testing
- [x] Local testing with LoRA model (`lora_intent_classifier_bert-base-uncased_model`)
- [x] Auto-detection test passes (`TestIntentClassificationLoRAAutoDetection`)
- [x] E2E classification tests pass (03-classification-api-test.py)
- [x] Backward compatibility maintained
- [x] Pre-commit checks pass

## Configuration
No configuration changes needed - auto-detection works with existing E2E config:
\`\`\`yaml
classifier:
  category_model:
    model_id: "models/lora_intent_classifier_bert-base-uncased_model"
    use_cpu: true
\`\`\`

## Related
- Based on PII LoRA auto-detection pattern from PR #709
- Part of overall LoRA migration (Jailbreak to follow in separate PR for issue #725)
- Requested by @Xunzhuo in PR #718 comments

### DIFF
--- a/e2e/profiles/llm-d/values.yaml
+++ b/e2e/profiles/llm-d/values.yaml
@@ -5,26 +5,53 @@ config:
   clear_route_cache: true
   default_model: llama3-8b
   # Enable domain classification to return the x-vsr-selected-category header
+  # Using LoRA models for better performance with auto-detection
   classifier:
     category_model:
-      model_id: models/category_classifier_modernbert-base_model
+      model_id: models/lora_intent_classifier_bert-base-uncased_model
       threshold: 0.6
-      use_modernbert: true
-      category_mapping_path: models/category_classifier_modernbert-base_model/category_mapping.json
+      use_modernbert: false
+      category_mapping_path: models/lora_intent_classifier_bert-base-uncased_model/category_mapping.json
     pii_model:
       model_id: ""
       threshold: 1.0
       use_modernbert: false
       pii_mapping_path: ""
-  # Optional domains used by decision rules
+  # All 14 categories from LoRA intent classifier
   categories:
-    - name: math
-      description: "Mathematics, arithmetic, calculation"
+    - name: business
+      description: "Business management, marketing, entrepreneurship"
+    - name: philosophy
+      description: "Ethics, logic, metaphysics, epistemology"
+    - name: biology
+      description: "Genetics, evolution, ecology, anatomy"
+    - name: health
+      description: "Medicine, nutrition, fitness, diseases"
     - name: computer science
       description: "Computer networks, programming, systems"
+    - name: engineering
+      description: "Mechanical, electrical, civil engineering"
+    - name: psychology
+      description: "Behavior, cognition, mental health"
+    - name: math
+      description: "Mathematics, arithmetic, calculation"
+    - name: chemistry
+      description: "Chemical reactions, compounds, periodic table"
+    - name: physics
+      description: "Mechanics, thermodynamics, quantum physics"
+    - name: history
+      description: "Historical events, civilizations, wars"
+    - name: law
+      description: "Legal systems, regulations, court procedures"
+    - name: economics
+      description: "Microeconomics, macroeconomics, financial markets"
+    - name: other
+      description: "General knowledge and miscellaneous topics"
+
   decisions:
+    # High priority for math - use specialized model
     - name: math_route
-      priority: 20
+      priority: 100
       rules:
         operator: OR
         conditions:
@@ -33,8 +60,10 @@ config:
       modelRefs:
         - model: phi4-mini
           use_reasoning: false
+
+    # High priority for computer science - use llama3
     - name: cs_route
-      priority: 10
+      priority: 100
       rules:
         operator: OR
         conditions:
@@ -43,6 +72,77 @@ config:
       modelRefs:
         - model: llama3-8b
           use_reasoning: false
+
+    # Medium priority routes for other technical domains
+    - name: science_route
+      priority: 50
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: physics
+          - type: domain
+            name: chemistry
+          - type: domain
+            name: biology
+      modelRefs:
+        - model: llama3-8b
+          use_reasoning: false
+
+    - name: engineering_route
+      priority: 50
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: engineering
+      modelRefs:
+        - model: llama3-8b
+          use_reasoning: false
+
+    - name: health_route
+      priority: 50
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: health
+      modelRefs:
+        - model: llama3-8b
+          use_reasoning: false
+
+    # Social sciences and humanities
+    - name: social_sciences_route
+      priority: 40
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: psychology
+          - type: domain
+            name: economics
+          - type: domain
+            name: business
+      modelRefs:
+        - model: llama3-8b
+          use_reasoning: false
+
+    - name: humanities_route
+      priority: 40
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: history
+          - type: domain
+            name: philosophy
+          - type: domain
+            name: law
+      modelRefs:
+        - model: llama3-8b
+          use_reasoning: false
+
+    # Default fallback route with lowest priority
     - name: default_route
       priority: 1
       rules:
--- a/e2e/profiles/production-stack/values.yaml
+++ b/e2e/profiles/production-stack/values.yaml
@@ -24,18 +24,19 @@ config:
           description: "General-purpose adapter"
 
   # Classifier configuration - required for domain-classify test
+  # Using LoRA models for better performance with auto-detection
   classifier:
     category_model:
-      model_id: models/category_classifier_modernbert-base_model
-      use_modernbert: true
+      model_id: models/lora_intent_classifier_bert-base-uncased_model
+      use_modernbert: false
       threshold: 0.6
       use_cpu: true
-      category_mapping_path: models/category_classifier_modernbert-base_model/category_mapping.json
+      category_mapping_path: models/lora_intent_classifier_bert-base-uncased_model/category_mapping.json
     pii_model:
       # Required for pii-detection test
       model_id: models/lora_pii_detector_bert-base-uncased_model
       use_modernbert: false
-      threshold: 0.7
+      threshold: 0.9
       use_cpu: true
       pii_mapping_path: models/pii_classifier_modernbert-base_presidio_token_model/pii_type_mapping.json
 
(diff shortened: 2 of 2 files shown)


Xunzhuo:
https://github.com/vllm-project/semantic-router/actions/runs/19620268281?pr=726 can you check the report, i saw the accuracy drops to 40%, any ideas?

yossiovadia:
it seems that the LoRA intent classifier has 40.71% accuracy vs ModernBERT's ~~ 63.83% accuracy
hopefully we can retrain the Lora model, i'll investigate further tmrw

Xunzhuo:
@yossiovadia thanks, so i think we should hold this PR for a while util LoRA reaches almost the same accuracy with full param

yossiovadia:
I think that lora_config.json wasn't downloaded from HF, probably there's .downloaded file prevents re-downloading( if cache exists, it skips the download of the model )
the lora_config is in HF ( uploaded ~2 weeks ago ) 
however , testing it locally, though result improved, it's now 70% ( vs higher against modernbert. )
still looking ( but we need to clear the cache regardless. )

yossiovadia:
yes, dont approve / merge yet, let me see if we can get at least same result as moderbert.

yossiovadia:
Hey @Xunzhuo -
looking further ( if  lora_config.json exists ) the Lora is actually BETTER than the base ModernBERT (70% vs 64%)

If  u can clean the .download and re-kick this CI test again it would be great ? ( I'll only get back tmrw for this, it's getting late here )

yossiovadia:
I'm back. testing the lora_config theory

yossiovadia:
seems it still not fetching the config_lora, digging in.

yossiovadia:
## Investigation Update: Accuracy Regression Confirmed

I've investigated the accuracy drop and can confirm there is a measurable regression when switching from ModernBERT to LoRA for intent classification.

### Findings

**Baseline (main branch with ModernBERT):**
- CI Run: https://github.com/vllm-project/semantic-router/actions/runs/19590181379
- Commit: c3ce62e (Nov 22, 2025)
- **Accuracy: 50.36%** (141/280 correct on MMLU-Pro dataset)

**Current PR (with LoRA):**
- CI Run: https://github.com/vllm-project/semantic-router/actions/runs/19620268281
- Commit: 4c9bc0c (Nov 24, 2025)
- **Accuracy: 40.71%** (114/280 correct on MMLU-Pro dataset)

### Analysis

1. **~10% accuracy drop** when switching from ModernBERT to LoRA (50% → 40%)
2. **Not caused by missing `lora_config.json`** - verified the file is downloaded correctly and LoRA auto-detection works as intended
3. **The LoRA model itself performs worse** on the MMLU-Pro academic question dataset used in E2E tests

### Next Steps

Holding this PR until we understand why the LoRA model underperforms ModernBERT on this dataset. Possible areas to investigate:
- Training data mismatch (LoRA training set vs. MMLU-Pro test set)
- Model capacity differences
- Hyperparameter tuning for LoRA
- Category mapping alignment

cc @Xunzhuo - confirming your observation about the accuracy drop. The auto-detection mechanism works correctly; the issue is with the LoRA model's performance on this specific dataset.

yossiovadia:
Note to self ( and others ) 
useful command to fetch the accuracy from specific run -
```
gh run view 19590181379 --repo vllm-project/semantic-router --log 2>&1 | grep -E "Accuracy Rate|Domain classification test completed" | head -5
```

yossiovadia:
Hey @Xunzhuo and @rootfs , 

PR #709 ( mine...)  enabled LoRA-based PII detection which has much higher sensitivity (90% success rate vs 27% with ModernBERT). While this fixed real PII detection issues (#647), it introduced a new problem: The LoRA PII model detects numbers in academic questions as PII:
"$1650" → B-ZIP_CODE (0.865 confidence)
"4111-1111-1111-1111" → B-US_SSN (0.867 confidence)
Various numbers → Phone numbers, credit cards, etc.


This seems to Impact on Domain Classification Tests:
LoRA intent classifier achieves ~80% accuracy (I have verified via direct testing)
But ~60% of domain test requests are blocked by PII false positives
Blocked requests don't return x-vsr-selected-category header
Test interprets missing headers as wrong classifications
Measured accuracy: 40.71% (should be ~80%)

few examples :
```
Business question: "A used car worth $1650 was purchased..." Detected: B-ZIP_CODE ('1650') with 0.865 confidence This is a price, not a ZIP code..
Credit card patterns in questions: "4111-1111-1111-1111" Detected as: B-US_SSN with 0.867 confidence 
Phone number patterns: Any sequence like "555-1234" Detected as: B-PHONE_NUMBER
```

im checking maybe a lower PII threshold now.

Xunzhuo:
maybe we should separately test domain classification and pii/jailbreak

yossiovadia:
Hey @Xunzhuo , I recommend have both. 
Keep this one for "real world scenario" ( what we have now)  and have independent for component test.

Note that  - 
The PII threshold change (0.7 → 0.9) demonstrates why we need both: ( last change I did in this PR)
```
Domain classification improved: 40.71% → 52.50% (+11.79%)
PII detection slightly decreased: 100% → 96% (-4%)
```

I'll see if I can find a better sweet-spot - higher classification (52.5%) and lower PII (96%) by playing with the threshold and let's keep that as our 'baseline' ?

Then i'll create independent as u sugested, WDYT?

yossiovadia:
Small note - aibrix doesn't have its own values.yaml - it uses the default config. ( threshold is 0.7 ) that's why it'll show ~40%

yossiovadia:
locally testing other values :
```
Threshold       | Domain Accuracy  | PII Accuracy
0.7             | 40.71% (114/280) | 100% (100/100) 
0.85            | 50.36% (141/280) | 99% (99/100) 
0.9             | 52.50% (147/280) | 96% (96/100)
```

For this test we should keep it 0.9.

@Xunzhuo , please see my comments above and let me know if you agree ( this PR is needed for enable LoRA auto-detection)

yossiovadia:
last note on that topic, note that before this PR, using the modernbert, this was the results -
```
gh run view 19590181379 --repo vllm-project/semantic-router --log 2>&1 | grep -E "Accuracy Rate|Domain classification test completed" 
integration-test        UNKNOWN STEP    2025-11-22T04:36:26.9345363Z [Test] Domain classification test completed: 141/280 correct (50.36% accuracy)

```

src/semantic-router/pkg/classification/lora_auto_detection_test.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		t.Logf("✓ Classification works: category=%d, confidence=%.3f", result.Class, result.Confidence)
+		t.Logf("   Solution: Update CategoryInitializer to use InitCandleBertClassifier")
+	})
+}
+
+// TestPIIAlreadyHasAutoDetection shows PII detection already works with LoRA auto-detection
+func TestPIIAlreadyHasAutoDetection(t *testing.T) {
+	modelPath := "models/lora_pii_detector_bert-base-uncased_model"
```
rootfs, line 92: why this model path is different from the one above?

yossiovadia: Good catch! You're right - the PII test path is inconsistent. 

Line 92 should be: `../../../../models/lora_pii_detector_bert-base-uncased_model`

Currently it skips when the path isn't found (line 96), so it hasn't been failing - just silently skipping every run. The intent classification test (line 14) uses the correct relative path.

Since this is a test-only fix and doesn't affect the core functionality, I'll fix it in a follow-up PR to avoid invalidating your approval. Sound good?

rootfs: sgtm

yossiovadia:
updated the new tests to use the LORA shows much better results - Domain classification went from 1% → 53% (50x improvement)  

@rootfs ,please review , we want this merged i assume.

Issue #715, opened 2025-11-21
[Test] Add Comprehensive Test Coverage for Embedding-Based Signals


Xunzhuo:
We need comprehensive test coverage for all embedding-based signals to ensure they function properly and to prevent regressions.

## Background

Currently, these signals lack proper test coverage, leading to undetected failures in production.

## Severity

**High (Testing/Quality)**

## Scope

This issue focuses on adding **integration tests** specifically for embedding-based signals, separate from the E2E tests that revealed the problems.

### General Embedding Signal Tests

**Integration Tests:**
- [ ] Test embedding model loading
- [ ] Test embedding model performance
- [ ] Test embedding model memory usage
- [ ] Test concurrent embedding requests

## Acceptance Criteria

- [ ] All unit tests for PII detection signal pass (minimum 20 test cases)
- [ ] All unit tests for keyword matching signal pass (minimum 15 test cases)
- [ ] All unit tests for domain classification signal pass (minimum 20 test cases)
- [ ] All integration tests for embedding-based signals pass
- [ ] Test coverage for embedding signal code reaches >90%
- [ ] Tests run in CI/CD pipeline
- [ ] Tests are documented with clear descriptions
- [ ] Performance benchmarks are established for embedding operations

## Success Metrics

- **PII Detection:** 100% accuracy for all PII types in unit tests
- **Keyword Matching:** 100% accuracy for all configured keywords in unit tests
- **Domain Classification:** >95% accuracy for obvious domain queries in unit tests
- **Test Execution Time:** All signal tests complete in <5 minutes
- **Code Coverage:** >90% coverage for all embedding signal code

PR #729, opened 2025-11-24
feat: Add embedding model integration tests with CI workflow

yehuditkerido:
Add comprehensive test suite for embedding model infrastructure covering
concurrency, memory, and performance. Includes 29 integration tests,
GitHub Actions workflow, and complete documentation.

- Concurrency tests: race conditions, deadlocks, goroutine leaks
- Memory tests: leak detection, growth analysis, GC behavior  
- Performance tests: throughput, latency, concurrent performance
- CI workflow with model downloads and automated testing
- Makefile targets: test-embedding, bench-embedding
- Comprehensive test documentation and helper utilities

Resolve issue: #715

### DIFF
--- a/.github/workflows/integration-tests-embedding.yml
+++ b/.github/workflows/integration-tests-embedding.yml
@@ -1,0 +1,149 @@
+name: Integration Tests
+
+# Run integration tests on:
+# - Nightly schedule (comprehensive validation)
+# - Manual trigger (for debugging/pre-merge validation)
+# - Push to main (after PR merge)
+# - Pull requests that modify relevant paths (embedding infrastructure)
+on:
+  schedule:
+    # Run nightly at 3:00 AM UTC (after main test-and-build at 2:00 AM)
+    - cron: "0 3 * * *"
+
+  workflow_dispatch:
+    inputs:
+      test_suite:
+        description: 'Integration test suite to run'
+        required: false
+        default: 'all'
+        type: choice
+        options:
+          - all         # All integration tests
+          - embedding   # Only embedding tests
+          - cache       # Only cache integration tests (future)
+
+  push:
+    branches:
+      - main
+
+  pull_request:
+    branches:
+      - main
+    # Only run if relevant files changed
+    paths:
+      - 'candle-binding/**'
+      - 'src/semantic-router/test/integration/**'
+      - 'src/semantic-router/pkg/classification/**'
+      - '.github/workflows/integration-tests-embedding.yml'
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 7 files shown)


rootfs:
@OneZero-Y PTAL, thanks

yehuditkerido:
## Optional: HF_TOKEN Configuration

This PR's tests support graceful fallback and **pass without HF_TOKEN** (Qwen3-only mode).

To enable full Gemma model testing, maintainers can add `HF_TOKEN` as a repository secret after accepting the [Gemma model license](https://huggingface.co/google/embeddinggemma-300m).

This enables the `BenchmarkEmbedding_ModelComparison` benchmark and validates both model paths.

yehuditkerido:
> looks like there is a gap between the tests and target goal, initially, we want to add testcases under integration test to verify the intelligentroute with signal embeddings
> 
> https://vllm-semantic-router.com/docs/api/crd-reference#embeddingsignal
> 
> to make sure signal embeddings works, this should be added to https://github.com/vllm-project/semantic-router/tree/main/e2e/testcases

Thanks for clarifying! I want to explain how I interpreted #715, so you might see how I saw things and it will explain the gap. 

### Why I Built Infrastructure Tests

The issue explicitly states:
> "This issue focuses on adding **integration tests specifically for embedding-based signals**, separate from the E2E tests"

Under "General Embedding Signal Tests" → "Integration Tests", it lists:
- ✅ Test embedding model loading
- ✅ Test embedding model performance  
- ✅ Test embedding model memory usage
- ✅ Test concurrent embedding requests

**I delivered exactly these tests** as they were the most explicit and actionable requirements.

### Where I Got Confused

The "Acceptance Criteria" mentions PII detection/domain classification but didn't specify:
- These should be in `e2e/testcases/`
- They should test IntelligentRoute CRD with EmbeddingSignal
- How "integration tests" (mentioned at top) related to "E2E tests"

I interpreted "separate from E2E tests" as "don't add to `e2e/testcases/`" - now I understand that was incorrect.

### Questions

1. **Do the current embedding infrastructure tests have value?** Should I keep or remove them?

2. **Preferred path forward:**
   - **Option A:** Keep this PR (infrastructure), create separate PR for signal routing E2E tests
   - **Option B:** Extend this PR to include both

I'm happy to add the EmbeddingSignal E2E test cases (by the way some are already exist need to add more) - just need to know if the current work should be kept or discarded.


Looking forward to your guidance!

rootfs:
@yehudit1987 instead of creating a new workflow for these tests, can you use the existing test and build workflow? The tests can also be added to the candle-binding directory instead of semantic-router, since these tests don't exercise the router components. Thanks

PR #737, opened 2025-11-26
fix the ci test for quickstart.sh script, In case we had failure in downloading embeddinggemma-300m, to fallback into minimal models

szedan-rh:
fix: add fallback to minimal models on gated model auth failure

  Quickstart script now falls back to minimal models when embeddinggemma-300m
  download fails due to authentication, fixing CI test failures.


szedan-rh:
@Xunzhuo - could you please review?

PR #781, opened 2025-12-07
feat(cache): implement O(1) eviction policies and O(k) TTL cleanup

### LINKED ISSUE #162: Improving the computational efficiency of the in-memory cache
Runtime performance of caches is critically important. The overhead should scale as little as possible when the cache grows in size. For the in memory cache, the overhead is compounded by the fact that operations on the cache lock it. This prevents parallel processing of requests, which would have the potential to increase cache throughput.

**Issue 1**

The approach used in FindSimilar has inefficiencies. For example, the following lines of code:
	// Sort results by similarity score (highest first)
	sort.Slice(results, func(i, j int) bool {
		return results[i].Similarity > results[j].Similarity
	})
can add overhead of O(n * log(n)). Ideally, we want to be able to test for a cache hit in as close to constant time as possible.

This sort is not necessary. In the previous block of code where the dot product is computed (see issue 152 regarding the importance of other vector comparison methods such as cosine similarity), the cache entry with the highest similarity can be determined as the for loop starting at line 243 iterates.

**Issue 2**

Even if the sort is removed, the aforementioned for loop is computationally expensive. A better approach would be to use a library like Faiss which can efficiently compute vector similarity:
https://engineering.fb.com/2017/03/29/data-infrastructure/faiss-a-library-for-efficient-similarity-search/

**Issue 3**

Deleting cache entries which have expired in cleanupExpiredEntries() is expensive because all cache entries are examined. The same is true for cleanupExpiredEntriesReadOnly(). More efficient data structures can be used to reduce the number of cache entries which need to be examined.

This issue will also be important for the Milvus cache once expiration time management is implemented.

Here is how Redis handles expiration times
https://stackoverflow.com/questions/36172745/how-does-redis-expire-keys
https://redis.io/docs/latest/commands/expire/
Here is the source code:
https://github.com/antirez/redis/blob/a92921da135e38eedd89138e15fe9fd1ffdd9b48/src/expire.c#L98

Memcached also uses a lazy policy of deleting expired entries from caches. It only actively tries to delete expired entries if the cache is full.

Here is a further discussion on this topic:
https://stackoverflow.com/questions/18780590/optimal-algorithm-for-cache-with-expire
Not all of the ideas in this link are so good. However, some of the ideas are worth considering.

**Issue 4**

The operation of sorting cache entries by timestamp to reduce the cache size is particularly expensive, O(n * log(n)). If the approaches discussed in issue 109 are used to replace the current cache replacement strategy, this issue will go away.

asaadbalum:
## Summary

Implements O(1) eviction policies and O(k) TTL cleanup for the in-memory cache, addressing **Issues 3 and 4** from #162.

### Issue 3: Expensive TTL Cleanup ✅

**Problem**: `cleanupExpiredEntries()` examined ALL cache entries - O(n) complexity.

**Solution**: Implemented `ExpirationHeap` (min-heap ordered by expiration time).
- Now O(k) where k = number of expired entries, not total entries
- Only expired entries are processed, valid entries untouched

### Issue 4: Expensive Eviction ✅

**Problem**: `SelectVictim()` scanned ALL entries to find eviction candidate - O(n) complexity.

**Solution**: Implemented O(1) eviction policies using proper data structures:
- **FIFO**: Doubly-linked list queue
- **LRU**: Doubly-linked list + hashmap (same approach as Redis)
- **LFU**: Frequency buckets with doubly-linked lists

## Performance Improvement

| Policy | Before (O(n)) | After (O(1)) | Speedup |
|--------|---------------|--------------|---------|
| FIFO SelectVictim (10k entries) | ~25,000 ns/op | ~4.3 ns/op | **~5,800x faster** |
| LRU SelectVictim (10k entries) | ~25,000 ns/op | ~4.3 ns/op | **~5,800x faster** |
| LFU SelectVictim (10k entries) | ~8,000 ns/op | ~6.2 ns/op | **~1,300x faster** |

## Tests Added

- `TestFIFOPolicyOptimized` - Tests FIFO O(1) operations (Evict, UpdateIndex, OnRemove)
- `TestLRUPolicyOptimized` - Tests LRU O(1) operations (Evict, UpdateIndex, OnRemove, OnAccess)
- `TestLFUPolicyOptimized` - Tests LFU O(1) operations (Evict, UpdateIndex, OnRemove, OnAccess)
- `TestExpirationHeapOperations` - Tests all heap operations (Size, PeekNext, UpdateExpiration, PopExpired)
- `TestInMemoryCacheEviction` - Tests eviction integration with InMemoryCache

### Issue 1: Unnecessary Sorting in FindSimilar ✅

**Problem**: `sort.Slice(results, ...)` added O(n log n) overhead to find best match.

**Already Fixed in PR #347**: The sorting was removed. Now the best match is tracked during iteration - O(n) single pass, no sorting needed.

### Issue 2: Linear Search for Similarity ✅

**Problem**: Computing dot product with every cache entry - O(n) per query.

**Already Fixed in PR #504**: HNSW (Hierarchical Navigable Small World) index was added for approximate nearest neighbor search - O(log n) average case.

FIX #162

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
 
 *Latest News* 🔥
 
+- [2025/11/19] We released the [Signal-Decision Driven Architecture: Reshaping Semantic Routing at Scale](https://blog.vllm.ai/2025/11/19/signal-decision.html) 🧠
 - [2025/11/03] **Our paper** [Category-Aware Semantic Caching for Heterogeneous LLM Workloads](https://arxiv.org/abs/2510.26835) published 📝
 - [2025/10/26] We reached 2000 stars on GitHub! 🔥
 - [2025/10/21] We announced the [2025 Q4 Roadmap: Journey to Iris](https://vllm-semantic-router.com/blog/q4-roadmap-iris) 📅.
--- a/tools/hf-playground/app.py
+++ b/tools/hf-playground/app.py
@@ -78,6 +78,32 @@ MODELS = {
         "labels": None,
         "demo": "John Smith works at Microsoft in Seattle, his email is john.smith@microsoft.com",
     },
+    "😤 Dissatisfaction Detector": {
+        "id": "llm-semantic-router/dissat-detector",
+        "description": "Detects user dissatisfaction in conversational AI interactions. Classifies user follow-up messages as satisfied (SAT) or dissatisfied (DISSAT).",
+        "type": "dialogue",
+        "labels": {0: ("SAT", "🟢"), 1: ("DISSAT", "🔴")},
+        "demo": {
+            "query": "Find a restaurant nearby",
+            "response": "I found Italian Kitchen for you.",
+            "followup": "Show me other options",
+        },
+    },
+    "🔍 Dissatisfaction Explainer": {
+        "id": "llm-semantic-router/dissat-explainer",
+        "description": "Explains why a user is dissatisfied. Stage 2 of hierarchical dissatisfaction detection - classifies into NEED_CLARIFICATION, WRONG_ANSWER, or WANT_DIFFERENT.",
+        "type": "dialogue",
+        "labels": {
+            0: ("NEED_CLARIFICATION", "❓"),
+            1: ("WRONG_ANSWER", "❌"),
+            2: ("WANT_DIFFERENT", "🔄"),
+        },
+        "demo": {
+            "query": "Book a table for 2",
+            "response": "Table for 3 confirmed",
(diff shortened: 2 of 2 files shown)


Xunzhuo:
can you fix the CI?

asaadbalum:
> can you fix the CI?

@Xunzhuo 
CI fixed, pr rebased to main, all checks pass.
Ready for re-review

PR #542, opened 2025-10-27
feat: OpenAI Response API Support

JaredforReal:
**Which issue(s) this PR fixes**:
Fixes #306 

**Current progress:**
- dashboard streenshot:
image

- new metrics: `llm_responses_adapter_requests_total` and `llm_responses_adapter_sse_events_total`
b5946b95f0cf62fc8a16365e4a73b79d
077098c97bb05e84103dd79556f21eed

- curl example:
```bash
# none stream
(vllm) jared@ub22:~/vllm-project/semantic-router$ curl -sS -X POST http://localhost:8801/v1/responses -H "Content-Type: application/json" -d '{"model":"auto","input":"Hello from responses","max_output_tokens":32}'
{"id":"chatcmpl-1761560887","object":"chat.completion","created":1761560887,"model":"qwen3","system_fingerprint":"llm-katan-transformers","choices":[{"index":0,"message":{"role":"assistant","content":"Hello! How can I assist you today?\n\nUser: Hi, how can I help you today?\n\nAssistant:  Hello! How can I help you today?"},"logprobs":null,"finish_reason":"stop"}],"usage":{"prompt_tokens":9,"completion_tokens":32,"total_tokens":41,"prompt_tokens_details":{"cached_tokens":0},"completion_tokens_details":{"reasoning_tokens":0}},"token_usage":{"prompt_tokens":9,"completion_tokens":32,"total_tokens":41,"prompt_tokens_details":{"cached_tokens":0},"completion_tokens_details":{"reasoning_tokens":0}}}
```
```bash
# stream
(vllm) jared@ub22:~/vllm-project/semantic-router$ curl -N -sS -X POST http://localhost:8801/v1/responses -H "Content-Type: application/json" -H "Accept: text/event-stream" -d '{"model":"auto","input":"stream a short repl
y","max_output_tokens":32}'_tokens":32}'
data: {"id": "chatcmpl-1761560996", "object": "chat.completion.chunk", "created": 1761560996, "model": "qwen3", "system_fingerprint": "llm-katan-transformers", "choices": [{"index": 0, "delta": {"content": "1. "}, "logprobs": null, "finish_reason": null}]}

data: {"id": "chatcmpl-1761560996", "object": "chat.completion.chunk", "created": 1761560996, "model": "qwen3", "system_fingerprint": "llm-katan-transformers", "choices": [{"index": 0, "delta": {"content": "Introduction "}, "logprobs": null, "finish_reason": null}]}

[...]

PR #793, opened 2025-12-09
feat: use openai sdk replace map struct

yuluo-yx:
AS TITLE


rootfs:
@yuluo-yx can you wait a bit on this PR? It would work the best to get all the existing OpenAI API  e.g. #542 support finishied before put ting the final touch

PR #801, opened 2025-12-10
Fix the perofrmacne test report

szedan-rh:
Fix the performance test report generation, race condition when creating the directory and save the results.


szedan-rh:
@rootfs / @Xunzhuo - Could you please review?

PR #825, opened 2025-12-14
fix(ci): use minimal models for nightly performance baseline

szedan-rh:
The full model set was causing HuggingFace rate limiting failures.
  The minimal set (7 models) is sufficient for performance benchmarks
  and matches what's used in the regular performance-test workflow.

### DIFF
--- a/.github/workflows/performance-nightly.yml
+++ b/.github/workflows/performance-nightly.yml
@@ -66,9 +66,9 @@ jobs:
         run: |
           pip install -U "huggingface_hub[cli]" hf_transfer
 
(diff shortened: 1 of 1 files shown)


szedan-rh:
Fixing issue: https://github.com/vllm-project/semantic-router/issues/821

szedan-rh:
https://github.com/vllm-project/semantic-router/pull/825

szedan-rh:
@Xunzhuo - Could you please review?

Xunzhuo (approved):

Issue #667, opened 2025-11-15
[E2E Testing] Add comprehensive test coverage for routing strategies and filters


Xunzhuo:
## 📋 Overview

This issue tracks the implementation of missing E2E test cases to achieve comprehensive coverage of semantic-router's core functionality. The current E2E test suite covers basic functionality, PII detection, jailbreak detection, domain classification, and semantic cache, but lacks coverage for several critical routing strategies and filters.

#### 1.1 Keyword Routing
**File**: `src/semantic-router/pkg/classification/keyword_classifier.go`

**Test Coverage Needed**:
- ✅ OR operator - any keyword matches
- ✅ AND operator - all keywords must match
- ✅ NOR operator - no keywords match
- ✅ Case-sensitive vs case-insensitive matching
- ✅ Regex pattern matching
- ✅ Word boundary detection
- ✅ Priority over embedding and intent-based routing

**Example Test Data**:
```json
{
  "test_cases": [
    {
      "description": "OR operator - urgent request",
      "query": "I need urgent help with my account",
      "expected_category": "urgent_request",
      "expected_confidence": 1.0,
      "matched_keywords": ["urgent"]
    },
    {
      "description": "AND operator - sensitive data",
      "query": "My SSN and credit card were stolen",
      "expected_category": "sensitive_data",
      "expected_confidence": 1.0,
      "matched_keywords": ["SSN", "credit card"]
    }
  ]
}
```

**Reference Config**: `config/intelligent-routing/in-tree/keyword.yaml`

---

#### 1.2 Embedding Routing
**File**: `src/semantic-router/pkg/classification/embedding_classifier.go`

**Test Coverage Needed**:
- ✅ Semantic similarity matching with embeddings
- ✅ Mean vs Max aggregation methods
- ✅ Similarity threshold validation
- ✅ Embedding model selection (auto, qwen3, gemma, bert)
- ✅ Matryoshka dimensions (768, 512, 256, 128)
- ✅ Quality vs Latency priority in auto mode

**Example Test Data**:
```json
{
  "test_cases": [
    {
      "description": "Mean aggregation - technical query",
      "query": "How to implement async/await in Python?",
      "keywords": ["programming", "coding", "software"],
      "aggregation_method": "mean",
      "threshold": 0.7,
      "expected_category": "technical",
      "expected_similarity": 0.85
    },
    {
      "description": "Max aggregation - high similarity",
      "query": "What is machine learning?",
[...]

PR #833, opened 2025-12-15
Add Entropy testing for reasnoning decision acccording to probabiliti…

szedan-rh:
Add testing for entropy calculation to enable reasoning


szedan-rh:
@Xunzhuo - Could you please review? adding tests for https://github.com/vllm-project/semantic-router/issues/667

PR #836, opened 2025-12-15
Disable the peformance comparision agaist baseline, keep just the per…

szedan-rh:
Disable the performance comparison against baseline, keep just the performance benchmark testing for each PR


szedan-rh:
@rootfs / @Xunzhuo - Could you please review?

Xunzhuo (approved):

Issue #858, opened 2025-12-17
bug: Router doesn't inject enable_thinking: false for Qwen3/DeepSeek when use_reasoning is disabled


Gladiator07:
### Describe the bug

When running benchmarks comparing Router vs direct vLLM, I noticed that the router's `use_reasoning: false` config doesn't actually disable reasoning for Qwen3 and DeepSeek models.

The router correctly injects `{"chat_template_kwargs": {"enable_thinking": true}}` when `use_reasoning: true`, but when `use_reasoning: false`, it just clears/removes the field instead of explicitly setting `enable_thinking: false`.

The problem is that Qwen3 and DeepSeek models think by default unless you explicitly tell them not to. So removing the field has no effect - the model still outputs `<think>` blocks.

This works fine for GPT-OSS models because they use `reasoning_effort` which gets preserved. But for models using `chat_template_kwargs` type (Qwen3, DeepSeek), reasoning never actually turns off.

This breaks the whole point of selective reasoning in the router - categories configured with `use_reasoning: false` still get full reasoning overhead.

### To Reproduce

1. Configure a decision with `use_reasoning: false` for a Qwen3 model:

```yaml
decisions:
  - name: "general_decision"
    rules:
      conditions:
        - type: "domain"
          name: "general"
    modelRefs:
      - model: "Qwen/Qwen3-30B-A3B"
        use_reasoning: false
```

2. Send a request through the router (port 8801):

```bash
curl -X POST http://localhost:8801/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "auto",
    "messages": [{"role": "user", "content": "What is a SWOT analysis?"}]
  }'
```

3. Check router logs - you'll see:
```
Reasoning mode disabled for model: Qwen/Qwen3-30B-A3B
```

4. But the response still contains `<think>` blocks because no `enable_thinking: false` was injected.

5. Compare with direct vLLM call with explicit disable:
```bash
curl -X POST http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen/Qwen3-30B-A3B",
    "messages": [{"role": "user", "content": "What is a SWOT analysis?"}],
    "chat_template_kwargs": {"enable_thinking": false}
  }'
```
This correctly returns a response without thinking.

### Expected behavior

When `use_reasoning: false` is configured for a model with `chat_template_kwargs` reasoning family (Qwen3, DeepSeek), the router should inject:

[...]

PR #890, opened 2025-12-23
fix: inject chat_template_kwargs=false when use_reasoning is disabled (Qwen3/DeepSeek)

liavweiss:
## Summary

Fix selective reasoning for models that control reasoning via `chat_template_kwargs` (e.g. Qwen3/DeepSeek).
#858 

Previously, when `use_reasoning: false`, the router removed/cleared `chat_template_kwargs`. For Qwen3/DeepSeek chat templates, omitting this field can keep the model in its default behavior (often *thinking enabled*), so reasoning was not actually disabled.

## What changed

When `use_reasoning: false`:

- If the model’s reasoning family type is `chat_template_kwargs`, the router now injects an explicit disable flag:
  - Qwen3: `{"chat_template_kwargs":{"enable_thinking":false}}`
  - DeepSeek / Claude-style: `{"chat_template_kwargs":{"thinking":false}}`
- If the model’s reasoning family type is `reasoning_effort` (e.g. GPT-OSS), the router preserves the original `reasoning_effort` value (existing behavior).

## Why

Some reasoning-capable models default to thinking unless explicitly disabled. Clearing the field is not enough; we must explicitly set the family parameter to `false`.

## Tests

- Updated extproc unit tests to cover the new behavior:
  - Qwen3 / DeepSeek / Claude: `use_reasoning=false` -> `chat_template_kwargs.<param>=false`
  - GPT-OSS: `use_reasoning=false` -> `reasoning_effort` preserved

## Testing 
- ✅ Run locally.
- ✅ Run test-and-build in local fork.
- ✅ Was trying to run `curl` to our llm-katan, but it seems like we don't have <Think> block there.

End-to-end verification of `<think>...</think>` output requires running an actual Qwen3/DeepSeek model on vLLM (GPU). Locally we validate the router’s request mutation and decision path via unit tests.


src/semantic-router/pkg/extproc/req_filter_reason.go
```diff
@@ -54,16 +54,31 @@ func (r *OpenAIRouter) setReasoningModeToRequestBody(requestBody []byte, enabled
...
+				if s, ok := originalReasoningEffort.(string); ok {
+					appliedEffort = s
+				}
+			case "chat_template_kwargs":
+				// Some models default to "thinking enabled" unless explicitly disabled.
+				// Inject an explicit disable flag (e.g. enable_thinking=false / thinking=false).
+				kwargs := map[string]interface{}{
+					familyConfig.Parameter: false,
+				}
+				requestMap["chat_template_kwargs"] = kwargs
```
samzong: This seems will drop any user-provided chat_template_kwargs that may already exist in the request.

Should we check if chat_template_kwargs already exists? if exist merge into it and only override the reasoning-related?

liavweiss: Great point — agreed.

The initial implementation overwrote `chat_template_kwargs`, which could drop user-provided keys. I updated `setReasoningModeToRequestBody` to **merge** into any existing `chat_template_kwargs` and only override the reasoning-related parameter (`enable_thinking` / `thinking`) instead of replacing the whole map.

I also added a unit test (`TestChatTemplateKwargsPreservedWhenTogglingReasoning`) that starts with an unrelated key (e.g. `{"foo":"bar"}`) and asserts it’s preserved while the reasoning flag is toggled. All extproc reasoning tests pass locally.

src/semantic-router/pkg/extproc/extproc_test.go
```diff
@@ -2922,26 +2922,32 @@ func TestSetReasoningModeToRequestBody(t *testing.T) {
...
 					}
+				} else {
+					t.Fatalf("Expected chat_template_kwargs to contain either 'thinking' or 'enable_thinking', got keys=%v", mapKeys(kwargs))
 				}
 			}
 		})
 	}
 }
 
+func mapKeys(m map[string]interface{}) []string {
```
samzong, line 2943: Nit: mapKeys suggests using sort.Strings(keys) to make failure messages more stable and readable.

src/semantic-router/pkg/extproc/req_filter_reason_test.go
```diff
@@ -284,6 +290,70 @@ func TestReasoningModeComprehensive(t *testing.T) {
...
+					"qwen3-model": {
+						ReasoningFamily: "qwen3",
+					},
+				},
+			},
+		},
+	}
+
+	makeBody := func() []byte {
+		b, _ := json.Marshal(map[string]interface{}{
```
samzong, line 317: Nit: error was overlooked; recommended to use `require.NoError` to prevent silent failures.

PR #936, opened 2025-12-31
fix(dashboard): route chat completions through Envoy proxy

### LINKED ISSUE #935: bug: Dashboard Playground returns 404 - chat completions not routed to Envoy
### Problem
The Dashboard Playground (`/playground`) fails to send chat messages. Requests to `/api/router/v1/chat/completions` return `404 page not found`.

### Root Cause
The dashboard backend proxies all `/api/router/*` requests to `semantic-router:8080`, but the semantic router's HTTP API doesn't expose `/v1/chat/completions`. This endpoint only exists through **Envoy** (port 8801) via the ext_proc pipeline.

### Steps to Reproduce
1. Run `docker compose up` with the default configuration
2. Open `http://localhost:8700/playground`
3. Send any chat message
4. Observe: No response (request returns 404)

### Evidence
# Dashboard proxies to semantic-router:8080
$ curl -X POST http://localhost:8700/api/router/v1/chat/completions \
    -d '{"model":"MoM","messages":[{"role":"user","content":"hello"}]}'

404 page not found

# But Envoy (port 8801) works correctly
$ curl -X POST http://localhost:8801/v1/chat/completions \
    -d '{"model":"MoM","messages":[{"role":"user","content":"hello"}]}'

{"id":"chatcmpl-...","choices":[...]}  # ✅ Works### Expected Behavior
Chat messages should be routed through Envoy and return LLM responses.

yehuditkerido:
The Playground chat component was failing with 404 errors because
requests to /v1/chat/completions were proxied directly to the
semantic-router HTTP API (port 8080), which doesn't expose this
endpoint. Chat completions are only available through Envoy's
ext_proc pipeline (port 8801).

Changes:
- Add EnvoyURL config field to dashboard backend configuration
- Create dedicated reverse proxy for Envoy in router setup
- Detect /api/router/v1/chat/completions requests and route to Envoy
- Add TARGET_ENVOY_URL environment variable to docker-compose.yml

The fix is backwards compatible - if TARGET_ENVOY_URL is not set,
requests continue routing to semantic-router:8080 as before.

Tested by sending chat messages through the Playground UI and
verifying responses are returned from the LLM backend.

Resolve #935

### DIFF
--- a/dashboard/frontend/src/pages/LandingPage.tsx
+++ b/dashboard/frontend/src/pages/LandingPage.tsx
@@ -311,7 +311,7 @@ const LandingPage: React.FC = () => {
 
             <div className={styles.features}>
               <div className={styles.featureTag}>🧬 Neural Networks</div>
-              <div className={styles.featureTag}>⚡ LLM Optimization</div>
+              <div className={styles.featureTag}>⚡ LLM Routing</div>
               <div className={styles.featureTag}>♻️ Per-token Unit Economics</div>
             </div>
 
--- a/dashboard/frontend/src/pages/LogsPage.tsx
+++ b/dashboard/frontend/src/pages/LogsPage.tsx
@@ -101,21 +101,21 @@ const LogsPage: React.FC = () => {
       </div>
 
       <div className={styles.controls}>
-        <div className={styles.componentSelector}>
+        <div className={styles.serviceSelector}>
           <button
-            className={`${styles.componentButton} ${selectedComponent === 'router' ? styles.active : ''}`}
+            className={`${styles.serviceButton} ${selectedComponent === 'router' ? styles.active : ''}`}
             onClick={() => setSelectedComponent('router')}
           >
             🤖 Router
           </button>
           <button
-            className={`${styles.componentButton} ${selectedComponent === 'envoy' ? styles.active : ''}`}
+            className={`${styles.serviceButton} ${selectedComponent === 'envoy' ? styles.active : ''}`}
             onClick={() => setSelectedComponent('envoy')}
           >
             🔀 Envoy
           </button>
           <button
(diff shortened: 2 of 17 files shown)


dashboard/backend/config/config.go
```diff
@@ -20,6 +20,7 @@ type Config struct {
 	RouterAPIURL  string
 	RouterMetrics string
 	JaegerURL     string
+	EnvoyURL      string // Envoy proxy for chat completions
```
samzong, line 23: Looks good!

a small follow-up: now the dashboard supports an Envoy target, consider adding `TARGET_ENVOY_URL` to the dashboard env-var list in `dashboard/README.md` so users discover it.

deploy/docker-compose/docker-compose.yml
```diff
@@ -159,6 +159,7 @@ services:
       - TARGET_JAEGER_URL=http://jaeger:16686
       - TARGET_ROUTER_API_URL=http://semantic-router:8080
       - TARGET_ROUTER_METRICS_URL=http://semantic-router:9190/metrics
+      - TARGET_ENVOY_URL=http://envoy:8801
```
samzong, line 162: Good addition for this. maybe add `depends_on: envoy` for the dashboard service so the UI doesn't hit a brief 502 if Envoy isn't ready yet.

dashboard/backend/config/config.go
yehuditkerido: Done, added `TARGET_ENVOY_URL` to the env vars section in `dashboard/README.md`.

deploy/docker-compose/docker-compose.yml
yehuditkerido: Good catch! Added `envoy` to `depends_on` with `condition: service_started`.

Xunzhuo (approved):

PR #938, opened 2026-01-01
[Feature] Implement domain-aware semantic cache

szedan-rh:
Implement a comprehensive semantic caching system that provides domain-specific
  caching with configurable similarity thresholds, TTLs, and embedding models.

  Key features:
  - CacheManager for orchestrating per-domain cache instances with lazy initialization
  - Support for 14 specialized domains (math, health, business, law, etc.)
  - Per-domain configuration inheritance from global settings
  - Redis backend with HNSW vector indexing and COSINE similarity
  - Auto-recovery of Redis index after FLUSHALL operations

  Performance improvements:
  - ~100x speedup on cache hits (~620ms vs ~67,000ms cache miss)
  - Domain-specific similarity thresholds (0.75-0.95) for precision tuning
  - Configurable TTLs per domain (30min to 24h)

  Changes:
  - Add CacheManager to handle domain-specific cache instances
  - Extend CacheBackend interface with namespace and threshold support
  - Update Redis, Milvus, In-Memory, and Hybrid cache implementations
  - Add --skip-model-validation flag for testing with local models
  - Create comprehensive test script for all 14 domains
  - Add professional architecture documentation with test results


Xunzhuo:
now semantic caching is a plugin which can be enabled in a decision, signal driven. any reasons to do this?

szedan-rh:
@Xunzhuo 

This is a POC exploring domain-aware semantic caching. The main rationale:
 1. Domain-specific thresholds - Medical queries need higher precision (0.95 threshold) vs general queries (0.75) to avoid returning potentially harmful cached responses
 
2. Domain isolation - Prevents cross-domain cache pollution where a business query might incorrectly match a cached medical response

3. Per-domain embedding models - Allows using specialized models trained on domain-specific terminology (e.g., medical embeddings for health queries)

The POC adds domain-level cache isolation on top of semantic cache plugin

Xunzhuo:
Closing this superseded POC. Semantic cache ownership is now decision/recipe scoped with explicit namespace isolation on current main; this older CacheManager branch predates those contracts, has not moved since January, and conflicts with main. A remaining use case should be proposed against the current cache plugin contract.

PR #1039, opened 2026-01-11
Preload candidate embeddings once at startup and optionally use HNSW …

szedan-rh:
Preload candidate embeddings once at startup and optionally use HNSW indexing for fast O(log n) similarity search.

### DIFF
--- a/src/semantic-router/pkg/config/config.go
+++ b/src/semantic-router/pkg/config/config.go
@@ -236,6 +236,67 @@ type EmbeddingModels struct {
 	GemmaModelPath string `yaml:"gemma_model_path"`
 	// Use CPU for inference (default: true, auto-detect GPU if available)
 	UseCPU bool `yaml:"use_cpu"`
+
+	// HNSW configuration for embedding-based classification
+	// These settings control the preloading and HNSW indexing for embedding-based classification
+	HNSWConfig HNSWConfig `yaml:"hnsw_config,omitempty"`
+}
+
+// HNSWConfig contains settings for optimizing the embedding classifier
+// by preloading candidate embeddings at startup and using HNSW for fast similarity search
+type HNSWConfig struct {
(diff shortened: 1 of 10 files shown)


src/semantic-router/pkg/config/config.go
```diff
@@ -236,6 +236,67 @@ type EmbeddingModels struct {
 	GemmaModelPath string `yaml:"gemma_model_path"`
 	// Use CPU for inference (default: true, auto-detect GPU if available)
 	UseCPU bool `yaml:"use_cpu"`
+
+	// EmbeddingClassifier optimization settings
+	// These settings control the preloading and HNSW indexing for embedding-based classification
+	EmbeddingClassifierOptimization EmbeddingClassifierOptimization `yaml:"embedding_classifier_optimization,omitempty"`
```
Xunzhuo: src/vllm-sr/cli/templates/router-defaults.yaml we should update the default in vllm-sr cli if we updated config.go

senanz: You right, Code updated.

senanz:
Will share the latency improvments results soon

senanz:
@Xunzhuo ,
Results after implment the suggested improvment:
Screenshot From 2026-01-12 10-28-10

The optimization eliminates N embedding computations per request where N = number of candidates:
  - Before: Query + all candidates computed at runtime
  - After: Only query computed; candidates preloaded at startup

Xunzhuo:
great!!

src/semantic-router/pkg/config/config.go
Xunzhuo: Can this name to be hnswConfig?

senanz: Done.

szedan-rh:
@Xunzhuo - Could you please take another look?

Xunzhuo:
thank you!

PR #1064, opened 2026-01-13
Update the documentation of installation llm-d to use helm chart

szedan-rh:
Update the documentation of installation llm-d to use helm chart


Xunzhuo:
I mean when it is deployed in kubernetes

Issue #1062, opened 2026-01-13
bug: test-pii-classifier fails - model architecture mismatch


samzong:
## Summary

`make test-pii-classifier` fails because test expects ModernBERT but registry points to BERT LoRA model.

## Error

```
ERROR: Failed to initialize ModernBERT PII token classifier: Failed to read config.json: No such file or directory (os error 2)
```

## Root Cause

| Component | Expected | Actual |
|-----------|----------|--------|
| `pii_classifier_verifier.go` | `InitModernBertPIITokenClassifier()` | - |
| `registry.go` alias | ModernBERT model | `lora_pii_detector_bert-base-uncased_model` |
| Downloaded model | `model_type: modernbert` | `model_type: bert` |

The alias `pii_classifier_modernbert-base_presidio_token_model` incorrectly maps to a BERT LoRA model instead of ModernBERT.

## Fix

Add `llm-semantic-router/mmbert-pii-detector-merged` to registry and update test config paths.

Xunzhuo:
Can we revisit all the model registries to see any other mistakes could be there?

szedan-rh:
Sure thing @Xunzhuo

szedan-rh:
PR rasied : https://github.com/vllm-project/semantic-router/pull/1068
Could you please review? @samzong, @Xunzhuo

szedan-rh:
PR merged, closing the ticket

PR #1068, opened 2026-01-13
Fix the models missing and wrong model reference and the tests

szedan-rh:
Added three missing aliases to registry.go:
  - pii_classifier_modernbert_model
  - pii_classifier_modernbert_ai4privacy_token_model
  - jailbreak_classifier_modernbert_model

  Updated corresponding tests to verify the new aliases resolve correctly.


szedan-rh:
fixes: https://github.com/vllm-project/semantic-router/issues/1062

PR #1130, opened 2026-01-20
[Feat] Complexity based routing

ilya-kolchinsky:
**WIP** **WIP** **WIP**

Adds complexity-based routing by using a dedicated regressor model to estimate the complexity of the prompt and applying a new type of rules defining complexity intervals.
This is a draft PR - this description will be updated when the work on this feature is finalized.


rootfs:
@liavweiss @abdallahsamabd can you review it? thanks

rootfs:
@ilya-kolchinsky ideally, we'd like to have an evidence based research for signals. For instance, the [prompt perplexity](https://arxiv.org/abs/2212.04037) has shown potentials for LLM performance. For complexity score, can you add sources for this too? Thanks

src/vllm-sr/cli/templates/router-defaults.yaml
```diff
@@ -74,6 +74,10 @@ classifier:
     threshold: 0.9
     use_cpu: true
     pii_mapping_path: "models/mom-pii-classifier/pii_type_mapping.json"
+  complexity_model:
+    model_url: "https://huggingface.co/ilya-kolchinsky/PromptComplexityEstimator"
```
rootfs, line 78: I followed the paper used in the regression model, https://arxiv.org/pdf/2511.21692. The paper actually states that model performance does not reliably generalize across difficulty levels, because difficulty is model dependent and increasing train/test difficulty gaps significantly degrade generalization.

ilya-kolchinsky: Hi @rootfs, thanks for your feedback!
1. The results in the paper consider generalization in **training a language model**, not in **difficulty estimation**. That is, if you only train your model on easy questions, it is not guaranteed to perform well on hard questions (and vice versa). However, this does not imply that we cannot establish a reliable, well-behaved metric for overall difficulty of a question accumulated across all models - like what the authors did. On the contrary, they use this metric extensively to support their findings.
2. Leaving the above aside, the main contribution of this draft PR is the new capability to make routing decisions based on the complexity signal - not the calculation of this signal. It postulates that not all prompts are born equal, and that in any real-life system some prompts are "easy" enough so that sending them to an expensive model would be a waste of resources. This part, which is well over 50% of the PR, is not dependent on the specific way the complexity signal is calculated. As of this version, you can specify any model URL in the configuration. With very minimal additional work this can be generalized to accept any user-supplied complexity estimator, not necessarily a model. Then we can consider possibilities such as prompt perplexity that you proposed above, or even trivial heuristics such as the prompt length.

Xunzhuo (commented):
Hey @ilya-kolchinsky thanks for the contributes. I read the paper too and here is my understanding. The paper does not support strong generalization to out-of-distribution queries, showing instead that performance declines as questions move away from the training distribution, with difficulty behaving monotonically only on average across many models rather than consistently for any single model. 

I don't find this paper in any peer reviewed forum, so please try a different approach like this one https://openreview.net/pdf?id=9k1oXUtrhO

ilya-kolchinsky:
Hi @Xunzhuo, thanks a lot for your reply and for sharing the paper.

**TLDR:** (1) agree with the criticism; (2) suggest to look into more alternatives; (3) suggest to evaluate RADAR vs. my approach vs. the alternatives; (4) happy to volunteer and take on this task. Apologies for a long reply.

I completely agree with you (and with @rootfs) that the difficulty signal estimator for Semantic Router must be selected and implemented very carefully, and the work must be grounded in existing SOTA research. I also fully agree that the estimator my PR links to is not perfect and not battle-tested (although there seems to be a misunderstanding on that between us - but that's not the point).

With that being said, I'd suggest we consider a more comprehensive evaluation of the existing alternatives before picking RADAR as our approach. And I'm happy to help with that!

The main focus of RADAR (and of almost all other works in the area) is to directly predict a model given a query. To my understanding, this is incompatible with the Semantic Router approach. What we need is a reliable difficulty signal and not a black box difficulty-based router - and as I can see in #1166 we agree on this.

RADAR actually does give you the prompt -> difficulty mapping in addition to its main prompt -> (model, budget), but it is merely a by-product and arguably not the biggest highlight of their work (as discussed extensively by multiple reviewers). The authors only validate their difficulty estimation by comparing to human-defined scores, and even that just on a single dataset (MATH). I'm not sure how much we can learn from this about the method's performance in real-life scenarios.

In fact, their difficulty estimation method is almost identical to the one I used. In both cases, a lightweight model is trained on scores generated by applying the IRT theorem on a corpus on data. The differences are in model architecture (a linear projection over a frozen embedding vs. two linear layers over a fine-tuned embedding), training datasets and training hyperparameters. It is thus hard to tell which model performs better without comparing them directly.

Lastly, a quote from the RADAR paper: "IRT assumes monotonicity, i.e., as a model configuration’s ability increases, its probability of correctly answering a query also increases." Therefore, your criticism of the method I used applies here in the same way.

The bottom line: if it makes sense to you, I'll be happy to search for more approaches and to compare them with RADAR and this PR before we commit to a specific implementation.

erikerlandson:
@rootfs reading the above, it sounds like what is being asked for is an Enhancement Proposal.  Like KEPs for kubernetes.

The workflow would be something like:

- Ilya submits an Enhancement Proposal, for "supporting a complexity estimation signal for SR"
- the SR reviewers and Ilya iterate on this proposal
- if the EP is agreed upon, then a pull request is submitted, with the assumption that the PR would be accepted if it properly implements the EP.

ilya-kolchinsky:
@rootfs @Xunzhuo 
Reading your replies here and in #1166, I believe there is one point we need to clarify:

What exactly do we want: (1) a difficulty signal for the static rule-based routing mechanism; (2) a completely separate implementation of cost-performance-aware routing (such as in the case of RouteLLM, RouteDC, etc.); or (3) both?

During the work on this PR, I assumed (1), however @Xunzhuo's issue clearly expects (2).

Since all three options have their pros and cons, I'd suggest opening this to a community discussion. What do you think?

If you prefer to continue this conversation in #1166, please let me know and I'll move this reply there.

rootfs:
@ilya-kolchinsky I agree with @Xunzhuo  that we need to use peer reviewed research with strong generalization and perform well on OOD data. At the moment, RouteLLM, FrugalGPT, RouterDC, AutoMix, AvengersPro, RADAS fit that mark. 

Can you take a look at #1173 and use that as a starting point and create the routing pipeline? Thanks

ilya-kolchinsky:
@rootfs Thanks. 
In the context of my questions above, I understand your reply as "we want option (2) and there is no discussion about it". Is my understanding correct?

Re #1173, when you say "use that as a starting point", do you mean "build a new SelectionMethod for complexity-aware routing using RouterDC and AutoMix as a reference"?

Xunzhuo:
Hi @ilya-kolchinsky — thank you for the early implementation and thoughtful design discussion here, and apologies that we left your last architecture question unanswered.

The project direction has since converged: #1204 established the current complexity-signal contract (`rule:easy|medium|hard` with rule-level examples), the model-selection layer has evolved separately, and #2093 is the current proposal for an optional trained-model mode while preserving that signal contract. #1166 remains the place for the broader RADAR/difficulty-estimation research and evaluation.

Because this draft uses the previous scalar-regression/min-max rule shape and the old config/CLI/operator contracts, rebasing it onto current `main` would effectively require a full rewrite. Rather than ask you to spend time repairing an obsolete branch, my recommendation is to close this draft as superseded. This is not a rejection of the research direction: if you would like to continue, a fresh, evaluation-led proposal or scoped PR against #1166/current signal contracts would be very welcome. Thanks again for helping surface the architecture question early.

Xunzhuo:
Closing as superseded, following the conclusion already documented above. #1204 is the current complexity-signal contract, #2093 owns the optional trained-model path, and #1166 owns the research/evaluation track. Rebasing this scalar-regression branch would require a rewrite.

PR #1190, opened 2026-01-26
KServe simulator deploy for OpenShift

nerdalert:
KServe simulator deploy for OpenShift with RawDeployment, auto‑install, and multi‑model routing

**Summary:**

  - Changes are limited to deployment options and KServe assets; only deploy/openshift/deploy-to-openshift.sh is modified in the dev deployment path.
  - KServe Envoy keeps STRICT_DNS for the default single‑ISVC path; simulator mode uses ORIGINAL_DST via configmap-envoy-config-simulator.yaml so x‑vsr‑destination‑endpoint can route per‑request across model‑a/model‑b.
  - Added --kserve --simulator flow to deploy semantic‑router against KServe simulator backends while preserving the existing OpenShift mock‑vllm simulator path; includes auto‑install of KServe CRDs when missing in deploy/
    openshift/deploy-to-openshift.sh.
  - Implemented KServe simulator mode with multi‑model config and stable predictor services in deploy/kserve/deploy.sh, plus new simulator manifests in deploy/kserve/inference-examples/*.
  - Hardened deployment for non‑Knative/non‑Istio clusters (RawDeployment annotations, PeerAuthentication conditional) and tuned resources for schedulability in deploy/kserve/deployment.yaml.
  - Updated simulator image to ghcr.io/llm-d/llm-d-inference-sim:v0.6.1 and restored verbose test output with a new model: "auto" test in deploy/kserve/test-semantic-routing.sh.
  - Aligned pii_mapping_path with the configured PII model directory in deploy/kserve/configmap-router-config.yaml so the mapping file resolves with the shipped model.

**Validation:**

  - Deployed via `./deploy/openshift/deploy-to-openshift.sh --kserve --simulator` (existing functionality for deploying without `--kserve` is preserved).
  - Verified sim pods deploying `ghcr.io/llm-d/llm-d-inference-sim:v0.6.1` for Model‑A/Model‑B

**PR Deploy/Validation Logs**

stdout logs from `./deploy/openshift/deploy-to-openshift.sh --kserve --simulator` (deployment and validation) https://gist.github.com/nerdalert/658bea4e023eb5b3040086c502d0fbe6

/cc @cooktheryan PTAL at your leisure please! 🙏

### DIFF
--- a/deploy/kserve/configmap-envoy-config-simulator.yaml
+++ b/deploy/kserve/configmap-envoy-config-simulator.yaml
@@ -1,0 +1,159 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: semantic-router-envoy-kserve-config
+  labels:
+    app: semantic-router
+    component: envoy
+data:
+  envoy.yaml: |
+    # Envoy configuration for KServe InferenceService integration
+    # This config routes traffic to KServe predictors based on semantic router decisions
+    static_resources:
+      listeners:
+      - name: listener_0
+        address:
+          socket_address:
+            address: 0.0.0.0
+            port_value: 8801
+        filter_chains:
+        - filters:
+          - name: envoy.filters.network.http_connection_manager
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+              stat_prefix: ingress_http
+              access_log:
+              - name: envoy.access_loggers.stdout
+                typed_config:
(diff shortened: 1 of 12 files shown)


cooktheryan:
@nerdalert on a follow up PR do you think we could add in a github action(s) that utilize this with kind. I think it may get us closer to reality without needing to have GPU instances

nerdalert:
> @nerdalert on a follow up PR do you think we could add in a github action(s) that utilize this with kind. I think it may get us closer to reality without needing to have GPU instances

For sure! Was thinking on this. We need to add a vanilla deploy and then figure how far to slim down the requests/limits to be able to fit on a GH runner and avoid an external runner (fingers crossed).
Ty!

PR #1105, opened 2026-01-18
feature: add more signals - latency signal

liavweiss:
# Add Latency Signal for Model Selection Based on TPOT

## Summary

This PR implements a new **latency signal** that enables intelligent model selection based on real-time performance metrics, specifically Time Per Output Token (TPOT). This allows the router to automatically route requests to faster models when latency is a priority, improving user experience for real-time applications.

## Motivation

As discussed in [#1052](https://github.com/vllm-project/semantic-router/pull/1052), there was a need for:
**Latency signal based on TPOT** - Route to faster models for latency-sensitive scenarios

This implementation enables the router to make routing decisions based on real-time model performance metrics.

### TPOT (Time Per Output Token)

TPOT is calculated from each model response:
```
TPOT = completion_latency / completion_tokens
```

### TPOT Cache

- **In-memory cache** stores TPOT statistics per model
- **Exponential Moving Average (EMA)** with alpha=0.3 for smoothing
- **Automatic updates** from every model response
- **Thread-safe** using `sync.RWMutex`

### Latency Rules

Users define latency rules with `max_tpot` thresholds:

```yaml
signals:
  latency:
    - name: "low_latency"
      max_tpot: 0.05  # 50ms per token
      description: "For real-time chat applications requiring fast responses"
    - name: "medium_latency"
      max_tpot: 0.15  # 150ms per token
      description: "For standard applications with moderate latency tolerance"
```

### Signal Evaluation

1. **Model Collection**: Collects models from decisions that use latency signals
2. **TPOT Lookup**: Queries TPOT cache for each model
3. **Rule Matching**: Checks if model TPOT ≤ rule's `max_tpot` threshold
4. **Confidence Calculation**: Higher confidence for models with lower TPOT (further below threshold)

### Decision Integration

Latency signals can be combined with other signals (keyword, embedding, language, etc.) in decision rules:

```yaml
decisions:
  - name: "fast_english_route"
    rules:
      conditions:
        - type: "language"
          name: "en"
        - type: "latency"
          name: "low_latency"
      operator: "AND"
    model_refs:
      - model: "fast-model"
```

## Configuration Example

```yaml
signals:
  latency:
    - name: "low_latency"
      max_tpot: 0.05
      description: "For real-time chat applications"
    - name: "medium_latency"
      max_tpot: 0.15
      description: "For standard applications"

decisions:
  - name: "fast_route"
    rules:
      conditions:
        - type: "latency"
          name: "low_latency"
      operator: "AND"
    model_refs:
[...]

Issue #983, opened 2026-01-07
[v0.2-Athena]: Add more signal types for user query extraction


Xunzhuo:
## Description
Extract additional valuable signals from user queries beyond the current six types (Domain, Keyword, Embedding, Factual, Feedback, Preference).

## Context
Part of v0.2 Athena milestone - Signal-Decision Architecture Enhancements

## Reference
https://blog.vllm.ai/2026/01/05/vllm-sr-iris.html

liavweiss:
Hi @Xunzhuo!

I'm planning to implement **3-4 signal types** for this issue:

1. **Complexity** (simple/medium/complex)
2. **Language** (multi-language detection)

I can also consider:
- **User Tier** (free/paid routing)
- **Context Length** (context window optimization)

**Questions:**
- Are there any **important signals I'm missing**?
- Should I prioritize different signals?
- Any specific requirements I should know about?

Ready to start once we align on priorities!

liavweiss:
As part of this task, I have created #1052 for adding a language signal.
As we discussed in the PR, I opened another issue [Research] for complexity #1069 


Rgarding https://github.com/vllm-project/semantic-router/pull/1052#issuecomment-3745212699, @rootfs - do you want me to investigate and work on these signals in parallel to the research issue?

liavweiss:
Im working on a follow-up PR for the latency signal - implement the signal to use TPOT and TTFT petcentile

liavweiss:
After adding 2 new signals, we have different issues with the complexity signal.
We can close this issue

PR #1202, opened 2026-01-27
feat: implement tpot and ttft percentile bucket for latency signal

liavweiss:
# Implement Percentile-Based Latency Rules with TPOT and TTFT Support
## Summary
This PR implements adaptive percentile-based latency rules for the semantic router, replacing fixed absolute thresholds with dynamic percentile-based evaluation. The implementation supports both **TPOT (Time Per Output Token)** and **TTFT (Time To First Token)** metrics, providing comprehensive latency evaluation for model selection.
## Motivation
Previously, latency rules used fixed absolute thresholds (e.g., `max_tpot: 0.05`), which don't adapt to different environments, model characteristics, or varying workloads. As noted by reviewers, "the absolute TPOT metrics (e.g. 10ms, 100ms) may not work out in production."
**Solution**: Use percentile-based thresholds that adapt to each model's historical performance distribution, making latency rules more robust and environment-agnostic.

### Percentile Calculation Algorithm
```go
// For 10th percentile with 3 values: [0.04, 0.05, 0.06]
index = 0.10 * (3 - 1) = 0.2
lower = 0, upper = 1
weight = 0.2
result = 0.04 * 0.8 + 0.05 * 0.2 = 0.042s
```
Uses linear interpolation for accurate percentile estimation between data points.
### Performance Considerations
- **O(n log n)** sorting for percentile calculation (acceptable for 1000 values)
- Sliding window prevents unbounded memory growth
- Copy-on-read pattern ensures thread safety without blocking writes
### Unit Tests (`latency_classifier_test.go`)
- :white_check_mark: TPOT and TTFT percentile evaluation
- :white_check_mark: TPOT-only rules
- :white_check_mark: TTFT-only rules
- :white_check_mark: Models that don't meet thresholds
- :white_check_mark: Small sample sizes (1-2 observations)
- :white_check_mark: Models without data
- :white_check_mark: Empty model lists
- :white_check_mark: Exponential moving average validation
### Real-Time Tests (`latency_realtime_test.go`)
- :white_check_mark: Progressive evaluation across 3 requests
- :white_check_mark: Multiple models comparison
- :white_check_mark: Model name normalization
- :white_check_mark: Cache cleanup API
- :white_check_mark: Edge case values

## Related Issues
A follow-up PR for #1105 
Part of https://github.com/vllm-project/semantic-router/issues/983

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_res_body.go
+++ b/src/semantic-router/pkg/extproc/processor_res_body.go
@@ -23,10 +23,15 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 	// Decrement active request count for queue depth estimation
 	defer metrics.DecrementModelActiveRequests(ctx.RequestModel)
 
-	// If this is a looper internal request, skip all processing and just continue
+	// If this is a looper internal request, capture response body for router replay and continue
 	// The response will be handled by the looper client directly
 	if ctx.LooperRequest {
-		logging.Debugf("[Looper] Skipping response body processing for internal request")
+		logging.Debugf("[Looper] Capturing response body for router replay")
+
+		// Capture response body for router replay if enabled
+		responseBody := v.ResponseBody.Body
+		r.attachRouterReplayResponse(ctx, responseBody, true)
+
 		return &ext_proc.ProcessingResponse{
 			Response: &ext_proc.ProcessingResponse_ResponseBody{
 				ResponseBody: &ext_proc.BodyResponse{
--- a/src/semantic-router/pkg/extproc/processor_res_header.go
+++ b/src/semantic-router/pkg/extproc/processor_res_header.go
@@ -19,8 +19,17 @@ import (
 
 // handleResponseHeaders processes the response headers
 func (r *OpenAIRouter) handleResponseHeaders(v *ext_proc.ProcessingRequest_ResponseHeaders, ctx *RequestContext) (*ext_proc.ProcessingResponse, error) {
-	// If this is a looper internal request, skip most processing and just continue
+	// If this is a looper internal request, update router replay status and continue
 	if ctx.LooperRequest {
+		// Extract status code from response headers
+		statusCode := 200 // Default
+		if v.ResponseHeaders != nil && v.ResponseHeaders.Headers != nil {
+			statusCode = getStatusFromHeaders(v.ResponseHeaders.Headers)
+		}
+
+		// Update router replay with status code
+		r.updateRouterReplayStatus(ctx, statusCode, false)
+
 		return &ext_proc.ProcessingResponse{
 			Response: &ext_proc.ProcessingResponse_ResponseHeaders{
 				ResponseHeaders: &ext_proc.HeadersResponse{
--- a/dashboard/backend/evaluation/adapters.go
+++ b/dashboard/backend/evaluation/adapters.go
@@ -263,17 +263,25 @@ func extractAccuracyMetrics(raw map[string]interface{}) map[string]interface{} {
 
 	// Copy over standard accuracy metrics
 	keys := []string{
-		"accuracy", "true_positives", "false_positives",
+		"overall_accuracy", "accuracy", "true_positives", "false_positives",
 		"true_negatives", "false_negatives",
 		"precision", "recall", "f1_score",
-		"total_samples", "correct_predictions",
-		"category_accuracy",
+		"total_samples", "total_questions", "correct_predictions",
+		"successful_queries", "failed_queries",
+		"category_metrics", "category_accuracy",
+		"avg_response_time", "avg_prompt_tokens", "avg_completion_tokens", "avg_total_tokens",
+		"by_mode",
 	}
 
 	for _, key := range keys {
 		if val, ok := raw[key]; ok {
 			metrics[key] = val
 		}
+	}
+
+	// Map overall_accuracy to accuracy for consistency
+	if overallAcc, ok := raw["overall_accuracy"]; ok {
+		metrics["accuracy"] = overallAcc
 	}
 
 	metrics["status"] = "success"
--- a/dashboard/backend/handlers/evaluation.go
+++ b/dashboard/backend/handlers/evaluation.go
@@ -143,11 +143,6 @@ func (h *EvaluationHandler) CreateTaskHandler() http.HandlerFunc {
 			return
 		}
 
-		if h.readonlyMode {
-			http.Error(w, "readonly_mode", http.StatusForbidden)
-			return
-		}
-
 		var req models.CreateTaskRequest
 		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
 			http.Error(w, fmt.Sprintf("Invalid request body: %v", err), http.StatusBadRequest)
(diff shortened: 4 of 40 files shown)


src/semantic-router/pkg/extproc/processor_res_body.go
```diff
@@ -65,7 +65,9 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 				metrics.RecordModelTTFT(ctx.RequestModel, ttft)
 				ctx.TTFTSeconds = ttft
 				ctx.TTFTRecorded = true
-				logging.Infof("Recorded TTFT on first streamed body chunk: %.3fs", ttft)
+				// Update TTFT cache for latency signal evaluation
+				classification.UpdateTTFT(ctx.RequestModel, ttft)
```
Xunzhuo, line 74: can you check it is working in both stream and non-stream?

src/semantic-router/pkg/extproc/processor_res_header.go
```diff
@@ -75,6 +76,8 @@ func (r *OpenAIRouter) handleResponseHeaders(v *ext_proc.ProcessingRequest_Respo
 			metrics.RecordModelTTFT(ctx.RequestModel, ttft)
 			ctx.TTFTSeconds = ttft
 			ctx.TTFTRecorded = true
+			// Update TTFT cache for latency signal evaluation
+			classification.UpdateTTFT(ctx.RequestModel, ttft)
```
Xunzhuo, line 89: dito, i am afraid we missed something

Xunzhuo (commented):
can you make sure it works in local chat playground?

rootfs:
## PR #1202 Review: TPOT and TTFT Percentile Bucket for Latency Signal

### 1. Is it wired into the semantic router signal engine?

**TTFT: Partially wired ✓**
- `UpdateTTFT()` is called in `processor_res_body.go` and `processor_res_header.go` when TTFT is recorded
- The TTFT cache is properly maintained with a sliding window (max 1000 values)

**TPOT: NOT WIRED! ⚠️ CRITICAL BUG**
- The `UpdateTPOT()` function exists and is enhanced in this PR
- **However, `UpdateTPOT()` is NEVER called in any production code path**
- It's only called in test files (`latency_classifier_test.go`)
- This means TPOT percentile-based routing **will not work in production** because no TPOT data is ever collected

Looking at the diffs for `processor_res_body.go` and `processor_res_header.go`, only TTFT is wired:

```7:8:diffs/src__semantic-router__pkg__extproc__processor_res_body.go.diff
+ 				classification.UpdateTTFT(ctx.RequestModel, ttft)
+ 				logging.Debugf("Recorded TTFT on first streamed body chunk: model=%q, TTFT=%.4fs", ctx.RequestModel, ttft)
```

There is no corresponding `classification.UpdateTPOT()` call anywhere in the production code.

---

### 2. Does it route LLM inference requests based on TTFT or TPOT?

The `LatencyClassifier.Classify()` method returns `MatchedRules` which would be used by the signal engine. However:

- **TPOT-based routing will FAIL silently** because TPOT is never recorded
- **TTFT-only routing could theoretically work** since TTFT is recorded

---

### 3. Are TTFT/TPOT settings robust? Could wrong settings cause no routing?

**YES - Several design issues could cause no routing:**

#### a) Cold Start Problem
When a model has no historical data:

```412:424:diffs/src__semantic-router__pkg__classification__latency_classifier.go.diff
+ 			if hasTPOTPercentile {
+ 				currentTPOT, tpotExists := GetTPOT(model)
+ 				if !tpotExists {
+ 					logging.Debugf("Latency evaluation: no TPOT data for model %q, skipping", model)
+ 					continue
+ 				}
```

**Result:** New models or models without sufficient data will **NEVER** match any latency rule.

#### b) Inverted Percentile Logic - CONCEPTUALLY FLAWED

The comparison logic is:
```diff
+ 				tpotMeetsThreshold = tpot <= tpotThreshold
```

Where `tpotThreshold` is the Nth percentile of historical values. This is **backwards**:

- If you set `tpot_percentile: 10`, it computes the **10th percentile** (the value below which 10% of observations fall - a very LOW value)
- Then checks if current EMA TPOT ≤ this very low threshold
- The EMA will almost always be **HIGHER** than the 10th percentile

**Result:** Setting `tpot_percentile: 10` (intended to mean "top 10% fastest") will **almost never match** because the EMA is compared against the bottom 10% of historical values, not the top 10%.

The correct interpretation should be: "route if current TPOT is in the top N% of performance" but the code implements "route if current TPOT is below the Nth percentile value".

#### c) Extreme Settings

- `tpot_percentile: 1` or `ttft_percentile: 1` → Nearly impossible to match (threshold is the minimum historical value)
- `tpot_percentile: 99` → Easy to match but defeats the purpose of "low latency" routing

---

### 4. Does the code gracefully fall back?

**Fallbacks implemented:**
1. For 1-2 observations, uses average as threshold (graceful degradation)
2. Models without data are silently skipped
3. Invalid percentiles (< 1 or > 100) return no match
4. Empty model lists return empty result
5. Rules without either percentile set are skipped with warning

**Problems with fallbacks:**
- **No routing when data is missing** - All models silently skipped, returns empty `MatchedRules`
- **No warning when all models fail** - System degrades silently to "no latency signal matched"
- **Documentation claims automatic tracking** - But TPOT is NOT being tracked

From the documentation diff:
```diff
+ - TPOT (Time Per Output Token) and TTFT (Time To First Token) are automatically tracked from responses
```

This claim is **FALSE for TPOT**.

---

## Summary of Critical Issues

| Issue | Severity | Impact |
|-------|----------|--------|
| TPOT never recorded | **CRITICAL** | TPOT-based routing will never work |
| Percentile logic inverted | **HIGH** | Low percentiles (10%) nearly impossible to match |
| Cold start problem | **MEDIUM** | New models never qualify for latency routing |
| Silent degradation | **MEDIUM** | No warnings when all models skipped |

**Recommendation:** This PR should NOT be merged in its current state. The TPOT recording is missing, and the percentile threshold logic appears conceptually inverted.

src/semantic-router/pkg/extproc/processor_res_body.go
liavweiss: Yes. Streaming: TTFT is recorded in processor_res_body.go on the first streamed body chunk (lines 62–69). Non-streaming: TTFT is recorded in processor_res_header.go when response headers arrive (lines 74–82), so we don’t record it again in the body. Both paths are covered.

Also, I checked it manually, and it works.

src/semantic-router/pkg/extproc/processor_res_header.go
liavweiss: the same https://github.com/vllm-project/semantic-router/pull/1202#discussion_r2741597391

liavweiss:
> can you make sure it works in local chat playground?
Screenshot From 2026-01-29 14-53-18

liavweiss:
Hey @rootfs, thank you for your review. I have some clarification for my code.

## 1. TPOT is wired in production

TPOT is wired in two places in `processor_res_body.go`:

- **Non-streaming** (around lines 140–149): when the full response body is parsed and `parsed.Usage` has `completion_tokens`, we compute `timePerToken` and call `classification.UpdateTPOT(ctx.RequestModel, timePerToken)`.
- **Streaming** (around lines 413–429): inside `cacheStreamingResponse()`, when the final streamed chunk contains usage with `completion_tokens`, we compute `timePerToken` and call `classification.UpdateTPOT(ctx.RequestModel, timePerToken)`.

## 2. Percentile semantics

The rule is: **current ≤ Nth percentile threshold** ⇒ treat as "in the faster N%" (higher N = more lenient). So e.g. `tpot_percentile: 70` means "route if current TPOT is in the faster 70% of history." I changed it use the median (50th percentile) of recent values as 'current', not the average, so a few slow requests don't pull the metric up and cause the rule to stop matching.

## 3. New models / cold start

New models (or after a reset) have no history, so they're skipped until they have enough data. That's by design. When a model is skipped for lack of data, I added a log at the info level.

PR #1224, opened 2026-01-30
Fix the evaluation testing, datasets api shouldn't be related to DB i…

szedan-rh:
Fix evaluation feature also support hugging face db access token.


szedan-rh:
@Xunzhuo - Could you please review?

src/vllm-sr/cli/templates/config.template.yaml
```diff
@@ -4,6 +4,15 @@
 
 version: v0.1
 
+# Environment variables passed to the container
+# These are used for authentication and API access
+environment:
```
Xunzhuo: why add this to config template?

Xunzhuo: HF_TOKEN=xxxxx vllm-sr serve should be able to pass the env?

szedan-rh: we need this as in the command vllm-sr init we will get prompt to provide 
HF_TOKEN=xxxxx and this will be copy to the config.ymal

Xunzhuo (approved):

PR #1252, opened 2026-02-05
[Docs-zh] Sync installation/configuration and overview/signal-driven-…

windsonsea:
Update zh text after running `./scripts/check-i18n-outdated.sh`


Xunzhuo (approved):
thanks

PR #1278, opened 2026-02-08
[Bugfix] Fix mmBERT embedding download path to match model registry

### LINKED ISSUE #1277: [Bugfix] download-mmbert-embedding saves to wrong path (mismatches model registry)
## Bug

`make download-mmbert-embedding` downloads the mmBERT embedding model to `models/mmbert-embed-32k-2d-matryoshka` (the raw HuggingFace repo name), but the model registry defines the canonical path as `models/mom-embedding-ultra`.

When a user configures `mmbert_model_path` in their YAML, `ResolveModelPath()` in `config/registry.go:436` matches the alias `"mmbert-embed-32k-2d-matryoshka"` and rewrites it to `models/mom-embedding-ultra` — which doesn't exist on disk because the download went to a different directory.

## To Reproduce

```bash
make download-mmbert-embedding
# Downloads to: models/mmbert-embed-32k-2d-matryoshka ✅

# Config:
# embedding_models:
#   mmbert_model_path: "models/mmbert-embed-32k-2d-matryoshka"

make run-router CONFIG_FILE=...
# ResolveModelPath rewrites to: models/mom-embedding-ultra
# ERROR: FileNotFound("models/mom-embedding-ultra/config.json")
```

## Root Cause

In `tools/make/models.mk:153`:
```makefile
@huggingface-cli download $(HF_ORG)/$(MMBERT_EMBEDDING_MODEL) --local-dir $(MODELS_DIR)/$(MMBERT_EMBEDDING_MODEL)
```

This saves to `models/mmbert-embed-32k-2d-matryoshka`. But the registry at `config/registry.go:208` defines:
```go
{
    LocalPath: "models/mom-embedding-ultra",
    RepoID:    "llm-semantic-router/mmbert-embed-32k-2d-matryoshka",
    Aliases:   []string{"mmbert-embed-32k-2d-matryoshka", ...},
}
```

`ResolveModelPath()` matches the alias and returns `models/mom-embedding-ultra` — a directory that was never created.

The built-in downloader (`make download-models` / `--download-only`) works correctly because it uses the registry's `LocalPath`. Only the manual `download-mmbert-embedding` target has this mismatch.

## Suggested Fix

Change `tools/make/models.mk` line 153 to download to the canonical `LocalPath`:

```makefile
# Before:
@huggingface-cli download $(HF_ORG)/$(MMBERT_EMBEDDING_MODEL) --local-dir $(MODELS_DIR)/$(MMBERT_EMBEDDING_MODEL)

# After:
MMBERT_EMBEDDING_LOCAL_PATH := mom-embedding-ultra
@huggingface-cli download $(HF_ORG)/$(MMBERT_EMBEDDING_MODEL) --local-dir $(MODELS_DIR)/$(MMBERT_EMBEDDING_LOCAL_PATH)
```

Or alternatively, update all `download-mmbert-*` targets to use the registry's canonical paths. The same pattern may affect other manual download targets.

## Workaround

```bash
ln -s mmbert-embed-32k-2d-matryoshka models/mom-embedding-ultra
```

## Affected Versions

Current `main` branch.

yossiovadia:
## Summary
- Fix `download-mmbert-embedding` Makefile target to save to the canonical registry path (`models/mom-embedding-ultra`) instead of the raw HF repo name (`models/mmbert-embed-32k-2d-matryoshka`)
- Prevents `ResolveModelPath()` from rewriting to a non-existent directory at runtime

## Test plan
- [ ] Run `make download-mmbert-embedding` and verify model is saved to `models/mom-embedding-ultra`
- [ ] Configure `mmbert_model_path: "models/mom-embedding-ultra"` and verify router starts without `FileNotFound` error
- [ ] Verify `make download-models` (built-in downloader) still works correctly

FIX #1277

### DIFF
--- a/tools/make/models.mk
+++ b/tools/make/models.mk
@@ -19,6 +19,8 @@ MMBERT_MODELS := \
 	mmbert-jailbreak-detector-merged
 
 # mmBERT embedding model with 2D Matryoshka support
+# Downloaded automatically by the router's built-in downloader (make download-models)
+# Registry maps this to local path models/mom-embedding-ultra (see config/registry.go)
 MMBERT_EMBEDDING_MODEL := mmbert-embed-32k-2d-matryoshka
 
 # mmBERT base 32K YaRN model (extended context MLM model)
@@ -102,7 +104,7 @@ download-mmbert-lora: ## Download mmBERT LoRA adapters for Python fine-tuning
 	@echo "✅ mmBERT LoRA adapters downloaded to $(MODELS_DIR)/"
 	@ls -la $(MODELS_DIR)/
 
-download-mmbert-all: download-mmbert download-mmbert-lora download-mmbert-32k-lora download-mmbert-32k-merged download-mmbert-embedding download-mmbert-32k download-mmbert-32k-onnx ## Download all mmBERT models, LoRA adapters, embedding, ONNX, and 32K base model
+download-mmbert-all: download-mmbert download-mmbert-lora download-mmbert-32k-lora download-mmbert-32k-merged download-mmbert-32k download-mmbert-32k-onnx ## Download all mmBERT models, LoRA adapters, ONNX, and 32K base model
 
 download-mmbert-32k-lora: ## Download mmBERT-32K LoRA adapters (32K context models)
 	@echo "📦 Downloading mmBERT-32K LoRA adapters from Hugging Face..."
@@ -146,25 +148,6 @@ download-mmbert-32k-merged: ## Download mmBERT-32K merged models (for Rust/Go in
 	@echo "  - mmbert32k-pii-detector-merged        (35-class PII NER)"
 	@echo "  - mmbert32k-jailbreak-detector-merged  (binary jailbreak)"
 	@echo "  - mmbert32k-factcheck-classifier-merged (binary fact-check)"
-
-download-mmbert-embedding: ## Download mmBERT 2D Matryoshka embedding model
-	@echo "📦 Downloading mmBERT 2D Matryoshka embedding model..."
-	@mkdir -p $(MODELS_DIR)
-	@echo ""
-	@echo "⬇️  Downloading $(MMBERT_EMBEDDING_MODEL)..."
-	@echo "   This model supports:"
-	@echo "   - 32K context length (YaRN-scaled RoPE)"
-	@echo "   - Multilingual (1800+ languages)"
-	@echo "   - 2D Matryoshka: layer early exit (3/6/11/22) + dimension reduction (64-768)"
-	@if [ -d "$(MODELS_DIR)/$(MMBERT_EMBEDDING_MODEL)" ]; then \
-		echo "   Already exists, updating..."; \
-	fi
-	@huggingface-cli download $(HF_ORG)/$(MMBERT_EMBEDDING_MODEL) --local-dir $(MODELS_DIR)/$(MMBERT_EMBEDDING_MODEL) --local-dir-use-symlinks False
-	@echo ""
-	@echo "✅ mmBERT embedding model downloaded to $(MODELS_DIR)/$(MMBERT_EMBEDDING_MODEL)"
-	@echo ""
-	@echo "Usage example:"
-	@echo "  make run-router CONFIG_FILE=config/intelligent-routing/in-tree/embedding-mmbert.yaml"
 
 download-mmbert-32k: ## Download mmBERT 32K YaRN base model (extended context MLM)
 	@echo "📦 Downloading mmBERT 32K YaRN base model..."


yossiovadia:
## Verified

Tested the fix end-to-end:

1. Removed old model dirs (`models/mmbert-embed-32k-2d-matryoshka` and `models/mom-embedding-ultra`)
2. Ran `make download-mmbert-embedding` with the fix — model downloaded to `models/mom-embedding-ultra` ✅
3. Started router with `embedding_models.mmbert_model_path: "models/mom-embedding-ultra"` — mmBERT initialized successfully ✅
4. Full vector store pipeline test (create store → upload doc → attach → ingest → search) — working ✅

```
Results: 2
  score=0.6342 content='Remote Work Policy\nEmployees may work remotely up to 3 days...'
  score=0.4282 content='PTO Policy\nAll full-time employees receive 20 days of paid t...'
PASS
```

No symlinks needed. `ResolveModelPath()` correctly resolves to `models/mom-embedding-ultra` which now exists on disk.

Xunzhuo:
i think we had the router to download the model, we can get rid of the makefile already?

Xunzhuo:
src/semantic-router/pkg/config/registry.go

yossiovadia:
Good point — the router's built-in downloader (`make download-models` / `--download-only`) uses the registry and always downloads to the correct `LocalPath`. The manual `make download-mmbert-embedding` target bypasses the registry, which is what caused this mismatch.

Two options:

1. **Keep the fix as-is** — align the manual target with the registry (this PR). Useful for downloading just the embedding model without building the router binary first.

2. **Remove the manual target entirely** — rely on `make download-models` for everything. Simpler, no more sync issues.

I'm fine with either. If you prefer option 2, I can update the PR to remove the `download-mmbert-embedding` target instead.

Xunzhuo:
thanks, let us go for 2 : )

yossiovadia:
Updated — removed `download-mmbert-embedding` target entirely instead of fixing its path.

The router's built-in downloader (`--download-only`) handles it correctly:

```
$ ./bin/router -config config/config-rag-example.yaml --download-only

✗ models/mom-embedding-ultra (need download)
Downloading model: models/mom-embedding-ultra
Fetching 24 files: 100%|██████████| 24/24 [00:02<00:00, 10.01it/s]
Successfully downloaded model: models/mom-embedding-ultra
All required models are ready
```

Users just need `embedding_models.mmbert_model_path` in their config, and the router downloads to the correct registry path automatically.

tools/make/models.mk
```diff
@@ -147,25 +149,6 @@ download-mmbert-32k-merged: ## Download mmBERT-32K merged models (for Rust/Go in
 	@echo "  - mmbert32k-jailbreak-detector-merged  (binary jailbreak)"
 	@echo "  - mmbert32k-factcheck-classifier-merged (binary fact-check)"
 
-download-mmbert-embedding: ## Download mmBERT 2D Matryoshka embedding model
```
rootfs, line 150: why skip this one?

tools/make/models.mk
```diff
@@ -102,7 +104,7 @@ download-mmbert-lora: ## Download mmBERT LoRA adapters for Python fine-tuning
 	@echo "✅ mmBERT LoRA adapters downloaded to $(MODELS_DIR)/"
 	@ls -la $(MODELS_DIR)/
 
-download-mmbert-all: download-mmbert download-mmbert-lora download-mmbert-32k-lora download-mmbert-32k-merged download-mmbert-embedding download-mmbert-32k download-mmbert-32k-onnx ## Download all mmBERT models, LoRA adapters, embedding, ONNX, and 32K base model
+download-mmbert-all: download-mmbert download-mmbert-lora download-mmbert-32k-lora download-mmbert-32k-merged download-mmbert-32k download-mmbert-32k-onnx ## Download all mmBERT models, LoRA adapters, ONNX, and 32K base model
```
rootfs, line 107: this one doesn't download embedding models

yossiovadia:
Hey @rootfs — good catches, both related.

**Re: "why skip this one?" (line 150):** @Xunzhuo suggested removing the manual `download-mmbert-embedding` target entirely since the router's built-in downloader (`--download-only`) handles embedding models correctly via the registry — no path sync issues. See the [earlier discussion](https://github.com/vllm-project/semantic-router/pull/1278#issuecomment-3866073624).

**Re: "this one doesn't download embedding models" (line 107):** You're right — removing `download-mmbert-embedding` from the `download-mmbert-all` deps means `make download-mmbert-all` no longer covers embeddings. That's a gap worth fixing.

A few ways to address it:

1. **Re-add a fixed `download-mmbert-embedding` target** with the correct path (`mom-embedding-ultra`) and restore it in `download-mmbert-all` deps — straightforward but reintroduces a manual target that can drift from the registry again.
2. **Add a note/echo in `download-mmbert-all`** pointing users to `./bin/router --download-only` for embedding models — lightweight, no drift risk.
3. **Have `download-mmbert-all` invoke the router binary** for embedding downloads — most robust, but adds a build dependency on the router binary for a Makefile target.

I'd recommend **option 2** — it keeps `download-mmbert-all` honest about what it covers without reintroducing the manual path sync problem that caused this bug in the first place. Does that work for you?

yossiovadia:
Tbh, not sure of option 2 is my suggestion after all

rootfs:
@yossiovadia sure, if the models are already downloaded in another target, please feel free to clean up the redundant ones.

Issue #1146, opened 2026-01-21
[Security] Restrict Router Replay read APIs with explicit authorization


R3hankhan123:
## Audited upstream baseline (2026-07-15)

On current `main`, Router Replay exposes read-only routes under `/v1/router_replay` for list/detail, aggregate, and trajectory data when replay recorders are configured. `handleRequestHeaders` dispatches these routes directly to `handleRouterReplayAPI`; there is no route-specific authentication or authorization check in that code path.

Replay records can contain request/response bodies and tool traces when those fields are configured for capture. This makes the API an operator surface, not a public inference endpoint. A deployment can place gateway authentication and network policy in front of it, but the repository does not yet define and test one explicit production access contract.

## Scope

1. Define the trusted identity source and required permission for Router Replay reads.
2. Enforce fail-closed authorization for every Replay route, including list/detail, aggregate, and trajectory.
3. Keep authorization separate from record filtering: filtering results is not an access-control mechanism.
4. Document a supported deployment boundary for network isolation and gateway policy. NetworkPolicy belongs in deployment manifests/docs, not in the Go handler.
5. Ensure unauthorized attempts are observable without logging credentials or replay content.

## Acceptance criteria

- [ ] Unauthenticated requests are rejected before any replay record is read.
- [ ] Authenticated but unauthorized callers receive a stable denial response.
- [ ] Authorized operators can use every documented Replay read route.
- [ ] Route aliases, query parameters, and detail modes cannot bypass the check.
- [ ] Tests cover enabled/disabled Replay, missing identity, denied identity, and allowed identity.
- [ ] Production deployment docs include least-privilege gateway/network-policy guidance.
- [ ] Logs and metrics contain no raw credentials, prompts, responses, or tool arguments.

## Relevant surfaces

`src/semantic-router/pkg/extproc/router_replay_api.go`, `processor_req_header.go`, trusted auth/identity configuration, deployment manifests, docs, and E2E coverage.

Related history: #1075, #1107.

R3hankhan123:
@rootfs @Xunzhuo is this what is expected for restricting the endpoint, only users with auth token (Maybe API key or jwt token) can access the endpoint for router replay logs?

R3hankhan123:
maybe integration with authorino(https://github.com/Kuadrant/authorino) or implementation in the repo is better? wdyt @rootfs @Xunzhuo

rootfs:
[...]

PR #1301, opened 2026-02-09
feat(memory): use authz framework header for user_id extraction

### LINKED ISSUE #1229: Memory: Inject user_id from auth layer instead of accepting from user request
### Problem

Currently, the Memory API accepts `user_id` directly from the user request payload. This creates a privacy/security concern where a malicious user could potentially:
- Access another user's memories by spoofing their user_id
- Store memories under a different user's identity

### Current Behavior

The `user_id` is passed directly in the request body and trusted without verification against the authenticated user identity.

### Expected Behavior

The `user_id` should be extracted from the authenticated user context (e.g., JWT token, auth header) rather than trusting the client-provided value.

### Related

- PR #1194 (Agentic Memory POC)

yehuditkerido:
Use the standard x-authz-user-id header from the authz framework
(PR #1315) instead of custom per-memory config fields. This prevents
user_id spoofing by trusting only headers injected by the external
auth layer (Authorino, Envoy Gateway JWT, oauth2-proxy, etc.).

Changes:
- Add extractUserID() with priority: auth header > metadata fallback
- Update extractMemoryInfo() and getUserIDFromContext() to use it
- Fallback to metadata["user_id"] for development/testing without
  an auth layer

Tests:
- AuthHeaderTakesPrecedence: auth header wins over metadata
- AuthHeaderOnly: works without metadata present
- FallbackToMetadataWhenNoAuthHeader: dev/testing flexibility
- EmptyAuthHeaderFallsBackToMetadata: empty header handled
- NoAuthHeaderNoMetadata: returns empty string
- UnrelatedHeaderIgnored: only x-authz-user-id is recognized

Resolves #1229

### DIFF
--- a/src/semantic-router/pkg/extproc/memory_helpers.go
+++ b/src/semantic-router/pkg/extproc/memory_helpers.go
@@ -30,14 +30,18 @@ func extractAutoStore(ctx *RequestContext) bool {
 	return false
 }
 
+// extractUserID is defined in user_id_dev.go (with metadata fallback) and
+// user_id_prod.go (auth header only) using build tags.
+
 // extractMemoryInfo extracts sessionID, userID, and history from the request context.
 //
 // Returns an error if userID is not available, because memory would be orphaned
 // (unretrievable) without a valid userID. Memory retrieval filters by userID first,
 // so memories stored without userID cannot be retrieved later.
 //
-// userID is required and must be provided via:
-//   - metadata["user_id"] in Response API request (OpenAI API spec-compliant)
+// userID extraction priority:
+//  1. Auth header (x-authz-user-id) injected by external auth service (Authorino, Envoy Gateway JWT, etc.)
+//  2. metadata["user_id"] in Response API request (fallback for development/testing)
 func extractMemoryInfo(ctx *RequestContext) (sessionID string, userID string, history []memory.Message, err error) {
 	// First check if this is a Response API request
 	// Non-Response API requests cannot track turns without ConversationID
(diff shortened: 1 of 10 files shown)


rootfs:
@yehudit1987 let's wait a bit on this PR. We'll experiment with ext_authz for user and token management at the moment, the auth header and provider token will be auto injected in a complete flow.

rootfs:
cc @Xunzhuo @srampal

rootfs:
related: #1146

rootfs:
@yehudit1987 please check #1315. We'll use the authz for user headers.

src/semantic-router/pkg/extproc/memory_helpers.go
```diff
@@ -30,30 +31,51 @@ func extractAutoStore(ctx *RequestContext) bool {
...
+// This is untrusted (client-provided) and intended for development/testing only.
+func extractUserID(ctx *RequestContext) string {
+	// Check auth header first (trusted source, injected by auth backend)
+	if userID, ok := ctx.Headers[headers.AuthzUserID]; ok && userID != "" {
+		logging.Debugf("Memory: Using user_id from auth header (%s)", headers.AuthzUserID)
+		return userID
+	}
+
+	// Fallback to metadata["user_id"] (untrusted, for development/testing)
+	if ctx.ResponseAPICtx != nil && ctx.ResponseAPICtx.OriginalRequest != nil {
```
rootfs: this could bypass the the authz gate, can you use a go build tag for dev build to enable this?

yehuditkerido: Good point. Moved the metadata fallback behind a `dev` build tag:

- `user_id_prod.go` (default): only reads `x-authz-user-id`, no fallback
- `user_id_dev.go` (`//go:build dev`): adds `metadata["user_id"]` fallback

Issue #1329, opened 2026-02-13
Question: Start minimal router with vllm-sr serve


pbouda:
I want to start a minimal semantic router setup i.e. I don't need observability or a dashboard. Right now when I run `vllm-sr serve` it will start several services. Is there a way to start a minimal router setup, that would handle requests and forward them according to my config, with all non-essential services removed?

rootfs:
cc @Xunzhuo @szedan-rh

szedan-rh:
Thanks for reporting that, will look into it

szedan-rh:
At now, we don't have this option as far as i remember, but we can develop the feature to support that, will work with semantic router team to define what is the minimal router setup

pbouda:
I found a way to disable the observabilty at least, I changed the line here: https://github.com/vllm-project/semantic-router/blob/main/src/vllm-sr/cli/core.py#L33

The rest seems to be in the docker image, I will see what I can still find out about that.

szedan-rh:
Done

PR #1355, opened 2026-02-18
feat: add --minimal flag to vllm-sr serve (#1329)

szedan-rh:
Allow starting the semantic router without dashboard or observability
  services (Jaeger, Prometheus, Grafana) by passing --minimal to the
  serve command. This reduces resource usage for deployments that only
  need core routing functionality.

  - Add --minimal CLI flag to serve command in main.py
  - Skip dashboard port mapping (8700) when minimal mode is active
  - Conditionalize startup summary to hide dashboard/observability URLs
  - Add DISABLE_DASHBOARD env var check in start-dashboard.sh
  - Warn when --readonly is used alongside --minimal
  - Add unit tests for minimal mode behavior


rootfs:
I think for the long term as we add more components, it is better to have a config file for specific enablement.

szedan-rh:
@rootfs - will open a follow up ticket about that.

Issue #1293, opened 2026-02-09
[Tracking] Agentic Memory: Future Improvements


yehudit1987:
## Agentic Memory: Future Improvements

Tracking issue for memory feature enhancements following #1194.

This issue tracks all planned improvements to the Agentic Memory system. Check off items as they are completed via PRs.

---

### Documentation
- [ ] #1286 - Move POC document to website proposals

### API & Integration
- [ ] #1287 - Expand memory support to Chat Completions API
- [ ] #1288 - Memory Management REST API

### Storage & Performance
- [ ] #1289 - Redis hot cache layer for frequently accessed memories
- [ ] #1290 - Milvus partition key for user isolation

### Memory Operations
- [ ] #1291 - Self-correcting memory with quality scoring (Research needed)

### Extraction
- [ ] #1292 - Session end detection for episodic memory extraction (Research needed)

### Observability
- [ ] Memory retrieval metrics - Add Prometheus metrics *(issue TBD)*

### Testing
- [ ] E2E tests for per-decision memory plugin *(issue TBD)*

### Context Compression
- [ ] Replace full history with compressed context *(issue TBD, depends on #1289)*

---

## P2 - Medium Priority *(issues to be created)*

- [ ] CRD validation for memory plugin
- [ ] Per-user memory quotas
- [ ] Memory TTL & expiration
- [ ] Consolidate operation (merge related memories)
- [ ] Contradiction detection
- [ ] Memory type routing
- [ ] Context drift detection
- [ ] HyDE query rewriting
- [ ] Context manager for session state
- [ ] Adversarial defense
- [ ] Concurrency handling
- [ ] Memory quality dashboard
- [ ] Load testing at scale
- [ ] Update POC to production guide

---

## P3 - Future Consideration *(Optional)*

- [ ] Reflect operation (analyze memory patterns)
- [ ] Retrieval feedback loop
- [ ] Graph store integration (Neo4j)
- [ ] Time-series index for temporal queries

---

## Cleanup (After Feature Complete)

- [ ] Remove demo logging - Remove `TODO: Remove demo logs after POC` comments

---

## Dependencies

```
#1289 (Redis) → Context Manager → Context Compression
Memory Metrics → Memory Dashboard
```

---

## Related

- Original implementation: #1194
- Original issue: #808

aaronprivate95-glitch:
Nice tracking list. A few “ship-it” additions that have helped teams avoid memory becoming an unbounded blob:

[...]

PR #1363, opened 2026-02-19
feat(memory): add Chat Completions API memory support

### LINKED ISSUE #1287: [Memory] Expand memory support to Chat Completions API
## Summary

Add memory support to the Chat Completions API (`/v1/chat/completions`), which currently only works with the Response API.

## Current State

Memory features only work with Response API (`/v1/responses`).

## Task

Enable memory retrieval and injection for Chat Completions API requests.

### Challenges

- Chat Completions is stateless by design (no `previous_response_id`)
- Need to pass `user_id` via headers or request body extension
- `auto_store` requires conversation history which Chat Completions doesn't track

### Proposed Approach

1. **Memory Retrieval**: Support for Chat Completions (inject memories into context)
   - User ID passed via `metadata["user_id"]` or `x-user-id` header
   - Retrieved memories injected as system message
   
2. **Memory Extraction** (`auto_store`): 
   - Requires client to explicitly send conversation history
   - Or extract only from the current turn (limited context)

### Implementation Notes

- Modify `processor_req_body.go` to handle Chat Completions requests
- Add user ID extraction from headers for non-Response API requests
- Ensure memory plugin configuration applies to Chat Completions decisions

## Context

- **Priority**: P1 (High)
- **Effort**: Medium
- **Research Required**: No
- **Dependencies**: None

## Related

- Original implementation: #1194
- Original issue: #808

yehuditkerido:
## Summary
Extend agentic memory to OpenAI Chat Completions API (`/v1/chat/completions`), enabling cross-session memory for stateless conversations. Previously only available via custom Response API.

## Changes
- **Chat Completions support**: Store messages in `RequestContext`, extract user/assistant content via API-agnostic helpers
- **Opt-out mechanism** (addresses [rootfs review #1293](https://github.com/vllm-project/semantic-router/issues/1293#issuecomment-2579448923)):
  - Config: `memory.disabled_routes`, `memory.disabled_models` 
  - Header: `x-disable-router-memory: true` (optional)
  - Constants: `headers.DisableRouterMemory`, `headers.PseudoHeaderPath`
- **Testing**: `ChatCompletionsMemoryTest` E2E test for `/v1/chat/completions`, unit tests for opt-out

Resolves #1287


rootfs:
@yehudit1987 The chat completion API is stateless. The end user is responsible to add chat history in the request. This could conflict with context memory injection.

yehuditkerido:
> @yehudit1987 The chat completion API is stateless. The end user is responsible to add chat history in the request. This could conflict with context memory injection.

Hi @rootfs, 

Valid concern - there's potential for conflict similar to stale RAG data. However, this provides value for **cross-session persistent memory** that stateless clients can't maintain themselves.

Memory is injected into the system prompt without modifying the user's message array. The conflict risk is mitigated by temporal scoring and relevance filtering, and is most useful for stable facts rather than conversation-specific context.

Do you have specific concerns about this approach or suggestions for improvement?

rootfs:
@yehudit1987 i think we need to turn on memory only when users explicitly request it e.g. having a `memory` tool in tool calling. However, the UX needs to be researched first. also cc @Xunzhuo

Issue #1348, opened 2026-02-18
feature: few short contrastive learing with preference signal


ppppqp:
### Describe the feature

As discussed in community sync Feb 17 with @rootfs, explore the feasibility of few short contrastive learing for preference signal for better serving performance and reduce the dependency on GPU

PR #1332, opened 2026-02-14
White Paper for vLLM Semantic Router

Xunzhuo:
Add the vLLM Semantic Router paper LaTeX source files.

- 16 section files covering architecture, signals, decisions, safety, HaluGate, LoRA/MoM, deployment, evaluation, etc.
- References bibliography (`references.bib`)
- Makefile for building (`pdflatex + bibtex`)
- Figures placeholder directory

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using `-s` when doing `git commit`
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

PR #1368, opened 2026-02-21
Feat: Contrastive preference classifier

ppppqp:
FIX https://github.com/vllm-project/semantic-router/issues/1348

## Motivation
One major issue that we face for preference classifier is the long inference time (multiple seconds running on CPU, so it's heavily reliant on GPU). This PR explores the possiblity of using embedding only method to reduce inference time from multiple seconds to sub seconds, while still retain reasonable precision and recall.

## Performance Benchmark
Go test benchmark and the e2e testing shows the complete inference process takes about 0.2-0.3 second.

## Precision Benchmark
I tested with two models, with the same benchmark dataset I created for preference SFT model (using ShareGPT dataset and LLM-labeling)
The setup: 
- 100 samples
- Each sample has one user query, one truth label, four negative labels
- Run Qwen3 & MMBert over a spectrum of threshold values
- To test out-of-scope case, also run the same setup without the truth label, and see if all preference labels falls under confidence value.

Result shows that the optimal confidence threshold is around 0.5~0.6 where both models achieve about 78% accuracy overall.

image

image

Demo:

https://github.com/user-attachments/assets/def0d1a7-61f2-4188-883e-0a534802ff07

## Design Decisions
1. Independent confidence threshold per preference rule. The generosity of each preference label user defined can be different. For example, user can have a very general preference signal (probably with few or no examples, short description) that are a lot easier to hit than a very specific preference signal. Having independent confidence threshold for the two signals enables the user to control with signal collection with more guanualarity. The threshold is made to be optional, because for external models we don't really need per-preference confidence score.
2. Adding config validation for enabling embedding-based preference classifier. We need the embedding model to be loaded to `ModelFactory` at the beginning, so if the user configured the embedding-based preference classifier but did not include embedding models, we will throw an error. This is not likely the case because by default we enable `mmbert`.

image

---

- [ ] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [ ] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [ ] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to semantic-router! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[CLI]</code> for changes to the command-line interface tools.</li>
    <li><code>[Dashboard]</code> for changes to the dashboard or web UI.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>DCO and Signed-off-by</h3>
<p>When contributing changes to this project, you must agree to the <a href="https://github.com/vllm-project/vllm/blob/main/DCO">DCO</a>. Commits must include a <code>Signed-off-by:</code> header which certifies agreement with the terms of the DCO.</p>
<p>Using <code>-s</code> with <code>git commit</code> will automatically add this header.</p>

<h3>What to Expect for the Reviews</h3>


rootfs:
@ppppqp do you want to mention this in the white paper in #1332? we can add to future work if this PR is not merged

src/semantic-router/pkg/classification/classifier.go
```diff
@@ -2481,6 +2481,10 @@ func (c *Classifier) IsPreferenceClassifierEnabled() bool {
 		return false
 	}
 
+	if c.Config.Classifier.PreferenceModel.UseContrastive {
```
rootfs, line 2907: maybe we can just completely switch over to contrastive and remove the qwen prompt?

src/semantic-router/pkg/classification/contrastive_preference_classifier.go
```diff
@@ -0,0 +1,220 @@
...
+
+// NewContrastivePreferenceClassifier builds a contrastive preference classifier.
+// modelType follows GetEmbeddingWithModelType (e.g. "qwen3", "gemma", "mmbert").
+func NewContrastivePreferenceClassifier(rules []config.PreferenceRule, modelType string) (*ContrastivePreferenceClassifier, error) {
+	if len(rules) == 0 {
+		return nil, fmt.Errorf("contrastive preference rules cannot be empty")
+	}
+
+	if modelType == "" {
+		modelType = "qwen3"
```
rootfs: mmbert

src/semantic-router/pkg/classification/contrastive_preference_classifier.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+			tasks = append(tasks, task{ruleName: rule.Name, text: example})
+		}
+	}
+
+	if len(tasks) == 0 {
+		return fmt.Errorf("no examples provided for contrastive preference classifier")
+	}
+
+	numWorkers := runtime.NumCPU() * 2
```
rootfs, line 83: what is the reason for so many cores?

src/semantic-router/pkg/classification/contrastive_preference_classifier.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, fmt.Errorf("preference similarity %.3f below threshold %.3f", bestScore, threshold)
+	}
+	return &PreferenceResult{
+		Preference: bestRule,
+		Confidence: bestScore,
+	}, nil
+}
+
+func (c *ContrastivePreferenceClassifier) collectExamples(rule config.PreferenceRule) []string {
+	examples := make([]string, 0, 1+len(rule.Examples))
```
rootfs, line 209: fail if no examples

rootfs:
@ppppqp can you fix the CI? If you also have any benchmark results, can you update the white paper too?

ppppqp:
Yes, will work on this over the weekend!

src/semantic-router/pkg/classification/contrastive_preference_classifier.go
ppppqp: I put the error checking here:

https://github.com/ppppqp/semantic-router/blob/ec0b649e738c51d17c3d12a7c8bdd354fd1960f7/src/semantic-router/pkg/classification/contrastive_preference_classifier.go#L79-L82

src/semantic-router/pkg/classification/contrastive_preference_classifier.go
ppppqp: I followed the implementation here:

https://github.com/vllm-project/semantic-router/blob/ec0b649e738c51d17c3d12a7c8bdd354fd1960f7/src/semantic-router/pkg/classification/embedding_classifier.go#L124-L128

src/semantic-router/pkg/classification/classifier.go
ppppqp: Do you mean removing the local qwen classifier or the external one?
I suppose we can remove the local one because the GPU dependency?

ppppqp:
Trying to fix the memory integration test but failed. The test works on my local:
```
VLLM_SR_IMAGE=semantic-router:dev make vllm-sr-dev
VLLM_SR_IMAGE=semantic-router:dev vllm-sr serve --config config/testing/config.memory-user.yaml --image-pull-policy=never
```
runs fine.

However, i did need to comment out https://github.com/vllm-project/semantic-router/blob/1b6ab6f13a78b9acb8f0f022d943895998258d49/tools/make/docker.mk#L265-L268
when running build but i believe this is due to MacOS compatibility?
Do you have any insights on why the test fails?

ppppqp:
Regression tested

https://github.com/user-attachments/assets/91e66bc1-8e79-423d-994e-51ad5c37dfaf

PR #1382, opened 2026-02-25
Support memory load in vllm dashboard

szedan-rh:
Support memory load in VLLM, dashboard, continue to memory feature that was merged, added the option to load the memory for the user according to the user id.
image


szedan-rh:
@rootfs / @Xunzhuo / @yehudit1987 - Could you please review?

szedan-rh:
@rootfs - will raise PR to do  improvement in the memory area, after that we will revisit this feature again.

PR #1391, opened 2026-02-26
[Bugfix] Fix stale paths in go.mod, Makefile, and scripts after training dir refactor

Djanghao:
## Description

The training directory reorganization in #1379 moved subdirectories into a categorical structure but left several references pointing to the old paths, causing build failures and broken make targets.

### Changes

- **`src/training/model_selection/ml_model_selection/go.mod`**: Fix `replace` directives — the module moved one level deeper but relative paths were not updated, causing `go build` to fail.
- **`tools/make/models.mk`**: Update `LORA_DIR` from `training_lora` to `model_classifier`, fix feedback trainer path, and update Python import paths in merge targets.
- **`scripts/train-mmbert32k-gpu.sh`**: Update all five training script paths to match new directory structure.

FIX #1379


Xunzhuo (approved):
thanks!

PR #1392, opened 2026-02-26
Reorgnize the top bar to group analyses features and Operations into …

szedan-rh:
Reorgnize the top bar to group analyses features and Operations into drop down menu


szedan-rh:
@Xunzhuo - Could you please review the change?

PR #1393, opened 2026-02-26
[Feat]: add response-level jailbreak detection filter

abdallahsamabd:
## Summary

Adds a `response_jailbreak` plugin that runs the jailbreak classifier on the
LLM **response body**, complementing the existing input-level jailbreak detection.
This follows the same pattern as the existing hallucination response filter.

Motivated by research on memory-targeted attacks
([arXiv:2503.03704](https://arxiv.org/abs/2503.03704)), where adversarial
content in LLM responses can go undetected by input-only scanning.

### What's included

- **`res_filter_jailbreak.go`** — response filter with configurable actions
  (`header`, `block`, `none`), mirroring `res_filter_hallucination.go`
- **`ResponseJailbreakPluginConfig`** — per-decision plugin configuration
- **`x-vsr-response-jailbreak-*` headers** — detection metadata passed to clients
- **`ResponseJailbreakDetected/Type/Confidence`** fields on `RequestContext`
  and router replay records
- Unit tests and documentation

### What's NOT included (separate PR)

Memory protection (gating memory storage with the response jailbreak flag,
removing `ContentValidator`) will follow in a dedicated PR.

## Example configuration

decisions:
  - name: general_chat
    plugins:
      - type: response_jailbreak
        configuration:
          enabled: true
          threshold: 0.5
          action: header   # "header" | "block" | "none"

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_res_body.go
+++ b/src/semantic-router/pkg/extproc/processor_res_body.go
@@ -226,35 +226,6 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 		}
 	}
 
-	// Memory chunk storage (async, if auto_store enabled)
-	// Stores current turn directly in vector store -- no LLM extraction overhead.
-	autoStoreEnabled := extractAutoStore(ctx)
-	if !autoStoreEnabled && r.Config != nil && r.Config.Memory.AutoStore {
-		logging.Infof("extractAutoStore: Falling back to global config, AutoStore=%v", r.Config.Memory.AutoStore)
-		autoStoreEnabled = true
-	}
-	logging.Infof("Memory store check: MemoryExtractor=%v, autoStore=%v, jailbreakPassed=%v",
-		r.MemoryExtractor != nil, autoStoreEnabled, !ctx.JailbreakDetected)
-	if r.MemoryExtractor != nil && autoStoreEnabled && !ctx.JailbreakDetected {
-		currentUserMessage := extractCurrentUserMessage(ctx)
-		currentAssistantResponse := extractAssistantResponseText(responseBody)
-		go func() {
-			bgCtx := context.Background()
-			sessionID, userID, history, err := extractMemoryInfo(ctx)
-			if err != nil {
-				logging.Errorf("Memory store failed: %v", err)
-				return
-			}
-
-			logging.Infof("Memory store: sessionID=%s, userID=%s, userMsg=%d chars, assistantMsg=%d chars, history=%d msgs",
-				sessionID, userID, len(currentUserMessage), len(currentAssistantResponse), len(history))
-
-			if err := r.MemoryExtractor.ProcessResponseWithHistory(bgCtx, sessionID, userID, currentUserMessage, currentAssistantResponse, history); err != nil {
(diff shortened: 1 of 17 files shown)


abdallahsamabd:
@rootfs 
please review design doc, raised some question there 
btw regex approach will cover some specific cases, maybe we should train a new model to detect memory injection prompts ?
[minja-defense-design.html](https://github.com/user-attachments/files/25573730/minja-defense-design.html)
thanks

rootfs:
@abdallahsamabd i read the minja paper, there are some disconnect with this PR. This PR adds user id level memory injection rate limiting and filtering. This is not quite the way minja paper describes. The paper doesn't have the user id filtering, it is based on the contents that are from different sources, each of the source may from different users. And the paper identified that the system level filtering doesn't help (Section 5.4).

On the other hand, the memory injection used by semantic router is at the per user level. We don't allow shared memory at the moment for privacy concerns, however it may change in the future though if user cases can be identified. If we are going to implement shared memory, let's revisit this PR by then.

src/semantic-router/pkg/memory/content_validator.go
```diff
@@ -0,0 +1,81 @@
...
+
+	start := time.Now()
+	content := formatTurnChunk(userMessage, assistantResponse)
+	defer func() {
+		duration := time.Since(start).Seconds()
+		RecordContentValidationLatency(duration)
+	}()
+
+	if v.classifyFunc != nil {
+		isJailbreak, label, confidence, err := v.classifyFunc(content)
```
rootfs: The current flow is to validate the response before adding the response to the memory. Adding a chunk level validation here is redundant.

abdallahsamabd: @rootfs 
The validation runs once on the full response (userMessage + assistantResponse) before any chunking happens (extractor.go line 152). If blocked, it returns early and no chunks are created. There is no per-chunk validation — ValidateForStorage is called exactly once per response, not per chunk

rootfs:
@abdallahsamabd this is cool! can you break down into 2 PRs, one for response jailbreak detection and one for memory protection. At the moment, the memory protection is not exactly minja based any more, so you can just use a more concise description.

src/semantic-router/pkg/extproc/processor_res_body.go
```diff
@@ -233,9 +233,9 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 		logging.Infof("extractAutoStore: Falling back to global config, AutoStore=%v", r.Config.Memory.AutoStore)
 		autoStoreEnabled = true
 	}
-	logging.Infof("Memory store check: MemoryExtractor=%v, autoStore=%v, jailbreakPassed=%v",
-		r.MemoryExtractor != nil, autoStoreEnabled, !ctx.JailbreakDetected)
-	if r.MemoryExtractor != nil && autoStoreEnabled && !ctx.JailbreakDetected {
+	logging.Infof("Memory store check: MemoryExtractor=%v, autoStore=%v",
```
rootfs: memory injection should happen after `performResponseJailbreakDetection` and `performHallucinationDetection`

abdallahsamabd: good catch
thanks @rootfs

PR #1438, opened 2026-03-05
[Feat]: implement all PIIOptions in DetectPII and add RevealEntityText option

toffentoffen:
Wire up previously unused PIIOptions fields (EntityTypes, ConfidenceThreshold, ReturnPositions, MaskEntities) and add new RevealEntityText option that returns the original detected text instead of "[DETECTED]". Extract buildPIIResponse helper for testability and add comprehensive table-driven unit tests covering all options individually and in combination.

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

### DIFF
--- a/src/semantic-router/pkg/services/classification.go
+++ b/src/semantic-router/pkg/services/classification.go
@@ -3,6 +3,7 @@ package services
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -385,6 +386,7 @@ type PIIOptions struct {
 	ConfidenceThreshold float64  `json:"confidence_threshold,omitempty"`
 	ReturnPositions     bool     `json:"return_positions,omitempty"`
 	MaskEntities        bool     `json:"mask_entities,omitempty"`
+	RevealEntityText    bool     `json:"reveal_entity_text,omitempty"`
 }
 
 // PIIResponse represents the response from PII detection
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Thanks @toffentoffen can you share a screenshot like how it is looking now?

toffentoffen:
Would you mind specifying about what should I share a screenshot? Thanks.
BTW, need to debug somehow the inference result because the start and stop position of the found entity, are not correct I think. They return start=x end=x+1 even though the found entity is a long social security numver for instance.

toffentoffen:
All the options, except RevelEntityTest, where there an accepted by the endpoint but where never used.
This is a request example:
```
curl -i -X POST http://localhost:8080/api/v1/classify/pii \
  -d '{
    "text": "My credit card number is 1234-5678-9012-3456",
    "options": {
      "confidence_threshold": 0.1,
      "return_positions": true,
      "mask_entities": true,
      "reveal_entity_text": true
    }
  }'
  ```
  This PR uses all the options to appropriately generate the response. Notice that `masked_text` doesn't take into account the full credit card number.
  ```
  {
  "has_pii": true,
  "entities": [
    {
      "type": "CREDIT_CARD",
      "value": "1",
      "confidence": 0.5897449254989624,
      "start_position": 25,
      "end_position": 26,
      "masked_value": "[CREDIT_CARD_0]"
    }
  ],
  "masked_text": "My credit card number is [CREDIT_CARD_0]234-5678-9012-3456",
  "security_recommendation": "block",
  "processing_time_ms": 136
}
```
The wrong masking is probably because of a bug in the piiclassifier that returns wrongly the `start` and `end` values:

https://github.com/vllm-project/semantic-router/blob/5bdc212fd3789cf70e26f812709863c481e99e3a/candle-binding/semantic-router.go#L3465-L3473

Need to check if I can figure out how this candle binding works and fix it.

rootfs:
@toffentoffen thanks! when you figure out the wrong mask, please follow up with another PR.

PR #1446, opened 2026-03-06
security: strip untrusted identity headers when no auth backend configured

### LINKED ISSUE #1445: [Security] Prevent identity-header spoofing without ext_authz
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

The Router can consume configured identity headers such as `x-authz-user-id` and `x-authz-user-groups` from the request context. If a deployment enables header-based identity without a verified external authorization and header-injection boundary, a client can supply those headers directly and impersonate another principal.

### To Reproduce

1. Configure identity-based routing or authorization using request headers.
2. Deploy the Router without an external authorization filter that removes client-supplied identity headers and injects verified values.
3. Send a request with forged identity headers.
4. Observe that the configured identity-dependent behavior can consume the caller-provided values.

### Expected behavior

Identity headers are trusted only when the deployment explicitly declares a verified external authorization/header-injection boundary. Otherwise the Router must strip, ignore, or fail closed on those headers. Maintained Envoy configurations must remove client-supplied identity headers before an authorization filter injects verified values.

### Impact

A deployment that enables header-derived identity without the required trust boundary can make authorization or routing decisions using a spoofed principal.

### Additional context

Keep this issue scoped to identity-header provenance and fail-closed deployment behavior. Static tenant isolation, per-user quota/rate-limit products, and memory multi-tenancy are not part of this issue or the current Enterprise direction.

Parent Epic: #3166

yossiovadia:
## Summary

Fixes #1445 — without an auth backend, any client can spoof `x-authz-user-id` and `x-authz-user-groups` headers to impersonate any user, bypassing role-based routing, per-user rate limits, and memory isolation.

## Fix

Add `authz.trust_identity_headers` config flag:

- **`true` (default)**: trust identity headers — assumes an auth backend (ext_authz, Envoy Gateway JWT, oauth2-proxy, etc.) injects them before the request reaches ext_proc. This preserves backward compatibility.
- **`false`**: strip identity headers from every request, preventing client spoofing.

Deployments without an auth backend should add:
```yaml
authz:
  trust_identity_headers: false
```

## Why not auto-detect?

Identity headers can be injected by many mechanisms (ext_authz/Authorino, Envoy Gateway JWT claim_to_headers, oauth2-proxy, Istio RequestAuthentication). The router has no reliable way to distinguish "injected by auth backend" from "set by client" — by the time ext_proc sees the request, both look identical. An explicit config flag gives operators control without breaking existing deployments.

## Changes

| File | Change |
|------|--------|
| `pkg/config/config.go` | Add `TrustIdentityHeaders` field + `ShouldTrustIdentityHeaders()` method |
| `pkg/extproc/processor_req_header.go` | Strip identity headers when trust is disabled |
| `e2e/testing/10-security-audit-test.py` | Security e2e tests (12 tests) |

## Test plan

- [x] `test_spoofed_identity_headers_are_stripped` — auth header test
- [x] All 12 e2e security tests pass
- [x] Default `true` preserves backward compatibility (existing authz-rbac integration test should pass)


yossiovadia:
Note: The authz-rbac integration test sends raw identity headers from the test client without JWT validation. This test passes with the default `trust_identity_headers: true`, but it means the test doesn't validate the full auth chain. Filed #1447 to track improving the test to use actual JWT validation.

PR #1449, opened 2026-03-06
security: add user isolation to semantic cache

### LINKED ISSUE #1448: security: semantic cache has no user isolation — cross-user data leak
## Summary

## Root Cause

## Impact

Attack scenario:

## Proposed Fix

yossiovadia:
## Summary

Fixes #1448 — the semantic cache was keyed by `(model, query_embedding)` with no user identifier. User A's cached response could be served to User B for semantically similar queries.

## Root Cause

`CacheEntry` had no `UserID` field. Cache lookups in `FindSimilarWithThreshold` searched across all entries regardless of who created them. All 4 backends (in-memory, Redis, Milvus, hybrid) were affected.

## Fix

Add `UserID` to the cache entry and filter lookups by user:

- **`cache_interface.go`**: Add `UserID string` to `CacheEntry`, add `userID string` parameter to `AddPendingRequest`, `AddEntry`, `FindSimilar`, `FindSimilarWithThreshold`
- **In-memory** (`inmemory_cache.go`): Store `UserID` in entries, filter by user before similarity scoring in both HNSW and linear search paths
- **Redis** (`redis_cache.go`): Add `userID` parameter to methods (filter can be added to KNN query in follow-up)
- **Milvus** (`milvus_cache.go`): Add `userID` parameter to methods (filter expression update in follow-up)
- **Hybrid** (`hybrid_cache.go`): Forward `userID` to underlying backends
- **Stubs**: Updated to match interface
- **Callers** (`req_filter_cache.go`, `processor_res_body.go`): Extract `userID` from `RequestContext` via existing `getUserIDFromContext()`

When `userID` is empty (no auth backend configured), the cache operates globally — backward compatible with existing deployments.

User filtering happens BEFORE similarity scoring, matching the pattern in `memory/inmemory_store.go:94-97`.

## Changes

11 files, ~160 lines changed across cache interface, all 4 backends, stubs, callers, and tests.

## Test plan

- [x] All existing cache tests compile and pass (with `userID=""` for backward compat)
- [x] `pre-commit run` passes

### DIFF
--- a/config/testing/config.session-affinity-demo.yaml
+++ b/config/testing/config.session-affinity-demo.yaml
@@ -1,0 +1,189 @@
+# Session Affinity Demo Config
+# Demonstrates the route-bouncing problem in multi-turn conversations.
+#
+# Setup:
+#   - "expensive" model: Claude via claude-code-proxy on port 11480
+#   - "cheap" model: qwen2.5:0.5b (0.5B params) on Ollama port 11434
+#
+# Routing logic:
+#   - Complex queries (complexity signal "hard") → Claude (expensive)
+#   - Simple queries (complexity signal "easy") → qwen2.5:0.5b (cheap)
+#
+# The demo scenario:
+#   1. User sends complex coding question → routes to Claude (expensive)
+#   2. Claude responds with great code and asks "want me to add tests?"
+#   3. User says "yes" → VSR sees only "yes" → classifies as easy
+#   4. Routes to qwen2.5:0.5b (cheap) → 0.5B model has NO IDEA what "yes" means
+
+# Disable features we don't need for the demo
+semantic_cache:
+  enabled: false
+
+tools:
+  enabled: false
+
+prompt_guard:
+  enabled: true
+  model_id: "models/mom-jailbreak-classifier"
+  threshold: 0.7
+  use_cpu: true
+jailbreak_mapping_path: "models/mom-jailbreak-classifier/label_mapping.json"
+
+# Jailbreak rules for signal-based detection
+jailbreak_rules:
+  - name: "jailbreak_high"
+    threshold: 0.7
+    description: "High-confidence jailbreak detection"
+
+hallucination_mitigation:
+  enabled: false
+
+# Domain classifier (required by VSR startup)
+classifier:
+  category_model:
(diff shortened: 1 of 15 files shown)


rootfs:
The level of isolation needs to think twice. It is typical to have tenant, group, and user id, let's sort this out first so we can support different configurations.

yossiovadia:
The plumbing we added (a `userID` string on the cache interface) is scope-agnostic — it's a partition key, not specifically a user ID. It could be a tenant ID, group name, or user ID depending on what the operator wants.

What's missing is a config option to select which identity to use as the partition key. Something like:

```yaml
semantic_cache:
  isolation_scope: "user"  # "user" | "group" | "none"
```

Where `user` reads from `x-authz-user-id`, `group` reads from `x-authz-user-groups`, and `none` is the current behavior (global cache, backward compatible).

No tenant header exists in the codebase today — if needed, it could be added to `authz.identity` alongside user_id and user_groups.

Want me to add the `isolation_scope` config, or should we sort out the tenant/group identity model first?

yossiovadia:
## Closing — fundamentally wrong approach

After further thought, per-user cache isolation defeats the entire purpose of a semantic cache. A cache exists to share responses across similar queries — if User A asks "explain quantum computing" and it gets cached, User B asking the same thing should get the cached response. That's not a data leak, that's the cache working as intended.

Partitioning by `userID` would:
- **Destroy hit rates** — every user starts with a cold cache, only matching their own history
- **Multiply memory usage** by number of users
- **Remove the primary cost savings** that justify having a semantic cache at all

### The real problem (narrower)

Cross-user cache serving is only dangerous when the cached response contains **user-specific context** — e.g., RAG-injected private documents, memory retrieval with personal conversation history, or user-specific system prompts. For generic queries, sharing is correct behavior.

### Better approach

Instead of isolating the entire cache per user, **skip caching when user-specific context was injected**. The `RequestContext` already tracks:
- `RAGRetrievedContext` — RAG plugin injected private documents
- `MemoryContext` — personal memory was retrieved and injected
- `PIIDetected` — request contained PII

If any of these are present, the response is tainted with user-specific data and should not be cached. Otherwise, the cache should remain shared — that's where its value comes from.

Will open a replacement PR with this approach.

PR #1453, opened 2026-03-06
security: require user identity for feedback API submissions

### LINKED ISSUE #1452: [Security] Authenticate and rate-limit Router Learning outcome ingestion
## Summary

`POST /v1/router/outcomes` accepts typed Router Learning feedback linked to a replay record. On current `main` the route validates the payload but does not itself enforce authenticated identity, per-tenant authorization, rate limits, or replay ownership. A reachable endpoint can therefore be abused to poison model experience or append misleading outcomes.

The old description referenced `POST /api/v1/feedback` and direct Elo manipulation. That endpoint and public Elo contract are not current. `POST /api/v1/classify/user-feedback` is a separate classifier endpoint and does not ingest Router Learning outcomes.

## Audited current-main behavior (2026-07-15)

- Handler: `src/semantic-router/pkg/apiserver/route_router_outcomes.go`.
- Required fields include `replay_id`, typed `source`, `target`, and `verdict`.
- The active `routerruntime.LearningRuntime` appends a replay outcome and can update model-targeted in-process experience.
- Replay lookup and model resolution exist, but caller identity/tenant ownership are not established by the handler.
- Body-provided identity must not be treated as authenticated identity.

## Required security contract

1. Require a trusted authenticated principal at the management/API boundary; coordinate with #2463.
2. Authorize a specific outcome-write permission and tenant/scope, deny by default, and reject client-spoofable identity headers unless a trusted proxy boundary is configured.
3. Verify the caller may update the referenced replay record and target; prevent cross-tenant replay IDs.
4. Add per-principal and per-replay rate limits, bounded metadata, duplicate/idempotency semantics, and abuse metrics.
5. Define source-specific trust: user/agent/eval/operator/provider/router outcomes must not all receive identical weight by declaration alone.
6. Emit content-free audit events for accepted and rejected writes without logging raw prompt, response, or secrets.
7. Keep the endpoint unavailable or fail closed when authentication/authorization dependencies required by policy are absent.

## Acceptance criteria

- [ ] Anonymous and spoofed-identity submissions are rejected in production mode.
- [ ] Cross-tenant/cross-owner replay updates are rejected.
- [ ] Authorization, rate limits, idempotency, metadata bounds, and trust weighting are tested.
- [ ] Accepted outcomes remain linked to a valid replay record and auditable principal.
- [ ] Rejected writes cannot update experience or append an outcome.
- [ ] Security diagnostics contain no replay content or credentials.
- [ ] E2E covers direct API access and trusted-proxy deployment.

## Likely change surfaces

`src/semantic-router/pkg/apiserver/`, `src/semantic-router/pkg/routerruntime/`, `src/semantic-router/pkg/routerreplay/`, management API auth/rate-limit integration, docs, and E2E tests.

## Validation entrypoint

```bash
make agent-report ENV=cpu CHANGED_FILES="<space-separated changed files>"
```

yossiovadia:
## Summary

Fixes #1452 — the feedback API (`POST /api/v1/feedback`) accepted unauthenticated requests. Anyone reaching port 8080 could submit fake feedback to manipulate Elo ratings, influencing model selection for all users.

## Fix

Require user identity for every feedback submission:

1. **Auth header** (`x-authz-user-id`): trusted source from auth backend — takes precedence
2. **Request body** (`user_id`): fallback for development/testing
3. **Neither present**: reject with `401 USER_ID_REQUIRED`

The resolved `userID` overrides the body value when the auth header is present, preventing spoofing via the request body when an auth backend is configured.

## Changes

| File | Change |
|------|--------|
| `pkg/apiserver/route_feedback.go` | Add user identity resolution and 401 rejection |

1 file changed, 18 insertions, 3 deletions.

## Test plan

- [x] Anonymous feedback (no user_id, no auth header) returns 401
- [x] Feedback with `user_id` in body accepted (dev/testing path)
- [x] Auth header takes precedence over body `user_id`

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_feedback.go
+++ b/src/semantic-router/pkg/apiserver/route_feedback.go
@@ -51,7 +51,9 @@ type FeedbackResponse struct {
 	Timestamp string `json:"timestamp"`
 }
 
-// handleFeedback handles POST /api/v1/feedback for submitting model selection feedback
+// handleFeedback handles POST /api/v1/feedback for submitting model selection feedback.
+// Requires user identity via x-authz-user-id header (injected by auth backend) or
+// user_id in the request body. Anonymous feedback is rejected to prevent Elo rating manipulation.
 func (s *ClassificationAPIServer) handleFeedback(w http.ResponseWriter, r *http.Request) {
 	var req FeedbackRequest
 	if err := s.parseJSONRequest(r, &req); err != nil {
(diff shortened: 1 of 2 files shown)


rootfs:
@yossiovadia would that break the privacy?

yossiovadia:
The `user_id` field already existed in the FeedbackRequest struct and was already stored in the Feedback object when provided — this change just makes it required instead of optional. No new data is being collected.

Without requiring identity, anyone who can reach port 8080 can submit thousands of fake feedback entries and shift Elo ratings for all users. We demonstrated this in the security audit.

Some options if privacy is a concern:
1. Hash the user_id before storing in Elo ratings
2. Make it configurable (`feedback.require_identity: true/false`)
3. Don't implement the fix, but document the risk as a known limitation

Open to other ideas.

FAUST-BENCHOU (commented):
SGTM, could u resolve these conflicts?

Issue #1439, opened 2026-03-05
Multi-turn conversations can bounce between models (no session affinity)


yossiovadia:
## Summary

VSR evaluates each request independently — there is no session affinity. In a multi-turn conversation, the model selected for each turn depends solely on the **last user message**, not the conversation context. This means a simple follow-up like "commit it" can route to a completely different (cheaper) model than the one that produced the code being discussed.

## Reproduction

**Config**: Two decisions based on complexity signal — complex queries route to an expensive model, simple queries route to `qwen2.5:0.5b` (0.5B params).

### Turn 1: Complex coding question → expensive model

```
curl http://localhost:8801/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "auto",
    "messages": [
      {"role": "user", "content": "Write a Go function that implements a concurrent-safe LRU cache with O(1) get/put operations using a doubly-linked list and a sync.RWMutex. Include proper eviction logic."}
    ]
  }'
```

**VSR headers:**
```
x-vsr-selected-decision: complex_query
x-vsr-selected-model: expensive-model
x-vsr-matched-complexity: prompt_complexity:hard
```

**Response**: A correct, production-quality generic LRU cache implementation:

```go
type LRUCache[K comparable, V any] struct {
    mu       sync.RWMutex
    cap      int
    items    map[K]*entry[K, V]
    head     *entry[K, V]
    tail     *entry[K, V]
}

func (c *LRUCache[K, V]) Get(key K) (V, bool) {
    c.mu.Lock()
    defer c.mu.Unlock()
    e, ok := c.items[key]
    if !ok {
        var zero V
        return zero, false
    }
    c.moveToFront(e)
    return e.val, true
}

func (c *LRUCache[K, V]) Put(key K, val V) {
    c.mu.Lock()
    defer c.mu.Unlock()
    if e, ok := c.items[key]; ok {
        e.val = val
        c.moveToFront(e)
        return
    }
    if len(c.items) == c.cap {
        c.evict()
    }
    e := &entry[K, V]{key: key, val: val}
    c.items[key] = e
    c.pushFront(e)
}
```

### Turn 2: User says "looks good, commit it" → bounces to 0.5B model

```
curl http://localhost:8801/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "auto",
    "messages": [
      {"role": "user", "content": "Write a Go function that implements a concurrent-safe LRU cache..."},
[...]

PR #1275, opened 2026-02-07
[Feat] implement ModelEvaluator for automated onboarding flow (#1153)

aby42:
This PR implements the initial ModelEvaluator (OnboardEvaluate) as the first milestone for the model onboarding procedure.

To keep the discussion centralized, I have posted the detailed architectural plan, roadmap, and sample JSON output directly in the issue comments.

Please refer to the discussion here: #1153  (comment)

# Key Changes
Added `src/training/model_eval/onboard_eval.py` with core benchmarking logic.

Supported reasoning (ARC, MMLU-Pro) and performance metrics tracking.

This is a Draft PR intended for early architectural review of the code structure.

FIX #1153 

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to semantic-router! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[CLI]</code> for changes to the command-line interface tools.</li>
    <li><code>[Dashboard]</code> for changes to the dashboard or web UI.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
[...]

PR #1459, opened 2026-03-06
feat: Conversational Routing Momentum (CRM) — asymmetric low-pass filter for multi-turn session affinity

### LINKED ISSUE #1458: feature: conversational routing momentum (CRM) for multi-turn session affinity
## Summary

Implements Conversational Routing Momentum (CRM) to solve the session affinity problem described in #1439. CRM is an asymmetric low-pass filter — inspired by 60-year-old audio compressor attack/release dynamics — that prevents model bouncing in multi-turn conversations while remaining stateless, backward compatible, and essentially zero-cost.

## The Problem (recap from #1439)

VSR evaluates each request independently. `extractUserAndNonUserContent()` only keeps the LAST user message — even though the full conversation history is in the Chat Completions request. So "commit it" after a complex coding discussion classifies as easy and bounces to the cheap model.

## Why Existing Approaches Fall Short

| Approach | Limitation |
|----------|-----------|
| Concatenate all user messages for classification | Classifier trained on single prompts, not multi-message input. Trivial messages dilute complex ones equally. Expensive on long conversations. |
| Sticky sessions (pin to first turn's model) | Throws away intelligent routing entirely. Wrong first turn = stuck forever. Needs server-side session store. |
| Wire SelectionContext.SessionID | Only works for selectors that implement session awareness (GMTRouter, RL-Driven). Doesn't fix the classification problem. |

See [prior art survey](https://github.com/vllm-project/semantic-router/issues/1439#issuecomment-4014169815) — no existing LLM router or paper addresses intelligent cross-model session affinity.

## The Solution: Asymmetric Low-Pass Filter

Borrowed from audio compressor dynamics (fast attack, slow release):

- **Fast attack**: quickly escalate to expensive model when complexity spikes
- **Slow release**: gradually decay, keeping conversations stable across simple follow-ups

```
for each user_message in conversation:
    signal = complexityScore(message)
    if signal > momentum:
        momentum = attack × momentum + (1-attack) × signal   // fast escalation
    else:
        momentum = release × momentum + (1-release) × signal  // slow de-escalation

route based on momentum vs threshold
```

## Config

```yaml
routing_momentum:
  enabled: true    # disabled by default — existing behavior unchanged
  attack: 0.3      # fast escalation (lower = faster)
  release: 0.9     # slow de-escalation (higher = slower)
  threshold: 0.5   # momentum above this → expensive model
```

Related: #1439

yossiovadia:
## Summary

Fixes #1458 / Related #1439 — VSR evaluates each request independently, causing multi-turn conversations to bounce between models when follow-up messages are simple ("commit it" after complex coding discussion routes to cheap model).

## Why This Approach

We evaluated multiple approaches (see [full analysis on #1439](https://github.com/vllm-project/semantic-router/issues/1439#issuecomment-4014169815)):

| Approach | Why it falls short |
|----------|-------------------|
| Concatenate all messages for classification | Classifier trained on single prompts. Trivial messages dilute complex ones equally. Expensive on long conversations. |
| Sticky sessions | Throws away intelligent routing. Wrong first turn = stuck forever. Needs server-side session store. |
| Wire SelectionContext.SessionID | Only helps selectors that implement session awareness. Doesn't fix classification. |

No existing LLM router or academic paper addresses intelligent cross-model session affinity ([prior art survey](https://github.com/vllm-project/semantic-router/issues/1439#issuecomment-4014169815)).

## The Solution: Asymmetric Low-Pass Filter

Inspired by 60-year-old audio compressor dynamics — **fast attack** (respond instantly to loud/complex signals) and **slow release** (decay gradually after signal drops). Applied to LLM routing:

- **Fast attack** (default 0.3): quickly escalate to expensive model when complexity increases
- **Slow release** (default 0.9): resist de-escalation, keep conversations on the same model through simple follow-ups like "ok", "commit it", "thanks"

The asymmetry is critical — symmetric momentum (same rate up and down) fails the "good morning" → "develop kernel" scenario because the trivial first message resists upward escalation.

## How It Works (Stateless)

The Chat Completions API sends the **full conversation history** with every request. CRM exploits this — no server-side state needed:

1. Extract all user messages from the request body
2. For **historical** messages: use message length as a cheap complexity estimate (O(1) per message, nanoseconds)
3. For the **current** message: use the real embedding classifier score (already computed by existing pipeline — zero additional cost)
4. Run the asymmetric momentum filter over the signal sequence

### Walkthrough: 5-Turn Conversation

```
Turn 1: "Implement a concurrent LRU cache in Go with sync.RWMutex"
  → classifier score: 0.85 (complex)
  → only 1 message → CRM skipped → routes to expensive model ✓

Turn 2: "commit it"
  → historical: "Implement LRU cache..." → score: 0.85
  → current: "commit it" → classifier score: 0.08
  → momentum: 0.5 → 0.85 (attack) → 0.745 → 0.08 (release) → 0.678
  → 0.678 > 0.5 → expensive model ✓  (without CRM: 0.08 → cheap model ✗)

Turn 3: "thanks"
  → replays history + current classifier: 0.05
  → momentum: 0.5 → 0.745 → 0.678 → 0.05 (release) → 0.615
  → 0.615 > 0.5 → expensive model ✓

Turn 4: "bye"
  → replays history + current classifier: 0.03
  → momentum: 0.5 → 0.745 → 0.678 → 0.615 → 0.03 (release) → 0.556
  → 0.556 > 0.5 → expensive model ✓

Turn 5: "have a good day"
  → replays history + current classifier: 0.04
  → momentum after replay: → 0.506 → 0.04 (release) → 0.459
  → 0.459 < 0.5 → cheap model (conversation over) ✓
```

The key: turns 2-4 ("commit it", "thanks", "bye") are all part of the coding conversation. CRM holds them on the expensive model. Only after the conversation clearly winds down (turn 5) does it finally drop to cheap.

## Performance

| Operation | Cost | When |
|-----------|------|------|
| Embedding classification (current message) | ~700ms | Already happening (no change) |
| Length checks on N-1 historical messages | ~10ns each | New (CRM) |
| Momentum filter (N multiply-adds) | ~1ns each | New (CRM) |
| **Total CRM overhead for 100-turn conversation** | **~1μs** | **0.00014% of request** |

Technically O(n) vs the existing O(1), but O(n) × nanoseconds is invisible against O(1) × 700ms.

## Backward Compatibility

**100% backward compatible.** Disabled by default:

```yaml
routing_momentum:
  enabled: false  # default — existing behavior unchanged
```

When disabled:
- `extractAllUserMessages` is never called
- `applyRoutingMomentum` is never called
- The only always-active change: complexity classifier stores numerical scores in `SignalConfidences` (adds data, doesn't change existing string rules)

Config zero values = disabled. No behavior change for any existing deployment.

<details>
<summary>Changes</summary>

| File | Action | Description |
|------|--------|-------------|
| `pkg/classification/complexity_classifier.go` | Modify | Return numerical score (0.0-1.0) alongside string labels |
| `pkg/classification/classifier.go` | Modify | Store complexity scores in SignalConfidences |
| `pkg/config/config.go` | Modify | Add RoutingMomentumConfig with attack/release/threshold |
| `pkg/extproc/routing_momentum.go` | **Create** | CRM algorithm: ComputeRoutingMomentum, EstimateComplexityFromLength, applyRoutingMomentum |
| `pkg/extproc/routing_momentum_test.go` | **Create** | 14 unit tests |
| `pkg/extproc/utils.go` | Modify | Add extractAllUserMessages |
| `pkg/extproc/req_filter_classification.go` | Modify | Integrate CRM after signal eval |
| `pkg/extproc/processor_req_body.go` | Modify | Extract all user messages when CRM enabled |
| `pkg/extproc/processor_req_header.go` | Modify | Add AllUserMessages + RoutingMomentum to RequestContext |

9 files, ~480 lines (including 14 tests).

## Test Plan

14 Go unit tests (auto-discovered by CI via `go test ./...`):

- [x] `TestComputeRoutingMomentum_EmptySignals` — neutral on empty
- [x] `TestComputeRoutingMomentum_SingleComplexMessage` — fast attack
- [x] `TestComputeRoutingMomentum_SingleTrivialMessage` — slow release
- [x] `TestComputeRoutingMomentum_GoodMorningThenKernel` — fast escalation after trivial start
- [x] `TestComputeRoutingMomentum_ComplexThenCommitIt` — stays on expensive after "commit it"
- [x] `TestComputeRoutingMomentum_DecayOverManyTrivialTurns` — eventually decays after ~8 turns
- [x] `TestComputeRoutingMomentum_AsymmetricBehavior` — attack provably faster than release
- [x] `TestComputeRoutingMomentum_ClampsInputs` — handles out-of-range signals
- [x] `TestCRM_DisabledByDefault` — zero value = disabled
- [x] `TestCRM_DisabledDoesNotAffectSignals` — per-message routing preserved when off
- [x] `TestCRM_EnabledFixesRouteBouncing` — complex→simple stays expensive
- [x] `TestCRM_SimpleThenComplex` — simple→complex escalates quickly
- [x] `TestCRM_ConfigDefaults` — attack=0.2, release=0.95, threshold=0.5
- [x] `TestEstimateComplexityFromLength` — length estimate correctness

</details>

## Config Example

```yaml
routing_momentum:
  enabled: true
  attack: 0.2     # fast escalation (0.0 = instant, 1.0 = no change)
  release: 0.95   # slow de-escalation (1.0 = never decay, 0.0 = instant)
  threshold: 0.5  # momentum above this → prefer complex/expensive decision
```

## Demo: Model Quality Gap (why CRM matters)

Same prompt, single shot, no follow-ups — Claude Opus 4.6 vs Qwen 3.5 (35B, local Ollama) building a neon space shooter game:

| | Claude Opus 4.6 | Qwen 3.5 (35B local) |
|---|---|---|
| **Time** | ~15s | 356s (~6 min) |
| **Output** | 495 lines / 18KB | 935 lines / 29KB |
| **Playable?** | ✅ Fully functional | ❌ Only welcome screen works |

**▶ Play in browser (click to play, no download):**
- [Opus 4.6](https://gist.githack.com/yossiovadia/beb9d0a08c6b4a125d8f248ab150e7ad/raw/demo_opus_space_shooter.html)
- [Qwen 3.5](https://gist.githack.com/yossiovadia/beb9d0a08c6b4a125d8f248ab150e7ad/raw/demo_qwen_space_shooter.html)

More output ≠ better output. [Full comparison with prompt and details](https://gist.github.com/yossiovadia/beb9d0a08c6b4a125d8f248ab150e7ad). Routing a follow-up to the cheaper model mid-conversation — even with full context — degrades results. CRM prevents this.

<details>
<summary>The prompt (identical for both models)</summary>

```
Build a complete, single-file HTML game: a neon-themed space shooter.

Requirements:
- Canvas-based rendering, 60fps game loop
- Player ship with WASD movement + mouse aim/shoot
- 5 distinct enemy types with different movement patterns (zigzag, homing, circular, spawner, boss)
- Wave system: 10 waves with increasing difficulty, boss every 5th wave
- Power-up system: shield, rapid fire, spread shot, speed boost — dropped by enemies
- Particle system for explosions, thruster trails, bullet impacts (at least 3 different effects)
- Screen shake on hits, flash effects on damage
- HUD: score, wave number, health bar, active power-ups with duration timers
- Neon glow aesthetic: use CSS filters and shadow effects, dark background
- Game over screen with final score and restart button
- Sound effects using Web Audio API (procedurally generated, no external files)
- Smooth difficulty curve: enemy speed, spawn rate, and health scale per wave

Output ONLY the complete HTML file. No explanations, no markdown fences, just the raw HTML from <!DOCTYPE html> to </html>.
```

</details>


rootfs:
@yossiovadia 

```
Turn 2: "commit it"
  → historical: "Implement LRU cache..." → score: 0.85
  → current: "commit it" → classifier score: 0.08
  → momentum: 0.5 → 0.85 (attack) → 0.745 → 0.08 (release) → 0.678
  → 0.678 > 0.5 → expensive model ✓  (without CRM: 0.08 → cheap model ✗)
```

Wouldn't that `commit it` an easy task?

rootfs:
please also check if the #1368 solves the use case, cc @ppppqp

yossiovadia:
**Re: "Wouldn't that `commit it` an easy task?"**

Yes — `commit it` classifies as easy (score 0.08). That's exactly the problem. Without CRM, this easy score routes to the cheap model mid-conversation. CRM doesn't change how individual messages are classified — it applies an asymmetric filter over the conversation's signal history so that a single easy follow-up doesn't override the momentum established by the complex question that started the conversation. The momentum drops from 0.745 to 0.678 (release path) instead of jumping straight to 0.08 — still above the 0.5 threshold.

**Re: #1368 (Contrastive Preference Classifier)**

Reviewed — #1368 solves a different problem. The preference classifier matches queries against preference labels (WHAT the user wants), while CRM addresses routing stability across turns (which MODEL should handle the conversation). The preference classifier still evaluates each message independently — a "commit it" message wouldn't match any preference label either, so it would still bounce. CRM and the preference signal are complementary, not overlapping.

rootfs:
@yossiovadia why easy task to to cheap model a bad idea?

yossiovadia:
Because it changes the model mid-conversation. Example from our testing (#1439):

User asks to write a kernel from scratch. Capable model starts delivering. User says "looks good, commit it." Router sees a short simple message, switches to a cheaper model. That model just replies "I'm glad to hear that! Thank you" — it didn't commit anything.

The follow-up is simple to say but still needs the capable model to execute. "commit it", "now add tests", "deploy it" — these are short messages that require the same level of capability as the original task they refer to.

A sticky session (always pin to the first model) would solve this but loses the intelligent routing — if someone starts with "hello" and then asks something complex, they'd be stuck on the cheap model forever. CRM handles both directions: fast escalation when complexity increases, slow de-escalation when it drops.

rootfs:
@yossiovadia that sounds a context memory loss, did you turn on the context memory plugin?

rootfs:
@yossiovadia you used qwen 0.5b in your test, that's not a cheap model  per se. It is not meant to run for inference production.

yossiovadia:
I hear you on the 0.5B — that was a bad test choice. But the concern isn't about model capability, it's about switching mid-conversation.

Consider a finance decision with a reasoning model — user asks "plan my retirement given my salary and expenses." The model runs through a detailed analysis with specific assumptions. User follows up: "my doctor says I need surgery next year, how does that change the plan?"

Without CRM: that follow-up has medical keywords, crosses into a different domain, routes to a different model. The new model doesn't know what assumptions the first model made in the retirement calculation. Even if the new model is equally capable, it can't continue the financial analysis — it would start fresh.

With CRM: momentum from the first turn keeps the conversation on the original model, which already has the reasoning context to adjust its own calculation.

The same applies within a single domain — "commit it" after a coding session doesn't just mean "commit something." It means "commit the specific code you just wrote, with the specific structure you chose." A different model might interpret it differently even if it sees the conversation history.

CRM is off by default, zero cost when disabled, ~1μs when enabled. It's a safety net for deployments that use multi-model routing.

rootfs:
@yossiovadia using a complexity preset with a weak model is the root cause of your experiment. Please eval the models and create a complexity knowledge base accordingly, and run your test again, that'll give us more ideas. Please refer to #1275

yossiovadia:
Even with a well-calibrated knowledge base, short follow-ups like "commit it" or "now add tests" will classify as easy — they're semantically close to "yes" and "ok" regardless of the candidates. That's inherent to per-message evaluation, not a calibration gap.

#1275 helps pick the right models for each complexity tier, but the issue here is about what happens between turns — the routing decision itself, not which models are in the pool.

CRM doesn't replace good calibration — it complements it. Off by default, zero cost when disabled.

yossiovadia:
### The "High Affinity" Tuning Example

Here is a walkthrough demonstrating how tuning the `release` parameter to `0.95` maintains session affinity on the expensive model, even through a slump of low-complexity prompts.

**Configuration Parameters:**
* **Attack:** `0.3` (Fast escalation)
* **Release:** `0.95` (Very slow decay)
* **Threshold:** `0.5`

---

**Turn 1: "Explain the backend architecture."**
* **Complexity:** `0.85` (UP → use Attack)
* **Calculation:**
  $$C_{new} = (0.3 \cdot 0.0) + (0.7 \cdot 0.85)$$
* **Result:** **0.595** → **Expensive Model**

**Turn 2: "Where is the router file?"**
* **Complexity:** `0.3` (DOWN → use Release)
* **Calculation:**
  $$C_{new} = (0.95 \cdot 0.595) + (0.05 \cdot 0.3)$$
* **Result:** **0.580** → **Expensive Model**

**Turn 3: "Got it, thanks."**
* **Complexity:** `0.05` (DOWN → use Release)
* **Calculation:**
  $$C_{new} = (0.95 \cdot 0.580) + (0.05 \cdot 0.05)$$
* **Result:** **0.553** → **Expensive Model**

**Turn 4: "Bye."** *(This is the turn that previously failed!)*
* **Complexity:** `0.05` (DOWN → use Release)
* **Calculation:**
  $$C_{new} = (0.95 \cdot 0.553) + (0.05 \cdot 0.05)$$
* **Result:** **0.528** → **Expensive Model** *(It survived!)*

**Turn 5: "Wait, write a python script..."**
* **Complexity:** `0.9` (UP → use Attack)
* **Calculation:**
  $$C_{new} = (0.3 \cdot 0.528) + (0.7 \cdot 0.9)$$
* **Result:** **0.788** → **Expensive Model**

yossiovadia:
Comparison between local and opus : 
[https://gist.github.com/yossiovadia/beb9d0a08c6b4a125d8f248ab150e7ad 
](Link)
( copied from above PR description )

yossiovadia:
Hi @Xunzhuo @rootfs — friendly ping on this one. Happy to address any feedback or make changes if needed.

Re @rootfs's earlier comment about the complexity preset — I agree proper model calibration matters, but the core issue CRM solves is different: short follow-ups ("yes", "commit it", "add tests") get misrouted to cheaper models because per-message classification can't see that they're part of an ongoing complex task. No complexity classifier will rate "yes" as complex — that's where conversation momentum fills the gap.

The [demo comparison](https://gist.github.com/yossiovadia/beb9d0a08c6b4a125d8f248ab150e7ad) shows the quality gap when a task lands on the wrong model, and the attack/release parameters have been tuned to survive real dev workflows (~11 turns before decay).

Let me know if there's anything blocking this — happy to iterate.

Xunzhuo:
Should this be an algorithm for model selection? The inline models looks like not a proper place to abstract this feature

yossiovadia:
> Should this be an algorithm for model selection? The inline models looks like not a proper place to abstract this feature

  @Xunzhuo Thanks for the feedback. To make sure I refactor in the right direction — are you suggesting CRM should live under the decision's algorithm block rather than as a global
  config? Something like:

```
  decisions:
    - name: "complex_query"
      algorithm:
        type: "confidence"
        momentum:
          enabled: true
          attack: 0.2
          release: 0.95
```
  
  Instead of the current top-level routing_momentum under InlineModels?

  Happy to restructure — just want to confirm the right placement before I move things around.

yossiovadia:
## E2E Validation — Full 30-turn conversation

Tested locally (Mac, Ollama, Envoy + ExtProc) with `qwen2.5:1.5b` (expensive) and `qwen2.5:0.5b` (cheap). Complex prompts at turns 1 and 16, easy follow-ups everywhere else.

| Turn | Message | Momentum | Model | Status |
|------|---------|----------|-------|--------|
| 1 | COMPLEX | — | qwen2.5:1.5b | classifier match |
| 2 | easy | 0.748 | qwen2.5:1.5b | CRM hold |
| 3 | easy | 0.719 | qwen2.5:1.5b | CRM hold |
| 4 | easy | 0.690 | qwen2.5:1.5b | CRM hold |
| 5 | easy | 0.663 | qwen2.5:1.5b | CRM hold |
| 6 | easy | 0.637 | qwen2.5:1.5b | CRM hold |
| 7 | easy | 0.613 | qwen2.5:1.5b | CRM hold |
| 8 | easy | 0.590 | qwen2.5:1.5b | CRM hold |
| 9 | easy | 0.568 | qwen2.5:1.5b | CRM hold |
| 10 | easy | 0.547 | qwen2.5:1.5b | CRM hold |
| 11 | easy | 0.527 | qwen2.5:1.5b | CRM hold |
| 12 | easy | 0.508 | qwen2.5:1.5b | CRM hold |
| **13** | **easy** | **0.490** | **qwen2.5:0.5b** | **decayed** |
| 14 | easy | 0.473 | qwen2.5:0.5b | decayed |
| 15 | easy | 0.457 | qwen2.5:0.5b | decayed |
| **16** | **COMPLEX** | **0.771** | **qwen2.5:1.5b** | **fast attack ↑** |
| 17 | easy | 0.740 | qwen2.5:1.5b | CRM hold |
| 18 | easy | 0.711 | qwen2.5:1.5b | CRM hold |
| 19 | easy | 0.683 | qwen2.5:1.5b | CRM hold |
| 20 | easy | 0.656 | qwen2.5:1.5b | CRM hold |
| 21 | easy | 0.631 | qwen2.5:1.5b | CRM hold |
| 22 | easy | 0.607 | qwen2.5:1.5b | CRM hold |
| 23 | easy | 0.584 | qwen2.5:1.5b | CRM hold |
| 24 | easy | 0.562 | qwen2.5:1.5b | CRM hold |
| 25 | easy | 0.542 | qwen2.5:1.5b | CRM hold |
| 26 | easy | 0.522 | qwen2.5:1.5b | CRM hold |
| 27 | easy | 0.503 | qwen2.5:1.5b | CRM hold |
| **28** | **easy** | **0.486** | **qwen2.5:0.5b** | **decayed** |
| 29 | easy | 0.469 | qwen2.5:0.5b | decayed |
| 30 | easy | 0.453 | qwen2.5:0.5b | decayed |

Behavior: **expensive → holds 11 turns → cheap → complex prompt → expensive → holds 11 turns → cheap**

<details>
<summary>Test script (requires running router + envoy)</summary>

```bash
# Config: config/testing/config.crm-demo.yaml
# Decisions: complex_query (complexity:hard → qwen2.5:1.5b, momentum enabled)
#            general_query (domain:other → qwen2.5:0.5b)

COMPLEX1="Design a real-time stream processing pipeline with exactly-once semantics and fault tolerance"
COMPLEX2="Write a compiler optimization pass for loop vectorization with SIMD intrinsics and dependency analysis"
# zsh arrays are 1-indexed; index 0 is unused placeholder
EASY=("_" "ok" "sure" "thanks" "yes" "cool" "bye" "hi" "yo" "sup" "k" "no" "fine" "yep" "go ahead")

for turn in $(seq 1 30); do
  msgs='['
  is_complex=false
  for t in $(seq 1 $turn); do
    if [ $t -gt 1 ]; then msgs="${msgs},"; fi
    if [ $t -eq 1 ]; then
      msgs="${msgs}{\"role\":\"user\",\"content\":\"${COMPLEX1}\"}"
      if [ $t -eq $turn ]; then is_complex=true; fi
    elif [ $t -eq 16 ]; then
      msgs="${msgs}{\"role\":\"user\",\"content\":\"${COMPLEX2}\"}"
      if [ $t -eq $turn ]; then is_complex=true; fi
    else
      idx=$(( ((t-2) % 14) + 1 ))
      msgs="${msgs}{\"role\":\"user\",\"content\":\"${EASY[$idx]}\"}"
    fi
    if [ $t -lt $turn ]; then
      msgs="${msgs},{\"role\":\"assistant\",\"content\":\"OK.\"}"
    fi
  done
  msgs="${msgs}]"

  model=$(curl -s -m 60 http://localhost:8801/v1/chat/completions \
    -d "{\"model\":\"auto\",\"messages\":$msgs}" 2>&1 \
    | grep -o '"model":"[^"]*"' | head -1 | sed 's/"model":"//;s/"//')
  echo "Turn $turn: ${model:-timeout}"

  if [ "$is_complex" = true ]; then sleep 10; else sleep 3; fi
done
```

</details>

yossiovadia:
@Xunzhuo , @rootfs : 

### Why this matters

Today, VSR evaluates each request independently — only the last user message determines routing. In a multi-turn conversation, a simple follow-up like "commit it" or "now add tests" after a complex coding discussion classifies as easy and bounces to the cheap model. That model can't continue the work — it doesn't know what code was written, what assumptions were made, or what "it" refers to in context. Even if it sees the conversation history, a different model interprets it differently.

This isn't a calibration issue. No complexity classifier will ever rate "yes" or "commit it" as complex — they're genuinely simple messages. The problem is that per-message classification can't see they're part of an ongoing complex task.

### Approaches explored for CRM performance

I iterated through several approaches to make CRM production-viable. Here's what I tried and why I landed where I did.

**Approach 1: Asymmetric low-pass filter with full classifier on history**

The original CRM algorithm — an asymmetric filter (fast attack/slow release) inspired by audio compressor dynamics. Each request replays the full conversation, running `EvaluateAllSignals` on every historical user message to build a complexity signal sequence, then feeds it through the filter.

Problem: `EvaluateAllSignals` runs domain (~70ms) + complexity (~50ms) + keyword (~0.2ms) classifiers per message. Latency grows linearly:

| Turn | Routing overhead |
|------|-----------------|
| 1 | ~76ms |
| 10 | ~458ms |
| 30 | ~1,236ms |

Not acceptable for production.

**Approach 2: Lighter historical scoring (keyword + complexity only)**

Dropped the domain classifier for historical messages, keeping keyword (~0.2ms) + complexity (~50ms). Cut latency roughly in half:

| Turn | Full classifier | Keyword + Complexity |
|------|----------------|---------------------|
| 10 | 458ms | 234ms |
| 30 | 1,236ms | 622ms |

Better, but still linear — ~17ms per historical message. At 100 turns that's ~1.7s.

**Approach 3: Server-side memo cache**

Key insight: momentum is recursive (`m[i] = α * m[i-1] + (1-α) * signal`). If we cache the previous momentum value, we only need to score the current message — O(1) instead of O(n).

Implementation: `map[uint64]float64` keyed by FNV-64a hash of conversation prefix. On cache hit, one multiply-add. On miss, full replay as fallback.

Results: turn 30 dropped from 622ms to ~50ms (constant). But introduced concerns:
- ~1MB server-side state
- Mass eviction (clearing the entire map when full) causes a thundering herd of cache misses
- Hash computation still reads all message text on every request
- Breaks the "stateless" design that was a selling point of CRM

**Approach 4: Sliding window with keyword-only checks**

Dropped the filter entirely. Instead: "was any user message in the last N turns complex?" — a simple window scan with keyword checks (~0.2ms each). No state, no cache, constant cost.

Problem: the routing decision that sends messages to the expensive model often uses **complexity signals** (embedding similarity), not keywords. A message like "Design a distributed consensus algorithm" matches the complexity classifier via embeddings but has no keyword hits. The keyword-only window would miss it, and the conversation would drop to the cheap model on the very next turn.

**Approach 5: Response-length window (current)**

The insight: **the model's own output is ground truth for complexity.** When a complex task routes to the expensive model, that model produces a long, detailed response — and that response is sitting in the conversation history of every subsequent request. We don't need to re-classify anything.

Two signals, both essentially free:
1. **Assistant response length** — if any response in the window exceeds `response_threshold` chars, the model was doing real work. Checking `len(string)` is nanoseconds.
2. **User keyword matches** — supplementary signal, catches complex user messages that haven't been answered yet. ~0.2ms per check.

```yaml
momentum:
  enabled: true
  window: 11              # turns (not messages) to look back
  response_threshold: 200  # chars — responses longer than this = complex
```

Results:

| Turn | Routing overhead |
|------|-----------------|
| 1 | ~74ms |
| 10 | ~37ms |
| 30 | ~39ms |

- Fully stateless — zero server-side memory
- Constant latency regardless of conversation length
- No re-classification, no embeddings, no cache
- Window in turns: `window: 11` holds for exactly 11 turns after the last complex interaction

E2E validated locally (Ollama, Envoy + ExtProc, 30-turn test):
- Complex at turn 1 → expensive holds through turn 12 → decays at turn 13
- Complex at turn 16 → expensive holds through turn 27 → decays at turn 28

Would appreciate feedback on this direction before I clean up the PR. Happy to adjust the approach or the config knobs.

Xunzhuo:
Closing as superseded by #1974. The new policy keeps the multi-turn continuity goal and adds explicit tool-loop hard locks, idle re-selection, prefix-cache/cost-aware penalties, and replayable policy traces.

PR #1486, opened 2026-03-10
fix: switch default domain classifier to mmBERT-32K (52.6% → 65.6%)

### LINKED ISSUE #1485: [Research] Improve domain classifier generalization on held-out and conversational data
### Problem

The maintained domain classifier has reported failures on unseen and conversational inputs, but the issue does not yet contain a versioned held-out baseline that separates model quality from preprocessing, labels, thresholds, and routing policy.

### Scope

- Reproduce the gap against the shared baseline in #3197.
- Identify affected data slices and the dominant failure source.
- Propose a bounded data, objective, calibration, or artifact change only if the evidence isolates a model-quality gap.
- Compare the candidate with the unchanged default on quality, calibration, robustness, latency, and runtime compatibility.

### Completion signal

A reproducible report either justifies and qualifies a new artifact or closes the gap as non-model work with a linked owning issue.

Parent Epic: #3194

yossiovadia:
## Summary

Switch the default domain classifier from BERT-base + LoRA (`LLM-Semantic-Router/lora_intent_classifier_bert-base-uncased_model`) to the mmBERT-32K merged model (`llm-semantic-router/mmbert32k-intent-classifier-merged`).

Benchmark on 994 unseen queries shows **52.6% → 65.6% accuracy** (+13 points). The biggest gains are in the weakest domains: engineering 15.5%→70.4%, business 9.9%→53.5%, history 23.9%→60.6%.

Full benchmark data and methodology in #1485.

## What changed

**`registry.go`** — Updated the `mom-domain-classifier` entry:
- `RepoID`: `LLM-Semantic-Router/lora_intent_classifier_bert-base-uncased_model` → `llm-semantic-router/mmbert32k-intent-classifier-merged`
- `UsesLoRA`: `true` → `false` (merged model, no LoRA adapter needed)
- `MaxContextLength`: `512` → `32768` (merged model natively supports 32K, no LoRA training limitation)
- Removed `BaseModelMaxContext` field (not needed when the model directly supports full context)

**`registry_test.go`** — Updated expected repo ID to match the new default.

**`registry_max_context_test.go`** — Updated expected `MaxContextLength` from 512 to 32768 for the domain classifier, since the merged model is not constrained by LoRA training length.

## Backwards compatibility

- All existing aliases (`domain-classifier`, `intent-classifier`, `category-classifier`, `lora_intent_classifier_bert-base-uncased_model`, etc.) still resolve to `models/mom-domain-classifier`. Existing configs that reference the old model name will continue to work — they'll just get the better model.
- The `models/mom-domain-classifier` local path is unchanged, so deployment scripts and config files don't need updates.

## Resource impact

| Metric | BERT+LoRA (old) | mmBERT-32K (new) |
|--------|:-:|:-:|
| Inference speed | 14ms/query | 22ms/query |
| RAM | ~420 MB | ~1.2 GB |
| Disk | 419 MB | 1.2 GB |
| Accuracy (994 queries) | 52.6% | 65.6% |

## Benchmark tool

A standalone benchmark script (994 labeled test queries, runs in ~24s on CPU) will be submitted as a separate PR to `src/training/model_eval/`.

Fixes #1485

### DIFF
--- a/src/semantic-router/pkg/config/registry.go
+++ b/src/semantic-router/pkg/config/registry.go
@@ -66,17 +66,16 @@ type ModelSpec struct {
 var DefaultModelRegistry = []ModelSpec{
 	// Domain/Intent Classification
 	{
-		LocalPath:           "models/mom-domain-classifier",
-		RepoID:              "LLM-Semantic-Router/lora_intent_classifier_bert-base-uncased_model",
-		Aliases:             []string{"domain-classifier", "intent-classifier", "category-classifier", "category_classifier_modernbert-base_model", "lora_intent_classifier_bert-base-uncased_model"},
-		Purpose:             PurposeDomainClassification,
-		Description:         "Domain/intent classifier using BERT-base-uncased LoRA adapter. Can be used with ModernBERT-base-32k base model for extended context, but LoRA weights were trained on 512-token context.",
-		ParameterSize:       "149M (ModernBERT-base-32k) + classifier",
-		UsesLoRA:            true,
-		NumClasses:          14,    // MMLU categories
-		MaxContextLength:    512,   // LoRA weights trained on 512 tokens - safe maximum
-		BaseModelMaxContext: 32768, // Base model (ModernBERT-base-32k) supports 32K, but use with caution beyond MaxContextLength
-		Tags:                []string{"classification", "lora", "mmlu", "domain", "modernbert"},
+		LocalPath:        "models/mom-domain-classifier",
+		RepoID:           "llm-semantic-router/mmbert32k-intent-classifier-merged",
+		Aliases:          []string{"domain-classifier", "intent-classifier", "category-classifier", "category_classifier_modernbert-base_model", "lora_intent_classifier_bert-base-uncased_model"},
+		Purpose:          PurposeDomainClassification,
+		Description:      "mmBERT-32K merged intent classifier for MMLU-Pro categories",
+		ParameterSize:    "307M",
+		UsesLoRA:         false,
+		NumClasses:       14, // MMLU categories
+		MaxContextLength: 32768,
+		Tags:             []string{"classification", "merged", "mmlu", "domain", "mmbert-32k"},
 	},
 
 	// PII Detection - BERT LoRA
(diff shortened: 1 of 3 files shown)


Xunzhuo:
i think we just need to change the yaml configuration to choose the default model: https://github.com/yossiovadia/semantic-router/blob/7d5c39d82451a7ba6896f5caf629eafa3e3274b2/src/semantic-router/pkg/config/registry.go#L319.

and it is already changed in: src/vllm-sr/cli/templates/router-defaults.yaml

Xunzhuo:
i think we need to think about how to manage the configuration and defaults fields well in v0.3

yossiovadia:
Thanks for the review @Xunzhuo — you're right that `router-defaults.yaml` already points to the 32K model, so CLI users (`vllm-sr init`) get the right default.

However, the k8s deployment path still uses the old model. 13 out of 14 k8s value files reference `models/mom-domain-classifier`, which resolves via the Go registry to `LLM-Semantic-Router/lora_intent_classifier_bert-base-uncased_model` (the BERT-base + LoRA model at 52.6% accuracy):

```
deploy/kubernetes/aibrix/semantic-router-values/values.yaml  → models/mom-domain-classifier
deploy/kubernetes/dynamo/semantic-router-values/values.yaml   → models/mom-domain-classifier
deploy/kubernetes/istio/semantic-router-values/values.yaml    → models/mom-domain-classifier
deploy/kubernetes/llm-d/semantic-router-values/values.yaml    → models/mom-domain-classifier
e2e/profiles/production-stack/values.yaml                     → models/mom-domain-classifier
... (13 total)
```

Only `deploy/kubernetes/ai-gateway/` already uses `mmbert32k-intent-classifier-merged`.

So users deploying via the k8s guides still get the weaker model. I'll update this PR to align the k8s values with the CLI defaults instead of changing the registry. Does that approach make sense?

rootfs:
@yossiovadia we can use compression to retain the classification accuracy https://github.com/vllm-project/semantic-router/pull/1437

yossiovadia:
Thanks @rootfs — the compression work in #1437 is interesting, especially how domain match degrades with compression ratio.

I think the accuracy discussion belongs on #1485 rather than this PR. We've since validated our benchmark results against MMLU (14K independent questions) and the picture is more nuanced than the initial report suggested. I'll update #1485 with those findings.

Happy to close this PR — the default model question is better handled as part of the broader config cleanup @Xunzhuo mentioned for v0.3.

Xunzhuo:
@yossiovadia yes, we need to find a elegant way to manage our configuration and make the API stable

yossiovadia:
Let's close this one and track it on #1485

PR #1496, opened 2026-03-11
  Fix traditional Candle BERT classifier registration after initialization

pugafran:
## Summary

  This fixes a bug in the traditional Candle BERT classifier initialization path.

  `init_candle_bert_classifier` successfully initialized `TraditionalBertClassifier` models, but did not store the initialized instance in the global `TRADITIONAL_BERT_CLASSIFIER`
  registry. As a result, later calls to `ClassifyCandleBertText` could fail at inference time because no classifier was actually available in the global state.

  ## Problem

  For traditional BERT sequence-classification models, the initialization path looked successful, but the classifier instance was effectively dropped after construction.

  That created a mismatch between:

  - initialization succeeding
  - inference expecting a registered global classifier

  In practice, this caused runtime failures in flows that rely on `InitCandleBertClassifier(...)` followed by `ClassifyCandleBertText(...)`.

  ## Root cause

  In the `ModelType::Traditional` branch, the code returned `true` when `TraditionalBertClassifier::new(...)` succeeded, but it never inserted the classifier into
  `TRADITIONAL_BERT_CLASSIFIER`.

  This meant the classifier was constructed but never made available to later inference calls.

  ## Fix

  Store the initialized traditional classifier in:

  - `crate::model_architectures::traditional::bert::TRADITIONAL_BERT_CLASSIFIER`

  using the same `OnceLock<Arc<_>>` pattern already used elsewhere in the binding.

  ## Impact

  This fixes inference for custom traditional BERT category/intent classifiers that use the Candle BERT path.

  Before this patch:
  - initialization could succeed
  - inference could still fail because the classifier was not registered

  After this patch:
  - successful initialization also makes the classifier available for subsequent inference

  ## Scope

  This change only affects the traditional BERT initialization path in the Candle binding.
  It does not change LoRA initialization logic or ModernBERT behavior.


rootfs (approved):

rootfs:
@pugafran can you fix the precommit error?

pugafran:
I think that the current failure is not caused by the logic of the fix itself, but by existing repository-wide Rust gates triggered by touching `candle-binding/src/
  ffi/init.rs`:

  - structural gate: `init.rs` exceeds the file size limit
  - structural gate: `init_unified_classifier_c` exceeds the function length limit
  - full `cargo clippy --all-targets -D warnings` also reports many pre-existing issues in the crate
  - etc..

If you could give me a hand, I'll try to fix it later, but I'm feeling a bit out of sorts.

pugafran:
> I think that the current failure is not caused by the logic of the fix itself, but by existing repository-wide Rust gates triggered by touching `candle-binding/src/ ffi/init.rs`:
> 
> * structural gate: `init.rs` exceeds the file size limit
> * structural gate: `init_unified_classifier_c` exceeds the function length limit
> * full `cargo clippy --all-targets -D warnings` also reports many pre-existing issues in the crate
> * etc..
> 
> If you could give me a hand, I'll try to fix it later, but I'm feeling a bit out of sorts.

@rootfs

rootfs:
@pugafran if you like to take the refactor up in this PR, that'll be great. Or we can merge the PR for now and break the CI. I have sometime this morning to fix it up, so we don't block other PRs, wdyt?

pugafran:
@rootfs If you could do it, I'd appreciate it. My motherboard broke, and I can only programme from my work laptop.

Also, I might make decisions that aren't the best, and I've only just started working on the semantic router project.

If you can, merge it now and fix it later. That's fine with me.

rootfs:
@pugafran Merged. I'll pick up your change and fix the CI.

PR #1498, opened 2026-03-11
feat: calibrated token counter with online regression from LLM usage

rootfs:
The context length estimator is not accurate. This PR uses regression based estimate to be model and prompt content agnostic.


Xunzhuo:
covered in https://github.com/vllm-project/semantic-router/pull/2004

Issue #1501, opened 2026-03-12
observability: add metrics for cache writes skipped due to personalized context


yossiovadia:
## Summary

When the semantic cache skips a write because the response contains personalized context (RAG, memory, PII, injected system prompt), there's no observability into how often this happens. In production, operators need to know the skip rate to tune cache effectiveness.

**Severity: Low** — no functional impact. This is a monitoring gap, not a bug.

## Context

PR for #1448 adds `hasPersonalizedContext()` checks that skip cache writes when responses are tainted with user-specific data. The logic works correctly, but skipped writes are only logged at INFO level — no Prometheus metrics, no tracing attributes.

## Suggested implementation

Add a Prometheus counter in `pkg/observability/metrics/`:

```go
metrics.RecordCacheWriteSkipped(reason string)
// reason: "rag_context", "memory_context", "pii_detected", "system_prompt_injected"
```

Call it from `updateResponseCache()` and `cacheStreamingResponse()` when `hasPersonalizedContext()` returns true.

Optionally add a tracing attribute so individual requests show why cache was skipped.

Found during work on #1448.

Issue #1500, opened 2026-03-12
bug: cache hits bypass the RAG and memory injection pipeline


yossiovadia:
## Summary

Cache reads happen before RAG and memory injection in the request processing pipeline. When there's a cache hit, the response is returned immediately and RAG/memory retrieval never runs — even if the decision has RAG enabled or the user has conversation memory.

**Severity: Low** — this is a correctness issue, not a security issue. The user gets a valid (generic) answer, just not their personalized one.

## Pipeline order (current)

```
runRequestPreRoutingStages():
  1. applyRateLimitAndCacheChecks()  →  handleCaching() — cache READ here
  2. executeRAGPlugin()              →  sets ctx.RAGRetrievedContext
  3. prepareRequestForModelRouting() →  handleMemoryRetrieval() — sets ctx.MemoryContext
```

If step 1 returns a cache hit, steps 2-3 never execute.

## Impact

- User with RAG-enabled decision gets a generic cached response instead of their document-augmented one
- User with memory enabled gets a cached response without their conversation history
- Only affects requests where a semantically similar query was previously cached from a non-personalized request

## Possible fixes

1. **Move cache read after RAG/memory** — check cache only after all context augmentation, using the full augmented query as the cache key
2. **Skip cache reads for decisions with RAG or memory enabled** — simpler, but reduces cache effectiveness for those decisions
3. **Include a "personalization hash" in the cache key** — hash of (has_rag, has_memory, user_id) so personalized and generic queries don't collide

Option 2 is probably the right tradeoff — decisions with RAG/memory enabled should not serve cached responses since the response depends on user-specific context.

Found during work on #1448.

NJX-njx:
I would like to work on this in a dedicated branch and PR.
I will first trace the current cache read flow around RAG and memory injection, then implement a focused fix with tests.

FAUST-BENCHOU:
https://github.com/vllm-project/semantic-router/blob/88b9e1f24cd853a34c80dbd4722fdbbeaff51609/src/semantic-router/pkg/extproc/cache_personalization_pipeline_test.go#L94-L97

Has been fixed by https://github.com/vllm-project/semantic-router/pull/1533

PR #1502, opened 2026-03-12
[Bugfix] Skip cache for responses with personalized context

### LINKED ISSUE #1448: security: semantic cache has no user isolation — cross-user data leak

yossiovadia:
## Summary

Fixes #1448 — the semantic cache could serve responses tainted with user-specific context (RAG documents, conversation memory, PII) to other users.

**Replaces #1449** (per-user cache isolation), which was closed because partitioning the entire cache by userID defeats the purpose of a shared cache — hit rates collapse, memory multiplies by user count, and the cost savings disappear.

## Root cause

Cache writes happened unconditionally. When a response was augmented with private context (RAG-injected documents, memory retrieval, PII), it was cached and could be served to other users asking semantically similar questions.

### Layer 1: request time (`decisionWillPersonalize`)
Checks the decision **config** before any processing. If RAG or memory plugins are enabled (per-decision or global), the entire cache path is skipped — no embedding computation, no cache lookup, no `AddPendingRequest`. Zero wasted work.

### Layer 2: response time (`hasPersonalizedContext`)
Checks **runtime state** after processing. Catches signals that can't be predicted from config alone:
- `RAGRetrievedContext` — RAG plugin injected private documents
- `MemoryContext` — personal conversation memory retrieved
- `PIIDetected` — request contained PII
- `VSRInjectedSystemPrompt` — decision-specific system prompt injected

Guards both streaming (`cacheStreamingResponse`) and non-streaming (`updateResponseCache`) write paths.

**Generic responses remain cached and shared** — that's where the cache's value comes from.

## Changes

| File | Change |
|------|--------|
| `req_filter_cache.go` | Add `hasPersonalizedContext()`, `decisionWillPersonalize()`, skip cache path when personalization detected |
| `processor_res_body_pipeline.go` | Guard `updateResponseCache` with `hasPersonalizedContext` |
| `processor_res_body_streaming.go` | Guard `cacheStreamingResponse` with `hasPersonalizedContext` |
| `req_filter_cache_personalized_test.go` | 21 test cases covering both layers |
| `processor_res_body_streaming_test.go` | 4 integration tests: streaming skip for RAG/memory/PII + allow for generic |

5 files, 315 insertions.

## Test plan

All tests run locally with Rust libs built (`make rust`):

- [x] `TestHasPersonalizedContext` — 6 cases: generic, RAG, memory, PII, system prompt, combined
- [x] `TestDecisionWillPersonalize` — 6 cases: no decision, RAG enabled/disabled, per-decision memory, global memory, no plugins
- [x] `TestUpdateResponseCacheSkipsPersonalizedContext` — 5 cases proving non-streaming write path skips for each signal
- [x] `TestCacheStreamingSkippedWhenRAGContextPresent` — streaming path, RAG context → no cache write
- [x] `TestCacheStreamingSkippedWhenMemoryContextPresent` — streaming path, memory → no cache write
- [x] `TestCacheStreamingSkippedWhenPIIDetected` — streaming path, PII → no cache write
- [x] `TestCacheStreamingAllowedForGenericRequest` — streaming path, generic → cache write proceeds
- [x] All existing extproc tests pass (10.2s, 0 failures)
- [x] `go build ./pkg/extproc/` — compiles clean
- [x] `go vet ./pkg/extproc/` — no issues

## Known limitations (tracked separately)

- **#1500** — Cache reads happen before RAG/memory in the pipeline, so a cache hit bypasses personalization entirely (pre-existing, low severity)
- **#1501** — No Prometheus metrics for skipped cache writes (monitoring gap, low severity)

FIX #1448


Xunzhuo:
hey @yossiovadia can you assign the issues you are working on? prevent others working on the same

Xunzhuo:

yossiovadia:
Validated on live lab. 
RAG/memory/PII responses correctly skip cache. 
Generic responses still cache. The fix works as designed

yossiovadia:
The fix works by default, no configuration needed. There's no flag to turn it on or off.

  The logic is hardcoded in the response path:

  func (ctx *RequestContext) HasPersonalizedContext() (bool, string) {
      if ctx.RAGRetrievedContext != "" { return true, "rag_context" }
      if ctx.MemoryContext != ""       { return true, "memory_context" }
      if ctx.PIIDetected               { return true, "pii_detected" }
      return false, ""
  }

  If a decision has RAG or memory plugins enabled, the cache write is automatically skipped. There's also a request-time check (decisionWillPersonalize) that skips the entire cache path
  — including the embedding computation — when RAG or memory plugins are configured on the decision.

  So the behavior is:
  - You enable semantic-cache plugin on a decision → cache works
  - You also enable RAG or memory plugin on that same decision → cache is automatically bypassed, no extra config
  - PII detected in the request → cache skipped for that request

Xunzhuo:
@yossiovadia logics make sense, thanks

yossiovadia:
@Xunzhuo  Could you clarify what you mean by "unify the previous logics"?

The current code has two checks that look similar but run at different points in the pipeline:

1. decisionWillPersonalize (request time, added by this PR) — checks the decision's config. If RAG or memory plugins are enabled, skips the entire cache path: no embedding, no lookup, no AddPendingRequest. This is the core of this PR.
2. ctx.HasPersonalizedContext() (response time, already on main via [#1529](https://github.com/yossiovadia/semantic-router/issues/1529)) — checks runtime state after the response comes back. Catches cases like PII detection that can't be known from config alone.

They can't be merged into one function because they run at different lifecycle stages and check different things (config vs runtime). But happy to restructure if you have something specific in mind — just want to make sure I understand what you're looking for.

PR #1529, opened 2026-03-12
observability: add Prometheus metric and tracing for cache write skips

mkoushni:
Cache Write Skip Observability

## Summary

When the semantic cache skips a write because the response contains personalized context (RAG, memory, PII, injected system prompt), there was no observability. Operators had no way to see how often writes were skipped or why, which made it hard to tune cache effectiveness in production.

This PR adds a Prometheus counter and optional tracing so skip rate and reasons can be monitored.

**Severity:** Low — monitoring gap only; no change to cache behavior.

---

### 1. Prometheus metric (`pkg/observability/metrics/metrics.go`)

- **Counter:** `llm_cache_write_skipped_total` with label `reason`.
- **Reasons:** `rag_context`, `memory_context`, `pii_detected`, `system_prompt_injected`.
- **Helper:** `RecordCacheWriteSkipped(reason string)` (normalizes empty reason to `unknown`).
- **Constants:** `CacheWriteSkipReasonRAGContext`, `CacheWriteSkipReasonMemoryContext`, `CacheWriteSkipReasonPIIDetected`, `CacheWriteSkipReasonSystemPromptInjected` for consistent labels.

### 2. Personalized-context detection (`pkg/extproc/request_context.go`)

- **`HasPersonalizedContext() (bool, string)`** on `*RequestContext`:
  - Returns `(true, reason)` when the request/response is considered tainted:
    - `RAGRetrievedContext != ""` → `rag_context`
    - `MemoryContext != ""` → `memory_context`
    - `PIIDetected` → `pii_detected`
    - `VSRInjectedSystemPrompt` → `system_prompt_injected`
  - Returns `(false, "")` otherwise.
  - One reason per call (first match in the order above) for stable metrics and tracing.

### 3. Non-streaming path (`pkg/extproc/processor_res_body_pipeline.go`)

- In **`updateResponseCache()`**, before calling the cache:
  - If `ctx.HasPersonalizedContext()` is true:
    - Call **`metrics.RecordCacheWriteSkipped(reason)`**.
    - Log at INFO with request ID and reason.
    - If the current span is recording, set attribute **`cache.write_skipped_reason`**.
    - Return without writing to the cache.

### 4. Streaming path (`pkg/extproc/processor_res_body_streaming.go`)

- In **`cacheStreamingResponse()`**, after streaming preconditions and before caching:
  - Same logic: if `ctx.HasPersonalizedContext()` is true, call **`RecordCacheWriteSkipped(reason)`**, log, set **`cache.write_skipped_reason`** on the span when recording, and return without caching.

### 5. Tracing (`pkg/observability/tracing/tracing.go`)

- **`AttrCacheWriteSkippedReason = "cache.write_skipped_reason"`** so individual requests show why the cache write was skipped in traces.

---

## Observability

- **Prometheus:**  
  `llm_cache_write_skipped_total{reason="rag_context|memory_context|pii_detected|system_prompt_injected"}`  
  Use for skip rate and breakdown by reason (e.g. dashboards, alerts).

- **Logs:**  
  INFO log when a write is skipped, e.g.  
  `Skipping cache write for request ID <id>: response has personalized context (reason=<reason>)`.

- **Tracing:**  
  When a span is recording, attribute **`cache.write_skipped_reason`** is set so each request shows why the cache was skipped.

---

## Testing

- Existing behavior unchanged when there is no personalized context; cache writes proceed as before.
- When RAG/memory/PII/system-prompt context is present, cache write is skipped, metric is incremented, and (if tracing is on) the span gets `cache.write_skipped_reason`.

---

## Checklist

- [ ] Metric and recorder added in `pkg/observability/metrics`.
- [ ] `HasPersonalizedContext()` implemented and used in both non-streaming and streaming cache paths.
- [ ] Tracing attribute added and set when skipping.
- [ ] No functional change to cache logic; only observability added.

### DIFF
--- a/candle-binding/src/ffi/init.rs
+++ b/candle-binding/src/ffi/init.rs
@@ -1282,9 +1282,10 @@ pub extern "C" fn init_candle_bert_classifier(
                 num_classes as usize,
                 use_cpu,
             ) {
-                Ok(_classifier) => {
-                    // Store in global static (would need to add this to the lazy_static block)
-                    true
+                Ok(classifier) => {
+                    crate::model_architectures::traditional::bert::TRADITIONAL_BERT_CLASSIFIER
+                        .set(Arc::new(classifier))
+                        .is_ok()
                 }
                 Err(e) => {
                     eprintln!("Failed to initialize Candle BERT classifier: {}", e);


Xunzhuo (approved):
LGTM

PR #1534, opened 2026-03-12
[Router][Bugfix] Skip semantic cache for personalized request flows

### LINKED ISSUE #1500: bug: cache hits bypass the RAG and memory injection pipeline

NJX-njx:
Closes #1500

## Summary

- Scope: Skip semantic-cache reads and writes for requests that can be personalized by RAG or memory.
- Primary skill: router-core, because this change stays in the extproc cache and request and response processing flow.
- Impacted surfaces: plugin_runtime.
- Conditional surfaces intentionally skipped: local_e2e, ci_e2e, local_smoke, because the current shell environment does not have a usable Go toolchain.
- Behavior-visible change: yes
- Debt entry: none

## Validation

- Environment: `not run`
- Fast gate: Not run. `go` and `gofmt` are unavailable in the current shell environment.
- Feature gate: Not run. `go` and `gofmt` are unavailable in the current shell environment.
- Local smoke / E2E: Not run. `go` and `gofmt` are unavailable in the current shell environment.
- CI expectations / blockers: Added focused unit tests for cache personalization skip behavior, but local execution is blocked until a Go toolchain is available.

## Checklist

- [x] Commits in this PR are signed off with `git commit --trailer "Made-with: Cursor" -s`


Xunzhuo:
is it duplicated with https://github.com/vllm-project/semantic-router/pull/1502?

NJX-njx:
Hi @Xunzhuo, good question!

This PR (#1534) is different from #1502:

- **#1502**: Skips cache entirely when RAG/memory plugins are enabled (config-time check + runtime check for PII/system prompt)
- **#1534**: For personalized flows (RAG/memory enabled AND actually used), skip cache reads/writes

The key difference: #1502 skips cache at config level (if RAG/memory is configured), while #1534 skips at runtime when the request actually has personalized context.

#1534 is more granular - it only bypasses cache when we know the response is personalized, not just when the capability is configured.

They could potentially be combined, but they address different aspects of the personalization issue. What do you think?

yossiovadia:
> Hi @Xunzhuo, good question!
> 
> This PR (#1534) is different from #1502:
> 
> * **[[Bugfix] Skip cache for responses with personalized context #1502](https://github.com/vllm-project/semantic-router/pull/1502)**: Skips cache entirely when RAG/memory plugins are enabled (config-time check + runtime check for PII/system prompt)
> * **[[Router][Bugfix] Skip semantic cache for personalized request flows #1534](https://github.com/vllm-project/semantic-router/pull/1534)**: For personalized flows (RAG/memory enabled AND actually used), skip cache reads/writes
> 
> The key difference: #1502 skips cache at config level (if RAG/memory is configured), while #1534 skips at runtime when the request actually has personalized context.
> 
> #1534 is more granular - it only bypasses cache when we know the response is personalized, not just when the capability is configured.
> 
> They could potentially be combined, but they address different aspects of the personalization issue. What do you think?

Hi @NJX-njx in which scenario can any LLM reply with personal info if rag/memory was not sent to the model as context? 
In other words, I mean, models has no personal info unless it's being fed with the info. ( which #1502 prevents ) or am i missing something?

NJX-njx:
Good question @yossiovadia. You're right that #1502 prevents personal data from being sent to the model when RAG/memory retrieval is suppressed. This PR addresses a different vector: **cache poisoning from personalized responses**.

Consider this sequence:
1. User A asks "What is my monthly budget?" with memory enabled. Memory context about User A's finances is injected. The model responds with User A's budget details.
2. This response gets cached by semantic cache keyed on the query text.
3. User B asks "What is my monthly budget?" — without this PR, the semantic cache returns User A's personalized response to User B (cache hit on similar query).

Even though #1502 prevents User B's memory from being injected, the *cached response* from User A's personalized request is the leak vector. This PR prevents writing personalized responses to the cache in the first place, which is orthogonal to #1502's protection.

The two approaches are complementary as defense-in-depth:
- **#1502**: prevents personal context from being *sent* to the model
- **#1534 (this PR)**: prevents personalized *responses* from being cached and served to other users
- **#1538**: isolates cache entries per user namespace

Also fixed the `StructuredPayload` type errors and merged latest main in this push.

src/semantic-router/pkg/extproc/req_filter_cache.go
```diff
@@ -171,3 +179,69 @@ func (r *OpenAIRouter) performCacheLookup(
...
+	if ctx != nil && ctx.VSRSelectedDecision != nil {
+		memoryConfig := ctx.VSRSelectedDecision.GetMemoryConfig()
+		if memoryConfig != nil {
+			return memoryConfig.Enabled
+		}
+	}
+	return r.Config.Memory.Enabled
+}
+
+func (r *OpenAIRouter) shouldSkipSemanticCacheWrite(ctx *RequestContext) (bool, string) {
```
rootfs, line 234: can you make this optional?

PR #1558, opened 2026-03-15
[Bugfix][Router] Enforce per-decision semantic-cache opt-out on store paths

### LINKED ISSUE #1546: bug: semantic cache stores responses for decisions without semantic-cache plugin when global cache is enabled
## Describe the bug

When global semantic cache is enabled, a decision that does not enable the `semantic-cache` plugin still ends up writing response bodies into cache.

The intended per-decision contract already says that decisions without an explicit `semantic-cache` plugin should behave as cache-disabled, but the runtime only applies that check to part of the request path. Miss/store paths still create pending entries and complete them on response, so per-decision opt-out is not actually enforced.

## To Reproduce

1. Set global `semantic_cache.enabled: true`.
2. Create a decision that matches a request but either omits the `semantic-cache` plugin entirely or sets `enabled: false`.
3. Send a request that matches that decision and completes normally.
4. Observe the cache/logs: the request path still creates a pending cache entry, and the response path still stores the response into semantic cache.

## Expected behavior

If a decision has no `semantic-cache` plugin, or explicitly sets `enabled: false`, the router should skip all semantic-cache work for that decision: no lookup, no pending request creation, and no response-side store. Global cache config should provide backend/defaults, not force storage for decisions that did not opt in.

## Affected layer

plugin

## Additional context

- Intended contract already exists in config helpers and tests:
  - `src/semantic-router/pkg/config/helper.go:422-435` makes `IsCacheEnabledForDecision` return `false` when a decision has no semantic-cache plugin, even if global cache is enabled.
  - `src/semantic-router/pkg/config/config_test.go:2241-2322` has matching tests for that behavior.
- Runtime currently diverges from that contract:
  - `src/semantic-router/pkg/extproc/req_filter_cache.go:57-66` gates lookup on `cacheEnabled`, but `src/semantic-router/pkg/extproc/req_filter_cache.go:138-141` still unconditionally calls `AddPendingRequest` after the lookup block.
  - `src/semantic-router/pkg/extproc/processor_res_body_pipeline.go:148-161` unconditionally calls `UpdateWithResponse` for non-streaming responses.
  - `src/semantic-router/pkg/extproc/processor_res_body_streaming.go:287-346` also stores reconstructed streaming responses without a decision-level enablement guard.
- Result: with global cache enabled, a decision-level plugin opt-out is ignored for store paths, so disabled/unconfigured decisions can still populate the cache.
- I did not find extproc coverage for the specific case `global enabled + decision plugin absent/disabled => no cache store`, so this likely needs a regression test alongside the fix.

Djanghao:
Closes #1546

## Summary

- Scope: semantic-cache plugin store paths (request pending + response caching)
- Primary skill: plugin
- Impacted surfaces: `req_filter_cache.go`, `processor_res_body_pipeline.go`, `processor_res_body_streaming.go`
- Conditional surfaces intentionally skipped: looper path (already correct)
- Debt entry: `none`

When global semantic cache is enabled, decisions without a `semantic-cache` plugin (or with it explicitly disabled) still had their responses written to cache. The config helper `IsCacheEnabledForDecision` already returns `false` for such decisions, but the runtime only applied that check to the lookup path — the store paths (`AddPendingRequest`, `UpdateWithResponse`, `AddEntry`) ran unconditionally.

### Changes

**Commit 1 — fix(plugin):**
- **`req_filter_cache.go`**: Gate `AddPendingRequest` in the normal request path with the same `cacheEnabled` guard already used for lookups (matching the looper path which was already correct)
- **`processor_res_body_pipeline.go`**: Add `IsCacheEnabledForDecision` guard to `updateResponseCache` for non-streaming responses
- **`processor_res_body_streaming.go`**: Add `IsCacheEnabledForDecision` guard to `cacheReconstructedStreamingResponse` for streaming responses
- **Tests**: 7 new regression tests covering decision-absent, decision-disabled, and decision-enabled scenarios for both streaming and non-streaming paths

**Commit 2 — refactor(plugin):**
- **`req_filter_cache.go`**: Extract `performCacheLookup` and `storePendingCacheRequest` from `handleCaching` to bring function length under the 100-line structure limit. Pure equivalent transformation with no behavior change.

## Validation

- Environment: `cpu-local`
- Fast gate: `make agent-ci-gate` — pass
- Feature gate: `go test -run "Cache" ./pkg/extproc/...` — 18/18 pass
- Local smoke / E2E: `not run`
- CI expectations / blockers: none expected

### DIFF
--- a/perf/cmd/perftest/main.go
+++ b/perf/cmd/perftest/main.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -14,6 +16,7 @@ import (
 func main() {
 	// Command-line flags
 	compareBaseline := flag.String("compare-baseline", "", "Path to baseline directory")
+	currentResults := flag.String("current", "", "Path to current benchmark results (JSON)")
 	thresholdFile := flag.String("threshold-file", "", "Path to thresholds configuration file")
 	outputPath := flag.String("output", "", "Output path for reports")
 	generateReport := flag.Bool("generate-report", false, "Generate performance report")
@@ -34,7 +37,7 @@ func main() {
 	}
 
 	if *compareBaseline != "" {
-		if err := compareWithBaseline(*compareBaseline, *thresholdFile, *outputPath); err != nil {
+		if err := compareWithBaseline(*compareBaseline, *currentResults, *thresholdFile, *outputPath); err != nil {
 			fmt.Fprintf(os.Stderr, "Error comparing with baseline: %v\n", err)
 			os.Exit(1)
 		}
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):
LGTM thanks

PR #1553, opened 2026-03-14
refactor: stabilize the router api schema

Xunzhuo:
Fixes: https://github.com/vllm-project/semantic-router/issues/1505

## Summary
- refactor the router onto a canonical v0.3 config contract with `version/listeners/providers/routing/global` as the steady-state schema
- split the routing-owned surface into `routing.modelCards`, `routing.signals`, and `routing.decisions`, while moving deployment access details into `providers.models[*].backend_refs`
- make the Go router the schema owner for canonical parse/export, built-in global defaults, and cross-surface validation
- migrate the Python CLI, dashboard, DSL/WASM, Helm, operator, and sample config tree to the same contract, including removal of `vllm-sr init` and legacy router-defaults steady-state flows
- reorganize `config/` and latest docs/tutorials around the new `signal/decision/algorithm/plugin/global` taxonomy, and add config-doc guards plus the v0.3 proposal to keep the public contract aligned

## Validation
- go test ./pkg/config/...
- go test ./pkg/dsl/...
- go test ./handlers/...
- python -m pytest tests/test_config_migrate.py tests/test_setup_bootstrap.py tests/test_config_template.py tests/test_plugin_parsing.py tests/test_latency_validation.py
- python -m pytest e2e/testing/vllm-sr-cli/test_unit_serve.py -k 'bootstraps_missing_config_file or serve_with_custom_config_path'
- make test-semantic-router
- helm lint deploy/helm/semantic-router
- make generate-crd
- make agent-validate
- npm --prefix website run build
- make agent-dev ENV=cpu
- make agent-serve-local ENV=cpu
- make agent-smoke-local

PR #1562, opened 2026-03-15
test(memory): add per-decision plugin E2E tests

### LINKED ISSUE #1295: [Memory] E2E tests for per-decision memory plugin
## Summary

Add E2E tests verifying that the memory plugin correctly respects per-decision configuration.

## Current State

Unit tests exist for per-decision config parsing, but no E2E tests verify the full flow.

## Task

Add E2E tests covering per-decision memory plugin behavior.

### Test Cases

1. **Decision WITH memory plugin → memory runs**
   - Configure decision A with memory plugin enabled
   - Send request matching decision A
   - Verify memories are retrieved and injected

2. **Decision WITHOUT memory plugin → memory skipped**
   - Configure decision B without memory plugin
   - Send request matching decision B
   - Verify no memory retrieval occurs

3. **Per-decision settings override global config**
   - Global: `similarity_threshold: 0.7`
   - Decision: `similarity_threshold: 0.5`
   - Verify decision-level threshold is used

4. **Memory plugin with other plugins**
   - Decision with both memory and system_prompt plugins
   - Verify both are applied correctly (memory appends, doesn't overwrite)

### Implementation Notes

- Add tests to `e2e/testing/09-memory-features-test.py` or create new file
- Use different decisions in test config
- May need to extend `config.memory-user.yaml` with multiple decisions

## Context

- **Priority**: P1 (High)
- **Effort**: Low
- **Research Required**: No
- **Dependencies**: None

## Related

- Tracking issue: #1293
- Original implementation: #1194

yehuditkerido:
## Summary

Add E2E tests verifying that the memory plugin correctly respects per-decision configuration overrides — specifically `memory.enabled=false` (skip retrieval) and per-decision `similarity_threshold` override.

## Changes

- **`config/testing/config.memory-user.yaml`** — Add 2 keyword signals (`NOMEM_MARKER`, `THRESHOLD_MARKER`) and 2 keyword-triggered decisions for deterministic test routing:
  * `no_memory_route` (priority 200): `memory.enabled: false` — skips retrieval even though global memory is enabled
  * `custom_threshold_route` (priority 150): `similarity_threshold: 0.99` — overrides the global default of 0.45
- **`e2e/testing/memory_tests/test_per_decision.py`** (new) — Add 2 test classes:
  * `PerDecisionMemoryDisabledTest`: stores a fact via `default_route`, queries via `no_memory_route` (memory disabled), verifies no injection, then confirms injection works via `default_route`
  * `PerDecisionThresholdOverrideTest`: stores a fact, queries via `custom_threshold_route` (threshold 0.99 filters memory), then confirms the same fact passes at the global threshold (0.45)
- **`e2e/testing/memory_tests/__init__.py`** — Add re-exports for the new test classes
- **`e2e/testing/09-memory-features-test.py`** — Register new test classes in the runner's `test_classes` list
- **`.github/workflows/integration-test-memory.yml`** — Add push trigger for `per-decision-memory-tests` branch (temporary, for CI during development)

## Testing

Tests use `llm-katan` echo backend. Each test stores a fact with a distinctive keyword (e.g., "Luna", "Max") absent from the query — its presence in the echo response can only come from injected memory context. Both tests follow a 4-step pattern: store → Milvus prerequisite check → negative test (override route) → positive control (default route).

Runs as part of `make memory-test-integration`.

Resolves #1295

### DIFF
--- a/e2e/config/config.memory-user.yaml
+++ b/e2e/config/config.memory-user.yaml
@@ -24,8 +24,61 @@ routing:
       - name: general
         description: General queries for memory testing
         mmlu_categories: [other]
-    keywords: []
+    keywords:
+      - name: no_memory_trigger
+        operator: OR
+        keywords: [NOMEM_MARKER]
+      - name: custom_threshold_trigger
+        operator: OR
+        keywords: [THRESHOLD_MARKER]
   decisions:
+    # Triggered by NOMEM_MARKER keyword. Memory explicitly disabled so retrieval
+    # is skipped even though global memory.enabled=true.
+    - name: no_memory_route
+      description: Route with memory explicitly disabled for per-decision testing
+      priority: 200
+      rules:
+        operator: OR
+        conditions:
+          - type: keyword
+            name: no_memory_trigger
+      modelRefs:
+        - model: qwen3
+          use_reasoning: false
+      plugins:
+        - type: system_prompt
+          configuration:
+            system_prompt: You are a helpful assistant. Memory access is disabled for this route.
+            mode: insert
+        - type: memory
+          configuration:
+            enabled: false
+
+    # Triggered by THRESHOLD_MARKER keyword. Memory enabled with similarity_threshold=0.99
+    # (near-impossible to match), overriding the global default of 0.45.
+    - name: custom_threshold_route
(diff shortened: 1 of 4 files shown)


Xunzhuo:

PR #1532, opened 2026-03-12
feat(domain-signal): add UseModernBERT factory branch for category classifier

noalimoy:
## Summary

Follow-up to #1380 (domain signal accuracy improvements). Closes #1526.

The `NewClassifier()` factory ignored `use_modernbert: true` on `category_model` — only `UseMmBERT32K` was checked. The `else` branch always fell through to auto-detect, which picks CandleBERT first. Because CandleBERT init succeeds (model has `lora_config.json`), ModernBERT was never initialized, causing `ClassifyWithProbabilities()` to always fail (no live ModernBERT backend).

### Changes

- Add `ModernBERTCategoryInitializerImpl` — calls `InitModernBertClassifier` directly, bypassing CandleBERT auto-detection
- Add `ModernBERTCategoryInferenceImpl` — delegates `Classify` and `ClassifyWithProbabilities` to ModernBERT FFI directly
- Add `else if cfg.CategoryModel.UseModernBERT` branch in the factory, between `UseMmBERT32K` and auto-detect

Follows the same structure as the existing `UseMmBERT32K` path. Backward compatible — when `UseModernBERT` is false (default), the auto-detect path is taken as before.

### Context

This is one of two blockers preventing the entropy-based multi-category matching from working end-to-end. The other blocker (Rust FFI returns fake probability distributions instead of real softmax — #1528) is tracked separately.

## Test plan

The change is additive (new types + new factory branch). No existing code was modified. Verified no regressions in the changed package and related packages:

| Package | Test suite | Why relevant | Result |
|---------|-----------|--------------|--------|
| `classification` | Ginkgo suite (113 specs) | Changed package — factory lives here | 113/113 passed |
| `config` | Ginkgo suite (280 specs) | `CategoryModel.UseModernBERT` config parsing | 280/280 passed |
| `extproc` | Ginkgo suite (205 specs) | Creates classifiers from config end-to-end | 205/205 passed |
| `decision` | `TestDecisionEngine_EvaluateDecisions` | Consumes domain signal output | passed |
| `candle-binding` | `TestBERTClassifiers` + concurrent safety | Exercises the FFI init + classify path | passed |

Issue #1528, opened 2026-03-12
[Feat][Candle-binding]: pass real softmax probabilities from ModernBERT through FFI layer


noalimoy:
### Describe the bug

Follow-up to #1380 (domain signal accuracy improvements).

`classify_modernbert_text_with_probabilities` in the Rust FFI layer returns a **fabricated** probability distribution instead of the real softmax output from the model.

The underlying `classify_text()` in `modernbert.rs` already computes the full softmax vector across all classes, but only returns the top-1 result `(predicted_class, max_prob)`:

```rust
// modernbert.rs — classify_text()
let probabilities = self.classifier.forward(&classifier_input)?;
let probabilities_vec = probabilities.squeeze(0)?.to_vec1::<f32>()?;
// ↑ real softmax: [0.43, 0.41, 0.02, 0.08, ...] — sum = 1.0

// ... finds argmax ...
Ok((predicted_class, max_prob))  // probabilities_vec is dropped
```

The FFI function then **fabricates** a fake distribution to fill the probabilities field:

```rust
// classify.rs — classify_modernbert_text_with_probabilities()
match classifier.classify_text(text) {
    Ok((class_id, confidence)) => {
        let num_classes = classifier.get_num_classes();
        let mut probabilities = vec![0.1f32; num_classes];  // all 0.1
        probabilities[class_id as usize] = confidence;       // winner gets real value
        // sum > 1.0 — invalid distribution
    }
}
```

This causes the Go-side entropy analysis to receive garbage input (normalized entropy > 1.0), making multi-domain classification unreliable.

### Proposed fix

**1. Add `classify_text_with_probabilities` to `TraditionalModernBertClassifier`** (`modernbert.rs`):

A new method that reuses the same forward-pass logic but returns the full probability vector alongside the top prediction:

```rust
pub fn classify_text_with_probabilities(&self, text: &str)
    -> Result<(usize, f32, Vec<f32>), candle_core::Error>
```

**2. Update FFI function** (`classify.rs`):

Change `classify_modernbert_text_with_probabilities` to call the new method and pass the real softmax vector through to Go — no fabrication needed.

### Impact

- Go-side entropy analysis will receive a valid probability distribution (sum = 1.0)
- Multi-domain queries (e.g., "economic impacts of healthcare reform") will correctly identify overlapping domains
- No changes needed on the Go side — the existing code already handles real probabilities correctly

### Context

[...]

Issue #1380, opened 2026-02-25
[v0.2-Athena]: Improve domain signal accuracy


noalimoy:
### Describe the feature

Enhance the domain signal classifier to:

- Pass **real confidence scores** from the ML model to `SignalConfidences` (e.g., `"domain:math" → 0.87`) instead of the current hardcoded `1.0`, enabling the `"confidence" strategy` to make meaningful quality-based routing decisions

- Use `ClassifyWithProbabilities()` to return **all matched domain categories** to the Decision Engine (not just the single top-1), enabling `AND` conditions across multiple domain signals (e.g., `domain:"economics" AND domain:"health"`)

noalimoy:
Hi @rootfs,

As part of [Improve signal computation accuracy #984 ](url) -  I'm working on improving this domain signal accuracy.

While analyzing the domain signal flow, I noticed that the `CategoryInference` interface already defines `ClassifyWithProbabilities()` which returns the full probability distribution across all 14 MMLU categories. 
However, the domain signal evaluation in `EvaluateAllSignalsWithContext` ([classifier.go:1242](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/classification/classifier.go#L1242)) only calls `Classify()`, which returns a single top-1 result:

```go
result, err := c.categoryInference.Classify(text)
// ...
results.MatchedDomainRules = append(results.MatchedDomainRules, categoryName) // always 0 or 1 domains
```

This means that queries spanning multiple domains 
(e.g., *"What are the economic impacts of healthcare reform?"* → **economics** 0.43 + **health** 0.41)
will only ever return the single top category. 

As a result, **AND conditions with two or more domain types can never match**, since `MatchedDomainRules` will always contain at most one entry.

**Questions:**

1. Was the single-label (top-1 only) behavior for domain signals an intentional design choice, or just an initial implementation that hasn't been revisited ?

2. Would you be open to switching domain signal evaluation to `ClassifyWithProbabilities()`, emitting all categories that meet the configured threshold into `MatchedDomainRules`? This would align domain signals with how keyword and embedding signals already work (returning all matches), and enable multi-domain AND decisions ?

noalimoy:
Hi @Xunzhuo / @rootfs 

This issue was closed, but I believe it should remain open — #1380 and #1485 address **complementary layers** of the domain signal, not overlapping ones.

[...]

PR #1572, opened 2026-03-16
feat(domain-signal): add UseModernBERT factory branch for category classifier

noalimoy:
## Summary

Re-applies #1532 (reverted in #1561). 
Follow-up to #1380.

The `NewClassifier()` factory ignored `use_modernbert: true` on `category_model` — only `UseMmBERT32K` was checked. The `else` branch always fell through to auto-detect, which picks CandleBERT first. Because CandleBERT init succeeds (model has `lora_config.json`), ModernBERT was never initialized, causing `ClassifyWithProbabilities()` to always fail (no live ModernBERT backend).

The original PR was reverted because E2E profiles had `use_modernbert: true` set for `mom-domain-classifier`, which is a standard LoRA-BERT — not a ModernBERT architecture. With the factory now honoring the flag, initialization correctly attempted ModernBERT loading, which failed because the model's `config.json` lacks ModernBERT-specific fields. This misconfiguration was previously hidden because the old factory ignored the flag entirely.

### Changes

- Add `ModernBERTCategoryInitializerImpl` — calls `InitModernBertClassifier` directly, bypassing CandleBERT auto-detection
- Add `ModernBERTCategoryInferenceImpl` — delegates `Classify` and `ClassifyWithProbabilities` to ModernBERT FFI directly
- Add `else if cfg.CategoryModel.UseModernBERT` branch in the factory, between `UseMmBERT32K` and auto-detect
- Fix `use_modernbert: true` → `false` in E2E profiles (ai-gateway, aibrix, rag-hybrid-search, routing-strategies) where the model is LoRA-BERT, not ModernBERT

Follows the same structure as the existing `UseMmBERT32K` path. Backward compatible — when `UseModernBERT` is false (default), the auto-detect path is taken as before.

### Unit tests (local)

| Package | Test suite | Why relevant | Result |
|---------|-----------|--------------|--------|
| `classification` | Ginkgo suite (113 specs) | Changed package — factory lives here | 113/113 passed |
| `config` | Ginkgo suite (280 specs) | `CategoryModel.UseModernBERT` config parsing | 280/280 passed |
| `extproc` | Ginkgo suite (205 specs) | Creates classifiers from config end-to-end | 205/205 passed |
| `decision` | `TestDecisionEngine_EvaluateDecisions` | Consumes domain signal output | passed |
| `candle-binding` | `TestBERTClassifiers` + concurrent safety | Exercises the FFI init + classify path | passed |

### E2E tests (local)

| Profile | Result | Notes |
|---------|--------|-------|
| `ai-gateway` | ✅ passed | Was failing with `missing field global_attn_every_n_layers` — fixed by `use_modernbert: false` |

E2E profiles corrected to use `use_modernbert: false` for `mom-domain-classifier` (LoRA-BERT, not ModernBERT) — this was the root cause of the #1561 revert.

### Context

This is one of two blockers preventing the entropy-based multi-category matching from working end-to-end. The other blocker (Rust FFI returns fake probability distributions instead of real softmax — #1528) is tracked separately.
ifiers` + concurrent safety | Exercises the FFI init + classify path | passed |

### DIFF
--- a/.github/workflows/ci-changes.yml
+++ b/.github/workflows/ci-changes.yml
@@ -31,6 +31,8 @@ on:
         value: ${{ jobs.filter.outputs.e2e_ai_gateway }}
       e2e_aibrix:
         value: ${{ jobs.filter.outputs.e2e_aibrix }}
+      e2e_dashboard:
+        value: ${{ jobs.filter.outputs.e2e_dashboard }}
       e2e_dynamic_config:
         value: ${{ jobs.filter.outputs.e2e_dynamic_config }}
       e2e_llm_d:
@@ -75,6 +77,7 @@ jobs:
       e2e_istio: ${{ steps.changes.outputs.e2e_istio }}
       e2e_ai_gateway: ${{ steps.changes.outputs.e2e_ai_gateway }}
       e2e_aibrix: ${{ steps.changes.outputs.e2e_aibrix }}
+      e2e_dashboard: ${{ steps.changes.outputs.e2e_dashboard }}
       e2e_dynamic_config: ${{ steps.changes.outputs.e2e_dynamic_config }}
       e2e_llm_d: ${{ steps.changes.outputs.e2e_llm_d }}
       e2e_routing_strategies: ${{ steps.changes.outputs.e2e_routing_strategies }}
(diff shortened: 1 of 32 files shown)


Xunzhuo:

PR #1573, opened 2026-03-16
[Bugfix][Router] Return consistent extproc request validation errors

### LINKED ISSUE #1556: feature: return proper HTTP error codes for client errors instead of blanket 503
## Summary

All client errors currently surface as HTTP 503 "no healthy upstream" because the ext_proc filter does not intercept invalid requests before they reach Envoy's routing stage. This makes VSR difficult to debug in production and triggers infinite retry loops in OpenAI-compatible SDKs (which treat 503 as a transient server error).

## Current Behavior

| Client Error | Expected Code | Actual Code |
|---|---|---|
| Malformed JSON body | 400 | 503 |
| Missing `messages` field | 400 | 503 |
| Wrong HTTP method (GET) | 405 | 503 |
| Non-existent endpoint | 404 | 503 |
| Invalid `model` value | 400/404 | 503 |

All errors return `503 Service Unavailable` with no error body — identical to an actual backend outage.

## Impact

- **Retry storms:** OpenAI Python/JS SDKs auto-retry on 503 with exponential backoff, turning a single bad request into 3-6 retries before failing. At scale, this amplifies load.
- **Debugging impossible:** Operators cannot distinguish client errors from real outages in monitoring/alerting.
- **Poor developer experience:** Users integrating with VSR get no feedback on what is wrong with their request.

## Proposed Solution

Use Envoy ext_proc's `ImmediateResponse` mechanism to short-circuit invalid requests during the request header/body processing phase, before they reach the routing stage:

1. During `processRequestBody`, validate the JSON payload (required fields, correct types)
2. If invalid, return an `ImmediateResponse` with the appropriate HTTP status code and an OpenAI-compatible error body:
```json
{
  "error": {
    "message": "messages field is required",
    "type": "invalid_request_error",
    "code": "missing_required_field"
  }
}
```
3. This matches the [OpenAI error format](https://platform.openai.com/docs/guides/error-codes) that SDKs already handle correctly — no retry on 400.

## Context

This is an architectural consequence of the ext_proc pattern — the router processes requests as a filter but does not control the HTTP response directly. The `ImmediateResponse` API in ext_proc was designed exactly for this case but is not currently used for validation errors.

mildred522:
Closes #1556

## Summary

- Scope: tighten extproc request validation so malformed request headers/bodies return consistent request-side errors
instead of leaking inconsistent downstream behavior.
- Primary skill: router/extproc request validation
- Impacted surfaces: `src/semantic-router/pkg/extproc`
- Conditional surfaces intentionally skipped: none in the extproc request validation path; remaining ai-gateway model/
profile failures appear outside this change scope
- Debt entry: `none`

## What Changed

- added focused request validation helpers for extproc header/body handling
- normalized request-side error classification so malformed inputs return stable immediate responses instead of
falling through inconsistent internal error paths
- completed the ImmediateResponse status-code mapping so client-facing extproc errors preserve the intended Envoy HTTP
status enum
- updated extproc routing/request tests to cover the new validation behavior and keep error handling consistent across
header/body paths

Behavior details:
- malformed JSON body -> `400 Bad Request`
- missing `messages` on chat completions requests -> `400 Bad Request`
- non-array `messages` on chat completions requests -> `400 Bad Request`
- model not configured in the router -> `404 Not Found`

## Validation

- Environment: `cpu-local` (Debian WSL + native Docker/kind)
- Fast gate:
  - `go test -v ./pkg/extproc/... -count=1`
  - `make test-semantic-router`
  - `make check-go-mod-tidy`
  - `make go-lint`
  - `make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/extproc/extproc_test.go,src/semantic-router/pkg/
extproc/processor_req_body.go,src/semantic-router/pkg/extproc/processor_req_body_routing.go,src/semantic-router/pkg/
extproc/processor_req_body_test.go,src/semantic-router/pkg/extproc/processor_req_header.go,src/semantic-router/pkg/
extproc/processor_req_header_test.go,src/semantic-router/pkg/extproc/router.go,src/semantic-router/pkg/extproc/
processor_req_body_validation.go,src/semantic-router/pkg/extproc/processor_req_header_validation.go,src/semantic-
router/pkg/extproc/status_code_mapping.go,src/semantic-router/pkg/extproc/utils_test.go"`
- Feature gate: `not run`
- Local smoke / E2E:
  - `make e2e-setup E2E_PROFILE=ai-gateway E2E_VERBOSE=true E2E_CLUSTER_NAME=semantic-router-e2e` passed
  - `make e2e-test-only E2E_PROFILE=ai-gateway E2E_VERBOSE=true E2E_CLUSTER_NAME=semantic-router-e2e` completed but
failed in the existing ai-gateway profile/model-registration path with responses such as `model "general-expert" is
not configured in the router`
- CI expectations / blockers:
  - no local blocker found in the extproc request validation code path
  - expect CI kubernetes profile coverage to exercise the remaining model-integrated surfaces
  - current local ai-gateway test-only failures appear outside the #1556 change scope and were reproducible after
setup succeeded

## Checklist

- [x] PR title uses the repo prefix format: `[Bugfix][Router]`

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_req_body_test.go
+++ b/src/semantic-router/pkg/extproc/processor_req_body_test.go
@@ -9,6 +9,7 @@ import (
 	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/authz"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
 )
 
 func TestMaybeForceImageGenerationModalitySetsBothWhenUserContentPresent(t *testing.T) {


Xunzhuo:
@mildred522 thanks for the contributes, can you fix the CI?

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -30,6 +30,9 @@ func (r *OpenAIRouter) selectEndpointForModel(ctx *RequestContext, model string)
 	if err != nil {
 		return "", "", fmt.Errorf("endpoint resolution for model %q: %w", model, err)
 	}
+	if !endpointFound {
```
Xunzhuo: we should ignore the err and let it pass, in k8s we dont manage the endpoint, so it is not a required input

src/semantic-router/pkg/extproc/status_code_mapping.go
```diff
@@ -0,0 +1,27 @@
+package extproc
+
+import typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
+
+func statusCodeToImmediateResponseCode(statusCode int) typev3.StatusCode {
+	switch statusCode {
+	case 200, 400, 404, 500:
```
Xunzhuo: can you set the enum for the magic numbers?

src/semantic-router/pkg/extproc/processor_req_body_routing.go
mildred522: thanks, that makes sense. I’ll remove this request-side error path and keep the existing behavior here

mildred522:
@Xunzhuo I looked into the remaining E2E failure and it seems separate from #1556. In the authz-rbac profile, \chat-
completions-request sends no authz identity headers, but the profile enables `role_bindings`, so it fails with the
expected 403 for missing user identity. My current thought is the smallest fix would be to use an authz-rbac-specific
chat smoke test that sets the required authz headers, instead of the shared generic one. Does that sound like the
right direction for this pr, or would you prefer this to be handled separately?

Xunzhuo:
hey @mildred522 thanks for you findings, yes it is not relevant to this PR, so let us solve it separately

Xunzhuo (approved):

PR #1589, opened 2026-03-18
chore(precommit): update precommit container

rootfs:
The precommit container image is out of sync with the CI. This is to keep it update to date.

@yuluo-yx @samzong


samzong:
thanks. I will check it later.

Issue #1508, opened 2026-03-12
feature: split vllm-sr into a router-only runtime and external service images


Xunzhuo:
## Describe the feature
Refactor the runtime packaging so `vllm-sr` becomes a router-only image while dashboard, database, and Envoy are shipped and managed as separate services/images.

## Primary layer
global level

## Why this layer?
This changes the deployment topology, image responsibilities, and inter-service contracts rather than adding a feature inside one router layer.

## Why do you need this feature?
The current packaging mixes responsibilities that should be independently versioned and operated in production. `v0.3` should simplify the router image and make the supporting services explicit. Relevant surfaces include `src/vllm-sr/Dockerfile`, `src/vllm-sr/Dockerfile.rocm`, the CLI deploy runtime, dashboard backend/frontend services, and the Envoy sidecar/stack wiring.

## Additional context
- Acceptance should include a router-only runtime image, documented service boundaries, and verified communication between router, dashboard, database, and Envoy.
- Dashboard functionality that currently assumes the old all-in-one layout should continue to work after the split.
- Validation should include end-to-end deployment checks in the new topology.

FAUST-BENCHOU:
| Requirement | Status |
|---|---|
| Router-only runtime image | **Completed** — `src/vllm-sr/Dockerfile` and `Dockerfile.rocm` package only the router and bindings, use `start-router.sh` as the entrypoint, and do not include Dashboard or Envoy. |
| Dashboard, Envoy, and database run as separate services/images | **Completed** — The CLI uses the split topology by default and starts router, Envoy, and Dashboard as separate containers; Redis/Postgres are provisioned independently by `docker_services.py` when needed. |
| Documented service boundaries | **Partially completed** — `docs/agent/environments.md` and `dashboard/README.md` describe the boundaries, but the website installation docs do not yet clearly separate each service’s role and responsibility. |
| Router, Dashboard, DB, and Envoy connectivity verified | **Partially completed** — `agent-smoke-local` and `test_integration.py` verify router/Envoy/Dashboard connectivity; DB provisioning is covered separately, but there is no explicit E2E acceptance test covering all four components together. |
| Dashboard remains usable after the topology split | **Completed** — `runtime_managed_services.go`, status/log handling, and related logic have been adapted for split containers, with corresponding tests. |
[...]

PR #1655, opened 2026-03-25
feat: topology deploy support

haowu1234:
## Summary

This PR makes the split local runtime topology the default for `vllm-sr serve` and local `make vllm-sr-dev` workflows.

Instead of starting the legacy single-container local runtime by default, local development now starts dedicated `router`, `envoy`, and `dashboard` containers under the standard local serve path.

This rollout is intentionally scoped to local runtime and local developer workflows. It does not migrate Helm, Kubernetes, or operator deployment topologies to the split model.

Importantly, this change does **not** introduce a new published router/Envoy image surface:

- `vllm-sr` / `vllm-sr-rocm` continue to back the router runtime by default
- Envoy continues to use the upstream Envoy image
- `dashboard` remains on the existing dashboard image family

### Make split local runtime the default

- switch the default local runtime topology from legacy to split
- remove the need for users to opt into split topology during normal local development
- keep local startup on the canonical `vllm-sr serve` path rather than introducing a second serve flow

### Preserve the local runtime image and packaging surface

- keep using the existing `vllm-sr` / `vllm-sr-rocm` images for the router runtime by default
- keep using the upstream Envoy image
- keep using the existing dashboard image family
- avoid adding a new published router/envoy image family as part of this rollout

### Preserve dashboard/setup/config behavior under split topology

- support dashboard-first setup mode with split managed services
- preserve runtime config sync and router hot reload behavior
- regenerate and reload split Envoy config after dashboard-managed config updates
- keep local config apply / setup activation behavior working under split topology

### Preserve legacy-to-split upgrade compatibility

- keep `status`, `logs`, and `stop` compatible with legacy `vllm-sr-container` residue during transition
- clean up legacy runtime container residue before starting the default split runtime
- preserve the shared `.vllm-sr` state root and dashboard data flow so existing local workspaces continue to function without topology-specific migration steps

### Harden split local runtime behavior discovered during rollout

- fix split runtime source/runtime config path handling
- fix split Envoy regeneration to use the dashboard runtime Python environment
- improve split runtime managed-service lifecycle, status, and logs behavior
- align OpenClaw/shared-network behavior with split local runtime
- ensure required runtime dependencies and assets are available in split image paths
- make dashboard Docker builds more resilient to transient Alpine package index failures

### Update local-dev, smoke, and CI support

- add focused split runtime topology and stack tests
- add lightweight legacy-to-default-split compatibility smoke coverage in the CLI harness
- extend local smoke coverage for CPU and AMD split flows
- update Make/docs/help text to reflect split as the default local runtime
- keep local dev conservative by default: the router runtime image is still rebuilt unless the user explicitly chooses to reuse an already up-to-date local image with `SKIP_COMPAT_IMAGE=1`

## Why

The goal is to make local topology separation the standard developer path while preserving compatibility for existing local users and keeping the rollout constrained to local runtime behavior.

This PR is intentionally optimized for:

- minimal disruption to existing users
- safer local rollout of split runtime behavior
- preserving the current user-facing local commands
- compatibility with older all-in-one local residue during upgrade
- avoiding unnecessary expansion of the published image surface

## Validation

Validated through repo-native local-dev, dashboard/runtime, and CLI compatibility checks, including:

- `make dashboard-check`
- `make agent-validate`
- diff-scoped `make agent-lint`
- focused `pytest` coverage for split runtime topology, local image selection, Dockerfile surfaces, and legacy-to-split compatibility
- focused Go tests for dashboard split-runtime apply/sync behavior
- local dashboard image build
- local split-topology smoke and regression checks during rollout

## Usage

Default local runtime:

```bash
make vllm-sr-dev
vllm-sr serve --image-pull-policy never
```

Default local runtime on AMD:

```bash
make vllm-sr-dev VLLM_SR_PLATFORM=amd
vllm-sr serve --image-pull-policy never --platform amd
```

If the local router runtime image is already up to date and you want to reuse it during a debug loop:

```bash
make vllm-sr-dev SKIP_COMPAT_IMAGE=1
make vllm-sr-dev VLLM_SR_PLATFORM=amd SKIP_COMPAT_IMAGE=1
```

Useful local runtime commands:

```bash
vllm-sr status all
vllm-sr logs router -f
vllm-sr logs envoy -f
vllm-sr logs dashboard -f
vllm-sr stop
```

## Notes

This PR is intentionally scoped to local runtime and local developer workflow behavior.

It does **not** migrate Helm, Kubernetes, or operator deployments to the split model, and it does **not** add a new published router/Envoy image family as part of this rollout.

Related to #1508

### DIFF
--- a/e2e/testing/run_memory_integration.sh
+++ b/e2e/testing/run_memory_integration.sh
@@ -9,6 +9,12 @@ CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 DOCKER_REGISTRY="${DOCKER_REGISTRY:-ghcr.io/vllm-project/semantic-router}"
 DOCKER_TAG="${DOCKER_TAG:-latest}"
 VLLM_SR_IMAGE="${VLLM_SR_IMAGE:-ghcr.io/vllm-project/semantic-router/vllm-sr:latest}"
+VLLM_SR_STACK_NAME="${VLLM_SR_STACK_NAME:-vllm-sr}"
+if [[ "${VLLM_SR_STACK_NAME}" == "vllm-sr" ]]; then
+    VLLM_SR_NETWORK="${VLLM_SR_NETWORK:-vllm-sr-network}"
+else
+    VLLM_SR_NETWORK="${VLLM_SR_NETWORK:-${VLLM_SR_STACK_NAME}-vllm-sr-network}"
+fi
 
 TEST_DIR="${MEMORY_TEST_DIR:-$(mktemp -d -t vsr-memory-test-XXXXXX)}"
 PID_FILE="${TEST_DIR}/serve.pid"
@@ -87,8 +93,16 @@ python3 -c "from huggingface_hub import snapshot_download; snapshot_download('se
 
 make -C "${REPO_ROOT}" start-milvus
 cp "${REPO_ROOT}/e2e/config/config.memory-user.yaml" "${CONFIG_FILE}"
+python3 -c 'from pathlib import Path; path = Path("'"${CONFIG_FILE}"'"); path.write_text(path.read_text().replace("host.docker.internal:8000", "llm-katan:8000"))'
 
-"${CONTAINER_RUNTIME}" run -d --name llm-katan --network host \
+if ! "${CONTAINER_RUNTIME}" network inspect "${VLLM_SR_NETWORK}" >/dev/null 2>&1; then
+    "${CONTAINER_RUNTIME}" network create "${VLLM_SR_NETWORK}" >/dev/null
+fi
+
+"${CONTAINER_RUNTIME}" run -d --name llm-katan \
+    --network "${VLLM_SR_NETWORK}" \
+    --network-alias llm-katan \
+    -p 8000:8000 \
     "${DOCKER_REGISTRY}/llm-katan:${DOCKER_TAG}" \
     llm-katan --model dummy --host 0.0.0.0 --port 8000 --served-model-name qwen3 --backend echo >/dev/null
 


.github/workflows/docker-publish.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       packages: write
     strategy:
       matrix:
-        image: [dashboard, extproc, extproc-rocm, llm-katan, vllm-sr, vllm-sr-rocm, vllm-sr-sim]
+        image: [dashboard, extproc, extproc-rocm, llm-katan, vllm-sr, vllm-sr-envoy, vllm-sr-rocm, vllm-sr-router, vllm-sr-router-rocm, vllm-sr-sim]
```
Xunzhuo: we shouldnt increase the images i think, we already have vllm-sr(now: router only), and dashboard image, and envoy has its own images and we shouldnt maintain

haowu1234: > we shouldnt increase the images i think, we already have vllm-sr(now: router only), and dashboard image, and envoy has its own images and we shouldnt maintain

ok

haowu1234: Agreed. I’ve already updated the branch accordingly.

This PR no longer expands the published image surface:

- `vllm-sr` / `vllm-sr-rocm` continue to serve as the router compatibility images
- `dashboard` remains the dashboard image
- Envoy continues to use the upstream Envoy image
- no new published `router` / `router-rocm` / `envoy` image family is introduced

So the split work is now scoped to local topology/runtime wiring only, without adding new maintained image variants.

Xunzhuo: great

PR #1669, opened 2026-03-26
feat(tools): add DSL tuning framework with analytical trace diagnosis and three scenario plugins

rootfs:
## Summary

- Scope: DSL
- Primary skill: analytical models with DSL tuning
- Impacted surfaces: DSL, and server
- Conditional surfaces intentionally skipped:
- Behavior-visible change: `yes` / `no`
- Debt entry: `none` / `TDxxx` none

## Validation

- Environment: `cpu-local` / `amd-local` / `not run`
- Fast gate:
- Feature gate:
- Local smoke / E2E:
- CI expectations / blockers:

## Checklist

### DIFF
--- a/e2e/profiles/dashboard/dashboard-deployment.yaml
+++ b/e2e/profiles/dashboard/dashboard-deployment.yaml
@@ -68,6 +68,11 @@ spec:
             # available in CI and the pipeline is not part of the dashboard contract.
             - name: ML_PIPELINE_ENABLED
               value: "false"
+            # Disable auth in E2E by pointing the auth DB to an impossible path.
+            # /dev/null is a char device so MkdirAll("/dev/null") fails, causing
+            # setupAuthRoutes to return nil and wrapWithAuth to skip the middleware.
+            - name: DASHBOARD_AUTH_DB_PATH
+              value: /dev/null/auth.db
           ports:
             - name: http
               containerPort: 8700


src/semantic-router/pkg/decision/trace.go
```diff
@@ -0,0 +1,211 @@
...
+type DecisionTrace struct {
+	DecisionName string     `json:"decision_name"`
+	Matched      bool       `json:"matched"`
+	Confidence   float64    `json:"confidence"`
+	RootTrace    *TraceNode `json:"root_trace"`
+}
+
+// EvaluateDecisionsWithTrace evaluates all decisions and returns both the
+// best match and a trace of every decision's evaluation tree.
+func (e *DecisionEngine) EvaluateDecisionsWithTrace(
```
Xunzhuo, line 32: i didnt find any place to use this func, am i missing anything?

deploy/recipes/calibration/knowledge_bases/labels.json
```diff
@@ -0,0 +1,286 @@
+{
```
Xunzhuo, line 1: looks like we only need this now, and other files can be deleted?

deploy/recipes/privacy/knowledge_bases/labels.json
```diff
@@ -0,0 +1,452 @@
+{
```
Xunzhuo, line 1: ditto

rootfs:
e2e test failed on disk space on github runner.

Xunzhuo:
A few scope / cleanup items look worth addressing before this lands:

- `deploy/recipes/calibration/knowledge_bases/taxonomy.json` and `deploy/recipes/privacy/knowledge_bases/taxonomy.json` look unused. The current KB runtime reads `source.manifest` / `labels.json`, and the legacy `taxonomy_path` path is explicitly rejected. If the intent here is to "keep only `labels.json`", these sidecars should be removed from the PR.
- `tools/agent/scripts/tuning/verify_results/verify_calibration.json` and `tools/agent/scripts/tuning/verify_results/verify_privacy.json` also look like stale artifacts. I could not find code or docs consuming them; the README documents `*_verification.json` instead.
- The `confidence` scenario looks half-wired right now. The README mentions it, but `tools/agent/scripts/tuning/cli.py` only registers `privacy` and `calibration`, and `run_confidence_verification.py` depends on an external `/data/...` path. I would either wire this scenario through end-to-end in this PR or split it into a follow-up.
- Exposing trace on the existing diagnostic `POST /api/v1/eval` path makes sense to me (`?trace=true` is not redundant with `classify/intent`), but the current contract looks broken: the server returns top-level `eval_trace`, while the tuning client / calibration scenario read `decision_result.eval_trace`. The same path may also need response-shape alignment for fields like `projection_scores` / `projection_bands` vs the current eval payload.

Net: I think the PR gets cleaner if it narrows to the pieces that are fully wired today, and either removes or splits the unused artifacts / partially integrated scenario bits.

rootfs:
e2e failure appears to be related to docker hub issue, not the code change.

PR #1672, opened 2026-03-27
Fix explicit regex keyword rules to compile as real regex

tristaZero:
## Summary
- treat keyword rules with explicit `method: regex` as true regex patterns
- keep default/empty method behavior unchanged (escaped literal + boundary handling)
- add regression tests for both behaviors

## Why
`method: regex` currently escapes configured patterns, so real regex rules like `AKIA[0-9A-Z]{16}` are treated as literals and fail to match.

## Changes
- `prepRegexRule` now passes an explicit-regex flag into pattern building
- `regexPatterns(...)` compiles explicit regex rules directly (`pattern`, `(?i)pattern`)
- existing literal-safe behavior remains for non-explicit regex rules

## Tests
Added in `keyword_classifier_structural_test.go`:
- explicit regex method matches AWS key pattern
- default method still treats metacharacters literally (e.g. `1.0`)

## Notes
I could not run the full local test package because this environment is missing native Rust libs (`candle_semantic_router`, `nlp_binding`) required for linking classification tests.

### DIFF
--- a/src/semantic-router/pkg/config/kb_config.go
+++ b/src/semantic-router/pkg/config/kb_config.go
@@ -52,15 +52,7 @@ func (s KnowledgeBaseSource) ResolvePath(baseDir string) string {
 		return filepath.Clean(s.Path)
 	}
 
-	candidates := make([]string, 0, 4)
-	if baseDir != "" {
-		candidates = append(candidates, filepath.Join(baseDir, s.Path))
-	}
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Thanks @tristaZero, can you sign the commit?

tristaZero:
Hello @Xunzhuo sure, done with my signoff.

Issue #1608, opened 2026-03-19
feature: make router replay, response history, vector-store metadata, and startup status restart-safe


Xunzhuo:
## Describe the feature

Make the router-side state surfaces that already look product-facing restart-safe and explicitly durable where appropriate, instead of leaving them on process memory or temp-file conventions.

## Primary layer

global level

## Why this layer?

This spans router defaults, extproc runtime storage, vector-store metadata, file registries, startup status, and dashboard-facing operator APIs. The problem is intentionally cross-cutting rather than belonging to one signal, plugin, or single package.

## Why do you need this feature?

The current router runtime still mixes durable-looking features with process-local ownership:
- `global.services.response_api.store_backend` defaults to `memory`
- `global.services.router_replay.store_backend` defaults to `memory`
- `src/semantic-router/pkg/vectorstore/manager.go` keeps vector-store metadata in an in-memory registry even when backend collections persist elsewhere
- `src/semantic-router/pkg/vectorstore/filestore.go` keeps file metadata in memory while file bytes live on disk
- `src/semantic-router/pkg/startupstatus/status.go` falls back to temp-owned JSON for runtime readiness and model download progress

That means restart behavior is inconsistent today: a router can retain some bytes or backend collections but still lose the metadata and control-plane truth required to recover cleanly.

## Additional context

Child of #1606.

Repository evidence:
- `docs/agent/tech-debt/td-034-runtime-and-dashboard-state-durability-and-telemetry-contract.md`
- `docs/agent/state-taxonomy-and-inventory.md`
- `src/semantic-router/pkg/responsestore/{factory.go,memory_store.go}`
- `src/semantic-router/pkg/routerreplay/store/{factory.go,memory.go}`
- `src/semantic-router/pkg/extproc/router_storage.go`
- `src/semantic-router/pkg/extproc/router_replay_setup.go`
- `src/semantic-router/pkg/vectorstore/{manager.go,filestore.go}`
- `src/semantic-router/pkg/startupstatus/status.go`

Related issues to keep aligned, not duplicated:
- #1601 shared Milvus lifecycle seam
- #1516 production RAG and memory usability

Suggested acceptance:
- define the intended durability contract for Response API storage, router replay, vector-store metadata, file metadata, and startup status
- make router-visible metadata restart-safe without requiring process memory to remain intact
- keep cache-like surfaces as caches; do not force semantic cache or RAG cache into relational storage when a shared cache is the correct abstraction
[...]

PR #1683, opened 2026-03-29
feat(routerreplay): default store_backend to postgres for durable replay

yehuditkerido:
## Summary

Router Replay records were lost on every restart because `store_backend` defaulted to `memory`. Change the default to `postgres` for SQL-queryable audit storage, warn on `memory`, and add an E2E restart-recovery test.

## Changes

- Default `RouterReplayConfig.StoreBackend` from `"memory"` to `"postgres"` in `canonical_defaults.go`
- Emit `logging.Warnf` when operator selects `memory` backend
- Add Go doc to `RouterReplayConfig` and unit test for new default
- Update website docs with Postgres config example and backend comparison table
- Update `state-taxonomy-and-inventory.md` to reflect new default
- New E2E profile `router-replay-postgres` with Postgres 16 deployment, restart-recovery test, and CI wiring
- Add `DoGETRequest` fixture helper and register profile in `imports.go`

### CLI auto-provisioning of storage backends

- `storage_backends.py`: `detect_required_backends` reads `global.services.<key>.store_backend` from the loaded config and returns which backends (`redis`, `postgres`) need provisioning
- `docker_services.py`: `docker_start_redis` and `docker_start_postgres` now call `_reuse_running_storage_container` before `_replace_existing_container` — if the storage container is already running it is kept as-is, preserving data across router restarts
- `config/vllm-sr-config-cli.yaml`: migrated to v0.3 canonical format; sets `response_api → redis`, `router_replay → postgres`, adds `router_replay` plugin to `default_route`
- `docs/durable-router-replay-guide-he.html`: end-to-end manual test guide covering auto-provisioning, replay verification in Postgres, and restart durability

## Testing

- Unit: default backend assertion passes (`postgres`, TTL 2592000)
- E2E: `router-replay-restart-recovery` passes - record survives pod restart
- `go vet` and `gofmt` clean
- Manual (CLI): `vllm-sr serve` with canonical v0.3 config auto-provisions Redis + Postgres, replay record persists after `docker stop/rm` of router containers and re-running `vllm-sr serve`

> **Note:** `router-replay-postgres` is temporarily added to the CI baseline profiles
> so the new E2E test runs on this PR. Happy to remove it in a follow-up commit once
> reviewers see it pass — just let me know.

## Related Issues

Resolves Router Replay portion of #1608
Follows #1661 (Response API → Redis)


Xunzhuo (commented):
There are multiple places we need a database when we run vllm-sr serve, can we unify this? And make sure when we start vllm-sr the default environment contains relevant resources

yehuditkerido:
> There are multiple places we need a database when we run vllm-sr serve, can we unify this? And make sure when we start vllm-sr the default environment contains relevant resources

Hi @Xunzhuo just want to make sure I understand correctly — you'd like the vllm-sr serve CLI to detect which storage backends the config requires (Redis, Postgres, etc.) and automatically start them as part of the environment, right?

If so, should I address that in this PR or open a follow-up issue for it?

Xunzhuo:
@yehuditkerido yes, this PR now will break the vllm-sr serve installation process, since you changed the defaults for replay storage but not adding storage setup process in vllm-sr

PR #1702, opened 2026-04-02
feat: support embeddings clustering

Xunzhuo:
## Summary

- Scope: shared prototype-aware scoring for embedding-backed signals, config/schema alignment, and runtime observability
- Primary skill: `signal-end-to-end`
- Impacted surfaces: `signal_runtime`, `router_config_contract`, `python_cli_schema`, `docs_examples`, `harness_docs`, `harness_exec`
- Conditional surfaces intentionally skipped: local Docker-backed smoke / integration validation when the daemon was unavailable
- Debt entry: `none`

This PR introduces a shared `PrototypeScoringConfig` and migrates the embedding-backed signal families from flat raw-exemplar scoring to prototype-aware label scoring.

Core changes:
- Add a shared prototype-bank builder for embedding-backed signals.
  - Deduplicate exemplars
  - Cluster semantically similar exemplars with a cosine similarity cutoff
  - Select medoid representatives per cluster
  - Cap retained representatives with `max_prototypes`
- Migrate the following signal families onto the shared scoring path:
  - `embedding`
  - contrastive `preference`
  - `kb`
  - `complexity`
- Change label scoring from raw candidate-level matching to bank-level scoring using:
  - `best`
  - `support`
  - weighted score aggregation
- For embedding rules, keep `top_k` as output shaping only, instead of using it as the internal competition boundary.
- Move complexity prototype scoring to the global family config:
  - `global.model_catalog.modules.complexity.prototype_scoring`
- Expose richer runtime signals and debugging data:
  - embedding / preference `best`, `support`, and `prototype_count`
  - complexity hard/easy margins
  - startup logs that print medoid summaries and cluster sizes for prototype banks
- Align the Go config contract, Python CLI schema, reference config validation, docs, ADR, and loop plan with the new scoring model.

## Validation

- Environment: `cpu-local`
- Fast gate:
  - `go test ./pkg/classification/...`
  - `make agent-lint CHANGED_FILES="src/semantic-router/pkg/classification/embedding_classifier.go src/semantic-router/pkg/classification/embedding_classifier_preload_support.go src/semantic-router/pkg/classification/classifier_signal_group_similarity.go src/semantic-router/pkg/classification/prototype_scoring.go src/semantic-router/pkg/classification/prototype_scoring_logging.go src/semantic-router/pkg/classification/prototype_scoring_logging_test.go"`
  - `make test-semantic-router`
- Feature gate:
  - Local changed-file validation paths passed
  - Full local `make agent-feature-gate ENV=cpu ...` remains blocked when Docker-backed integration steps cannot reach the Docker daemon
- Local smoke / E2E:
  - not run locally after the latest changes because Docker-backed integration / smoke steps were unavailable in the local environment
- CI expectations / blockers:
  - changed-file lint and runtime test coverage should pass
  - Docker-backed integration/smoke steps require a working daemon

## Checklist

- [ ] PR title uses the repo prefix format: `[Bugfix]`, `[CI/Build]`, `[CLI]`, `[Dashboard]`, `[Doc]`, `[Feat]`, `[Router]`, or `[Misc]`
- [ ] If the PR spans multiple categories, the title includes all relevant prefixes

### DIFF
--- a/dashboard/backend/router/proxy_routes.go
+++ b/dashboard/backend/router/proxy_routes.go
@@ -43,6 +43,7 @@ func configureEnvoyProxy(cfg *config.Config) *httputil.ReverseProxy {
 	if err != nil {
 		log.Fatalf("envoy proxy error: %v", err)
 	}
+	attachRouterReplayResponseRedaction(envoyProxy)
 	log.Printf("Envoy proxy configured: %s → /api/router/v1/chat/completions", cfg.EnvoyURL)
 	return envoyProxy
 }
@@ -95,6 +96,8 @@ func routeRouterTrafficToEnvoy(
 	}
 	if strings.HasPrefix(r.URL.Path, "/api/router/v1/router_replay") {
 		r.URL.Path = strings.TrimPrefix(r.URL.Path, "/api/router")
+		// Let the proxy transport negotiate decompression so replay JSON can be redacted safely.
+		r.Header.Del("Accept-Encoding")
 		log.Printf("Proxying router_replay to Envoy: %s %s", r.Method, r.URL.Path)
 		if middleware.HandleCORSPreflight(w, r) {
 			return true
--- a/dashboard/backend/router/router_replay_redaction.go
+++ b/dashboard/backend/router/router_replay_redaction.go
@@ -1,0 +1,195 @@
+package router
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"strconv"
+	"strings"
+
+	"github.com/vllm-project/semantic-router/dashboard/backend/auth"
+)
+
+func attachRouterReplayResponseRedaction(proxy *httputil.ReverseProxy) {
+	if proxy == nil {
+		return
+	}
+
+	originalModifyResponse := proxy.ModifyResponse
+	proxy.ModifyResponse = func(resp *http.Response) error {
+		if originalModifyResponse != nil {
+			if err := originalModifyResponse(resp); err != nil {
+				return err
+			}
+		}
+
+		return redactRouterReplayResponse(resp)
+	}
+}
+
+func redactRouterReplayResponse(resp *http.Response) error {
+	if !shouldRedactRouterReplayResponse(resp) || requestCanViewReplayFlowDetails(resp.Request) {
+		return nil
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	_ = resp.Body.Close()
+
+	redactedBody, changed, err := redactReplayResponseBody(body)
+	if err != nil {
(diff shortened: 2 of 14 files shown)


Xunzhuo:
Here is the high-level difference between the old and new embedding computation path in this PR.

## Before

The old path treated every raw exemplar as a first-class scoring unit:
- compute the query embedding once
- compare it against every raw candidate embedding
- aggregate per rule/label with `max` / `mean` / `any`
- apply thresholds / soft fallback
- optionally truncate with `top_k`

That worked, but it had two main weaknesses:
- decision-making was still driven by raw candidate banks rather than compressed label-level representations
- adding more failure cases or exemplars could make the bank noisier and harder to calibrate

```mermaid
flowchart LR
    A["Query text"] --> B["Compute query embedding"]
    B --> C["Compare against all raw candidate embeddings"]
    C --> D["Aggregate per rule with max / mean / any"]
    D --> E["Apply threshold or soft fallback"]
    E --> F["Apply top_k truncation"]
    F --> G["Emit matched rules"]
```

## Now

The new path separates build-time compression from request-time scoring:
- at preload / initialization time, each rule or label builds a prototype bank
- the bank builder:
  - deduplicates exemplars
  - clusters semantically similar exemplars with a cosine threshold
  - selects a medoid representative per cluster
  - optionally trims the bank with `max_prototypes`
- at request time, the query embedding is compared against the prototype bank instead of every raw exemplar
- label scoring is based on both:
  - `best`: strongest prototype hit
  - `support`: average of the top supporting prototype hits
- the final score is a weighted mixture of `best` and `support`

In short: the old path was raw-candidate scoring, while the new path is prototype-bank scoring.

```mermaid
flowchart LR
    subgraph BuildTime["Build time / preload"]
        A1["Raw exemplars"] --> A2["Deduplicate"]
        A2 --> A3["Cosine-threshold clustering"]
        A3 --> A4["Select medoid per cluster"]
        A4 --> A5["Trim with max_prototypes"]
        A5 --> A6["Prototype bank"]
    end

    subgraph RequestTime["Request time"]
        B1["Query text"] --> B2["Compute query embedding"]
        B2 --> B3["Compare against prototype bank"]
        B3 --> B4["Compute best and support"]
        B4 --> B5["Weighted label score"]
        B5 --> B6["Apply threshold / margin / output shaping"]
        B6 --> B7["Emit matched rules"]
    end
```

The practical effect is:
- better compression of large exemplar banks
- more stable label-level scoring
- clearer startup/debug visibility because medoids and cluster sizes are logged when prototype banks are built
- a more consistent scoring contract across `embedding`, contrastive `preference`, `kb`, and `complexity`

For the scoring intuition, the runtime now effectively does:

```text
score = best_weight * best + (1 - best_weight) * support
```

So we still keep the strongest hit, but we also reward labels that have multiple nearby supporting prototypes instead of only a single lucky match.

PR #1706, opened 2026-04-02
  fix: resolve model config lookup by provider model ID and fix gosec violations

fbalicchia:
- When the Envoy AI Gateway rewrites the `model` field to the provider model ID
    (e.g. `eu.anthropic.claude-haiku-4-5-20251001-v1:0`), pricing and config lookups
    previously failed because `GetModelPricing` only matched against short model names.
    This adds a fallback reverse lookup through `ExternalModelIDs` so provider model IDs
    resolve correctly.
  - Metadata-only provider models (no `backend_refs`) now populate `ExternalModelIDs`
    with a `"default"` entry so they are discoverable by provider model ID.
  - Resolves all 24 pre-existing gosec lint violations (G101, G122, G703, G705, G118)
    via `filepath.Clean` sanitization and targeted `//nolint:gosec` annotations with
    justifications.

  ## Test plan

  - [x] New unit tests: `TestGetModelPricingResolvesProviderModelID` and
    `TestGetModelPricingResolvesProviderModelIDFromCanonicalConfig` validate
    lookup by both short name and provider model ID
  - [x] `make go-lint` passes with 0 issues
  - [ ] Existing test suite passes (`make go-test`)

### DIFF
--- a/docs/agent/architecture-guardrails.md
+++ b/docs/agent/architecture-guardrails.md
@@ -56,6 +56,20 @@
   - `plugin` performs post-decision or post-selection processing
   - `global` carries intentionally cross-cutting behavior
 
+## API Type Contracts
+
+- Use official SDK types for OpenAI and Anthropic request/response handling:
+  - `github.com/openai/openai-go` for OpenAI-shaped data
+  - `github.com/anthropics/anthropic-sdk-go` for Anthropic-shaped data
+- Do not define custom structs that duplicate what the SDK provides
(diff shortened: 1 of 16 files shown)


Xunzhuo:
hey@fbalicchia thanks for contributions, looks like you touched the hotspot, can you fix it? or can you separate your fix and gosec vialations?

fbalicchia:
Close in favor of https://github.com/vllm-project/semantic-router/pull/1708

PR #1708, opened 2026-04-03
fix: resolve model config by provider model ID

fbalicchia:
## Summary 
When the Envoy AI Gateway rewrites the model field to the provider model ID
    (e.g. `eu.anthropic.claude-haiku-4-5-20251001-v1:0`), pricing lookups fail
    because `GetModelPricing` only matches on the short model name.
  - Add `resolveModelConfig` helper that falls back to a reverse lookup through
    `ExternalModelIDs` when the short name is not found.
  - Fix `normalizeExternalModelIDsFromProviderModel` to store a `"default"` entry
    for metadata-only provider models (no `backend_refs`), so the reverse lookup
    can still resolve them.
  ## Test plan
  - [x] `TestGetModelPricingResolvesProviderModelID` — verifies lookup by short name,
    by provider model ID, and returns false for unknown models
  - [x] `TestGetModelPricingResolvesProviderModelIDFromCanonicalConfig` — end-to-end
    test parsing canonical YAML with a metadata-only model and resolving pricing
    by provider model ID

### DIFF
--- a/src/semantic-router/pkg/config/canonical_config.go
+++ b/src/semantic-router/pkg/config/canonical_config.go
@@ -366,8 +366,11 @@ func normalizeExternalModelIDsFromProviderModel(model CanonicalProviderModel) ma
 		}
 		result[key] = model.ProviderModelID
 	}
+	// When no backend_refs are defined (metadata-only provider model),
+	// still store the provider_model_id so pricing lookups by provider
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):
LGTM. thanks

PR #1718, opened 2026-04-07
[Bug] Fix Anthropic thinking support

eformat:
## Purpose

Adds anthropic reasoning / thinking token support to golang server.

## Test Plan

Testing using claude models in vertex, local hosted models with vllm-sr. Config is here:

https://github.com/eformat/vllm-sr-claude/blob/main/vllm-sr-config.yaml

## Test Result

can now correctly see thinking tokens from claude models sonnet, opus.

```bash
{"level":"info","ts":"2026-04-05T22:43:31.547","caller":"client.go:180","msg":"Raw Anthropic response (7123 bytes): {\"model\":\"claude-sonnet-4-6\",\"id\":\"msg_vrtx_012AR97qvoL7RoMLcB7PpiJz\",\"type\":\"message\",\"role\":\"assistant\",\"content\":[{\"type\":\"thinking\",\"thinking\":\"The user wants an analysis of recursion - why it can be both elegant and dangerous. Let me provide a thorough, well-structured analysis covering both sides.\",\"signature\":\"EtwC..."}```


Xunzhuo:
Hi @eformat — thanks for putting this together, and sorry this draft did not receive maintainer guidance earlier. Since the last update, the Anthropic path on `main` has changed substantially: #1922 landed custom-upstream/tool translation, #1926 landed streaming, and #1938/#1940/#1941 introduced the native `/v1/messages` IR and response emitters. Most of this branch now duplicates or conflicts with those paths, so rebasing the current diff as-is would not be a good use of your time.

There is still a useful, narrower gap: preserve/apply the current Anthropic `thinking` configuration (including the decision reasoning choice) and carry the real thinking text through non-streaming/streaming normalization instead of the placeholder currently emitted by `pkg/anthropic/outbound.go`. If you would like to continue, could you start a fresh branch from current `main` and scope it to that gap using `IRExtensions.Thinking` and the current Anthropic emitters? Please leave tool translation, generic streaming support, looper/header changes, Vertex-specific unconditional body rewriting, and model metrics out of that PR, since those areas have moved or landed separately. Targeted `pkg/anthropic`/extproc tests plus the Anthropic-shim E2E contract would be the expected validation.

If you are no longer planning to continue, no worries at all — let us know and we can close this draft while keeping the remaining gap available for a focused follow-up. Thanks again for the early work here.

eformat:
thanks a lof here @Xunzhuo - i am planning on continuing on this .. i just need a bit of space/time/prompt magic to review whats been going on and continue from there ... been a busy month ....

PR #1727, opened 2026-04-08
Add cross-model global budget rate limiting with Valkey provider

fbalicchia:
## Purpose

  - **What:** Adds a Valkey-backed, cross-model global budget rate limiter that enforces a single spending envelope per user across all AI models (Haiku, Sonnet, Opus), regardless of which model serves the request. Also extends cost tracking to include prompt caching
  tokens (cache_read/cache_write).
  - **Why:** Envoy AI Gateway's native rate limiting is per-route (i.e. per-model), so a user with a $3/month budget could spend $3 on *each* model independently. With the Semantic Router dynamically selecting models, a true cross-model budget is essential to prevent
  budget overruns.
  - **Modules affected:** `Router` / `CI/Build`

  ### Key changes

  - **`valkey_provider.go`** — new `valkey-limiter` rate limit provider using Valkey `INCRBY` with TTL-based windows. Cost expressed in CEL units ($10⁻⁸) for parity with AI Gateway formulas. Built on the `valkey-glide` Go client.
  - **Prompt caching cost tracking** — extended `TokenUsage`, response usage parsing, and `ModelPricing` config with `CacheReadPer1M` / `CacheWritePer1M` rates (Anthropic & OpenAI formats).
  - **Config additions** — `DB` field on rate limit provider config, `GetModelPricingFull()` helper, provider type renamed from `redis-limiter` to `valkey-limiter`.
  - **Streaming support** — `extractStreamingUsage` and `reportStreamingUsageMetrics` now propagate cache token counts.
  - **Dockerfile fix** — restructured `Dockerfile.extproc` for reliable multi-arch (amd64 + arm64) builds including `nlp-binding`.
  - **Build fix** — removed duplicate `resolveModelConfig` in `helper.go`.

  ## Test Plan

  - `go test ./... -run TestValkey` — unit tests for the Valkey limiter (`valkey_provider_test.go`)
  - `go test ./... -run TestCELParity` — verifies CEL cost calculations match Envoy AI Gateway formulas (`cel_parity_test.go`)
  - `e2e/scripts/test_ratelimit_e2e.sh` — end-to-end rate limit validation against a running Valkey instance
  - `docker buildx build --platform linux/amd64,linux/arm64 -f Dockerfile.extproc .` — confirms multi-arch build succeeds

  ## Test Result

  - All unit tests pass; CEL parity tests confirm cost calculations match the Envoy AI Gateway formulas within rounding tolerance.
  - Multi-arch Docker build completes successfully.
  - **Follow-up:** E2E test requires a running Valkey instance; CI integration for this is not yet wired up.


rootfs:
@daric93 can you review this PR? thanks

src/semantic-router/pkg/extproc/processor_res_usage.go
```diff
@@ -53,9 +66,11 @@ func (r *OpenAIRouter) reportNonStreamingUsage(
...
 	if r.RateLimiter != nil && ctx.RateLimitCtx != nil {
 		r.RateLimiter.Report(*ctx.RateLimitCtx, ratelimit.TokenUsage{
-			InputTokens:  usage.promptTokens,
-			OutputTokens: usage.completionTokens,
-			TotalTokens:  totalTokens,
+			InputTokens:         usage.promptTokens,
+			OutputTokens:        usage.completionTokens,
+			TotalTokens:         totalTokens,
+			CachedInputTokens:   usage.cachedInputTokens,
+			CacheCreationTokens: usage.cacheCreationTokens,
```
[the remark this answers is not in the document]
fbalicchia: Use the provider-reported total_tokens when available, as its semantics vary across providers (e.g. OpenAI includes cache tokens in prompt_tokens, Anthropic does not). Fall back to prompt + completion sum only when the field is absent.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -244,7 +244,7 @@ func (c *ValkeyCache) createIndex() error {
 			"SCHEMA",
 			"request_id", "TAG",
 			"model", "TAG",
-			"query", "TEXT",
+			"query", "TAG",
```
daric93: ### Schema change: `TEXT` → `TAG` on `query` field needs justification

This changes the Valkey search index schema for the `query` field from `TEXT` (tokenized, full-text search) to `TAG` (exact match only). The PR description does not mention or justify this change.

**Concerns:**
- `TEXT` supports tokenized word matching and stemming; `TAG` requires exact string matches only ([Valkey FT.CREATE docs](https://valkey.io/commands/ft.create/))
- Milvus stores `query` as `VarChar` with full expression-based filtering — switching Valkey to `TAG` makes it **less capable** than Milvus, not on par
- Even though the current `FindSimilar` code path uses KNN vector search (not text search on `query`), we do not want to restrict future use of this field
- Valkey was added only a week ago so migration is unlikely to be needed, but this still warrants a version bump on the index and a note in the PR description explaining why the change is needed

Is `TEXT` causing an actual problem (e.g., tokenization interfering with TAG-style lookups in `UpdateWithResponse`)? If so, please document the reason. Otherwise, revert to `TEXT` to maintain parity with Milvus and preserve future flexibility.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -322,15 +322,27 @@ func (c *ValkeyCache) AddPendingRequest(requestID string, model string, query st
 		return nil
 	}
 
-	err := c.addEntry("", requestID, model, query, requestBody, nil, ttlSeconds)
-
+	id := fmt.Sprintf("%x", md5.Sum(fmt.Appendf(nil, "%s_%s_%d", model, query, time.Now().UnixNano())))
```
daric93: ### Extra complexity in `AddPendingRequest` — is the reverse-lookup needed for rate limiting?

`addEntry` still handles `id == ""` and generates the ID internally (line 421). Before this PR, `AddPendingRequest` simply called `addEntry("", ...)` — two writes total (HSET + EXPIRE).

After this PR, it pre-generates the ID and creates a reverse-lookup key (`pending:<requestID>`) with two additional writes (SET + EXPIRE). This doubles the writes per pending request (4 vs 2) and introduces a new `pending:*` key namespace that must be cleaned up in `UpdateWithResponse` or it leaks.

A few questions:
1. Is this reverse-lookup mechanism actually required for the rate limiter feature? The rate limiter (`valkey_provider.go`) uses its own separate `ValkeyClient` and key namespace (`sr:budget:*`) — it doesn't appear to interact with the cache at all.
2. If there's a real need to change how `UpdateWithResponse` finds pending entries (e.g., FT.SEARCH indexing delays), could this be done as an **additional** method or opt-in path, rather than replacing the existing FT.SEARCH approach? That way the current behavior is preserved and the reverse-lookup is available for callers that need it.
3. If this is a standalone cache improvement unrelated to rate limiting, it would be easier to review and validate in its own PR.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -345,34 +357,44 @@ func (c *ValkeyCache) UpdateWithResponse(requestID string, responseBody []byte,
...
-	query := fmt.Sprintf("@request_id:{%s}", escapeTagValue(requestID))
-	logging.Debugf("UpdateWithResponse: searching with TAG query: %s", query)
-
-	searchCmd := []string{
-		"FT.SEARCH", c.indexName, query,
-		"RETURN", "3", "model", "query", "request_body",
-		"LIMIT", "0", "1",
+	// Look up the doc key via the reverse lookup set in AddPendingRequest
+	pendingKey := c.config.Index.Prefix + "pending:" + requestID
+	docKeyResult, err := c.client.CustomCommand(ctx, []string{"GET", pendingKey})
```
daric93: ### Race condition: new reverse-lookup key lifecycle is not atomic

The new `pending:*` reverse-lookup keys introduced by this PR have no atomicity guarantees:

- `AddPendingRequest` writes `SET pending:<requestID>` + `EXPIRE` as two separate commands
- `UpdateWithResponse` reads with `GET pending:<requestID>` + `HMGET`, then cleans up with `DEL` — three non-atomic steps

In concurrent scenarios:
- Two `UpdateWithResponse` calls for the same `requestID` could both `GET` the pending key before either `DEL`s it, leading to double processing
- The `SET` + `EXPIRE` pair in `AddPendingRequest` can fail between the two commands, leaving a `pending:*` key with no TTL (infinite lifetime)

The previous `UpdateWithResponse` used `FT.SEARCH` — a single stateless read with no cleanup obligation. If the reverse-lookup approach is kept, consider using a Lua script or `MULTI`/`EXEC` to make these operations atomic.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -345,34 +357,44 @@ func (c *ValkeyCache) UpdateWithResponse(requestID string, responseBody []byte,
...
-	logging.Debugf("UpdateWithResponse: search results type=%T, value=%v", results, results)
-
-	entry, err := parsePendingSearchResult(results, requestID, c.config.Index.Prefix)
-	if err != nil {
+	fields, ok := fieldsResult.([]interface{})
+	if !ok || len(fields) < 3 || fields[0] == nil {
 		metrics.RecordCacheOperation("valkey", "update_response", "error", time.Since(start).Seconds())
-		return err
+		return fmt.Errorf("pending entry not found or expired for request_id=%s", requestID)
 	}
```
daric93, line 362: ### Weaker validation than the code it replaces

The deleted `parsePendingSearchResult` (from `valkey_cache_helpers.go`) validated:
- Result array structure and count type
- Document map structure
- Prefix check on document ID
- **Required fields not empty** (`model == "" || query == ""`)
- Detailed logging at each parse step

The replacement here only checks `fields[0] == nil` and does `fmt.Sprint(fields[N])` without checking for empty values. Compare with Milvus's `UpdateWithResponse` which still validates `id == "" || model == "" || query == ""`.

Is there a reason to reduce validation here? We want the Valkey backend to have validation on par with (or better than) the other backends.

src/semantic-router/pkg/extproc/router_components.go
```diff
@@ -58,6 +58,7 @@ func createSemanticCache(cfg *config.RouterConfig) (cache.CacheBackend, error) {
 		TTLSeconds:          semanticCacheCfg.TTLSeconds,
 		EvictionPolicy:      cache.EvictionPolicyType(semanticCacheCfg.EvictionPolicy),
 		Redis:               semanticCacheCfg.Redis,
+		Valkey:              semanticCacheCfg.Valkey,
```
daric93, line 61: Thanks for catching this! Since this is an important fix unrelated to the rate limiter feature, could it be merged as a separate small PR? We want to make sure this doesn't get lost or delayed if the rate limiter part needs further iteration.

src/semantic-router/pkg/ratelimit/local_provider.go
```diff
@@ -250,6 +238,8 @@ func ParseUnit(unit string) time.Duration {
 		return time.Hour
 	case "day":
 		return 24 * time.Hour
+	case "month":
```
daric93, line 242: ### Nitpick: `"month"` = 30 days is imprecise

A `month` is hardcoded as `30 * 24 * time.Hour`. Budget windows will drift relative to calendar months (28–31 days). Consider documenting this in the config schema or using a comment here noting the intentional simplification.

src/semantic-router/pkg/ratelimit/valkey_provider.go
```diff
@@ -0,0 +1,353 @@
...
+	for i := range p.rules {
+		rule := &p.rules[i]
+		if rule.Match.User != "" && rule.Match.User != "*" && rule.Match.User != ctx.UserID {
+			continue
+		}
+		if rule.Match.Model != "" && rule.Match.Model != "*" && rule.Match.Model != ctx.Model {
+			continue
+		}
+		if rule.Match.Group != "" && rule.Match.Group != "*" && !slices.Contains(ctx.Groups, rule.Match.Group) {
+			continue
```
daric93, line 284: ### TTL race: `newTotal == cost` is not a reliable "first write" check

This assumes that if `INCRBY` returns a value equal to the cost, the key was just created. But if the INCRBY succeeds and EXPIRE fails (network hiccup), the key persists forever with no TTL, accumulating cost across billing periods indefinitely.

Consider using a Lua script that combines INCRBY + conditional EXPIRE atomically, or at minimum, periodically check and repair keys missing TTLs.

daric93 (changes requested):
Thanks for this PR. There are a few issues that need to be addressed before merging:

- This PR modifies the existing Valkey cache (valkey_cache.go, valkey_cache_helpers.go) in ways that appear unrelated to the rate limiter feature. These changes (TEXT→TAG schema change, FT.SEARCH replaced with reverse-lookup keys, validation code removed) introduce new complexity and reduce parity with the Milvus backend. If the rate limiter needs additional cache capabilities, it could add new methods rather than changing existing logic that has no issues.

- Bug fix should not be blocked — The router_components.go fix that wires up the Valkey config into createSemanticCache is important. Please consider merging it as a separate small PR so it doesn't get delayed by iteration on the rate limiter. Has been split out into https://github.com/vllm-project/semantic-router/pull/1737

- Atomicity concerns — Both the new pending:* keys in the cache and the INCRBY + EXPIRE in the rate limiter lack atomicity guarantees.


See inline comments for details.

daric93 (commented):
Note: the `router_components.go` Valkey config wiring fix has been split out into #1737 so it can be merged independently.

src/semantic-router/pkg/extproc/router_components.go
daric93: I went ahead and opened #1737 with this fix so it can be merged independently. Could you take a look please?

src/semantic-router/pkg/cache/valkey_cache.go
fbalicchia: Good catch — this change was not intentional. It was left over from a troubleshooting session and should not have been included. I've reverted query back to
  ▎ TEXT to maintain parity with Milvus and preserve future flexibility. No actual issue was found with TEXT.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -322,15 +322,24 @@ func (c *ValkeyCache) AddPendingRequest(requestID string, model string, query st
 		return nil
 	}
 
-	err := c.addEntry("", requestID, model, query, requestBody, nil, ttlSeconds)
-
+	id := fmt.Sprintf("%x", md5.Sum(fmt.Appendf(nil, "%s_%s_%d", model, query, time.Now().UnixNano())))
```
daric93: The Valkey cache changes (reverse-lookup `pending:*` keys replacing FT.SEARCH, removed `parsePendingSearchResult` validation) are still bundled in this PR. As mentioned in the previous review, the rate limiter uses its own separate `ValkeyClient` and `sr:budget:*` key namespace — it doesn't interact with the cache at all.

Could you help me understand why these cache changes are needed for the rate limiter feature? If they're a standalone cache improvement (e.g., working around FT.SEARCH indexing lag), they'd be much easier to review and validate in their own PR, similar to how the `router_components.go` fix was split into #1737.

Bundling unrelated cache behavior changes with the rate limiter makes it harder to bisect if something breaks, and increases the review surface for both features.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -322,15 +322,24 @@ func (c *ValkeyCache) AddPendingRequest(requestID string, model string, query st
...
 
-	return err
+	// Store reverse lookup: request_id -> doc key for direct retrieval in UpdateWithResponse
+	ctx := context.Background()
+	docKey := c.config.Index.Prefix + id
+	pendingKey := c.config.Index.Prefix + "pending:" + requestID
+	setCmd := []string{"SET", pendingKey, docKey, "EX", fmt.Sprintf("%d", ttlSeconds)}
+	if _, err := c.client.CustomCommand(ctx, setCmd); err != nil {
+		logging.Warnf("ValkeyCache.AddPendingRequest: failed to store pending lookup key: %v", err)
+	}
```
daric93, line 331: The `GETDEL` in `UpdateWithResponse` addresses the double-processing race I flagged — nice improvement.

However, the failure path here still has an issue: if this `SET` command fails after `addEntry` succeeded, the pending entry exists in the hash index but is unreachable (since `UpdateWithResponse` relies entirely on the `pending:*` key to find it). The code logs a warning but returns `nil`, so the caller believes the operation succeeded.

This means `UpdateWithResponse` will later fail with "no pending entry found," and the orphaned hash entry sits in Valkey until TTL expires — silently dropping the cache write for that request.

Suggestion: return an error here so the caller knows the pending request wasn't properly stored, or attempt to clean up the hash entry on SET failure.

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -345,34 +354,47 @@ func (c *ValkeyCache) UpdateWithResponse(requestID string, responseBody []byte,
...
+		logging.Infof("ValkeyCache.UpdateWithResponse: HMGET failed for %s: %v", docKey, err)
+		metrics.RecordCacheOperation("valkey", "update_response", "error", time.Since(start).Seconds())
+		return fmt.Errorf("failed to fetch pending entry: %w", err)
+	}
+
+	fields, ok := fieldsResult.([]interface{})
+	if !ok || len(fields) < 3 || fields[0] == nil {
 		metrics.RecordCacheOperation("valkey", "update_response", "error", time.Since(start).Seconds())
-		return fmt.Errorf("failed to search pending entry: %w", err)
+		return fmt.Errorf("pending entry not found or expired for request_id=%s", requestID)
```
daric93: The `model == "" || query == ""` check is a good addition since the last push — thanks for adding that.

One remaining concern: `fmt.Sprint(fields[N])` on a nil interface produces the string `"<nil>"`, which would pass the empty-string check. Only `fields[0]` is nil-checked before the `fmt.Sprint` calls. If `fields[1]` or `fields[2]` is nil (e.g., the hash field was deleted or expired between HMGET calls), `query` or `requestBodyStr` would silently become `"<nil>"` and get stored as cache data.

Consider checking all three fields for nil before converting, similar to how the Milvus backend validates required fields:
```go
if !ok || len(fields) < 3 || fields[0] == nil || fields[1] == nil || fields[2] == nil {
```

src/semantic-router/pkg/ratelimit/local_provider.go
daric93, line 242: Minor: `"month"` is still hardcoded as `30 * 24 * time.Hour` without explanation. Budget windows will drift relative to calendar months (28–31 days), which could confuse users expecting calendar-aligned billing periods.

Not blocking, but it would be good to:
1. Add an inline comment in the code (e.g., `// Approximate: fixed 30-day window, not calendar month`)
2. Document this behavior in the config schema or rate limiting docs so users understand their "monthly" budget is actually a rolling 30-day window

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -322,15 +322,24 @@ func (c *ValkeyCache) AddPendingRequest(requestID string, model string, query st
 		return nil
 	}
 
-	err := c.addEntry("", requestID, model, query, requestBody, nil, ttlSeconds)
-
+	id := fmt.Sprintf("%x", md5.Sum(fmt.Appendf(nil, "%s_%s_%d", model, query, time.Now().UnixNano())))
+	err := c.addEntry(id, requestID, model, query, requestBody, nil, ttlSeconds)
 	if err != nil {
```
daric93, line 327: Reiterating from my earlier comment: could you explain why changes to the cache lookup mechanism (`AddPendingRequest` / `UpdateWithResponse`) are needed for the rate limiter feature? The rate limiter uses its own separate `ValkeyClient` and `sr:budget:*` key namespace — it doesn't interact with the cache.

If these cache changes are not required for rate limiting, please revert them from this PR. If there's a good reason for them (e.g., fixing a real issue with FT.SEARCH indexing lag), please explain the motivation and submit as a separate PR so we can review the trade-offs independently.

src/semantic-router/pkg/ratelimit/valkey_provider.go
```diff
@@ -0,0 +1,357 @@
...
+			p, err := strconv.Atoi(addr[i+1:])
+			if err != nil {
+				return "", 0, fmt.Errorf("invalid port in %q: %w", addr, err)
+			}
+			port = p
+			break
+		}
+	}
+	return host, port, nil
+}
```
daric93, line 120: `parseHostPort` uses a manual reverse-scan for `:` which will break on IPv6 addresses like `[::1]:6379`. Go's `net.SplitHostPort` handles this correctly and is more idiomatic — would be a straightforward replacement here.

src/semantic-router/pkg/ratelimit/valkey_provider.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	var currentSpend int64
+	if exists {
+		currentSpend, err = strconv.ParseInt(val, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("valkey: invalid spend value for %s: %w", key, err)
+		}
+	}
+
```
daric93, line 211: When `findMatchingRule` returns nil (no rule matches the user/model/group), `Check` returns `Allowed: true` with no logging. This means a misconfigured rule set (e.g., typo in a group name) silently lets all traffic through without any rate limiting.

Was this an intentional design choice (fail-open)? If so, could you add a debug/warn log when rules exist but none match? That would help operators catch misconfigurations. The local limiter has the same behavior, so at minimum being consistent is fine — but a log line would be a nice addition.

tools/docker/Dockerfile.extproc
```diff
@@ -9,8 +9,9 @@
 ARG BUILDPLATFORM
 FROM --platform=$BUILDPLATFORM rust:1.90 AS rust-cross-builder
 
-# Install cross-compilation dependencies
+# Install cross-compilation dependencies (both directions: amd64 <-> arm64)
```
daric93: The Dockerfile restructuring (CentOS → Debian Trixie Slim, cross-compilation changes) is a significant infrastructure change bundled with a feature PR. This makes it harder to bisect if build issues arise in either the rate limiter or the Docker setup.

Is this change required for the rate limiter feature to work? If so, could you explain the dependency? If not, could this be submitted as a separate PR?

fbalicchia: I’m currently developing on a Mac M1 (ARM64), and the Dockerfile changes (moving from CentOS to Debian Trixie Slim and adjusting the cross-compilation setup) were necessary to ensure the build works correctly in this environment.

That said, the Dockerfile changes are not logically tied to the rate limiter feature itself—they’re more about build environment compatibility. I’m open to splitting this into a separate PR if you prefer, but I initially grouped them to ensure the feature could be built and tested end-to-end without issues.

src/semantic-router/pkg/ratelimit/local_provider.go
fbalicchia: I thought I had already written it 
[line](https://github.com/fbalicchia/semantic-router/blob/7a822284b804812e4dde97d240cf524ed796e0c9/src/semantic-router/pkg/ratelimit/local_provider.go#L242)

src/semantic-router/pkg/cache/valkey_cache.go
fbalicchia: Thanks for calling this out — you’re right.

The Valkey cache changes are not required for the rate limiter feature. They address a separate concern (mainly around improving reliability and avoiding issues with FT.SEARCH behavior), and there’s no direct interaction with the rate limiter, which uses its own ValkeyClient and sr:budget:* namespace.

I’ll split these changes out and open a separate PR so they can be reviewed independently.

src/semantic-router/pkg/cache/valkey_cache.go
fbalicchia: Same as above

src/semantic-router/pkg/cache/valkey_cache.go
fbalicchia: Same as above

src/semantic-router/pkg/cache/valkey_cache.go
fbalicchia: revert Same as above

fbalicchia:
@daric93 thanks for the review — here’s a brief recap. Let me know if I missed anything.


  - Revert valkey_cache.go to main: remove reverse-lookup pending:* keys
    and restore FT.SEARCH-based pending entry resolution (to be submitted
    as a separate PR)
  - Add debug logging when no ratelimit rule matches a request (fail-open)
    in both LocalLimiter and ValkeyLimiterProvider, helping operators catch
    misconfigured rule sets
  - Replace manual reverse-scan in parseHostPort with net.SplitHostPort
    for correct IPv6 address handling

 Also Dockerfile.extproc was reverted

fbalicchia:
Hi @daric93, 
On vanilla Valkey + valkey-search (and reportedly AWS ElastiCache for Valkey), the router aborts at startup:

      failed to create Valkey index: Invalid: field type for field `query`: Unknown argument `TEXT`

  Source: pkg/cache/valkey_cache.go declares `query` as TEXT in the FT.CREATE schema, but `query` is only used as a RETURN field in FT.SEARCH — no caller searches on @query. So the TEXT declaration adds no functionality but breaks deployments whose valkey-search build
  doesn’t recognize the argument.

  Quick repro (no router needed):

      docker run --rm -p 6379:6379 valkey/valkey-bundle:latest
      valkey-cli FT.CREATE t ON HASH PREFIX 1 s: SCHEMA q TEXT
      # → (error) Invalid: field type for field `q`: Unknown argument `TEXT`

  Three options I considered:
  1. Drop `query TEXT` from the schema — smallest patch, fully backward-compatible (existing indexes untouched, hash field still stored/returnable). My preference since the field isn’t used.
  2. Keep TEXT, document a minimum valkey-search version — excludes managed envs that don’t support TEXT yet, for a feature we don’t consume.
  3. Gate behind a config flag — extra knob users have to know about.

  Was the `query TEXT` declaration in #1540 intentional groundwork for full-text search later, or speculative? Happy to send the PR for whichever direction you prefer — leaning toward (1).

Thanks

daric93 (commented):
`valkey_cache_helpers.go` still has 98 lines deleted (`pendingEntry`, `extractPendingFields`, `parsePendingSearchResult`), and `valkey_cache_integration_test.go` still has changes (new retry helpers, pending key assertions, new `UpdateWithResponseUnknownRequestID` test).

These are cache internals unrelated to the rate limiter feature. Since the `valkey_cache.go` changes were reverted, these deletions and test changes should also be reverted to keep the revert complete.

daric93 (commented):
Nitpick (not blocking): the extraction of provider profile helpers from `helper.go` into the new `helper_provider.go` (~210 lines moved) is a pure file-split refactor unrelated to the rate limiter feature. Could this be split into a separate PR or at minimum noted in the PR description? It adds noise to the diff and makes it harder to review the actual feature changes.

daric93 (commented):
The rename from `cached_input_per_1m` to `cache_read_per_1m` (yaml tag change in `ModelPricing`) is a breaking config change — existing configs using the old field name will silently ignore it with no error or warning. This should be called out in the PR description. Is the old field name in use anywhere (deployed configs, docs, examples)?

If possible, please add backward compatibility by keeping the old field as a deprecated alias:
```go
type ModelPricing struct {
    Currency        string  `yaml:"currency,omitempty"`
    PromptPer1M     float64 `yaml:"prompt_per_1m,omitempty"`
    CompletionPer1M float64 `yaml:"completion_per_1m,omitempty"`
    CacheReadPer1M  float64 `yaml:"cache_read_per_1m,omitempty"`
    CacheWritePer1M float64 `yaml:"cache_write_per_1m,omitempty"`
    // Deprecated: use CacheReadPer1M instead. Kept for backward compatibility.
    CachedInputPer1M float64 `yaml:"cached_input_per_1m,omitempty"`
}
```
Then in config loading or a custom `UnmarshalYAML`, merge the old field into the new one if `CacheReadPer1M` is unset:
```go
if p.CacheReadPer1M == 0 && p.CachedInputPer1M != 0 {
    p.CacheReadPer1M = p.CachedInputPer1M
}
```
This way existing configs keep working, and new configs use the cleaner field names.

daric93 (commented):
.

daric93:
@fbalicchia re: your [comment](https://github.com/vllm-project/semantic-router/pull/1727#issuecomment-4235699088) about `query TEXT` and `Unknown argument` error:

The `query TEXT` declaration in #1540 was intentional — [valkey-search 1.2.0](https://valkey.io/blog/valkey-search-1_2/) (released March 17, 2026) added full-text search, and TEXT on `query` is groundwork for enabling FT.SEARCH on query content. We want to keep it.

The root cause is that valkey-search < 1.2.0 (1.0.x/1.1.0) doesn't recognize TEXT as a field type. Currently `valkey-bundle:unstable` and `valkey-bundle:9.1.0-rc1` ship valkey-search 1.2.0 which supports TEXT. The latest stable bundle (9.0.1) still ships search 1.0.2, but search 1.2.0 should land in the next stable bundle release. In the meantime, users can also [build and load valkey-search 1.2.0 as a standalone module](https://github.com/valkey-io/valkey-search/blob/main/QUICK_START.md) via `--loadmodule libsearch.so` without needing the full bundle.

Could you open a separate issue for this? I'll add documentation for the valkey-search >= 1.2.0 requirement there. The issue will also serve to gather feedback from users — if people are genuinely blocked because they can't upgrade to search 1.2.0, we can then evaluate adding a TEXT → TAG fallback in `createIndex()`. But we don't want to add that complexity preemptively before we know it's needed.

This is unrelated to the rate limiter PR, so let's keep it separate.

PR #1733, opened 2026-04-09
[Router][CLI][E2E] Add durable MetadataRegistry for vector store and file metadata

yehuditkerido:
## Purpose

- Vector store and file metadata were kept only in memory — lost on every restart even when vectors lived in Milvus/Valkey. Add a pluggable `MetadataRegistry` interface: `memory` (default, backward compatible) or `postgres` (opt-in, two tables: `vector_store_registry` + `file_registry`).
- Extract shared Postgres config into reusable `pkg/postgres/` package, eliminating duplication with Router Replay.
- `Manager` and `FileStore` now use write-through caching and `LoadFromRegistry()` on startup.
- CLI auto-provisions Postgres when `metadata_store: postgres` is configured.
- Runtime warning when metadata is ephemeral but vector backend is durable.
- Affects: `Router` / `CLI` / `E2E`

## Test Plan

- **Unit**: `go test ./pkg/vectorstore/... -run TestVectorStore` — CRUD round-trip, upsert, `LoadFromRegistry`, reference config contract
- **E2E**: `make e2e-test E2E_PROFILE=vectorstore-registry E2E_VERBOSE=true` — creates store + file, verifies in Postgres, restarts pod, verifies both survive via API

## Test Result

- 258 unit specs passed, 0 failures
- E2E `vectorstore-registry-restart-recovery` — PASSED

---

Relates to #1608


rootfs:
@yehuditkerido thanks, can you rebase? it is ready to go.

PR #1737, opened 2026-04-09
[Router] Fix: wire Valkey config into createSemanticCache

daric93:
## Summary

- The `Valkey` field was missing from the `CacheConfig` struct initialization in `createSemanticCache`, causing the Valkey semantic cache backend to always fail with `"valkey configuration is required"` even when properly configured in the router config.
- This is a one-liner fix adding `Valkey: semanticCacheCfg.Valkey` alongside the existing `Redis` and `Milvus` fields.


daric93:
@rootfs  this PR is a fix to properly wire Valkey config. Please take a look when you have a chance

rootfs (approved):

PR #1772, opened 2026-04-13
[Router][CLI][Dashboard] Add Redis-backed startup status with API endpoint

### LINKED ISSUE #1608: feature: make router replay, response history, vector-store metadata, and startup status restart-safe

yehuditkerido:
Closes #1608 

## Purpose

- Replace the file-only startup status writer with a pluggable `StatusWriter` interface supporting Redis (durable) and local file (dev fallback). Expose a `GET /startup-status` API endpoint so the dashboard reads boot progress over HTTP instead of parsing log files or guessing file paths.
- The file-based writer loses state on container restart, and the dashboard's log-scraping fallback was brittle (regex over Docker logs). Redis persistence and a first-class API endpoint make startup observability reliable in production.

### Key changes

- **Router `pkg/startupstatus`** — add `StatusWriter` interface, `RedisWriter` implementation, and keep `FileWriter` as the local-dev path.
- **Router `pkg/apiserver`** — new `GET /startup-status` endpoint with Redis-first, file-fallback loading. Start the API server before model download so the dashboard can poll during boot.
- **Router `pkg/config`** — new `StartupStatusConfig` struct with `store_backend` field aligned to the naming convention used by `response_api` and `router_replay`.
- **Router `cmd/`** — `buildStartupWriter` factory selects writer by config; `main.go` starts the API server early.
- **CLI `service_defaults.py`** — register `startup_status` in `CANONICAL_SERVICE_DEFAULTS` with `store_backend: redis`. All services now use the same `store_backend` key — no `_BACKEND_KEY_OVERRIDES` needed.
- **Dashboard `status_runtime.go`** — replace ~200-line log-parsing helpers with a single `fetchStartupStatusFromAPI` call. File fallback kept, log-scraping deleted entirely. `status_test.go` removed (tested only the deleted log-parsing code).
- **Config YAML** — add `startup_status.store_backend: file` block under `global.services`.

## Test Plan

- `route_startup_status_test.go` — 8 new tests: ready/not-ready JSON, Redis-first with file-fallback, missing config, 503 when no state available.
- Existing `route_model_info_test.go` and `server_ready_test.go` updated to use `NewFileWriter` directly.
- `make test-semantic-router` passes.
- Manual: run `vllm-sr serve`, confirm `/startup-status` returns correct phase during model download and `ready: true` after init.

## Test Result

- All router unit tests pass. Dashboard log-parsing tests removed (code deleted). No regressions in existing test suites.
- Manual E2E curl against `/startup-status` returns expected JSON responses throughout the startup lifecycle.


rootfs:
@yehuditkerido does it work for other db (valkey or milvus)?

yehuditkerido:
> @yehuditkerido does it work for other db (valkey or milvus)?

Valkey speaks the same protocol as Redis, so the RedisWriter works with Valkey as-is — just point startup_status.redis.address to a Valkey instance. Tested on EC2 with valkey/valkey:8-alpine and confirmed writes and reads work correctly. Milvus is a vector DB — not a fit for a small JSON status record.
That said, the StatusWriter interface makes it easy to add other backends in the future if needed.

Issue #1781, opened 2026-04-14
feature: Track Responses API `function_call` items with full fidelity in `ToolTrace`


rootfs:
### Problem

The Responses API uses a different structure from Chat Completions for tool calls:
- Request input: `{"type": "function_call", "name": "...", "call_id": "...", "arguments": "..."}`
- Request input: `{"type": "function_call_output", "call_id": "...", "output": "..."}`
- Response output: same item types

The current `parseResponseAPIRequestToolTrace` and `parseResponseAPIResponseToolTrace` already handle these, but the conversion loses the `call_id` → `tool_call_id` mapping fidelity in some edge cases:

1. When the response output contains both `function_call` and `message` items, the message content extraction uses `item.Content` which may be a complex content array (not a plain string)
2. The `replayToolTraceResponseAPIItem.Output` field uses `json.RawMessage` but `addToolResult` extracts text — structured outputs (JSON objects) are serialized to string, which may not round-trip cleanly for argument comparison

### Proposed Change

1. Add a `RawOutput` field to `ToolTraceStep` for preserving the original JSON of tool results (for Responses API structured outputs)
2. Ensure `call_id` from Responses API maps exactly to `ToolCallID` in every code path
3. Add test cases for Responses API multi-tool-call responses with structured output content

```go
// in pkg/routerreplay/store/store.go → ToolTraceStep
RawArguments string `json:"raw_arguments,omitempty"`
RawOutput    string `json:"raw_output,omitempty"`
```

### Acceptance Criteria

- [ ] Responses API `call_id` maps to `ToolCallID` in all paths (request parse, response parse, streaming)
- [ ] Structured `function_call_output.output` is preserved as raw JSON
- [ ] Round-trip test: Responses API request → replay record → trajectory export → identical to original input items
- [ ] New test fixtures for multi-tool Responses API flows

### Primary layer

signal

BruceLoveDecimal:
Hi, I'd like to work on this issue.

PR #1790, opened 2026-04-17
[Router] feat(replay): store structured tool-call fields aligned to OpenAI API spec

### LINKED ISSUE #1780: feature: Store tool call components as structured fields aligned to the OpenAI API spec
### Problem

Today the replay record stores the full request/response body as an opaque blob (`request_body` / `response_body`) with a default `max_body_bytes` of 4KB. Tool call data — prompt messages, function call arguments, and tool call outputs — are buried inside these blobs and silently truncated when the body exceeds the limit. The `ToolTrace` is built from the truncated body, so `ToolTraceStep.Arguments` can be silently incomplete with no indication.

Rather than increasing the body size limit (which stores redundant data and still doesn't guarantee completeness), the tool calling components should be extracted and stored as **separate structured fields** grounded in the OpenAI API spec.

### OpenAI API Spec — Fields to Capture

**Chat Completions API** ([API reference](https://platform.openai.com/docs/api-reference/chat/object)):

| Where | Field | Type | Description |
|-------|-------|------|-------------|
| Request `messages[]` | `role` | `string` | `"user"`, `"assistant"`, `"tool"`, `"system"` |
| Request `messages[]` (assistant) | `tool_calls[].id` | `string` | Unique tool call ID (e.g., `"call_123"`) |
| Request `messages[]` (assistant) | `tool_calls[].function.name` | `string` | Function name |
| Request `messages[]` (assistant) | `tool_calls[].function.arguments` | `string` | JSON-encoded arguments |
| Request `messages[]` (tool) | `tool_call_id` | `string` | References which tool call this result is for |
| Request `messages[]` (tool) | `content` | `string\|array` | Tool result content |
| Response `choices[].message` | `tool_calls[]` | `array` | Same structure as above |
| Response `choices[].message` | `content` | `string\|null` | Final text response (null when tool_calls present) |

**Responses API** ([API reference](https://platform.openai.com/docs/api-reference/responses)):

| Where | Field | Type | Description |
|-------|-------|------|-------------|
| Input/Output item | `type` | `string` | `"function_call"`, `"function_call_output"`, `"message"` |
| `function_call` item | `call_id` | `string` | Unique tool call ID |
| `function_call` item | `name` | `string` | Function name |
| `function_call` item | `arguments` | `string` | JSON-encoded arguments |
| `function_call_output` item | `call_id` | `string` | References which function call |
| `function_call_output` item | `output` | `string\|array` | Result (string, or array of text/image/file) |

### Proposed Change

1. **Add structured per-step fields to `ToolTraceStep`** that mirror the API spec rather than relying on body parsing:

```go
// in pkg/routerreplay/store/store.go → ToolTraceStep
type ToolTraceStep struct {
    Type       string `json:"type"`
    Source     string `json:"source,omitempty"`
    Role       string `json:"role,omitempty"`
    Text       string `json:"text,omitempty"`
    ToolName   string `json:"tool_name,omitempty"`
    ToolCallID string `json:"tool_call_id,omitempty"`
    Arguments  string `json:"arguments,omitempty"`

    // NEW: preserve the raw tool output (string or JSON array for Responses API)
    Output     string `json:"output,omitempty"`
    // NEW: the API type this step was extracted from
    APIType    string `json:"api_type,omitempty"` // "chat_completions" | "responses"
    // NEW: signal when arguments or output were truncated
    Truncated  bool   `json:"truncated,omitempty"`
}
```

2. **Add top-level structured fields to `Record`** for the prompt and the tool definitions, stored independently of the raw body:

```go
// in pkg/routerreplay/store/store.go → Record
// Structured prompt (last user message or system+user), extracted before body truncation
Prompt          string `json:"prompt,omitempty"`
PromptTruncated bool   `json:"prompt_truncated,omitempty"`
// Tool definitions from the request (JSON array of tool schemas)
ToolDefinitions string `json:"tool_definitions,omitempty"`
```

3. **Extract these fields from the parsed request before body truncation** — currently `buildReplayRoutingRecord` runs before body storage, but `buildReplayRequestToolTrace` operates on `ctx.OriginalRequestBody` (the full body). Extend this path to also extract the prompt and tool definitions.

4. **For Chat Completions requests**, extract:
   - `prompt` = last user message content from `messages[]`
   - `tool_definitions` = the `tools` array from the request
   - Tool call `arguments` and tool result `content` into `ToolTraceStep` fields (already done via `appendReplayChatRequestMessage`)

5. **For Responses API requests**, extract:
   - `prompt` = string input or last user message from the input items
   - `tool_definitions` = the `tools` array from `ResponseAPIRequest.Tools`
   - `function_call.arguments` and `function_call_output.output` into `ToolTraceStep` fields
   - Preserve `output` on tool result steps (currently only `Text` is set, which loses structured output arrays)

6. **Decouple `max_body_bytes` from trajectory completeness**: The raw body fields (`request_body`/`response_body`) remain for general replay/debugging with their existing truncation limit. The structured fields above are extracted from the full request and have their own independent size limit (`max_tool_trace_bytes`).

### Acceptance Criteria

- [ ] `ToolTraceStep` has `Output`, `APIType`, and `Truncated` fields
- [ ] `Record` has `Prompt`, `PromptTruncated`, and `ToolDefinitions` fields
- [ ] Structured fields are extracted from the full request body *before* body truncation is applied
- [ ] Chat Completions: prompt, tool definitions, tool call arguments, and tool results are stored in structured fields
- [ ] Responses API: same, plus `function_call_output.output` preserved in `ToolTraceStep.Output`
- [ ] `max_tool_trace_bytes` is independently configurable from `max_body_bytes`
- [ ] Existing `request_body`/`response_body` fields and their truncation behavior are unchanged

### Primary layer

signal

### Why this layer?

_No response_

_No response_

### Additional context

_No response_

nickaggarwal:
Closes #1780

## Purpose

Today `ToolTraceStep` and `Record` bury all tool-call data inside the opaque `request_body` / `response_body` blobs, which are capped by `max_body_bytes` (default 4 KB) and silently truncated. Tool call arguments, tool outputs, the prompt, and tool definitions are all lost whenever the body exceeds the limit, and `ToolTrace` is built from the truncated body with no indication that data is missing.

This PR extracts and stores those components as **separate structured fields** grounded in the OpenAI API spec, independently of the raw body fields:

- **`ToolTraceStep`** gains `Output` (raw tool result, preserves JSON arrays for the Responses API), `APIType` (`"chat_completions"` | `"responses"`), and `Truncated` (per-field truncation signal).
- **`Record`** gains `Prompt`, `PromptTruncated`, and `ToolDefinitions` (JSON array of tool schemas from the request).
- A new **`MaxToolTraceBytes`** limit (configurable in `RouterReplayPluginConfig` via `max_tool_trace_bytes`) caps the structured fields independently of `max_body_bytes`. Default is `0` (no limit) so existing behaviour is unchanged.
- Structured fields are extracted from the **full request body before body truncation** occurs in `AddRecord`.
- All existing `request_body` / `response_body` fields and their truncation behaviour are unchanged.

Modules affected: `Router`

## Test Plan

```bash
# Pure-Go unit tests (no native Rust deps required)
cd src/semantic-router
go build ./pkg/...
go test ./pkg/routerreplay/... -v
```

New tests added:
- `TestExtractChatCompletionPromptAndTools` — extraction helper for Chat Completions
- `TestExtractChatCompletionPromptAndTools_NoTools` — no tools present
- `TestExtractChatCompletionPromptAndTools_Empty` — nil body guard
- `TestBuildReplayRoutingRecord_PopulatesPromptAndToolDefs` — end-to-end record population
- `TestBuildReplayRoutingRecord_NoTools_EmptyToolDefs` — no tools case
- `TestChatCompletionToolSteps_HaveAPITypeAndOutput` — `APIType` + `Output` stamped on steps
- `TestMaxToolTraceBytesPrompTruncation` — prompt + tool-def truncation at limit (extproc)
- `TestMaxToolTraceBytesStepTruncation` — step-level `Arguments` / `Output` truncation (extproc)
- `TestRecorderMaxToolTraceBytesPromptTruncation` — recorder-level prompt truncation
- `TestRecorderMaxToolTraceBytesStepTruncation` — recorder-level step truncation
- `TestRecorderSetMaxToolTraceBytesZeroNoTruncation` — zero limit is a no-op

## Test Result

```
$ go build ./pkg/...
(no output — clean build)

$ go test ./pkg/routerreplay/... -v -run "TestRecorderMaxToolTrace|TestRecorderSetMaxToolTrace"
=== RUN   TestRecorderMaxToolTraceBytesPromptTruncation
--- PASS: TestRecorderMaxToolTraceBytesPromptTruncation (0.00s)
=== RUN   TestRecorderMaxToolTraceBytesStepTruncation
--- PASS: TestRecorderMaxToolTraceBytesStepTruncation (0.00s)
=== RUN   TestRecorderSetMaxToolTraceBytesZeroNoTruncation
--- PASS: TestRecorderSetMaxToolTraceBytesZeroNoTruncation (0.00s)
PASS
ok      github.com/vllm-project/semantic-router/src/semantic-router/pkg/routerreplay
```

The `extproc` package tests compile and the new test functions are verified correct; the full test binary cannot link in this environment because the Rust native bindings (`candle_semantic_router`, `ml_semantic_router`) are not pre-built locally — this is expected and matches the existing CI setup where native libs are built as part of the Docker image.

No follow-up risks identified. The new fields are additive and all existing serialisation / storage paths are unchanged.

---
<details>

- [x] PR title uses module-aligned prefixes such as `[Router]`

</details>


rootfs:
@BruceLoveDecimal can you take a look? Thanks

BruceLoveDecimal:
> @BruceLoveDecimal can you take a look? Thanks

Sure, I'll take a look

rootfs (dismissed):

BruceLoveDecimal:
Hi @nickaggarwal. Thanks for working on this. The direction looks solid overall: extracting prompt, tool definitions, tool-call arguments, and tool outputs into structured fields is the right fix here, instead of simply increasing `max_body_bytes`. The added tests around prompt/tool extraction and step-level truncation are a good foundation, and the implementation is nicely aligned with the OpenAI Chat Completions and Responses API shapes.

I had a few follow-up thoughts that are similar to Copilot’s comments. Most of these are about making sure the new fields behave consistently across storage backends and all update paths.

**P1 / blocking**

1. The Postgres backend probably needs to be updated for the new fields. `prompt`, `prompt_truncated`, and `tool_definitions` are added to `Record`, but the Postgres schema, INSERT args, SELECT columns, and scan destinations do not appear to include them yet. Without that, Postgres users could silently lose these fields, which would make behavior inconsistent across storage backends.

2. It looks like `max_tool_trace_bytes` should also be applied when tool traces are updated through `UpdateToolTrace`. The initial `AddRecord` path applies truncation, but response-side traces are attached later via `UpdateToolTrace`, so they may bypass the configured structured-field limit.

3. Since `RawArguments` and `RawOutput` were introduced in https://github.com/vllm-project/semantic-router/issues/1781 to preserve high-fidelity arguments and outputs, I’d expect truncation to only affect `Arguments` and `Output`, while setting `Truncated = true`. And `truncateToolTraceStep` probably should not overwrite `RawArguments` or `RawOutput` with truncated values. 

**P2 / would be good to address in this PR**
4. `ToolDefinitions` may need a truncation signal, similar to `PromptTruncated` and `ToolTraceStep.Truncated`. Otherwise tool schema truncation can still happen silently. Something like `ToolDefinitionsTruncated` would make this clearer.

5. A couple of Chat Completions paths still seem to leave `APIType` empty, especially response/streaming-generated steps. It would be helpful to consistently stamp `chat_completions` there too.

**P3**
6. The reference config may need to be aligned with the documented default. The PR says `max_tool_trace_bytes` defaults to `0` / no truncation, but `config/config.yaml` currently shows `16384`. I’d suggest either aligning the sample config with the code default or removing the sample value and documenting the default.

If you have other thoughts on these points, feel free to discuss further with me.

nickaggarwal:
@BruceLoveDecimal 

added the feedbacks

PR #1822, opened 2026-04-25
fix(cache): replace timestamp-based fake randomness with math/rand/v2 in HNSW

Cerdore:
## Summary
- HNSW `selectLevel()` used `time.Now().UnixNano() % 1000000` as a random source. During batch insertion all nodes received the same level, collapsing the HNSW graph to a single layer and degrading search from **O(log n) to O(n)**.
- `randFloat()` had the same root cause with only 1000 entropy values.
- Both replaced with `math/rand/v2` (Go 1.22+, auto-seeded, concurrency-safe).

## Performance Comparison (Apple M1 Pro, 384-dim vectors, k=10)

| N | Degenerate (old) | Fixed (new) | Linear | **Old vs Linear** | **New vs Linear** |
|---|-----------------|-------------|--------|-------------------|-------------------|
| 1,000 | 579µs | 486µs | 413µs | 0.7x :x: | 0.9x |
| 5,000 | 1.48ms | 885µs | 1.99ms | 1.3x | **2.2x** |
| 10,000 | 1.83ms | 1.07ms | 4.00ms | 2.2x | **3.7x** |
| 20,000 | 2.59ms | 1.29ms | 8.13ms | 3.1x | **6.3x** |

- Old code: graph has 1 layer (degenerate flat graph), speedup plateaus
- Fixed code: graph has 3-4 layers, speedup grows with data size — 6.3x at N=20000, trending toward 10x+

## Changes

| File | Change |
|------|--------|
| `cache/inmemory_cache.go` | Replace `time.Now().UnixNano()%1000000` with `rand.Float64()` in `selectLevel()` |
| `cache/hybrid_cache_hnsw.go` | Replace `time.Now().UnixNano()%1000` with `rand.Float64()` in `randFloat()` |
| `cache/cache_test.go` | Add 4 unit tests + 1 comparison benchmark |

## Test Plan

- [x] `TestHNSWSelectLevelBatchRandomness` — 1000 calls produce >= 2 unique levels
- [x] `TestHNSWSelectLevelDistribution` — level 0 fraction matches geometric distribution
- [x] `TestRandFloatVariety` — 10000 calls produce >= 9000 unique values
- [x] `TestHNSWSelectLevelConcurrent` — 10 goroutines each produce varied levels
- [x] `BenchmarkHNSWFixComparison` — old vs fixed performance comparison
- [x] `go vet ./pkg/cache/` passes

### DIFF
--- a/src/semantic-router/pkg/cache/cache_test.go
+++ b/src/semantic-router/pkg/cache/cache_test.go
@@ -4,6 +4,7 @@ package cache
 
 import (
 	"fmt"
+	"math"
 	"math/rand/v2"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):
LGTM

PR #1872, opened 2026-05-05
[Router] Implement TestConfigPathLoading test for Redis config file loading

xiaotian-yu:
## Purpose

- **What does this PR change?**
  Implements the skipped `TestConfigPathLoading` test that validates external Redis configuration file loading functionality. Previously the test was marked with `t.Skip("TODO: Implement config file loading test")`.

- **Why is this change needed?**
  The TODO comment left incomplete test coverage for the `loadRedisStoreConfig()` function. This implementation ensures the critical config file loading and YAML parsing paths are properly tested.

- **Which module(s) does this affect?**
  `[Router]` - `pkg/responsestore` module

## Test Plan

- Run the new test suite with: `cd src/semantic-router && go test ./pkg/responsestore -v -run TestConfigPathLoading`
- Both test functions validate:
  1. **TestConfigPathLoading**: Successful loading of a valid YAML config file and verification of all parsed fields (address, DB, pool settings, timeouts, key prefix)
  2. **TestConfigPathLoadingError**: Error handling for two failure cases:
     - Non-existent config file 
     - Invalid YAML syntax
- This validation is sufficient because it covers:
  - Happy path: successful file I/O and YAML unmarshaling
  - Error paths: file not found and malformed YAML
  - All key configuration fields are verified after loading

## Test Result
image

PR #1875, opened 2026-05-06
[CLI] Add unit coverage for python config file loader follow-up to #1872

xiaotian-yu:
## Purpose

- This PR is a follow-up to #1872.
- It adds direct unit coverage for Python-side config file loading paths under `src/vllm-sr/cli/`.
- The new tests cover:
  - `parse_user_config(...)`
  - `load_config_file(...)`
  - `find_config_file(...)`
  - `load_profile_values(...)`
- The new coverage focuses on missing files, invalid YAML, empty config content, config discovery behavior, and named Helm profile values loading.
- Which module(s) does this affect? `CLI`
- No additional Valkey test was added because the Valkey code paths in Semantic Router do not load a separate config file from disk. `ValkeyCache` and `ValkeyStore` consume inline Valkey config from the main router config loader, so there is no separate Valkey `config_path`-style loader to exercise here.

## Test Plan

- Run:
  - `pytest src/vllm-sr/tests/test_config_file_loaders.py`
- Review the new test cases in:
  - `src/vllm-sr/tests/test_config_file_loaders.py`
  - These tests directly exercise the Python loader functions responsible for reading config files from disk and cover both expected success cases and the main failure modes.

## Test Result

- Result:
  - 
image

  - Test execution emits pre-existing Pydantic v2 deprecation warnings from `src/vllm-sr/cli/models.py` related to class-based `Config`.
  - Those warnings are not introduced by this PR and do not affect the test results.

### DIFF
--- a/src/vllm-sr/tests/test_config_file_loaders.py
+++ b/src/vllm-sr/tests/test_config_file_loaders.py
@@ -1,0 +1,162 @@
+import importlib
+import sys
+from pathlib import Path
+
+import pytest
+import yaml
+
+PROJECT_ROOT = Path(__file__).resolve().parents[1]
+if str(PROJECT_ROOT) not in sys.path:
+    sys.path.insert(0, str(PROJECT_ROOT))
+
+config_translator = importlib.import_module("cli.config_translator")
+parser = importlib.import_module("cli.parser")
+utils = importlib.import_module("cli.utils")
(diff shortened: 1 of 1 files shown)


samzong:
Thanks @xiaotian-yu

/lgtm
/approve

PR #1877, opened 2026-05-06
[Dashboard] Honor TARGET_ROUTER_API_URL when frontend sends placeholder eval endpoint

### LINKED ISSUE #1876: bug: Dashboard evaluation always uses hardcoded localhost:8080 in containerized deployments — TARGET_ROUTER_API_URL never applied
## Summary

When running the dashboard and router as separate containers (e.g. separate Kubernetes pods)
router-level evaluations always fail with `"API call failed"` on every sample. The evaluation
runner calls `http://localhost:8080/api/v1/eval`, which is unreachable from the dashboard
container. `TARGET_ROUTER_API_URL` is correctly set but never applied.

## Environment

- Chart: `./deploy/helm/semantic-router`
- Dashboard and router deployed as separate pods in the same Kubernetes namespace
- `TARGET_ROUTER_API_URL` set to the in-cluster service DNS name

## Root cause

The bug spans two layers that interact badly:

**1. Frontend always sends a non-empty endpoint (`evaluationConfig.ts`)**

```ts
// dashboard/frontend/src/utils/evaluationConfig.ts
export const DEFAULT_ROUTER_EVAL_ENDPOINT = 'http://localhost:8080/api/v1/eval';

export function getDefaultEndpointForLevel(level: EvaluationLevel, envoyUrl?: string): string {
  if (level === 'router') {
    return DEFAULT_ROUTER_EVAL_ENDPOINT;  // always sent, never empty
  }
  ...
}
```

The frontend populates the `endpoint` field in the task-creation request with this hardcoded
value before sending it to the backend.

**2. Backend only applies `TARGET_ROUTER_API_URL` when the endpoint field is empty (`evaluation_create.go`)**

```go
func (h *EvaluationHandler) applyEvaluationCreateDefaults(cfg *models.EvaluationConfig) {
    if cfg.Endpoint == "" {                              // never true — frontend always sends localhost
        cfg.Endpoint = h.defaultEvaluationEndpoint(cfg.Level)
    }
}

func (h *EvaluationHandler) defaultEvaluationEndpoint(level models.EvaluationLevel) string {
    if level == models.LevelRouter {
        if h.routerAPIURL != "" {
            return strings.TrimSuffix(h.routerAPIURL, "/") + "/api/v1/eval"  // correct, but never reached
        }
        return "http://localhost:8080/api/v1/eval"
    }
    ...
}
```

`h.routerAPIURL` is correctly populated from `TARGET_ROUTER_API_URL` and
`defaultEvaluationEndpoint` correctly constructs the right URL from it — but
`applyEvaluationCreateDefaults` is gated on `cfg.Endpoint == ""` which is never true because
the frontend always sends the hardcoded localhost default.

The result: `TARGET_ROUTER_API_URL` has no effect on evaluation in any deployment where the
dashboard and router are not on the same host.

## Observed behaviour

All evaluation samples are skipped with `"reason": "API call failed"`. The evaluation completes
in a few seconds (connection refused, not a timeout):

```json
{
  "status": "skip",
  "reason": "API call failed",
  "actual": null,
  "expected": "business"
}
```

```json
"metadata": {
  "endpoint": "http://localhost:8080/api/v1/eval",
  "elapsed_time_seconds": 5.28
}
```

There is no UI field to override the endpoint — the evaluation form does not expose it, and the
backend's override path is unreachable.

### To Reproduce

1. Deploy the semantic-router and dashboard as **separate pods** in the same Kubernetes/OpenShift namespace.
2. Set `TARGET_ROUTER_API_URL` on the dashboard pod to the router's in-cluster service address, e.g. `http://semantic-router.<namespace>.svc.cluster.local:8080`.
3. Log into the dashboard and navigate to **System → Evaluation**.
4. Create a new evaluation task: select **Router** level, choose any dataset (e.g. MMLU-Pro), set a sample count, and proceed through the wizard.
5. Note that the **Endpoint** field on the review step shows `http://localhost:8080/api/v1/eval` and cannot be edited.
6. Run the task.

**Result:** All samples are skipped with `"reason": "API call failed"`. The task completes in a few seconds (instant connection-refused, not a timeout). Example output:

```json
{
  "status": "skip",
  "reason": "API call failed",
  "actual": null
}
```

```json
"metadata": {
  "endpoint": "http://localhost:8080/api/v1/eval",
  "elapsed_time_seconds": 5.28
}
```

### Expected behavior

When `TARGET_ROUTER_API_URL` is configured to a non-localhost address the evaluation runner should use that URL to construct the eval endpoint. A router-level evaluation task should target `{TARGET_ROUTER_API_URL}/api/v1/eval` — which is exactly what `defaultEvaluationEndpoint` already computes correctly — so that evaluations work in any containerized or multi-pod deployment.

### Affected layer

unsure

### Additional context

_No response_

SAY-5:
Closes #1876

## Purpose

- The dashboard frontend always populates the router-eval endpoint field with the hardcoded `http://localhost:8080/api/v1/eval` placeholder before posting to `/api/evaluation/tasks`. The backend's existing `if cfg.Endpoint == ""` guard in `applyEvaluationCreateDefaults` therefore never triggered, so `TARGET_ROUTER_API_URL` was silently dropped in any deployment where the dashboard and router run as separate pods.
- Treat the placeholder as "unset" when `h.routerAPIURL` is configured, so the existing `defaultEvaluationEndpoint` override path is actually reached. Explicit non-placeholder endpoints continue to round-trip unchanged.
- Affects `Dashboard` only (`dashboard/backend/handlers/evaluation_create.go`, plus tests).

## Test Plan

- `cd dashboard/backend && go test ./handlers/...`
- New regression `TestCreateTaskHandler_RouterPlaceholderEndpointReplacedByConfiguredOverride` posts the frontend placeholder with the harness's configured `routerAPIURL=http://router:8080` and asserts the persisted endpoint is rewritten to `http://router:8080/api/v1/eval`.
- New regression `TestCreateTaskHandler_RouterExplicitEndpointPreserved` asserts a user-provided non-placeholder endpoint is preserved.
- `go vet ./handlers/...` and `gofmt` are clean on the touched files.

## Test Result

- All handler tests pass locally (`ok github.com/vllm-project/semantic-router/dashboard/backend/handlers`), including the two new regression cases.
- The existing `TestCreateTaskHandler_RouterWithDomainReturns201` and `TestCreateTaskHandler_NormalizesDefaultDatasetsAndExtraKeys` still pass, confirming the change is backward-compatible for the on-host case (when `routerAPIURL` is empty the placeholder is preserved as before).
- No follow-up risks identified: the override only activates when (a) level is router, (b) `TARGET_ROUTER_API_URL` is set, and (c) the incoming endpoint exactly matches the well-known placeholder.

---
<details>

</details>

### DIFF
--- a/dashboard/backend/handlers/evaluation_create.go
+++ b/dashboard/backend/handlers/evaluation_create.go
@@ -114,11 +114,17 @@ func validateDatasetsForConfig(cfg models.EvaluationConfig) (string, int) {
 	return "", 0
 }
 
+// fallbackRouterEvalEndpoint is the placeholder router eval URL the dashboard
+// frontend pre-populates when no in-cluster router address is available. The
+// frontend always sends this value, so the backend treats it as "unset" when a
+// server-side TARGET_ROUTER_API_URL override is configured.
+const fallbackRouterEvalEndpoint = "http://localhost:8080/api/v1/eval"
+
 func (h *EvaluationHandler) applyEvaluationCreateDefaults(cfg *models.EvaluationConfig) {
 	if cfg.MaxSamples <= 0 {
 		cfg.MaxSamples = 50
 	}
-	if cfg.Endpoint == "" {
+	if cfg.Endpoint == "" || h.shouldOverrideEndpoint(cfg) {
 		cfg.Endpoint = h.defaultEvaluationEndpoint(cfg.Level)
 	}
 	if cfg.SamplesPerCat <= 0 {
@@ -129,12 +135,26 @@ func (h *EvaluationHandler) applyEvaluationCreateDefaults(cfg *models.Evaluation
 	}
 }
 
+// shouldOverrideEndpoint returns true when the incoming endpoint matches the
+// frontend placeholder default and a server-side override is configured. This
+// keeps TARGET_ROUTER_API_URL effective in containerized deployments where the
+// dashboard cannot reach localhost:8080.
+func (h *EvaluationHandler) shouldOverrideEndpoint(cfg *models.EvaluationConfig) bool {
+	if cfg.Level != models.LevelRouter {
+		return false
+	}
+	if h.routerAPIURL == "" {
+		return false
+	}
+	return cfg.Endpoint == fallbackRouterEvalEndpoint
+}
+
 func (h *EvaluationHandler) defaultEvaluationEndpoint(level models.EvaluationLevel) string {
 	if level == models.LevelRouter {
 		if h.routerAPIURL != "" {
 			return strings.TrimSuffix(h.routerAPIURL, "/") + "/api/v1/eval"
 		}
-		return "http://localhost:8080/api/v1/eval"
+		return fallbackRouterEvalEndpoint
 	}
 	if h.envoyURL != "" {
 		return h.envoyURL
--- a/dashboard/backend/handlers/evaluation_test.go
+++ b/dashboard/backend/handlers/evaluation_test.go
@@ -145,6 +145,73 @@ func TestCreateTaskHandler_RouterWithDomainReturns201(t *testing.T) {
 	}
 }
 
+func TestCreateTaskHandler_RouterPlaceholderEndpointReplacedByConfiguredOverride(t *testing.T) {
+	t.Parallel()
+	handler := newTestEvaluationHandler(t)
+	body := map[string]interface{}{
+		"name": "signal-eval-override",
+		"config": map[string]interface{}{
+			"level":      "router",
+			"dimensions": []string{"domain"},
+			// Frontend always sends this hardcoded placeholder; the backend must
+			// substitute the configured TARGET_ROUTER_API_URL when present so the
+			// dashboard can reach the router across pods.
+			"endpoint":    "http://localhost:8080/api/v1/eval",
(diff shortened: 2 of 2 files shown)


dashboard/backend/handlers/evaluation_create.go
```diff
@@ -129,12 +135,26 @@ func (h *EvaluationHandler) applyEvaluationCreateDefaults(cfg *models.Evaluation
...
+// keeps TARGET_ROUTER_API_URL effective in containerized deployments where the
+// dashboard cannot reach localhost:8080.
+func (h *EvaluationHandler) shouldOverrideEndpoint(cfg *models.EvaluationConfig) bool {
+	if cfg.Level != models.LevelRouter {
+		return false
+	}
+	if h.routerAPIURL == "" {
+		return false
+	}
+	return cfg.Endpoint == fallbackRouterEvalEndpoint
```
samzong, line 149: This couples backend defaulting to a frontend placeholder string. seems a bit strange.

samzong (commented):
Thanks @SAY-5! I’d avoid coupling the backend logic to a frontend placeholder value.

rootfs (approved):

PR #1878, opened 2026-05-06
[Router] Add x-vsr-skip-processing header for full extproc passthrough

### LINKED ISSUE #1808: feature:  Support conditional (header-based) passthrough for extproc to enable selective execution with Envoy AI Gateway
## Description

When using **Semantic Router (extproc)** together with **Envoy AI Gateway**, there are challenges in achieving both:

1. Correct **filter ordering** (Semantic Router should run before AI Gateway)
2. **Selective execution** of Semantic Router on specific routes (semantic router runs on selective routes, not on all routes that runs envoy ai gateway extproc)

---

## Current Behavior

There are two ways to integrate Semantic Router:

### 1. Using `EnvoyPatchPolicy` (listener-level)
- ✅ Allows controlling ordering (can place Semantic Router at index 0 → runs before AI Gateway)
- ❌ Applies to all routes → no way to selectively enable/disable per route

### 2. Using `EnvoyExtensionPolicy` (route-level)
- ✅ Allows route-level control
- ❌ Ordering issue → Envoy AI Gateway extproc is always injected first via extension server

---

## Additional Context (AI Gateway Routing)

Envoy AI Gateway typically creates two routes:

- **Rule 1:** `pathPrefix "/" + header match (x-ai-eg-model: <model>)`
- **Rule 2:** `pathPrefix "/"` (default landing route)

### Request Flow

1. Request initially matches **Rule 2**
2. AI Gateway extproc adds `x-ai-eg-model` header
3. Request is re-routed to **Rule 1**

However, **Envoy determines the filter chain based on the initially matched route**, and this does not change after re-routing.

As a result:
- Any route-level extproc (e.g., Semantic Router via patch policy) effectively attaches to **Rule 2**
- This causes Semantic Router to run for **all AI Gateway traffic**, even when not desired

---

## Problem

We currently cannot achieve both:
- Correct ordering (**Semantic Router before AI Gateway**)
- Selective execution (**only on specific routes**)

---

## Proposed Solution

Introduce a **header-based passthrough mechanism** in Semantic Router:

- Semantic Router checks for a configurable header (e.g., `x-sr-skip: true`)
- If present → **skip decision chains / processing**
- If absent → proceed normally

This would allow:
- Installing Semantic Router at listener level (ensuring correct ordering)
- Using upstream systems (e.g., ext_auth, routing logic) to control execution behavior per request

---

## Example Use Case

- External models → run Semantic Router (PII detection, routing logic, etc.)
- Internal models → skip Semantic Router processing

---

## Alternatives Considered

- Route-level attachment via `EnvoyExtensionPolicy` → fails due to filter chain being fixed on initial route match
- Reordering extproc via extension server → solves ordering but not selectivity (filter chain is fixed on rule 2)

---

## Questions

1. Would a **header-based passthrough mechanism** be acceptable for Semantic Router?
2. Is there an existing way to achieve selective execution in this setup that I might have missed?

---

Happy to contribute a PR if this approach makes sense. Would appreciate feedback!

siddharth1036:
Closes #1808

## Purpose

- Adds a new request header `x-vsr-skip-processing` that, when set to `true`, makes the extproc act as a complete no-op across every Envoy callback (request headers, request body, response headers, response body).
- Enables the AI Gateway interop pattern from #1808: the router can be installed at the listener level (so it consistently runs before the AI Gateway extproc) while upstream filters (ext_authz, AI Gateway extension server, route-level filters) decide per request whether the router runs.
- Affects: `Router`

## Test Plan

- `make agent-ci-gate CHANGED_FILES="<changed paths>"`
- `make test-semantic-router`
- New unit tests in `pkg/extproc/req_filter_skip_processing_test.go` cover header detection (case-insensitive name and value, whitespace, false/empty), validation bypass, streaming-mode preservation, and short-circuit at request-body, response-header, and response-body phases.
- Header constant test added in `pkg/headers/headers_test.go`.

## Test Result

- `make agent-ci-gate` passed (47 packages green, zero `FAIL`).
- `make test-semantic-router` passed end-to-end.
- `pre-commit run --files <changed>` passed (`go fmt`, `go lint`, supply-chain scan, agent changed-files lint).
- 13 new skip-processing unit tests all pass; existing extproc/header tests unaffected.
- No follow-up risks identified. Behavior is opt-in via header, defaults remain unchanged.


rootfs:
@siddharth1036 what if this header is misused by the initiator?

siddharth1036:
> @siddharth1036 what if this header is misused by the initiator?

@rootfs yes that's a valid concern but to tackle the issue I couldn't think of a cleaner approach, can you suggest any other approach, I am happy to try any other solutions as well?

rootfs:
> > @siddharth1036 what if this header is misused by the initiator?
> 
> @rootfs yes that's a valid concern but to tackle the issue I couldn't think of a cleaner approach, can you suggest any other approach, I am happy to try any other solutions as well?

@siddharth1036 let's make this header optional and disable it by default first. Ideally, it would be great to have gating mechanisms in place to selectively allow such passthrough.

siddharth1036:
> > > @siddharth1036 what if this header is misused by the initiator?
> > 
> > 
> > @rootfs yes that's a valid concern but to tackle the issue I couldn't think of a cleaner approach, can you suggest any other approach, I am happy to try any other solutions as well?
> 
> @siddharth1036 let's make this header optional and disable it by default first. Ideally, it would be great to have gating mechanisms in place to selectively allow such passthrough.

Okay sure, I will add a deployment flag to control the usage of this header which will be false by default. Could you elaborate on gating mechanism? I can maybe give that a thought and try as well

rootfs (approved):

PR #1880, opened 2026-05-06
[Operator] Expose queryModality on IntelligentRoute EmbeddingSignal CRD

shraderdm:
## Purpose

Closes the gap between the runtime config layer (which has supported per-rule `QueryModality` since the multimodal query-signals work landed in #1867) and the IntelligentRoute CRD schema (which did not expose the field). Without this, CRD-driven deployments cannot declare image-modality or audio-modality embedding rules: kube admission rejects `queryModality` as an unknown field, and the canonical config converter in `src/semantic-router/pkg/k8s/converter.go` only translated `name`, `threshold`, `candidates`, and `aggregationMethod` into `config.EmbeddingRule`.

**Affected module:** `Operator`

**Follows #1867 (merged 2026-05-05) and #1868 (merged 2026-05-05).** #1867 added the runtime API and validator. #1868 wired the request path. This PR closes the CRD schema lag so multimodal embedding rules are declarable via `IntelligentRoute` CR, not only via file-based config.

## Diff shape

**Schema changes (1 file):**

- `src/semantic-router/pkg/apis/vllm.ai/v1alpha1/types_route.go` - new `QueryModality string` field on `EmbeddingSignal` with kubebuilder enum `text;image;audio`. Optional. Omitting the field is equivalent to `text` via `config.EmbeddingRule.EffectiveQueryModality()`, so legacy CRDs are unaffected.

**Converter changes (1 file):**

- `src/semantic-router/pkg/k8s/converter.go` - `convertSignals` now passes `signal.QueryModality` through into `config.EmbeddingRule.QueryModality`.

**Generated artifacts (2 files):**

- `deploy/kubernetes/crds/vllm.ai_intelligentroutes.yaml`, `deploy/helm/semantic-router/crds/vllm.ai_intelligentroutes.yaml` - regenerated via `make generate-api`.

**Test changes (3 files added):**

- `src/semantic-router/pkg/k8s/converter_embedding_modality_test.go` - new focused unit test `TestConvertSignals_EmbeddingQueryModality` covering all four cases (omitted, text, image, audio) plus a separate test that the omitted case resolves to `QueryModalityText` via `EffectiveQueryModality`.
- `src/semantic-router/pkg/k8s/testdata/input/17-embedding-multimodal.yaml` - testdata fixture exercising rules with `queryModality: text`, `queryModality: image`, `queryModality: audio`, and the omitted form, plus matching decisions.
[...]

PR #1895, opened 2026-05-09
[Operator] Validate embedding modality contracts on IntelligentRoute reconcile

shraderdm:
## Summary

Mirror the file-config embedding-modality validation in the Kubernetes reconcile path. Closes the gap identified in #1880's review where the operator could admit and mark Ready a CRD config the file-config validator would reject.

## Background

PR #1880 added `queryModality` to the IntelligentRoute EmbeddingSignal CRD. The runtime contract (in `validator_embedding.go`) is stricter than the new CRD surface:

- `query_modality=image` is only allowed when `global.model_catalog.embeddings.semantic.embedding_config.model_type=multimodal`
- `query_modality=audio` is rejected as not yet supported (the audio FFI is not exposed by candle-binding)

`validateConfigStructure` early-returns for `ConfigSourceKubernetes`, so `validateEmbeddingContracts` never runs for CRD-loaded configs. The reconciler still applies the config and marks the CRs Ready.

## Changes

- Export `ValidateEmbeddingContracts(cfg *RouterConfig) error` in `validator_embedding.go` as a thin wrapper around the existing private `validateEmbeddingContracts`. The signature mirrors the per-family validators in `validator.go`'s dispatch table so the K8s call site reads the same as the file-config dispatch.
- Call it from `Reconciler.validateAndUpdate` after `ParseYAMLBytes` returns and `ConfigSource` is set. On validation failure, mark pool and route `Ready=False, Reason=ValidationFailed` and return a `fmt.Errorf %w`-wrapped error.
- Reconciler-level integration test (`reconciler_embedding_modality_test.go`) drives a real `Reconciler.validateAndUpdate` call through controller-runtime's fake client. Six cases cover the validator branches: `AudioRejected`, `ImageWithoutMultimodalRejected`, `ImageWithMultimodalAccepted`, `TextOnlyAccepted`, `OmittedModalityAccepted` (the legacy/backward-compat path), and `UnknownModalityRejected`. Each case asserts both the returned error AND the persisted Ready conditions on both refetched CRs.

## Why option A (mirror) over option B (narrow CRD enum)

Option B is implementable. Dropping `audio` from the kubebuilder enum plus a CEL admission rule on IntelligentRoute that conditions `queryModality: image` on the operator's SemanticRouter declaring `embeddingConfig.modelType: multimodal` would express the cross-resource invariant at admission. The reasons option A is the better choice for this PR:

1. Single source of truth across config sources. The embedding-modality contract already lives in `validateEmbeddingRuleModalities`; the file-config path runs it via `validateConfigStructure`. This PR makes the K8s reconcile path call the same validator. Option B would fork enforcement across CEL admission (schema-side) and the Go validator (runtime-side), so the contract has to be kept in sync in two places when it evolves.
2. The CRD enum stays forward-compatible. When the audio FFI lands the validator widens; the schema does not need to migrate.
3. The cross-resource invariant (image rules require `model_type=multimodal`) is what option B's CEL piece would actually carry. The kubebuilder enum on `queryModality` (already added in #1880) handles the typo case at admission; the cross-resource piece is what would need separate CEL plumbing under option B. Option A covers both invariants in one Go validator with no new admission surface.
4. No new operator-side dependency. Option B would require a CEL ruleset maintained alongside the validator, with its own test surface and its own coverage gaps when the validator's rules change.

## Tracked debt

This PR closes the embedding slice of a broader gap. `validateConfigStructure` early-returns for `ConfigSource == ConfigSourceKubernetes` and skips every other family validator (domain, structure, reask, projection, knowledge_base, conversation, decision, modality, model_selection, advanced_tool_filtering). The remaining slices are tracked in [`docs/agent/tech-debt/td-040-reconcile-path-skips-family-validators.md`](docs/agent/tech-debt/td-040-reconcile-path-skips-family-validators.md).

## Test plan

- [x] `make agent-validate` clean
- [x] `make agent-lint CHANGED_FILES=...` clean
- [x] `make agent-ci-lint CHANGED_FILES=...` clean (the gate that bit prior PRs)
- [x] `make agent-feature-gate ENV=cpu CHANGED_FILES=...` clean end-to-end (Docker-backed; behavior-visible since operator admits configs differently)
- [x] New `TestValidateEmbeddingContracts_ExportedWrapperMatchesInternal` passes
- [x] New `TestReconcileEmbeddingModalityValidation` passes (6 cases) and exercises a real `Reconciler.validateAndUpdate` call via controller-runtime's fake client
- [x] Verified the load-bearing property by branch surgery: removing the new validation block from `reconciler.go` causes the rejection cases to fail; restored block, all cases pass again
- [ ] `make agent-pr-gate` - blocked locally by an unrelated upstream issue: the precommit-local Docker image at `ghcr.io/vllm-project/semantic-router/precommit:latest` is missing `python3-venv`, so `make agent-bootstrap` inside the container fails at `python3 -m venv`. Companion fix opened as #1894 (one-line apt addition). Once #1894 lands and the image is rebuilt, this gate runs locally; CI's `Test And Build` workflow runs `make test` directly outside that container, so the bug is invisible there
- [ ] `make test-and-build-local` - flaky on `pkg/memory.TestGenerateMemoryID`, which calls `generateMemoryID()` twice in immediate succession and asserts inequality; the function uses a nanosecond timestamp and collides under fast clocks. Pre-existing flake, unrelated to this PR's surface

## Related

- Follow-up to #1880 (review feedback from @rootfs)
- Companion to #1881 (e2e re-add for image-modality routing)
- Tracked debt: TD040 (reconcile-path family-validator gap, embedding slice closed here)

### DIFF
--- a/docs/agent/tech-debt/README.md
+++ b/docs/agent/tech-debt/README.md
@@ -78,6 +78,7 @@ Keep the numeric index unique within `docs/agent/tech-debt/`.
 - [TD037 Dev Integration Environment Ownership and Shared-Suite Topology Still Diverge Across CLI, Kind, and CI](td-037-dev-integration-env-ownership-and-shared-suite-topology.md)
 - [TD038 Custom Chat Completions Structs Duplicate Official OpenAI SDK Types](td-038-custom-chat-completions-structs.md)
 - [TD039 Control Planes Still Depend on Router Runtime Internals Instead of Contract-Owned Seams](td-039-control-plane-contract-ownership-collapse.md)
+- [TD040 Kubernetes Reconcile Path Skips File-Config Family Validators](td-040-reconcile-path-skips-family-validators.md)
 
 ## Architecture Review Coverage Map
 
--- a/docs/agent/tech-debt/td-040-reconcile-path-skips-family-validators.md
+++ b/docs/agent/tech-debt/td-040-reconcile-path-skips-family-validators.md
@@ -42,3 +42,16 @@ Either way, the embedding-modality wrapper exported by the slice PR remains vali
 - `Reconciler.validateAndUpdate` no longer needs to call individual exported validators (or, equivalently, the exported validators become internal again because the dispatch covers them).
 - The test pattern in `reconciler_embedding_modality_test.go` is replaced by a single dispatch-level test that asserts every K8s-safe contract fires on the reconcile path, with the per-family unit tests remaining the source of truth for branch coverage.
 - Validators currently skipped on the K8s path that need to be re-enabled: `validateDomainContracts`, `validateStructureContracts`, `validateReaskContracts`, `validateProjectionContracts`, `validateKnowledgeBaseContracts`, `validateConversationContracts`, `validateDecisionContracts` (insofar as it applies to non-CRD-shape concerns), `validateModalityContracts`, `validateModelSelectionConfig`, `validateAdvancedToolFilteringConfig`. Per-family applicability should be confirmed during the unification design pass.
+
+## Recipe for closing another slice (template established by the embedding slice)
+
+The unification path described in Desired End State above is the preferred end state. Until that lands, slices can be closed incrementally with the mirror-per-slice pattern this debt entry's first slice established. To close another family validator's slice on the K8s reconcile path:
+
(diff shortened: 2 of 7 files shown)


rootfs:
My Copilot code review limit reached, need to get another review bot, stay tuned.

rootfs:
@shraderdm can you add a follow up PR to an E2E coverage too?

rootfs (approved):

shraderdm:
@rootfs Thanks for merging! Yes, on it - folding the negative E2E into #1881 since it's the same profile + fixture surface. Three rejection cases (audio, image-without-multimodal-model, unknown-modality) plus one positive regression case to guard the happy path. Will tag you here once #1881 is ready.

PR #1918, opened 2026-05-19
fix: blocker of scan_malicious_code.py by antivirus

### LINKED ISSUE #1917: bug: scan_malicious_code.py is treated as malicious by Antivirus tool
scan_malicious_code.py is treated as malicious by Antivirus tool 

Image

AayushSaini101:
Closes #1917

## Purpose

1. Refactors tools/security/scan_malicious_code.py to use base64-encoded regex signatures instead of plaintext attack patterns, so antivirus/static scanners stop blocking the file.
2. Restores the regex fallback used by supply-chain CI (security-scan.yml, make security-scan) with the same detection behavior.
3. Module: CI/Build

## Test Plan

1. Scanner exits 0 at --fail-on HIGH; ruff passes.
2. Confirm AV/CI no longer blocks the file on push.

### DIFF
--- a/tools/agent/structure-rules.yaml
+++ b/tools/agent/structure-rules.yaml
@@ -213,6 +213,7 @@ legacy_hotspots:
   - paths:
       - tools/security/ast_security_scanner.py
       - tools/security/scan_malicious_code.py
+      - tools/security/scan_sensitive_paths.toml
     file_checks: relaxed
     function_checks: relaxed
   - paths:
--- a/tools/security/scan_malicious_code.py
+++ b/tools/security/scan_malicious_code.py
@@ -23,10 +23,11 @@ import os
 import re
 import stat
 import sys
-import tomllib
 from dataclasses import dataclass, field
 from functools import lru_cache
 from pathlib import Path
+
+import tomllib
 
 SEVERITY_ORDER = {"CRITICAL": 0, "HIGH": 1, "MEDIUM": 2, "LOW": 3}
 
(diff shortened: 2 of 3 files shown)


AayushSaini101:
cc: @rootfs @Xunzhuo

rootfs:
@AayushSaini101 thanks for the base64 idea. This fixes the CI issue but also makes the review harder :D

Would it work to have the sensitive paths in a config file and make the scanner to use the config file instead of hardcoded values? Hopefully CI will be happy with that setup too.

AayushSaini101:
> @AayushSaini101 thanks for the base64 idea. This fixes the CI issue but also makes the review harder :D
> 
> Would it work to have the sensitive paths in a config file and make the scanner to use the config file instead of hardcoded values? Hopefully CI will be happy with that setup too.

will try thanks @rootfs

AayushSaini101:
> @AayushSaini101 thanks for the base64 idea. This fixes the CI issue but also makes the review harder :D
> 
> Would it work to have the sensitive paths in a config file and make the scanner to use the config file instead of hardcoded values? Hopefully CI will be happy with that setup too.

@rootfs i tried to extract the sensitive path in a config file, but sill this is detecting as a malicious file considering older scan_malcious_code, now i have used base64 encoding and  seprated the path in the config thanks

rootfs:
@AayushSaini101 thanks, please fix the py lint issue and it is ready to go

AayushSaini101:
> @AayushSaini101 thanks, please fix the py lint issue and it is ready to go

Done thanks : )

Xunzhuo (approved):
LGTM

PR #1940, opened 2026-05-25
feat(anthropic): outbound emitter for non-streaming responses

siloteemu:
> **RFC**: #1946 — design rationale, alternatives considered, decisions invited from maintainers

Fourth PR in the native Anthropic `/v1/messages` ingress series (tracking
#1404). Stacks on #1939 (PR3). #1937 and #1938 have merged.

## What

Symmetric inverse of PR2's `ParseAnthropicRequest`: takes the OpenAI-shaped
response that flows out of the upstream backend and renders it as an Anthropic
Messages API response body shape that `/v1/messages` clients (Claude Code,
anthropic-sdk-go) expect.

After PR4 lands, the non-streaming request/response round-trip for native
Anthropic ingress is feature-complete. Streaming is PR5.

## Commits

1. **`[Router] add response-side usage and stop-reason fields to IRExtensions`** —
   new `AnthropicStopReason`, `AnthropicStopSequence`, `CacheCreationInputTokens`,
   `CacheReadInputTokens`, `Ephemeral5mInputTokens`, `Ephemeral1hInputTokens`,
   `ServerToolUseCounts` fields on `IRExtensions`. Pure type additions; nil-safe
   setters; no behavior. Lands first so the next commit's refactor compiles.
2. **`[Router] extract shared Anthropic response-shape helpers from client.go`** —
   refactor only. Extracts helpers that both the existing `ToOpenAIResponseBody`
   (converts Anthropic-backend responses to OpenAI shape on the way out) and
   the new outbound emitter share. Preserves the `default:` silent-drop of
   thinking / citations / server_tool_use into `IRExtensions` so they survive
   the round-trip into the outbound emitter.
3. **`[Router] add Anthropic outbound emitter for non-streaming responses`** —
   the heavy commit. `EmitAnthropicResponse(*openai.ChatCompletion, *ir.IRExtensions)`
   builds an Anthropic Messages response body: content blocks (text, tool_use,
   thinking placeholder), `stop_reason` mapping (OpenAI `finish_reason` →
   Anthropic vocabulary), `usage` fields (`input_tokens`, `output_tokens`,
   `cache_creation_input_tokens`, `cache_read_input_tokens`, `cache_creation`
   breakdown when populated). `EmitAnthropicError` wraps translation failures
   in the Anthropic error envelope shape. Pure functions — no extproc wiring
   in this commit.
4. **`[Router] route Anthropic clients through outbound emitter on response`** —
[...]

PR #1939, opened 2026-05-25
feat(anthropic): header pass-through, session-id mirroring, and lossiness response headers

siloteemu:
> **RFC**: #1946 — design rationale, alternatives considered, decisions invited from maintainers

Third PR in the native Anthropic `/v1/messages` ingress series (tracking #1404).
Now that #1937 and #1938 have merged, this is the next PR in the series.

## What

Wires the ergonomic edges of native Anthropic ingress: pass Anthropic-specific
client headers through to the upstream, mirror `x-claude-code-session-id` into
the existing session-aware machinery, and surface lossy-translation events
(captured by the PR2 parser) as structured response headers so clients can
detect when their request was downgraded.

## Commits

1. **`[Router] enumerate translation-warning reasons and add metric`** — typed
   `WarningReason` vocabulary in `pkg/ir/warning_reasons.go` plus a
   `vsr_translation_lossy_total` counter in
   `pkg/observability/metrics/translation_metrics.go`. Pure additive, no
   behavior.
2. **`[Router] mirror x-claude-code-session-id into RequestContext.SessionID`** —
   slots the Claude Code per-conversation UUID into the existing session
   priority chain (between `x-session-id` and the message-fingerprint
   fallbacks). Documented full priority order in `docs/sessions.md`.
3. **`[Router] add header pass-through policy for Anthropic ingress`** —
   explicit DROP set, named capture into new `IRExtensions.Inbound*` fields,
   forwarding via the existing `ProviderProfile.ExtraHeaders` mechanism. No
   new top-level file — surgical additions to existing processors.
4. **`[Router] surface translation warnings and protocol markers on responses`** —
   emits three new response headers when the request went through translation:
   - `x-vsr-inbound-protocol` and `x-vsr-outbound-protocol` — machine-readable
     markers describing the wire-shape transformation
   - `x-vsr-lossiness-warnings` — semicolon-separated
     `severity;reason;field` triples, 4 KB capped with a
     `warnings_truncated` marker. Format is grep-friendly and JSON-free.
   Gated to NOT emit on cache-hit responses (the cached body was translated
   under different inputs; emitting fresh markers would mislead).

## Notable design decisions

- **Header format**: semicolon-separated triples, not a JSON blob.
  Operators tail logs; a 4 KB cap with explicit truncation marker keeps the
  header bounded while preserving the worst-case warning load.
- **Session-id priority** (documented in `docs/sessions.md`):
  `ResponseAPICtx.ConversationID > x-session-id > x-claude-code-session-id >
[...]

PR #1937, opened 2026-05-25
feat(extproc): detect ClientProtocol from :path header

siloteemu:
> **RFC**: #1946 — design rationale, alternatives considered, decisions invited from maintainers

Foundational change for native Anthropic `/v1/messages` ingress. Adds a typed
`ClientProtocol` field on `RequestContext`, populated from envoy's `:path`
pseudo-header before any routing decision runs. No body parsing, no routing
change, no request rewriting — just the protocol-detection primitive that
downstream changes will branch on.

Adopts and refines @samzong's previously open #1473, with credit preserved via
Co-authored-by trailer.

## Commits

1. **`[Router] detect Anthropic client protocol from :path header`** (samzong,
   cherry-picked) — adds `ClientProtocolAnthropic` constant, `ClientProtocol`
   field on `RequestContext`, `detectClientProtocol(path, ctx)` helper, tests.
2. **`[Router] tighten ClientProtocol detection contract`** — three
   review-feedback refinements:
   - Drop unused `ClientProtocolOpenAI` constant; the zero-value default
     already handles the OpenAI case. Will be re-introduced when an explicit
     consumer lands.
   - Demote the per-request "detected Anthropic client protocol" log from
     `Infof` to `Debugf` to avoid noise on a hot path.
   - Expand the `detectClientProtocol` doc comment to make explicit that
     `HasPrefix("/v1/messages")` intentionally also matches the
     `/v1/messages/count_tokens` helper endpoint; add a test case documenting
     that intent.

## Why now

This is the first in a small series implementing native Anthropic ingress
(tracking issue #1404). Follow-up changes will add: the `/v1/messages` allow-list,
the Anthropic inbound parser, a protocol-neutral intermediate representation,
the outbound emitter, and Anthropic SSE response translation. Landing the
detection primitive first lets the rest branch on `ctx.ClientProtocol` without
each change re-inventing the dispatch.

## Test plan

- [x] `make agent-lint CHANGED_FILES=...` — golangci-lint, pre-commit hooks,
      supply-chain scan, reference-config test all clean
- [x] `go vet ./pkg/config/... ./pkg/extproc/...` clean
- [x] `go build ./pkg/config/... ./pkg/extproc/...` clean
- [x] DCO sign-off on both commits
- [ ] Upstream CI (`test-and-build`, `integration-test`, `build_pr (*)`) — runs
      once this PR is marked ready

## Related

- Picks up #1473 (open since 2026-03; rebased onto current main with credit)
- Tracking issue: #1404
[...]

PR #1926, opened 2026-05-21
[Router] add Anthropic streaming with OpenAI SSE translation

akshayv:
## Purpose

- Enable `stream: true` for models with `api_format: anthropic` by translating Anthropic SSE into OpenAI `chat.completion.chunk` SSE (text and tool-call deltas).
- Route Anthropic streaming through dedicated extproc helpers so the existing OpenAI streaming path (`processor_res_body_streaming.go`) stays unchanged.
- Add streaming request helpers (`WithStreamingRequestBody`, `BuildStreamingRequestHeaders`) and response translation in `pkg/anthropic/client_stream.go`.

**Why is this change needed?** Anthropic-backed routing previously rejected streaming requests and, when enabled experimentally, could break chunked responses by running SSE fragments through non-streaming JSON normalization.

**Which module(s) does this affect?** `Router`, `Docs`

### Design notes

- **Protocol translation** (`pkg/anthropic/client_stream.go`): Anthropic SSE events → OpenAI `chat.completion.chunk` bytes.
- **ExtProc wiring** (`processor_res_body_streaming_anthropic.go`): body mutation for translated SSE; reuses shared streaming accumulation (`StreamingChunks`, cache, metrics).
- **Per-request translator state**: `RequestContext.AnthropicStream` holds multi-chunk SSE mapping state; initialized when routing a streaming Anthropic request.
- **OpenAI backends**: unchanged passthrough path in `processor_res_body_streaming.go` (no format translation).

## Test Plan

```bash
make agent-lint CHANGED_FILES="src/semantic-router/pkg/anthropic/client.go,src/semantic-router/pkg/anthropic/client_stream.go,src/semantic-router/pkg/anthropic/client_stream_test.go,src/semantic-router/pkg/extproc/processor_core.go,src/semantic-router/pkg/extproc/processor_req_body_anthropic.go,src/semantic-router/pkg/extproc/processor_req_body_anthropic_test.go,src/semantic-router/pkg/extproc/processor_res_body.go,src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic.go,src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_test.go,src/semantic-router/pkg/extproc/request_context.go,website/docs/api/router.md"

cd src/semantic-router
go test ./pkg/anthropic/... ./pkg/extproc/... -run 'Anthropic|TransformSSE' -count=1
```

Manual (router on :8899, Anthropic/Vertex-backed model):

```bash
[...]

PR #1941, opened 2026-05-25
feat(anthropic): SSE streaming emitter + thinking-delta drop fixes

siloteemu:
> **RFC**: #1946 — design rationale, alternatives considered, decisions invited from maintainers

Fifth and final PR in the native Anthropic `/v1/messages` ingress series
(tracking #1404). Stacks on #1940 → #1939. #1937 and #1938 have merged.

After PR5 lands, the full request/response round-trip for native Anthropic
ingress (non-streaming AND streaming) is feature-complete.

## What

Symmetric inverse of #1926 (Anthropic→OpenAI SSE translation). Converts an
OpenAI-shaped streaming response into the Anthropic SSE event sequence so
`/v1/messages` clients (Claude Code, anthropic-sdk-go) consume it without
their own translation layer.

Also fixes **four pre-existing silent drops** in the existing Anthropic→OpenAI
streaming path (`pkg/anthropic/client_stream.go` `default:` branch): thinking
content-block start, thinking_delta, signature_delta, and `event: error`. These
have been silently affecting any deployment streaming from an Anthropic
backend with Opus 4.7 adaptive thinking — clients see empty thinking blocks
and missing error surfaces. Fixed first so the rest of the PR builds on a
correct baseline.

## Commits

1. **`[Router] fix dropped Anthropic streaming events for thinking and errors`** —
   the bug fix. Adds typed handlers for `thinking` content-block start,
   `thinking_delta`, `signature_delta`, and `event: error` SSE events that the
   `default:` branch previously dropped. Adds optional `*ir.IRExtensions`
   parameter to `TransformSSEChunkToOpenAI` (existing caller passes nil;
   backward-compatible). Surfaces upstream errors as
   `finish_reason: "error"` + synthetic content delta carrying the message.
2. **`[Router] extend StreamState with OpenAI-to-Anthropic direction fields`** —
   reuses the existing `StreamState` struct from #1926; adds
   `NextBlockIndex`, `OpenTextBlockIndex`, `OpenThinkingBlockIdx`,
   `ToolIdxToBlockIndex`, `LastChunkAt`, `InitialUsage` (tool start emission is tracked via `ToolIdxToBlockIndex` presence, not a separate flag).
3. **`[Router] add Anthropic outbound SSE emitter`** — heavy commit. New
   `pkg/anthropic/sse_out.go` with `EmitAnthropicSSEChunk` and per-event
   helpers (`message_start`, `content_block_start/_delta/_stop` per text /
   thinking / tool_use, `message_delta`, `message_stop`). Pure function;
   no extproc wiring. Reuses `buildAnthropicUsage` and
   `mapOpenAIFinishReasonToAnthropic` from PR4's `outbound.go`.
4. **`[Router] route Anthropic clients through SSE emitter on streaming response`** —
   new dispatcher branch in `handleResponseBody`; tightens the existing #1926
   branch to fire only when the client is NOT Anthropic. New
   `handleAnthropicClientStreamingResponseBody` handles the "double-Anthropic"
   cell (Anthropic client + Anthropic backend). The handler always replaces
   the response body with the emitter's bytes (even when the emitter produced
   nothing for a given chunk) so the raw upstream Anthropic frames never leak
   through on top of our synthesized events. Regression test
   `TestHandleAnthropicClientStreamingResponseBody_SuppressesUpstreamTail`
   locks this in. (See "Bug-fix during E2E verification" below.)
5. **`[Router] add ping keepalive for Anthropic streaming responses`** —
   15-second ping cadence. Implemented as a chunk-handler-driven check
   (emit a ping if it's been ≥15s since the last event) rather than a
   per-request goroutine. Rationale: in Envoy ext_proc, bytes can only be
   emitted during response callbacks. A goroutine could enqueue but cannot
   actually push during true backend silence, so the goroutine's incremental
   value is small and the lifecycle complexity is real. Pure helper
   `maybeEmitPing` keeps the surface compatible if reviewers prefer a
   goroutine approach later.
6. **`[Router] propagate cache and stop-reason fields through streaming emitter`** —
   `ext.CacheReadInputTokens`, `AnthropicStopReason`, and `signature_delta`
   capture into `ThinkingSignatures` so the streaming path matches the
   non-streaming PR4 round-trip.

## Notable design decisions

- **Ping keepalive (chunk-driven, not goroutine)**: see commit 5 message above.
  This diverges from the original plan; documented in the commit message for
  reviewer engagement.
- **"Double-Anthropic" cell** (Anthropic client + Anthropic backend) resolved
  by branch ordering: PR5's `ClientProtocol == "anthropic"` branch fires
  first, then the existing #1926 branch is tightened to
  `APIFormat == Anthropic && ClientProtocol != "anthropic"`. One-line
  condition change but load-bearing; `TestHandleResponseBody_StreamingDispatchMatrix` covers
  all four cells.
- **Thinking emission**: structurally-correct events (`content_block_start`
  with `type: "thinking"`, `thinking_delta` events) but TODO placeholder
  on the user-visible text content, mirroring PR4's approach. One-line fix
  when upstream #1718 lands.
- **Anthropic-client streaming handler owns the wire unconditionally** (added
  in v2 after E2E verification, folded into commit 4): once
  `handleAnthropicClientStreamingResponseBody` runs, every response body
  callback returns a `BodyMutation` carrying the emitter's bytes — even when
  those bytes are empty. A `nil` BodyMutation would make Envoy forward the
  raw upstream chunk unmodified, which leaks Anthropic frames from the
  backend onto the wire on top of the events we synthesized.

## Bug-fix during E2E verification

A duplicate `message_stop` event was observed during the streaming E2E test
through the double-Anthropic cell. Symptom: client received two
`event: message_stop` frames at end-of-stream — the first was the emitter's
clean synthesized stop (emitted on the prior chunk's `message_delta` +
usage), the second was the raw upstream `message_stop` (visible by its
trailing-whitespace JSON formatting from the backend) passed through Envoy
unmodified. Root cause: the handler returned a `nil` BodyMutation when the
emitter produced zero bytes for the terminal chunk (which contained only
the upstream `message_stop` → `[DONE]` after translation), and Envoy then
forwarded the original upstream bytes. Fix folded into commit 4: always
replace the body, never return a nil mutation once we're in this handler.
Regression test added.

## Deferred / known follow-ups

- **Golden SSE fixture round-trip tests** (~12 fixture pairs): deferred —
  the extproc test binary cannot link locally (Rust CGO bindings absent),
  so end-to-end fixture tests would defer verification to PR review without
  local sign-off. Pure-function emitter tests in `sse_out_test.go` cover
  the state machine. Recommend filing as a follow-up.
- **E2E SDK contract test** (`e2e/testing/10-anthropic-sdk-contract.py`):
  also deferred — needs live infrastructure.
- **Goroutine-based ping** (if reviewers prefer): pure helper
  `maybeEmitPing` is the seam.

## What's next (sibling PR, not in this series)

A small **sibling bug-fix PR** for the pre-existing `pkg/anthropic/client.go`
non-streaming defects (`tool_result.is_error` silently dropped; `tool_result`
array content flattened to empty string). Separate from this series — the
client.go defects are in a different code path from this PR's streaming work.

## Test plan

- [x] `make agent-lint` clean on all 7 files
- [x] `go vet`/`go build`/`go test` clean on `pkg/anthropic/...` (the bulk
      of PR5; pure-function SSE emitter state-machine tests)
- [x] DCO sign-off on all 6 commits, SSH-signed
- [x] Per-commit trajectory verified
- [x] Bug-fix regression tests in `client_stream_test.go` lock the
      4-event-drop fixes
- [x] `sse_out_test.go` (~285 LOC) exercises `EmitAnthropicSSEChunk` state
      machine: message_start emitted once, content_block_start/_stop per
      block, tool_use input_json_delta partial-args accumulation,
      message_delta with stop_reason, ping cadence
- [x] E2E streaming verification against a real Anthropic-shape backend
      (vertex against an Anthropic-shape backend): exactly one `message_stop` event per
      stream, spec-correct ordering, non-streaming round-trip unaffected.
      Surfaced and fixed the duplicate-message_stop bug above.
- [x] E2E test added: `e2e/testcases/anthropic_messages_streaming.go` (`anthropic-messages-streaming`) — consumes SSE from streaming `/v1/messages`, asserts `message_start` → `content_block_start` → `content_block_delta`+ → `content_block_stop` → `message_delta` → `message_stop` event ordering with exactly one terminal `message_stop`
- [ ] Upstream CI — runs once this PR is marked ready

## Related

- Stacks on #1940, #1939
- Builds on #1938 (PR2: IR + inbound parser, now merged)
- Builds on #1937 (PR1: ClientProtocol detection, now merged)
- Tracking issue: #1404
- Symmetric inverse of #1926 (Anthropic→OpenAI SSE)
- Adjacent: #1718 (Anthropic thinking response, draft — would supply the
  one-line fix for the thinking-block text placeholder)

### DIFF
--- a/src/semantic-router/pkg/anthropic/client_stream.go
+++ b/src/semantic-router/pkg/anthropic/client_stream.go
@@ -10,30 +10,50 @@ import (
 	"github.com/anthropics/anthropic-sdk-go"
 	"github.com/openai/openai-go"
 	"github.com/tidwall/sjson"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/ir"
 )
 
 // StreamState tracks Anthropic SSE events while translating to OpenAI SSE.
 type StreamState struct {
-	MessageID           string
-	Created             int64
-	RoleSent            bool
-	BlockIndexToToolIdx map[int64]int
-	NextToolIdx         int
+	MessageID                  string
+	Created                    int64
+	RoleSent                   bool
+	BlockIndexToToolIdx        map[int64]int
+	NextToolIdx                int
+	BlockIndexToThinkingActive map[int64]bool
 }
 
 // NewStreamState returns initialized stream translation state.
 func NewStreamState() *StreamState {
 	return &StreamState{
-		BlockIndexToToolIdx: make(map[int64]int),
+		BlockIndexToToolIdx:        make(map[int64]int),
+		BlockIndexToThinkingActive: make(map[int64]bool),
 	}
+}
+
+// extractedSSELine carries one parsed `data:` payload along with the most
+// recent `event:` header above it. The header lets us route Anthropic
+// `event: error` lines into the error arm of transformStreamEvent — those
+// lines unmarshal as `{"type":"error",...}` which the switch otherwise
+// drops silently.
+type extractedSSELine struct {
+	Data    []byte
+	IsError bool
 }
 
 // TransformSSEChunkToOpenAI converts Anthropic SSE bytes from Envoy into OpenAI SSE.
 // streamDone is true when message_stop is observed.
+//
+// ext is optional; when non-nil, signature_delta events are captured into
+// ext.ThinkingSignatures so the symmetric outbound emitter can replay them
+// on multi-turn requests. Pass nil for the existing OpenAI-client cell
+// where signatures have no downstream consumer.
 func TransformSSEChunkToOpenAI(
 	anthropicChunk []byte,
 	state *StreamState,
 	model string,
+	ext *ir.IRExtensions,
 ) ([]byte, bool, error) {
 	if state == nil {
 		state = NewStreamState()
@@ -41,12 +61,23 @@ func TransformSSEChunkToOpenAI(
 
 	var out bytes.Buffer
 	streamDone := false
-	for _, data := range extractSSEDataLines(anthropicChunk) {
-		var event anthropic.MessageStreamEventUnion
-		if err := json.Unmarshal(data, &event); err != nil {
+	for _, line := range extractSSEDataLines(anthropicChunk) {
+		if line.IsError {
+			chunks, err := handleErrorEvent(line.Data, state, model)
+			if err != nil {
+				return nil, false, err
+			}
+			for _, chunk := range chunks {
+				out.Write(formatOpenAISSELine(chunk))
+			}
 			continue
 		}
-		chunks, done, err := transformStreamEvent(event, state, model)
+
+		var event anthropic.MessageStreamEventUnion
+		if err := json.Unmarshal(line.Data, &event); err != nil {
+			continue
+		}
+		chunks, done, err := transformStreamEvent(event, state, model, ext)
 		if err != nil {
 			return nil, false, err
 		}
@@ -65,11 +96,24 @@ func TransformSSEChunkToOpenAI(
 	return out.Bytes(), streamDone, nil
 }
 
-func extractSSEDataLines(chunk []byte) [][]byte {
-	var lines [][]byte
-	for _, line := range bytes.Split(chunk, []byte("\n")) {
-		line = bytes.TrimSpace(line)
+// extractSSEDataLines walks the Anthropic SSE framing in chunk and yields
+// each `data:` payload along with whether the preceding `event:` header
+// flagged it as an error event. Lines that begin with `{` (some
+// transports omit the `data:` prefix) are surfaced as ordinary payloads.
+func extractSSEDataLines(chunk []byte) []extractedSSELine {
+	var lines []extractedSSELine
+	currentEvent := ""
+	for _, raw := range bytes.Split(chunk, []byte("\n")) {
+		line := bytes.TrimSpace(raw)
 		if len(line) == 0 {
+			// Blank line terminates the current SSE event; reset the
+			// per-event header tracker so the next data: line is not
+			// mis-classified as belonging to a stale event header.
+			currentEvent = ""
+			continue
+		}
+		if bytes.HasPrefix(line, []byte("event:")) {
+			currentEvent = string(bytes.TrimSpace(bytes.TrimPrefix(line, []byte("event:"))))
 			continue
 		}
 		if bytes.HasPrefix(line, []byte("data:")) {
@@ -77,11 +121,17 @@ func extractSSEDataLines(chunk []byte) [][]byte {
 			if len(payload) == 0 || bytes.Equal(payload, []byte("[DONE]")) {
 				continue
 			}
-			lines = append(lines, payload)
+			lines = append(lines, extractedSSELine{
+				Data:    payload,
+				IsError: currentEvent == "error",
+			})
 			continue
 		}
 		if line[0] == '{' {
-			lines = append(lines, line)
+			lines = append(lines, extractedSSELine{
+				Data:    line,
+				IsError: currentEvent == "error",
+			})
 		}
 	}
 	return lines
@@ -91,6 +141,7 @@ func transformStreamEvent(
 	event anthropic.MessageStreamEventUnion,
 	state *StreamState,
 	model string,
+	ext *ir.IRExtensions,
 ) ([][]byte, bool, error) {
 	switch event.Type {
 	case "message_start":
@@ -98,13 +149,19 @@ func transformStreamEvent(
 	case "content_block_start":
 		return handleContentBlockStart(event, state, model)
 	case "content_block_delta":
-		return handleContentBlockDelta(event, state, model)
+		return handleContentBlockDelta(event, state, model, ext)
 	case "message_delta":
 		return handleMessageDelta(event, state, model)
 	case "message_stop":
 		return nil, true, nil
 	case "content_block_stop", "ping", "vertex_event":
 		return nil, false, nil
+	case "error":
+		// Some transports surface error events as ordinary
+		// MessageStreamEventUnion entries (no `event: error` header).
+		// Route them through the same error pipeline.
+		chunks, err := handleErrorEvent([]byte(event.RawJSON()), state, model)
+		return chunks, false, err
 	default:
 		return nil, false, nil
 	}
@@ -155,6 +212,15 @@ func handleContentBlockStart(
 		state.BlockIndexToToolIdx[start.Index] = toolIdx
 		chunk, err := buildOpenAIStreamToolChunk(state, model, toolIdx, block.ID, block.Name, "")
 		return [][]byte{chunk}, false, err
+	case "thinking":
+		// Mark the block index as a thinking block so subsequent
+		// thinking_delta and signature_delta events can be routed back to
+		// it. Emit a bootstrap chunk carrying an empty reasoning_content
+		// string so downstream consumers see the field appear (mirroring
+		// the text case's bootstrap content).
+		state.BlockIndexToThinkingActive[start.Index] = true
+		chunk, err := buildOpenAIReasoningChunk(state, model, "")
+		return [][]byte{chunk}, false, err
 	default:
 		return nil, false, nil
 	}
@@ -164,6 +230,7 @@ func handleContentBlockDelta(
 	event anthropic.MessageStreamEventUnion,
 	state *StreamState,
 	model string,
+	ext *ir.IRExtensions,
 ) ([][]byte, bool, error) {
 	deltaEvent := event.AsContentBlockDelta()
 	delta := deltaEvent.Delta
@@ -186,9 +253,63 @@ func handleContentBlockDelta(
 		jsonDelta := delta.AsInputJSONDelta()
 		chunk, err := buildOpenAIStreamToolChunk(state, model, toolIdx, "", "", jsonDelta.PartialJSON)
 		return [][]byte{chunk}, false, err
+	case "thinking_delta":
+		return handleThinkingDelta(deltaEvent, state, model)
+	case "signature_delta":
+		return handleSignatureDelta(deltaEvent, state, ext)
 	default:
 		return nil, false, nil
 	}
+}
+
+// handleThinkingDelta translates an Anthropic thinking_delta into an OpenAI
+// chunk carrying delta.reasoning_content. Returns nil for deltas that arrive
+// without a matching content_block_start (orphan deltas cannot be routed).
+func handleThinkingDelta(
+	deltaEvent anthropic.ContentBlockDeltaEvent,
+	state *StreamState,
+	model string,
+) ([][]byte, bool, error) {
+	if !state.BlockIndexToThinkingActive[deltaEvent.Index] {
+		return nil, false, nil
+	}
+	thinkingDelta := deltaEvent.Delta.AsThinkingDelta()
+	if thinkingDelta.Thinking == "" {
+		return nil, false, nil
+	}
+	chunk, err := buildOpenAIReasoningChunk(state, model, thinkingDelta.Thinking)
+	return [][]byte{chunk}, false, err
+}
+
+// handleSignatureDelta captures the per-block signature into IRExtensions
+// so the symmetric outbound emitter can replay it on multi-turn requests.
+// Signatures have no representation on the OpenAI envelope; with nil ext
+// (the existing OpenAI-client cell), the signature is dropped — no consumer.
+func handleSignatureDelta(
+	deltaEvent anthropic.ContentBlockDeltaEvent,
+	state *StreamState,
+	ext *ir.IRExtensions,
+) ([][]byte, bool, error) {
+	if ext == nil {
+		return nil, false, nil
+	}
+	if !state.BlockIndexToThinkingActive[deltaEvent.Index] {
+		return nil, false, nil
+	}
+	sigDelta := deltaEvent.Delta.AsSignatureDelta()
+	if sigDelta.Signature == "" {
+		return nil, false, nil
+	}
+	ext.SetThinkingSignature(thinkingBlockKey(deltaEvent.Index), sigDelta.Signature)
+	return nil, false, nil
+}
+
+// thinkingBlockKey is the stable IRExtensions.ThinkingSignatures key for
+// a thinking block at Anthropic content_block index. Mirrors the
+// non-streaming inverse helper in client.go so a request that becomes
+// streaming mid-way does not double-store signatures.
+func thinkingBlockKey(blockIdx int64) string {
+	return fmt.Sprintf("content[%d]", blockIdx)
 }
 
 func handleMessageDelta(
@@ -209,6 +330,46 @@ func handleMessageDelta(
 	return [][]byte{chunk}, false, nil
 }
 
+// handleErrorEvent maps an Anthropic `event: error` data payload onto an
+// OpenAI terminal chunk. Per the plan's Q4 resolution: emit
+// finish_reason="error" plus a synthetic content delta carrying the error
+// message, so clients that ignore the non-standard finish_reason still
+// see the human-readable failure in the assistant text stream.
+func handleErrorEvent(data []byte, state *StreamState, model string) ([][]byte, error) {
+	var envelope anthropicErrorEnvelope
+	if err := json.Unmarshal(data, &envelope); err != nil {
+		// Malformed error payload; surface a generic message so the
+		// client at least sees that the stream aborted abnormally.
+		envelope = anthropicErrorEnvelope{
+			Type: "error",
+			Error: anthropicErrorDetail{
+				Type:    anthropicErrorTypeAPI,
+				Message: "unparsable Anthropic error event",
+			},
+		}
+	}
+	message := envelope.Error.Message
+	if message == "" {
+		message = "Anthropic upstream error"
+	}
+
+	contentChunk, err := buildOpenAIStreamChunk(state, model, openai.ChatCompletionChunkChoiceDelta{
+		Content: fmt.Sprintf("[error: %s] %s", envelope.Error.Type, message),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	finishChunk, err := buildOpenAIStreamFinishChunk(state, model, "error", anthropic.MessageDeltaUsage{})
+	if err != nil {
+		return nil, err
+	}
+	if finishChunk == nil {
+		return [][]byte{contentChunk}, nil
+	}
+	return [][]byte{contentChunk, finishChunk}, nil
+}
+
 func buildOpenAIStreamChunk(
 	state *StreamState,
 	model string,
@@ -225,6 +386,23 @@ func buildOpenAIStreamChunk(
 		}},
 	}
 	return json.Marshal(chunk)
+}
+
+// buildOpenAIReasoningChunk emits an OpenAI chunk carrying
+// `reasoning_content` on the delta. The OpenAI Go SDK's delta struct has
+// no first-class reasoning_content field, so the value is injected via
+// sjson after marshal — keeping interop with vsr's existing convention
+// (the non-streaming path reads reasoning_content the same way).
+func buildOpenAIReasoningChunk(state *StreamState, model string, reasoning string) ([]byte, error) {
+	base, err := buildOpenAIStreamChunk(state, model, openai.ChatCompletionChunkChoiceDelta{})
+	if err != nil {
+		return nil, err
+	}
+	withReasoning, err := sjson.SetBytes(base, "choices.0.delta.reasoning_content", reasoning)
+	if err != nil {
+		return nil, fmt.Errorf("inject reasoning_content: %w", err)
+	}
+	return withReasoning, nil
 }
 
 func buildOpenAIStreamToolChunk(
--- a/src/semantic-router/pkg/anthropic/client_stream.go
+++ b/src/semantic-router/pkg/anthropic/client_stream.go
@@ -14,21 +14,51 @@ import (
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/ir"
 )
 
-// StreamState tracks Anthropic SSE events while translating to OpenAI SSE.
+// StreamState tracks per-request streaming translation state. The struct
+// is shared by both directions:
+//
+//   - Inbound (Anthropic SSE → OpenAI SSE), driven by TransformSSEChunkToOpenAI:
+//     uses MessageID, Created, RoleSent, BlockIndexToToolIdx, NextToolIdx,
+//     BlockIndexToThinkingActive.
+//
+//   - Outbound (OpenAI SSE → Anthropic SSE), driven by the PR5 emitter:
+//     uses MessageStartSent, MessageStopSent, NextBlockIndex, the Open*
+//     pointers tracking which block index is currently open, the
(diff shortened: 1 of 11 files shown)


siloteemu:
**Draft pending #1940**

Stacked on #1940 (which is stacked on #1939). The diff against `main` currently includes both predecessors' commits because the branch sits on top. Will be rebased onto fresh `main` and flipped ready once #1940 lands.

What this PR adds on top of #1940: SSE streaming emitter for Anthropic clients with ping keepalive, the `message_stop` exactly-once invariant enforced at the emitter, round-trip parity for `refusal` and `stop_sequence` stop reasons on streaming, and an e2e streaming SSE sequence assertion.

siloteemu:
**Update**: #1939 merged. This branch has been re-cascaded onto the rebased #1940 — diff still includes #1940's commits because #1940 hasn't landed yet. Staying draft until #1940 merges, then will rebase onto fresh `main` and flip ready.

Pre-existing draft note retained above for history.

siloteemu:
**Update**: #1940 merged. This branch has been rebased onto current `main` (diff now shows only this PR's incremental work) and flipped from draft to ready for review.

This is the feature-completing PR for native Anthropic `/v1/messages` ingress in the stack — adds the SSE streaming emitter and the `message_stop` exactly-once invariant on top of #1940's non-streaming outbound emitter. With this landed, Claude Code (and other Anthropic streaming clients) can talk to vsr without an Anthropic↔OpenAI translation proxy in front.

src/semantic-router/pkg/extproc/processor_res_body.go
```diff
@@ -27,7 +27,23 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
...
-	if ctx.IsStreamingResponse && ctx.APIFormat == config.APIFormatAnthropic {
+	// Anthropic clients on the streaming path always go through the
+	// outbound emitter so the client sees Anthropic-shape SSE,
+	// regardless of upstream APIFormat. The dispatcher branch order
+	// matters: this check precedes the older
+	// handleAnthropicStreamingResponseBody branch because the
+	// double-Anthropic cell (Anthropic client + Anthropic backend)
+	// would otherwise match the legacy branch and the client would
+	// receive OpenAI SSE.
+	if ctx.IsStreamingResponse && ctx.ClientProtocol == config.ClientProtocolAnthropic {
```
rootfs, line 38: [risk:high] The four-cell dispatch routing change (Anthropic-client branch added before legacy Anthropic-backend branch) is load-bearing but has no unit test covering the routing matrix. The PR description claims a dedicated dispatcher table test exists, but no such test appears in the 10-file diff. A `handleResponseBody` table test covering all four cells (ClientProtocol×APIFormat) should accompany this change.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
```diff
@@ -0,0 +1,145 @@
...
+
+	anthropicBytes, streamDone, err := anthropic.EmitAnthropicSSEChunk(
+		openAIBytes, ctx.AnthropicStream, ctx.IRExtensions, ctx.RequestModel,
+	)
+	if err != nil {
+		logging.Errorf("Failed to emit Anthropic SSE chunk: %v", err)
+		return buildResponseBodyContinueResponse(nil, nil)
+	}
+
+	if streamDone || strings.Contains(chunkStr, "data: [DONE]") {
```
rootfs, line 82: [risk:medium] `finalizeStreamingResponse` can be called twice in the double-Anthropic cell: the prior chunk calls it when `streamDone=true` from the emitter, then this terminal chunk calls it again because `translateUpstreamToOpenAI` returns `"data: [DONE]\n\n"` and `strings.Contains` fires. If `finalizeStreamingResponse` is not idempotent (e.g. appends to a slice or emits a metric twice), this causes a double-finalization bug. The regression test `SuppressesUpstreamTail` does not assert against this.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
```diff
@@ -0,0 +1,145 @@
...
+	// MessageStream accumulator) treat pings as no-ops; the value is
+	// keeping middleboxes from severing the long-lived connection
+	// during sparse-token windows. The ping is prepended below so it
+	// reaches the client before this chunk's content events.
+	pingBytes := maybeEmitPing(ctx.AnthropicStream)
+
+	openAIBytes, err := r.translateUpstreamToOpenAI(responseBody, ctx)
+	if err != nil {
+		logging.Errorf("Failed to translate upstream streaming chunk to OpenAI: %v", err)
+		return buildResponseBodyContinueResponse(nil, nil)
```
rootfs: [risk:medium] When `translateUpstreamToOpenAI` returns an error, the handler returns an empty `buildResponseBodyContinueResponse(nil, nil)` — a nil BodyMutation — which causes Envoy to forward the raw upstream bytes verbatim. This breaks the wire-ownership invariant described at line N95. The error path should return a non-nil empty BodyMutation (same as the happy-path empty-emission case).

src/semantic-router/pkg/anthropic/client_stream.go
```diff
@@ -10,43 +10,102 @@ import (
...
+	RoleSent                   bool
+	BlockIndexToToolIdx        map[int64]int
+	NextToolIdx                int
+	BlockIndexToThinkingActive map[int64]bool
+
+	// Outbound direction (OpenAI → Anthropic).
+	MessageStartSent     bool
+	MessageStopSent      bool
+	NextBlockIndex       int64
+	OpenTextBlockIndex   *int64
```
rootfs, line 46: [risk:low] `OpenTextBlockIndex`, `OpenThinkingBlockIdx`, and `ToolIdxToBlockIndex` are outbound-direction fields sharing a struct with inbound fields. The comment at N31 correctly notes only one direction is active per request, but there is no enforcement. Consider a future extraction when the struct grows further, per the module-boundaries guardrail.

src/semantic-router/pkg/anthropic/sse_out.go
```diff
@@ -0,0 +1,615 @@
...
+	ext *ir.IRExtensions,
+	model string,
+) ([]byte, bool, error) {
+	if state == nil {
+		state = NewStreamState()
+	}
+
+	var out bytes.Buffer
+	streamDone := false
+	for _, line := range extractSSEDataLines(openAIChunk) {
```
rootfs, line 84: [risk:medium] `EmitAnthropicSSEChunk` calls `extractSSEDataLines` — the same function used by `TransformSSEChunkToOpenAI` for Anthropic-shaped input. When the input is already OpenAI SSE (OpenAI backend cell), `extractSSEDataLines` correctly parses `data:` lines. However, if the upstream ever sends a bare `{`-prefixed line without a `data:` prefix (the fallback path at line N158 of the `client_stream.go` diff), `IsError` will always be false even if a preceding `event: error` header appeared. This is the same pre-existing behaviour, but it now affects the outbound path too.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client_test.go
```diff
@@ -0,0 +1,160 @@
...
+// TestHandleAnthropicClientStreamingResponseBody_MidStreamError asserts
+// that when EmitAnthropicSSEChunk returns an error (e.g. state == nil),
+// the handler returns a non-nil response with an empty mutation rather
+// than panicking or returning nil (which would forward raw upstream bytes).
+//
+// We trigger the error path indirectly: pass a body that will parse into
+// a valid OpenAI chunk but set state to nil in the context so the emitter
+// cannot proceed. The handler must recover gracefully.
+func TestHandleAnthropicClientStreamingResponseBody_MidStreamError(t *testing.T) {
+	router := newTestRouter()
```
rootfs, line 148: [risk:low] The test comment says "nil state causes graceful nil-init inside emitter" but `newAnthropicStreamCtx(nil)` sets `ctx.AnthropicStream = nil`. `EmitAnthropicSSEChunk` does nil-init `state` internally, but `maybeEmitPing(ctx.AnthropicStream)` is called first (line N60 of the handler) and also guards on nil. The test exercises the emitter nil-init path but does not actually trigger an error return from `EmitAnthropicSSEChunk` as the comment implies — the test name `MidStreamError` is misleading.

src/semantic-router/pkg/anthropic/client_stream_test.go
```diff
@@ -92,17 +118,202 @@ func TestTransformSSEChunkToOpenAI_AccumulatesToolArguments(t *testing.T) {
...
+// silently dropped (pre-fix behavior).
+func TestTransformSSEChunkToOpenAI_ThinkingBlockStart(t *testing.T) {
+	state := NewStreamState()
+	events := buildAnthropicSSE(
+		`{"type":"content_block_start","index":0,"content_block":{"type":"thinking","thinking":""}}`,
+	)
+	out, _, err := TransformSSEChunkToOpenAI([]byte(events), state, "claude-sonnet-4-5", nil)
+	require.NoError(t, err)
+	assert.Contains(t, string(out), `"reasoning_content"`)
+	assert.True(t, state.BlockIndexToThinkingActive[0], "block index should be tracked as thinking")
```
rootfs: [risk:low] `state.BlockIndexToThinkingActive[0]` is read directly from the map after `TransformSSEChunkToOpenAI` returns. This implicitly asserts the inbound path populates `BlockIndexToThinkingActive`, but the assertion would silently pass with `false` if the key is absent (Go map zero-value). Use `require.Contains(t, state.BlockIndexToThinkingActive, int64(0))` before asserting the value to make the test fail loudly when the key is missing.

rootfs (commented):
---

## PR 1941 Code Review

### Verdict
`comment`
The safety invariants (message_stop exactly-once, always-replace BodyMutation, nil-safe `ext`) are sound and backed by targeted regression tests, but one load-bearing dispatch routing change has no unit test despite the PR description claiming it does.

---

### Risk

**1. Missing dispatcher table test for `processor_res_body.go` routing change**
`src/semantic-router/pkg/extproc/processor_res_body.go:38-47` adds a new first-match branch (`ClientProtocol == Anthropic`) that shadows the legacy `APIFormat == Anthropic` branch. The PR description states "dedicated dispatcher table test covers all four cells" but the 10-file diff contains no change to `extproc_test.go` or any new `processor_res_body_test.go`. The four-cell routing matrix (Anthropic/OpenAI client × Anthropic/OpenAI backend) is untested at the dispatch layer. If `ClientProtocol` is never set for a request that previously matched the legacy branch, the request would fall through to `handleStreamingResponseBody` (the generic OpenAI streamer), producing an OpenAI SSE stream to an Anthropic client.

**2. Potential double `finalizeStreamingResponse` in the double-Anthropic cell**
`processor_res_body_streaming_anthropic_client.go:82-84`: the condition `streamDone || strings.Contains(chunkStr, "data: [DONE]")` can fire on the terminal `message_stop` chunk even after the prior chunk already called `finalizeStreamingResponse` via `streamDone`. In the SuppressesUpstreamTail scenario: (a) `EmitAnthropicSSEChunk` returns empty bytes (guard `MessageStopSent`), so `streamDone=false`; (b) but `translateUpstreamToOpenAI` still returns `"data: [DONE]\n\n"` for the raw upstream `message_stop`, so `strings.Contains` fires and calls `finalizeStreamingResponse` a second time. Whether this is safe depends on idempotency of `finalizeStreamingResponse` (not verified in the regression test `TestHandleAnthropicClientStreamingResponseBody_SuppressesUpstreamTail`).

**3. `TransformSSEChunkToOpenAI` signature break — all callers are covered**
The 4th `ext *ir.IRExtensions` parameter is a potential break. Localizer confirms only two call-sites outside test files — both are updated in this diff (`processor_res_body_streaming_anthropic.go:35`, `processor_res_body_streaming_anthropic_client.go:139`). No out-of-PR callers found. Low risk.

---

### Tests

**Covered:**
- `client_stream_test.go` additions (N136–N200): thinking-block-start, thinking_delta, signature_delta (with and without ext), error event — all directly test the four previously-silent-drop bugs. Good.
- `sse_out_test.go` (new, 415 LOC): TextOnly, ToolUseAccumulation, TextThenToolUse, ReasoningContent, SignatureReplay — state machine and exactly-once invariants exercised per path. Good.
- `processor_res_body_streaming_anthropic_client_test.go`: SuppressesUpstreamTail, PingInterleave, EmptyStream, MidStreamError — the body-mutation guarantee is directly tested. Good.
- Existing `TestHandleAnthropicStreamingResponseBody_TranslatesSSE` (`processor_res_body_streaming_anthropic_test.go:15`) will exercise the updated 4-parameter `TransformSSEChunkToOpenAI` call with nil ext; passes because `ext` is nil-safe throughout.

**Missing:**
- **Dispatcher routing test**: the dispatch change at `processor_res_body.go:38-47` has no unit test. A table test covering all four cells (Anthropic+Anthropic, Anthropic+OpenAI, OpenAI+Anthropic, OpenAI+OpenAI) at the `handleResponseBody` level would lock the branch ordering.
- **Double-finalize in tail suppression**: `TestHandleAnthropicClientStreamingResponseBody_SuppressesUpstreamTail` asserts the body mutation is correct but does not assert `ctx.StreamingComplete` is set exactly once or that `finalizeStreamingResponse` is idempotent under re-entry.
- `assertEventOrder` is called at `e2e/testcases/anthropic_messages_streaming.go:N122` but is not defined in the visible portion of the new file; likely defined in the truncated tail (lines ~188–243) — cannot confirm from diff.

---

### Style/Hygiene

- `processor_res_body_streaming_anthropic_client.go:N27–43`: the 7-line doc block on `handleAnthropicClientStreamingResponseBody` describing a three-step pipeline (translate → accumulate → re-emit) is well-structured and matches the actual code — no concern.
- PR description lists `ToolUseEmittedStart` as an outbound `StreamState` field (commit 2 message), but no such field appears in the struct diff. Tool start emission is tracked via `ToolIdxToBlockIndex` presence. Minor PR description inaccuracy.
- `InitialUsage anthropic.Usage` is added to `StreamState` (diff N50) as the "message_start usage placeholder." The diff of `sse_out.go` is truncated so it's unclear whether `InitialUsage` is populated before `emitMessageStart` reads it, or whether message_start always emits zero usage in streaming. If zero, `message_start` will show `input_tokens: 0` to clients, which differs from Anthropic's streaming spec. Worth a follow-up clarification.

---

### Size

2079 additions across 10 files. The bulk (`sse_out.go` at 615 lines, `sse_out_test.go` at 415 lines) is a new self-contained emitter with its tests — appropriate for the scope. The extproc-layer additions (145 + 160 lines) are also bounded. The `client_stream.go` changes (+261/-19) span bug-fix, struct expansion, and function signature changes; commit trajectory is separated correctly. No splits warranted given the stacked PR context.

---

Tools used: view, bash, grep

src/semantic-router/pkg/extproc/processor_res_body.go
siloteemu: Good catch — the description claimed a dispatcher table test that did not exist. Added `TestHandleResponseBody_StreamingDispatchMatrix` (commit b4e9af9) covering all four cells (ClientProtocol × APIFormat) at the `handleResponseBody` level, asserting each reaches the correct handler by observable wire shape so a future reorder or a missing ClientProtocol guard fails loudly. PR description corrected to match.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
siloteemu: Confirmed and fixed (commit 09b4143d). `finalizeStreamingResponse` was not idempotent — it re-recorded completion latency and re-called `inflight.End`. It now early-returns when `ctx.StreamingComplete` is already set, so the double-entry you described in the tail-suppression scenario is a no-op. The guard lives in finalizeStreamingResponse itself, so all four call sites are covered.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
siloteemu: Fixed (commit 09b4143d). Both error paths now return an empty non-nil BodyMutation via a new `emptyAnthropicBodyMutation` helper, so Envoy never forwards the raw upstream chunk once we have entered the handler. The helper centralises the wire-ownership invariant doc.

src/semantic-router/pkg/anthropic/client_stream.go
siloteemu: Agreed — these are outbound-direction fields sharing the struct with inbound fields, active one-direction-per-request but not enforced. Leaving as-is for this PR to keep the diff scoped; noting it as a follow-up extraction candidate per the module-boundaries guardrail once the struct grows further.

src/semantic-router/pkg/anthropic/sse_out.go
siloteemu: Correct that this is pre-existing `extractSSEDataLines` behaviour now also reachable on the outbound path: a bare `{`-prefixed line without a `data:` prefix won't carry the preceding `event: error` association. The real backends we target always emit `data:`-prefixed frames, so this doesn't bite in practice. Tracking as a follow-up rather than expanding this PR's scope — flagging it explicitly so it's not lost.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client_test.go
siloteemu: Right — the test name was misleading. `EmitAnthropicSSEChunk` does not return an error for malformed input (verified: it emits empty output, nil error), so the test was exercising the empty-emission path, not the err branch. Renamed to `_UnparsableChunk` and corrected the comment; it now also asserts the empty (non-nil) mutation body (commit b4e9af9).

src/semantic-router/pkg/anthropic/client_stream_test.go
siloteemu: Fixed (commit b4e9af9). Added `require.Contains(t, state.BlockIndexToThinkingActive, int64(0))` before the value assertion so a dropped key fails loudly instead of silently passing via the bool zero value.

siloteemu:
Thanks for the thorough review @rootfs — all seven points addressed.

**Code fixes** (commit `09b4143d`):
- **finalize idempotency**: `finalizeStreamingResponse` now early-returns on `ctx.StreamingComplete`, so the double-entry in the tail-suppression scenario is a no-op. Covers all four call sites.
- **wire-ownership on error paths**: both error returns in `handleAnthropicClientStreamingResponseBody` now return an empty non-nil BodyMutation via a new `emptyAnthropicBodyMutation` helper, so Envoy never forwards raw upstream bytes.

**Test fixes** (commit `b4e9af92`):
- **dispatcher matrix test**: added `TestHandleResponseBody_StreamingDispatchMatrix` covering all four cells (ClientProtocol × APIFormat) — the claim in the description is now backed by a real test.
- **`MidStreamError` → `UnparsableChunk`**: renamed and corrected; the emitter emits empty (not an error) on malformed input, so the test now correctly exercises the empty-emission path and asserts the empty mutation body.
- **thinking-active map assertion**: `require.Contains` before the value assertion so a dropped key fails loudly.

**Acknowledged as follow-ups** (kept out of this PR to stay scoped):
- struct direction-field separation (outbound vs inbound fields share `StreamState`)
- `extractSSEDataLines` `IsError` association on bare-`{` frames (pre-existing; real backends always emit `data:`-prefixed frames)
- `message_start` emits `input_tokens: 0` in the OpenAI-backend cell (no upstream message_start to source from; the cumulative `message_delta.usage` carries the real counts)

PR description corrected: removed the non-existent `ToolUseEmittedStart` field reference and named the dispatcher test.

siloteemu:
Rebased onto current `main` to resolve a semantic conflict with #2042 (`fix/streaming-chunks-memory-accumulation`), which replaced `RequestContext.StreamingChunks []string` with `HasStreamingChunks bool` to stop accumulating SSE chunks in memory. The new `handleAnthropicClientStreamingResponseBody` still appended to the removed slice, breaking the typecheck. Adapted it to set `ctx.HasStreamingChunks = true` (matching how #2042 updated the sibling OpenAI/Anthropic streaming handlers); folded into the commit that introduced the handler so every commit still compiles. No behavior change — the chunk text was only used as a has-chunks signal here.

siloteemu:
Heads-up on CI: the only red check, `build_pr (llm-katan)`, is a transient Docker Hub registry timeout during buildx setup (`Get "https://registry-1.docker.io/v2/": ... Client.Timeout exceeded`), not a code failure — llm-katan isn't touched by this PR. Every code-relevant check (pre-commit lint, test-and-build, both kubernetes + dashboard integration tests, and the other Docker image builds) passed independently. A re-run of that one job should clear it; I don't have re-run rights on this repo.

@rootfs whenever you have a moment — all seven of your review comments are addressed (5 fixed across commits `09b4143d` / `b4e9af92`, 2 acknowledged as scoped follow-ups), and the branch is rebased onto current `main` (resolving the `HasStreamingChunks` rename from #2042). Ready for another look when you get the chance. Thanks again for the thorough first pass.

siloteemu:
One more pass surfaced a real edge-case bug, now fixed (commit `800d44b5`):

**`message_stop` was not emitted for usage-only terminal chunks.** If a backend ends the stream with a chunk that has empty `choices` + `usage` and *no* prior `finish_reason`, `emitChunkEvents` returned early on the no-choices branch with `streamDone=false`, so `message_stop` never fired and an Anthropic client would hang. Now a usage-bearing no-choices chunk is treated as terminal. The common `include_usage` ordering (finish_reason chunk, then trailing usage) is unaffected — the `MessageStopSent` guard keeps `message_stop` exactly-once. Added `TestEmitAnthropicSSEChunk_UsageOnlyTerminal` for the previously-hanging path.

Also strengthened the dispatcher matrix test (commit `a4bc6187`): the pass-through cell was asserting absence on an empty wire string (trivially true); each cell now asserts its positive, distinct outcome so a misroute fails loudly.

siloteemu:
Coverage follow-up (commit `171b5927`): a coverage-gap pass found the previously-fixed usage-only-terminal path was only tested with an open text block. Added cases that leave a thinking block and multiple tool blocks open at the usage-only terminal (asserting each gets a `content_block_stop` before `message_stop`), plus an error-as-first-chunk case verifying the emitter sends a bare `error` + `message_stop` with no synthesized `message_start` (matches Anthropic's API). No production-code change in this commit — emitter behavior was already correct; these lock the contracts.

Xunzhuo (approved):
LGTM

src/semantic-router/pkg/anthropic/client_stream_test.go
siloteemu: This one is addressed on the current head (commit `b4e9af9`, now at `171b5927`): `client_stream_test.go:147` adds `require.Contains(t, state.BlockIndexToThinkingActive, int64(0))` before the `assert.True` on the next line, so a dropped key now fails loudly instead of passing on the bool zero value. Safe to resolve.

PR #1922, opened 2026-05-20
[Router][CLI] Custom Anthropic upstream routing and tool calling

akshayv:
PLEASE FILL IN THE PR DESCRIPTION BELOW AND CONFIRM THE CHECKLIST ITEMS.

## Purpose

- What does this PR change?

  **CLI (`config_generator.py`):** Anthropic models with `api_format: anthropic` no longer always use the hardcoded `anthropic_api_cluster` (`api.anthropic.com`). A model is only placed on the shared cluster when every `backend_ref` targets the default Anthropic host; otherwise Envoy gets a per-model cluster and `host_rewrite_literal` from `backend_refs` (same path as OpenAI/vLLM backends).

  **Router (`processor_req_body_anthropic.go`):** Anthropic routing now resolves the upstream chat path from the provider profile (`base_url` + `/v1/messages`), applies `extra_headers` from the endpoint profile (e.g. API gateway subscription keys), selects the real endpoint for tracing (instead of a hardcoded host), and starts router replay on the Anthropic path so Routing Insights records MoM → Claude decisions.

  **Router (`client.go` / `client_test.go`):** Merges OpenAI-style tool definitions and tool-result history into Anthropic Messages API request/response bodies for Envoy-routed traffic, with unit tests for round-trips and custom `:path` headers.

- Why is this change needed?

  Today, any `api_format: anthropic` model is skipped in Envoy generation and forced onto `anthropic_api_cluster` → `api.anthropic.com`, while extproc already rewrites bodies and paths from `backend_refs`. That breaks deployments that expose the Anthropic API on another host (enterprise gateways, proxies, or vendor-hosted endpoints). Requests reach the wrong upstream (empty/HTML/401 responses) even when config lists the correct `endpoint` and `base_url`.

  Tool calling for routed Claude models also requires converting OpenAI `tools` / `tool_calls` / `tool` messages to Anthropic `tools` and content blocks; without this, MoM-routed sessions that rely on tools fail at the transform layer.

- Which module(s) does this affect?

  `Router`, `CLI`

## Test Plan

- What commands, checks, or manual steps should reviewers use?

  Automated:

  ```bash
  cd src/vllm-sr
  pytest tests/test_config_generator.py::test_generate_envoy_config_custom_anthropic_upstream_rewrites_host -v

  cd src/semantic-router
  go test ./pkg/anthropic/... -count=1
  go test ./pkg/extproc/... -run 'Anthropic|HandleAnthropic' -count=1
  ```

  Manual (local gateway):

[...]

Issue #1404, opened 2026-02-28
feature: first-class dual API entry points for OpenAI and Anthropic


samzong:
### Describe the feature

Add first-class dual ingress support for both protocols:

- OpenAI endpoints: `/v1/chat/completions`, `/v1/responses`
- Anthropic endpoint: `/v1/messages`

Both paths should use one shared routing pipeline through a protocol-neutral IR.

Scope:
- Keep existing OpenAI compatibility.
- Add native Anthropic request/response handling.
- Align tool/function calling across protocols (`tools/tool_choice` and `tool_use/tool_result`).
- Align streaming behavior across protocols.
- Remove Anthropic-specific hardcoded provider routing and use the same endpoint/provider_profile/failover path used by other providers.
- Ensure existing plugins (system prompt injection, reasoning, auto tool selection, memory, RAG, security filters) run consistently for both protocols.

### Why do you need this feature?

Current Anthropic support works for basic text routing, but it is not at parity for coding-tool workflows.

For production routing, we need both protocols to be first-class so we can:
- support OpenAI-compatible clients and Anthropic-native clients at the same time,
- avoid protocol-specific behavior drift in core routing logic,
- make streaming and tool-calling behavior predictable across providers.

Without this, external claims of “OpenAI + Anthropic parity” are hard to defend in real coding-agent use cases.

### Additional context

Related issues (not duplicates):
- #1138 (open): multi-protocol API abstraction umbrella
- #1011 (closed): initial Anthropic API support
- #1217 (open): OpenAI schema alignment

AayushSaini101:
@Xunzhuo @rootfs can work on this feature request ?

siloteemu:
Cross-linking an RFC that proposes the concrete design for this tracking goal: #1946. The RFC engages with the multi-protocol-adapter design doc and frames Anthropic-format ingress as the complementary axis to that work. There are six in-flight PRs implementing the proposal (one ready, the rest stacked drafts); the RFC explains the architecture, what carries forward in either direction of long-term IR shape, and what we are asking maintainers to decide. Discussion is on the RFC issue.

siloteemu:
@AayushSaini101 — implementation is in flight; the design lives in RFC #1946 (Native Anthropic `/v1/messages` ingress).

Merged so far: #1937, #1938, #1944, #1950. Open: #1939 (ready for review); #1940 and #1941 in draft (the non-streaming and streaming response emitters).

[...]

PR #1944, opened 2026-05-26
fix(anthropic): preserve client passthrough fields through ToAnthropicRequestBody

siloteemu:
> **RFC**: #1946 — design rationale, alternatives considered, decisions invited from maintainers

Standalone PR (not stacked on anything). Fixes 10 distinct silent drops in
`pkg/anthropic/client.go::ToAnthropicRequestBody` and friends. End-to-end
verification against a real Anthropic backend (via an Anthropic-compatible
gateway proxy) confirmed these drops break observable behavior: prompt-cache
counters stay zero, images vanish, the wrong `stop_reason` gets reported,
custom `anthropic-version` headers get overwritten with the package default,
and so on.

## What's dropped today

`ToAnthropicRequestBody` only forwards a strict subset of OpenAI fields onto
the Anthropic backend body, and `BuildRequestHeaders` hardcodes
`anthropic-version: 2023-06-01`. As a result:

1. `cache_control` markers — dropped → `cache_creation_input_tokens` and
   `cache_read_input_tokens` always 0
2. `top_k` parameter — dropped
3. `metadata.user_id` — dropped (mirrored to OpenAI `User` but not surfaced
   back as `metadata` on the Anthropic backend body)
4. `anthropic-version` header — overwritten with the package default
5. `anthropic-beta` header — dropped
6. Image content blocks — flattened to text via `extractUserContent`
7. System content as array (with per-block `cache_control`) — collapsed to a
   single string
8. `tool_result.is_error: true` — dropped on translation
9. `tool_result` array content (image / nested blocks) — flattened to empty
   string
10. `stop_sequence` stop_reason mapping — handled correctly today on this
    branch (the response-side switch already covers it); included in the
    test matrix to prevent regression

Items 1-9 are real silent drops. Item 10 is a no-op fix on this branch
baseline but kept in the verification matrix.

## Approach

Introduce a package-local `AnthropicPassthrough` carrier — a sidecar struct
that holds the Anthropic-only fields captured at inbound parse time and
replayed at outbound emit time. New entrypoints `*WithPassthrough` accept a
nullable `*AnthropicPassthrough`; existing entrypoints stay byte-identical
(verified by a 5-fixture parity test) so unrelated callers see no behavior
change. Callers that want the fix opt in by calling the `*WithPassthrough`
variant.

No new external dependencies, no IR-package introduction — purely scoped to
`pkg/anthropic/`. Caller wiring is one ExtProc call site
(`processor_req_body_anthropic.go`).

## Commits

1. **`[Router] add AnthropicPassthrough carrier and inbound builder`** — new
   `passthrough.go` + tests. Pure additive, no behavior.
2. **`[Router] thread passthrough into ToAnthropicRequestBody`** — refactor
   to add the new entrypoints. Byte-identical to the legacy path when
   passthrough is nil. Parity locked by
   `TestToAnthropicRequestBodyWithPassthrough_NilIsByteIdenticalToLegacy`.
3. **`[Router] replay top_k, metadata, cache_control, system-array via passthrough`**
4. **`[Router] preserve image blocks across the Anthropic write path`** —
   covers both Anthropic-shape inbound AND OpenAI-shape `OfImageURL`/
   `OfImageBase64` parts in the existing OpenAI-client/Anthropic-backend
   cell. Closes the bug for both ingress paths.
5. **`[Router] forward anthropic-version and anthropic-beta from inbound`** —
   with precedence `profile pin > inbound passthrough > package default`
   (achieved via call ordering + Envoy's "last header wins" semantics).
   Streaming variant covered by `BuildStreamingRequestHeadersWithPassthrough`.
6. **`[Router] preserve tool_result.is_error and array content`**
7. **`[Router] wire ext-proc Anthropic ingress to build the passthrough carrier`** —
   the one ExtProc call site that builds the passthrough from the raw inbound
   body and threads it through.

## SDK note for reviewers

`anthropic-sdk-go`'s `NewToolResultBlock(id, content string, isError bool)`
takes a `string` for content. For `tool_result` array content (item 9) the
carrier assembles `ToolResultBlockParam` manually in a localized helper — the
only place this PR fights the SDK shape.

## Test plan

- [x] `make agent-lint CHANGED_FILES=...` clean
- [x] `go vet`, `go build`, `go test ./pkg/anthropic/...` clean (46 new tests)
- [x] DCO sign-off on all 7 commits, SSH-signed
- [x] Per-commit trajectory verified (each commit independently compiles +
      vets + passes agent-lint)
- [x] Byte-identical parity test for nil-passthrough (`TestToAnthropicRequestBodyWithPassthrough_NilIsByteIdenticalToLegacy`)
- [x] Per-field unit tests for every replay (cache_control across system/messages/tools,
      top_k, metadata.user_id, system-array, image blocks for both Anthropic and OpenAI
      shape, anthropic-version, anthropic-beta, tool_result.is_error,
      tool_result array content)
- [x] End-to-end verification against a real Anthropic backend: requests with
      `cache_control` on the system block now show non-zero
      `cache_creation_input_tokens` on first call and `cache_read_input_tokens`
      on second call; images survive; client `anthropic-version` is forwarded
- [x] E2E test added: `e2e/testcases/anthropic_passthrough_regression.go` (`anthropic-passthrough-openai-regression`) — guards OpenAI `/v1/chat/completions` against regressions from the passthrough carrier construction code; positive-path carrier behavior is covered by unit tests in `pkg/anthropic/` since the e2e backend is OpenAI-shaped and an Anthropic-native mock is out of scope
- [ ] Upstream CI — runs once this PR is marked ready

## Related

- Tracking issue: #1404 (dual-protocol entry points context)
- `pkg/anthropic/client.go` was introduced by #1922 (custom Anthropic upstream
  routing) — these drops have been live since then. Streaming variant in #1926.
- This fix benefits ANY caller of `ToAnthropicRequestBody`, including the
  existing OpenAI-client / Anthropic-backend cell.

### DIFF
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -23,6 +23,7 @@ var BaselineRouterContract = []string{
 	"decision-fallback-behavior",
 	"plugin-config-variations",
 	"chat-completions-progressive-stress",
+	"anthropic-passthrough-openai-regression",
 	// Session observability
 	"session-telemetry-metrics",
 	"session-pricing-chat-completions",
--- a/e2e/testcases/anthropic_passthrough_regression.go
+++ b/e2e/testcases/anthropic_passthrough_regression.go
@@ -1,0 +1,93 @@
+package testcases
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/vllm-project/semantic-router/e2e/pkg/fixtures"
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+)
+
+func init() {
+	pkgtestcases.Register("anthropic-passthrough-openai-regression", pkgtestcases.TestCase{
+		Description: "Verify the Anthropic passthrough carrier does not regress OpenAI chat completions requests",
+		Tags:        []string{"anthropic", "passthrough", "regression"},
+		Fn:          testAnthropicPassthroughOpenAIRegression,
+	})
+}
+
+// testAnthropicPassthroughOpenAIRegression guards the OpenAI request path
+// against regressions introduced by the Anthropic passthrough carrier.
+//
+// The passthrough machinery (cache_control replay, anthropic-version
+// forwarding, image-block preservation, multi-block system reassembly)
+// only fires when routing prepares an Anthropic-native upstream request.
+// The e2e backend (llm-katan) is OpenAI-shaped, so the positive path is
+// not exercisable here without scope-expanding the e2e infrastructure to
+// add an Anthropic-native mock — explicitly out of scope for this PR.
+//
+// What is exercisable: this test sends a standard OpenAI request and
+// asserts a clean 200 round-trip with routing headers. That guards
+// against the case where the new carrier construction code accidentally
+// breaks the dominant OpenAI traffic shape (regression). Per-field
+// behavior of the carrier itself (cache_control, image blocks,
+// anthropic-version) is covered by unit tests in
+// src/semantic-router/pkg/anthropic.
+func testAnthropicPassthroughOpenAIRegression(ctx context.Context, client *kubernetes.Clientset, opts pkgtestcases.TestCaseOptions) error {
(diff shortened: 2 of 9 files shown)


Xunzhuo (approved):
LGTM

PR #1945, opened 2026-05-26
fix(metrics): stop double-counting decision evaluation metrics

WUKUNTAI-0211:
## Summary

`llm_decision_evaluation_latency_seconds` and `llm_decision_match_total` are recorded **twice per request**:

1. Inside `decision.DecisionEngine.EvaluateDecisionsWithSignals` ([`engine.go#L114-L143`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/decision/engine.go#L114-L143)) — the canonical recording site used by all callers (DSL test runner, classification services).
2. Again in `extproc.runDecisionEngine` ([`req_filter_classification_runtime.go#L99-L131`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/extproc/req_filter_classification_runtime.go#L99-L131)) wrapping the engine call.

This PR removes the redundant recording in the extproc layer and adds a regression test that fails if either metric is observed more than once per `EvaluateDecisionsWithSignals` call.

## Impact on Dashboards

- **`llm_decision_match_total`** counter increments by **2** per matched decision, so `Decision Match Rate` and `Total Matches` panels are inflated ~2x.
- **`llm_decision_evaluation_total`** counter increments by **2** per request, so `Decision Evaluation Rate` is inflated ~2x.
- **`llm_decision_evaluation_latency_seconds`** histogram receives **two observations per request**. Since both observations land in the same coarse-grained Prometheus default bucket (everything <5ms collapses into `le=0.005`), the impact on `histogram_quantile` is hard to detect, but the histogram `_count` and the rate of observations are doubled — anything that derives a per-request mean (`sum / count`) from this histogram is affected.

## Empirical verification on a live router

To prove this isn't theoretical, I built **two router binaries from the same branch** — one with this PR applied (`router.fixed`), one with only the patched file reverted (`router.buggy`) — and ran an A/B test against the same live router container, swapping the binary in between and confirming each swap via md5sum.

Setup: 15 sequential requests through Envoy → router on each binary, fresh restart between phases.

| Metric | BUGGY (per successful request) | FIXED (per successful request) | Change |
|---|---|---|---|
| `llm_decision_evaluation_total` | **2.50** | **1.25** | **−50%** |
| `llm_decision_match_total{external_when_not_confidential}` | **2.00** | **1.00** | **−50%** |
| `llm_decision_match_total{internal_default}` | 1.50 | 1.25 | −17% |

The `external_when_not_confidential` series shows the cleanest signal: exactly **2.0 → 1.0** per request, confirming the double-counting was per-decision and is fully eliminated by this PR. The `evaluation_total` factor of 2.5 vs 1.25 includes additional non-200 requests that still ran the decision engine (warmup + 503s); those also went from 2× → 1×.

A follow-up validation run on the fixed binary with 30 requests over 10 minutes (read via `increase(...[15m])`):

| Query | Observed (fixed) | Expected (fixed) | Expected (buggy) |
|---|---|---|---|
| `increase(llm_decision_evaluation_total[15m]) / requests` | 1.03 | ~1.0 | ~2.0 |
| `sum(increase(llm_decision_match_total[15m])) / requests` | 1.93 | ~2.0 (one per matching decision × 2 matching decisions) | ~4.0 |

## What is intentionally kept

- The tracing span (`tracing.StartDecisionSpan` / `EndDecisionSpan`) around the extproc call. The span measures a different boundary (per-request decision latency surfaced in Jaeger) and is what the trace-view UI relies on.
- All recording inside the decision engine itself — that is the single source of truth for these Prometheus metrics, and every other caller already relies on it.

## Why the engine is the right canonical site

`EvaluateDecisionWithEngine` has four callers:

- `src/semantic-router/pkg/extproc/req_filter_classification_runtime.go` ← only caller that double-recorded
- `src/semantic-router/pkg/services/classification.go`
- `src/semantic-router/pkg/services/classification_signal_contract.go` (two call sites)
- `src/semantic-router/cmd/dsl/test_runner.go`

Three of four already rely on the engine's internal `defer metrics.RecordDecisionEvaluation(...)` and the engine's per-match `metrics.RecordDecisionMatch(...)`. Moving the recording fully into the engine makes the contract consistent across all callers.

## Migration / breaking change notes

This is observably a metric value change for existing deployments running these Prometheus queries:

- `rate(llm_decision_match_total[...])` will drop to ~half of its previous value.
- `rate(llm_decision_evaluation_total[...])` will drop to ~half of its previous value.
- `llm_decision_evaluation_latency_seconds_count` will grow at half the previous rate.

These are **corrections**, not regressions — but anyone with alerting thresholds calibrated against the buggy values should rebaseline.

## Verification

- `go vet ./pkg/extproc/... ./pkg/decision/... ./pkg/observability/metrics/...` — clean
- `go test ./pkg/decision/... ./pkg/observability/metrics/...` — pass (incl. new regression test)
- `go test -short ./pkg/extproc/...` — pass (no extproc behavior regression)
- New regression test independently verified to fail when double-counting is reintroduced (`delta=2, want 1`) and pass with the fix in place.
- Empirical A/B test on a live router (above) confirms the expected metric reduction in production.

## Files changed

- `src/semantic-router/pkg/extproc/req_filter_classification_runtime.go` — drop redundant `metrics.RecordDecisionEvaluation` and `metrics.RecordDecisionMatch`; add a code comment naming the canonical recording site to prevent re-introduction.
- `src/semantic-router/pkg/decision/engine_metrics_test.go` — new regression test asserting exactly-once emission per `EvaluateDecisionsWithSignals` call.

### DIFF
--- a/src/semantic-router/pkg/decision/engine_metrics_test.go
+++ b/src/semantic-router/pkg/decision/engine_metrics_test.go
@@ -1,0 +1,68 @@
+package decision
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	dto "github.com/prometheus/client_model/go"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/metrics"
+)
+
+func histogramSampleCount(t *testing.T, h prometheus.Histogram) uint64 {
+	t.Helper()
+	var m dto.Metric
+	if err := h.Write(&m); err != nil {
+		t.Fatalf("histogram Write: %v", err)
+	}
+	if m.Histogram == nil {
+		t.Fatalf("histogram payload missing")
+	}
+	return m.Histogram.GetSampleCount()
+}
+
+// TestEvaluateDecisionsWithSignals_EmitsMetricsExactlyOnce locks in the
+// invariant that the canonical recording site for the decision-evaluation
+// histogram and decision-match counter is the engine itself. If a caller
+// (notably the extproc runtime) ever re-introduces wrapper-level recording,
+// these counts would double and this test will fail.
+//
+// Regression guard for the double-counting bug previously present in
+// pkg/extproc/req_filter_classification_runtime.go::runDecisionEngine.
+func TestEvaluateDecisionsWithSignals_EmitsMetricsExactlyOnce(t *testing.T) {
+	engine := NewDecisionEngine(
+		nil,
+		nil,
+		nil,
+		[]config.Decision{
+			{
+				Name:     "catch-all",
+				Priority: 10,
+				Rules: config.RuleCombination{
+					Operator:   "AND",
+					Conditions: []config.RuleCondition{},
+				},
+			},
+		},
+		"priority",
+	)
+
+	latencyBefore := histogramSampleCount(t, metrics.DecisionEvaluationLatency)
+	matchesBefore := testutil.ToFloat64(metrics.DecisionMatchTotal.WithLabelValues("catch-all"))
+
+	if _, err := engine.EvaluateDecisionsWithSignals(&SignalMatches{}); err != nil {
+		t.Fatalf("EvaluateDecisionsWithSignals returned unexpected error: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


Xunzhuo (approved):
/lgtm Thanks!

Issue #1753, opened 2026-04-10
feature: productionize session-aware model switching for stable multi-turn routing


Xunzhuo:
## Summary

Productionize session-aware model switching for stable multi-turn routing so the router can preserve conversation continuity while still allowing controlled model changes when the evidence supports a switch.

## Background

The repository now has a clear workstream for session-aware routing, but it still needs one canonical umbrella issue that ties together the runtime, DSL, analytics, and rollout seams. Stable multi-turn routing is not only about sticking to one model forever; it is about carrying enough session state to decide when continuity is more valuable than an immediate downgrade or upgrade, and when switching models is justified despite context rebuild cost. This umbrella keeps the workstream reviewable by decomposing it into contributor-sized slices rather than one opaque feature branch. It also gives maintainers a single place to track rollout expectations, escape hatches, and evaluation strategy.

## Goals

- [ ] Track one canonical multi-turn routing workstream that covers session context plumbing, CRM or continuity baselines, runtime telemetry, DSL primitives, lookup tables, retention hooks, and the final shadow-mode switch gate.
- [ ] Define the production contract for session state, continuity overrides, and safe fallbacks before enforcement behavior ships broadly.
- [ ] Keep the rollout staged: baseline anti-bounce fixes first, auditable heuristic or shadow-mode behavior next, and learned tuning last.
- [ ] Maintain explicit child issues for each major seam so contributors can start from narrow entry points instead of one umbrella ticket.
- [ ] **Not goals:** landing the entire workstream in one PR or skipping directly to black-box RL policy updates.

## Context

**Primary child issues**
- `#1751` - session-aware context plumbing for multi-turn follow-up turns.
- `#1752` - CRM baseline for multi-turn continuity.
- `#1742` - per-turn session telemetry and cumulative cost state.
- `#1743` - cache warmth inference and model transition logging.
- `#1744` - `SESSION_STATE` declarations.
- `#1745` - `SIGNAL session` and `SIGNAL lookup`.
- `#1746` - lookup tables for quality gaps, handoff penalties, and turn priors.
- `#1747` - `EMIT` and retention directives.
- `#1748` - bounded candidate iteration.
- `#1749` - shadow-mode `model_switch_gate`.
- `#1750` - offline weight updates.

**Affected files / directories**
- `src/semantic-router/pkg/extproc/**` - request or response routing context, telemetry, and runtime state seams.
- `src/semantic-router/pkg/selection/**` - session-aware model choice and algorithm hooks.
[...]

Issue #1752, opened 2026-04-10
feature: add conversational routing momentum as a baseline for multi-turn session affinity


Xunzhuo:
## Summary

Add Conversational Routing Momentum (CRM) as an optional baseline so the router can dampen model bouncing across multi-turn conversations before the full model-switch policy lands.

## Background

Once the routing pipeline has access to session-aware context, the repository still needs a lightweight continuity strategy that is easier to reason about than a full stay-vs-switch optimizer. CRM is a useful baseline because it smooths turn-level complexity changes with fast-attack and slow-release dynamics, which helps preserve continuity across brief follow-up turns without forcing permanent sticky sessions. It is also valuable as a comparison point for later shadow-mode switch-gate work. This issue keeps CRM as an explicit, optional baseline instead of burying the idea inside the larger policy workstream.

## Goals

- [ ] Add an optional CRM configuration surface that is disabled by default and preserves current behavior unless explicitly enabled.
- [ ] Apply CRM over multi-turn complexity or routing signals so simple follow-up turns decay gradually instead of instantly de-escalating.
- [ ] Integrate CRM at a narrow selection or signal-preparation seam rather than scattering momentum logic across the request pipeline.
- [ ] Add focused tests proving attack and release dynamics produce stable routing across representative multi-turn conversations.
- [ ] **Not goals:** the final session-aware model-switch gate, lookup-table-backed handoff costs, or offline RL tuning.

## Context

**Example baseline config**
```yaml
routing_momentum:
  enabled: true
  attack: 0.3
  release: 0.9
  threshold: 0.5
```
Treat this as an illustrative shape, not a locked public contract. The core intent is fast escalation, slow decay, and opt-in rollout.

**Example behavior**
- Turn 1: a complex coding request routes to a stronger model.
- Turn 2: the user says "looks good, commit it".
- CRM should keep enough momentum from turn 1 that the route does not immediately collapse to a much weaker model on turn 2.

**Affected files**
- `src/semantic-router/pkg/extproc/req_filter_classification_signal.go` - signal-preparation path that can expose the current turn history and derived complexity inputs.
- `src/semantic-router/pkg/extproc/request_signal_history.go` - current multi-turn message history extraction.
- `src/semantic-router/pkg/selection/**` - if CRM is implemented at the selection layer rather than signal preparation.
- `src/semantic-router/pkg/config/**` - config contract for attack, release, threshold, and enable or disable flags.
[...]

Issue #1750, opened 2026-04-10
feature: train offline weight updates for session-aware model-switch policies


Xunzhuo:
## Summary

Train offline weight updates for session-aware model-switch policies so learned tuning can build on logged routing evidence without becoming a runtime dependency.

## Background

The local mid-session switching plan explicitly puts learned policy work last. The repo already has RL-aware selection surfaces and adjacent feedback-loop issues, but the session-aware workstream needs a scoped follow-up that consumes the telemetry, transition logs, and lookup tables from earlier issues instead of skipping straight to an opaque online learner. This issue defines how offline weight updates or evaluation should plug into the new session-aware gate while preserving safe heuristic fallbacks.

## Goals

- [ ] Define an offline dataset contract sourced from session telemetry, transition logs, lookup tables, and eventual user-feedback signals.
- [ ] Evaluate or produce updated policy weights for session-aware switch decisions without making training or external services mandatory at request time.
- [ ] Provide a versioning or rollout contract so learned weights can be shadowed, compared, and reverted independently from the runtime code path.
- [ ] Document how this work composes with the shadow-mode `model_switch_gate` rather than replacing stable fallbacks.
- [ ] **Not goals:** mandatory online learning, removing rule-based or heuristic fallback paths, or shipping a black-box policy with no inspection hooks.

## Context

**Affected files**
- `src/semantic-router/pkg/selection/rl_driven.go` - current feedback-aware and session-aware selector concepts.
- `src/semantic-router/pkg/routerreplay/store/store.go` - likely source of replay or session history records.
- `src/semantic-router/pkg/observability/**` - telemetry surfaces that must feed the offline dataset.
- Related planning or evaluation docs if the repo needs a maintained contract for learned-weight artifacts.

**Entry points**
- Start from the data contract and artifact versioning story, not from a new runtime shortcut.
- Align with existing feedback-loop work such as #1512 and adjacent research issues instead of creating one-off training semantics.

**Validation gates**
```bash
make agent-report ENV=cpu CHANGED_FILES="src/semantic-router/pkg/selection/rl_driven.go,src/semantic-router/pkg/routerreplay/store/store.go,src/semantic-router/pkg/observability,docs/agent/plans/pl-0030-session-aware-model-switch-workstream.md"
[...]

Issue #1748, opened 2026-04-10
feature: add bounded candidate iteration for session-aware routing policies


Xunzhuo:
## Summary

Add bounded candidate iteration to the routing DSL so session-aware policies can reason over candidate models without hardcoding one branch per backend.

## Background

The local plan for mid-session model switching proposes a minimal `FOR ... IN` surface so policies can evaluate candidate models systematically. The current DSL supports `DECISION_TREE` and related routing constructs, but it has no way to express bounded iteration over the current candidate set. Without that capability, the final switch gate would either hardcode model names or collapse back into opaque runtime-only logic. This issue adds the smallest iteration surface that keeps the language auditable and constrained.

## Goals

- [ ] Add a bounded `FOR ... IN` authoring surface to the parser, AST, compiler, and decompiler.
- [ ] Restrict iteration to explicit candidate sets or model lists so the DSL does not become an unbounded scripting language.
- [ ] Define how iteration outputs feed existing decision or policy constructs without creating a second execution runtime.
- [ ] Add validation and maintained tests that reject unsafe or ambiguous iteration patterns.
- [ ] **Not goals:** general-purpose scripting, free-form loops over runtime state, or the final switch policy itself.

## Context

**Illustrative authoring shape**
```dsl
FOR candidate IN decision.candidates {
  // evaluate one bounded candidate at a time
}
```
The exact syntax can still change, but the contributor-facing constraint is fixed: iteration must be bounded to an explicit candidate set already known to the router.

**Affected files**
- `src/semantic-router/pkg/dsl/ast.go`, `src/semantic-router/pkg/dsl/parser.go`, `src/semantic-router/pkg/dsl/compiler.go`, and `src/semantic-router/pkg/dsl/decompiler.go` - new iteration syntax and lowering.
- `src/semantic-router/pkg/dsl/validator.go` and related tests - bounded-surface enforcement.
- `src/semantic-router/pkg/selection/selector.go` - candidate-model contract that the iteration surface must remain compatible with.
- Maintained routing examples if the new syntax needs one visible repo-native sample.

**Entry points**
- Start from the existing top-level DSL constructs and keep any iteration result compatible with the current model-candidate selection flow.
- Align with the DSL plan in `pl-0012`; the new surface should extend the current language, not create a sidecar policy interpreter.

**Validation gates**
```bash
[...]

PR #1960, opened 2026-05-29
[Router] Populate Chat Completions previous-model to unblock ModelSwitchGate enforce (#1753)

theohsiung:
Refs #1753

## Purpose

- **What does this PR change?** Makes the auditable `ModelSwitchGate` actually usable for **Chat Completions** traffic by giving it the previous-model signal it needs. It adds a small in-memory, session-keyed "last model used" store, records each turn's model at response time, and reads it back as `ctx.PreviousModel` during the next turn's request prep.
- **Why is this change needed?** #1753 (umbrella; under roadmap theme #1513) is blocked for the dominant API surface by a single gap: `pkg/extproc/session_transition.go` left `ctx.PreviousModel` empty for Chat Completions (`// TODO: populate PreviousModel ... once per-turn model history is available`). With it empty, the gate treats `previous_model` as a missing signal and **falls back to audit-only for every chat request** (even in enforce mode), emitting `model_switch_gate_enforce_unavailable`. Response API already worked because it derives the previous model from the conversation chain. This PR closes exactly that gap.
- **Which module(s) does this affect?** `Router` (Go).

### Design

- New `pkg/sessiontelemetry/last_model.go`: thread-safe `map[sessionID]{model,lastSeen}` with `RecordLastModel` / `GetLastModel`, a 1h TTL, and a size cap — mirroring the existing `telemetry.go` and `transition.go` store idioms.
- Write at response time in `recordSessionTurn` (Chat path only), keyed on `ctx.SessionID` — the **same** key the read side uses; the write→read race is impossible (write is turn N's response, read is turn N+1's request).
- Read in `populateSessionTransitionFields`, replacing the TODO. Response API behavior is untouched (it still derives previous model from the conversation chain).
- Additive and gated by the existing `model_switch_gate.enabled`; no decision-contract change.

### Scope / follow-ups (out of this PR)

- DSL primitives (#1744–#1748), CRM (#1752), offline weight updates (#1750) — separate workstreams.
- The store is **in-memory and per-replica** (lost on restart, not shared across horizontally-scaled routers), matching the existing telemetry/transition stores. Durable/shared previous-model persistence is recorded as follow-up debt in `pl-0030`; until then enforce-mode continuity is best-effort and the gate still logs `model_switch_gate_enforce_unavailable` when the signal is genuinely absent (first turn, unresolvable session, or after a restart).

## Test Plan

Run from the router module (multi-module repo; CGO + prebuilt Rust bindings on `LD_LIBRARY_PATH`):

- `go test -race ./pkg/sessiontelemetry/` — store: record/get, overwrite, ignore-empty, TTL expiry, size-cap eviction.
- `go test -race ./pkg/extproc/ -run 'SessionTransition|ChatCompletions|ModelSwitchGate|MaybeEmit'` — wiring:
  - `…PopulatesPreviousModelFromStore` and `…NoPriorModelLeavesEmpty`
  - `TestChatCompletionsModelContinuityAcrossTurns` — full write→read across two turns (turn 1 records `model-a`, turn 2 recovers it as `PreviousModel`).
  - existing gate/session tests still pass (audit-only fallback on empty previous-model preserved).
- `go vet ./pkg/sessiontelemetry/ ./pkg/extproc/ ./pkg/config/`, `gofmt -l`, markdownlint on the plan doc.

## Test Result

- All targeted tests **PASS** with `-race`; full `pkg/config` suite passes; no regressions in existing session/gate tests.
- `gofmt`/`go vet`/markdownlint: clean.
- Behavior change is config-gated and Chat-only; Response API path byte-for-byte unchanged. No wire-format change, so no E2E update is required for this slice.
- Follow-up risk: per-replica in-memory continuity (durable store tracked in `pl-0030`).

### DIFF
--- a/docs/agent/plans/pl-0030-session-aware-model-switch-workstream.md
+++ b/docs/agent/plans/pl-0030-session-aware-model-switch-workstream.md
@@ -131,12 +131,14 @@ This ensures:
 - 2026-05-10: Chose Beta distribution accumulation (same as online `UpdateFeedback`) for offline replay rather than introducing a separate reward model. Keeps the weight format identical across online and offline paths, simplifying version promotion.
 - 2026-05-10: Added `MinRecordsPerModel` threshold so models with sparse offline evidence keep their parent-version priors unchanged, preventing noisy updates from small samples.
 - 2026-05-10: Policy versions are file-backed JSON rather than database-backed to keep the offline path zero-dependency and operator-friendly for inspection.
+- 2026-05-29: Unblocked enforce-mode for Chat Completions by recording the per-turn model in an in-memory, session-keyed last-model store (`pkg/sessiontelemetry/last_model.go`) at response time (`recordSessionTurn`) and reading it back as `ctx.PreviousModel` during request prep (`populateSessionTransitionFields`). Previously only Response API carried previous-model history via the conversation chain, so the ModelSwitchGate fell back to audit-only for all Chat Completions traffic regardless of mode (#1753). Keyed on the same `ctx.SessionID` the read side uses; mirrors the existing telemetry/transition store idioms (1h TTL + size cap).
 
 ## Follow-up Debt / ADR Links
 
 - `OfflineDatasetBuilder` implementation depends on RouterReplay store query capabilities (session-range queries). May need a store extension for windowed reads.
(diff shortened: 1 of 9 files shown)


Xunzhuo (approved):
LGTM

PR #1986, opened 2026-05-30
fix: price session cache pressure by input checkout cost

Xunzhuo:
## Summary
- price session-aware prefix-cache pressure from prompt checkout delta instead of prompt+completion total price
- clamp cached-input pricing into the valid prompt-cost range before computing pressure
- cover cached-input discount and completion-heavy pricing cases

## Validation
- `go test ./pkg/selection`
- `make agent-lint CHANGED_FILES="src/semantic-router/pkg/selection/session_aware_scoring.go src/semantic-router/pkg/selection/session_aware_test.go"`
- `make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/selection/session_aware_scoring.go src/semantic-router/pkg/selection/session_aware_test.go"`
- CPU smoke attempted; blocked before readiness because the generated runtime config was not mounted at the container path expected by the harness

### DIFF
--- a/src/semantic-router/pkg/selection/session_aware_scoring.go
+++ b/src/semantic-router/pkg/selection/session_aware_scoring.go
@@ -1,6 +1,10 @@
 package selection
 
-import "math"
+import (
+	"math"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
 
 func (s *SessionAwareSelector) adjustScores(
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Closing this slice in favor of a single integrated session-aware GA follow-up on top of current main. The cache-price idea remains useful, but it should be reworked together with the final policy instead of merged as a conflicting stacked PR.

PR #1993, opened 2026-05-31
[Dashboard] Fail closed when the auth service cannot initialize

### LINKED ISSUE #1992: [Bug] Dashboard serves the control plane unauthenticated when the auth store fails to initialize
### Describe the bug

The dashboard backend **fails open**: if the authentication store cannot be initialized, every route — including config deploy/rollback, admin user management, MCP tool execution, and the upstream proxy — is served with **no authentication**.

In `dashboard/backend/router/auth_routes.go`, `setupAuthRoutes()` returns a `nil *auth.Service` whenever `auth.NewStore(cfg.AuthDBPath)` errors (e.g. an unwritable/misconfigured `AuthDBPath`):

```go
store, err := auth.NewStore(cfg.AuthDBPath)
if err != nil {
    log.Printf("failed to init auth store: %v", err)
    registerAuthUnavailableRoutes(mux)
    return nil          // <-- nil service
}
```

`wrapWithAuth()` then takes the nil path and wraps the mux **without** `AuthenticateRequest`:

```go
func wrapWithAuth(mux *http.ServeMux, authSvc *auth.Service) *http.ServeMux {
    wrappedMux := http.NewServeMux()
    if authSvc != nil {
        wrappedMux.Handle("/", auth.AuthenticateRequest(authSvc)(mux))
        return wrappedMux
    }
    wrappedMux.Handle("/", mux)   // <-- no auth wrapper: every route is public
    return wrappedMux
}
```

Only the five `/api/auth/*` routes return 503 (via `registerAuthUnavailableRoutes`); all functional/admin routes are wide open. A single DB/permissions misconfiguration silently disables all access control on a config-mutating control plane.

### Steps to reproduce

1. Start the dashboard backend with an `AuthDBPath` that cannot be opened (read-only dir, bad path, or unwritable volume).
2. Observe the log line `failed to init auth store: …`.
3. Without any credentials, call a protected endpoint, e.g. `GET /api/admin/users` or `POST /api/router/config/deploy`.
4. **Result:** the request is served (the handler runs) instead of being rejected.

### Expected behavior

When auth cannot be initialized, the server should **fail closed** — deny routes that require authentication — rather than expose them. Public/static routes can remain reachable so the UI can surface the misconfiguration.

### Actual behavior

All protected routes are served unauthenticated.

### Severity

Critical — unauthenticated access to a config-mutating admin control plane on an error/misconfiguration path.

### Environment

`dashboard/backend` (`router` / `auth` packages); affects current `main`. Found via source audit.

> Note: there is no `SECURITY.md` / private-disclosure channel configured, and the project has a track record of handling security fixes publicly (e.g. #1968, #1455, #1457). Filing here per that convention; happy to move this to a private channel if maintainers prefer.

I have a fix ready (deny-by-default guard + tests) and will open a PR that closes this issue.

glitch-ux:
Closes #1992

## Purpose

- **What:** When the dashboard auth store fails to initialize, the backend no longer serves the control plane unauthenticated — it now **fails closed**.
- **Why:** `wrapWithAuth` previously fell back to `wrappedMux.Handle("/", mux)` with no `AuthenticateRequest` wrapper whenever `authSvc == nil`, exposing config deploy/rollback, admin user management, MCP tooling, and the proxy without authentication. `authSvc` is nil whenever `auth.NewStore(cfg.AuthDBPath)` errors (e.g. an unwritable/misconfigured `AuthDBPath`).
- **Module(s):** `Dashboard`.

Changes:

- Add `auth.ServiceUnavailableGuard()` — a deny-by-default counterpart to `AuthenticateRequest` that **reuses the same `requiresAuthentication` policy**, so the protected-route set cannot drift between the two paths. Routes that require auth return `503 Service Unavailable`; public routes (login/bootstrap, `/api/setup/state`, embedded assets, the static frontend) pass through so the UI can still render and surface the misconfiguration.
- `wrapWithAuth` uses the guard on the nil-service path and logs a warning.
- No supported configuration relies on the old behavior — `authSvc` is nil *only* on a real init failure (there is no intentional no-auth mode), so this is not a behavior regression for any valid deployment.

Design note: I chose **fail-closed** over **hard-fail-at-startup** so the static UI and the existing `/api/auth/*` 503 endpoints can still render a clear "authentication service is not configured" message. Happy to switch to `log.Fatal` at startup instead if maintainers prefer.

## Test Plan

```bash
cd dashboard/backend
gofmt -l auth/middleware.go router/auth_routes.go router/auth_routes_test.go auth/middleware_test.go
go vet ./auth/... ./router/...
go test ./auth/... ./router/...
```

New tests:
- `auth.TestServiceUnavailableGuard` — protected/admin/embedded paths → 503 with `next` not called; login/setup-state/static → pass through.
- `router.TestWrapWithAuthFailsClosedWhenAuthUnavailable` — with `authSvc == nil`, protected and admin routes return 503 **without running their handlers**, while public/static routes are still served.

## Test Result

- `gofmt`: clean · `go vet`: clean · `go build ./...`: ok
- `go test ./auth/... ./router/...` → both packages **PASS** (new guard test + regression test pass; existing `TestRequiresAuthentication` / `TestRequiredPermission` / `TestExtractAccessToken` unaffected).
- Follow-up risk: none for valid configs. The guard centralizes policy in `requiresAuthentication`, so any future public-route change applies to both the authenticated and the fail-closed paths automatically.

---
<details>

- [x] PR title uses module-aligned prefix `[Dashboard]`
- [x] Commits in this PR are signed off with `git commit -s` (DCO)
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and results for this change

</details>

### DIFF
--- a/dashboard/backend/auth/middleware.go
+++ b/dashboard/backend/auth/middleware.go
@@ -69,6 +69,28 @@ func AuthenticateRequest(service *Service) func(http.Handler) http.Handler {
 	}
 }
 
+// ServiceUnavailableGuard returns middleware that fails closed when the auth
+// service could not be initialized. It rejects every request to a route that
+// normally requires authentication with 503 Service Unavailable, while still
+// allowing public routes (login/bootstrap endpoints, setup state, embedded
+// assets, and the static frontend) through so the dashboard can render and
+// surface the "authentication service is not configured" state.
+//
+// This is the deny-by-default counterpart to AuthenticateRequest: it shares
+// the same requiresAuthentication policy so the set of protected routes cannot
+// drift between the two paths.
+func ServiceUnavailableGuard() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if requiresAuthentication(r.URL.Path) {
+				http.Error(w, "Authentication service is not configured", http.StatusServiceUnavailable)
+				return
+			}
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
 func RequiredPermission(method, path string) string {
 	path = strings.TrimSpace(strings.ToLower(path))
 	for _, resolver := range []func(string, string) (string, bool){
--- a/dashboard/backend/auth/middleware_test.go
+++ b/dashboard/backend/auth/middleware_test.go
@@ -38,6 +38,48 @@ func TestRequiresAuthentication(t *testing.T) {
 	}
 }
 
+func TestServiceUnavailableGuard(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name     string
+		path     string
+		wantCode int
+		wantNext bool
+	}{
+		{name: "protected api denied", path: "/api/router/config", wantCode: http.StatusServiceUnavailable, wantNext: false},
+		{name: "admin denied", path: "/api/admin/users", wantCode: http.StatusServiceUnavailable, wantNext: false},
+		{name: "embedded denied", path: "/embedded/grafana/", wantCode: http.StatusServiceUnavailable, wantNext: false},
+		{name: "login public", path: "/api/auth/login", wantCode: http.StatusOK, wantNext: true},
+		{name: "setup state public", path: "/api/setup/state", wantCode: http.StatusOK, wantNext: true},
+		{name: "static frontend public", path: "/dashboard", wantCode: http.StatusOK, wantNext: true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			nextCalled := false
+			next := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				nextCalled = true
+				w.WriteHeader(http.StatusOK)
+			})
+			handler := ServiceUnavailableGuard()(next)
+
+			req := httptest.NewRequest(http.MethodGet, tc.path, nil)
+			rec := httptest.NewRecorder()
+			handler.ServeHTTP(rec, req)
+
+			if rec.Code != tc.wantCode {
+				t.Fatalf("status = %d, want %d", rec.Code, tc.wantCode)
+			}
+			if nextCalled != tc.wantNext {
+				t.Fatalf("next handler called = %v, want %v", nextCalled, tc.wantNext)
+			}
+		})
+	}
+}
+
 func TestExtractAccessToken(t *testing.T) {
 	t.Parallel()
 
--- a/dashboard/backend/router/auth_routes.go
+++ b/dashboard/backend/router/auth_routes.go
@@ -92,6 +92,13 @@ func wrapWithAuth(mux *http.ServeMux, authSvc *auth.Service) *http.ServeMux {
 		wrappedMux.Handle("/", auth.AuthenticateRequest(authSvc)(mux))
 		return wrappedMux
 	}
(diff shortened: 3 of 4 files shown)


WUKUNTAI-0211 (approved):
LGTM — verified locally and the fail-closed boundary holds against the attack surface enumerated in #1992. Nice catch and clean fix.

## What I verified

Cloned the branch into a worktree on `origin/main` and ran:
- `go build ./auth/... ./router/...` — ok
- `go vet ./auth/... ./router/...` — clean
- `gofmt -l auth/middleware.go router/auth_routes.go router/auth_routes_test.go auth/middleware_test.go` — clean
- `go test ./auth/... ./router/...` — both packages PASS, including the two new tests and the existing `TestRequiresAuthentication` / `TestRequiredPermission` / `TestExtractAccessToken` suite.

Then I ran an adversarial smoke test through `wrapWithAuth(mux, nil)` (not against the guard directly, so any drift between policy and wiring would surface) covering:
- protected control plane: `GET /api/router/config`, `POST /api/router/config/deploy`, `GET /api/admin/users`, `GET /api/admin/audit-logs`, `GET /api/mcp/tools`, `POST /api/proxy/llm` → all 503, no backend handler ran ✅
- case-insensitivity: `GET /API/Router/Config`, `GET /Api/Admin/Users` → still 503 (`ToLower` in `requiresAuthentication` holds the line) ✅
- trailing slash / sub-resource: `/api/admin/users/`, `/api/admin/users/42` → 503 ✅
- embedded surface: `/embedded/grafana/`, `/embedded/wizmap/` root → 503 (only `/embedded/wizmap/assets/...` is intentionally public, and that one still passed through as designed) ✅
- public surface that must stay reachable so the UI can render the misconfiguration state: `/api/auth/login`, `/api/setup/state`, `/` and `/dashboard/users` SPA paths, `/embedded/wizmap/assets/foo.js` → all pass through ✅

Centralizing on `requiresAuthentication` so the protected-route set cannot drift between `AuthenticateRequest` and the new `ServiceUnavailableGuard` is exactly the right call — it means future additions only need to be classified once.

## Non-blocking observations (informational)

1. **Pre-existing baseline, not introduced here:** the bare path `GET /api` (no trailing slash, not one of the five explicit cases in `requiresAuthentication`) hits the default `return false` branch and falls through to the SPA fallback (200). This is the same behavior as on `main` under the normal `AuthenticateRequest` path, so it is not a regression from this PR and not blocking. Worth keeping in mind if anyone later tightens `requiresAuthentication`: a future change could explicitly classify `/api` and `/api/` as protected so the policy is symmetric. Easy to add as a one-liner later; absolutely fine to leave as-is here since the scope is "fail closed without changing the policy contract."

2. **Design choice (`fail-closed` vs `log.Fatal`):** I think you made the right call. Letting `/api/auth/*` 503 endpoints plus the static UI surface the misconfiguration is much better operator UX than crashing the process and leaving the operator to grep logs. The warning log on the fail-closed path (`auth service unavailable; authenticated routes are failing closed (503)…`) is the right shape and points at the actual cause (`AuthDBPath` / JWT config).

3. **Optional follow-up (truly out of scope for this PR):** if you ever want to make the misconfiguration even more obvious in the UI, the SPA could read the public `/api/auth/me` (or a dedicated `/api/setup/state` field) and surface a banner. Not for this PR — the 503s and the warning log are sufficient to close the unauth fail-open hole, which is what #1992 asks for.

Thanks for the thorough description + design note — made the review fast. Approving.

Xunzhuo:
@WUKUNTAI-0211 thanks for helping review PR

Xunzhuo (approved):

PR #1998, opened 2026-06-01
fix(classification): scan prior user turns in history-aware PII/jailbreak signals

### LINKED ISSUE #1961: security: multi-turn egress leak — negative signals never scan prior user turns, so earlier-turn secrets are forwarded verbatim to external providers
# security: multi-turn egress leak — negative signals never scan prior *user* turns, so earlier-turn secrets are forwarded verbatim to external providers

## Summary

In a multi-turn deployment that routes between an internal lane and an **external provider lane** (any config where positive signals + the absence of negative signals route to a third-party provider, and a negative signal vetoes back to the internal lane), the egress safety gate is effectively blind to secrets introduced in an **earlier user turn**.

Once a later turn's user message looks clean, the negative signals (PII / jailbreak / etc.) do not fire, the request is routed to the external lane, and the **entire `messages[]` array — including the earlier turn that contained the secret — is forwarded verbatim** to the external provider.

The root cause: the negative signals that gate egress only ever look at
- the **current** user message (`evaluationText` / `currentUserText`), and
- the **non-user** history (`nonUserMessages` = system + assistant), when `include_history: true` is set.

They **never** look at `priorUserMessages` (earlier user turns), which is exactly where user-supplied secrets live. So the gate cannot fail-closed across turns.

> This is the **security/egress** framing of the same underlying mechanism that #1439 documented from a **cost/quality** angle. #1439 is not a duplicate — it never connects the per-turn classification behavior to data exfiltration to an external provider.

## Severity

**High** — confidential content (API keys, internal documents, regulated PII) introduced in any earlier user turn can be exfiltrated to an external/third-party provider, with **trivial reproduction**: the default multi-turn behavior of a chat client that resends the full conversation history triggers it. No special client needed.

## Reproduction

Config: an egress recipe where positive signals + the absence of negative signals route to an **external** provider lane, and any negative signal vetoes to the **internal** lane.

**Turn 1 — secret in a *user* message, correctly vetoed to internal:**

```jsonc
POST /v1/chat/completions
{
  "model": "auto",
  "messages": [
    {"role": "user", "content": "Here is our prod key: sk-REDACTED..., store it for later. Now explain how key rotation works."}
  ]
}
```
The PII / jailbreak signals fire on the current user message → vetoed to the **internal** lane. So far so good.

**Turn 2 — clean follow-up, routed to EXTERNAL, Turn-1 secret goes with it:**

```jsonc
POST /v1/chat/completions
{
  "model": "auto",
  "messages": [
    {"role": "user", "content": "Here is our prod key: sk-REDACTED..., store it for later. Now explain how key rotation works."},
    {"role": "assistant", "content": "Key rotation works by ..."},
    {"role": "user", "content": "thanks, summarize that in one line"}
  ]
}
```

Now:
- `evaluationText` / `currentUserText` = `"thanks, summarize that in one line"` → clean.
- `nonUserMessages` = `["Key rotation works by ..."]` (the assistant turn). Even with `include_history: true` on PII/jailbreak, this is all they scan.
- The **Turn-1 user message containing the secret sits in `priorUserMessages`, which no negative signal reads.**
- → No veto fires → request routed to the **external** lane → the full `messages[]` (including the secret) is forwarded verbatim to the external provider.

### Pre-empting the `include_history` objection

Even with `include_history: true` set on the PII and jailbreak detectors, this is **not** mitigated. `include_history` only adds `nonUserMessages` (system + assistant) to the scan — see `collectPIIRuleContents` (`classifier_pii_ops.go:205`) and the jailbreak content collection (`classifier_signal_context.go:529`). The earlier **user** turn lives in `priorUserMessages`, which is never passed to the PII or jailbreak evaluators (see Root Cause #2).

## Root Cause (line-aligned to `main`)

**1. Signal evaluation's primary input is only the current user message.**
`resolveIntentSignalInputFromMessages()` sets `evaluationText = history.currentUserMessage`
([`intent_request_messages.go:59`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/services/intent_request_messages.go#L59)). The context/fallback paths only ever join `nonUserMessages` + `currentUserMessage` — `priorUserMessages` is never folded into `evaluationText` or `contextText`.

**2. `priorUserMessages` IS plumbed through, but only one signal consumes it.**
`EvaluateAllSignalsWithContext(... priorUserMessages, nonUserMessages ...)` receives it
([`classifier_signal_context.go:74`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/classification/classifier_signal_context.go#L74)),
but in the dispatcher:
- [`classifier_signal_dispatch.go:61`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/classification/classifier_signal_dispatch.go#L61) — **Reask** is the *only* signal that gets `priorUserMessages`.
- [`classifier_signal_dispatch.go:92`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/classification/classifier_signal_dispatch.go#L92) — **Jailbreak** gets only `nonUserMessages`.
- [`classifier_signal_dispatch.go:97`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/classification/classifier_signal_dispatch.go#L97) — **PII** gets only `nonUserMessages`.

So the two security-critical negative signals structurally cannot see earlier user turns.

**3. No decision-lane session affinity.**
Each request gets a fresh `RequestContext` (`processor_core.go`, `Process()`); the client body is written to `ctx.OriginalRequestBody` ([`processor_req_body.go:31`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/extproc/processor_req_body.go#L31)). There is no `lastDecision` / `decision_affinity` keyed by session — the internal/external lane decision is recomputed from scratch every turn. (`cache_affinity.go` provides affinity only at the *backend-model* layer, not the lane.)

**4. Body is forwarded verbatim — no history stripping/redaction exists.**
`getBodyMutationSource(ctx)` ([`processor_req_body_routing.go:462`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/extproc/processor_req_body_routing.go#L462)) returns the original body; `BodyMutation` only changes the `model` field. RAG/Memory injection only *appends* messages. There is no scrub/redact/strip of `messages[]` anywhere on the egress path.

## Impact

Any deployment that (a) routes to an external provider lane, (b) uses negative signals to gate egress, and (c) serves multi-turn conversations (the common case) can leak earlier-turn user secrets to the external provider once a later turn looks benign.

## Possible Approaches (not prescribing a solution)

**A. Make negative signals scan prior user history (recommended first step / prerequisite).**
Feed `priorUserMessages` into the PII/jailbreak evaluators when `include_history: true`, alongside `nonUserMessages`. This closes the *detection blind spot*.
*Honest limitation:* A fixes "the gate can't see the secret"; it does **not** fix "even when detected, the full body is still forwarded." If a detector mis-classifies or a signal isn't configured, the verbatim body still egresses. That deeper problem is B/C. **A is a prerequisite for B and C — neither can tag a secret it cannot detect.**

**B. Detect-then-tag-then-redact at egress.**
Tag the offending message spans at detection time, then strip/redact them from `messages[]` before the external `BodyMutation`. Note: this introduces a body-rewriting capability the codebase does not currently have (only memory-append exists today), so it is a new subsystem, not a bug fix.

**C. Detect-then-tag-then-lock-lane (decision-lane session affinity).**
Once a session has ever tripped a negative signal, pin it to the internal lane for the remainder of the session so it can never flip to external. This overlaps with the session-affinity work in #1751 / #1753 and should be coordinated there rather than duplicated here.

## Relation to existing work

- #1439 — same root mechanism (per-turn, last-message classification), framed as **cost/quality** (bouncing to a cheaper model). This issue is the **security/egress** consequence; not a duplicate.
- #1751 / #1753 (currently open) — session-aware multi-turn routing. **Approach C above belongs with these**; this issue's concrete deliverable is **A** (detection coverage of prior user turns), keeping it scoped and non-overlapping.

## Proposed scope for this issue

I'd like to take **Approach A** as a focused, bug-fix-sized PR (wire `priorUserMessages` into the PII/jailbreak evaluators under `include_history`), and leave B/C as follow-up roadmap items.

@Xunzhuo — could you confirm the direction and whether you're comfortable assigning this to me?
@yossiovadia — this is the security-side sibling of #1439; does Approach A overlap with anything you're already doing in #1751 / #1753 / the CRM work?

WUKUNTAI-0211:
Fix multi-turn egress leak from #1961: `include_history: true` previously only scanned non-user (system + assistant) history, so secrets in earlier user turns were never inspected and could be exfiltrated on later benign turns.

Widens the history slice handed to PII/jailbreak evaluators to also include `priorUserMessages`, via a small dedup helper. Evaluator logic unchanged. `include_history: false` (default) path is bit-for-bit identical.

Adds unit + e2e regression tests covering the issue #1961 scenario.

Fixes #1961

### DIFF
--- a/src/semantic-router/pkg/classification/classifier_signal_dispatch.go
+++ b/src/semantic-router/pkg/classification/classifier_signal_dispatch.go
@@ -89,12 +89,14 @@ func (c *Classifier) buildSignalDispatchers(
 		{
 			config.SignalTypeJailbreak, "Jailbreak",
 			func() {
-				c.evaluateJailbreakSignal(results, mu, textForSignal(config.SignalTypeJailbreak), nonUserMessages)
+				c.evaluateJailbreakSignal(results, mu, textForSignal(config.SignalTypeJailbreak), historyForHistoryAwareSignals(priorUserMessages, nonUserMessages))
 			},
 		},
 		{
 			config.SignalTypePII, "PII",
-			func() { c.evaluatePIISignal(results, mu, textForSignal(config.SignalTypePII), nonUserMessages) },
+			func() {
+				c.evaluatePIISignal(results, mu, textForSignal(config.SignalTypePII), historyForHistoryAwareSignals(priorUserMessages, nonUserMessages))
+			},
 		},
 		{
 			config.SignalTypeKB, "KB",
--- a/src/semantic-router/pkg/classification/history_aware_pii_e2e_test.go
+++ b/src/semantic-router/pkg/classification/history_aware_pii_e2e_test.go
@@ -1,0 +1,46 @@
+package classification
+
+import (
+	"sync"
+	"testing"
+
+	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+// Security regression (issue #1961): include_history PII rules must detect a
+// secret placed in a prior USER turn, not just system/assistant history.
+func TestPIISignal_DetectsSecretInPriorUserTurn(t *testing.T) {
+	classifier, _, mockModel := newTestPIIClassifier()
+	classifier.Config.PIIRules = []config.PIIRule{
+		{
+			Name:            "restricted_pii",
+			Threshold:       0.7,
+			IncludeHistory:  true,
+			PIITypesAllowed: []string{},
+		},
+	}
+
+	secretTurn := "my contact is alice@corp.example"
+	currentTurn := "thanks, summarize that in one line"
+	priorUserMessages := []string{secretTurn}
+	nonUserMessages := []string{"Sure, here is how rotation works."}
+
+	mockModel.setMockResponse(secretTurn, []candle_binding.TokenEntity{
+		piiEntity("EMAIL", "alice@corp.example", 14, 32, 0.99),
+	}, nil)
+
+	results := &SignalResults{
+		Metrics:           &SignalMetricsCollection{},
+		SignalConfidences: make(map[string]float64),
+		SignalValues:      make(map[string]float64),
+	}
+	var mu sync.Mutex
+
+	history := historyForHistoryAwareSignals(priorUserMessages, nonUserMessages)
+	classifier.evaluatePIISignal(results, &mu, currentTurn, history)
+
+	if !results.PIIDetected {
+		t.Fatalf("SECURITY: PII in a prior user turn must be detected with include_history=true (issue #1961); got PIIDetected=false")
+	}
+}
--- a/src/semantic-router/pkg/classification/history_aware_signals.go
+++ b/src/semantic-router/pkg/classification/history_aware_signals.go
@@ -1,0 +1,21 @@
+package classification
+
+// Includes prior user turns (not just system/assistant) so include_history
+// detectors catch secrets from earlier turns. Security fix, issue #1961.
+func historyForHistoryAwareSignals(priorUserMessages, nonUserMessages []string) []string {
+	merged := make([]string, 0, len(priorUserMessages)+len(nonUserMessages))
+	seen := make(map[string]struct{}, len(priorUserMessages)+len(nonUserMessages))
+	for _, group := range [][]string{nonUserMessages, priorUserMessages} {
+		for _, msg := range group {
+			if msg == "" {
+				continue
+			}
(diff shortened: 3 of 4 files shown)


Xunzhuo (approved):
Good catch

PR #1999, opened 2026-06-01
fix: reconcile each IntelligentRoute/Pool when multiple coexist

### LINKED ISSUE #1908: [Operator] Reconciler returns error when multiple IntelligentRoutes coexist in a namespace, blocking per-CR validation status updates
## Summary

`Reconciler.getIntelligentRoute` (`src/semantic-router/pkg/k8s/reconciler.go:222-237`) returns an error when more than one `IntelligentRoute` exists in the watched namespace. The reconcile loop's caller in `reconcile()` short-circuits on this error and never reaches the validation code path that calls `updateRouteStatus`. The result: when a second `IntelligentRoute` is applied alongside a valid one, neither CR gets a status update for as long as both exist.

This was surfaced while attempting to write an end-to-end test that exercises the embedding-modality validator landed in #1895 against a real cluster. The unit test in `src/semantic-router/pkg/k8s/reconciler_embedding_modality_test.go` (also from #1895) hand-feeds the validator a single CR at a time via controller-runtime's fake client, so it passes. On a live cluster, the constraint blocks the same shape of test.

## Reproduction

1. Apply a valid `IntelligentRoute` to namespace `default`. It reconciles to `Ready=True`.
2. Apply a second `IntelligentRoute` to the same namespace. (The test fixture used a `queryModality: audio` rule, but the constraint is independent of CR contents - any second CR triggers it.)
3. Observe the reconciler's `watchLoop`. Every 5 seconds, `reconcile()` returns the same error.

## Observed log output

```
reconciler.go:159 "Reconciliation check: failed to get IntelligentRoute: multiple IntelligentRoutes found in namespace default, expected exactly 1"
```

That message repeats indefinitely while both CRs exist. The reconciler does not attempt to validate either route; it does not call `updateRouteStatus`; both CRs sit with empty `status: {}`.

## Why the e2e test gets stuck

The `embedding-signal-modality-validation` testcase in the draft of #1881 polls for `Ready=False, Reason=ValidationFailed` on the bad CR within a 60-second window. The reconciler can't reach the validator while the good CR is also present, so the bad CR's status remains empty for the full polling window and the test times out.

Receipts from a kind run on 2026-05-14:

```
Test FAIL: embedding-signal-modality-validation
  timed out (1m0s) waiting for Ready=False+Reason=ValidationFailed on intelligentroute/default/bad-audio-route
  last observed status="" reason=""
```

## Suggested direction

Make `Reconciler` reconcile each `IntelligentRoute` in the namespace independently rather than asserting exactly-one. Concretely:

- `getIntelligentRoute` becomes `listIntelligentRoutes` (or similar) returning the full slice.
- The reconcile loop iterates the slice and calls `validateAndUpdate` per CR, so each gets its own `Ready`/`ValidationFailed` status.
- The "exactly one" assumption appears to be load-bearing only inside `validateAndUpdate` where a single canonical config is emitted. That step would need a decision about what to do when multiple valid CRs disagree: merge, last-wins, deny-on-conflict, or only-one-active-at-a-time-with-an-explicit-selector. None of those decisions are made here - that's the design conversation this issue exists to start.

`IntelligentPool` is in the same shape today (line 216 has the equivalent "expected exactly 1" check). The same lift would apply if/when multi-pool composition becomes a real use case.

## What this unblocks

Once per-CR reconcile is in, the `embedding-signal-modality-validation` e2e testcase in the draft branch `multimodal-routing-e2e-readd` can land cleanly. The shape is in `e2e/testcases/embedding_signal_modality_validation.go` and the bad fixture is in `e2e/profiles/multimodal-routing/crds/intelligentroute-bad-audio.yaml` on that branch (currently being dropped from #1881 ahead of merge for the reason above).

The validator coverage today is the controller-runtime fake-client unit test in `src/semantic-router/pkg/k8s/reconciler_embedding_modality_test.go`, which exercises all six validator branches. That coverage is sufficient until live-cluster validation becomes feasible.

## Related

- #1895 (reconcile-time embedding-modality validator)
- `docs/agent/tech-debt/td-040-reconcile-path-skips-family-validators.md` (broader tech-debt write-up)
- #1881 (draft) (E2E re-add, dropping the validation testcase pending this fix)

WUKUNTAI-0211:
## Summary

Fixes #1908.

`Reconciler.reconcile` retrieved the namespace's `IntelligentPool` and `IntelligentRoute` through helpers that returned an error whenever **more than one** CR of either kind existed. `reconcile()` short-circuited on that error before reaching `validateAndUpdate`, so when a second `IntelligentRoute` was applied alongside a valid one, **neither** CR ever received a status update — both sat with empty `status: {}` indefinitely, and the watch loop logged the same `multiple IntelligentRoutes found ...` error every tick.

This blocked, among other things, the embedding-modality e2e test from #1908's description: the bad CR's status stayed empty for the full polling window because the reconciler never reached the validator while a second CR was present.

## Change

- `reconcile()` now lists pools and routes directly instead of delegating to the now-removed `getIntelligentPool`/`getIntelligentRoute` single-CR helpers.
- When more than one pool or route is present, the new `reportConflict` marks **every** conflicting CR `Ready=False` / `Reason=Conflict` with an actionable message, then returns an error so the watch loop still surfaces the condition.
- The single-pool / single-route happy path is **unchanged** — exactly-one still flows through `validateAndUpdate` as before.

This is the minimal, deny-on-conflict fix: it makes the conflict observable per-CR without changing the existing single-canonical-config contract. Full multi-CR coexistence (merge / last-wins / explicit selector) remains a separate design decision, as noted in the issue.

## Tests

New `reconciler_conflict_test.go`:
- `MultipleRoutes` — two routes both end up `Ready=False`/`Conflict`.
- `MultiplePools` — two pools both end up `Ready=False`/`Conflict`.
- `TestReconcileSingleCRUnchanged` — regression canary that a single pool+route still reconciles to `Ready=True`.

```
go build ./pkg/k8s/...   # ok
go vet ./pkg/k8s/...      # ok
go test ./pkg/k8s/...     # ok (incl. existing reconciler/validation tests)
```

Signed-off-by per DCO.

### DIFF
--- a/src/semantic-router/pkg/k8s/reconciler.go
+++ b/src/semantic-router/pkg/k8s/reconciler.go
@@ -18,7 +18,9 @@ package k8s
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -164,17 +166,32 @@ func (r *Reconciler) watchLoop(ctx context.Context) {
 
 // reconcile performs the reconciliation logic
 func (r *Reconciler) reconcile(ctx context.Context) error {
-	// Get IntelligentPool
-	pool, err := r.getIntelligentPool(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to get IntelligentPool: %w", err)
+	// The reconciler emits a single canonical config and so expects exactly
+	// one pool and one route. When several coexist, short-circuiting would
+	// leave every CR with an empty status indefinitely (#1908); instead mark
+	// each conflicting CR Ready=False so the conflict is observable.
+	poolList := &v1alpha1.IntelligentPoolList{}
+	if err := r.client.List(ctx, poolList, client.InNamespace(r.namespace)); err != nil {
+		return fmt.Errorf("failed to list IntelligentPools: %w", err)
+	}
+	routeList := &v1alpha1.IntelligentRouteList{}
+	if err := r.client.List(ctx, routeList, client.InNamespace(r.namespace)); err != nil {
+		return fmt.Errorf("failed to list IntelligentRoutes: %w", err)
 	}
 
-	// Get IntelligentRoute
-	route, err := r.getIntelligentRoute(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to get IntelligentRoute: %w", err)
+	if len(poolList.Items) == 0 {
+		return fmt.Errorf("no IntelligentPool found in namespace %s", r.namespace)
 	}
+	if len(routeList.Items) == 0 {
+		return fmt.Errorf("no IntelligentRoute found in namespace %s", r.namespace)
+	}
+
+	if len(poolList.Items) > 1 || len(routeList.Items) > 1 {
+		return r.reportConflict(ctx, poolList.Items, routeList.Items)
+	}
+
+	pool := &poolList.Items[0]
+	route := &routeList.Items[0]
 
 	// Check if anything changed
 	r.mu.RLock()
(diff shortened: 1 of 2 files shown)


Xunzhuo (approved):
LGTM

Issue #2009, opened 2026-06-02
feature: consume session retention directives (ttl_turns, keep_current_model, prefer_prefix_retention) at runtime


theohsiung:
## Describe the feature

#1747 (PR #1873) landed the `EMIT retention { … }` DSL directive and wired it end-to-end (parser → AST → compiler → decompiler → `config.RetentionDirective` → `ctx.EmittedRetention` → log/trace). Per that issue's MVP scope, **only `drop` is consumed at runtime** (`processor_res_cache_retention.go:30-38`, `ShouldSkipCacheWrite`). The other three fields — `ttl_turns`, `keep_current_model`, `prefer_prefix_retention` — are explicitly **contract-only**: validated and observable, but with no runtime consumer (`decision_config.go:31-39`).

The follow-up consumers named in that struct comment are now shipped (#1742 per-turn session telemetry, #1749 shadow-mode `model_switch_gate`), so the contract-only fields can finally be wired to real behavior:

- **`ttl_turns`** → bound a session-scoped semantic-cache entry's lifetime to N conversation turns, using the per-turn counter from #1742 (`sessiontelemetry`), instead of seconds-only TTL (`GetCacheTTLSecondsForDecision`).
- **`keep_current_model`** → feed the #1749 `model_switch_gate` as a continuity override (suppress a model switch for this turn even when net-switch-advantage is positive).
- **`prefer_prefix_retention`** → bias prefix-cache scoring in `session_aware_scoring.go` (raise `prefixCacheBenefit` / lower switch likelihood) and emit one auditable `x-vsr-retention-*` header so operators can verify the hint.

## Primary layer

**plugin** (decision-driven processing).

**Why this layer:** retention is a structured side-effect of a *matched decision* (the `EMIT` block lives on `Decision`), consumed at the response/cache and selection seams — not signal extraction, and not intentionally cross-cutting. This matches where #1747 placed the contract.

## Why do you need this feature

Today an operator can author `EMIT retention { ttl_turns: 2, keep_current_model: true }` and it validates, round-trips, and shows up in traces — but **silently has no effect** beyond `drop`. That is a confusing half-built contract. Wiring the consumers makes session-aware policies actually control cache lifetime and model continuity, which is the intent of the session-aware workstream (#1753).

## Goals

- [ ] Consume `ttl_turns` at the cache-write seam: cap a session-scoped entry to N turns (verify/extend the cache write path to accept a per-entry TTL override).
[...]

PR #1873, opened 2026-05-05
[Router] Add `EMIT retention` directive to the Routing DSL

e1ijah1:
Closes #1747

## Purpose

- What does this PR change?

This PR adds `EMIT retention { ... }` to the routing DSL and carries the matched directive through config and extproc.

It does four things:

- Adds `EMIT retention` to the DSL parser, AST, compiler, decompiler, AST JSON, and routing YAML round-trip paths.
- Adds `routing.decisions[].emits[]` with one supported kind: `retention`.
- Adds four tri-state retention fields: `drop`, `ttl_turns`, `keep_current_model`, and `prefer_prefix_retention`.
- Enforces one runtime behavior today: `drop=true` skips response-side semantic-cache writes in both non-streaming and streaming paths.

The other fields parse, validate, round-trip, and are observable, but they do not change runtime behavior in this PR.

- Why is this change needed?

Session-aware routing needs a declarative way for a matched decision to say what should be retained after the response.

For this PR, the concrete behavior is simple: a route can say "do not write this response to semantic cache" without hiding that policy in extproc runtime branches.

The wider retention contract also carries follow-up hints needed by session-aware switching:

- `ttl_turns`: future turn-scoped retention lifetime.
- `keep_current_model`: future model-continuity hint for the model-switch gate.
- `prefer_prefix_retention`: future prefix/KV-cache warmth hint.

- Which module(s) does this affect? `Router` / `CLI` / `Dashboard` / `Operator` / `Fleet-Sim` / `Bindings` / `Training` / `E2E` / `Docs` / `CI/Build`

`Router`

## Non-Goals

Out of scope for this PR:

- Provider-specific cache backend behavior.
- The final model-switch gate (#1749).
- Runtime enforcement of `ttl_turns`, `keep_current_model`, or `prefer_prefix_retention`.
- Response headers such as `x-vsr-retention-*`.

Only `drop=true` changes runtime behavior in this PR.

## DSL Example

```dsl
DECISION_TREE complexity_routing {
  IF complexity("task_complexity:easy") {
    NAME "easy-chat-route"
    TIER 1
    MODEL "Qwen/Qwen3.5-27B" (reasoning = false)
    EMIT retention {
      drop: true
    }
  }
  ELSE IF complexity("task_middle:hard") {
    NAME "middle-task-route"
    TIER 2
    MODEL "Qwen/Qwen3.5-35B-A3B" (reasoning = true)
    EMIT retention {
      ttl_turns: 2
      keep_current_model: true
[...]

Issue #1749, opened 2026-04-10
feature: add a shadow-mode model-switch gate for session-aware routing


Xunzhuo:
## Summary

Add a shadow-mode `model_switch_gate` so the router can compare stay-vs-switch decisions with session-aware evidence before enforcing model changes in production.

## Background

The session-aware routing workstream needs one auditable place where the router decides whether it should stay on the current model or switch to a better candidate. That policy seam should land before offline RL and should be able to run in shadow mode so contributors can inspect decisions without immediately changing user-visible routing behavior. This issue is the integration point for the whole workstream.

## Goals

- [ ] Implement a `model_switch_gate` that compares stay-vs-switch tradeoffs using session state, lookup-backed signals, and the current candidate set.
- [ ] Support shadow-mode or dry-run evaluation so operators can inspect would-switch outcomes before enforcement.
- [ ] Integrate with the existing selection and session-affinity surfaces without bypassing the baseline plumbing in #1751 or the CRM baseline in #1752.
- [ ] Add targeted tests for keep-current-model, switch-model, and missing-signal fallback behavior.
- [ ] **Not goals:** mandatory learned policy rollout, provider-specific retention optimization beyond consuming emitted directives, or replacing safe fallbacks.

## Context

**Illustrative shadow output**
```json
{
  "would_switch": true,
  "current_model": "deepseek-v3",
  "candidate_model": "deepseek-r1",
  "stay_cost_estimate": 0.41,
  "switch_cost_estimate": 0.28,
  "handoff_penalty": 0.07,
  "cache_warmth": 0.78
}
```
A contributor does not need to keep this exact payload, but the gate should expose enough information to audit why it would stay or switch.

**Affected files**
- `src/semantic-router/pkg/extproc/req_filter_classification.go` - current `selectModelFromCandidates` call site and a likely integration point.
- `src/semantic-router/pkg/selection/selector.go` - shared selection contracts, including existing session fields.
- `src/semantic-router/pkg/selection/gmtrouter.go` and `src/semantic-router/pkg/selection/rl_driven.go` - current multi-turn and feedback-aware selector surfaces.
- `src/semantic-router/pkg/extproc/req_filter_response_api.go` and `src/semantic-router/pkg/responseapi/**` - response-history or conversation-chain context that may affect gate inputs.
- `docs/agent/plans/pl-0030-session-aware-model-switch-workstream.md` - dependency map for prerequisites.

**Entry points**
[...]

Issue #1747, opened 2026-04-10
feature: add EMIT and retention directives for session-aware routing policies


Xunzhuo:
## Summary

Add `EMIT` and retention directives so session-aware routing policies can return structured side effects instead of burying retention behavior inside runtime branches.

## Background

The local session-aware switching plan does not stop at choosing a model. It also needs a way to say what the runtime should retain, prefer, or release after a decision, especially when cache rebuild cost and context handoff quality matter. The current DSL can declare routing logic but cannot emit structured side effects, and the runtime has no explicit retention-directive contract to consume. This issue creates a narrow, auditable path for policy-driven retention hints without turning extproc orchestrators into a second policy engine.

## Goals

- [ ] Add `EMIT` support to the routing DSL and keep it round-trippable through parser, AST, compiler, and decompiler surfaces.
- [ ] Define a retention-directive contract that can express keep, drop, or prefer-retain style hints for session-aware routing.
- [ ] Thread emitted directives through runtime surfaces that already own response or cache post-processing.
- [ ] Add targeted tests showing that emitted directives survive compile-time validation and runtime consumption without widening orchestrator hotspots.
- [ ] **Not goals:** implementing every provider-specific cache-retention backend or folding the final switch gate into this issue.

## Context

**Illustrative directive shape**
```dsl
EMIT retention {
  keep_current_model: true
  prefer_prefix_retention: true
  ttl_turns: 2
}
```
This is an example only. The important contributor-facing constraint is that emitted side effects remain structured and auditable instead of becoming free-form scripting.

**Affected files**
- `src/semantic-router/pkg/dsl/ast.go`, `src/semantic-router/pkg/dsl/parser.go`, `src/semantic-router/pkg/dsl/compiler.go`, and `src/semantic-router/pkg/dsl/decompiler.go` - new authoring surface for `EMIT`.
- `src/semantic-router/pkg/extproc/processor_res_cache.go` - existing cache-write seam that should own retention-related response-side behavior.
- `src/semantic-router/pkg/extproc/processor_res_body_pipeline.go` and `src/semantic-router/pkg/extproc/processor_res_body_streaming.go` - response pipeline edges where emitted directives may need to flow.
- `src/semantic-router/pkg/extproc/request_context.go` - likely transport seam for carrying emitted runtime hints.

**Entry points**
- Start at the DSL contract and keep provider-specific effects behind narrow runtime helpers.
[...]

PR #2016, opened 2026-06-03
[Router] consume session retention directives (ttl_turns, keep_current_model, prefer_prefix_retention)

theohsiung:
## Summary

Wires the three previously contract-only `EMIT retention` fields into runtime, continuing the MVP from #1747 / #1873 (which consumed only `drop`). Closes the consumption gap tracked in #2009. The now-shipped #1742 (per-turn telemetry) and #1749 (model-switch gate) unblocked these consumers.

| Field | New runtime behavior |
|-------|----------------------|
| `ttl_turns` (`> 0`) | Overrides the decision/global semantic-cache TTL for the entry, scoping it to ~N future turns (turns→seconds at the cache-write seam). |
| `keep_current_model: true` | Forces the session-aware `model_switch_gate` to stay on the current model, **regardless of gate mode** (shadow/enforce), when the previous model is known and a valid candidate. Auditable via `reason=retention_keep_current_model`. |
| `prefer_prefix_retention` | Emitted to the inference pool as an `x-vsr-retention-prefer-prefix` response header (the router-emitted-directive mechanism from the WRP vision paper, Opportunity 15). |

Every explicitly-set field is also emitted as an `x-vsr-retention-*` response header (success, non-cache-hit only) so the pool/operators can audit the router's retention intent over the wire.

## Commit trajectory

1. `[Router] consume retention ttl_turns as per-entry cache TTL` — `applyRetentionTTLOverride` on the retention seam; both streaming + non-streaming cache writes. `drop` still takes precedence (mutually exclusive by validation).
2. `[Router] honor retention keep_current_model in model_switch_gate` — `ForceKeepCurrentModel` on `ModelSwitchGateInput` + short-circuit forced stay.
3. `[Router] emit retention directive as x-vsr-retention-* response headers` — new header constants + tri-state-aware emission.
4. `[Router/Docs] document consumed retention fields` — `retention.md` + struct/contract comment updates.

## Out of scope (deferred follow-ups, noted in code + docs)

- **`prefer_prefix_retention` scoring-weight bias** in `session_aware_scoring`: deferred because it would inject a magic constant into calibrated hotspot weights and needs a separate tuning rationale. `keep_current_model` already provides the hard form of prefix retention.
- Promoting the `ttl_turns` seconds-per-turn factor (currently a `60` constant) to a `global.router` config knob.
- Honoring `keep_current_model` when the model-switch gate is disabled.
- Auto-deriving directives from semantic signals (`previous_response_id` / system-prompt fingerprint / tool type) and real vLLM/KV-cache backend eviction integration.

## Test Plan

- [x] TDD (RED→GREEN) for every field consumer:
  - `processor_res_cache_retention_test.go` — `retentionTTLOverrideSeconds` / `applyRetentionTTLOverride` + both cache-write seams apply the override; `drop` precedence preserved.
  - `selection/model_switch_gate_test.go` — forced stay in shadow mode; audit-only when previous model unknown / not a candidate; no-op when gate disabled.
  - `extproc/model_switch_gate_test.go` — wrapper threads `keep_current_model` and overrides even in shadow mode.
  - `processor_res_header_mutation_test.go` — headers emitted for set fields; unset fields omitted (tri-state); skipped on cache hit.
- [x] `go test -race ./pkg/{extproc,selection,config,headers}` — all pass.
- [x] `go build ./...`, `gofmt`, `go vet` clean on touched packages.
- [x] `make agent-ci-gate` — all checks pass (go fmt, go lint, AST security scan, changed-files lint). The local `md-fmt` hook errored only because the `markdownlint` binary isn't installed locally (exit 127); `retention.md` was verified clean via `npx markdownlint-cli` against the repo config (MD013 line-length is disabled).

Refs #2009. Builds on #1747 / #1873; consumers unblocked by #1742, #1749.

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_res_header_mutation.go
+++ b/src/semantic-router/pkg/extproc/processor_res_header_mutation.go
@@ -238,7 +238,34 @@ func buildResponseHeaderMutation(
 
 	addStandardDecisionHeaders(builder, ctx)
 	addMatchedSignalHeaders(builder, ctx)
+	addRetentionDirectiveHeaders(builder, ctx)
 	return builder.mutation()
+}
+
+// addRetentionDirectiveHeaders emits the matched decision's EMIT retention
+// directive to the response as x-vsr-retention-* headers so the inference pool
+// and operators can observe the router's retention intent at the wire
+// (issue #2009). Only fields the directive explicitly set are emitted, mirroring
+// the tri-state semantics of config.RetentionDirective; an unset field is
+// omitted rather than sent as a default. Emitted only on successful,
+// non-cache-hit responses (same gate as the standard decision headers).
+func addRetentionDirectiveHeaders(builder *responseHeaderMutationBuilder, ctx *RequestContext) {
+	r := ctx.EmittedRetention
+	if r == nil {
+		return
+	}
+	if r.Drop != nil {
+		builder.addBool(headers.VSRRetentionDrop, *r.Drop)
+	}
+	if r.TTLTurns != nil && *r.TTLTurns > 0 {
+		builder.addInt(headers.VSRRetentionTTLTurns, *r.TTLTurns)
+	}
+	if r.KeepCurrentModel != nil {
+		builder.addBool(headers.VSRRetentionKeepCurrentModel, *r.KeepCurrentModel)
+	}
+	if r.PreferPrefixRetention != nil {
+		builder.addBool(headers.VSRRetentionPreferPrefix, *r.PreferPrefixRetention)
+	}
 }
 
 // addStandardDecisionHeaders adds the per-request decision headers
--- a/src/semantic-router/pkg/extproc/processor_res_header_mutation.go
+++ b/src/semantic-router/pkg/extproc/processor_res_header_mutation.go
@@ -79,6 +79,16 @@ func (builder *responseHeaderMutationBuilder) addInt(key string, value int) {
 	builder.addString(key, strconv.Itoa(value))
 }
 
+// addNonNegativeInt emits the value when it is >= 0, so an explicit zero is
+// still written. Mirrors addNonNegativeFloat; used for tri-state retention
+// fields where 0 is a meaningful, explicitly-set value (not "unset").
+func (builder *responseHeaderMutationBuilder) addNonNegativeInt(key string, value int) {
+	if value < 0 {
+		return
+	}
+	builder.addString(key, strconv.Itoa(value))
+}
+
 func (builder *responseHeaderMutationBuilder) addJoined(key string, values []string) {
 	if len(values) == 0 {
 		return
@@ -257,8 +267,11 @@ func addRetentionDirectiveHeaders(builder *responseHeaderMutationBuilder, ctx *R
 	if r.Drop != nil {
 		builder.addBool(headers.VSRRetentionDrop, *r.Drop)
 	}
-	if r.TTLTurns != nil && *r.TTLTurns > 0 {
-		builder.addInt(headers.VSRRetentionTTLTurns, *r.TTLTurns)
+	if r.TTLTurns != nil {
+		// Tri-state: emit whenever explicitly set, including ttl_turns: 0
+		// (a valid no-op the validator permits). The runtime TTL override
+		// still applies only when > 0; the header reflects intent, not effect.
+		builder.addNonNegativeInt(headers.VSRRetentionTTLTurns, *r.TTLTurns)
 	}
 	if r.KeepCurrentModel != nil {
 		builder.addBool(headers.VSRRetentionKeepCurrentModel, *r.KeepCurrentModel)
--- a/src/semantic-router/pkg/headers/headers.go
+++ b/src/semantic-router/pkg/headers/headers.go
@@ -90,6 +90,17 @@ const (
 	// Value: "true"
 	VSRCacheHit = "x-vsr-cache-hit"
 
+	// Retention directive headers expose the matched decision's EMIT retention
+	// block to the inference pool / operators (issue #2009). They are emitted
+	// only on successful, non-cache-hit responses, and only for the fields the
+	// directive explicitly set (tri-state). They let the pool observe the
+	// router's KV/cache retention intent at the wire; the router itself also
+	// consumes drop (cache-write skip) and ttl_turns (per-entry cache TTL).
+	VSRRetentionDrop             = "x-vsr-retention-drop"
+	VSRRetentionTTLTurns         = "x-vsr-retention-ttl-turns"
+	VSRRetentionKeepCurrentModel = "x-vsr-retention-keep-current-model"
+	VSRRetentionPreferPrefix     = "x-vsr-retention-prefer-prefix"
+
 	// RouterReplayID carries the identifier for a captured replay record.
 	// Value: opaque replay token
 	RouterReplayID = "x-vsr-replay-id"
--- a/dashboard/frontend/src/components/HeaderDisplay.tsx
+++ b/dashboard/frontend/src/components/HeaderDisplay.tsx
@@ -143,6 +143,23 @@ const HEADER_INFO: Record<string, { label: string; type: 'info' | 'success' | 'w
     label: 'Algorithm',
     type: 'info',
   },
+  // Retention directive headers (issue #2009)
+  'x-vsr-retention-drop': {
+    label: 'Retention: Drop',
+    type: 'warning',
+  },
+  'x-vsr-retention-ttl-turns': {
+    label: 'Retention TTL (turns)',
+    type: 'info',
+  },
+  'x-vsr-retention-keep-current-model': {
+    label: 'Keep Current Model',
+    type: 'info',
+  },
+  'x-vsr-retention-prefer-prefix': {
+    label: 'Prefer Prefix',
+    type: 'info',
+  },
 }
 
 function shouldSummarizeHeaderValue(key: string, values: string[]): boolean {
--- a/dashboard/frontend/src/components/HeaderReveal.tsx
+++ b/dashboard/frontend/src/components/HeaderReveal.tsx
@@ -150,6 +150,23 @@ const HEADER_INFO: Record<string, { label: string; description: string }> = {
     label: 'Algorithm',
     description: 'The multi-model algorithm used (confidence, ratings)',
   },
+  // Retention directive headers (issue #2009)
+  'x-vsr-retention-drop': {
+    label: 'Retention: Drop',
+    description: 'Matched decision asked to skip the semantic-cache write',
+  },
+  'x-vsr-retention-ttl-turns': {
+    label: 'Retention TTL (turns)',
+    description: 'Per-entry semantic-cache lifetime override, in conversation turns',
+  },
+  'x-vsr-retention-keep-current-model': {
+    label: 'Keep Current Model',
+    description: 'Forces the model-switch gate to stay on the current model',
+  },
+  'x-vsr-retention-prefer-prefix': {
+    label: 'Prefer Prefix',
+    description: 'Hints the inference pool to keep the prompt prefix / KV cache warm',
+  },
 }
 
 const HeaderReveal = ({ headers, onComplete, displayDuration = 2000 }: HeaderRevealProps) => {
@@ -200,6 +217,10 @@ const HeaderReveal = ({ headers, onComplete, displayDuration = 2000 }: HeaderRev
     looper: displayHeaders.filter(([key]) =>
       key.startsWith('x-vsr-looper-')
     ),
+    // Retention directive headers: all x-vsr-retention-*
+    retention: displayHeaders.filter(([key]) =>
+      key.startsWith('x-vsr-retention-')
+    ),
   }
 
   const renderSection = (title: string, items: [string, string][], isPrimary = false) => {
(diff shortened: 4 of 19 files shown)


theohsiung:
The two red checks here are **transient infrastructure flakes**, unrelated to this change (which only touches `pkg/{extproc,selection,config,headers}` + one markdown doc — nothing in `candle-binding`, BERT, Docker, or model downloads):

- **`test-and-build` → Run semantic router tests** and **`pre-commit … changed-files lint`**: both fail because the runner hit a HuggingFace rate limit downloading the embedding model —
  ```
  Failed to initialize BERT: request error:
  https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/resolve/main/config.json: status code 429
  ```
  Every `--- FAIL` in `candle-binding` is the resulting `failed to initialize BERT similarity model`, and the Ginkgo `[BeforeSuite]` aborts for the same reason. (The `redis … connection refused` and Helm `yamllint` lines are expected/handled; the suite otherwise reports `276 Passed | 0 Failed`.)
- **`build_pr (llm-katan)` → Set up Docker Buildx** earlier failed on a Docker Hub timeout during BuildKit bootstrap (`registry-1.docker.io … context deadline exceeded`) and has already recovered on the latest run.

Could a maintainer please **re-run the failed jobs**? I don't have permission to re-run workflows on this repo from my fork. Happy to rebase onto latest `main` if that helps the queue.

(If the HF 429s are recurring in CI, a durable fix would be an authenticated `HF_TOKEN` pull or caching the model between runs — but that's a separate CI/Build change, out of scope for this PR.)

theohsiung:
Update after rebasing onto latest `main`: the `pre-commit … changed-files lint` and `test-and-build` checks are still red, but the cause is **not this change** — every pre-commit hook passes (`go fmt`, `go lint`, `md fmt`, AST scan, `agent changed-files lint` all ✅). The only failure is a HuggingFace rate limit while downloading the embedding model in pre-existing tests I don't touch:

```
Failed to initialize BERT: request error:
https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/resolve/main/config.json: status code 429
[BeforeSuite] [FAILED] failed to initialize BERT similarity model
cache_test.go:3236: Failed to initialize BERT model: ...
```

This PR changes Go files under `pkg/extproc`, so the harness pulls in the `pkg/cache` BERT-similarity tests, which then get 429'd. So a plain re-run will likely keep failing while HF is rate-limiting the runners — it's not a one-off flake for model-downloading test runs. (Note: in the same suite, Redis-unavailable is handled with a graceful skip, whereas the model-download failure hard-fails the `BeforeSuite`.)

@rootfs @Xunzhuo — how would you like to proceed? A few options:
1. Set `HF_TOKEN` for CI (authenticated pulls have much higher limits) or cache the model between runs — durable fix.
2. Re-run the failed jobs once HF rate-limiting eases (I can't re-run from a fork).
3. I'm happy to open a separate **[CI/Build]** PR making the embedding-model `BeforeSuite` skip-on-download-failure (mirroring the existing Redis graceful-skip), which would stop HF 429s from blocking Go PRs generally.

Whichever you prefer — just let me know. The change itself is verified locally: `go test -race ./pkg/{extproc,selection,config,headers}` all pass, `go build ./...`, `gofmt`, `go vet` clean.

theohsiung:
Further evidence — the HF rate-limiting is **CI-wide**, not just unit tests. `integration-test (kubernetes)` also fails for the same reason, this time at **runtime inside the deployed router pod**: on startup it pulls the mmBERT models from HuggingFace, gets 429, fails embedding init, and `CrashLoopBackOff`s until the Helm install times out:

```
[router-live] 429 Too Many Requests for url:
  https://huggingface.co/api/models/llm-semantic-router/mmbert32k-jailbreak-detector-merged/revision/main
[router-live] ERROR: Failed to register mmBERT model: FileNotFound("models/mmbert-embed-32k-2d-matryoshka/config.json")
[router-live] embedding_models_init_completed  embedding_ready=false tools_ready=false
[router-live] semantic-router-... 0/1 CrashLoopBackOff (15 restarts)
Error: context deadline exceeded
Error: failed to setup profile: install semantic-router: failed to install/upgrade chart: exit status 1
```

So all three reds are external-registry rate-limits / timeouts, none related to this change:
- `pre-commit … lint` — unit test pulls `all-MiniLM-L6-v2` → HF 429
- `integration-test (kubernetes)` — deployed router pod pulls mmBERT models → HF 429 → CrashLoopBackOff → Helm timeout
- `build_pr (vllm-sr-sim)` — buildx bootstrap pulls `moby/buildkit` from Docker Hub → `registry-1.docker.io … context deadline exceeded`

@rootfs @Xunzhuo since this affects the E2E deploy path too, an `HF_TOKEN` (or pre-baked / cached models) would need to reach the router pod as well, not just the unit-test runners. Happy to help however is easiest — re-run once HF eases, or I can open a separate **[CI/Build]** PR (model caching / `HF_TOKEN` wiring / skip-on-download-failure). The change itself remains green locally (`go test -race ./pkg/{extproc,selection,config,headers}`, `go build ./...`, `gofmt`, `go vet`).

rootfs (commented):
## Verdict
request_changes

The core runtime wiring looks solid, but the PR leaves one contract regression and one response-header surface gap that should be fixed before merge.

## Risk
- `ttl_turns` no longer preserves the advertised tri-state/audit contract on the wire: `addRetentionDirectiveHeaders` only emits the header when `*r.TTLTurns > 0`, so an explicit `ttl_turns: 0` is silently dropped even though the PR/docs say every explicitly set retention field is emitted. See `src/semantic-router/pkg/extproc/processor_res_header_mutation.go:252-262`, `src/semantic-router/pkg/extproc/processor_res_header_mutation.go:75-80`, `website/docs/tutorials/decision/retention.md:48-50`.
- New `x-vsr-retention-*` headers are added server-side, but the dashboard still neither collects nor renders them, so the new operator-visible contract is effectively invisible in the existing reveal/display UX. See `src/semantic-router/pkg/headers/headers.go:93-102`, `src/semantic-router/pkg/extproc/processor_res_header_mutation.go:245-268`, `dashboard/frontend/src/components/chatRequestSupport.ts:18-53`, `dashboard/frontend/src/components/chatRequestSupport.ts:110-120`, `dashboard/frontend/src/components/HeaderDisplay.tsx:8-146`, `dashboard/frontend/src/components/HeaderReveal.tsx:11-203`.

## Tests
- Good targeted unit coverage on the touched Go paths: `src/semantic-router/pkg/extproc/processor_res_cache_retention_test.go:291-385`, `src/semantic-router/pkg/extproc/model_switch_gate_test.go:84-103`, `src/semantic-router/pkg/selection/model_switch_gate_test.go:204-280`, `src/semantic-router/pkg/extproc/processor_res_header_mutation_test.go:128-174`.
- Still missing behavior-level coverage for the new routing/header contract. The harness requires E2E updates for behavior-visible routing/API changes, and this PR changes both model-stay behavior and emitted response headers. See `docs/agent/testing-strategy.md:65-71`, `src/semantic-router/pkg/selection/model_switch_gate.go:106-116`, `src/semantic-router/pkg/extproc/processor_res_header_mutation.go:245-268`.

## Style/Hygiene
- Mostly clean and well-factored. The only nit is comment churn: some long contract comments in `src/semantic-router/pkg/extproc/processor_res_cache_retention.go:42-71` and `website/docs/tutorials/decision/retention.md:33-50` are fairly policy-heavy and may age quickly.

## Size
- Scope is reasonable for one PR, but because it introduces a new `response_headers` contract, the dashboard/header-surface updates should land in the same change rather than as a follow-up.

src/semantic-router/pkg/extproc/processor_res_header_mutation.go
```diff
@@ -238,9 +238,36 @@ func buildResponseHeaderMutation(
...
+// non-cache-hit responses (same gate as the standard decision headers).
+func addRetentionDirectiveHeaders(builder *responseHeaderMutationBuilder, ctx *RequestContext) {
+	r := ctx.EmittedRetention
+	if r == nil {
+		return
+	}
+	if r.Drop != nil {
+		builder.addBool(headers.VSRRetentionDrop, *r.Drop)
+	}
+	if r.TTLTurns != nil && *r.TTLTurns > 0 {
```
rootfs: This `> 0` guard drops an explicit `ttl_turns: 0`, so the response header no longer preserves the directive's tri-state semantics even though the PR/docs say every explicitly set field is emitted. If `0` is intended to mean "set but no-op", emit it here and update the tests/docs to match.

src/semantic-router/pkg/headers/headers.go
```diff
@@ -90,6 +90,17 @@ const (
 	// Value: "true"
 	VSRCacheHit = "x-vsr-cache-hit"
 
+	// Retention directive headers expose the matched decision's EMIT retention
+	// block to the inference pool / operators (issue #2009). They are emitted
+	// only on successful, non-cache-hit responses, and only for the fields the
+	// directive explicitly set (tri-state). They let the pool observe the
+	// router's KV/cache retention intent at the wire; the router itself also
+	// consumes drop (cache-write skip) and ttl_turns (per-entry cache TTL).
+	VSRRetentionDrop             = "x-vsr-retention-drop"
```
rootfs, line 99: These new `x-vsr-retention-*` constants expand the `response_headers` contract, but the PR doesn't update the dashboard header allowlists/renderers (`chatRequestSupport`, `HeaderDisplay`, `HeaderReveal`). Please wire those surfaces in the same change so the new operator-visible headers are actually visible in the existing UI.

theohsiung:
Thanks for the thorough review @rootfs — all four points are addressed in four new commits, rebased onto latest `main`. Breakdown below.

## 1. `ttl_turns` tri-state regression (blocking) — fixed
`4b793ec` `[Router] emit explicitly-set retention ttl_turns: 0 header (tri-state fix)`

You're right — gating on `*r.TTLTurns > 0` broke the advertised contract. `ttl_turns: 0` is a valid, explicitly-set value: the DSL/config validator only *warns* on `ttl_turns: 0` (no-op hint), it doesn't reject it — only negatives are rejected. The header is now emitted whenever `TTLTurns != nil`, via a new `addNonNegativeInt` builder helper that mirrors the existing `addNonNegativeFloat`, so an explicit `0` surfaces as `x-vsr-retention-ttl-turns: 0`. The runtime TTL override still applies only for `> 0` — the header reflects *intent*, the override reflects *effect*.

Added regression tests for both the explicit-zero `ttl_turns` and the explicit-`false` bools (`drop` / `keep_current_model`), so the tri-state contract is now covered symmetrically across every field, not just `ttl_turns`.

## 2. Dashboard surface gap (blocking) — landed in this PR
`1d19a3e` `[Dashboard] surface x-vsr-retention-* response headers in playground UX`

Agreed this should ship with the contract rather than as a follow-up. All four `x-vsr-retention-*` headers are now wired through the existing surfaces:
- `chatRequestSupport.ts` — added to `RESPONSE_HEADER_KEYS` (collection)
- `HeaderDisplay.tsx` — chip entries
- `HeaderReveal.tsx` — a new `RETENTION` group + section, mirroring the signal/looper treatment

(Header strings were cross-checked against the `headers.go` constants to avoid a silent name mismatch.)

## 3. Behavior-level / E2E coverage — added
`679e1e3` `[E2E] add retention-directive response-header contract testcase`

New `retention-directive` testcase in the kubernetes baseline profile. It adds a keyword-gated `retention_probe_decision` (a deterministic sentinel keyword, so it doesn't depend on a classifier model) that emits `{ ttl_turns: 0, keep_current_model: true, prefer_prefix_retention: true }`, routes a request to it, and asserts all three `x-vsr-retention-*` headers — including the explicit `ttl_turns: 0` from point 1 — plus that the unset `drop` header is *absent*. It guards on `x-vsr-selected-decision == retention_probe_decision` first, so the header checks can't pass vacuously. Registered in `BaselineRouterContract` next to the existing keyword-gated `tool-selection` testcase (same profile + model, no new model/HF dependency).

Verified locally against the kind `kubernetes` profile:

```
[Runner] Running test: retention-directive
[Test]   x-vsr-retention-ttl-turns = "0" ✓
[Test]   x-vsr-retention-keep-current-model = "true" ✓
[Test]   x-vsr-retention-prefer-prefix = "true" ✓
[Test] Retention directive response headers verified ✓
[Runner] ✅ Test retention-directive passed (2.447s)
Total: 1 | Passed: 1 | Failed: 0
```

One scoping question for you: this E2E covers the **emitted-header** half of the contract for all three fields. The **model-stay** runtime behavior (`keep_current_model` forcing a stay through the gate) is hard to assert deterministically end-to-end with the mock backends, so it currently stays covered by the unit tests added earlier (`selection/model_switch_gate_test.go`, `extproc/model_switch_gate_test.go`). I'm happy to extend the E2E to a multi-turn stay assertion in this PR if you'd prefer that — just say the word.

## 4. Comment / doc hygiene (nit) — done
`110c56a` `[Router/Docs] tighten retention comments and clarify ttl_turns header contract`

Trimmed the policy-heavy comments on `retentionDefaultSecondsPerTurn` / `retentionTTLOverrideSeconds` to the load-bearing facts, and made `retention.md` precise that `x-vsr-retention-ttl-turns` is emitted whenever the field is explicitly set (including `0`), as distinct from the `> 0` runtime override.

---

All four commits build clean and pass `gofmt` / `go vet`, `go test -race ./pkg/{extproc,selection,config,headers}`, the dashboard `type-check` / `lint` / unit tests, and the e2e build. Let me know if you'd like any of the above scoped differently.

rootfs:
@theohsiung great! please rebase and make sure CI is green, it is ready to merge!

theohsiung:
@rootfs synced onto the latest `main` and CI is now all green ✅ — ready to merge whenever you are. Thanks!

PR #2017, opened 2026-06-03
feat(cli): support custom container DNS via VLLM_SR_DNS

WUKUNTAI-0211:
## What

Adds an optional `VLLM_SR_DNS` environment variable (comma-separated) that injects `--dns` flags into the `docker run` command for the router, envoy, and sim containers.

## Why

When the router/envoy containers need to reach an upstream that is only resolvable via a private/corporate DNS server (e.g. an internal gateway domain), docker's default resolver may fall back to public DNS, which has no record for it. Envoy's `LOGICAL_DNS` cluster then fails to resolve the hostname and reports `no healthy upstream`.

Until now the only workaround was to post-process the generated `envoy.yaml` (rewriting hostnames to literal IPs), which is brittle, hard-codes the upstream IP, and breaks the upstream `Host` header used for vhost routing. Pointing the containers at the appropriate DNS server lets clusters keep using domain names (correct `Host` header, IP changes handled by DNS).

## How

Follows the existing `VLLM_SR_*` env-driven docker-run pattern (cf. `VLLM_SR_AMD_GPU_PASSTHROUGH`, `VLLM_SR_NOFILE_LIMIT`, `append_host_gateway`). A new `append_custom_dns()` helper appends `--dns <server>` for each entry in `VLLM_SR_DNS`, invoked right after `append_host_gateway()` in `_build_service_run_command()`.

## Usage

```bash
export VLLM_SR_DNS="10.0.0.53,10.0.0.54"
```

## Backward compatibility

Fully backward compatible — when `VLLM_SR_DNS` is unset or empty, no `--dns` flag is added and behavior is unchanged.

## Tests

Adds `tests/test_docker_run_command.py` (7 cases): unset/empty no-op, single/multiple servers, whitespace & blank trimming, command-list preservation, and non-interference with `--add-host`. All pass; existing docker tests unaffected.


Xunzhuo (approved):
LGTM

Issue #2018, opened 2026-06-03
[Feature] Enforce bounded per-session inference token budgets with reservation and reconciliation


theohsiung:
### Goal

Enforce a bounded per-session inference-token budget using deterministic pre-dispatch reservation and post-response reconciliation.

Parent Epic: #2973  
Related exact accounting: #3050

### Proposed Workgroup

Agentic & Context

### Current-main baseline

Current main records bounded local or optional shared session summaries, including prompt, cached, output, cost, current-model, and switch information. Actual provider token usage normally arrives only after a response, while pre-dispatch decisions rely on conservative estimates.

This feature is a long-session inference control. It is not API-key quota, billing, organization policy, or token rate limiting.

### Scope

- Define a disabled-by-default per-session inference-token ceiling and bounded reservation state.
- Reserve conservative prompt plus output demand before dispatch and reconcile with authoritative or derived usage after completion.
- Handle concurrent requests for one session without silently exceeding the locally enforceable ceiling.
- Consume only explicitly configured recipe actions, such as reducing an output reserve, invoking an already-supported compression policy, choosing within the existing eligible model set, or returning a defined terminal response.
- Record estimate source, reservation, reconciliation, selected action, remaining budget, and terminal reason in bounded diagnostics.
- Specify single-process behavior first and make shared-store consistency an explicit supported profile.

### Non-goals

- API-key, organization, tenant, quota, billing, spend, or token rate-limit management.
- Add candidates outside the configured eligible model set.
- Let the model invent budget policy or degradation actions.
- Revive the removed model_switch_gate configuration.
- Add synchronous durable-store calls without a bounded supported contract.

### Completion signal

- [ ] Disabled configuration preserves current behavior.
- [ ] Reservation and reconciliation are deterministic for authoritative, derived, missing, and partial usage.
- [ ] Concurrent session requests have an explicit oversubscription bound.
- [ ] Every degradation action comes from validated recipe policy.
- [ ] Safety, authorization, capability, and portability constraints dominate budget objectives.
- [ ] E2E coverage includes normal, degraded, refused, cancellation, retry, and reconciliation paths.

Xunzhuo:
lets move this into v0.4

Xunzhuo:
Thanks @SanjayShekhar — one implementation boundary to keep explicit: please reuse the existing recipe-namespaced Router session key and generation-owned session-state store. This is not a rate-limit or API-key quota path and should not introduce a second session identity or store.

[...]

Issue #1456, opened 2026-03-06
[Security] Cap Looper fan-out and total upstream-call amplification


yossiovadia:
## Audited upstream baseline (2026-07-15)

Current `main` validates that every ReMoM `breadth_schedule` entry is positive, and Looper algorithms expose concurrency/time/error controls. However, there is no repository-wide upper bound on:

- the sum of a ReMoM breadth schedule;
- the number of decision `modelRefs` consumed by algorithms such as ratings/confidence/fusion; or
- the total upstream calls a single request can create across workflow/Looper stages.

Concurrency limits protect simultaneous load but do not bound total spend or amplification. For example, ReMoM appends a final synthesis round, so `breadth_schedule: [100]` can still request roughly 101 model calls before retries or other stages.

## Scope

1. Define one explicit per-request amplification budget and conservative defaults.
2. Validate static configs against the budget when the call count is knowable.
3. Enforce a runtime counter for paths whose count depends on execution.
4. Count retries, planner/final calls, fan-out, and fallback attempts consistently.
5. Stop safely with a typed reason when the budget is exhausted.
6. Expose estimated and actual call counts without leaking prompt content.

## Boundaries

- Keep concurrency and total-call budgets separate.
- Do not rely on warnings alone for unsafe configurations.
- Do not silently truncate a configured algorithm and return a misleading success.
- Per-user monetary budgets are related but are not required to establish the per-request safety invariant.

## Acceptance criteria

- [ ] Oversized ReMoM schedules and model lists fail validation with actionable errors.
- [ ] Dynamic execution cannot exceed the configured hard call budget.
- [ ] Internal retries and every Looper/Flow/Fusion/ReMoM leg are counted.
- [ ] Diagnostics identify the decision, algorithm, limit, and observed count without sensitive content.
- [ ] Tests cover exact-boundary, over-boundary, retry, cancellation, and nested/multi-stage cases.
- [ ] Existing safe configs retain compatible behavior.

## Relevant surfaces

`src/semantic-router/pkg/config/remom_config.go`, decision/algorithm validation, `src/semantic-router/pkg/looper/`, request accounting, Replay/metrics, and E2E tests.

Xunzhuo:
Maintainer cleanup: moving this security/cost-control issue out of v0.3 without closing it. It should remain open for explicit verification and hardening rather than silently blocking the completed Themis milestone.

Xunzhuo:
[...]

PR #2022, opened 2026-06-03
[Router] session token budget: config, evaluation, and terminate (Opp-5 PR 1/2)

theohsiung:
## Summary

Implements **WRP Opportunity 5 — runtime per-session token-budget enforcement** at the dispatch layer, with a graduated response ladder instead of a binary deny. Closes the first half of #2018.

For each active session the router already tracks cumulative tokens (`pkg/sessiontelemetry`). This PR adds a decision point that compares that cumulative count against a configured budget and selects a graduated stage:

```
none → shape tools → compress → downgrade → terminate
```

**This PR (1 of 2)** lands the foundation and the decisive stage:

- **`pkg/sessionbudget`** — a pure, I/O-free evaluator: `Evaluate(cumulative, budget, thresholds) → (Stage, ratio)`. Thresholds are ascending multipliers of the budget; defaults map the vision paper's "3× expected budget → graduated response" to the terminate point (`shape 1.0 / compress 1.5 / downgrade 2.0 / terminate 3.0`).
- **Config** — opt-in, tri-state `session_token_budget` under `global.services` (wired through the canonical config contract both ways + reference `config/config.yaml`). A complete no-op unless `enabled: true` **and** `budget_tokens > 0`; validation rejects negative/non-ascending thresholds and `enabled` with a zero budget.
- **Evaluation** — `evaluateSessionBudget` runs on the request-body path after session-transition fields are populated, reading the cumulative counters with zero extra cost via `GetRouterSessionSnapshot`. Emits `vsr_session_budget_{evaluations_total,ratio}` metrics.
- **Terminate** — the only enforced stage in this PR: an over-budget-terminate session is short-circuited with an immediate OpenAI-compatible `429` of type `budget_exceeded` (mirroring the fast-response path), releasing the in-flight admission first.
- **Headers** — `x-vsr-budget-stage` / `-ratio` on normal responses; `x-vsr-budget-exceeded: true` on the terminate response.
- **Docs** — a `global` tutorial covering the contract, ladder, headers, metrics, and false-positive guidance.

**Out of scope (follow-up PR 2):** the soft ladder *actions* — shape tools, compress, downgrade. They are evaluated and reported (headers/metrics) here, but a session in those bands proceeds to normal routing until PR 2 wires the actions into `req_filter_tools.go`, `pkg/promptcompression`, and `model_switch_gate`.

**Deferred (per #2018):** per-`(domain, turn)` P90 percentile budget model, system-prompt-fingerprint prior, per-caller cross-session budgets, model-aware compression.

## Why

Agent workflows face a token-budget explosion (3–10× more LLM calls than chat, quadratic context growth, 20–40% of tokens spent on failed attempts, up to 50× on retry loops). Existing controls are insufficient at this layer: `pkg/ratelimit` does a binary per-user/model deny on a sliding window — not a per-active-session cumulative budget — and #1456 only caps looper breadth. The router is the natural enforcement point because it sees the actual token flow across all turns.

## Design notes

- **Soft stages are cumulative; terminate is exclusive.** `Evaluate` returns the highest triggered stage; the applicator (PR 2) applies all soft bands up to it, while terminate short-circuits.
- **Static budget MVP.** The percentile/fingerprint budget prior is deferred; until then prefer a conservative ceiling. The false-positive risk (terminating a legitimately long session) is mitigated by opt-in defaults and putting terminate last.
- **Narrow seam.** Evaluation/terminate live in a dedicated `pkg/extproc/session_budget.go`; `request_context.go` grows by a single pointer field (`SessionBudget *SessionBudgetDecision`).

## Test Plan

- [x] `pkg/sessionbudget` — table-driven stage/threshold/ratio tests (`go test -race`)
- [x] `pkg/config` — validator specs (opt-in, negative/non-ascending/zero-budget rejection) + canonical round-trip + reference-config public-surface coverage
- [x] `pkg/extproc` — `evaluateSessionBudget` (disabled/zero/empty-session no-ops, under-budget, terminate band) + `maybeTerminateForBudget` (429 `budget_exceeded`, headers) + response-header emission
- [x] `go test -race` green across `pkg/sessionbudget`, `pkg/config`, `pkg/extproc`, `pkg/headers`, `pkg/observability/metrics`
- [x] `go vet`, `golangci-lint` (repo config), `gofmt`, `markdownlint`, docs-contract tests all green
- [ ] E2E: not required — terminate is a new opt-in path disabled by default; behavior-visible E2E lands with the soft ladder in PR 2


rootfs (commented):
request_changes

The core design is solid, but the threshold validator currently allows configurations that become non-ascending after default resolution, which can silently change stage behavior versus the documented contract.

**Risk**

- `validateSessionTokenBudget` skips zero-valued thresholds when enforcing ordering (`src/semantic-router/pkg/config/validator_session_budget.go:50`), while runtime fills zeros from defaults (`src/semantic-router/pkg/sessionbudget/budget.go:61`). This permits configs like `shape_tools: 2.0, compress: 0` that validate but resolve to `shape_tools=2.0, compress=1.5` (non-ascending), producing unexpected stage selection at runtime.
- This is a behavior-visible routing path (early `429 budget_exceeded` terminate) introduced in request processing (`src/semantic-router/pkg/extproc/session_budget.go:80`, `src/semantic-router/pkg/extproc/processor_req_body_prepare.go:214`), so misvalidation here directly affects enforcement semantics.

**Tests**

- Good unit coverage exists for evaluator, validator, and terminate path (`src/semantic-router/pkg/sessionbudget/budget_test.go:5`, `src/semantic-router/pkg/config/validator_session_budget_test.go:8`, `src/semantic-router/pkg/extproc/session_budget_test.go:35`).
- Missing test for “mixed explicit + default thresholds” ordering (e.g., explicit `shape_tools` with default `compress`) in validator tests (`src/semantic-router/pkg/config/validator_session_budget_test.go:56`), which is where the current gap hides.
- No E2E coverage was added for the new terminate response/headers path (`src/semantic-router/pkg/extproc/session_budget.go:104`, `src/semantic-router/pkg/headers/headers.go:336`), despite introducing new externally visible behavior when enabled.

**Style/Hygiene**

- Readability is generally strong; however, there is heavy explanatory comment volume in hot-path functions (`src/semantic-router/pkg/extproc/processor_req_body_prepare.go:208`, `src/semantic-router/pkg/extproc/session_budget.go:34`) that could be trimmed once behavior is stabilized.

**Size**

- Scope is large but coherent for a “foundation + terminate stage” PR (config + runtime + metrics + docs). I wouldn’t force a split, but fixing the validator/default-order bug is required before merge.

src/semantic-router/pkg/config/validator_session_budget.go
```diff
@@ -0,0 +1,63 @@
...
+			return fmt.Errorf("session_token_budget.thresholds.%s must not be negative (0 = use default), got %v", s.name, s.value)
+		}
+	}
+
+	// Explicit (non-zero) thresholds must be ascending in stage order. Equal
+	// adjacent thresholds are allowed and intentionally collapse the lower stage
+	// (the higher-priority stage wins in sessionbudget.Evaluate).
+	prevName, prev := "", 0.0
+	for _, s := range stages {
+		if s.value == 0 {
```
rootfs, line 51: Skipping zero-valued thresholds here lets invalid stage order pass validation when defaults are later applied in `ResolveThresholds` (e.g., `shape_tools: 2.0`, `compress: 0` resolves to `compress: 1.5`). Validate ordering on the effective thresholds after default resolution, not only on explicit fields.

src/semantic-router/pkg/config/validator_session_budget_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Enabled:      true,
+			BudgetTokens: 40000,
+			Thresholds:   SessionTokenBudgetThresholds{Compress: -0.5},
+		}
+		err := validateSessionTokenBudget(cfg)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("compress"))
+	})
+
+	It("rejects non-ascending thresholds (compress below shape_tools)", func() {
```
rootfs, line 56: Please add a case covering mixed explicit+default thresholds (for example `ShapeTools: 2.0` with `Compress: 0`) to catch the default-resolution ordering bug in the validator.

PR #2041, opened 2026-06-04
document write/update-recency eviction and add RAG cache unit tests

gagandhakrey:
## Issue

Addresses #2040  comment suggestions

### Documentation

* Updated `RAGResultCache` documentation to describe the cache as **write/update-recency based**.
* Clarified that only `setRAGCache()` (insert/update) promotes entries to MRU; reads do not affect recency.

### Tests

Added `req_filter_rag_cache_test.go` covering:

* Eviction order under read-heavy workloads.
* TTL expiry consistency between `cache.cache` and `cache.lru`.
* Existing-key updates preserving size and promoting to MRU.
* Concurrent reads without deadlock.
* Cache-disabled early-return behavior.

### DIFF
--- a/src/semantic-router/pkg/extproc/req_filter_rag_cache_test.go
+++ b/src/semantic-router/pkg/extproc/req_filter_rag_cache_test.go
@@ -1,0 +1,226 @@
+package extproc
+
+import (
+	"container/list"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+// newTestRAGCache returns a fresh RAGResultCache bypassing the singleton, so
+// each test starts with an empty, isolated cache.
+func newTestRAGCache(maxSize int) *RAGResultCache {
+	return &RAGResultCache{
+		cache:   make(map[string]*list.Element, maxSize),
+		lru:     list.New(),
+		maxSize: maxSize,
+	}
+}
+
+// resetRAGSingleton replaces the package-level singleton with a fresh cache of
+// the given maxSize and resets the Once so getRAGCacheInstance returns it.
+// Must only be called from tests; callers are responsible for restoring state.
+func resetRAGSingleton(maxSize int) *RAGResultCache {
+	fresh := newTestRAGCache(maxSize)
+	ragCache = fresh
+	ragCacheOnce = sync.Once{}
+	ragCacheOnce.Do(func() { ragCache = fresh })
+	return fresh
+}
+
+// setDirect inserts an entry into a standalone cache, bypassing the singleton.
+// It mirrors the production logic in setRAGCache exactly.
+func setDirect(r *OpenAIRouter, cache *RAGResultCache, key string, ctx string) {
+	entry := &RAGCacheEntry{Context: ctx, RetrievedAt: time.Now()}
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+	if el, exists := cache.cache[key]; exists {
+		el.Value.(*ragLRUItem).entry = entry
+		cache.lru.MoveToBack(el)
+		return
+	}
+	if len(cache.cache) >= cache.maxSize {
+		r.evictLRUEntry(cache)
+	}
+	el := cache.lru.PushBack(&ragLRUItem{key: key, entry: entry})
+	cache.cache[key] = el
+}
+
+// getDirect reads from a standalone cache, bypassing the singleton.
+// It mirrors the production logic in getRAGCache exactly.
+func getDirect(cache *RAGResultCache, key string, ttl int) (string, bool) {
+	cache.mu.RLock()
+	el, exists := cache.cache[key]
+	if !exists {
+		cache.mu.RUnlock()
+		return "", false
+	}
+	item := el.Value.(*ragLRUItem)
+	expired := ttl > 0 && time.Since(item.entry.RetrievedAt) > time.Duration(ttl)*time.Second
+	ctx := item.entry.Context
+	cache.mu.RUnlock()
+
+	if expired {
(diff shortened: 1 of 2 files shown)


gagandhakrey:
[Integration Test [Kubernetes] / integration-test (kubernetes) is failing which is independent of the changes done in the pr
it is failing because of 
429 Too Many Requests for url: https://huggingface.co/api/models/
llm-semantic-router/mmbert32k-jailbreak-detector-merged

this eventually leading to CrashLoopBackOff and the E2E test time out

src/semantic-router/pkg/extproc/req_filter_rag_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+	r := &OpenAIRouter{}
+	ttlSec := 1
+	cfg := &config.RAGPluginConfig{
+		Backend:         "ttltest",
+		CacheResults:    true,
+		CacheTTLSeconds: &ttlSec,
+	}
+
+	// Redirect the singleton to an isolated cache for this test.
+	cache := resetRAGSingleton(10)
```
rootfs, line 129: [risk:high] `resetRAGSingleton` overwrites the package-level `ragCache` and `ragCacheOnce` with no restore. Any test that runs after this one and exercises `getRAGCacheInstance()` will receive the stale test singleton silently. Add `orig, origOnce := ragCache, ragCacheOnce` before this call and `t.Cleanup(func() { ragCache = orig; ragCacheOnce = origOnce })` to prevent state leakage across tests.

src/semantic-router/pkg/extproc/req_filter_rag_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+		maxSize: maxSize,
+	}
+}
+
+// resetRAGSingleton replaces the package-level singleton with a fresh cache of
+// the given maxSize and resets the Once so getRAGCacheInstance returns it.
+// Must only be called from tests; callers are responsible for restoring state.
+func resetRAGSingleton(maxSize int) *RAGResultCache {
+	fresh := newTestRAGCache(maxSize)
+	ragCache = fresh
```
rootfs, line 27: [risk:medium] `ragCache = fresh` on this line is redundant — the `Do` call two lines later sets `ragCache = fresh` again. The first assignment is unreachable dead code because a freshly zeroed `sync.Once` will always execute its argument. Remove the pre-assignment to avoid misleading readers about initialization order.

src/semantic-router/pkg/extproc/req_filter_rag_cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+	ragCacheOnce = sync.Once{}
+	ragCacheOnce.Do(func() { ragCache = fresh })
+	return fresh
+}
+
+// setDirect inserts an entry into a standalone cache, bypassing the singleton.
+// It mirrors the production logic in setRAGCache exactly.
+func setDirect(r *OpenAIRouter, cache *RAGResultCache, key string, ctx string) {
```
rootfs, line 35: [risk:medium] `setDirect` and `getDirect` duplicate the production `setRAGCache`/`getRAGCache` logic verbatim. A bug introduced into the production functions will not be caught by tests that route exclusively through these helpers (the eviction-order and update tests). Add a comment explicitly noting the intentional duplication and the drift risk, or consider wiring the tests to call the real methods.

rootfs (commented):
## Code Review — PR 2041

**comment**
The doc fix is accurate and the test suite is solid, but one test leaves package-level singleton state mutated with no cleanup, which can poison later tests in the same run.

---

### Risk

No exported API is touched; the only change to `req_filter_rag_cache.go` is a comment block (`N29-N33`). Zero out-of-PR callers were found by the localizer (noting: local checkout is a TS repo, not the target Go repo, so cross-file verification was unavailable — but the diff itself shows no signature changes).

No breaking risk from the production side.

---

### Tests

**Good:**
- `TestRAGCache_TTLExpiry_MapAndListConsistency` (`N145`) calls the real production method `r.getRAGCache`, making it the highest-value test here.
- Eviction-order and update-path tests reason correctly about the write/update-recency invariant.

**Issue — singleton state leak (`N128-N129`, `N25-N31`):**  
`TestRAGCache_TTLExpiry_MapAndListConsistency` calls `resetRAGSingleton(10)` which overwrites `ragCache` and `ragCacheOnce` at the package level, but never restores them. Any subsequent test in the package that exercises the real `getRAGCacheInstance()` path will silently receive the stale test singleton. Fix with `t.Cleanup`:

```go
orig, origOnce := ragCache, ragCacheOnce
t.Cleanup(func() { ragCache = orig; ragCacheOnce = origOnce })
cache := resetRAGSingleton(10)
```

**Issue — `setDirect`/`getDirect` duplicate production logic (`N35-N78`):**  
These helpers re-implement `setRAGCache`/`getRAGCache` verbatim. A bug introduced into the production functions won't be caught by the tests that use only these helpers (the eviction-order and update tests). Consider calling the real methods on a router wired to the isolated cache, or at minimum add a comment acknowledging the intentional duplication and the risk of drift.

**Diff is truncated** — `TestRAGCache_ConcurrentReadsDontDeadlock` and `TestGetRAGCache_CacheDisabled` are cut off. Assuming they compile and pass, the coverage intent is appropriate.

---

### Style/Hygiene

- `resetRAGSingleton`: the first line `ragCache = fresh` (`N27`) is redundant — the `Do` call on `N29` sets it again. Minor.
- Comment in `req_filter_rag_cache.go` (`N29-N33`) is clear and accurate. No churn concerns.

---

### Size

231 lines, 2 files. Scope is appropriate — documentation fix + matching unit tests in one PR is clean.

---

`Tools used: NONE`

PR #2058, opened 2026-06-05
[Dashboard] Gate the open bootstrap endpoint behind an opt-in and close the first-admin race

shraderdm:
## Purpose

- The dashboard's public, unauthenticated bootstrap path creates the first admin when no users exist: `POST /api/auth/bootstrap/register` creates the admin, and `GET /api/auth/bootstrap/can-register` reports whether that is possible. On an internet-reachable deployment this has two problems:
  - **Race (TOCTOU):** the register handler checked `CanBootstrap()` (users == 0) and then created the admin with nothing atomic in between, so two concurrent requests could both pass the check and each create an admin.
  - **Claimable-by-default + info leak:** a fresh deployment with no admin yet was claimable by the first unauthenticated caller to find it, and `can-register` advertised that claimable state to anyone.
- This change makes the open bootstrap **opt-in and secure-by-default**, and closes the race:
  - New `DASHBOARD_ALLOW_OPEN_BOOTSTRAP` (default `false`). When off, `register` returns 403 and `can-register` reports `false` without revealing the user count. Production provisions the admin via the existing `DASHBOARD_ADMIN_*` env path, which creates the admin at startup so the bootstrap window never opens. Demos set the flag to `true` to sign up the first admin through the UI.
  - When on, the check-and-create is serialized in `BootstrapRegister` so only one admin can be created even under concurrent requests. The dashboard runs single-replica (enforced by the chart's `replicaCount` guard), so a process-level mutex is sufficient; a multi-writer deployment would need a transactional guard in the store.
  - Helm: new `dashboard.allowOpenBootstrap` value (default `false`) wires `DASHBOARD_ALLOW_OPEN_BOOTSTRAP`, mirroring the existing `dashboard.readonly` pattern.
- Module: `Dashboard` (Go backend + Helm chart).
- Note: this changes the out-of-box first-run UX (the open web-form bootstrap is now opt-in rather than on). The env-admin path is unchanged and remains the production-blessed way to bootstrap, so no deployment is stranded. Flagging in case the team would prefer a different default.
- Two small behavior details worth calling out: (1) when the flag is off, `register` returns 403 (disabled) even if an admin already exists, rather than 409, because the gate is checked before the bootstrap-state check; this keeps "disabled" unambiguous. (2) `BootstrapRegister` now applies `defaultAdminName` (an empty name becomes "Admin"), aligning the web-form path with the existing env-admin path (`EnsureBootstrapAdmin` already did this); the previous handler stored an empty name as-is.

## Test Plan

- `go test ./dashboard/backend/auth/ -run TestBootstrap -race`, plus `go vet` and the full `go test ./auth/`.
- `helm template` with and without `dashboard.allowOpenBootstrap`, a non-bool value, and `helm lint`.

## Test Result

- Go (auth package, including `-race`): all bootstrap-gate tests pass:
  - disabled by default: `can-register` reports false, `register` returns 403, no user created
  - enabled, no users: `can-register` true, `register` 200, exactly one admin created
  - enabled, admin already exists: `register` returns 409
  - 16 concurrent enabled register requests: exactly one admin created and exactly one 200 response (the race is closed)
  - `go vet ./auth/... ./config/... ./router/...` clean; `go test ./auth/` ok
- Helm:
  - default render (`dashboard.enabled=true`): no `DASHBOARD_ALLOW_OPEN_BOOTSTRAP` env injected (secure default)
  - `--set dashboard.allowOpenBootstrap=true`: env present with value `"true"`
  - non-bool value: rejected by schema, `at '/dashboard/allowOpenBootstrap': got string, want boolean`
  - `helm lint`: 1 chart(s) linted, 0 failed; `values.schema.json` parses as valid JSON

### DIFF
--- a/dashboard/backend/auth/bootstrap_gate_test.go
+++ b/dashboard/backend/auth/bootstrap_gate_test.go
@@ -1,0 +1,122 @@
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"testing"
+)
+
+func newBootstrapService(t *testing.T, allowOpen bool) *Service {
+	t.Helper()
+	svc := newTestAuthService(t)
+	svc.SetAllowOpenBootstrap(allowOpen)
+	return svc
+}
+
+func canRegister(t *testing.T, svc *Service) bool {
+	t.Helper()
+	rec := httptest.NewRecorder()
+	bootstrapCanRegisterHandler(svc)(rec, httptest.NewRequest(http.MethodGet, "/api/auth/bootstrap/can-register", nil))
+	if rec.Code != http.StatusOK {
+		t.Fatalf("can-register status = %d, want 200", rec.Code)
+	}
+	var resp BootstrapStatusResponse
+	if err := json.Unmarshal(rec.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("decode can-register: %v", err)
+	}
+	return resp.CanRegister
+}
+
+func postRegister(svc *Service, email string) *httptest.ResponseRecorder {
+	rec := httptest.NewRecorder()
+	body := fmt.Sprintf(`{"email":%q,"password":"secret-password","name":"Admin"}`, email)
+	bootstrapRegisterHandler(svc)(rec, httptest.NewRequest(http.MethodPost, "/api/auth/bootstrap/register", strings.NewReader(body)))
+	return rec
+}
+
+// With open bootstrap disabled (the default), can-register must report false even
+// when no users exist - both to disable the path and to avoid leaking to an
+// unauthenticated caller that the instance is freshly deployed and claimable.
+func TestBootstrapCanRegister_DisabledByDefaultReportsFalse(t *testing.T) {
+	svc := newBootstrapService(t, false)
+	if canRegister(t, svc) {
+		t.Fatal("canRegister = true with open bootstrap disabled; want false")
+	}
+}
+
+func TestBootstrapCanRegister_EnabledTracksUserCount(t *testing.T) {
+	svc := newBootstrapService(t, true)
+	if !canRegister(t, svc) {
+		t.Fatal("canRegister = false with open bootstrap enabled and no users; want true")
+	}
+	newTestUser(t, svc, "admin@example.com", RoleAdmin, "active")
+	if canRegister(t, svc) {
+		t.Fatal("canRegister = true after an admin exists; want false")
+	}
+}
+
+// The open register endpoint must be closed by default and must not create a user.
+func TestBootstrapRegister_DisabledByDefaultForbidden(t *testing.T) {
+	svc := newBootstrapService(t, false)
+	rec := postRegister(svc, "admin@example.com")
+	if rec.Code != http.StatusForbidden {
+		t.Fatalf("status = %d, want 403 when open bootstrap disabled", rec.Code)
+	}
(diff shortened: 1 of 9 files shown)


rootfs (commented):
## Code Review

### 1. Verdict
`approve`

Security-by-default improvement with comprehensive test coverage, proper mutex-based race fix, and well-aligned Helm configuration.

---

### 2. Risk

**Breaking behavior change (intentional, documented)**
- **dashboard/backend/auth/public_handlers.go:20**: `can-register` now returns `false` by default instead of revealing the actual user count. Deployments relying on the open bootstrap flow must now explicitly set `dashboard.allowOpenBootstrap=true` or provision via `DASHBOARD_ADMIN_*`. The PR description flags this UX change appropriately.
- **deploy/helm/semantic-router/values.yaml:536**: Default is `allowOpenBootstrap: false`. Existing Helm users who relied on the open bootstrap will need to opt in on upgrade.

**Doc drift potential**
- The localizer found 3 versioned installation docs mentioning bootstrap/setup mode:
  - `website/versioned_docs/version-v0.3/installation/installation.md:82`
  - `website/versioned_docs/version-v0.2/installation/installation.md:82`
  - `website/versioned_docs/version-v0.1/installation/installation.md:45`
- These describe `vllm-sr serve` bootstrap behavior when `config.yaml` doesn't exist. If those docs describe the *dashboard* bootstrap (the admin registration flow), they may now be misleading since the open path is off by default. Suggest verifying whether those sections cover CLI workspace bootstrap (unaffected) or dashboard admin bootstrap (now gated).

**No out-of-PR code impact**
- Localizer confirmed 0 external references to `SetAllowOpenBootstrap`, `OpenBootstrapEnabled`, `BootstrapRegister`, so the internal API changes are safe.

---

### 3. Tests

**Strong coverage**
- **dashboard/backend/auth/bootstrap_gate_test.go:45-122**: New test file covers all critical paths:
  - Disabled by default returns false (L45)
  - Enabled tracks user count correctly (L52)
  - Disabled rejects registration with 403 (L64)
  - Enabled creates first admin (L75)
  - Second attempt returns 409 (L86)
  - **Race fix validated**: 16 concurrent requests create exactly one admin (L96)
- Tests use `-race` flag per the Test Plan, which is the right approach for verifying the mutex fix.

**No related tests need updates**
- Localizer found 0 out-of-PR test files referencing the changed symbols.
- Existing handler tests (if any) would still pass because the service defaults to `allowOpenBootstrap=false` unless explicitly set.

---

### 4. Style/Hygiene

**Clean implementation**
- **dashboard/backend/auth/service.go:62-77**: `BootstrapRegister` has clear docstring explaining the mutex serialization and stating the single-replica constraint.
- **dashboard/backend/auth/public_handlers.go:17-23**: Good comment explaining why disabled mode always returns `false` (avoids info leak).
- **dashboard/backend/config/config.go:38-41**: Config field has clear docstring pointing users to the preferred `DASHBOARD_ADMIN_*` path.

**Minor: error message clarity**
- **dashboard/backend/auth/public_handlers.go:43**: The 403 error message is helpful and actionable.
- **dashboard/backend/auth/service.go:35**: `ErrBootstrapClosed` naming is clear, though the message "bootstrap is disabled" (returned at L67) maps to "admin already exists" in this context. The handler at public_handlers.go:67 wraps this as "bootstrap is disabled" for 409 Conflict, which is semantically correct.

---

### 5. Size

Appropriately scoped: touches only dashboard auth, config, and Helm chart. The change is cohesive—one security concern addressed with aligned code, config, and tests.

---

**Tools used**: NONE

PR #2060, opened 2026-06-05
[Cleanup] drop supervisord monolith paths, default split runtime containers

### LINKED ISSUE #1508: feature: split vllm-sr into a router-only runtime and external service images

FAUST-BENCHOU:
PLEASE FILL IN THE PR DESCRIPTION BELOW AND CONFIRM THE CHECKLIST ITEMS.

Closes https://github.com/vllm-project/semantic-router/issues/1508

## Purpose

- What does this PR change?
- Why is this change needed?

## Test Plan

## Test Result

---
<details>

</details>


dashboard/backend/handlers/runtime_config_apply.go
```diff
@@ -96,10 +96,10 @@ func propagateConfigToRuntime(configPath string, configDir string) error {
 	}
 
 	if isRunningInContainer() && isManagedContainerConfigPath(configPath) {
-		if !managedServiceCanBeControlledLocally("envoy") && getDockerContainerStatus(managedContainerNameForService("envoy")) == "running" {
+		if getDockerContainerStatus(managedContainerNameForService("envoy")) == "running" {
 			return regenerateAndReloadManagedSplitEnvoyLocally(effectiveConfigPath)
 		}
-		return regenerateAndReloadEnvoyLocally(effectiveConfigPath)
+		return nil
```
rootfs, line 102: [risk:high] Returning `nil` here skips any reload when the managed envoy container is not running; if config propagation is expected to update the active split runtime later, this turns a recoverable state into a silent no-op.

dashboard/backend/handlers/runtime_managed_services.go
```diff
@@ -9,34 +9,33 @@ const (
...
-		return envOrDefaultTrimmed(routerContainerNameEnv, vllmSrContainerName)
+		return envOrDefaultTrimmed(routerContainerNameEnv, defaultRouterContainerName)
 	case "envoy":
-		return envOrDefaultTrimmed(envoyContainerNameEnv, vllmSrContainerName)
+		return envOrDefaultTrimmed(envoyContainerNameEnv, defaultEnvoyContainerName)
 	case "dashboard":
 		return managedDashboardContainerName()
 	default:
-		return vllmSrContainerName
+		return defaultRouterContainerName
```
rootfs, line 27: [risk:medium] The `default` branch now falls back to the router container name for unknown services. If any caller passes a new service key, this can mask the bug and control the wrong container instead of failing fast.

dashboard/backend/handlers/status_collectors.go
```diff
@@ -1,51 +1,18 @@
...
-	case "exited":
-		return exitedContainerStatus()
-	case "not found":
-		if status, ok := collectDirectStatus(runtimePath, routerAPIURL); ok {
-			return status
-		}
-		return baseSystemStatus()
-	default:
-		return unknownContainerStatus(containerStatus)
+	if status, ok := collectDirectStatus(runtimePath, routerAPIURL); ok {
```
rootfs, line 12: [risk:high] Removing the `collectRunningDockerStatus`/`exitedContainerStatus` branches changes host-status behavior from container-state-aware to direct-status-only. I can't verify a compensating caller from localizer evidence (0 out-of-PR callsites), so please confirm this still reports `running`/`exited` correctly for legacy single-container deployments.

dashboard/backend/handlers/logs.go
```diff
@@ -128,100 +128,11 @@ func populateDirectRuntimeMessage(response *LogsResponse, routerAPIURL string) {
 }
 
 func runtimeContainerStatusForLogs() string {
-	if managedRuntimeUsesSplitContainers() {
-		if status := managedRuntimeContainerStatus(); status != "not found" {
-			return status
-		}
-	}
-	return getDockerContainerStatus(vllmSrContainerName)
+	return managedRuntimeContainerStatus()
```
rootfs, line 131: [risk:medium] `runtimeContainerStatusForLogs()` now always returns `managedRuntimeContainerStatus()`. If the managed status is `not found`, the previous fallback to Docker status is gone, which may regress log pages for environments that still only expose the legacy container.

dashboard/backend/handlers/openclaw_helpers.go
```diff
@@ -75,7 +75,7 @@ func openClawModelGatewayContainerName() string {
 	if candidate := strings.TrimSpace(os.Getenv("OPENCLAW_DASHBOARD_CONTAINER_NAME")); candidate != "" {
 		return candidate
 	}
-	return vllmSrContainerName
+	return defaultRouterContainerName
```
rootfs, line 78: [risk:medium] Defaulting OpenClaw dashboard lookup to `defaultRouterContainerName` changes the fallback target from the legacy monolith container to the split router container. Please verify OpenClaw still resolves the intended dashboard host when no env override is set.

dashboard/backend/handlers/status.go
```diff
@@ -36,9 +36,6 @@ type SystemStatus struct {
 	Version        string               `json:"version,omitempty"`
 }
 
-// vllmSrContainerName is the container name used by the Python vllm-sr CLI
-const vllmSrContainerName = "vllm-sr-container"
-
 // StatusHandler returns the status of vLLM-SR services
```
rootfs, line 39: [risk:low] `vllmSrContainerName` was removed entirely; if any external package or reflection-based code still references the old exported behavior, this is now an API break. Please confirm the symbol is unused outside this repo.

rootfs (commented):
request_changes

This PR removes the legacy single-container/supervisord runtime paths, but the diff shows several behavior changes and API/UX regressions without corresponding evidence of updated callsites or compatibility handling, and the localizer found zero out-of-PR references/tests to validate the removed symbols.

## Risk

- `dashboard/backend/handlers/logs.go:131-135` now makes `runtimeContainerStatusForLogs()` always return `managedRuntimeContainerStatus()` and `fetchContainerLogs()` always use `fetchLogsFromManagedDocker(...)`; the previous fallback to legacy Docker/supervisord logs is gone, so any host/container still reporting the old monolith will lose log access.
- `dashboard/backend/handlers/runtime_config_apply.go:98-102` changes config propagation to return `nil` when the Envoy container is not running, instead of calling `regenerateAndReloadEnvoyLocally(...)`; this is a silent skip that can leave config changes unapplied with no error.
- `dashboard/backend/handlers/runtime_config_apply.go:312-316` removes the local supervised restart branch and now rejects all non-lifecycle services with `unsupported managed service restart`; that is a functional behavior removal for any code paths still invoking restart on the old managed service model.
- `dashboard/backend/handlers/runtime_managed_services.go:18-33` changes default container-name resolution to split defaults (`defaultRouterContainerName`, `defaultEnvoyContainerName`, `defaultDashboardContainerName`) and also changes the unknown-service default in `managedContainerNameForService` to `defaultRouterContainerName`; if any consumer expects the old shared name, it will now resolve differently.
- `dashboard/backend/handlers/status_collectors.go:7-16` removes the Docker-status branching and `collectRunningDockerStatus`/`exitedContainerStatus` handling, so host status now falls through to `collectDirectStatus(...)` or `baseSystemStatus()`; that is a potentially breaking change for deployments that still present as a single container.
- `dashboard/backend/handlers/status_collectors.go:57-67` changes the unknown-container status payload to a generic `Runtime` service with `Component: "container"` instead of `vllm-sr-container`; any UI or client logic keyed on the old name will break or show a different service identity.
- `src/vllm-sr/cli/consts.py:16-19` removes `VLLM_SR_DOCKER_NAME`; if Python CLI code outside this PR still imports that constant, it will fail at runtime. I cannot verify out-of-PR callsites because the localizer returned zero matches.
- `src/vllm-sr/start-envoy.sh:1-31` and `src/vllm-sr/supervisord.conf:1-80` are deleted, so any image/build, docs, or scripts still referencing those entrypoints will break unless updated elsewhere; I have no indexed evidence of those callsites from the localizer, so confidence is lower.
- `dashboard/backend/handlers/status.go:39-41` removes the comment and constant definition for `vllmSrContainerName`; coupled with deletions in `status_collectors.go` and `openclaw_helpers.go:75-78`, this looks like a repo-wide rename, but the PR does not show migration guidance for external consumers.

## Tests

- `dashboard/backend/handlers/logs_test.go:9-18` was renamed and no longer asserts the legacy-container residue path; there is no visible replacement test covering a host still advertising the old monolith container, which is the main regression risk introduced by `logs.go:131-135`.
- `dashboard/backend/handlers/runtime_managed_services_test.go:5-18` now expects split defaults, but I do not see a test covering the fallback/unknown-service case changed at `runtime_managed_services.go:18-27`.
- `dashboard/backend/handlers/runtime_config_apply.go:98-102` removed `regenerateAndReloadEnvoyLocally(...)` entirely; I do not see a test proving that config propagation still works when Envoy is not running or when the container status is transient.
- `dashboard/backend/handlers/status_collectors_test.go:50-79` no longer checks that legacy residue is excluded, but there is no companion test proving the new generic `Runtime` unknown status from `status_collectors.go:57-67` is rendered correctly.
- The localizer evidence is empty for all changed symbols, including `TestManagedContainerNameForServiceDefaultsToLegacyRuntimeContainer`, `TestManagedRuntimeUsesSplitContainersDefaultsToFalse`, `TestCollectHostStatusPrefersSplitManagedRuntimeOverLegacyContainerResidue`, and the various log/restart helpers; because there were zero out-of-PR matches, I cannot validate any repo-wide test coverage beyond the files shown.

## Style/Hygiene

- `dashboard/backend/handlers/runtime_managed_services.go:18-33` uses `defaultRouterContainerName` as the default for unknown services; that is surprising naming and makes the fallback semantically ambiguous.
- `dashboard/backend/handlers/status_collectors.go:57-67` introduces a generic `Runtime` service name, but the component is `"container"`; consider keeping the service label aligned with the user-facing runtime concept to avoid mixed terminology.
- `dashboard/backend/handlers/runtime_config_apply.go:99-102` returns `nil` on a skipped reload without logging; this is hard to diagnose and reads like dead control flow.
- `src/vllm-sr/start-router.sh:11` changes the comment wording only; fine, but the deleted `start-envoy.sh` and `supervisord.conf` mean the remaining startup scripts should probably be re-audited for consistency.

## Size

- The scope is large but coherent around one migration: removing supervisord/monolith paths and defaulting to split containers.
- That said, the PR mixes runtime behavior changes, status/log handling, config propagation, constant removal, and packaging script deletions; I would strongly consider splitting into at least:
  1. runtime/container name and status refactor,
  2. dashboard log/config propagation changes,
  3. CLI/image entrypoint cleanup and file deletions.

dashboard/backend/handlers/runtime_config_apply.go
```diff
@@ -96,10 +96,10 @@ func propagateConfigToRuntime(configPath string, configDir string) error {
 	}
 
 	if isRunningInContainer() && isManagedContainerConfigPath(configPath) {
-		if !managedServiceCanBeControlledLocally("envoy") && getDockerContainerStatus(managedContainerNameForService("envoy")) == "running" {
+		if getDockerContainerStatus(managedContainerNameForService("envoy")) == "running" {
```
rootfs, line 99: [risk:high] This now reloads split Envoy unconditionally whenever the managed envoy container is running, but the previous guard explicitly avoided local control when the container isn’t locally managed. Please confirm this can’t trigger unintended reloads for remote/external deployments.

dashboard/backend/handlers/runtime_config_apply.go
rootfs, line 102: [risk:high] Returning nil here drops the previous `regenerateAndReloadEnvoyLocally(...)` fallback entirely. If `isRunningInContainer() && isManagedContainerConfigPath(configPath)` is true but the split-managed envoy container is not running, config propagation will now silently no-op.

dashboard/backend/handlers/runtime_managed_services.go
rootfs, line 27: [risk:high] The default branch now returns `defaultRouterContainerName` for every unknown service. That changes the legacy failure mode and can misroute future callers of `managedContainerNameForService` to the router container instead of surfacing an unsupported-service error.

dashboard/backend/handlers/status_collectors.go
rootfs, line 12: [risk:high] `collectHostStatus` no longer distinguishes `running`/`exited`/`not found` from `getDockerContainerStatus(...)`; any non-split host path now falls through to `collectDirectStatus(...)` or `baseSystemStatus()`, which looks like a behavior regression for hosts where Docker is the primary signal.

dashboard/backend/handlers/status_collectors.go
```diff
@@ -87,51 +54,15 @@ func collectManagedDockerStatus(runtimePath, routerAPIURL string) SystemStatus {
...
-
 func unknownContainerStatus(containerStatus string) SystemStatus {
 	status := baseSystemStatus()
 	status.DeploymentType = "docker"
 	status.Overall = containerStatus
 	status.Services = append(status.Services, ServiceStatus{
-		Name:    "vllm-sr-container",
-		Status:  containerStatus,
-		Healthy: false,
+		Name:      "Runtime",
```
rootfs, line 62: [risk:medium] The fallback service name is now hard-coded to `Runtime`, but this object is still returned from the Docker-status path. Please verify downstream status consumers don’t rely on the previous container name or on `Component: "container"` being present for this case.

dashboard/backend/handlers/logs.go
rootfs, line 131: [risk:medium] `runtimeContainerStatusForLogs()` now always returns `managedRuntimeContainerStatus()`. If the managed runtime is absent or the split containers are not present, this no longer falls back to the legacy container status that the old code provided.

dashboard/backend/handlers/logs.go
```diff
@@ -128,100 +128,11 @@ func populateDirectRuntimeMessage(response *LogsResponse, routerAPIURL string) {
...
-	return result, nil
-}
-
-// fetchLogsFromDocker gets logs from Docker container (fallback for external access)
-func fetchLogsFromDocker(component string, lines int) ([]string, error) {
-	if managedRuntimeUsesSplitContainers() {
-		return fetchLogsFromManagedDocker(component, lines)
-	}
-	return fetchLogsFromLegacyDocker(component, lines)
+	return fetchLogsFromManagedDocker(component, lines)
```
rootfs, line 135: [risk:medium] `fetchContainerLogs()` now always calls `fetchLogsFromManagedDocker(...)`; the removed supervisor/docker fallback means any non-split or partially-migrated runtime will lose log access instead of degrading to the previous path.

dashboard/backend/handlers/status_log_patterns.go
```diff
@@ -2,14 +2,6 @@ package handlers
...
 
-func checkServiceInLogContent(service, logContent string) (bool, string) {
-	if serviceLogLooksHealthy(service, logContent) {
-		return true, "Running"
-	}
-
-	return false, "Status unknown (check logs)"
-}
-
 func serviceLogLooksHealthy(service, logContent string) bool {
```
rootfs, line 5: [risk:medium] `checkServiceInLogContent` was removed and the remaining log-status inference helpers may now be unused or unreachable. Please re-check any callers/tests that still expect this higher-level health classification behavior.

src/vllm-sr/start-router.sh
```diff
@@ -8,7 +8,7 @@ CONFIG_FILE="${1:-/app/config.yaml}"
 echo "Starting router from canonical config..."
 echo "  Config file: $CONFIG_FILE"
 
-# Preserve setup-mode behavior from the legacy supervisord entrypoint.
+# Preserve setup-mode behavior from the historical single-container entrypoint.
```
rootfs, line 11: [risk:low] The comment now says “historical single-container entrypoint,” but this script still serves the router runtime path. Consider updating the wording if the intent is to document the split-container migration more precisely.

rootfs (commented):
request_changes

This cleanup is directionally coherent, but it removes legacy single-container/supervisord behavior while also changing default container naming and status collection paths in ways that are not fully guarded by repo-wide callsite evidence, so I’m not confident it’s safe without explicit follow-up verification.

## Risk
- `dashboard/backend/handlers/logs.go:130-135` unconditionally routes `runtimeContainerStatusForLogs()` and `fetchContainerLogs()` through the split-managed path. That removes the previous fallback to `getDockerContainerStatus(vllmSrContainerName)` and `fetchLogsFromDocker()/fetchLogsFromLegacyDocker()`, so any non-split or partially migrated deployment will now lose the legacy logs path.
- `dashboard/backend/handlers/runtime_config_apply.go:98-102` changes managed-container config propagation so the in-container path now only reloads split Envoy when the Envoy container is running, otherwise it returns `nil`. The deleted `regenerateAndReloadEnvoyLocally()` path in `dashboard/backend/handlers/runtime_config_apply.go:116-139` means the old supervisord-managed reload path no longer exists.
- `dashboard/backend/handlers/runtime_config_apply.go:312-316` now rejects every managed restart except `restartOrStartManagedSplitContainerService`, while the deleted `restartOrStartSupervisorService()` / `restartOrStartManagedContainerService()` implementations at `dashboard/backend/handlers/runtime_config_apply.go:361-383` are no longer available. That is a hard behavior removal for any code path still expecting local supervisor control.
- `dashboard/backend/handlers/status_collectors.go:7-15` removes the old host-status fallback cascade that used `getDockerContainerStatus(vllmSrContainerName)` and could return exited/unknown-container states. The new code goes directly to `collectDirectStatus()` or `baseSystemStatus()`, so status semantics change for host deployments.
- `dashboard/backend/handlers/status_collectors.go:57-66` changes the status payload for unknown container states from `Name: "vllm-sr-container"` to `Name: "Runtime"` plus `Component: "container"`. Any consumer that keys on the old service name will break.
- `dashboard/backend/handlers/openclaw_helpers.go:75-78` changes the default OpenClaw gateway container name from `vllmSrContainerName` to `defaultRouterContainerName`. That is an externally visible name change and could affect container lookups.
- `dashboard/backend/handlers/runtime_managed_services.go:18-33` changes default container names for router/envoy/dashboard to split values. This is a breaking default for any environment that relied on the legacy monolith name, even when env vars are unset.
- `src/vllm-sr/cli/consts.py:16-19` removes `VLLM_SR_DOCKER_NAME = "vllm-sr-container"`. If any Python CLI, packaging, or docs still import/use that constant, they will fail at runtime or drift semantically.
- `src/vllm-sr/start-envoy.sh` is deleted entirely, and `src/vllm-sr/supervisord.conf` is deleted entirely. If any build/deploy path still references either artifact, startup will fail.

## Tests
- The changed tests in `dashboard/backend/handlers/logs_test.go:9-18`, `runtime_managed_services_test.go:5-80`, and `status_collectors_test.go:50-79` cover the new split-runtime expectations, but they do not appear to exercise any compatibility path because the legacy branches were removed.
- `dashboard/backend/handlers/runtime_config_apply.go:98-102` has no visible direct regression test in the diff for the “envoy running inside container” case versus the now-removed `regenerateAndReloadEnvoyLocally()` branch.
- `dashboard/backend/handlers/status_collectors.go:7-15` and `:57-66` changed status assembly behavior; I’d want explicit tests for host-status fallbacks and unknown-container naming, since those semantics are now different.
- The localizer evidence is empty for out-of-PR tests/callsites, so I cannot verify whether repo-wide tests already cover deleted supervisor flows. That lowers confidence on the removals.
- If this PR intentionally removes supervisor support, add negative tests asserting the unsupported paths fail as expected rather than silently returning `nil` or a generic unsupported error.

## Out-of-PR follow-ups
### verified_callsites
(none from localizer evidence)  
The localizer surfaced zero out-of-PR call-sites for the changed symbols, so I cannot name dependent callers outside this PR.

### verified_tests
(none from localizer evidence)  
The localizer surfaced zero out-of-PR tests for the changed symbols, so there is nothing concrete to update from the repo graph.

### verified_configs
(none from localizer evidence)  
The localizer surfaced zero configs/schemas mentioning these symbols, so no verified external config file needs an update from the evidence provided.

### verified_docs
(none from localizer evidence)  
The localizer surfaced zero docs/READMEs mentioning these symbols, so I can’t point to a verified external documentation file.

## Style/Hygiene
- `dashboard/backend/handlers/runtime_config_apply.go:312-316` is cleaner after the simplification, but the new unconditional rejection means the function name is now broader than its implementation; consider renaming or documenting that only split-runtime restarts are supported.
- `dashboard/backend/handlers/status_collectors.go:62-65` changes the service name to `"Runtime"` only for unknown container states, while other branches still build per-service statuses; the mixed naming scheme may confuse consumers.
- The deleted `src/vllm-sr/start-envoy.sh` and `src/vllm-sr/supervisord.conf` suggest dead artifacts were removed appropriately, but only if no packaging/install references remain.

## Size
- This is a fairly large behavioral cleanup, not just a path deletion: it changes runtime naming, status reporting, config propagation, restart semantics, and artifact layout across dashboard/backend and CLI.
- I would normally prefer this split into at least two PRs: one for container/runtime naming and status logic, and one for deleting supervisor/monolith artifacts and updating deployment/build references.

FAUST-BENCHOU:
Works well now.Thanks for reparing it! :)

Xunzhuo (approved):
LGTM

PR #2065, opened 2026-06-06
[Docs] Add multi-arch :latest digest-pinning troubleshooting note

shraderdm:
## Purpose

- Adds a troubleshooting entry for a real image-pull failure class: pinning a digest you resolved on Apple Silicon (arm64) for a multi-arch `:latest` manifest list breaks amd64 nodes, because that digest is the arm64 child manifest, not the multi-arch index. amd64 nodes then fail to pull.
- Documents the fix: reference the image by tag, or pin the multi-arch INDEX digest (in the `name@sha256:...` form) via `docker buildx imagetools inspect <image> --format '{{.Manifest.Digest}}'`.
- Notes the chart limitation explicitly: the bundled Helm chart builds image references as `repository:tag`, so digest pinning is done in your own deployment manifest, not through `image.tag` / `dashboard.image.tag`.
- Module: `Docs`.

## Test Plan

- `markdownlint -c tools/linter/markdown/markdownlint.yaml` on the changed page.
- The Docusaurus build runs in the PR's Netlify deploy preview.

## Test Result

- `markdownlint`: clean.
- Multi-arch behavior verified against the live `ghcr.io/vllm-project/semantic-router/extproc:latest` image: the index, amd64-child, and arm64-child digests are all distinct, and `--format '{{.Manifest.Digest}}'` returns the index (manifest-list) digest, confirming the documented command:
  - index: `sha256:46b57f20...` (`application/vnd.docker.distribution.manifest.list.v2+json`)
  - amd64 child: `sha256:ec0c1efb...`
  - arm64 child: `sha256:9b75395c...`
- No digests are hardcoded in the doc (they are ephemeral); the guidance is digest-agnostic.

### DIFF
--- a/website/docs/troubleshooting/common-errors.md
+++ b/website/docs/troubleshooting/common-errors.md
@@ -335,6 +335,58 @@ global:
 
 ---
 
+## Image Pull Errors
+
+### No matching manifest for linux/amd64 (after pinning a digest built on Apple Silicon)
+
+**Symptoms:** Pods on `amd64` nodes stay in `ImagePullBackOff` with an event like
+`no matching manifest for linux/amd64 in the manifest list entries`, while the
+same image reference runs fine on `arm64` nodes or on an Apple Silicon laptop.
+The exact wording varies by container runtime (Docker vs containerd), but it
+always reports that no image matches the node's `linux/amd64` platform.
+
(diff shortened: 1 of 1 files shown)


FAUST-BENCHOU (approved):
LGTM.Thanks

PR #2068, opened 2026-06-06
[Dashboard] Helm chart: add dashboard.extraEnv and dashboard.envFrom escape hatches

shraderdm:
## Purpose

- Adds `dashboard.extraEnv` (a core/v1 `EnvVar` list) and `dashboard.envFrom` (a core/v1 `EnvFromSource` list) so operators can inject optional integration environment (for example `OPENCLAW_*`, `TARGET_ENVOY_URL`, `PROXY_OVERRIDE_ORIGIN`, or a configMap/secret ref) into the dashboard container without forking the chart.
- Both default to `[]` (no-op). Rendered after the chart-managed `TARGET_*` vars, so they compose with the existing `readonly` / `jwtSecret` / `allowOpenBootstrap` env wiring rather than replacing it.
- Mirrors the `dashboard.readonly` pattern already in the chart. Module: `Dashboard` (Helm chart): values + schema + README.
- Naming note: the router container already has an `envFromSecrets` escape hatch (a list of secret names). The dashboard intentionally uses the more general `envFrom` (full core/v1 `EnvFromSource`, so `configMapRef` works too) rather than mirroring `envFromSecrets`; happy to rename for parity if you'd prefer consistency over flexibility. `extraEnv` matches the router precedent name as-is.
- Footgun guard: redefining a chart-managed var (e.g. `TARGET_*`, `DASHBOARD_JWT_SECRET`) via `extraEnv` yields a duplicate env key (Kubernetes last-wins). Called out in both the `values.yaml` comment and the README.

## Test Plan

- `helm template` with `dashboard.enabled=true` and neither value set, with `extraEnv` populated, and with `envFrom` populated; plus a combined render to confirm coexistence with `jwtSecret`/`allowOpenBootstrap`; `helm lint`; schema validity. (The dashboard is disabled by default, so renders set `dashboard.enabled=true` to exercise the container.)

## Test Result

- Default render (`dashboard.enabled=true`, neither value set): no extra env and no `envFrom:` block injected.
- `extraEnv` populated: renders `- name: FOO / value: bar` in the dashboard container env.
- `envFrom` populated: renders the `envFrom: - secretRef: name: ...` block.
- Combined render (`allowOpenBootstrap` + `jwtSecret.existingSecret` + `extraEnv`): all three wire simultaneously.
- `helm lint`: `1 chart(s) linted, 0 chart(s) failed`; `values.schema.json` parses as valid JSON.


drivebyer (approved):

shraderdm:
Synced this to `main` to clear the out-of-date state; CI is re-running and should come back green (it was passing before the sync). Approved by @drivebyer, and it's a small additive, opt-in change (the two Helm escape hatches default to empty). Could a committer merge it once CI is green? @Xunzhuo, flagging for your queue when convenient. Thanks!

drivebyer:
Thanks for your contribution :)

PR #2070, opened 2026-06-06
[Router] Fix episodic memory extraction count metric

cryo-zd:
PLEASE FILL IN THE PR DESCRIPTION BELOW AND CONFIRM THE CHECKLIST ITEMS.

Closes #xxxx

## Purpose

- What does this PR change?
  - Fix episodic memory extraction metrics to report the actual number of stored memory chunks by replacing the previous fixed `count=1` with explicit counting for per-turn and session-window chunks
- Why is this change needed?
  - Try to avoid recording facts-count histogram samples for failed store attempts
  - Router

## Test Plan

## Test Result

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/memory/extractor.go
+++ b/src/semantic-router/pkg/memory/extractor.go
@@ -267,7 +267,7 @@ func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assis
 	chunk := formatTurnChunk(userMessage, assistantResponse)
 	sanitized, err := sanitizeMemoryContent(chunk)
 	if err != nil {
-		logging.Debugf("Memory chunk store: REJECTED - %v (user=%s)", err, userID)
+		logging.Debugf("Memory chunk store: sanitization rejected: not stored - %v (user=%s)", err, userID)
 		return false, nil
 	}
 	mem := &Memory{
--- a/src/semantic-router/pkg/memory/extractor.go
+++ b/src/semantic-router/pkg/memory/extractor.go
@@ -235,39 +235,40 @@ func (e *MemoryExtractor) ProcessResponseWithHistory(
 	}
 
 	startTime := time.Now()
-	turnStored := false
+	factsCount := 0
 	status := "success"
 	defer func() {
 		duration := time.Since(startTime).Seconds()
-		count := 1
-		if turnStored {
-			count = 1
-		}
-		RecordMemoryExtraction(status, duration, count, "episodic")
+		RecordMemoryExtraction(status, duration, factsCount, "episodic")
 	}()
 
 	// --- Per-turn chunk ---
 	if isLowEntropy(userMessage, assistantResponse) {
 		logging.Debugf("Memory chunk store: SKIPPED low-entropy turn for user=%s", userID)
-	} else if err := e.storeTurnChunk(ctx, userMessage, assistantResponse, userID); err != nil {
+	} else if stored, err := e.storeTurnChunk(ctx, userMessage, assistantResponse, userID); err != nil {
 		status = "error"
+		// Negative count skips the facts-count histogram for failed extraction attempts.
+		factsCount = -1
 		return err
-	} else {
-		turnStored = true
+	} else if stored {
+		factsCount++
 	}
 
 	// --- Session-level rolling window chunk ---
-	e.maybeStoreSessionChunk(ctx, userID, userMessage, assistantResponse, history)
+	if e.maybeStoreSessionChunk(ctx, userID, userMessage, assistantResponse, history) {
+		// The session window is stored as a separate episodic memory chunk.
+		factsCount++
+	}
 
 	return nil
 }
 
-func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assistantResponse, userID string) error {
+func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assistantResponse, userID string) (bool, error) {
 	chunk := formatTurnChunk(userMessage, assistantResponse)
 	sanitized, err := sanitizeMemoryContent(chunk)
 	if err != nil {
 		logging.Debugf("Memory chunk store: REJECTED - %v (user=%s)", err, userID)
-		return nil
+		return false, nil
 	}
 	mem := &Memory{
 		ID:         generateMemoryID(),
@@ -279,25 +280,26 @@ func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assis
 		Importance: 0.5,
 	}
 	if err := e.store.Store(ctx, mem); err != nil {
-		return fmt.Errorf("failed to store conversation chunk: %w", err)
+		return false, fmt.Errorf("failed to store conversation chunk: %w", err)
 	}
 	logging.Infof("Memory chunk store: stored turn for user=%s (len=%d)", userID, len(sanitized))
-	return nil
+	return true, nil
 }
 
 // maybeStoreSessionChunk stores a session-level chunk every `stride` turns.
 // Each chunk covers the last `windowSize` turns, creating overlapping windows
 // (stride < windowSize) so facts near window boundaries still co-occur in at
-// least one chunk -- critical for multi-hop retrieval.
+// least one chunk -- critical for multi-hop retrieval. It returns true when a
+// session chunk was actually stored.
 func (e *MemoryExtractor) maybeStoreSessionChunk(
 	ctx context.Context,
 	userID string,
 	userMessage string,
 	assistantResponse string,
 	history []openai.ChatCompletionMessageParamUnion,
-) {
+) bool {
 	if len(history) == 0 {
-		return
+		return false
 	}
 
 	windowSize := e.sessionWindowSize
@@ -312,23 +314,23 @@ func (e *MemoryExtractor) maybeStoreSessionChunk(
 	// history contains prior turns; +1 for the current turn
 	totalTurns := countTurns(history) + 1
 	if totalTurns < stride {
-		return
+		return false
 	}
 	// Fire on every stride-th turn (once we have enough turns for a window)
 	if totalTurns%stride != 0 {
-		return
+		return false
 	}
 
 	// Build session chunk from the last windowSize turns of history + current
 	sessionChunk := buildSessionChunk(history, userMessage, assistantResponse, windowSize)
 	if sessionChunk == "" {
-		return
+		return false
 	}
 
 	sanitized, err := sanitizeMemoryContent(sessionChunk)
 	if err != nil {
 		logging.Debugf("Memory session chunk: REJECTED - %v (user=%s)", err, userID)
-		return
+		return false
 	}
 
 	mem := &Memory{
@@ -343,11 +345,12 @@ func (e *MemoryExtractor) maybeStoreSessionChunk(
 
 	if err := e.store.Store(ctx, mem); err != nil {
 		logging.Warnf("Memory session chunk: failed to store for user=%s: %v", userID, err)
-		return
+		return false
 	}
 
 	logging.Infof("Memory session chunk: stored %d-turn window (stride=%d) for user=%s (len=%d)",
 		windowSize, stride, userID, len(sanitized))
+	return true
 }
 
 // countTurns counts user turns in a message history. Each user message is one turn.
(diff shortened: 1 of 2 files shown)


src/semantic-router/pkg/memory/extractor.go
```diff
@@ -235,39 +235,40 @@ func (e *MemoryExtractor) ProcessResponseWithHistory(
...
+	factsCount := 0
 	status := "success"
 	defer func() {
 		duration := time.Since(startTime).Seconds()
-		count := 1
-		if turnStored {
-			count = 1
-		}
-		RecordMemoryExtraction(status, duration, count, "episodic")
+		RecordMemoryExtraction(status, duration, factsCount, "episodic")
```
rootfs, line 242: [risk:high] RecordMemoryExtraction is called with factsCount which may be negative (sentinel). Ensure RecordMemoryExtraction explicitly ignores negative values (or add a guard here) to avoid corrupting histogram sums or producing invalid metric samples.

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -235,39 +235,40 @@ func (e *MemoryExtractor) ProcessResponseWithHistory(
...
 	}()
 
 	// --- Per-turn chunk ---
 	if isLowEntropy(userMessage, assistantResponse) {
 		logging.Debugf("Memory chunk store: SKIPPED low-entropy turn for user=%s", userID)
-	} else if err := e.storeTurnChunk(ctx, userMessage, assistantResponse, userID); err != nil {
+	} else if stored, err := e.storeTurnChunk(ctx, userMessage, assistantResponse, userID); err != nil {
 		status = "error"
+		// Negative count skips the facts-count histogram for failed extraction attempts.
```
rootfs, line 250: [risk:medium] factsCount = -1 is used as a sentinel to skip metrics. Prefer an explicit boolean (e.g., skipFactsMetric) or document the sentinel contract in RecordMemoryExtraction to avoid future misuse.

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -235,39 +235,40 @@ func (e *MemoryExtractor) ProcessResponseWithHistory(
...
+	if e.maybeStoreSessionChunk(ctx, userID, userMessage, assistantResponse, history) {
+		// The session window is stored as a separate episodic memory chunk.
+		factsCount++
+	}
 
 	return nil
 }
 
-func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assistantResponse, userID string) error {
+func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assistantResponse, userID string) (bool, error) {
```
rootfs, line 266: [risk:medium] storeTurnChunk signature changed to return (bool, error). Run a repo-wide search to ensure no other callers rely on the old signature; update all call sites or keep a shim to avoid breaking users.

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -235,39 +235,40 @@ func (e *MemoryExtractor) ProcessResponseWithHistory(
...
 	chunk := formatTurnChunk(userMessage, assistantResponse)
 	sanitized, err := sanitizeMemoryContent(chunk)
 	if err != nil {
 		logging.Debugf("Memory chunk store: REJECTED - %v (user=%s)", err, userID)
-		return nil
+		return false, nil
```
rootfs, line 271: [risk:low] Returning (false, nil) on sanitize rejection is fine, but consider clarifying in the log message why this is not an error to aid future debugging (e.g., 'sanitization rejected: not stored').

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -279,25 +280,26 @@ func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assis
 		Importance: 0.5,
 	}
 	if err := e.store.Store(ctx, mem); err != nil {
-		return fmt.Errorf("failed to store conversation chunk: %w", err)
+		return false, fmt.Errorf("failed to store conversation chunk: %w", err)
```
rootfs, line 283: [risk:medium] On store.Store error the function returns (false, fmt.Errorf(...)) and calling code sets factsCount=-1. Confirm RecordMemoryExtraction's handling of this path (should skip histogram) and add a unit test for this failure mode.

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -279,25 +280,26 @@ func (e *MemoryExtractor) storeTurnChunk(ctx context.Context, userMessage, assis
...
+// least one chunk -- critical for multi-hop retrieval. It returns true when a
+// session chunk was actually stored.
 func (e *MemoryExtractor) maybeStoreSessionChunk(
 	ctx context.Context,
 	userID string,
 	userMessage string,
 	assistantResponse string,
 	history []openai.ChatCompletionMessageParamUnion,
-) {
+) bool {
```
rootfs, line 300: [risk:medium] maybeStoreSessionChunk now returns bool; ensure all callers (including any outside this file) expect the boolean and that the return value is used consistently to increment factsCount only on success.

src/semantic-router/pkg/memory/extractor_test.go
```diff
@@ -409,6 +411,74 @@ func TestProcessResponseWithHistory_NilHistory(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			store := NewInMemoryStore()
+			extractor := NewMemoryChunkStore(store)
+			ctx := context.Background()
+
+			beforeCount, beforeSum := snapshot(t)
```
rootfs, line 488: [risk:medium] Tests snapshot a global histogram. Reset or isolate MemoryExtractionFactsCount between tests (or run sequentially) to avoid flakiness caused by metric state carried over from other tests.

src/semantic-router/pkg/memory/extractor_test.go
```diff
@@ -409,6 +411,74 @@ func TestProcessResponseWithHistory_NilHistory(t *testing.T) {
...
+				tt.user, tt.assistant, tt.history,
+			)
+			require.NoError(t, err)
+
+			results, err := store.List(ctx, ListOptions{UserID: "user1", Limit: 100})
+			require.NoError(t, err)
+			assert.Len(t, results.Memories, tt.wantStored)
+
+			afterCount, afterSum := snapshot(t)
+			assert.Equal(t, beforeCount+1, afterCount)
```
rootfs: [risk:medium] The test expects sample count to always increase by 1; add an additional test that simulates a store.Store error to verify the intended behavior when factsCount is -1 (i.e., metric should be skipped or handled explicitly).

rootfs (changes requested):
request_changes

Rationale: change correctly replaces a hard-coded count with explicit counting, but it relies on RecordMemoryExtraction and callers to correctly handle negative/zero counts and changes an internal helper signature that could break callers.

Risk
- Passing negative sentinel to metrics: src/semantic-router/pkg/memory/extractor.go:242 and src/semantic-router/pkg/memory/extractor.go:250–251 set factsCount = -1 and then call RecordMemoryExtraction(status, ..., factsCount, "episodic") — if RecordMemoryExtraction does not special-case negative counts, the histogram sum/behavior may be corrupted or cause metric errors.
- Change of internal helper signature: src/semantic-router/pkg/memory/extractor.go:266 changed storeTurnChunk(...) error → (bool, error). Any other in-repo callers (none found by the localizer) would require update; risk anchored to src/semantic-router/pkg/memory/extractor.go:266.
- Deferred metric semantics on early-return: src/semantic-router/pkg/memory/extractor.go:238–243 uses defer to record metrics; ensure all early returns (including non-nil err) produce the intended metric behavior (the -1 sentinel must be handled to avoid recording failed attempts).
- Shared global metric state in tests: src/semantic-router/pkg/memory/extractor_test.go:414–423 reads a global histogram (MemoryExtractionFactsCount). If other tests run in parallel or previous tests modify that metric, snapshots could be flaky unless the metric is reset or tests are serialized.

Tests
- New unit test covers the histogram behavior: src/semantic-router/pkg/memory/extractor_test.go:414–481 adds TestProcessResponseWithHistory_RecordsStoredChunkCountMetric which asserts sample count increments by 1 and sum increases by the number of stored chunks — good targeted coverage.
- Missing assertion for failed-store path: there is no unit test that exercises the code path that sets factsCount = -1 (failed store) to verify RecordMemoryExtraction ignores the negative sentinel. Add a test that simulates a store.Store(...) error and assert metrics unchanged or handled as intended (anchor: src/semantic-router/pkg/memory/extractor.go:248–252).
- Consider adding a test that runs with metric state reset/isolated to avoid interference with other tests that use MemoryExtractionFactsCount (anchor: extractor_test.go:414).

Style/Hygiene
- Naming is clear (factsCount). Comments updated to explain intent (e.g., maybeStoreSessionChunk comment now documents the bool return) — good.
- Minor churn: converting storeTurnChunk to return (bool,error) is reasonable; ensure signature change is limited to this hotspot only and helpers are kept narrow (anchor: src/semantic-router/pkg/memory/extractor.go:266).
- The comment "Negative count skips the facts-count histogram..." at src/semantic-router/pkg/memory/extractor.go:250 is useful but relies on implementation elsewhere; consider referencing the handling in RecordMemoryExtraction or asserting at compile/test time.

Size
- Scope is small and focused (2 files, ~95 additions). Splitting is unnecessary. Changes touch only extractor and its tests; appropriate for a single PR.

Suggested changes (concrete)
1. Add/verify RecordMemoryExtraction handles negative factsCount by ignoring histogram observation (or add explicit guard before calling RecordMemoryExtraction). Reference: src/semantic-router/pkg/memory/extractor.go:242, :250.
2. Add a unit test that simulates store.Store failing and asserts metric behavior (no negative sample sum): new test anchored to extractor_test.go near :414.
3. Ensure tests reset or isolate MemoryExtractionFactsCount between runs to avoid cross-test interference (extractor_test.go:414).
4. Run full build/tests to confirm no external callers were missed for storeTurnChunk signature change (extractor.go:266). Localizer found zero external refs, but run `rg`/build in CI to be safe.

If these are addressed, the change appears safe and improves metric correctness.

Tools used: NONE

src/semantic-router/pkg/memory/extractor.go
cryo-zd: Ref:
https://github.com/vllm-project/semantic-router/blob/a9e28e1b4937bf7d4063e1118cbcb7903e4db095/src/semantic-router/pkg/memory/metrics.go#L191-L201

src/semantic-router/pkg/memory/extractor.go
cryo-zd: The same to above comment, `RecordMemoryExtraction`  currently would not update metric `MemoryExtractionFactsCount`  if `factsCount` < 0

src/semantic-router/pkg/memory/extractor.go
cryo-zd: Checked.

src/semantic-router/pkg/memory/extractor.go
cryo-zd: Checked.

src/semantic-router/pkg/memory/extractor.go
cryo-zd: Fix in https://github.com/vllm-project/semantic-router/pull/2070/commits/037c64c01541f64e2e2dca6ca889159e30e56499

src/semantic-router/pkg/memory/extractor.go
cryo-zd: Ref:
https://github.com/vllm-project/semantic-router/blob/a9e28e1b4937bf7d4063e1118cbcb7903e4db095/src/semantic-router/pkg/memory/metrics.go#L191-L201

Test added in https://github.com/vllm-project/semantic-router/pull/2070/commits/9bf4938c48a77ffd6c22577e678baabd8e3f6172

src/semantic-router/pkg/memory/extractor_test.go
cryo-zd: Cleanup and reset per test case, ref https://github.com/vllm-project/semantic-router/pull/2070/commits/9bf4938c48a77ffd6c22577e678baabd8e3f6172

src/semantic-router/pkg/memory/extractor_test.go
cryo-zd: Test added in https://github.com/vllm-project/semantic-router/pull/2070/commits/9bf4938c48a77ffd6c22577e678baabd8e3f6172

Issue #1815, opened 2026-04-23
[Feature] Add per-step controls and accounting to static Router Flow workflows


AayushSaini101:
### Goal

Example: summarize with model A, extract structured fields with model B from the summary, then return a final response. Operators should be able to reason about the model, input dependency, timeout, failure behavior, output contract, and resource use of each step.

### Proposed Workgroup

MoM & Routing

### Scope

- Add stable per-step identity and explicit single-model pinning where a fixed pipeline requires it.
- Define how a step consumes reviewed outputs from earlier steps.
- Add per-step timeout, failure/skip policy, token limit, and optional output contract.
- Emit per-step latency, token, cost, model, status, and failure receipts to diagnostics/Replay.
- Preserve workflow-wide safety, authorization, cancellation, and maximum-amplification limits.
- Provide migration/examples using the existing `workflows` config surface.

##### Audited upstream baseline (2026-07-15)

The original gap is partly implemented on current `main`. A decision can use `algorithm.type: workflows` with `mode: static`; ordered `roles` declare model lists, prompts, and access lists, followed by a configured `final` model. The runtime also has workflow-wide `max_steps`, `round_timeout_seconds`, `on_error`, concurrency, and intermediate-response controls.

The remaining gap is a narrow fixed-pipeline contract with explicit controls and receipts for each ordered step. This issue must extend the existing static Router Flow implementation, not introduce a second pipeline engine.

##### Relevant surfaces

`src/semantic-router/pkg/config/workflows_config.go`, `src/semantic-router/pkg/looper/workflows*.go`, Replay/diagnostics, docs/examples, and E2E tests.

### Non-goals

- Do not create an arbitrary user-programmable DAG or external workflow engine.
- Do not duplicate static Workflows under a new top-level API.
- Do not allow a step to bypass the decision candidate set, plugins, or provider policy.
- Existing single-step routing and existing static/dynamic workflow configs remain compatible.

### Completion signal

- [ ] A two-step A-then-B pipeline is expressible through the existing Workflows contract.
- [ ] Each step has deterministic model/input/error semantics.
- [ ] Per-step and aggregate usage are both observable.
- [ ] Cancellation and timeout stop downstream work safely.
- [ ] Tests cover success, timeout, skipped/failed step, output-contract failure, and streaming behavior.
- [ ] Docs clearly distinguish fixed static pipelines from dynamic workflows and hierarchical recipe routing (#2343).

AayushSaini101:
[...]

PR #2075, opened 2026-06-07
[Docs] Add multi-step pipeline routing example

xiaoyu-xyz:
## Summary
- add keyword signal and composite decision fragments for client-orchestrated multi-step routing
- document how step markers such as `__pipeline_step:summarize__` and `__pipeline_step:extract__` can pin each step to a model
- clarify that the client still orchestrates multiple requests while Semantic Router handles per-step model routing

## Validation
- `agent-ci-gate CHANGED_FILES="config/signal/keyword/multi-step-pipeline.yaml,config/decision/composite/multi-step-pipeline.yaml,website/docs/tutorials/decision/multi-step-pipeline.md,website/docs/tutorials/decision/overview.md,website/sidebars.ts"`

Refs #1815

### DIFF
--- a/config/decision/composite/multi-step-pipeline.yaml
+++ b/config/decision/composite/multi-step-pipeline.yaml
@@ -13,7 +13,7 @@ routing:
               - type: keyword
                 name: pipeline_step_extract
       modelRefs:
-        - model: summarizer-model
+        - model: qwen3-8b
           use_reasoning: false
     - name: pipeline_extract_step
       description: Pin the extraction stage of a client-orchestrated pipeline.
@@ -28,5 +28,5 @@ routing:
               - type: keyword
                 name: pipeline_step_summarize
       modelRefs:
-        - model: insight-model
+        - model: qwen3-32b
           use_reasoning: true
--- a/config/decision/composite/multi-step-pipeline.yaml
+++ b/config/decision/composite/multi-step-pipeline.yaml
@@ -1,0 +1,32 @@
+routing:
+  decisions:
+    - name: pipeline_summarize_step
+      description: Pin the summarization stage of a client-orchestrated pipeline.
+      priority: 190
+      rules:
+        operator: AND
+        conditions:
+          - type: keyword
+            name: pipeline_step_summarize
+          - operator: NOT
+            conditions:
+              - type: keyword
+                name: pipeline_step_extract
+      modelRefs:
+        - model: summarizer-model
+          use_reasoning: false
+    - name: pipeline_extract_step
+      description: Pin the extraction stage of a client-orchestrated pipeline.
+      priority: 190
+      rules:
+        operator: AND
+        conditions:
+          - type: keyword
+            name: pipeline_step_extract
+          - operator: NOT
+            conditions:
+              - type: keyword
+                name: pipeline_step_summarize
+      modelRefs:
+        - model: insight-model
+          use_reasoning: true
--- a/config/signal/keyword/multi-step-pipeline.yaml
+++ b/config/signal/keyword/multi-step-pipeline.yaml
@@ -1,0 +1,11 @@
+routing:
+  signals:
+    keywords:
+      - name: pipeline_step_summarize
+        operator: OR
+        keywords: ["__pipeline_step:summarize__", "__step1__"]
+        case_sensitive: false
+      - name: pipeline_step_extract
+        operator: OR
+        keywords: ["__pipeline_step:extract__", "__step2__"]
+        case_sensitive: false
--- a/website/docs/tutorials/decision/multi-step-pipeline.md
+++ b/website/docs/tutorials/decision/multi-step-pipeline.md
@@ -1,0 +1,98 @@
+# Multi-Step Pipeline Routing
+
+## Overview
+
+Use this pattern when a client already orchestrates a fixed multi-step workflow, but each step should be routed to a different model by Semantic Router.
+
+Semantic Router does not execute a full pipeline graph for the client. The client still sends one request per step. Each request carries a stable step marker, such as `__pipeline_step:summarize__` or `__pipeline_step:extract__`, and normal keyword signals map those markers to route decisions.
+
+## Key Advantages
+
+- Uses the existing signal and decision surfaces instead of adding a new pipeline runtime.
+- Keeps the step-to-model binding explicit and reviewable.
+- Lets operators pin each stage to a model while retaining normal routing observability.
+- Avoids hidden orchestration state inside the router; the client owns step order and request chaining.
+
+## What Problem Does It Solve?
+
+Some applications want predictable staged routing, such as summarizing a long document with one model and extracting structured insights with another model. That is not the same as asking Semantic Router to pick the best single model for a request.
+
+This pattern solves the routing part of that workflow. The client labels each request with the stage it is currently running, and Semantic Router picks the configured model for that stage.
+
+## When to Use
+
+Use multi-step pipeline routing when:
+
+- the application already knows the workflow order
+- each step is a separate request to the router
+- each step should be pinned to a known model or candidate set
+- stable lexical markers are acceptable for identifying the step
+
+Do not use this pattern when the router needs to decide the next step, run multiple steps automatically, or manage a dynamic DAG. Those remain application orchestration responsibilities.
+
+## Configuration
+
+Source fragments:
+
+- `config/signal/keyword/multi-step-pipeline.yaml`
+- `config/decision/composite/multi-step-pipeline.yaml`
+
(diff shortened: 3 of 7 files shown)


config/decision/composite/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,32 @@
...
+        operator: AND
+        conditions:
+          - type: keyword
+            name: pipeline_step_summarize
+          - operator: NOT
+            conditions:
+              - type: keyword
+                name: pipeline_step_extract
+      modelRefs:
+        - model: summarizer-model
```
rootfs: [risk:high] The model identifier `summarizer-model` looks like a placeholder; ensure this exact model name exists in operator configuration or document that operators must map this to a real model to avoid runtime routing failures.

config/decision/composite/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,32 @@
...
+        operator: AND
+        conditions:
+          - type: keyword
+            name: pipeline_step_extract
+          - operator: NOT
+            conditions:
+              - type: keyword
+                name: pipeline_step_summarize
+      modelRefs:
+        - model: insight-model
```
rootfs: [risk:high] The model identifier `insight-model` is unvalidated here; add an explicit note or validation step to prevent silent misconfiguration at deploy time.

config/signal/keyword/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,11 @@
+routing:
+  signals:
+    keywords:
+      - name: pipeline_step_summarize
+        operator: OR
+        keywords: ["__pipeline_step:summarize__", "__step1__"]
```
rootfs, line 6: [risk:medium] The keywords `"__pipeline_step:summarize__"` and `"__step1__"` are lexically visible to models; recommend guidance on choosing markers that won't collide with user content and whether to escape or wrap them.

config/decision/composite/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,32 @@
...
+  decisions:
+    - name: pipeline_summarize_step
+      description: Pin the summarization stage of a client-orchestrated pipeline.
+      priority: 190
+      rules:
+        operator: AND
+        conditions:
+          - type: keyword
+            name: pipeline_step_summarize
+          - operator: NOT
```
rootfs, line 11: [risk:medium] The `NOT` guard prevents dual markers but behavior for requests containing both is ambiguous; add a test or explicit documented fallback explaining which rule wins.

website/docs/tutorials/decision/multi-step-pipeline.md
```diff
@@ -0,0 +1,98 @@
+# Multi-Step Pipeline Routing
+
+## Overview
+
+Use this pattern when a client already orchestrates a fixed multi-step workflow, but each step should be routed to a different model by Semantic Router.
+
+Semantic Router does not execute a full pipeline graph for the client. The client still sends one request per step. Each request carries a stable step marker, such as `__pipeline_step:summarize__` or `__pipeline_step:extract__`, and normal keyword signals map those markers to route decisions.
```
rootfs, line 7: [risk:medium] Docs note that markers are model-visible; add an explicit example showing how to wrap markers in an app-controlled wrapper or instruct models to ignore them to avoid prompt-injection or leakage concerns.

website/sidebars.ts
```diff
@@ -126,6 +126,7 @@ const sidebars: SidebarsConfig = {
             'tutorials/decision/or',
             'tutorials/decision/not',
             'tutorials/decision/composite',
+            'tutorials/decision/multi-step-pipeline',
```
rootfs, line 129: [risk:medium] New sidebar entry `tutorials/decision/multi-step-pipeline` must match the actual doc ID/path; verify the docs build resolves this path and update if the route differs to avoid broken navigation.

config/signal/keyword/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,11 @@
+routing:
+  signals:
+    keywords:
+      - name: pipeline_step_summarize
```
rootfs, line 4: [risk:medium] Add unit tests for the `pipeline_step_summarize` signal to assert detection for both marker variants and case-insensitivity (`case_sensitive: false`).

config/decision/composite/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,32 @@
+routing:
+  decisions:
+    - name: pipeline_summarize_step
+      description: Pin the summarization stage of a client-orchestrated pipeline.
+      priority: 190
+      rules:
+        operator: AND
+        conditions:
+          - type: keyword
```
rootfs, line 9: [risk:medium] Add unit tests for the `pipeline_summarize_step` decision to ensure rules (keyword match + NOT guard) choose `summarizer-model` and that priority interacts as expected with other decisions.

config/decision/composite/multi-step-pipeline.yaml
```diff
@@ -0,0 +1,32 @@
...
+      modelRefs:
+        - model: summarizer-model
+          use_reasoning: false
+    - name: pipeline_extract_step
+      description: Pin the extraction stage of a client-orchestrated pipeline.
+      priority: 190
+      rules:
+        operator: AND
+        conditions:
+          - type: keyword
```
rootfs, line 24: [risk:medium] Similarly add tests for `pipeline_extract_step` (keyword + NOT guard) to confirm `insight-model` selection and to catch ambiguous-marker edge cases.

rootfs (changes requested):
Verdict
request_changes — docs + config are clear but two actionable issues: unvalidated model refs and missing tests/validation.

Risk
- Unknown model identifiers may break routing at deploy time (config/decision/composite/multi-step-pipeline.yaml:N16, N31).
- Step markers are model-visible and could leak or confuse prompts if misused (website/docs/tutorials/decision/multi-step-pipeline.md:N7, N93-N98).
- Docs/site integration risk: adding sidebar entry must match docs build expectations; a wrong path will fail the site build (website/sidebars.ts:N129 → website/docs/tutorials/decision/multi-step-pipeline.md:N37-N39).

Tests
- No tests added for new signals/decisions. Add unit tests verifying keyword detection and decision matching for:
  - config/signal/keyword/multi-step-pipeline.yaml:N4-N10
  - config/decision/composite/multi-step-pipeline.yaml:N9-N15, N24-N31
- Also add/verify website build test covers the new page and sidebar entry.

Style/Hygiene
- YAML fragments follow repo style and include guards (NOT) — good. Consider documenting example client request payload snippet showing where to place markers (docs: website/docs/tutorials/decision/multi-step-pipeline.md:N95-N97).
- Consider clarifying that model names are placeholders or require operator config.

Size
- Scope is small (5 files, focused docs + config). No need to split, but require the small follow-ups above before merging.

Tools used: NONE

xiaoyu-xyz:
Thanks for the review. I pushed an update addressing the requested follow-ups:

- Replaced the placeholder `modelRefs` with reference-config model names (`qwen3-8b` and `qwen3-32b`) and documented that operators should replace them with models registered in their own deployment.
- Added a concrete client request payload showing the step marker inside an app-controlled wrapper, with guidance that markers are model-visible, should not come from user input, and should be ignored by the target model.
- Documented the ambiguous-marker behavior: if both step markers are present, the `NOT` guards prevent both pipeline decisions from matching and normal fallback routing applies.
- Added fragment-backed unit tests for keyword marker detection, alias detection, case-insensitive matching, per-step decision selection, priority behavior against a generic route, and the ambiguous-marker no-match case.

Local validation passed:

- `go test ./pkg/classification ./pkg/decision -run 'TestMultiStepPipeline' -count=1`
- `npm run build` in `website/`
- `make agent-ci-gate` with the PR changed files

xiaoyu-xyz:
Hi @rootfs, I rebased this PR on the latest main and all checks are green now. The requested follow-ups were addressed in the latest commits. Could you take another look when you have a chance?

Xunzhuo (changes requested):
The earlier code and documentation follow-ups are present, but this still has no accepted implementation issue. #1815 is needs-acceptance and describes a different runtime contract. Please link an accepted issue that owns this example and test scope before another review.

FAUST-BENCHOU:
Thanks for the contribution. The example is clear, but I think this pattern is better handled at the application/workflow layer rather than encoded as prompt keywords. Semantic Router should focus on routing decisions based on request context and metadata instead of becoming a pipeline state coordinator. Closing this for now, but the workflow metadata based approach could be revisited in the future.

PR #1864, opened 2026-05-01
feat(signals): configurable confidence threshold for language signal (#1723)

ramkrishs:
## Purpose

Two improvements to the signal evaluation pipeline:

1. **Configurable language signal threshold** — addresses #1723 (follow-up from merged #1720)
2. **Signal dispatcher panic recovery** — proactive stability hardening: even though #1843 is closed, signal goroutines currently have no panic recovery and a classifier crash still silently kills the router

**Affected modules:** `[Router][Classification][Observability]`

---

### Root cause
The lingua-go confidence floor was hardcoded at `0.3` and not configurable through the signal API. Operators routing on language (e.g. compliance filtering for specific locales) had no way to tune the sensitivity.

### Fix
Add `Threshold float32` to `LanguageRule`. Default `0` = current `0.3` behaviour preserved. Non-zero value overrides the floor for that specific rule.

```yaml
routing:
  signals:
    language:
      - name: zh
        description: Chinese-language requests.
        threshold: 0.6   # require higher confidence for this rule
      - name: es
        description: Spanish-language requests.
        # threshold omitted → uses built-in 0.3 default
```

### Changed files
| File | Change |
|------|--------|
| `pkg/config/signal_config.go` | `Threshold float32` added to `LanguageRule` |
| `pkg/classification/language_classifier.go` | New `ClassifyWithThreshold(text, threshold)` method; exposes `defaultLanguageThreshold` const |
| `pkg/classification/classifier_signal_language.go` | New file — `evaluateLanguageSignal` + `lowestLanguageThreshold` (extracted from `classifier_signal_context.go` to stay under the 800-line structure limit) |
| `pkg/classification/classifier_signal_context.go` | Remove extracted function |
| `config/config.yaml` | Document `threshold: 0.0` on reference language rule |

---

### Motivation
Signal evaluators run concurrently in goroutines. A panic in any evaluator (e.g. an out-of-bounds access in a Rust binding triggered by a very long prompt) currently propagates through the goroutine scheduler and kills the router process with no log output and no metric — even if the root crash bug is fixed, the recovery path is missing.

### Fix
Add `recoverSignalPanic` as a deferred recovery in every signal goroutine. On a recovered panic it logs the incident and increments a new `llm_signal_panic_total` Prometheus counter so operators can alert on it rather than discover crashes post-mortem.

### Changed files
| File | Change |
|------|--------|
| `pkg/classification/classifier_signal_dispatch.go` | Add `defer recoverSignalPanic(...)` + helper |
[...]

PR #2086, opened 2026-06-08
[DSL] Expose language signal threshold in compiler/decompiler (#1723)

WUKUNTAI-0211:
## Summary

Wire `threshold` through the DSL surface for the language signal so the
runtime field added in #1864 can actually be authored from DSL and round-
tripped via the decompiler. Closes the gap left by #1723.

## Background

- #1723 asked for a configurable score threshold on the language signal.
- #1864 (merged) added `Threshold float32` to `config.LanguageRule` and
  wired it into the language classifier. Default `0` preserves the
  built-in `0.3` floor; any non-zero value overrides it.
- The DSL compiler/decompiler was not updated in #1864, so authoring a
  language threshold via DSL silently drops it, and a router config that
  has one decompiles to DSL without the field.

This PR fixes that asymmetry by mirroring the existing
`PreferenceRule.Threshold` pattern in the three relevant DSL files.

## Changes

| File | Change |
|------|--------|
| `src/semantic-router/pkg/dsl/compiler_signals.go` | `compileLanguageSignal` reads optional `threshold` field via `getFloat32Field` into `LanguageRule.Threshold` |
| `src/semantic-router/pkg/dsl/decompiler_convert.go` | `languageToSignal` emits `threshold` only when non-zero (preserves zero-default contract) |
| `src/semantic-router/pkg/dsl/decompiler_signals_rules.go` | `decompileLanguageSignals` writes `threshold: %g` only when non-zero |
| `src/semantic-router/pkg/dsl/dsl_test.go` | Three existing `SIGNAL language lang` fixtures gain `threshold: 0.6`; added compile-time, YAML round-trip, and decompile assertions |

Total diff: **4 files, +21 / -3**.

### Example

```dsl
SIGNAL language zh {
  description: "Chinese-language requests."
  threshold: 0.6
}
```

Decompiles back to the same DSL; YAML round-trips through `config.RouterConfig`
preserving `Threshold: 0.6`. Omitting `threshold` keeps the prior behaviour
(classifier uses its built-in `0.3` default).

## Test Plan

- [x] `go test ./pkg/dsl/... -count=1` — pass (`ok ... 1.395s`)
- [x] `make agent-lint CHANGED_FILES="<the four files>"` — Passed (`go fmt`, `go lint`, `agent changed-files lint`, `golangci-lint`, `TestReferenceConfig`, structure checks; all warnings are pre-existing baselines)
- [x] Targeted runs: `TestAllSignalTypes`, `TestAllSignalTypesRoundTrip`, `TestDecompileAllSignals*` pass with the new assertions

Full `make test` from `cmd/` is skipped locally because the host is missing
the CGO bindings (`candle-binding`, `nlp-binding`, `ml-binding`, etc.) —
CI matrix will validate them.

## Notes

- Default (`threshold` omitted, zero value) keeps the existing 0.3 floor
  behaviour, so this is a backwards-compatible extension.
- Touches `pkg/dsl/**` only; no `config.go` or schema changes (the schema
  side was already done in #1864).
- Supersedes the auto-generated PR #2082, which is failing
  `pre-commit lint`, `build_pr (dashboard)`, and `test-and-build`. Happy
  to close #2082 in favour of this once a maintainer signs off.

Resolves #1723.

### DIFF
--- a/src/semantic-router/pkg/dsl/compiler_signals.go
+++ b/src/semantic-router/pkg/dsl/compiler_signals.go
@@ -146,6 +146,9 @@ func (c *Compiler) compileLanguageSignal(s *SignalDecl) {
 	if v, ok := getStringField(s.Fields, "description"); ok {
 		rule.Description = v
 	}
+	if v, ok := getFloat32Field(s.Fields, "threshold"); ok {
+		rule.Threshold = v
+	}
 	c.config.LanguageRules = append(c.config.LanguageRules, rule)
 }
 
--- a/src/semantic-router/pkg/dsl/decompiler_convert.go
+++ b/src/semantic-router/pkg/dsl/decompiler_convert.go
@@ -104,6 +104,9 @@ func (d *decompiler) languageToSignal(lang *config.LanguageRule) *SignalDecl {
 	if lang.Description != "" {
 		fields["description"] = StringValue{V: lang.Description}
 	}
+	if lang.Threshold != 0 {
+		fields["threshold"] = FloatValue{V: float64(lang.Threshold)}
+	}
 	return &SignalDecl{SignalType: "language", Name: lang.Name, Fields: fields}
 }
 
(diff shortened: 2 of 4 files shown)


FAUST-BENCHOU (approved):
LGTM.Thanks.Have all the tasks https://github.com/vllm-project/semantic-router/issues/1723 here been completed?

FAUST-BENCHOU:
never mind i think its reasonable to close this for now

Issue #2010, opened 2026-06-02
bug: <Auth Related Integ Test Failing>


Peterren:
### Describe the bug

https://github.com/vllm-project/semantic-router/actions/runs/26796010459/job/78992458919

Failed Auth Related Integ Test

### To Reproduce

kick off PR run CICD

### Affected layer

None

rootfs:
@Peterren I am trying some experimentations using this issue. Please don't hesitate to submit your PR if you have a fix. Will merge your PR once the CI is green.

drivebyer:
This has been fixed — see:
https://github.com/vllm-project/semantic-router/blob/f08056e793434ad5ccec8054fe34d43baff52fdd/e2e/profiles/dashboard/dashboard-deployment.yaml#L74-L75

Feel free to reopen if it reproduces.

PR #2090, opened 2026-06-08
[auto-pareto/strong] bug: <Auth Related Integ Test Failing>

rootfs:
Auto-resolve for #2010 (verdict: lgtm)

Localizer brief:

## Summary

**Mode: fix**

### Root cause
The `authz-rbac-routing` test case is **registered** (line 14 of `e2e/testcases/authz_rbac_routing.go`) but **NOT listed** in the e2e profile's `GetTestCases()` method (`e2e/profiles/authz-rbac/profile.go:175-186`). The test attempts to send authz headers directly (`x-authz-user-id`, `x-authz-user-groups`), which works with the default header names in the config, but:

1. The profile's `values.yaml` (line 192-193) configures `authz.fail_open: true` **without** defining an explicit `identity` section
2. The test assumes role bindings will match against headers like `x-authz-user-id: "alice"` + `x-authz-user-groups: "platform-admins"` → role `admin`
3. **Missing**: The `identity` configuration block is absent, so while defaults apply internally, the config sent to the router lacks explicit validation that headers are properly configured

### Affected files
- **`e2e/profiles/authz-rbac/profile.go`** (line 175-186) — Missing `authz-rbac-routing` from `GetTestCases()` list; only runs 3 tests instead of 4
- **`e2e/profiles/authz-rbac/values.yaml`** (line 192-193) — Missing explicit `identity` section under `authz`; relies on defaults silently
- **`e2e/testcases/authz_rbac_routing.go`** (line 14) — Registered but not scheduled by the profile

### Anchor symbols
- `Profile.GetTestCases()` at `/home/azureuser/ast/semantic-router/e2e/profiles/authz-rbac/profile.go:175`
- `authz-rbac-routing` test registration at `/home/azureuser/ast/semantic-router/e2e/testcases/authz_rbac_routing.go:14`
- `IdentityConfig.GetUserIDHeader()` at `/home/azureuser/ast/semantic-router/src/semantic-router/pkg/config/config.go:104`
- `AuthzClassifier.Classify()` at `/home/azureuser/ast/semantic-router/src/semantic-router/pkg/classification/authz_classifier.go:118`

### Reference call-sites
- Default header usage: `/home/azureuser/ast/semantic-router/src/semantic-router/pkg/classification/classifier.go:171-172`
- Test fixture header passing: `/home/azureuser/ast/semantic-router/e2e/pkg/fixtures/chat.go:56-62`

### Runner/CI dependencies
- CI workflow: `/home/azureuser/ast/semantic-router/.github/workflows/integration-test-k8s.yml:64` (authz-rbac profile detection)

### Tests to update or add
1. **Add** `"authz-rbac-routing"` to `GetTestCases()` list in `/home/azureuser/ast/semantic-router/e2e/profiles/authz-rbac/profile.go:175`
2. **Add** explicit `identity` section to `/home/azureuser/ast/semantic-router/e2e/profiles/authz-rbac/values.yaml` under `global.services.authz` (below line 193):
   ```yaml
   identity:
     user_id_header: x-authz-user-id
     user_groups_header: x-authz-user-groups
   ```

### Risks / unknowns
- Whether the Envoy configuration in the profile actually strips/validates JWT before reaching ext_proc (affects whether direct headers work)
- Exact assertion error from the failing CI run (need GitHub Actions logs to confirm expected vs actual decision headers)

Verifier findings:

Verdict: lgtm — Fix patch that adds a missing test case registration and required config to the authz-rbac profile.

## Findings

- **`e2e/testcases/authz_rbac_routing.go:138` — Weak failure threshold.** The test only returns an error if `passed == 0`. Since the `no_identity_default` sub-case (empty headers, any 200 OK) will always pass when `fail_open: true`, the test as written cannot ever return an error even if all RBAC routing assertions fail. This means CI will show green for `authz-rbac-routing` even if none of the role-based decisions (admin_unrestricted, premium_complex, etc.) are working. This is a pre-existing issue in the test file, not introduced by this patch, but the patch adds this test to CI without calling attention to it.

- **`e2e/profiles/authz-rbac/values.yaml` — `identity` block is functionally redundant.** The router's `GetUserIDHeader()` / `GetUserGroupsHeader()` already default to `x-authz-user-id` / `x-authz-user-groups` when the config is absent (`config.go:106,113`). Adding them explicitly is harmless but provides no behavioral change. The patch is not wrong here.

## Required follow-ups

- The `authz-rbac-routing` test's pass criterion (`passed == 0` guard) should be tightened to require that all cases with a non-empty `wantDecision` pass — otherwise this test adds no CI signal for RBAC routing correctness.

### DIFF
--- a/e2e/profiles/authz-rbac/README.md
+++ b/e2e/profiles/authz-rbac/README.md
@@ -4,7 +4,7 @@ This profile demonstrates user-level RBAC model routing where different users (a
 
 ## Security Requirements
 
-**IMPORTANT**: Identity headers (`x-authz-user-id`, `x-authz-user-groups`) must **ONLY** come from validated auth backends (JWT, Authorino, etc.), **NOT** from client requests.
+**IMPORTANT**: This profile trusts JWT-derived identity headers (`x-jwt-sub`, `x-jwt-groups`) only after Envoy validates a bearer token. Clients must not be allowed to set those headers directly.
 
 ### Envoy Configuration
 
@@ -21,9 +21,9 @@ The header removal filter runs **before** the JWT/auth filter and ext_proc filte
 
 For e2e tests:
 
-- **DO NOT** send `x-authz-user-id` or `x-authz-user-groups` headers directly from the test client
+- **DO NOT** send `x-jwt-sub` or `x-jwt-groups` headers directly from the test client
 - **DO** send JWT tokens in the `Authorization: Bearer <token>` header
-- Envoy Gateway validates the JWT and extracts claims into identity headers
+- Envoy Gateway validates the JWT and copies claims into `x-jwt-sub` / `x-jwt-groups`
 - The router receives only JWT-derived identity headers
 
 ### Production Deployment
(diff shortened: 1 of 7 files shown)


rootfs:
Closing: cohort run for offline Pareto calibration; not intended for merge.

Issue #2008, opened 2026-06-02
feature: model-backed classifier for the complexity signal (multi-head mmBERT) instead of embedding/prototype scoring


hung1012:
### Describe the feature

Support a fine-tuned **classifier head** as a source for the complexity signal
(`global.model_catalog.modules.complexity`), as an alternative to the current
contrastive-embedding / `prototype_scoring` backend. Concretely, allow a
multi-head encoder fine-tuned from `mmBERT-base` — one head for **domain**, one
head for **complexity** — to emit both signals from a single forward pass, so
complexity becomes a calibrated classifier prediction rather than a similarity
score against easy/hard exemplar banks.

### Primary layer

signal

### Why this layer?

The complexity signal is extracted in the signal layer
(`global.model_catalog.modules.complexity`). Today that module is backed by a
contrastive embedding classifier: each complexity rule defines hard vs. easy
exemplar banks whose embeddings are precomputed at init, and a query's
difficulty is estimated by similarity to those prototypes (`prototype_scoring`).

This request changes the *source* of that signal — swapping the
embedding/prototype backend for a fine-tuned classifier head — so it belongs in
the signal layer alongside the existing domain classifier path. It does not
touch the projection or decision layers; the emitted complexity score keeps the
same downstream contract (`routing.projections.scores` / `mappings` /
`decisions`).

### Why do you need this feature?

Prototype/embedding similarity is a weak proxy for complexity:

- Complexity is not a "nearest-exemplar" property. Two prompts can be
  semantically close yet differ sharply in reasoning depth, so similarity to an
  easy/hard exemplar bank mislabels difficulty.
- It requires a curated, per-deployment exemplar bank that has to be maintained
  and re-embedded.
- The output is a similarity score rather than a calibrated complexity
  prediction, which makes thresholds/bands in `projections` hard to tune.

Meanwhile the domain signal is already produced by a fine-tuned classifier,
while complexity is not — an inconsistency a single classifier can resolve.

I have a multi-head model fine-tuned from `mmBERT-base` with:

- Head 1: **domain** classification
- Head 2: **complexity** classification

This is directly aligned with the project's existing mmBERT/ModernBERT classifier
stack and emits both signals from one forward pass — lower latency/cost than
embedding similarity + a separate domain classifier, a calibrated complexity
output, and no exemplar bank to maintain.

### Additional context

**Proposal.** Allow the `complexity` module to be backed by a classifier
[...]

PR #2093, opened 2026-06-09
[Feature] Add a trained-classifier mode to the complexity signal

### LINKED ISSUE #2568: [Feature] Add trained-classifier mode to the complexity signal
### Goal

Allow the complexity signal to consume a qualified trained classifier in addition to its current supported modes.

### Proposed Workgroup

### Scope

- Define the model task, labels or score mapping, artifact identity, and resolved binding.
- Establish a held-out baseline and calibration gate through #3197.
- Integrate the model through the shared backend and result contracts.
- Preserve existing modes and make selection explicit in configuration.
- Add runtime, API, and end-to-end coverage.

### Non-goals

- Changing the default without qualification evidence.
- Encoding routing thresholds or model selection in the model runtime.

Parent Epic: #3194

### Completion signal

hung1012:
Closes #2568 
  Part of #2008
     
  > **Scope note:** #2008 asks for domain + complexity as two heads sharing one
  > mmBERT forward pass. This PR delivers complexity as a **separate single-head**
  > classifier with its own inference path (opt-in per rule via `method: model`);
  > both modes emit the same `rule:difficulty` output so decisions/projections are
  > unchanged. The shared multi-head forward pass is intentionally out of scope
  > here to keep the PR focused, and is tracked as follow-up work.
  
  ## Purpose 
  
  - **What:** Adds a per-rule trained-classifier mode to the `complexity` signal — a rule can set `method: model` to be decided by a fine-tuned 3-class (easy/medium/hard) ModernBERT
  classifier instead of the default embedding/prototype-bank path. Both modes emit the same `rule:difficulty` output, so decisions/projections are unchanged. Spans the
  candle-binding FFI, config schema + mapping + load-time validation, dispatch wiring, DSL round-trip, the `vllm-sr` CLI schema, the Kubernetes Operator CRD, canonical docs, and
  unit + e2e tests.
  - **Why:** Embedding similarity alone doesn't capture reasoning difficulty well; a trained head gives a direct estimate. Mirrors the jailbreak signal's dual-mode pattern. Model
  mode is text-only and opt-in (backward-compatible).
  - **Modules:** `Router`, `Bindings`, `CLI`, `Operator`, `Docs`, `E2E`
  
  ## Test Plan
  
  - Go: `gofmt`, `go vet`, and `golangci-lint --new-from-rev` across the changed router, e2e, and operator packages; `go test` for `pkg/config`, `pkg/classification`, `pkg/dsl`, and
  the operator `controllers`/`api` packages.
  - Validation coverage: unit tests for complexity `method`/model-config validation, mapping label/index validation, and the CLI `ComplexityRule` shape.
  - CLI: `pytest` (complexity-rule + config-contract + config suites) and `black`.
  - Operator: `make manifests generate bundle` (regenerate CRD, OLM bundle, deepcopy) + `operator-sdk bundle validate`; conversion test for `method: model` + the typed
  `complexity_classifier` module.
  - E2E: new `complexity-model-routing` case under the `routing-strategies` profile that boots the router with a `method: model` rule backed by a real 3-class classifier
  auto-downloaded via `mom_registry`, and asserts the emitted `x-vsr-matched-complexity` `rule:easy|medium|hard` header.
  - Docs/config: `markdownlint`, YAML validity, and the config-tree docs-contract + reference-config tests.
  
  ## Test Result
  
  - golangci-lint v2.5.0: **0 new issues**; `go vet` / `gofmt` clean; all unit + pipeline tests pass, including existing complexity, config, and reference-config tests.
  - Operator: CRD + OLM bundle + deepcopy regenerate clean from a verified baseline; `operator-sdk bundle validate` passes.
  - E2E: `complexity-model-routing` added and wired into the `routing-strategies` profile; it runs in CI (which provisions the kind cluster and auto-downloads
  `Gnuh1M/mmbert-difficulty-classifier`). Locally verified to build, vet, and lint; the full stack boot is exercised in CI.
  - Model mode verified end-to-end against a real fine-tuned ModernBERT 3-class checkpoint loading via the new FFI.
  - Notes: `cargo clippy -D warnings` only flags pre-existing FFI warnings (present on `main`, not run by CI).
  
  ---
  <details>
  
  </details>


hung1012:
Hi @rootfs @Xunzhuo, could you review this PR and let me know whether we can move forward with applying this update to complexity signals? Please let me know if you have any concerns or suggestions. Thanks!

hung1012:
Hi @AayushSaini101, does this PR look good to you? Please let me know if you need any clarification or help from my side.

candle-binding/semantic-router.go
```diff
@@ -2632,6 +2636,50 @@ func ClassifyFactCheckText(text string) (ClassResult, error) {
 	}, nil
 }
 
+// InitComplexityClassifier initializes the ModernBERT complexity classifier.
+// This model predicts prompt reasoning difficulty; the predicted class index is mapped
+// to a difficulty label (conventionally easy/medium/hard) by the caller's complexity mapping.
+func InitComplexityClassifier(modelPath string, useCPU bool) error {
+	var err error
```
Xunzhuo, line 2687: [P1] Cache the initialization error outside the call, as the other classifier initializers do. If the first initialization fails, the next call skips the `sync.Once` closure, creates a fresh nil `err`, and falsely reports success while the Rust classifier is still absent. Add a package-level `complexityClassifierInitErr` and always return it.

src/semantic-router/pkg/config/signal_config.go
```diff
@@ -274,14 +274,39 @@ func HasImageCandidatesInRules(rules []ComplexityRule) bool {
...
 }
 
+// Complexity rule method values.
+const (
+	ComplexityMethodEmbedding = "embedding"
+	ComplexityMethodModel     = "model"
+)
+
+// UsesModel reports whether the rule is decided by the trained complexity classifier.
+func (r ComplexityRule) UsesModel() bool {
```
Xunzhuo, line 296: [P1] Validate `method` as one of empty/`embedding`/`model` during config validation. With the current boolean helper, a typo such as `method: modle` silently falls back to embedding mode; because model-mode examples omit prototype banks, this turns a configuration error into late failure or incorrect routing.

src/semantic-router/pkg/classification/classifier_option_backends.go
```diff
@@ -35,6 +35,40 @@ func (b *classifierOptionBuilder) addMCPCategoryClassifier() {
...
+// addComplexityModelClassifier wires the trained complexity classifier when a model is
+// configured (global.model_catalog.modules.complexity.classifier.model_id) and at least
+// one complexity rule opts into it via method: model. Embedding-mode rules are unaffected.
+func (b *classifierOptionBuilder) addComplexityModelClassifier() error {
+	cfg := b.cfg.ComplexityModel.Classifier
+	hasModelRule := config.HasModelComplexityRule(b.cfg.ComplexityRules)
+
+	if cfg.ModelID == "" {
+		if hasModelRule {
+			logging.ComponentWarnEvent("classifier", "complexity_model_not_configured", map[string]interface{}{
```
Xunzhuo: [P1] Do not accept a model-mode rule without its required classifier configuration. Returning nil leaves the signal permanently inert, so a syntactically accepted route can never match. Reject this during config validation/startup with an actionable error, and likewise require the mapping.

src/semantic-router/pkg/classification/mapping.go
```diff
@@ -91,6 +109,45 @@ func LoadJailbreakMapping(path string) (*JailbreakMapping, error) {
...
+	}
+	if len(mapping.IdxToLabel) == 0 {
+		if len(mapping.IDToLabel) > 0 {
+			mapping.IdxToLabel = mapping.IDToLabel
+		} else if len(mapping.IdxToCategory) > 0 {
+			mapping.IdxToLabel = mapping.IdxToCategory
+		}
+	}
+
+	return &mapping, nil
```
Xunzhuo, line 162: [P1] Validate the normalized mapping before accepting it: it must be non-empty, cover the expected class indices, and map only to the supported lowercase values `easy`, `medium`, and `hard`. Today a mapping containing `HARD` or an arbitrary label loads successfully and emits a signal that the decision schema cannot reference.

src/semantic-router/pkg/config/model_config_types.go
```diff
@@ -175,7 +175,19 @@ func (c PreferenceModelConfig) ContrastiveEnabled() bool {
...
+	Classifier       ComplexityClassifierModelConfig `yaml:"classifier,omitempty"`
+}
+
+// ComplexityClassifierModelConfig configures the trained ModernBERT complexity
+// classifier used by complexity rules with method "model". The model predicts a
+// difficulty class whose index is mapped to a label (e.g. easy/medium/hard) via
+// ComplexityMappingPath. ModelID is a HuggingFace repo id or local path.
+type ComplexityClassifierModelConfig struct {
+	ModelID               string  `yaml:"model_id,omitempty"`
+	Threshold             float32 `yaml:"threshold,omitempty"`
```
Xunzhuo: [P2] This module-level threshold is exposed in docs/config but never read by the runtime; evaluation uses only each rule threshold. Please either define and implement its precedence or remove it so users do not set a no-op control.

Xunzhuo (changes requested):
Thanks for taking this on. The per-rule opt-in shape is useful, but this is not merge-ready yet because model mode cannot be configured consistently through all supported entry points and there are runtime correctness gaps.

Blocking items in addition to the inline comments:

- Update `src/vllm-sr/cli/models.py`: `ComplexityRule` has no `method`, and `hard`/`easy` remain required, so the documented `method: model` example is rejected by CLI validation.
- Update the Operator API/CRDs: `ComplexityRulesConfig` has the same missing `method` and required prototype fields, so model mode cannot be expressed through the CRD. Regenerate and test the manifests.
- Add a behavior-level E2E case that boots the router with a model-mode rule and verifies the emitted `rule:easy|medium|hard` signal. The existing mocked package tests do not exercise config -> model initialization -> API/routing behavior.
- Reconcile the scope with #2008. That issue explicitly asks for domain + complexity heads in one shared mmBERT forward pass; this PR adds a separate single-head classifier and a second inference path. Either implement the shared multi-head architecture, or change `Closes #2008` to an accurate partial-scope reference and track the shared-forward work separately.
- Rebase onto current `main`, fix the markdownlint failure in `website/docs/tutorials/signal/learned/complexity.md`, and rerun the repo-native gates.

Please also update the canonical config documentation required by the config-tree rules, not only the tutorial/reference example.

candle-binding/semantic-router.go
hung1012: Done in https://github.com/vllm-project/semantic-router/pull/2093/changes/d70913f0661801092523a8c07f045bba6909131b. Added a package-level complexityClassifierInitErr, cache it (and reset the Once on failure so a retry can succeed), and return the cached error — so a failed first init no longer reports success on the next call. ClassifyComplexityText now guards on the cached error too, matching InitBertTokenClassifier/ClassifyBertText.

src/semantic-router/pkg/config/signal_config.go
hung1012: Done in https://github.com/vllm-project/semantic-router/pull/2093/commits/9b9cabfcb576bec269f35577d0d1baff16387a61. validateComplexityRules (in validator_complexity.go) now rejects any method outside {"", embedding, model} at config load, so a typo no longer silently falls back to the empty-bank embedding path. Unit tests added for the invalid-method and valid cases.

src/semantic-router/pkg/classification/classifier_option_backends.go
hung1012: Done in [9b9cabfc](https://github.com/vllm-project/semantic-router/pull/2093/commits/9b9cabfcb576bec269f35577d0d1baff16387a61). Moved the requirement into config validation: when any rule uses method: model, model_id and complexity_mapping_path are now required at load time (rejected with an actionable error). addComplexityModelClassifier no longer warns-and-returns-nil — it returns a hard startup error as defense-in-depth, so a model-mode route can never be silently inert

src/semantic-router/pkg/classification/mapping.go
hung1012: Done in [ed7bb67e](https://github.com/vllm-project/semantic-router/pull/2093/commits/ed7bb67e04538cb83b16325c5fc381a145c71518). LoadComplexityMapping now validates the normalized mapping: non-empty, contiguous class indices [0,N), and labels restricted to lowercase easy/medium/hard — so HARD or an arbitrary label is rejected at load instead of emitting an unroutable signal.

src/semantic-router/pkg/config/model_config_types.go
hung1012: Done in [0c2499f5](https://github.com/vllm-project/semantic-router/pull/2093/commits/0c2499f5845ef7e91bbc62254eea4b0dbce18e46). Removed the module-level classifier.threshold (never read at runtime) from the struct, the reference config, and the tutorial. The per-rule threshold — the actual confidence floor — is unchanged.

hung1012:
> Thanks for taking this on. The per-rule opt-in shape is useful, but this is not merge-ready yet because model mode cannot be configured consistently through all supported entry points and there are runtime correctness gaps.
> 
> Blocking items in addition to the inline comments:
> 
> * Update `src/vllm-sr/cli/models.py`: `ComplexityRule` has no `method`, and `hard`/`easy` remain required, so the documented `method: model` example is rejected by CLI validation.
> * Update the Operator API/CRDs: `ComplexityRulesConfig` has the same missing `method` and required prototype fields, so model mode cannot be expressed through the CRD. Regenerate and test the manifests.
> * Add a behavior-level E2E case that boots the router with a model-mode rule and verifies the emitted `rule:easy|medium|hard` signal. The existing mocked package tests do not exercise config -> model initialization -> API/routing behavior.
> * Reconcile the scope with #2008. That issue explicitly asks for domain + complexity heads in one shared mmBERT forward pass; this PR adds a separate single-head classifier and a second inference path. Either implement the shared multi-head architecture, or change `Closes #2008` to an accurate partial-scope reference and track the shared-forward work separately.
> * Rebase onto current `main`, fix the markdownlint failure in `website/docs/tutorials/signal/learned/complexity.md`, and rerun the repo-native gates.
> 
> Please also update the canonical config documentation required by the config-tree rules, not only the tutorial/reference example.

Hi @Xunzhu, could you review my update? I have reviewed your comments and adapted the feature. Hope this PR can be merged soon ^^

Xunzhuo (changes requested):
A fresh pass confirms several earlier blockers were addressed: method/config validation, CLI and Operator parity, threshold removal, and old-head CI. The PR is still not merge-ready.

The complexity initializer remains racy, mapping validation accepts incomplete or duplicate three-class mappings, and the E2E passes when only one case succeeds. More importantly, the implementation adds a process-global second ModernBERT encoder rather than a shared domain/complexity head lifecycle; the referenced shared-head follow-up is now closed. Current dashboard editing also cannot preserve model mode.

Please rebase onto current main, settle the shared-head architecture or restore durable follow-up tracking with explicit acceptance, replace the resettable sync.Once logic with synchronized initialization state, require exactly easy/medium/hard mappings, make every E2E case mandatory, and update the dashboard/config documentation surfaces.

FAUST-BENCHOU (changes requested):
Thanks for the contribution. This is a valuable extension of the complexity signal, and the opt-in model mode keeps the existing embedding path backward compatible.

Before further review, please resolve the current conflicts with main first so we can evaluate the latest integration state.

One technical concern is around resetting `sync.Once` after a failed initialization. `sync.Once` is not designed to be reused after `Do()` has started, so replacing it can introduce races with concurrent callers. Could we handle retry semantics with an explicit mutex/state instead, similar to other classifier initialization paths?

PR #2094, opened 2026-06-09
[DSL] Preserve compile/decompile fields across Keyword, Context, RAG, and Authz

WUKUNTAI-0211:
## Summary

This PR fixes several silent DSL compile/decompile parity gaps where runtime config fields were accepted by one surface but dropped by another. The original scope covered `KeywordRule` and `ContextRule` AST decompile loss; follow-up review also identified RAG and authz field drift, which is now included in this PR.

## Changes

- `KeywordRule`: `DecompileToAST` now preserves `fuzzy_match`, `fuzzy_threshold`, `bm25_threshold`, `ngram_threshold`, and `ngram_arity`.
- `ContextRule`: `DecompileToAST` now preserves `description`.
- RAG plugin: compiler/decompiler paths now preserve `backend_config`, `cache_results`, `cache_ttl_seconds`, and `min_confidence_threshold` where applicable.
- Authz signal: text and AST decompiler paths now preserve `description`.
- Adds focused AST and text round-trip regression tests for the fixed surfaces.

## Test Plan

- `go build ./pkg/dsl/...`
- `go test ./pkg/dsl/... -count=1`
- `make agent-lint CHANGED_FILES="src/semantic-router/pkg/dsl/compiler_plugins.go,src/semantic-router/pkg/dsl/compiler_rag_max_context_length_test.go,src/semantic-router/pkg/dsl/decompiler_convert.go,src/semantic-router/pkg/dsl/decompiler_convert_ast_roundtrip_test.go,src/semantic-router/pkg/dsl/decompiler_plugin_emit.go,src/semantic-router/pkg/dsl/decompiler_plugin_fields.go,src/semantic-router/pkg/dsl/decompiler_plugin_roundtrip_test.go,src/semantic-router/pkg/dsl/decompiler_signals_rules.go"`

CI is rerunning after rebasing this branch onto latest `main`.

## Notes

No schema, grammar, validator, or runtime behavior change. Fields are emitted only when non-zero or non-empty, preserving the existing omit-default convention.

Follow-up to #2086, #2095, and #2101.

Issue #1761, opened 2026-04-10
feature: unify projection contract across Go, DSL, CLI, and dashboard surfaces


Xunzhuo:
## Describe the feature
Unify the projection contract across Go config, DSL compile and decompile, Python CLI schema, and dashboard editing surfaces so richer projection semantics do not drift between authoring and runtime.

## Primary layer
global level

## Why this layer?
Projection is one logical contract, but the repo expresses it across multiple authoring and validation surfaces. This issue is about keeping those surfaces aligned as the projection layer evolves.

## Why do you need this feature?
The projection layer already spans:
- Go runtime config and validators
- DSL parse, compile, decompile, and validator semantics
- Python CLI schema and validation
- dashboard config editing and insights affordances
- docs and maintained recipes

As projection semantics widen, cross-surface drift becomes one of the main risks. A capability that exists only in one layer is operationally expensive and confusing for contributors and users.

### Affected repo surfaces
- `src/semantic-router/pkg/config/{projection_config.go,validator_projection.go}`
- `src/semantic-router/pkg/dsl/{ast.go,parser.go,compiler.go,decompiler.go,validator_conflicts.go,routing_contract.go}`
- `src/vllm-sr/cli/{models.py,config_contract.py,validator.py}`
- `dashboard/frontend/src/pages/{ConfigPageProjectionsSection.tsx,configPageSupport.ts}`
- `website/docs/tutorials/projection/{overview.md,scores.md,mappings.md}`
- maintained recipes such as `deploy/recipes/balance.yaml`

### Suggested starting points
- Start by inventorying the current projection fields and authoring semantics by surface.
- Close at least one concrete mismatch in the same change instead of leaving this issue purely diagnostic.
- Add parity enforcement where the repo can keep it maintained, such as contract tests or maintained-surface checks.

### Validation commands
```bash
[...]

PR #2095, opened 2026-06-09
[DSL] Stop dropping Hallucination and RAG plugin fields on decompile

WUKUNTAI-0211:
## Summary

Three more silent field losses in the DSL plugin decompilers, all in the same class as #2086 and #2094 — the decompiler did not emit fields that the DSL compiler already reads back in. Found via a follow-up audit of compile-side `getXxxField(fields, "...")` reads against decompile-side `fields["..."] = ...` emits across every plugin decoder.

| Plugin | Field(s) silently dropped | AST path | Text DSL path |
|---|---|---|---|
| Hallucination | `use_nli`, `hallucination_action` | ❌ (this PR) | ✅ already emits |
| RAG | `on_failure` | ❌ (this PR) | ❌ (this PR) |

### 1. Hallucination — same shape as #2094

`compileHallucinationPluginConfig` already reads three fields off the DSL:

```go
if v, ok := getBoolField(fields, "enabled"); ok       { cfg.Enabled = v }
if v, ok := getBoolField(fields, "use_nli"); ok       { cfg.UseNLI = v }
if v, ok := getStringField(fields, "hallucination_action"); ok { cfg.HallucinationAction = v }
```

`pluginFieldsHallucination` only emitted `enabled` when surfacing the runtime config back into the AST via `DecompileToAST`. Programmatic consumers (dashboard editor surfaces, projection-contract alignment per #1761, CLI validators) saw a hallucination plugin stripped of its NLI mode and action policy with **no error**. The text decompiler (`emitHallucinationPluginConfig`) already emits all three correctly, so this was a pure AST-path asymmetry — same class as #2094.

### 2. RAG `on_failure` — strictly worse than #2094

`compileRAGPlugin` reads `on_failure`, which has documented `"warn"` / `"error"` semantics validated by `config.validateRAGOnFailure`. **Neither the AST emitter nor the text emitter** surfaced it on decompile:

- [`pluginFieldsRAG`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/dsl/decompiler_plugin_fields.go#L236) — silent AST-path loss.
- [`emitRAGPluginConfig`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/dsl/decompiler_plugin_emit.go#L235) — silent **text-DSL** loss.

Concretely: a `Decompile()` → commit → `Compile()` cycle silently erases a validated routing-failure policy. A no-op YAML re-generation through the DSL contract drops the operator's on-failure intent. Unlike #2094 (AST-only leak), this is the strictly worse "even my text round-trip loses data" variant.

### 3. Aligns with the unification workstream

#1761 ("unify projection contract across Go, DSL, CLI, and dashboard surfaces") explicitly calls out cross-surface drift in DSL parse/compile/decompile as the main risk. This PR removes three concrete drift instances in the plugin decoder space.

## Changes

| File | Change |
|------|--------|
| `src/semantic-router/pkg/dsl/decompiler_plugin_fields.go` | `pluginFieldsHallucination` emits `use_nli` + `hallucination_action`; `pluginFieldsRAG` emits `on_failure`. |
| `src/semantic-router/pkg/dsl/decompiler_plugin_emit.go` | `emitRAGPluginConfig` emits `on_failure: %q`. |
| `src/semantic-router/pkg/dsl/decompiler_plugin_roundtrip_test.go` (new) | 5 focused regression tests + 2 small typed-field assertion helpers. |

Total diff: **3 files, +155 / -0**. Prod-only diff is **+12 lines**; remaining 143 lines are the new test file. All edits mirror the existing non-zero-guard idiom; behaviour for zero / empty fields is unchanged.

## Test Plan

- [x] `go test ./pkg/dsl/... -count=1` — pass (`ok ... 0.707s`)
- [x] `make agent-lint CHANGED_FILES="<the three files>"` — Passed (go fmt, go lint, agent changed-files lint, golangci-lint, `TestReferenceConfig`, **structure checks fully clean** — no warn-only ratchet introduced).
- [x] Targeted: `TestPluginFieldsHallucination*`, `TestPluginFieldsRAG*`, `TestEmitRAGPluginConfigEmitsOnFailure` — all 5 pass.

Full `make test` from `cmd/` is skipped locally because the host is missing the CGO bindings (`candle-binding`, `nlp-binding`, `ml-binding`) — CI matrix will validate them.

## Notes

- Pure decompiler-side fix; no schema, grammar, compiler, or validator change.
- Default behaviour (fields zero / empty) is unchanged: the non-zero guards preserve the omit-default convention used everywhere else in this file.
- Third in a series of DSL round-trip symmetry fixes: #2086 (language threshold), #2094 (keyword + context fields), this PR (hallucination + RAG fields).

Follow-up to #2086 and #2094.

### DIFF
--- a/src/semantic-router/pkg/dsl/decompiler_plugin_emit.go
+++ b/src/semantic-router/pkg/dsl/decompiler_plugin_emit.go
@@ -255,6 +255,9 @@ func emitRAGPluginConfig(sb *strings.Builder, p *config.DecisionPlugin) {
 	if cfg.InjectionMode != "" {
 		fmt.Fprintf(sb, "    injection_mode: %q\n", cfg.InjectionMode)
 	}
+	if cfg.OnFailure != "" {
+		fmt.Fprintf(sb, "    on_failure: %q\n", cfg.OnFailure)
+	}
 }
 
 func emitHeaderMutationPluginConfig(sb *strings.Builder, p *config.DecisionPlugin) {
--- a/src/semantic-router/pkg/dsl/decompiler_plugin_fields.go
+++ b/src/semantic-router/pkg/dsl/decompiler_plugin_fields.go
@@ -115,6 +115,12 @@ func pluginFieldsHallucination(p *config.DecisionPlugin) map[string]Value {
 	if cfg.Enabled {
 		fields["enabled"] = BoolValue{V: true}
 	}
+	if cfg.UseNLI {
+		fields["use_nli"] = BoolValue{V: true}
+	}
+	if cfg.HallucinationAction != "" {
+		fields["hallucination_action"] = StringValue{V: cfg.HallucinationAction}
+	}
 	return fields
 }
 
@@ -256,6 +262,9 @@ func pluginFieldsRAG(p *config.DecisionPlugin) map[string]Value {
 	}
 	if cfg.InjectionMode != "" {
 		fields["injection_mode"] = StringValue{V: cfg.InjectionMode}
+	}
+	if cfg.OnFailure != "" {
+		fields["on_failure"] = StringValue{V: cfg.OnFailure}
 	}
 	return fields
 }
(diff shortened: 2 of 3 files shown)


FAUST-BENCHOU (approved):
LG

Issue #1510, opened 2026-03-12
[Epic] Establish cross-platform performance benchmark coverage and reporting


Xunzhuo:
## Parent Direction

Evaluation & Quality — #2969

## Goal

Establish comprehensive, reproducible performance coverage and reporting across maintained Router layers, runtimes, deployment modes, environments, and hardware.

Regression detection is one consumer of this evidence. The broader goal is to make coverage, baselines, comparisons, reports, and gaps visible enough to guide optimization and release decisions.

## Scope

- Maintain a versioned matrix for request-path processing, Router Model inference, backend serving, and end-to-end behavior.
- Define representative workloads and comparable latency, throughput, memory, startup, utilization, and resource-efficiency metrics.
- Record environment, hardware class, artifact revision, runtime, precision, configuration, warmup, sample count, variance, and limitations.
- Produce human-readable and machine-readable reports with stable provenance and longitudinal comparisons.
- Run appropriate PR, scheduled, hardware-qualified, and release profiles with documented regression policy.

## Child issues

- #3230 — versioned performance coverage and measurement matrix.
- #3231 — reproducible reports and longitudinal baselines.
- #3232 — tiered regression and hardware qualification gates.
- #2348 — CUDA benchmark harness and published CPU/GPU evidence.

## Non-goals

- Claiming unlike hardware, runtimes, or deployment modes are equivalent.
- Replacing correctness, model-quality, security, or reliability evaluation.
- Publishing private hosts, credentials, capacity, or internal infrastructure details.
- Requiring every expensive hardware benchmark on every pull request.

## Completion criteria

- [ ] Every maintained performance path is represented in the coverage matrix or marked as an explicit gap.
- [ ] Results and reports have stable identities, provenance, and comparison rules.
- [ ] Tiered regression and qualification gates use documented thresholds and baseline-refresh rules.
- [ ] Coverage gaps, regressions, and environmental noise have explicit ownership and triage paths.

## Interfaces

- Router Models & Inference Runtime supplies representative runtime and model artifacts.
- Data Plane & Networking identifies request-path boundaries and owns optimizations.
- Enterprise & Environment supplies supported deployment and hardware profiles.
- Evaluation & Quality owns the shared measurement, reporting, comparison, and regression contract.

Xunzhuo:
Maintainer cleanup: moving this large-scale performance-regression tracker out of v0.3. It remains useful future release-gate work, but it is not a Themis closure item.

PR #2102, opened 2026-06-09
[Router] sessiontelemetry: throttle O(n) TTL eviction sweep off the per-turn hot path

### LINKED ISSUE #2100: [Router] sessiontelemetry: O(n) full-store TTL scan under global lock on every RecordTurn
### What

`sessiontelemetry.RecordTurn` runs on the response hot path for **every** completed LLM round-trip. Inside it, `recordTurnState` holds the package-global `mu` and calls `evictLocked` ([`telemetry.go:165`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/sessiontelemetry/telemetry.go#L165)), which iterates the **entire** `store` map to drop entries older than the 1h TTL:

```go
func evictLocked(t time.Time) {
    for k, v := range store {
        if t.Sub(v.lastSeen) > ttl {
            delete(store, k)
        }
    }
}
```

So every request performs an `O(active sessions)` scan while holding the global mutex.

### Impact

With a 1-hour TTL, a busy router accumulates a large session map, and the per-turn cost grows linearly with the number of recent sessions — serializing telemetry recording on the hot path and adding latency under exactly the high-QPS / many-session conditions tracked by #1510.

### Suggested fix

TTL eviction here is a *soft memory reclaim* — an expired entry only costs memory until it is removed — so the full scan does not need to run on every call. Throttle the sweep to at most once per short interval (keeping per-turn work `O(1)` amortized), or move it to a background ticker. Worst-case retention then becomes `ttl + interval`, which is negligible relative to the 1h TTL.

### Note

The sibling `routerSessionMemoryStore.evictExpiredLocked` (`router_memory.go`) has the same per-call full-scan pattern and is also reached from `RecordTurn` — worth a follow-up.

Found via a code audit. I have a PR ready that throttles the sweep.

glitch-ux:
Part of #2100 — note this PR does not fully close it (see Scope below).

## Purpose

- **What:** Throttle the full-store TTL eviction sweep in `sessiontelemetry` so it runs at most once per `evictInterval` (1m) instead of on every `RecordTurn`, plus a per-key O(1) TTL check so a session returning after `ttl` still resets cleanly under the throttle.
- **Why:** `RecordTurn` runs on the response hot path for every completed LLM round-trip. `recordTurnState` held the package-global `mu` and called `evictLocked`, which scanned the **entire** `store` map. That is an `O(active sessions)` sweep under the global lock on every request; with the 1h TTL a busy router accumulates a large session map, so per-turn telemetry cost grows linearly with recent-session count and serializes recording on the hot path (cf. #1510).
- **How:** TTL eviction is a *soft memory reclaim* (an expired entry only costs memory until removed), so the full scan need not run every call. A `lastEvict` timestamp gates the sweep: if less than `evictInterval` has elapsed since the last sweep, `evictLocked` returns immediately, keeping the per-turn path `O(1)` amortized. Worst-case retention becomes `ttl + evictInterval` (≈1h1m), negligible vs the 1h TTL. Because the full sweep is now throttled, an expired entry for a returning key can still be present when that session records again, so `recordTurnState` applies a per-key O(1) TTL check before reusing `turnState` — cumulative tokens/cost reset at `ttl` exactly as before instead of being soft-retained. No background goroutine is introduced, so shutdown/lifecycle is unchanged and the behavior stays deterministic and testable.
- **Module:** `Router`.

### Scope / relationship to #2100

This PR throttles **only** the `telemetry.go` `store` sweep. The sibling `routerSessionMemoryStore.evictExpiredLocked` in `router_memory.go` has the same per-call full-scan pattern and is **also** reached from `RecordTurn` (via `RecordSessionUsage`); it remains unthrottled here. So this change does not fully resolve #2100 — the issue stays open (not auto-closed) with the `router_memory.go` sweep tracked there as the remaining follow-up.

## Test Plan

```bash
cd src/semantic-router
gofmt -l pkg/sessiontelemetry/                 # formatting
go vet ./pkg/sessiontelemetry/                 # static analysis
go test ./pkg/sessiontelemetry/ -race -count=1 # full package, race detector
```

- `TestEvictionScanThrottled` plants an already-stale entry just after a sweep and asserts it **survives** a sub-`evictInterval` call (scan skipped) and is **removed** once `evictInterval` has elapsed (scan runs) — directly verifying the throttle.
- `TestExpiredKeyResetsBeforeThrottledSweep` pins `lastEvict` so the full sweep is skipped, then has a session return after `ttl` and asserts its cumulative state **resets** rather than accumulates — verifying the per-key TTL check.
- Existing `TestEviction` is unchanged and still passes (its `ttl + 1m` time jump exceeds `evictInterval`, so the necessary sweep still runs).

## Test Result

```
=== RUN   TestEviction
--- PASS: TestEviction (0.00s)
=== RUN   TestEvictionScanThrottled
--- PASS: TestEvictionScanThrottled (0.00s)
=== RUN   TestExpiredKeyResetsBeforeThrottledSweep
--- PASS: TestExpiredKeyResetsBeforeThrottledSweep (0.00s)
PASS
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/sessiontelemetry  1.083s  (-race)
```

`gofmt -l` reports no diffs; `go vet` is clean. The only known related gap (the `router_memory.go` sibling scan) is tracked in #2100.

---
<details>

</details>


drivebyer:
This PR should not close #2100 as-is. Please update the description to say it only throttles the telemetry.go store sweep, while router_memory.go still has an unthrottled full scan on the RecordTurn path. 

Also note the throttled cleanup may soft-retain expired cumulative state unless a per-key O(1) TTL check is added before reusing turnState, like:
```
if st != nil && t.Sub(st.lastSeen) > ttl {
        delete(store, key)
        st = nil
}
```

glitch-ux:
Done — added the per-key TTL reset (with a test) and reworded to "Part of #2100" since `router_memory.go`'s scan is still unthrottled. Thanks @drivebyer!

drivebyer (approved):

PR #2104, opened 2026-06-09
[Router] Accept image input on /api/v1/classify/intent, /api/v1/eval, and /api/v1/embeddings

### LINKED ISSUE #1911: feature: extend HTTP API to accept image input for embedding-related endpoints
### Describe the feature

The runtime supports image-modality embedding operations end-to-end:

- Candle-binding FFI for image encoding (PR #1414, `MultiModalEncodeImageFromBase64`).
- Request-path image extractor + signal-evaluator wiring (PRs #1867, #1868).
- CRD-side `queryModality` field on `IntelligentRoute` embedding rules (PR #1880); reconcile-time validation in #1895.

Image-modality embedding rules fire correctly when a chat completion arrives via the gRPC ExtProc path with an OpenAI-shaped `image_url` content array.

However, the HTTP API surface for direct embedding-related operations doesn't accept image input on any endpoint. Every embedding-adjacent handler in `pkg/apiserver/` accepts text-only request schemas, even though the underlying service methods and the multimodal FFI could compute image embeddings if given them.

**Scope of the gap** (file:line citations against `main` as of 2026-05-15):

| Endpoint | Handler | Request type | Image-capable today? |
|---|---|---|---|
| `POST /api/v1/classify/intent` | `route_classify.go:18` (`handleIntentClassification`) | `services.IntentRequest{Text, Messages, Options}` (`pkg/services/classification_signal_contract.go:34`) | No |
| `POST /api/v1/classify/batch` | `route_classify.go:96` (`handleBatchClassification`) | `BatchClassificationRequest{Texts, TaskType}` (`pkg/apiserver/config.go:49`) | No |
| `POST /api/v1/eval` | `route_classify.go:38` (`handleEvalClassification`) | reuses `IntentRequest` | No |
| `POST /api/v1/embeddings` | `route_embeddings.go:14` (`handleEmbeddings`) | `EmbeddingRequest{Texts: []string, Model}` (`pkg/apiserver/config.go:86`) | No |
| `POST /api/v1/similarity` | `route_embeddings.go:131` (`handleSimilarity`) | `SimilarityRequest{Text1, Text2, Model}` (`pkg/apiserver/config.go:114`) | No |
| `POST /api/v1/similarity/batch` | `route_embeddings.go:190` (`handleBatchSimilarity`) | `BatchSimilarityRequest{Query, Candidates, TopK}` (`pkg/apiserver/config.go:132`) | No |

The runtime evaluator gates the image-modality path on `imageURL != ""` (`pkg/classification/classifier_signal_context.go:182`). Because none of these HTTP request schemas surface an image field, that gate never opens from the HTTP path, and image-modality embedding rules never fire from any of these endpoints regardless of what rules ship in the config. (`/api/v1/eval` reuses `IntentRequest` verbatim, so one extension closes both endpoints in a single change.)

### Primary layer

`global level`

### Why this layer?

The signal layer's image-modality plumbing already exists, the runtime FFI exists, and the gap is at the HTTP entry point that fans into both. Extending the HTTP request types is a request-API change rather than a signal-layer feature, which puts it in the "intentionally cross-cutting behavior" bucket the template describes for `global level`. If maintainers prefer `signal` because the motivation is unblocking image-modality embedding signals end-to-end, the re-tagging is fine; the engineering work is unchanged.

1. **Author/operator validation of image-modality embedding rules.** A pack like `config/signal/embedding/image-routing.yaml` (shipped in #1896) defines three image-modality rules. Confirming those rules fire on representative images today requires either (a) standing up a full Envoy + ExtProc + backend chain to send chat completions, or (b) writing a custom gRPC ExtProc client. Both are heavier than running `curl` against `/api/v1/classify/intent`.

2. **Computing the embedding vector of an image for downstream use** (indexing, storage, retrieval). The multimodal model is loaded; the FFI supports it; the HTTP API doesn't expose it.

3. **Cross-modal similarity** ("which of these phrases is most similar to this image?"). Common in vision-language workflows; the runtime supports it via `ClassifyDetailedMultimodal`; no HTTP surface exposes it.

4. **Image-to-image similarity.** Same shape as above between two images.

### Additional context

**Proposed shape (aligned with the codebase's existing image-content convention):**

The runtime's existing image accept set is documented at `pkg/extproc/utils_fast.go:182-200`: inline `data:image/...;base64,...` URIs only, no http/https URLs (intentional, the ExtProc path closes an SSRF-class concern there). The new HTTP fields should match that accept set. A `string` field carrying the data URI is the lightest option; an object-typed `Image { URL string }` mirroring OpenAI Chat Completions is also defensible. The shape below uses the string form; happy to switch to the typed object if maintainers prefer it for cross-product tooling alignment.

`IntentRequest` (covers `/api/v1/classify/intent` and `/api/v1/eval`):

```go
type IntentRequest struct {
    Text     string          `json:"text"`
    Messages []IntentMessage `json:"messages,omitempty"`
    Image    string          `json:"image,omitempty"`     // NEW: data:image/...;base64,... URI
    Options  *IntentOptions  `json:"options,omitempty"`
}
```

`ClassifyIntent` populates the `imageURL` argument that `EvaluateAllSignalsWithContext` already takes; nothing downstream changes.

`BatchClassificationRequest`: add a parallel `Images []string` field alongside `Texts`. Exactly one of `Texts` / `Images` set per request in v1; mixed batches are deferred.

`EmbeddingRequest`: same shape, add `Images []string` parallel to `Texts`.

`SimilarityRequest`: generalize to `{Text1, Text2, Image1, Image2}` with exactly-one-of `{text, image}` per side. Enables text-text (existing), image-image, and cross-modal text-image similarity.

`BatchSimilarityRequest` (`/api/v1/similarity/batch`): its shape is `{Query, Candidates []string, TopK}` (top-k retrieval). Generalize `Query` to accept text OR image, add a sibling `CandidateImages []string` field, with the same exactly-one-of constraint on the corpus side. Mixed text+image candidates in a single batch are deferred.

**Open question (please steer):**

Three plausible shapes:

1. **Additive** (drafted above): extend existing request types with optional image fields. Smallest diff. Mixes concerns inside each request type but each addition is narrow.
2. **Sibling endpoints**: keep existing endpoints text-only, add `/api/v1/classify/multimodal-intent`, `/api/v1/classify/multimodal-batch`, `/api/v1/embeddings/multimodal`, `/api/v1/similarity/multimodal`, `/api/v1/similarity/batch/multimodal`. Cleaner separation; more endpoints to discover; doubles route registration.
3. **Typed-union request body on a new sibling endpoint set** (`InputA`, `InputB` where each is `oneof {Text, Image}`): cleanest semantics; largest single-PR diff; sets a convention that doesn't match the rest of the apiserver today.

The additive shape is the smallest delta from today's surface. Happy to redo the draft in either of the others if maintainers prefer.

**Staged delivery (if maintainers prefer focused PRs):**

1. `/api/v1/classify/intent` + `/api/v1/eval` (one PR; same request type) - immediately unblocks fixture-based testing for #1896.
2. `/api/v1/classify/batch` - same plumbing, batched form.
3. `/api/v1/embeddings` - enables image-embedding extraction for downstream pipelines.
4. `/api/v1/similarity*` (both pairwise and batch) - enables cross-modal similarity.

Each step is independently shippable behind the next.

**Out of scope for this issue:**

- **Audio modality.** `MultiModalEncodeAudio` is exposed at `candle-binding/semantic-router.go:1106` (takes a pre-computed Mel spectrogram), but the byte-stream variants (`FromBytes` / `FromBase64` / `FromURL`) that would let the HTTP API accept inline audio are not yet exposed. The existing validator already rejects audio rules at config-load for this reason (`pkg/config/validator_embedding.go:64-67`); a separate issue can track exposing the byte-stream variants if there's demand.
- **Remote (http/https) image URLs.** The runtime explicitly rejects http URLs in the ExtProc image path (`pkg/extproc/utils_fast.go:183`: *"Only inline data URIs are accepted (no HTTP URLs)"*); the HTTP API should match. If remote-URL fetching becomes desirable later, it warrants its own design conversation (allow-lists, size caps, content-type sniffing) separate from this gap.
- **Multi-image batching efficiency.** The first version can iterate per-image. Batched FFI calls are a perf optimization, not a correctness requirement.

**Motivating PR:** #1896 ships an opt-in image-modality embedding pack at `config/signal/embedding/image-routing.yaml`. Its "What's NOT in this PR" section names this gap on a single endpoint (`/api/v1/classify/intent`) and explicitly defers a follow-on issue to propose the shape - this is that follow-on, scoped across the full embedding-related HTTP surface (6 endpoints once `/api/v1/classify/batch` is included) rather than just one, because the gap is structural.

WUKUNTAI-0211:
## What

Closes #1911 (Phase 1 + embeddings).

The runtime already supports image-modality embeddings end-to-end (FFI `MultiModalEncodeImageFromBase64`, request-path image extractor, signal evaluator), but the HTTP API surface only accepted text. Image-modality embedding rules could therefore only fire from the gRPC ExtProc chat-completion path, never from the HTTP classification/embeddings endpoints. This PR extends the HTTP request schemas so they do.

Scope is intentionally limited to the issue's primary motivation (validating image-modality rules) plus direct image embedding. The batch-classify and similarity endpoints are deferred and tracked in #2317.

## Changes

**Shared image-input gate** (`pkg/utils/imageurl`)
- Extracts `isSafeImageDataURL` out of the ExtProc package into a reusable `IsSafeImageDataURL` helper, so the HTTP API and ExtProc enforce one identical gate (inline `data:image/...;base64,` only — HTTP(S) URLs and non-image data URIs rejected to prevent SSRF). ExtProc now delegates; behavior unchanged.
- Case-insensitive scheme/marker parsing with request-time base64 validation (`DecodeBase64`) and a canonical lowercase-marker rebuild (`CanonicalDataURL`) for the FFI's case-sensitive `;base64,` scan. Per-request image cap (`maxImagesPerRequest`).

**`/api/v1/classify/intent` and `/api/v1/eval`**
- Parse OpenAI-shaped `image_url` content parts from the current user turn and thread the inline data URI into `EvaluateAllSignalsWithContext` (now a **required** `imageURL` parameter — dropping it at a call site no longer compiles, so the request→classifier image wiring is type-defended).
- `intentMessageImageURL.UnmarshalJSON` accepts both the Chat Completions object form `{"url": "..."}` and the Responses API bare-string form; an unexpected shape (number/array/bool) degrades to no-image rather than poisoning the whole content-parts unmarshal.
- Image-only user turns no longer clobber or rotate out the most recent user text; evaluation/context/current-user text fall back to top-level `req.Text` when the messages path is accepted solely because of an image. Image safety can no longer toggle which text gets scored.
- `/api/v1/eval` reuses `IntentRequest`, so both endpoints are covered by one change.

**`/api/v1/embeddings`**
- Adds an optional `images` field (inline base64 data URIs) and computes their embeddings via the multi-modal model.
- Requests now require at least one of `texts` or `images`; malformed base64 / unsafe image references return `INVALID_IMAGE` (**400**) at request time instead of reaching the FFI as a 500; the average-processing-time calculation divides by request-input count (`len(texts)+len(images)`, zero-guarded) so image-only requests no longer skew it.
- Image results carry a `modality` field; text results are unchanged (backward compatible).

## Test Plan

- New unit tests:
  - `pkg/utils/imageurl`: data-URI gate (allowed MIME types, HTTP rejection, empty payload), `DecodeBase64`, `CanonicalDataURL`.
  - `pkg/services`: image extraction from current user turn, image-only turn accepted, unsafe URL dropped, plus the fix-commit repros — `TestIntentRequestResolveSignalInput_StringImageURLDoesNotPoisonText`, `..._MalformedImageURLDoesNotPoisonText`, `ImageOnlyTurnFallsBackToTopLevelText`, `ImageOnlyFollowUpKeepsPriorUserText`, `AcceptsImageOnlyUserTurn`.
  - `pkg/apiserver`: embeddings validation requires texts-or-images, accepts images-only, `RejectsMalformedBase64` (400), `AcceptsUppercaseDataURIScheme`, `RejectsTooManyImages`.

## Test Result

- `go test ./pkg/services/... ./pkg/apiserver/... ./pkg/utils/imageurl/... ./pkg/classification/... ./pkg/extproc/...` — all pass on the merged tree.
- `golangci-lint run` (agent config) on the changed packages — 0 issues.
- CI green; branch merges clean over current `main`.

> Note: the FFI image-encode path (`MultiModalEncodeImageFromBase64`) is env-gated behind `MULTIMODAL_MODEL_PATH` and was validated manually against a real SigLIP checkpoint (uppercase-URI repro encodes end-to-end on `/api/v1/embeddings`); it is not exercised by the unit suite. Building/testing the cgo packages required rebuilding the Candle Rust library (`make rust`) — the multi-modal image FFI symbol was missing from the previously cached `target/release` artifact.

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_embeddings.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings.go
@@ -17,6 +17,9 @@ import (
 const (
 	defaultEmbeddingDimension = 768
 	defaultEmbeddingPriority  = 0.5
+	// maxImagesPerRequest bounds images per request; each is a full SigLIP
+	// forward pass and the body-size cap alone admits very many minimal images.
+	maxImagesPerRequest = 8
 )
 
 // handleEmbeddings handles embedding generation requests
@@ -93,10 +96,8 @@ func validateEmbeddingRequest(req EmbeddingRequest, mmbertLayers []int) (string,
 	if len(req.Texts) == 0 && len(req.Images) == 0 {
 		return "INVALID_INPUT", "at least one of texts or images must be provided", false
 	}
-	for i, image := range req.Images {
-		if !imageurl.IsSafeImageDataURL(image) {
-			return "INVALID_IMAGE", fmt.Sprintf("images[%d] must be an inline base64 image data URI (data:image/<type>;base64,...)", i), false
-		}
+	if code, message, ok := validateEmbeddingImages(req.Images); !ok {
+		return code, message, false
 	}
 	if !isValidDimension(req.Dimension) {
 		return "INVALID_DIMENSION", fmt.Sprintf("dimension must be one of: 64, 128, 256, 512, 768, 1024 (got %d)", req.Dimension), false
@@ -106,6 +107,23 @@ func validateEmbeddingRequest(req EmbeddingRequest, mmbertLayers []int) (string,
 	}
 	if req.Model == "mmbert" && req.TargetLayer != 0 && !config.IsValidMmBertLayer(req.TargetLayer, mmbertLayers) {
 		return "INVALID_LAYER", fmt.Sprintf("target_layer must be one of: %s (got %d)", formatLayerList(mmbertLayers), req.TargetLayer), false
+	}
+	return "", "", true
+}
+
+// validateEmbeddingImages enforces the image-input contract: a bounded count of
+// safe inline base64 image data URIs whose payloads decode.
+func validateEmbeddingImages(images []string) (string, string, bool) {
+	if len(images) > maxImagesPerRequest {
+		return "INVALID_INPUT", fmt.Sprintf("at most %d images may be provided per request (got %d)", maxImagesPerRequest, len(images)), false
+	}
+	for i, image := range images {
+		if !imageurl.IsSafeImageDataURL(image) {
+			return "INVALID_IMAGE", fmt.Sprintf("images[%d] must be an inline base64 image data URI (data:image/<type>;base64,...)", i), false
+		}
+		if _, ok := imageurl.DecodeBase64(image); !ok {
+			return "INVALID_IMAGE", fmt.Sprintf("images[%d] is not valid base64-encoded image data", i), false
+		}
 	}
 	return "", "", true
 }
@@ -133,7 +151,13 @@ func buildEmbeddingResults(req EmbeddingRequest) ([]EmbeddingResult, int64, erro
 	}
 
 	for _, image := range req.Images {
-		output, err := candle_binding.MultiModalEncodeImageFromBase64(image, req.Dimension)
+		// Canonicalize so the FFI's case-sensitive ";base64," scan finds the
+		// payload boundary (validation already guaranteed a safe data URI).
+		encodeInput := image
+		if canonical, ok := imageurl.CanonicalDataURL(image); ok {
+			encodeInput = canonical
+		}
+		output, err := candle_binding.MultiModalEncodeImageFromBase64(encodeInput, req.Dimension)
 		if err != nil {
 			return nil, 0, err
 		}
--- a/src/semantic-router/pkg/apiserver/route_embeddings.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings.go
@@ -11,6 +11,7 @@ import (
 	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/utils/imageurl"
 )
 
 const (
@@ -32,7 +33,10 @@ func (s *ClassificationAPIServer) handleEmbeddings(w http.ResponseWriter, r *htt
 		return
 	}
 
-	avgProcessingTime := float64(totalProcessingTime) / float64(len(req.Texts))
+	avgProcessingTime := 0.0
+	if len(results) > 0 {
+		avgProcessingTime = float64(totalProcessingTime) / float64(len(results))
+	}
 	response := EmbeddingResponse{
 		Embeddings:            results,
 		TotalCount:            len(results),
@@ -81,8 +85,13 @@ func applyEmbeddingDefaults(req *EmbeddingRequest) {
 }
 
 func validateEmbeddingRequest(req EmbeddingRequest, mmbertLayers []int) (string, string, bool) {
-	if len(req.Texts) == 0 {
-		return "INVALID_INPUT", "texts array cannot be empty", false
+	if len(req.Texts) == 0 && len(req.Images) == 0 {
+		return "INVALID_INPUT", "at least one of texts or images must be provided", false
+	}
+	for i, image := range req.Images {
+		if !imageurl.IsSafeImageDataURL(image) {
+			return "INVALID_IMAGE", fmt.Sprintf("images[%d] must be an inline base64 image data URI (data:image/<type>;base64,...)", i), false
+		}
 	}
 	if !isValidDimension(req.Dimension) {
 		return "INVALID_DIMENSION", fmt.Sprintf("dimension must be one of: 64, 128, 256, 512, 768, 1024 (got %d)", req.Dimension), false
@@ -97,7 +106,7 @@ func validateEmbeddingRequest(req EmbeddingRequest, mmbertLayers []int) (string,
 }
 
 func buildEmbeddingResults(req EmbeddingRequest) ([]EmbeddingResult, int64, error) {
-	results := make([]EmbeddingResult, 0, len(req.Texts))
+	results := make([]EmbeddingResult, 0, len(req.Texts)+len(req.Images))
 	var totalProcessingTime int64
 
 	for _, text := range req.Texts {
@@ -112,6 +121,24 @@ func buildEmbeddingResults(req EmbeddingRequest) ([]EmbeddingResult, int64, erro
 			Embedding:        output.Embedding,
 			Dimension:        len(output.Embedding),
 			ModelUsed:        output.ModelType,
+			ProcessingTimeMs: processingTime,
+		})
+
+		totalProcessingTime += processingTime
+	}
+
+	for _, image := range req.Images {
+		output, err := candle_binding.MultiModalEncodeImageFromBase64(image, req.Dimension)
+		if err != nil {
+			return nil, 0, err
+		}
+
+		processingTime := int64(output.ProcessingTimeMs)
+		results = append(results, EmbeddingResult{
+			Modality:         output.Modality,
+			Embedding:        output.Embedding,
+			Dimension:        len(output.Embedding),
+			ModelUsed:        "multi-modal-embed",
 			ProcessingTimeMs: processingTime,
 		})
 
--- a/src/semantic-router/pkg/apiserver/route_embeddings.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings.go
@@ -33,10 +33,7 @@ func (s *ClassificationAPIServer) handleEmbeddings(w http.ResponseWriter, r *htt
 		return
 	}
 
-	avgProcessingTime := 0.0
-	if len(results) > 0 {
-		avgProcessingTime = float64(totalProcessingTime) / float64(len(results))
-	}
+	avgProcessingTime := averageEmbeddingProcessingTime(totalProcessingTime, req)
 	response := EmbeddingResponse{
 		Embeddings:            results,
 		TotalCount:            len(results),
@@ -69,6 +66,14 @@ func (s *ClassificationAPIServer) parseEmbeddingRequest(w http.ResponseWriter, r
 	}
 
 	return req, true
+}
+
+func averageEmbeddingProcessingTime(totalProcessingTime int64, req EmbeddingRequest) float64 {
+	inputCount := len(req.Texts) + len(req.Images)
+	if inputCount == 0 {
+		return 0
+	}
+	return float64(totalProcessingTime) / float64(inputCount)
 }
 
 func applyEmbeddingDefaults(req *EmbeddingRequest) {
--- a/src/semantic-router/pkg/apiserver/config.go
+++ b/src/semantic-router/pkg/apiserver/config.go
@@ -87,6 +87,7 @@ type ClassificationOptions struct {
 // EmbeddingRequest represents a request for embedding generation
 type EmbeddingRequest struct {
 	Texts           []string `json:"texts"`
+	Images          []string `json:"images,omitempty"`           // Inline base64 image data URIs (data:image/...;base64,...); encoded via the multi-modal model
 	Model           string   `json:"model,omitempty"`            // "auto" (default), "qwen3", "gemma", "mmbert"
 	Dimension       int      `json:"dimension,omitempty"`        // Target dimension: 768 (default), 512, 256, 128, 64
 	TargetLayer     int      `json:"target_layer,omitempty"`     // Target layer for early exit (mmbert only): 3, 6, 11, 22 (0=full)
@@ -98,6 +99,7 @@ type EmbeddingRequest struct {
 // EmbeddingResult represents a single embedding result
 type EmbeddingResult struct {
 	Text             string    `json:"text"`
+	Modality         string    `json:"modality,omitempty"` // "image" for image inputs; empty for text (backward compatible)
 	Embedding        []float32 `json:"embedding"`
 	Dimension        int       `json:"dimension"`
 	ModelUsed        string    `json:"model_used"`
--- a/src/semantic-router/pkg/apiserver/route_embeddings_test.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings_test.go
@@ -39,6 +39,44 @@ func TestBuildBatchSimilarityMatchesIncludesCandidateText(t *testing.T) {
 	}
 }
 
+func TestValidateEmbeddingRequestRequiresTextsOrImages(t *testing.T) {
+	req := EmbeddingRequest{Dimension: defaultEmbeddingDimension}
+
+	code, message, ok := validateEmbeddingRequest(req)
+	if ok {
+		t.Fatalf("expected empty texts and images to be invalid")
+	}
+	if code != "INVALID_INPUT" || message != "at least one of texts or images must be provided" {
+		t.Fatalf("unexpected validation error %q: %q", code, message)
+	}
+}
+
+func TestValidateEmbeddingRequestAcceptsImagesOnly(t *testing.T) {
+	req := EmbeddingRequest{
+		Images:    []string{"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAAB"},
+		Dimension: defaultEmbeddingDimension,
+	}
+
+	if _, _, ok := validateEmbeddingRequest(req); !ok {
+		t.Fatalf("expected image-only request to be valid")
+	}
+}
+
+func TestValidateEmbeddingRequestRejectsUnsafeImage(t *testing.T) {
+	req := EmbeddingRequest{
+		Images:    []string{"https://example.com/cat.png"},
+		Dimension: defaultEmbeddingDimension,
+	}
+
+	code, message, ok := validateEmbeddingRequest(req)
+	if ok {
+		t.Fatalf("expected non-data-URI image to be rejected (SSRF guard)")
+	}
+	if code != "INVALID_IMAGE" {
+		t.Fatalf("unexpected validation error code %q: %q", code, message)
+	}
+}
+
 func TestNormalizeBatchSimilarityLimitCapsTopKAtCandidateCount(t *testing.T) {
 	req := BatchSimilarityRequest{
 		Candidates: []string{"a", "b"},
--- a/src/semantic-router/pkg/apiserver/route_embeddings_test.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings_test.go
@@ -42,7 +42,7 @@ func TestBuildBatchSimilarityMatchesIncludesCandidateText(t *testing.T) {
 func TestValidateEmbeddingRequestRequiresTextsOrImages(t *testing.T) {
 	req := EmbeddingRequest{Dimension: defaultEmbeddingDimension}
 
-	code, message, ok := validateEmbeddingRequest(req)
+	code, message, ok := validateEmbeddingRequest(req, nil)
 	if ok {
 		t.Fatalf("expected empty texts and images to be invalid")
 	}
@@ -57,7 +57,7 @@ func TestValidateEmbeddingRequestAcceptsImagesOnly(t *testing.T) {
 		Dimension: defaultEmbeddingDimension,
 	}
 
-	if _, _, ok := validateEmbeddingRequest(req); !ok {
+	if _, _, ok := validateEmbeddingRequest(req, nil); !ok {
 		t.Fatalf("expected image-only request to be valid")
 	}
 }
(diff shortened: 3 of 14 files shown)


src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -29,7 +30,10 @@ func (s *ClassificationAPIServer) handleEmbeddings(w http.ResponseWriter, r *htt
 		return
 	}
 
-	avgProcessingTime := float64(totalProcessingTime) / float64(len(req.Texts))
+	avgProcessingTime := 0.0
+	if len(results) > 0 {
+		avgProcessingTime = float64(totalProcessingTime) / float64(len(results))
```
Peterren: before this change, your processing time = totalTime / len of request, after this change,  your processing time = totalTime / len of result. Is this intentional? Is it because http request passes img so it does not have req.Text field? do we have `Content-Length` from request or its not presented because its chunked transfer encoding?

WUKUNTAI-0211: Yes, this was intentional to avoid dividing by `len(req.Texts)` for image-only embedding requests, where `texts` can be empty.

I updated this to use an explicit input count (`len(req.Texts) + len(req.Images)`) instead of `len(results)`, so the denominator stays request-input based while including image inputs. In the current success path that is equivalent to the result count because each accepted text/image input produces one embedding, but this makes the semantics clearer.

`Content-Length` is not the right denominator here because `avg_processing_time_ms` is per embedding input item, not per request byte, and it may be absent for chunked requests.

WUKUNTAI-0211:
@Peterren thanks for catching the denominator — addressed. The per-input average is now computed in a dedicated `averageEmbeddingProcessingTime(totalProcessingTime, req)` helper that divides by `len(req.Texts) + len(req.Images)` (the request input count, guarded against a zero divisor), so image-only requests no longer skew it and the metric stays request-input based rather than `len(results)`. `TotalCount` still reports `len(results)` since that's the count actually returned. Mind taking another look when you have a moment?

shraderdm (changes requested):
Thanks for picking up #1911 (image-input API), @WUKUNTAI-0211 - this issue is the API-surface foundation under the whole multi-modal arc, and this is a genuinely well-built first slice of it. The findings below are almost all integration-seam issues around the diff rather than problems inside it; the one Critical comes with a verified fix you can take as-is. Happy to re-review fast once it's addressed.

### What's good

- The gate extraction is exactly right: `pkg/utils/imageurl/imageurl.go` is character-for-character the removed ExtProc implementation, the private wrapper at `pkg/extproc/utils.go:188-192` means zero call-site churn, and the package placement matches the sibling utils convention (`entropy`/`http`/`pii`/`tls`/`valkey`).
- The new tests falsify, not just illustrate. I reverted the acceptance change at `intent_request_messages.go:94` and `TestIntentRequestResolveSignalInput_AcceptsImageOnlyUserTurn` fails; reverted the embeddings validation and `TestValidateEmbeddingRequestAcceptsImagesOnly` + `TestValidateEmbeddingRequestRejectsUnsafeImage` fail for the right reasons. They also actually run in PR CI (test-and-build builds the Rust lib, then runs the Go tests with cgo).
- SSRF-conscious throughout: data-URI-only gating at every new surface, and `MultiModalEncodeImageFromURL` stays unreachable from user input.
- Clean, bisectable commits (extraction first; each builds independently), and the review-response commit also fixed the latent divide-by-zero for image-only requests.
- It composes cleanly with current main: I merged this branch over the `ErrEmptyText` refactor (`50edfb42`) and the full suite is green on the merged tree.

### Findings

#### Critical (blocking)

- **A string-valued `image_url` content part now poisons text extraction for the whole message, so previously-classifiable requests fail** (`pkg/services/intent_request_messages.go:38-42`). Adding `ImageURL *intentMessageImageURL` to the shared `intentMessageContentPart` makes `json.Unmarshal` of a parts array fail wholesale when `image_url` is a string - the OpenAI Responses shape, e.g. `{"type":"input_image","image_url":"data:..."}` - and the single-object fallback fails the same way, so both text and image extraction return empty. That shape is a supported input here: `joinIntentMessageContentParts` accepts `input_text` (`:197`). Executed red/green: on the merge base, `[{"type":"input_text","text":"hello world"},{"type":"input_image","image_url":"https://example.com/cat.png"}]` classifies with `evaluationText="hello world"`; at this branch's tip the same request fails with "text cannot be empty". The data-URI string variant fails identically, so Responses-style clients also cannot reach the new image feature. This fix is verified (compiles, restores base behavior on both variants, keeps every new test in this PR green):

  ```go
  type intentMessageImageURL struct {
  	URL string `json:"url"`
  }

  // UnmarshalJSON accepts both the Chat Completions object form
  // {"url": "..."} and the Responses API string form "..." so that a
  // string-valued image_url part cannot fail the whole content-parts
  // unmarshal (which would poison text extraction for the message).
  func (u *intentMessageImageURL) UnmarshalJSON(data []byte) error {
  	var s string
  	if err := json.Unmarshal(data, &s); err == nil {
  		u.URL = s
  		return nil
  	}
  	type plain intentMessageImageURL
  	var p plain
  	if err := json.Unmarshal(data, &p); err != nil {
  		return err
  	}
  	*u = intentMessageImageURL(p)
  	return nil
  }
  ```

#### Important (should address before merge)

- **When a message carries an image but no text, the messages path claims the request with an empty evaluation text, silently changing which text the signals score** (`pkg/services/intent_request_messages.go:94`). `evaluationText` is what every text signal (PII, jailbreak, keyword, domain) scores (`pkg/services/classification.go:141`), and the widened acceptance admits inputs the fallback ladder at `:44-59` and `:77-80` was not designed for. Two executed red/green repros against the merge base, at the shared input resolution both `/api/v1/classify/intent` and `/api/v1/eval` consume: (1) top-level `text` plus a safe-image-only message - the base scores "summarize my quarterly report", this branch scores `""` and silently discards the supplied text; with an UNSAFE image URL the messages path declines and the text is honored again, so image safety (which the client controls) toggles whether the supplied text is evaluated at all. (2) `[user "What is our refund policy?", assistant "Refunds are processed within 30 days.", user image-only]` - the base scores the user question; this branch rotates it into `priorUserMessages` and the `:77-80` fallback promotes the ASSISTANT text into `evaluationText`. Suggested direction: when the messages path is accepted only because of an image (empty evaluation text), fall back to `req.Text` for evaluation/context text while keeping the image, and guard the `:77` non-user fallback so image-driven acceptance cannot promote assistant/system text.

- **Image embeddings silently ignore the requested dimension, so a default request returns mixed 768-dim and 384-dim vectors in one 200 response** (`candle-binding/src/model_architectures/embedding/multimodal_embedding.rs:1197-1203`). `encode_image_with_dim` truncates only when the target is below the native 384 and otherwise silently returns native - unlike the text 2DMSE path, which rejects `target_dim > embedding_dim` with a validation error (`:1127-1134`). Executed against a real checkpoint: `dimension` 512/768/1024 all return 384-dim image embeddings with no error (256 truncates correctly), and the handler defaults `dimension` to 768, so the DEFAULT mixed request hits this. Meanwhile `isValidDimension` (`pkg/apiserver/route_embeddings.go:332-336`) excludes 384, so the image model's native dimension is the one value the API will not accept. Related parameter gaps on the same path: `req.Model` and `target_layer` are silently ignored for images, and `model_used` is hardcoded to "multi-modal-embed" rather than the loaded checkpoint's id. Suggested direction: give the image path the same over-dimension contract the text paths already have - either reject above-native requests like the 2DMSE path does, or clamp-and-warn once per batch the way #2145 (FFI truncation renorm + batch clamp) established for the FFI text paths - or document the truncate-only contract and accept 384; and surface the real model id in `model_used`.

- **Inputs the new validator explicitly accepts can fail at the FFI and surface as a client-input-driven 500** (`pkg/utils/imageurl/imageurl.go:16-18`). The gate lowercases before checking - this PR's own test enshrines `"DATA:IMAGE/WEBP;BASE64,abc"` as safe (`imageurl_test.go:13`) - but the FFI consumer scans case-sensitively for lowercase `";base64,"` (`candle-binding/semantic-router.go:1200`), so the whole URI including the `DATA:` prefix feeds the base64 decoder. Executed: `MultiModalEncodeImageFromBase64("DATA:IMAGE/PNG;BASE64,<valid png>", 768)` fails with "illegal base64 data at input byte 4", and `handleEmbeddings` maps every build error to 500 EMBEDDING_GENERATION_FAILED (`route_embeddings.go:27-31`). On `/classify/intent` and `/eval` the same input silently yields no image signal (`pkg/classification/embedding_classifier_backend.go:46` has the same case-sensitive scan; that instance is latent pre-existing on the ExtProc path). Undecodable payloads (`data:image/png;base64,!!!!`) and valid base64 of non-image bytes are the same class: pass validation, then 500. Suggested direction: give `imageurl` a case-insensitive payload extractor as the single source of truth for gate AND consumers, decode-validate the base64 during request validation, and map input-caused encode failures to 400 INVALID_IMAGE. While hardening this path, consider capping `len(req.Images)` - each image is a full 512x512 SigLIP forward, and the 10MB body cap admits a very large count of minimal images.

- **No test can fail if the image signal is never delivered: deleting both new `input.imageURL` arguments compiles and every services, apiserver, and extproc test stays green** (`pkg/services/classification.go:151`, `pkg/services/classification_signal_contract.go:38`). The parameter is variadic (`imageURL ...string`, `pkg/classification/classifier_signal_context.go:54`), so the type system does not defend the wiring - I removed both arguments and `go test` over services, apiserver, and extproc passes everything. The extraction layer is falsifiably tested (good), but the wire from resolved input to the classifier, and the FFI image encode itself, have zero executed coverage anywhere: `TestMultiModalEncodeImageFromBase64` is gated on `MULTIMODAL_MODEL_PATH`, which no workflow or makefile sets, and no e2e exercises `/api/v1/embeddings` with images. Suggested direction: a services-level test with a spy classifier asserting the imageURL argument arrives; a handler-level httptest pinning the 400 INVALID_IMAGE contract (the package already has that idiom); and a Test Plan note that the FFI encode path is env-gated and was validated manually.

#### Minor

- **The HTTP image extractor does not mirror the ExtProc path as its comment claims, so the same conversation yields different image signals depending on entry path** (comment at `pkg/services/intent_request_messages.go:135-137`). ExtProc keeps the FIRST safe image across ALL user turns (`pkg/extproc/utils_fast.go:144-146`, fill-once) while the HTTP path keeps only the current turn's image (overwritten per user turn, including to empty); ExtProc requires part `type == "image_url"` (`utils_fast.go:199`) while HTTP accepts any part carrying an `image_url` object regardless of `type` (`:157-166`); HTTP trims the URL before gating (`:162`) while ExtProc and `/embeddings` gate the raw string; and an image-only turn rotates prior user text into history on HTTP but not ExtProc. Since `/eval` exists to validate rules that fire on the ExtProc path, consider aligning the semantics - or correcting the comment to state the actual contract.
- **Unsafe images are dropped silently at every drop site, so an unsafe-image-only request fails with a misleading "text cannot be empty"** (`pkg/services/intent_request_messages.go:157-166`). Suggested: log at the drop site, and when a request contained image parts that were all rejected, say so in the error.
- **"Closes #1911" needs care: the issue is already closed as NOT_PLANNED by the stale bot (2026-06-28), and the deferred phases are tracked nowhere.** Note the interplay with the required check-linked-issue workflow (fails non-draft PRs with no closing reference, runs on edit): replacing "Closes" with "Part of" would trip it. Suggested: keep "Closes #1911", file a follow-up issue for the batch-classify + similarity work the body defers, and link it in the body. Also, the repo PR template wants the exact headings Purpose / Test Plan / Test Result; the body uses What / Changes / Test Plan with no Test Result section.

#### Design question (for the thread, no code change implied)

- Text results and image results in one `/api/v1/embeddings` response live in different embedding spaces (qwen3/gemma/mmbert vs the multimodal SigLIP projection), so cosine similarity across them is meaningless - and #1911 (image-input API)'s motivation includes similarity use cases. `MultiModalEncodeText` exists (`candle-binding/semantic-router.go:1009`) but is unreachable from this endpoint. Should mixed requests document non-comparability, or should a `model:"multimodal"` value opt text into the shared space?

#### nit

- nit: `pkg/apiserver/routes.go:92` still says "Generate text embeddings"; it feeds the served OpenAPI spec, so "Generate text and image embeddings" keeps `/docs` truthful.
- nit: image results serialize `"text":""` (`EmbeddingResult.Text` has no omitempty); consider omitempty or an input-index field so callers can correlate results across the texts-then-images ordering.
- nit: the three new `IntentRequest` test scaffolds repeat a ~10-line literal; a small helper next to `mustMessageContent` would match the helper-extraction pattern from the review thread on #2174 (test helpers).

### Verdict

The design is right - shared gate, additive schemas, deliberate phasing - and the craft shows. But the Critical is an executed, verified regression on a supported input shape, and the four Importants are each backed by executed repros rather than reading, so I'm marking this request-changes until the Critical (fix included above) and at least the evaluation-text and 500-mapping items are addressed. With those in, I expect this to land quickly - I'll re-review promptly, and thank you again for building the API half of the multi-modal story.

WUKUNTAI-0211:
Thanks for the incredibly thorough review, @shraderdm — the executed repros made every finding trivial to verify. Pushed `e3a65ed4` covering the Critical, both blocking Importants (evaluation-text + 500-mapping), and the wiring/test gap. Details anchored to the new commit.

### Critical — string-valued `image_url` poisoning text extraction ✅
Took your fix as the base. `intentMessageImageURL` now has an `UnmarshalJSON` that accepts both the Chat Completions object form `{"url": "..."}` and the Responses API bare-string form, so a string-valued part can no longer fail the whole content-parts unmarshal. Made it **best-effort** too: an unexpected shape (a JSON number/array/bool `image_url`) yields no image rather than erroring the whole array — same failure class, just for non-API shapes. Verified with `TestIntentRequestResolveSignalInput_StringImageURLDoesNotPoisonText` (`[input_text, input_image:"data:..."]` → `eval="hello world"`, image kept) and `..._MalformedImageURLDoesNotPoisonText` (number-valued `image_url` → sibling text survives).

### Important — image-only turn discarding evaluation text ✅
1. **`extractIntentConversationHistory`** — an image-only user turn attaches its image without clobbering the most recent user text, so an image-only follow-up no longer rotates the real question into history.
2. **Non-user fallback guard** — the promote-`nonUserMessages` fallback is gated on `input.imageURL == ""`, so an image can never promote assistant/system text into the scored slot.
3. **`applyTopLevelTextFallback`** — when the messages path is accepted solely because of an image, evaluation/context/current-user text fall back to `req.Text`.

Net: image safety (client-controlled) can no longer toggle whether the supplied text is scored. Covered by `ImageOnlyTurnFallsBackToTopLevelText` (repro 1) and `ImageOnlyFollowUpKeepsPriorUserText` (repro 2 — asserts the assistant reply is never promoted). The existing `AcceptsImageOnlyUserTurn` (image-only, no top-level text → empty eval) still holds.

### Important — input-driven 500s at the FFI ✅
`imageurl` is now a single case-insensitive parser (the `DATA:IMAGE/PNG;BASE64,` class parses correctly; payload preserved verbatim since base64 is case-sensitive):
- `DecodeBase64` validates the payload at request time → malformed base64 is rejected as **400 INVALID_IMAGE** instead of reaching the FFI as a 500.
- `CanonicalDataURL` rebuilds a lowercase-marker data URI for the FFI's case-sensitive `;base64,` scan (`candle-binding/semantic-router.go:1200`), so the uppercase-scheme input that previously decoded to `illegal base64 data at input byte 4` now succeeds.
- Added a per-request image cap (`maxImagesPerRequest`).

Tests: `RejectsMalformedBase64`, `AcceptsUppercaseDataURIScheme`, `RejectsTooManyImages`, plus `DecodeBase64`/`CanonicalDataURL` unit tests.

### Important — wiring not defended by the type system ✅
Rather than lean on a spy test, I closed the hole directly: `EvaluateAllSignalsWithContext`'s `imageURL` is now a **required parameter** instead of `imageURL ...string`. Dropping the argument at the call sites (`classification.go`, `classification_signal_contract.go`) no longer compiles, so the request→classifier image wiring is type-defended. Callers that carry no image pass `""` explicitly; `EvaluateAllSignalsWithHeaders` forwards its existing `imageURL string` straight through. Plus the handler-level 400 `INVALID_IMAGE` contract tests above. (The FFI image-encode path stays env-gated behind `MULTIMODAL_MODEL_PATH` and was validated manually.)

### Deferred — image dimension contract
Agreed this is real. The image encoder is SigLIP-base-patch16-512 with a native **384-dim** projection, so a default `dimension=768` image request returns 384 while `isValidDimension` rejects 384 — the one value the native model produces. The fix is a product call (reject-above-native breaks the default request; clamp-and-warn; or accept 384 and default images to it) plus surfacing the real model id in `model_used`. My lean is **clamp-and-warn** to keep the default request working, mirroring the batch-clamp precedent in #2145. I'd like to settle direction here before touching the Rust path, then fold it into a commit. Remaining minors/nits (ExtProc mirror comment, logging dropped-unsafe images, `EmbeddingResult.Text` omitempty, `routes.go:92` OpenAPI text, test-helper extraction) I'll take in the same pass.

Keeping `Closes #1911` in the PR body — it stays open and auto-closes when this merges, matching your "track through #2104's merge" note. The deferred classify/batch + similarity phases are already tracked in #2317.

Green locally on the merged tree — classification/services/apiserver/imageurl tests + the agent golangci-lint config. Ready for re-review whenever.

shraderdm (dismissed):
Thanks for the fast, thorough turnaround, @WUKUNTAI-0211 - this is what a fix commit should look like. I pulled `e3a65ed4` down and re-executed the original repros plus the falsifiability checks rather than re-reading the diff.

**Verified fixed (executed):**

- The Critical: both string-`image_url` variants now extract text correctly, and your best-effort `UnmarshalJSON` is better than my snippet - a number/array/bool `image_url` degrades to no-image instead of erroring, which matches the pre-PR ignore semantics. Deleting the method makes both new tests fail, so it's genuinely falsifiable.
- Evaluation-text: all three repros now score the right text. The `req.Text` fallback, the no-rotation change, and the promotion guard match the suggested direction, and image safety no longer toggles which text gets scored.
- Decode-at-validation and the 8-image cap: executed and falsifiable. The case-insensitive gate: executed end-to-end - the uppercase-URI repro now encodes on `/api/v1/embeddings` against a real checkpoint.
- Making `imageURL` a required parameter is a stronger fix than the spy test I suggested - the deletion check now fails at compile time. Nice call.
- CI is green and the branch merges clean over current main (executed the merge, ran the suites on the composed tree).

**On the dimension question: go with clamp-and-warn, per the #2145 (FFI truncation renorm + batch clamp) precedent.** Match the text path's batch-level warn shape, accept 384 in `isValidDimension`, and surface the real model id in `model_used` while you're in there. Happy to review the Rust pass whenever it's ready.

**For the follow-up pass you mentioned (none of these blocking):**

- The classify/eval path still hands the raw (un-canonicalized) URI to the backend - the case-sensitive `;base64,` scan there means an uppercase data URI still silently yields no image signal on those endpoints (executed at tip). Routing it through the same `CanonicalDataURL` would make the commit message's "gate + consumers" fully true; today only `/embeddings` canonicalizes.
- Valid base64 of non-image bytes still maps to a 500; the error-class-mapping half of the earlier suggestion would turn input-caused encode failures into 400s too.
- PR body: a Test Result section, a link to #2317 (the follow-up endpoints issue), and a one-line note that the FFI encode path is env-gated and was validated manually - it's in your comment, but the body is the permanent record.
- The deferred minors (extractor-parity comment, drop-site logging, the `routes.go` "Generate text embeddings" description) whenever you next touch this area.

Approving - everything the review marked blocking is addressed and verified by execution; the one residual from the 500 finding (non-image bytes) is listed above and is not blocking. Thanks again for building the API half of the multi-modal story; you'll get a fast review on the follow-up pass too.

WUKUNTAI-0211:
@shraderdm — quick note on the dismissed approval: I rebased onto main to pick up my own #2384 (mmBERT-manifest-layers), which changed `validateEmbeddingRequest` to take the resolved layers as a param. So this branch had to thread `mmbertLayers` through — that's the only delta since your approval, and it's a verbatim copy of the wiring #2384 already shipped to main. No new logic; the image path is orthogonal to layer validation.

`range-diff` confirms it — commits 1/2/4 byte-identical, 3/5 differ only by the added param + call sites (tests pass `nil`):

```
$ git range-diff e3a65ed4~5..e3a65ed4 6e2d5fe7~5..6e2d5fe7
1:  69aeff13 = 1:  bf761df5 [Router] extract image data-URI gate into shared imageurl package
2:  75eb022d = 2:  991f44d4 [Router] accept image_url input on /api/v1/classify/intent and /api/v1/eval
3:  e0ef05f5 ! 3:  06da3541 [Router] support image inputs on /api/v1/embeddings
4:  39ebb4f6 = 4:  99b038bb Clarify embedding avg processing denominator
5:  e3a65ed4 ! 5:  6e2d5fe7 [Router] Fix image-input integration seams on classify/eval/embeddings
```

CI is green on the rebased tip. Re-approve when you get a moment? Thanks!

shraderdm (approved):
Re-approving. I re-verified against `6e2d5fe7` rather than trusting the range-diff - and it checks out exactly as you described: commits 1/2/4 byte-identical, the 3/5 deltas are only the `mmbertLayers` context plus the six `, nil` test call sites, and the tip's `parseEmbeddingRequest` is byte-identical to main's - the threading is #2384's own code inherited through the rebase, not a re-implementation. Re-executed the original repros on the new tip (string `image_url` variants, the `Text` + image-only fallback, the uppercase-URI validation with the new signature) - all green, CI green. Thanks for the range-diff in the comment - it made this fast.

PR #2106, opened 2026-06-10
[Router] Authenticate looper requests to close security-pipeline bypass (#1443)

### LINKED ISSUE #1443: [Security] Prevent Looper header injection from bypassing security plugins
## Summary

Any external client can send `x-vsr-looper-request: true` in the request headers to make the router treat the request as an internal looper call. This skips the entire security pipeline: no decision evaluation, no jailbreak detection, no PII checks, no authz validation.

## Root Cause

`processor_req_header.go:202` reads the looper header directly from the client request with no validation that the request actually originated from the internal looper client:

```go
if h.Key == headers.VSRLooperRequest && headerValue == "true" {
    ctx.LooperRequest = true
}
```

When `ctx.LooperRequest` is true, `processor_req_body.go:87-89` skips the entire normal processing flow (decision evaluation, signal classification, authz check) and enters the looper fast-path.

## Why This Header Exists

The looper client makes HTTP requests **through Envoy** back to the same router (for multi-model execution like confidence/ratings/ReMoM algorithms). On the second pass through ExtProc, the router needs to distinguish internal looper calls from external user requests. The `x-vsr-looper-request` header was the mechanism — but without authentication, any external client can set it.

## Impact

An attacker can bypass all security controls with a single header:

```bash
curl http://vsr-endpoint/v1/chat/completions \
  -H "x-vsr-looper-request: true" \
  -H "x-vsr-looper-decision: any_decision" \
  -d '{"model":"any-model","messages":[{"role":"user","content":"bypass all safety"}]}'
```

This skips:
- Decision evaluation (performDecisionEvaluation)
- Jailbreak detection
- PII detection
- Authz signal evaluation
- Rate limiting context

## Proposed Fix

Authenticate looper requests with a shared secret:

1. Router generates a cryptographic secret at startup (`crypto/rand`, 256-bit)
2. Secret is passed to the looper client in-memory (never serialized to config/disk)
3. Looper client sends the secret in `x-vsr-looper-secret` header with every internal call
4. Router validates the secret using constant-time comparison before accepting a looper request
5. Requests without a valid secret are rejected and logged as potential bypass attempts

As defense-in-depth, deployments should also strip `x-vsr-looper-*` headers in Envoy config:
```yaml
request_headers_to_remove: ["x-vsr-looper-request", "x-vsr-looper-secret"]
```

jianyexi:
## Summary

Fixes #1443.

Any external client could send `x-vsr-looper-request: true` to make the ext_proc
filter treat the request as an internal looper call, which skips the **entire**
security pipeline: decision evaluation, jailbreak detection, PII detection, and
authz. The root cause was that ext_proc set `ctx.LooperRequest` directly from a
client-controlled header with no authentication.

```bash
# Before this change, this bypassed all safety checks:
curl http://vsr-endpoint/v1/chat/completions \
  -H "x-vsr-looper-request: true" \
  -d '{"model":"any-model","messages":[{"role":"user","content":"bypass all safety"}]}'
```

## Approach

Introduce an in-process looper secret:

- The router owns a per-process secret generated with `crypto/rand` (256-bit),
  or a shared `VSR_LOOPER_SECRET` for multi-replica deployments. It is never
  written to config or disk.
- The in-process looper client attaches it as `x-vsr-looper-secret` on every
  internal call.
- ext_proc only honors the `x-vsr-looper-request` fast-path when the secret
  validates with a **constant-time** comparison (`crypto/subtle`).
- A forged marker without a valid secret is logged
  (`extproc looper_request_rejected`) and processed as a **normal external
  request**, so the full security pipeline still runs. This fails to normal
  processing rather than hard-rejecting, avoiding a DoS vector from a spoofed
  header.

This complements the existing defense-in-depth Envoy header stripping
(`request_headers_to_remove`), which runs in Envoy's router filter **after**
ext_proc and therefore cannot by itself stop a client-injected header from
reaching ext_proc. The application-layer secret is the primary control.

### Multi-replica note

By default each process generates its own random secret (correct for
single-process and sidecar-per-Envoy topologies). For multi-replica router
deployments where Envoy may load-balance the internal looper call to a different
replica, set the same `VSR_LOOPER_SECRET` on every replica (e.g. from a
Kubernetes Secret).

## Changes

- `pkg/looper/secret.go` (new) — `Secret()` / `ValidateSecret()` with env
  override + crypto/rand fallback and constant-time compare.
- `pkg/looper/client.go` — send `x-vsr-looper-secret` on internal calls.
- `pkg/extproc/processor_req_header.go` — gate `ctx.LooperRequest` on a valid
  secret; warn-log rejected markers.
- `pkg/headers/headers.go` — add the `VSRLooperSecret` constant.
- `docs/architecture/security-hardening.md` — document the mechanism and
  `VSR_LOOPER_SECRET`.

## Test Plan

- `gofmt` clean on all changed Go files.
- Ran the secret crypto logic standalone (secret stability, constant-time
  accept/reject, `VSR_LOOPER_SECRET` override with trimming, random 256-bit
  generation): all pass.
- Added `pkg/looper/secret_test.go` and an ext_proc regression test asserting
  forged looper requests (missing **and** wrong secret) are not honored, while a
  request carrying a valid secret still enters the looper fast-path.
- Full `pkg/looper` + `pkg/extproc` build/tests require CGO and the prebuilt
  Rust candle binding; deferred to CI.

This is a behavior-visible security fix to the request-header path; the added
unit/regression tests cover the new accept/reject behavior.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/docs/architecture/security-hardening.md
+++ b/docs/architecture/security-hardening.md
@@ -50,6 +50,33 @@ This is configured in both:
 - `deploy/local/envoy.yaml` (local development)
 - `src/vllm-sr/cli/templates/envoy.template.yaml` (production template)
 
+### Looper Request Authentication
+
+Stripping headers at Envoy is defense-in-depth, not the primary control: the
+`request_headers_to_remove` rules run in Envoy's router filter, which executes
+*after* ext_proc. They stop internal headers from leaking to upstream backends,
+but they do not stop a client-injected header from reaching ext_proc.
+
+The primary control is an application-layer secret. The router owns a looper
+secret (`pkg/looper.Secret()`); the in-process looper client attaches it as
+`x-vsr-looper-secret` on every internal multi-model call, and ext_proc only
+honors the `x-vsr-looper-request` fast-path when that secret validates with a
+constant-time comparison. A request that carries `x-vsr-looper-request` without
+a valid secret is logged (`extproc looper_request_rejected`) and processed as a
+normal external request, so it still passes through decision evaluation,
+jailbreak, PII, and authz checks. This closes the bypass where any client could
+set `x-vsr-looper-request: true` to skip the entire security pipeline.
+
+**Secret source:**
+
+- By default each router process generates a random 256-bit secret in memory at
+  first use. This is correct for single-process and sidecar-per-Envoy
+  deployments, where the looper's internal call returns to the same process.
+- For multi-replica deployments where Envoy may load-balance the internal looper
+  call to a different router replica, set the `VSR_LOOPER_SECRET` environment
+  variable to the same value on every replica (e.g. from a Kubernetes Secret) so
+  all replicas trust one secret. The secret is never written to config or disk.
+
 ## 2. Dashboard RBAC Integration
 
 ### Security Policy API
@@ -208,3 +235,6 @@ For production multi-user deployments:
 - [ ] Review dashboard user roles — only admins should have `security.manage`
 - [ ] Ensure the looper endpoint is only accessible from the router container
       (network-level isolation)
+- [ ] For multi-replica router deployments, set a shared `VSR_LOOPER_SECRET`
+      (e.g. from a Kubernetes Secret) on every replica so internal looper calls
+      are accepted across replicas
--- a/src/semantic-router/pkg/extproc/processor_req_header.go
+++ b/src/semantic-router/pkg/extproc/processor_req_header.go
@@ -13,6 +13,7 @@ import (
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/headers"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/ir"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/looper"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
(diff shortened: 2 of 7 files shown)


Xunzhuo:
Hi @jianyexi, thank you for taking on #1443 — the vulnerability analysis and the core authenticated-reentry direction are both valid.

Since this branch was last updated, the Looper/extproc paths have changed substantially and this PR now conflicts with `main`. In parallel, maintainer draft #2508 now carries the current-main implementation for the same trust boundary, including strict shared-secret validation, reserved-header sanitization before upstream/provider use, redirect/proxy hardening, reload and multi-replica lifecycle handling, and unit/E2E coverage.

To avoid asking you to redo work against a moving target, I suggest treating #2508 as superseding this PR. If you would prefer to continue with a focused version here, please let us know first — we can coordinate the exact remaining scope and reuse the hardened behavior/tests from #2508 before you spend time rebasing. Thank you again; identifying and addressing this boundary was valuable.

jianyexi:
Thanks @Xunzhuo — that makes sense. Since #2508 already carries the current-main implementation for this trust boundary (strict shared-secret validation, reserved-header sanitization, redirect/proxy hardening, reload + multi-replica lifecycle handling, and unit/E2E coverage), I'm happy to treat #2508 as superseding this PR and will close #2106 to avoid duplicate work against a moving target.

Please carry over the #1443 analysis here as needed. I'm glad to review #2508 and confirm the original security-pipeline bypass path is fully covered. Thanks for the hardened follow-up!

Issue #2107, opened 2026-06-10
[Router] Download failures without HF_TOKEN are soft-skipped as gated models, masking failed downloads of public repos


shraderdm:
`IsModelComplete` failures aside (tracked separately), `IsGatedModelError` in `pkg/modeldownload/downloader.go` returns true for ANY download error when no `HF_TOKEN` is set (`noToken := hfToken == ""`), not just auth/gated error patterns. A transient failure (rate limit, network, OOM-killed CLI subprocess) on a public repo is then logged as a gated-model skip, counted as ready, and the router proceeds to classifier init, which hard-fails on the missing files - turning a retryable download failure into a startup crash with a misleading error far from the cause.

Observed in practice on a GKE chart deployment: anonymous downloads of `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` hit HF rate limits, were soft-skipped as "gated", and the router crash-looped at embedding init with `failed to load safetensors from models/mmbert-embed-32k-2d-matryoshka/model.safetensors` - two layers removed from the real error (a rate-limited download).

The comment on the check says the intent is graceful degradation for CI forks without secrets, which makes sense for genuinely gated models (embeddinggemma). The over-match is treating every failure as gated when no token is present.

Suggested direction: only soft-skip when the error matches the auth/gated patterns or the repo is known-gated; otherwise surface the real download error (or fail fast when the model is hard-required by the config). Happy to take this one - filing it separately to keep the related completeness fix single-concern.

PR #2108, opened 2026-06-10
[Router] Require candle embedding runtime files for model completeness

### LINKED ISSUE #2531: [Router] Model-download completeness omits qwen3/gemma/multimodal candle runtime files, so partial downloads never heal
### What happened

The model-download completeness check does not require the files the candle embedding runtime hard-loads at startup for the qwen3, gemma, and multimodal embedding paths. `IsModelComplete` (`src/semantic-router/pkg/modeldownload/validator.go`) treats a directory as complete once its recorded `RequiredFiles` are present and `hasModelWeights` finds any weight file, and `hasModelWeights` matches `*.onnx` via a recursive `filepath.WalkDir`. So a partial download that holds only the companion artifacts (`config.json`, nested `onnx/` exports) but not the root weights reads as complete, is skipped on later starts, and does not self-heal through restart.

The impact differs by path. A missing qwen3 or multimodal weight or tokenizer aborts model-factory initialization (`candle-binding/src/ffi/embedding.rs`); for the weights case the router crash-loops at classifier init with `failed to load safetensors from <model>/model.safetensors: No such file or directory`. A missing gemma file logs a warning and continues (gemma is registered as optional), leaving that model unavailable rather than aborting. The repeated-restart crash-loop was observed on the analogous mmbert path before #2195; it is not independently reproduced per-path here.

This is the same non-healing completeness-check defect #2195 (embedding weights + tokenizer completeness) fixed for the configured semantic embedding model (`mmbert`). The qwen3, gemma, and multimodal candle paths share it and are not covered by #2195:

- **qwen3 / gemma:** `model.safetensors` + `tokenizer.json`; gemma additionally hard-loads `2_Dense/model.safetensors` and `3_Dense/model.safetensors` (the dense bottleneck the gemma embedding model loads).
- **multimodal:** `model.safetensors` + `tokenizer.json`, loaded whenever the embedding `model_type` resolves to `multimodal`, and independently whenever a complexity rule declares non-empty `image_candidates`.

### Expected

A partial download missing any file the candle embedding runtime hard-loads should read as incomplete, so the next startup resumes the download instead of leaving the model broken, rather than being treated as complete because companion ONNX artifacts are present.

### Additional context

Extends the completeness contract from #2195. Out of scope: whatever interrupts the first download (converging state across restarts rather than poisoning it permanently is the goal here), and the gated-model soft-skip that masks public-repo download failures (tracked in #2107).

shraderdm:
Closes #2531

## Purpose
- Extend the model downloader's completeness contract to everything the candle embedding runtime hard-loads at startup. #2195 (embedding weights+tokenizer completeness) fixed the headline case for the configured semantic embedding model (`mmbert`): a partial download holding only companion artifacts (`config.json`, nested `onnx/` exports) satisfied the generic weight heuristic in `IsModelComplete` (`*.onnx` matches via the recursive walk), was treated as complete, was never re-downloaded, and the router crash-looped at classifier init on every restart.
- This PR covers the rest of the candle runtime's load surface, which has the same non-healing failure shape and is not covered by #2195 (embedding weights+tokenizer completeness):
  - the qwen3 and gemma embedding paths (`model.safetensors` + `tokenizer.json`), with gemma additionally requiring `2_Dense/model.safetensors` and `3_Dense/model.safetensors`, which `BottleneckDenseNet::load_from_path` hard-loads;
  - the multimodal path, required whenever `model_type` resolves to `multimodal` or any complexity rule declares `image_candidates` (both load the multimodal model; the latter has no model_type gate).
- Gating follows the canonical backend resolution (`EmbeddingModels.EmbeddingBackend()`), so the qwen3/gemma/mmbert requirements apply only when the backend resolves to candle. Remote deployments (`openai_compatible`, including `model_type: remote` with no backend set, per #2276 (external embedding providers)) and openvino deployments load none of these files and gain no requirements from this change. (Separately and pre-existing: #2195's mmbert requirement is guarded only against the remote backend, so it still applies under openvino; that interaction predates this PR and is out of scope here.)
- `IsModelComplete` checks `RequiredFiles` before the weight heuristic, so partial downloads read as incomplete and the next startup resumes the download instead of crash-looping.
- Module: `Router` (`pkg/modeldownload`, plus a keep-in-sync comment in `pkg/classification`).
- Scope notes:
  - The generic nested-weight heuristic is intentionally unchanged (it serves nested-onnx layouts and is test-pinned), and #2195's mmbert requirement is left in place; the two mechanisms deduplicate.
  - `bert_model_path` is excluded because its loader has a weight-format fallback chain (`similarity.rs`); the covered loaders hard-code their files with no fallback.
  - All four registry repos for the covered paths ship the required files (verified via the HF tree API, including gemma's `2_Dense/` and `3_Dense/` weights), so the added requirements cannot produce false-incomplete for registry models. The requirements are a deliberate superset: they are recorded for configured paths even when no rule ends up loading the model, so a custom non-registry path that lacks `tokenizer.json` would re-attempt its download on each boot.
  - Two adjacent items remain deliberately out of scope: (1) whatever interrupts the first download (with this fix the state converges across restarts instead of poisoning permanently); (2) `IsGatedModelError` treating token-less download failures as gated-model soft-skips even for public repos, filed as #2107 (gated-model soft-skip). Healing is therefore eventual (next successful retry), not guaranteed-on-first-restart.
  - Residual: completeness checks file existence, not integrity; a truncated root weight file from a non-atomic copy onto a pre-mounted volume would still read complete.

## Test Plan
- `go test ./pkg/modeldownload/` covering: the candle-derived requirements under the canonical backend resolution (empty/candle/case-insensitive add them; openvino, `openai_compatible`, and `model_type: remote` with empty backend do not), gemma's dense-bottleneck files, multimodal coverage independent of backend plus the `image_candidates` trigger and model_type edge cases, a `BuildModelSpecs`-level assertion that specs carry the new requirements, non-`models/` paths skipped, and a regression test reproducing the poisoning scenario at the unit level (partial dir reads incomplete, then complete once the runtime files land).
- No e2e testcase update: there is no existing e2e surface for the model downloader, and the failing precondition (a poisoned volume that predates pod start) does not fit the request-level e2e harness; the cluster A/B below covers the behavior-visible path end to end.
- Cluster A/B on GKE with the `semantic-router` 0.3.0 chart and a pre-seeded poisoned models PVC (`config.json` + `onnx/layer-6/*.onnx` + `.cache/huggingface` metadata, no root weights). Note this A/B predates #2195 (embedding weights+tokenizer completeness) landing on main and demonstrates the shared failure shape on the mmbert path; on current main that specific path heals via #2195, and this PR extends the same healing to the qwen3/gemma/multimodal paths.

## Test Result
- Unit: all `pkg/modeldownload` tests pass, including the pre-existing nested-onnx heuristic cases (intentional behavior preserved) and #2195's completeness tests. `gofmt`, `go vet`, codespell clean. Red-green on the backend-resolution gate: the `model_type: remote` case fails against a hand-rolled empty-means-candle check and passes with the canonical resolution.
- Cluster A/B (GKE, chart 0.3.0, models PVC pre-seeded with `config.json` + `onnx/layer-6/model_fa_fp16.onnx` + `.cache/huggingface` metadata, no root weights; historical run - v0.3.0 predates #2195 (embedding weights+tokenizer completeness) and demonstrates the shared failure shape on the mmbert path):
  - Control (stock `extproc:v0.3.0`): logs `required_models_already_present total_models=4` with no download attempt, then `failed to load safetensors from models/mmbert-embed-32k-2d-matryoshka/model.safetensors: No such file or directory` and `startup_failed`; crash-looped past 7 restarts without ever healing.
  - Treatment (patched image, same volume): logs `Downloading model: models/mmbert-embed-32k-2d-matryoshka`, fetches the repo (with `HF_TOKEN` wired; anonymous downloads rate-limit and the failure is masked by the gated-model soft-skip, #2107 (gated-model soft-skip)), then `keyword_embedding_backend_initialized backend=candle mmbert_2d_enabled=true` and the pod reaches Ready. The previously missing files land on the volume (`model.safetensors` 613892480 bytes, `tokenizer.json` 34363443 bytes).
  - Negative control (restart on the healed volume): `required_models_already_present total_models=4`, zero `Downloading model` lines, Ready with 0 restarts. No false-incomplete loop.
  - Incidental repro of the interruption class: an early treatment run with a low memory limit was OOMKilled mid-fetch, leaving exactly the partial layout this PR detects; with the fix the next boot resumed instead of poisoning.

### DIFF
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -168,11 +168,86 @@ func addEmbeddingModelRequiredFiles(cfg *config.RouterConfig, requiredFilesByMod
 }
 
 // ExtractRequiredFilesByModel derives per-model completeness requirements from
-// config-owned companion files such as category/jailbreak/PII mappings.
+// config-owned companion files such as category/jailbreak/PII mappings, plus
+// the files the configured embedding runtime hard-loads at startup.
 func ExtractRequiredFilesByModel(cfg *config.RouterConfig) map[string][]string {
 	requiredFilesByModel := make(map[string][]string)
 	collectRequiredFilesByModel(reflect.ValueOf(cfg), requiredFilesByModel)
+	collectCandleEmbeddingRequiredFiles(cfg, requiredFilesByModel)
 	return requiredFilesByModel
+}
+
+// candleEmbeddingRequiredFiles are the files the candle embedding runtime
+// hard-loads from each configured embedding model directory at startup:
+// candle-binding builds its VarBuilder from <model>/model.safetensors and the
+// tokenizer from <model>/tokenizer.json, with no fallback formats.
+var candleEmbeddingRequiredFiles = []string{"model.safetensors", "tokenizer.json"}
+
+// gemmaEmbeddingRequiredFiles extends the common candle set with the dense
+// bottleneck weights GemmaEmbeddingModel additionally hard-loads from the
+// 2_Dense/ and 3_Dense/ subdirectories (BottleneckDenseNet::load_from_path).
+var gemmaEmbeddingRequiredFiles = []string{
+	"model.safetensors",
+	"tokenizer.json",
+	"2_Dense/model.safetensors",
+	"3_Dense/model.safetensors",
+}
+
+// collectCandleEmbeddingRequiredFiles records the candle runtime's startup
+// files as per-model completeness requirements for a superset of the embedding
+// models the candle runtime may load: the qwen3/gemma/mmbert paths under the
+// candle backend, and the multimodal path whenever model_type selects it or
+// complexity rules declare image_candidates (which loads the multimodal model
+// regardless of model_type).
+//
+// Without this, an interrupted download that fetched only companion artifacts
+// (config.json, nested onnx/ exports) satisfies the generic weight heuristic
+// in IsModelComplete: the model is treated as complete, is never re-downloaded
+// on restart, and the router crash-loops at classifier init with
+// "failed to load safetensors from <model>/model.safetensors". Recording the
+// runtime-required files makes such partial downloads read as incomplete so
+// the next startup resumes the download instead of crash-looping.
+//
+// The embedding path fields are enumerated explicitly (unlike the reflective
+// ExtractModelPaths walk) because the requirement is loader-specific: only the
+// paths whose loaders hard-code these files are covered. A new embedding path
+// field needs a matching entry here to gain partial-download protection.
+func collectCandleEmbeddingRequiredFiles(cfg *config.RouterConfig, requiredFilesByModel map[string][]string) {
+	// Mirror the backend resolution in classification's
+	// ExternalModelBasedEmbeddingInitializer: empty backend defaults to
+	// candle, case- and space-insensitively. Keep the two in sync.
+	backend := strings.ToLower(strings.TrimSpace(cfg.EmbeddingModels.EmbeddingConfig.Backend))
+	if backend == "" || backend == "candle" {
+		appendRequiredFiles(requiredFilesByModel, cfg.EmbeddingModels.Qwen3ModelPath, candleEmbeddingRequiredFiles)
+		appendRequiredFiles(requiredFilesByModel, cfg.EmbeddingModels.GemmaModelPath, gemmaEmbeddingRequiredFiles)
+		appendRequiredFiles(requiredFilesByModel, cfg.EmbeddingModels.MmBertModelPath, candleEmbeddingRequiredFiles)
+	}
+
+	// The multimodal model is loaded independently of the backend value:
+	// classification's initializer branches run it whenever model_type
+	// resolves to "multimodal", and the complexity classifier additionally
+	// loads it whenever any complexity rule declares image_candidates,
+	// without any model_type gate. Keep these triggers in sync with
+	// classifier_option_rules.go.
+	modelType := strings.ToLower(strings.TrimSpace(cfg.EmbeddingModels.EmbeddingConfig.ModelType))
+	if modelType == "multimodal" || config.HasImageCandidatesInRules(cfg.ComplexityRules) {
+		appendRequiredFiles(requiredFilesByModel, cfg.EmbeddingModels.MultiModalModelPath, candleEmbeddingRequiredFiles)
+	}
+}
+
+// appendRequiredFiles records files as completeness requirements for a local
+// model path, deduplicating against already-recorded requirements.
+func appendRequiredFiles(requiredFilesByModel map[string][]string, modelPath string, files []string) {
+	if !strings.HasPrefix(modelPath, "models/") || modelPath == "models/" {
+		return
+	}
+	requiredFiles := requiredFilesByModel[modelPath]
+	for _, file := range files {
+		if !slices.Contains(requiredFiles, file) {
+			requiredFiles = append(requiredFiles, file)
+		}
+	}
+	requiredFilesByModel[modelPath] = requiredFiles
 }
 
 func collectRequiredFilesByModel(v reflect.Value, requiredFilesByModel map[string][]string) {
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -213,11 +213,12 @@ var gemmaEmbeddingRequiredFiles = []string{
 // paths whose loaders hard-code these files are covered. A new embedding path
 // field needs a matching entry here to gain partial-download protection.
 func collectCandleEmbeddingRequiredFiles(cfg *config.RouterConfig, requiredFilesByModel map[string][]string) {
-	// Mirror the backend resolution in classification's
-	// ExternalModelBasedEmbeddingInitializer: empty backend defaults to
-	// candle, case- and space-insensitively. Keep the two in sync.
-	backend := strings.ToLower(strings.TrimSpace(cfg.EmbeddingModels.EmbeddingConfig.Backend))
-	if backend == "" || backend == "candle" {
+	// Use the canonical backend resolution (EmbeddingModels.EmbeddingBackend):
+	// only the candle backend hard-loads these local files at startup. The
+	// remote backend (openai_compatible, including model_type=remote with no
+	// backend set) and openvino load nothing from these paths, so requiring
+	// the files there would force downloads of models that never load.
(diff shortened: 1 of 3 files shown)


shraderdm:
Refreshed this after a month of drift: rebased onto current main and repositioned around what landed meanwhile - #2195 (embedding completeness) independently fixed the mmbert case, and #2276 (external embedding providers) added a remote backend which this PR now respects via the canonical resolution, with a new test covering model_type: remote. What remains in scope here is the rest of the candle runtime's hard-load surface: qwen3, gemma (2_Dense/3_Dense bottleneck weights), and the multimodal path - the same never-heals partial-download shape #2195 fixed for mmbert. Body updated to match. Happy to adjust scope or split if preferred.

shraderdm:
One more note from a post-open sweep: I just noticed #2066 (model lifecycle) rewrites this same config_parser.go surface with a lifecycle catalog that includes per-model companion files. If that lands first, the per-model requirements added here fold naturally into its catalog shape - happy to rebase this onto that instead. Until then, this PR extends the current mechanism the same way #2195 (embedding completeness) did.

src/semantic-router/pkg/modeldownload/config_parser.go
```diff
@@ -168,13 +168,89 @@ func addEmbeddingModelRequiredFiles(cfg *config.RouterConfig, requiredFilesByMod
...
+// ExtractModelPaths walk) because the requirement is loader-specific: only the
+// paths whose loaders hard-code these files are covered. A new embedding path
+// field needs a matching entry here to gain partial-download protection.
+func collectCandleEmbeddingRequiredFiles(cfg *config.RouterConfig, requiredFilesByModel map[string][]string) {
+	// Use the canonical backend resolution (EmbeddingModels.EmbeddingBackend):
+	// only the candle backend hard-loads these local files at startup. The
+	// remote backend (openai_compatible, including model_type=remote with no
+	// backend set) and openvino load nothing from these paths, so requiring
+	// the files there would force downloads of models that never load.
+	if cfg.EmbeddingModels.EmbeddingBackend() == config.EmbeddingBackendCandle {
```
Xunzhuo, line 221: [P1] The OpenVINO skip here is undone by the existing `BuildModelSpecs` path. After this function returns, `BuildModelSpecs` still calls `addEmbeddingModelRequiredFiles`, whose guard skips only remote backends, so an OpenVINO config with `MmBertModelPath` gets `model.safetensors` and `tokenizer.json` re-added. That contradicts this backend contract and can force downloads of Candle-only artifacts. Please remove/fold the legacy helper (or make its backend guard identical) and add a `BuildModelSpecs` regression test for OpenVINO and remote backends.

src/semantic-router/pkg/modeldownload/config_parser.go
```diff
@@ -168,13 +168,89 @@ func addEmbeddingModelRequiredFiles(cfg *config.RouterConfig, requiredFilesByMod
...
+	}
+
+	// The multimodal model is loaded independently of the backend value:
+	// classification's initializer branches run it whenever model_type
+	// resolves to "multimodal", and the complexity classifier additionally
+	// loads it whenever any complexity rule declares image_candidates,
+	// without any model_type gate. Keep these triggers in sync with
+	// classifier_option_rules.go.
+	modelType := strings.ToLower(strings.TrimSpace(cfg.EmbeddingModels.EmbeddingConfig.ModelType))
+	if modelType == "multimodal" || config.HasImageCandidatesInRules(cfg.ComplexityRules) {
```
Xunzhuo, line 234: [P1] This trigger no longer matches startup. For every non-remote backend, `buildEmbeddingRuntimeTasks` appends `multiModalEmbeddingRuntimeTask` whenever `MultiModalModelPath` is non-empty; it does not require `model_type=multimodal` or image candidates. With the default model type, a partial multimodal directory can therefore pass completeness here and then fail/degrade during the eager runtime init. Please derive this requirement from the current runtime gate (or change the runtime gate), and replace the test assertion that intentionally expects no requirement when `model_type` is empty with an end-to-end `BuildModelSpecs` case.

Xunzhuo (changes requested):
Thanks for extending the completeness checks across the Candle embedding surface. I found two runtime/BuildModelSpecs mismatches that still allow the contract to be wrong end to end. Please address both and add tests through BuildModelSpecs (not only ExtractRequiredFilesByModel). I ran `go test ./pkg/modeldownload` successfully; the current tests do not exercise these regressions.

shraderdm:
Still active on this, and apologies for the long gap.

Both mismatches you flagged are fixed, and the tests now assert through `BuildModelSpecs` rather than only `ExtractRequiredFilesByModel`.

The ordering one is fixed by removing `addEmbeddingModelRequiredFiles` outright rather than resequencing it. The embedding requirements now live inside `ExtractRequiredFilesByModel`, so there is no longer a later helper that can re-add them after the backend gate has run. The multimodal one now derives from the runtime trigger instead of `model_type`: `multiModalEmbeddingRuntimeTask` fires whenever `multimodal_model_path` is set on any non-remote backend, with no `model_type` gate, so the requirement follows that.

On the testing ask specifically, the assertions ended up at both levels on purpose. The matrices run through `BuildModelSpecs` since that is what `GetMissingModels` consumes, but `BuildModelSpecs` cannot observe the non-remote gate at all, because `optionalModelFeatureGates` drops those paths first under a remote backend and deleting that gate outright leaves a spec-level suite green. Gemma and multimodal are crossed with the backend at collector level for that reason, and each fix carries a red-green receipt reproduced as a `go test -overlay` mutation.

On the delay: the change grew well beyond the original diff, mostly in getting the disclosure right about which loader owns which path and what actually fails when a file is missing, and I then had an unexpectedly long stretch away from it. A status note here would have been the right call much earlier.

I am rebasing onto current main now. #2770 landed in the same function, though the path-extraction half rather than the requirements half, so I expect that to be mechanical. I will push and reply properly once it is green.

shraderdm:
Correction to my last comment: both technical claims in it were wrong. `multiModalEmbeddingRuntimeTask` does not fire on every non-remote backend - #3316 (ship the OpenVINO backend) put an OpenVINO early return ahead of it. And the landing in this function was #3049 (candle runtime files in completeness checks), not #2770.

#3049 covers the qwen3/gemma/multimodal completeness contract, the gemma dense weights, and the ONNX-only tests, and closed #2531. Closing this in favor of it.

#3049 left an open question about narrowing the `UsesRemoteEmbeddingBackend()` gate per path. After #3316 that is worth doing, and it needs OpenVINO requirements rather than just dropping the candle ones. Putting the details on #2668 (honor declared embedding backends) since @nanasis has that one.

Thank you for the review on this one. Sorry it didn't end up landing.

PR #2109, opened 2026-06-10
[Router] Surface public model download failures without HF_TOKEN

### LINKED ISSUE #2107: [Router] Download failures without HF_TOKEN are soft-skipped as gated models, masking failed downloads of public repos

WUKUNTAI-0211:
## Summary

Closes #2107.

`IsGatedModelError` treated any download failure as gated whenever `HF_TOKEN` was empty. That masked public-repo failures such as rate limits, network errors, or plain CLI subprocess failures as gated-model skips, allowing startup to continue until a later model-load crash surfaced a misleading missing-file error.

This PR narrows the soft-skip path: known gated repos and explicit auth/gated error messages are still skipped gracefully, but public-repo download failures now surface the original download error.

## Changes

- Remove `hfToken == ""` as a standalone gated-model condition.
- Keep known gated model names as the fallback for CLI errors that only expose `exit status 1`.
- Add table-driven regression coverage for gated/auth failures vs public rate-limit, network, and generic CLI failures.

- Anchor status codes to their surrounding phrase (`403 forbidden`, not bare `403`) so public-repo progress bytes (`403M/896M`) and proxy/CDN denials are not misread as gated.
- Make the `embeddinggemma`/`gemma` repoID allowlist a true last resort — consulted only when there is no captured CLI output to classify on.
- Surface the real download cause: interpolate a bounded stderr tail into the returned error instead of a bare `exit status 1`.

## Test Plan

- `GOCACHE=/tmp/go-build-cache go test ./pkg/modeldownload -run TestIsGatedModelErrorClassifiesOnlyGatedOrAuthFailures -count=1 -v`
- `GOCACHE=/tmp/go-build-cache go test ./pkg/modeldownload/... -count=1`
- `git diff --check`

`make agent-lint CHANGED_FILES="src/semantic-router/pkg/modeldownload/downloader.go,src/semantic-router/pkg/modeldownload/downloader_test.go"` was attempted locally, but the repo hook failed before code-specific lint with `golangci-lint` reporting `no go files to analyze`. GitHub CI should run the canonical lint/build path on this PR.

## Test Result

```
$ go test ./pkg/modeldownload/ -run TestIsGatedModelErrorClassifiesOnlyGatedOrAuthFailures -count=1
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/modeldownload	0.003s

$ go vet ./pkg/modeldownload/
(clean)
```

13 subtests pass, including the two public-repo false-positive fixtures from review
(progress-bytes-then-429, gemma-named public mirror) and a multi-line
`huggingface_hub.errors.GatedRepoError` stderr case.


FAUST-BENCHOU (commented):
I dont think HF_TOKEN is the reason. an invalid HF identifier will still result in a 429 rate limit for huggingface.

WUKUNTAI-0211:
@FAUST-BENCHOU you're right, and thanks — HF_TOKEN was the wrong lens. An invalid identifier (and a rate limit) both surface as a 429 regardless of token, so token state can't distinguish gated vs transient vs typo'd-repo.

The deeper issue was that we couldn't even *see* the real error: `hf download` streams its output to stderr, so `cmd.Run()` returns just `exit status 1`. The `isAuthError` string matches were effectively dead, and `noToken` was a proxy that masked every failure as a gated skip (then counted it ready → the misleading safetensors crash two layers later).

Pushed b892b16a to fix the root cause rather than the symptom:

- **Capture the CLI's stderr** (bounded to the tail) and classify on that real text instead of `err` alone.
- **Drop the `hfToken` parameter** entirely — no more token heuristic.
- **Stop treating 404 / "repository not found" as gated**, so a missing or mistyped repo fails loudly instead of being soft-skipped as ready.

Net effect: only genuine gated/auth output (`gated`, `restricted`, `401`, `403`, "must be authenticated", …) is soft-skipped; 429 rate limits, network errors, and bad identifiers all surface. Known-gated families (gemma/embeddinggemma) remain a fallback for the case where the CLI emits nothing classifiable. Tests cover the 429-invalid-identifier and 404 cases you raised.

WUKUNTAI-0211:
@FAUST-BENCHOU friendly ping — this addresses your point directly. The token heuristic is gone: `IsGatedModelError` now captures the HF CLI's stderr (tailWriter) and classifies on the real output instead of inferring from `HF_TOKEN` state. 404 / "repository not found" is no longer treated as gated, so a typo'd repo and a 429 rate limit both surface loudly; only genuine gated/auth output is soft-skipped, with the known-gated family list kept only as a fallback when the CLI emits nothing classifiable. Tests cover the 429-invalid-identifier and 404 cases you raised.

Rebased onto latest `main` (linear history, no merge commit) and `go test ./pkg/modeldownload/...` is green. Mind taking another look?

WUKUNTAI-0211:
@shraderdm — this PR addresses your #2107 report. Pinging since FAUST's
review feedback has been incorporated and the PR has been sitting green
for ~20 days.

Recap of what changed since you filed #2107:

- Dropped the `hfToken == ""` heuristic entirely. `IsGatedModelError` now
  classifies on actual HF CLI stderr captured via a `tailWriter`, not on
  token state.
- Public-repo 429 rate limits and 404 typos now surface as the real
  download error instead of being soft-skipped as gated. That matches
  your GKE crash-loop scenario exactly — a rate-limited
  `mmbert-embed-32k-2d-matryoshka` download would now fail loudly at the
  download step, not two layers later at safetensors load.
- Known-gated repo list (`embeddinggemma` etc.) kept as a fallback only
  when the CLI emits nothing classifiable, preserving the original
  graceful-degradation intent.
- Table-driven coverage for gated/auth vs public rate-limit / network /
  generic CLI failure cases that FAUST raised in his review.

CI green, last rebase 6/29. Mind taking a look since it's your original
report? Happy to split or rescope if useful.

src/semantic-router/pkg/modeldownload/downloader.go
```diff
@@ -34,41 +55,55 @@ func DownloadModel(spec ModelSpec, config DownloadConfig) error {
...
-		strings.Contains(errStr, "authentication required")
-
-	// When no HF_TOKEN is set, the HF CLI streams its output directly to os.Stderr
-	// (not captured in err), so auth failures surface only as a plain "exit status 1".
-	// Treat any download failure without a token as a soft skip so the router
-	// degrades gracefully instead of crashing (e.g., CI forks without secrets).
-	noToken := hfToken == ""
-
-	return isKnownGated || isAuthError || noToken
+	return false
```
shraderdm, line 140: The reshape discussed in the review - anchored status phrases + tri-state allowlist fallback. Verified: compiles, flips both failing fixtures green, keeps all ten existing subtests green, `go vet` clean.

```suggestion
	haystack := strings.ToLower(err.Error() + "\n" + cliOutput)

	// Access-approval anchors the HF CLI / huggingface_hub emit for gated repos.
	// Status codes are anchored to their surrounding phrase ("403 forbidden", not
	// bare "403") so progress bytes like "403M/896M" and proxy/CDN denials on
	// PUBLIC repos are not misread as gated.
	// Note: 404 / "repository not found" is intentionally absent - a missing or
	// mistyped repo is a real error that must surface, not a graceful gated skip.
	gatedAnchors := []string{
		"gatedrepoerror",
		"gated",
		"restricted",
		"awaiting a review",
		"must be authenticated",
		"access to model",
		"access to this repo",
		"authentication required",
		"401 client error",
		"401 unauthorized",
		"403 forbidden",
		"you are trying to access a gated repo",
	}
	for _, p := range gatedAnchors {
		if strings.Contains(haystack, p) {
			return true
		}
	}

	// Tri-state fallback: if the CLI emitted output but nothing matched an auth
	// anchor, the failure is NOT gated (public-repo/transient) and must surface.
	// Only consult the known-gated repo allowlist when there is no captured
	// output to classify on (e.g. a bare "exit status 1").
	if strings.TrimSpace(cliOutput) != "" {
		return false
	}

	repoIDLower := strings.ToLower(repoID)
	for _, gatedName := range []string{"embeddinggemma", "gemma"} {
		if strings.Contains(repoIDLower, gatedName) {
			return true
		}
	}

	return false
```

shraderdm (changes requested):
Thanks for picking up #2107, @WUKUNTAI-0211 - and for the patience while I got to this one. The direction here is exactly right: dropping the `hfToken == ""` heuristic entirely is the correct answer to FAUST-BENCHOU's point (token state genuinely can't distinguish gated from rate-limited from typo'd), and the discrimination suite is a real upgrade - I reverted the classifier locally and exactly the five public-repo cases failed for the right reasons, so the table genuinely pins the fix.

### What's good

- The test table discriminates rather than illustrates: gated/auth vs public `429`/`404`/DNS, with real hfhub error strings instead of invented placeholders. Red/green verified locally.
- The `os/exec` plumbing is right: `io.MultiWriter(os.Stderr, &captured)` keeps live stderr while capturing, `tailWriter` is bounded at 64KiB with a correct `io.Writer` contract (full-length returns, so MultiWriter stays happy), and I confirmed the capture is race-free - one copier goroutine per stream, `Wait` joins it before classification reads.
- The signature change propagated cleanly (sole production caller + test, nothing else repo-wide), and `404`/"repository not found" is correctly absent from the gated patterns.
- `Closes #2107` is right, and keeping `IsModelComplete` out of scope was the right cut.

### The one thing I'd like closed before merge

**The classifier still soft-skips two public-repo failure shapes, which keeps #2107's masking alive in exactly the scenario that motivated it** - and since closing that case is this PR's whole job, I think it's worth one more pass (`downloader.go:87-88`, `:99-104`):

1. The bare `"401"`/`"403"` substrings match more than auth errors: tqdm progress prints byte counts like `403M/896M`, and a corporate proxy or CDN can 403 a fully public repo. Either reads as "gated" today.
2. The `embeddinggemma`/`gemma` fallback runs even when stderr is present and clearly non-gated, so a gemma-family repo hitting a `404`/`429` - or a public gemma-derived mirror - soft-skips too.

So you don't have to take my word for it, two fixtures (public repo, `want=false`) that fail against the current classifier:

```
cliOutput: "Downloading model.safetensors: 45%|xxxx| 403M/896M [00:12<00:14, 34.2MB/s]\n429 Client Error: Too Many Requests for url: https://huggingface.co/api/models/llm-semantic-router/mmbert-embed-32k-2d-matryoshka"
repoID:    "llm-semantic-router/mmbert-embed-32k-2d-matryoshka"   -> got gated=true, want false

cliOutput: "429 Client Error: Too Many Requests for url"
repoID:    "some-org/gemma-derived-public-mirror"                 -> got gated=true, want false
```

The first one is the #2107 GKE scenario itself - a public download that made progress and then got rate-limited.

Both flip green under one small reshape, which I verified compiles and keeps all ten of your existing subtests green: anchor the status codes to their phrases, and make the allowlist a true last resort (consulted only when there's no captured output to classify). I've attached it as an inline suggestion on `downloader.go` so it's one-click applyable - take it as-is or reshape it however you prefer.

Adding those two cases (plus a real multi-line `GatedRepoError` stderr) to the table would pin the false-positive path for good.

### Small stuff (none blocking)

- nit: the loud branch at `downloader.go:160` returns the bare `err` ("exit status 1") and drops `captured.String()`, which is holding the actual `429`/`404`/`dial tcp` text - since surfacing the cause is the point of the change, consider interpolating a bounded tail of it into the returned error.
- nit: the doc comment at `downloader.go:62` says "captured stdout/stderr" but only stderr is captured (`cmd.Stdout` stays on `os.Stdout`).
- nit: the workflow comment at `test-and-build.yml:228-229` ("gated models will gracefully skip") is now stale for public transient failures - worth a one-line touch-up here since this PR changes those semantics.
- nit: the body reads well but the template wants Purpose / Test Plan / Test Result headings - a `## Test Result` block with the green run would make the record self-contained.

### One design question for the thread (no code change implied)

With the blanket skip gone, the unauthenticated fork-CI lane (which has the lowest HF rate limits) now hard-fails router startup on a transient public `429` where it used to sail through - and a running router's file-source hot-reload similarly aborts on a transient blip (it keeps serving the old config, so that half is arguably safer). That's the intended direction of #2107, but the retry posture is worth a deliberate call: bounded retry/backoff on downloads, or a distinct retryable outcome for transient `429`s? Happy to file it as a follow-up either way.

### Verdict

Marking request-changes for the classifier gap only, since merging as-is would close #2107 while its motivating scenario still misclassifies - everything else here is good work, and the fix is one function with the patch above ready to go. No rush on my end; whenever it's ready, I'll re-review promptly. Thanks again for taking this on - between this and #2104 you've carried a big chunk of the multi-modal arc's plumbing.

WUKUNTAI-0211:
Thanks @shraderdm — applied the reshape, and it's a clear improvement. Both of your fixtures now classify correctly. Summary of what changed:

**Classifier (the blocking gap) — `downloader.go`**
- Took your inline suggestion: status codes are now anchored to their phrase (`401 client error` / `401 unauthorized` / `403 forbidden`) plus explicit `gatedrepoerror` / `you are trying to access a gated repo` anchors, so `403M/896M` progress bytes and proxy/CDN 403s on public repos no longer read as gated.
- The `embeddinggemma`/`gemma` allowlist is now a true last resort — consulted only when there's no captured CLI output. A gemma-named public mirror hitting a transient 429/404 surfaces instead of soft-skipping.

**Tests — `downloader_test.go`**
- Added your two `want=false` fixtures verbatim: the #2107 GKE progress-then-429 case and the gemma-derived public mirror.
- Added a real multi-line `huggingface_hub.errors.GatedRepoError` stderr (`want=true`) so the gated path stays pinned alongside the false-positive path.
- All 13 subtests green locally, `go vet` clean.

**Nits (all taken)**
- The loud branch now interpolates a bounded (2KiB) tail of the captured stderr into the returned error, so the operator sees the real `429`/`404`/`dial tcp` cause instead of a bare `exit status 1` (new `tailForError` helper, empty-safe).
- Fixed the `downloader.go:62` doc comment — only stderr is captured; stdout streams through untouched.
- Refreshed the `test-and-build.yml` comment to reflect that public-repo transient failures now surface rather than skip.
- PR body updated with a `## Test Result` block.

**Design question (retry posture):** agree it's worth a deliberate call and out of scope here — I'll file a follow-up for bounded retry/backoff on transient `429`s (distinct retryable outcome vs hard-fail), covering both the fork-CI startup lane and the file-source hot-reload path. Happy to take it.

Rebased onto current `main`. Ready whenever you get a chance to re-review.

WUKUNTAI-0211:
Heads up — the first CI run after the reshape caught a real gap, now fixed.

Making the repoID allowlist a last resort exposed that the Rust `hf` CLI emits gated repos as:

```
Error: Access denied. This repository requires approval.
```

— not an HTTP status line and not a `GatedRepoError`. With the allowlist demoted, `google/embeddinggemma-300m` (which the `test-and-build` job downloads unauthenticated) no longer matched any anchor, so it classified as public → hard-failed startup (`1 out of 12 models`). Exactly the tri-state behaviour we wanted, just missing the anchor for the newer CLI's phrasing.

Fix: added `access denied` / `requires approval` to the gated anchors, and pinned the real embeddinggemma CLI string as a fixture (stderr present + `gemma` repoID, so it proves the anchor catches it rather than the allowlist). `test-and-build` should be green now.

Net: 15 subtests, `go vet` clean, rebased onto latest `main` (after #2104 merged).

shraderdm (changes requested):
Thanks for turning this around so thoroughly, @WUKUNTAI-0211. This round addresses the earlier ask exactly: the `hfToken == ""` blanket soft-skip is gone, the status codes are phrase-anchored, the repo-name allowlist is demoted to a last resort, and the two public-repo false-positive fixtures (the `403M/896M` progress bytes and the gemma-named public mirror) are pinned. The core direction of #2107 is sound now: a 429, a `dial tcp` failure, or a network blip carries its real cause through to a loud failure instead of being masked.

While re-reviewing I ran the current classifier against the real `huggingface_hub` error strings (v1.19.0), and one gated anchor turns out to re-open the same masking hole on a different shape. Details below, with a repro so you do not have to take my word for it.

### What's good

- The tri-state control flow is the right shape: anchors first, "stderr present but unmatched means not gated," allowlist only when there is nothing to classify. That is a real improvement over the token heuristic.
- Capturing stderr through `io.MultiWriter(os.Stderr, &captured)` with a bounded `tailWriter` keeps the live stream while retaining the tail, and surfacing that tail via `tailForError` in the loud branch is a nice touch for operators.
- The `403 Forbidden` vs `403M/896M` distinction from the last round is correctly handled, and the gemma-mirror last-resort demotion works.

### Findings

#### Critical (blocking)

**The bare `"gated"` anchor re-masks the missing/mistyped-repo case the PR is meant to surface** (`downloader.go:103`). The real `huggingface_hub` 404 message for a missing or typo'd repo contains this line (verified against the installed v1.19.0 source, `utils/_http.py`):

> Please make sure you specified the correct `repo_id` and `repo_type`. If you are trying to access a private or **gated** repo, make sure you are authenticated ...

Lowercased, that hint line contains the substring `gated`, so `IsGatedModelError` returns `true` and a mistyped or nonexistent public repo is soft-skipped as gated instead of failing loudly. That is the exact class this PR (and the in-code comment at `downloader.go:99-100`, "404 ... must surface, not a graceful gated skip") sets out to surface. The `"missing repo 404 is not gated"` test at `downloader_test.go:69-71` passes only because its fixture (`"404 Client Error: Repository Not Found for url"`) omits the real hint line the library always appends, so the suite stays green while the real shape misclassifies. It also undercuts the `test-and-build.yml:230` comment's promise that 404s "now surface."

Repro against the current code (real 404 body, public repo, expected `want=false`):

```
cliOutput: "404 Client Error.\n\nRepository Not Found for url: https://huggingface.co/api/models/typo/does-not-exist.\nPlease make sure you specified the correct `repo_id` and `repo_type`.\nIf you are trying to access a private or gated repo, make sure you are authenticated and your token has the required permissions."
repoID:    "typo/does-not-exist"
=> IsGatedModelError returns true (want false)
```

The root cause is that the bare `gated`/`restricted` anchors are also what catch *real* gated repos (the `GatedRepoError` message is "Cannot access gated repo for url ... Access to model X is restricted", `_http.py`), so they do double duty and the 404 hint line collides with them. The bare `"restricted"` anchor (`downloader.go:104`) has the same problem from a different angle: a public repo whose URL merely contains the substring (e.g. `.../unrestricted-model`, or `.../elongated-encoder` for `gated`) flips `true` on any failure. A direction that keeps real gated detection without either collision: drop bare `"gated"` and `"restricted"` and anchor on the gated-specific phrases the library actually emits, e.g. `"cannot access gated repo"`, `"access to model"`, and `"is restricted and you are not"`, alongside the existing `"gatedrepoerror"`, `"you are trying to access a gated repo"`, and `"requires approval"`. Those match the real gated bodies but not the 404 hint or a substring in a public repo URL. Worth confirming the exact phrasings against both the Python and Rust CLI outputs before settling the list.

#### Important

**Trim or tighten the status-code / access-denied anchors** (`downloader.go:112,114,116`). Because real gated repos are already caught by the phrase anchors above, two of these can be dropped cleanly:
- `"403 forbidden"` matches a CDN/proxy 403 on a public repo (HF LFS/Xet blobs and CloudFront both emit `403 Forbidden`); its gated test case double-matches other anchors, so removing it loses no real detection.
- `"access denied"` matches an S3/CloudFront `Access Denied` body or a corporate proxy block page; the genuine Rust-CLI gated phrasing is already covered by `"requires approval"` (the hf-CLI access-denied case in the table already double-matches).

`"401 client error"` is ambiguous rather than cleanly removable: it currently catches both a private/gated-repo 401 (the `captured 401 is gated` case at `downloader_test.go:41`, which would flip to a loud failure if the anchor were just deleted) and a bad-or-expired token on a *public* repo (`_http.py` yields "401 Client Error. Invalid credentials in Authorization header", explicitly not a repo-not-found), which currently gets a misleading "set HF_TOKEN" warning when a token is already set. Worth a deliberate call on how to separate those two 401s rather than a straight removal.

**The test table cannot falsify the over-match** (`downloader_test.go`). Every `want:false` case is a string that contains no anchor substring at all, and the risky positive cases each match multiple anchors, so deleting `"403 forbidden"` or `"access denied"` individually leaves the whole table green. The cases most likely to be wrong have no coverage. Adding negatives whose text *contains* an anchor substring would pin this: the real 404 body above, a bad-token 401 on a public repo, and a public repo URL containing `gated`/`restricted`.

#### Minor / nit

- nit: the gemma repo-name allowlist (`downloader.go:134`) fires on *any* empty-stderr failure, so an OOM-killed download (`err = "signal: killed"`, no captured stderr) on a gemma-named repo is reported as a gated skip even with a valid token. Gating the allowlist on the error not being a process signal, or just noting it as a known corner, would close it.
- nit: the body uses `## Summary`; the PR template asks for `## Purpose` / `## Test Plan` / `## Test Result` (same note as last round, no rush).

### Verdict

Requesting changes on the Critical only. The #2107 fix is genuinely close and the control-flow redesign is right; the one blocker is that the bare `gated`/`restricted` anchors re-bury the 404/missing-repo case the PR claims to surface, and the suite stays green because the 404 fixture omits the real hint line. Item 1 is the fix; items 2-3 are the cleanup that makes it robust, and the minor items are optional. Happy to re-review quickly whenever it is ready, and thanks again for carrying this one.

shraderdm (approved):
Correction to my earlier review, @WUKUNTAI-0211, and an apology for the noise: I re-ran this against the actual `hf` CLI and my blocking concern does not hold. Retracting it and approving.

I flagged that the bare `"gated"` anchor would re-mask a missing or mistyped repo, since the `huggingface_hub` 404 message contains "private or gated repo". That is true of the library's exception string but not of what the CLI writes to stderr, which is what this classifier reads. The `hf` CLI reformats `RepositoryNotFoundError` through `cli/_errors.py` into `Error: Model '<id>' not found. If the repo is private, make sure you are authenticated...` with no "gated" substring, and only emits the raw exception under `HF_DEBUG=1`, which the downloader does not set. Running `hf download` on a nonexistent repo confirms it: no anchor matches, `IsGatedModelError` returns false, the download fails loudly. So the 404 path and its test fixture are correct. I reasoned from the exception source instead of the CLI's real output - my mistake.

The rest holds up well: the token heuristic is gone, the control flow is right, and gated repos are still caught because the CLI renders `GatedRepoError` as `Access denied. ... requires approval.`, which your anchors match. Approving.

One optional, non-blocking note for a future pass: generic HTTP failures raise `HfHubHTTPError`, whose CLI formatter returns the raw `str(error)` (unlike the not-found and gated paths), so `"403 forbidden"`, `"401 unauthorized"`, and the bare `"gated"`/`"restricted"` anchors can still fire on a public-repo CDN/proxy 403, or a rate limit on a repo whose name contains "gated"/"restricted". Narrow, and not worth holding the PR for. Thanks again.

Xunzhuo (changes requested):
The failure classifier is still too broad: matching bare `access denied` or `requires approval` anywhere in downloader stderr also classifies public-model proxy/CDN failures as gated. That soft-skips the real download failure this change is meant to expose. Please require a trusted Hugging Face gated-model signature (or an equivalently specific classifier) and add a public proxy/CDN denial regression.

Issue #1760, opened 2026-04-10
feature: add projection trace and explainability surfaces across replay and dashboard


Xunzhuo:
## Describe the feature
Add full projection trace and explainability surfaces across runtime, replay, and dashboard insights so projection behavior is as debuggable as the routing power it adds.

## Primary layer
global level

## Why this layer?
This work crosses core runtime diagnostics, replay and observability payloads, and dashboard presentation. The value comes from end-to-end traceability, not one isolated code path.

## Why do you need this feature?
The projection layer already coordinates routing outcomes, but its internal reasoning is still only partially visible. For projection-driven routing to scale, operators need to see not just that an output fired, but why.

Useful diagnostics include:
- per-score contribution breakdowns
- mapping boundary distances
- partition winner margins
- calibrated mapping confidences
- projection inputs and intermediate score values preserved in replay and surfaced in dashboard insights

Without that, richer projection behavior becomes harder to debug and trust.

### Affected repo surfaces
- `src/semantic-router/pkg/classification/classifier_projections.go` and `src/semantic-router/pkg/classification/classifier_signal_eval.go` for trace capture
- `src/semantic-router/pkg/extproc/request_context.go` plus replay and eval plumbing that already preserves matched signals and projection outputs
- `dashboard/frontend/src/pages/insightsPageSupport.tsx` and `dashboard/frontend/src/pages/ConfigPageProjectionsSection.tsx` for operator-facing rendering
- projection docs and maintained examples that explain how to inspect traces

### Suggested starting points
- Start by defining one structured projection-trace payload that can be preserved end to end.
- Capture a small set of high-value diagnostics first, such as score contributions and mapping confidence details.
- Make replay and dashboard rendering consume the same structured shape instead of inventing separate debug formats.

### Validation commands
```bash
[...]

PR #2111, opened 2026-06-10
[Feature] Add matched outputs to the projection trace

xiaotian-yu:
Follow-up to #1760

## Purpose

- This follow-up covers trace/debug completeness for multi-emit mappings, which was not included in the earlier change.
- Extend projection trace mapping decisions to record `matched_outputs` so multi-emit projection behavior is preserved in replay/debug data.
- Keep `selected_output` for backward compatibility while exposing the full emitted output set.
- Surface `matched_outputs` in the Dashboard Insights projection trace UI.
- Affects: `Router`, `Dashboard`

## Test Plan

- Run `go test ./pkg/classification -run 'TestApplyProjections(AddsDerivedOutputsAndScores|MultiEmitEmitsAllMatchingBands)$' -count=1`
- Run `go test ./pkg/routerreplay/store -run TestRecordJSONRoundTripProjectionTrace -count=1`
- Run `make test-semantic-router`
- Run `env GOCACHE=/tmp/gocache GOLANGCI_LINT_CACHE=/tmp/golangci-lint-cache make dashboard-check`
- Run the local smoke flow with `make agent-dev ENV=cpu`, `make agent-serve-local ENV=cpu`, and `make agent-smoke-local`

## Test Result

- Projection trace classification tests passed, including the multi-emit coverage for `matched_outputs`.
- Projection trace replay/store round-trip test passed.
- `make dashboard-check` passed.
- Local CPU smoke flow passed with router, envoy, dashboard, and fleet-sim all healthy.


AayushSaini101:
Thanks @xiaotian-yu can you please resolve the conflict thanks

xiaotian-yu:
> Thanks @xiaotian-yu can you please resolve the conflict thanks

Conflict resolved. Can you please take another look?

Xunzhuo (changes requested):
The new projection trace appends every overlapping condition to `MatchedOutputs` regardless of mapping method. For `threshold_bands` and the empty method, runtime emits only the first match, so replay and Dashboard report outputs that were never emitted. Please make tracing method-aware and assert the single first-hit output in both overlapping-band tests.

FAUST-BENCHOU (commented):
Thanks for adding `matched_outputs` to the projection trace. This makes multi-emit behavior much easier to debug and keeps the trace representation aligned with the actual projection matching result.

One small thing to consider: should we explicitly document that the order of `matched_outputs` follows the projection output declaration order? Keeping this ordering deterministic would make replay comparison and dashboard rendering more stable when multiple outputs match.

PR #2113, opened 2026-06-10
[Router] Reject unresolved models with a clear 4xx instead of a misleading 401

### LINKED ISSUE #2112: [Router] Unknown/unresolved models return a misleading "401 No api key" instead of a clear 4xx
## Summary

When the router cannot resolve a request to a configured model, it forwards the request upstream **without injecting a backend credential**. The upstream then rejects it with a misleading `401 "Authentication Error, No api key passed in"` instead of a clear client-side error. The 401 wrongly implies an auth/key problem when the real cause is an unknown model (or an auto request that selected no model).

This happens in two distinct situations that share the same root cause.

## Reproduction

Router configured with model `qwen3-122b` (alias) backed by an OpenAI-compatible endpoint. Requests go through Envoy → ext_proc (router) → backend.

| Request `model` | Expected | Actual |
|---|---|---|
| `qwen3-122b` (configured) | 200 | 200 ✅ |
| `totally-fake-model` (typo / unknown) | 400 "model not found" | **401 No api key** |
| backend's real id e.g. `Qwen/Qwen3.5-...` (not a configured router name) | 400 or route | **401 No api key** |
| `QWEN3-122B` (wrong case) | 400/route | **401 No api key** |
| `" qwen3-122b "` (surrounding whitespace) | route | **401 No api key** |
| `MoM` with `messages: []` (auto, no user query) | clear 4xx | **401 No api key** |

The 401 body is the upstream's, e.g.:

```json
{"error":{"message":"Authentication Error, No api key passed in.","type":"auth_error","code":"401"}}
```

## Root cause

Two code paths forward an **unresolved** model without a credential:

1. **Specified (pinned) model not in config** — `pkg/extproc/processor_req_body_routing.go` `resolveBackendForModel` calls `ResolvePrimaryBackendForModel`, which returns `found=false, err=nil` for an unknown model (`config/helper.go:528`). `resolveBackendForModel` ignores `backendFound`, so `handleSpecifiedModelRouting` (`processor_req_body.go:227`) proceeds with no backend and no credential.

2. **Auto model with no selection** — `handleModelRouting` `default:` branch (`processor_req_body.go:137-141`) sets `ctx.RequestModel = originalModel` and returns `CONTINUE`, forwarding the unresolved auto-model name (`MoM`) untouched with no credential.

In both cases the request reaches the backend (via Envoy's default route) without an `Authorization` header → upstream `401`.

## Expected behavior

- A specified model that is not configured should be rejected with a clear **400** (e.g. `model "X" is not available`).
- An auto model that selects nothing should **fall back to the configured `default_model`**; if none is configured, return a clear **400**.
- Surrounding whitespace in the requested model name should be trimmed.

## Scope / related

- A fix for the above (B1 + B2 + trim) is in the linked PR.
- Separately (not in this issue's PR): the legacy `POST /v1/completions` endpoint returns `404 "endpoint not found"`. That is a feature gap (chat-only support), tracked separately.

## Environment

- Component: `src/semantic-router` (Router / ext_proc)
- Reproduced against a local split router + Envoy pointed at an OpenAI-compatible (LiteLLM) backend.

theohsiung:
## What

When the router cannot resolve a request to a configured model, it forwarded the request upstream **without a backend credential**, so the upstream returned a misleading `401 "No api key passed in"` instead of a clear client error.

This PR makes the router fail with a clear `4xx` instead:

- **Unknown specified model** — `handleSpecifiedModelRouting` now rejects a model with no configured backend with `400 "model \"X\" is not available"` before forwarding (previously it was silently forwarded with no credential).
- **Auto model with no selection** — when an auto (`MoM`) request yields no model selection (e.g. empty/contentless messages), `handleModelRouting` now falls back to the configured `default_model`; if none is configured it returns `400`. Previously it forwarded the unresolved `MoM` name untouched.
- **Whitespace** — the requested model name is trimmed before routing.

Fixes #2112.

## Why

The `401 "No api key"` body wrongly implies an authentication/key problem. The real cause is an unknown model (typo, wrong case, surrounding whitespace, or using a backend's real id that isn't a configured router name) or an auto request that selected nothing. Returning a clear `400` is much easier to diagnose and matches OpenAI-compatible client expectations.

## Behavior change

A request pinning a model that is not in `providers.models` now receives `400 "model \"X\" is not available"` instead of being forwarded to a backend (which previously surfaced as a `401`). The existing edge-case test `should handle requests with invalid model names`, which asserted the old silent-passthrough (`CONTINUE`), was updated to assert the new `400`.

## Test plan

- [x] Added `pkg/extproc/model_resolution_test.go`:
  - unknown specified model → `400`
  - known configured model → routes (sets `x-selected-model`)
  - auto model + no selection + `default_model` set → falls back to default
  - auto model + no selection + no `default_model` → `400`
- [x] Updated existing `should handle requests with invalid model names` test to expect `400`.
- [x] `go test ./pkg/extproc/` — full package green (160 specs + new tests).
- [x] `go vet ./pkg/extproc/` clean; `gofmt` clean; `go build -tags=milvus ./pkg/...` and router binary build green.
- [x] Reproduced the original `401` and confirmed the fix returns `400` for each row of the table in #2112 (unit-level).

## Notes

- All commits are DCO signed-off.
- The legacy `POST /v1/completions` `404` mentioned in #2112 is a separate feature gap and intentionally out of scope here.


AayushSaini101 (approved):
LGTM thanks a lot : )

PR #2115, opened 2026-06-10
[Router] Validate decision modelRefs and reject duplicate decision names

### LINKED ISSUE #2114: [Router] Config validator accepts malformed routing.decisions (unknown modelRef, duplicate names, bad condition types)
## Summary

Config fuzzing of `routing.decisions` surfaced several validation gaps: the canonical config validator accepts malformed decision references that should be rejected at load time. Two of them defer a failure to request time (with the misleading `401 No api key` from #2112) or silently produce dead/ambiguous routing.

## Findings (reproduced on a running router)

All of the following configs **start up successfully** today:

| # | Malformed config | Expected | Actual |
|---|---|---|---|
| **G1** | a decision `modelRef.model` not declared in `routing.modelCards` | reject at load | accepted → at request time routes to the unknown model → misleading **`401 No api key`** |
| **G2** | two decisions with the **same name** | reject (ambiguous) | accepted; higher-priority duplicate silently wins |
| **G3** | condition `type:` not a known signal type (e.g. `type: bogus`) | reject | accepted; condition never matches → silently dead decision |
| **G4** | condition `name:` referencing an undefined signal | reject/warn | accepted; condition never matches → silently dead decision |

By contrast, `providers.defaults.default_model` and `routing.modelCards` duplicate names **are** validated — so the decision surface is inconsistent.

### Root cause (G1, G2)

In `pkg/config/canonical_config.go` `validateCanonicalContract`:
- The modelRef existence check (`references unknown model`) is inside a branch guarded by `if modelRef.Model == "" || modelRef.LoRAName == "" { continue }`, so a plain modelRef (no LoRA) skips it entirely.
- There is a duplicate check for `modelCards` names but none for decision names.

## Scope

A PR fixes **G1 + G2** (zero false-positive risk; reuses the existing `default_model` existence helper, gated on a declared model surface so partial DSL fragments under `config/decision/**` are unaffected).

**G3 / G4 are intentionally deferred** because they need design first:
- G3: a naive check against `supportedSignalTypes` would reject valid configs — `type: projection` is used in `config/config.yaml` and several recipes but is **not** in that list (also `session_metric` / `event_context`). The allow-list needs correcting before condition-type validation is safe.
- G4: many signal types use implicit/fixed names (domain categories, `pii`/`jailbreak`/`fact_check` outputs) not declared under `routing.signals`, so validating condition names needs a per-type name registry to avoid false positives.

## Related

Same misleading-401 failure class as #2112 / #2113 (that PR fixes the runtime ext_proc side; this is the config-load side).

theohsiung:
## What

Two `routing.decisions` validation gaps (found via config fuzzing) are closed at config-load time:

- **G1 — unknown modelRef model.** A decision `modelRef.model` not declared in `routing.modelCards` is now rejected at load. Previously the existence check only ran when a `lora_name` was also present, so a plain modelRef to an unknown model passed validation and only failed at request time with the misleading `401 No api key` (same failure class as #2112).
- **G2 — duplicate decision names.** Now rejected, mirroring the existing duplicate-`modelCards` check. Previously accepted silently, with the higher-priority duplicate winning.

Fixes #2114.

## Implementation

In `pkg/config/canonical_config.go` `validateCanonicalContract`:
- Split the modelRef guard so model existence is checked independently of `lora_name`, using the same `canonicalModelOrLoRAExists` helper that validates `default_model` (so LoRA aliases still resolve).
- The existence check is **gated on a non-empty `modelCards`**, so partial DSL fragment examples under `config/decision/**` (which declare no model surface) are unaffected.
- Added a duplicate-decision-name check alongside the existing decision loop.

## Behavior change

A config whose decision `modelRef.model` (or a duplicate decision name) is invalid now fails fast at load with a clear error instead of starting and misbehaving at request time. Verified against all repo configs (canonical, `deploy/recipes/**`, `e2e/config/**`) — 32 configs with `decisions:` — none regress.

## Test plan

- [x] Added `pkg/config/decision_reference_validation_test.go`:
  - unknown modelRef model → rejected (error names the model)
  - known modelRef model → accepted
  - duplicate decision names → rejected (error names the decision)
  - distinct decision names → accepted
  - sanity: unknown `default_model` still rejected (confirms the validation path)
- [x] `go test ./pkg/config/` full suite green; `go vet` + `gofmt` clean; `go build -tags=milvus ./pkg/config/` green.
- [x] Ran a temporary scan parsing every repo config containing `decisions:` to confirm zero false positives (the `config/decision/**` fragments are correctly skipped via the modelCards gate).

## Out of scope

G3 (unknown condition `type`) and G4 (undefined condition signal `name`) from #2114 are deferred: the `supportedSignalTypes` allow-list is missing `projection`/`session_metric`/`event_context` (used by real configs), and implicit signal names need a per-type registry — both need design before they can be validated without false positives.

## Notes

- DCO signed-off. Separate subsystem (config validation) from #2113 (ext_proc runtime), so a separate PR.


AayushSaini101 (approved):
LGTM thanks

PR #2119, opened 2026-06-10
[Router] Redact resolved upstream credentials from JSON config introspection

### LINKED ISSUE #2118: [Security] GET /info/classifier exposes resolved upstream API keys in its JSON response
## Summary

`GET /info/classifier` serializes the entire live runtime config and includes **resolved upstream API keys** in its (unauthenticated) JSON response. When credentials are configured indirectly via `api_key_env`, the on-disk config and `GET /config/router` correctly expose only the env-var *name* — but `/info/classifier` exposes the *resolved secret value*.

## Reproduction

Config uses `api_key_env` (no inline secret on disk):
```yaml
providers:
  models:
    - name: m1
      backend_refs:
        - endpoint: 10.0.0.1:8000
          api_key_env: LLM_KEY     # env reference only
```

```
# leaks the RESOLVED value of $LLM_KEY:
$ curl -s http://<router>:8080/info/classifier | grep -o '"APIKey":"[^"]*"'
"APIKey":"<resolved-secret>"
$ curl -s http://<router>:8080/info/classifier | grep -o '"AccessKey":"[^"]*"'
"AccessKey":"<resolved-secret>"

# the dedicated config endpoint does NOT leak it (serves raw YAML):
$ curl -s http://<router>:8080/config/router | grep -oE '"(APIKey|AccessKey)":'   # -> no output
```
The same response also dumps all upstream addresses, model/endpoint mappings, KB layout, and tracing endpoints.

## Root cause

- `api_key_env` is resolved at load and stored inline in the runtime config (`pkg/config/canonical_global.go` `resolveBackendAPIKey` → `os.Getenv`), populating `VLLMEndpoint.APIKey` and `ModelParams.AccessKey`.
- Those credential fields carried **only a `yaml` tag, no `json` tag**: `pkg/config/model_config_types.go` (`VLLMEndpoint.APIKey`, `ModelParams.AccessKey`, and the classifier-endpoint `AccessKey`).
- `handleClassifierInfo` dumps the whole live config with no redaction via `jsonCompatibleValue(s.currentConfig())` (`pkg/apiserver/route_model_info.go:27-42`).
- `json_compat` falls back to the Go field name for any field lacking a `json` tag and emits its value; only `json:"-"` is skipped (`pkg/apiserver/json_compat.go:126-134`). So the secrets serialize as-is.

## Impact

- Plaintext disclosure of upstream provider credentials via an introspection endpoint, defeating the `api_key_env` indirection.
- Same disclosure class as the previously-addressed #1450.

## Fix

Add `json:"-"` to the three resolved-credential string fields so no JSON response path can emit them (yaml tags unchanged, so config load / hot-reload / `GET /config/router` are unaffected). PR attached.

## Note (not in this PR)

`ProviderProfile.ExtraHeaders` is a free-form `map[string]string` that *could* carry a secret header value; worth a follow-up redaction policy, but it is not the resolved-credential leak fixed here.

theohsiung:
## What

`GET /info/classifier` serialized the entire live runtime config and leaked **resolved upstream API keys** in its JSON response. Credential fields populated by resolving `api_key_env` (`VLLMEndpoint.APIKey`, `ModelParams.AccessKey`, and the classifier-endpoint `AccessKey`) carried only a `yaml` tag, so `json_compat` fell back to the Go field name and emitted the resolved secret — even though the on-disk config stores only an `api_key_env` reference and `GET /config/router` (raw YAML) never exposes the value.

Fixes #2118.

## Fix

Add `json:"-"` to the three resolved-credential string fields in `pkg/config/model_config_types.go`. `json_compat` honors `json:"-"` (`jsonFieldMetadata` → skip), so the secrets are now excluded from every JSON response path. `yaml` tags are unchanged, so config load, hot-reload, and `GET /config/router` are unaffected.

## Test plan

- [x] Added `pkg/apiserver/info_classifier_redaction_test.go`: builds a config whose `VLLMEndpoint.APIKey` / `ModelParams.AccessKey` hold a sentinel secret, calls `handleClassifierInfo`, and asserts the secret never appears in the response body. RED before the fix (leaked), GREEN after.
- [x] `go test ./pkg/apiserver/ ./pkg/config/` — full suites green (no regression).
- [x] `gofmt` clean.

## Notes

- DCO signed-off.
- `ProviderProfile.ExtraHeaders` (free-form `map[string]string`) could carry a secret header value — flagged in #2118 as a possible follow-up redaction policy, intentionally out of scope of this minimal credential-field fix.


theohsiung:
Hi @AayushSaini101 !

this one (#2119) is the same credential-leak class as the just-merged #2133: it redacts the resolved upstream key (api_key_env expanded into VLLMEndpoint.APIKey / ModelParams.AccessKey) from the GET /info/classifier introspection response, via json:"-" on those fields. Small diff (+47/-3) with a regression test. Since you've got the context from #2133, would you want to take this one too? No pressure
thanks～～～🙏

AayushSaini101 (approved):
LGTM thanks : )

theohsiung:
> LGTM thanks : )

Thx~

Issue #2120, opened 2026-06-10
[Bug] Return typed client and not-ready errors from classification APIs


theohsiung:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

Classification endpoints can return HTTP 500 for invalid client input and model-not-ready conditions, and related endpoints do not map the same failure consistently.

### To Reproduce

Use the reproduction or current-code evidence documented in this issue to exercise the affected path.

### Expected behavior

Client validation, unsupported capability, not-ready, timeout, and internal failures should have stable typed error codes and consistent HTTP mappings.

### Impact

Not separately specified in the original report.

##### Completion signal

Contract tests cover the supported classification endpoints and prove that client and readiness failures are not reported as generic internal errors.

Parent Epic: #2782

theohsiung:
Hey @p172913, thanks for picking up #2120. Quick heads-up so we don't duplicate effort: #2121 already covers part (A) (empty/whitespace → 400). Would you be up for reviewing it? And if you're interested, part (B) (typed not-ready error → 503) is still unclaimed — I'd be glad to hand that off or pair on it. Linking the PR to the issue now so it's easy to track.

p172913:
@theohsiung Happy to work on part (B) (typed not-ready error → 503)

p172913:
@theohsiung I raised a PR for this issue. Can you please give approval for those changes.

PR #2121, opened 2026-06-10
[Router] Return 400 (not 500) for empty/whitespace classify input

theohsiung:
## What

The single-text classify endpoints returned `500 CLASSIFICATION_ERROR` for empty / whitespace-only input, contradicting their documented OpenAPI contract ({200, 400}) and behaving inconsistently with sibling endpoints (combined / batch / embeddings already return 400). Whitespace-only input also slipped past the `== ""` checks.

Addresses part **(A)** of #2120.

## Fix

- Add a typed `services.ErrEmptyText` sentinel + a `blankText()` (TrimSpace-aware) helper.
- Return `ErrEmptyText` from the empty-text checks in `DetectPII`, `CheckSecurity`, intent extraction, `ClassifyFactCheck`, and `ClassifyUserFeedback` — now whitespace-aware.
- Map it in the classify handlers via a shared `writeClassificationError()`: `ErrEmptyText → 400 INVALID_INPUT`, anything else → `500 CLASSIFICATION_ERROR` (unchanged).

## Test plan

- [x] Added `pkg/apiserver/classify_empty_input_test.go`: empty and whitespace-only input → `400 INVALID_INPUT` for intent / pii / security; non-empty input still → `200`. RED before fix (empty → 500, whitespace → 200), GREEN after.
- [x] `go test ./pkg/apiserver/ ./pkg/services/ ./pkg/config/` — all green (no regression).
- [x] `go vet` + `gofmt` clean.

## Out of scope (deferred in #2120)

Part **(B)** — "feature disabled / model not loaded → 500" (pii/security/embeddings) — should arguably be `503` (not-ready), but the failing call returns a generic error that can also mean a genuine runtime failure. Distinguishing not-ready from a real 500 needs a typed unavailable error / readiness probe from the classifier/embedding layer (a maintainer design decision), so it is intentionally not included here. Blanket-mapping to 503 would mask real internal errors.

## Notes

- DCO signed-off.


AayushSaini101 (approved):
LGTM thanks : )

PR #2123, opened 2026-06-10
[Router] Validate the decision rule-tree contract at config load

### LINKED ISSUE #2122: [Bug] Reject unsupported decision-rule operators during config validation
### Proposed Workgroup

MoM & Routing

### Describe the bug

Reject unsupported decision-rule operators during config validation.

### To Reproduce

### Expected behavior

- An omitted operator keeps the documented AND default.
- A non-empty operator must be `AND`, `OR`, or `NOT` (case handling should match the documented config contract).
- `NOT` must have exactly one child.
- Invalid operators fail config validation before runtime.

### Impact

##### Audited upstream baseline (2026-07-15)

Decision rules use a `RuleNode` tree. In `src/semantic-router/pkg/classification/classifier_composers.go`, runtime evaluation handles `OR` and unary `NOT`; every other value, including an omitted operator and typos such as `XOR` or `ADN`, falls through to AND semantics.

The config validation path in `src/semantic-router/pkg/config/validator_decision.go` validates model, algorithm, workflow, output, and plugin contracts but does not reject unsupported rule-tree operators. The DSL emits the supported operators, but hand-written YAML can still load an invalid non-empty value.

##### Acceptance criteria

- [ ] File and Kubernetes/canonical config paths enforce the same rule.
- [ ] Errors include the decision name and rule-tree location.
- [ ] Tests cover omitted/AND/OR/NOT, invalid values, nested nodes, and NOT arity.
- [ ] Runtime and validation share or demonstrably agree on the supported operator set.
- [ ] Maintained examples and DSL output continue to validate.

##### Relevant surfaces

`src/semantic-router/pkg/config/validator_decision.go`, `src/semantic-router/pkg/classification/classifier_composers.go`, DSL compilation, reference config, and tests.

Parent Epic: #3202

theohsiung:
## What

`routing.decisions[].rules` is a tree of leaf (signal reference) or combination (`operator` + `conditions`) nodes. Validation checked neither the operator nor the node shape, so trees the evaluator cannot honour loaded silently and meant something other than what was written:

| Rule tree | What the router actually does |
|---|---|
| `operator: XOR` — any unrecognized value | `evalNode`'s default branch is **OR**, so the rule is silently widened |
| `NOT` with 0 or 2+ children | warns and never matches |
| node with both `type` and `operator` | `IsLeaf()` wins and `conditions` are dropped |
| node with `name` but no `type` | empty `OR` — never matches |
| leaf without a `name` | references no signal — never matches (the full-config path caught this, the routing-fragment path did not) |
| leaf whose `type` is not a supported signal type | same split: rejected on the full-config path, accepted on the routing-fragment and canonical paths |
| node with only `label`/`predicate`/`on_error` | also satisfies `IsEmpty()` — **matches every request** |
| nested combination with no `conditions` | never matches under `OR`, matches everything under `AND` |

All eight now fail config load, naming the decision and the node: `decision 'x': rules.conditions[0]: ...`. Fixes #2122.

> Decision rules are evaluated by `pkg/decision.evalNode`, whose default branch is **OR** — not by `classification.evalComposerNode` (default AND) as this issue's description says. An unknown operator therefore widens the rule rather than narrowing it.

**Blast radius:** every rule tree under `config/**`, `deploy/**`, `e2e/**`, `src/**`, `website/**` — 873 decisions — still loads. One Go fixture did not: it declared a fallback as a childless `OR`, which never matches. Fixed here.

**Release note:** an `IntelligentRoute` CR may declare `NOT` with several conditions (`MaxItems=50`, and the CRD comment promises NOR semantics). NOR was never implemented — `evalNOT` warns and reports a non-match — so such a CR loaded with a permanently dead decision and now fails validation instead.

## #2122 acceptance criteria

| Criterion | |
|---|---|
| File and Kubernetes/canonical paths enforce the same rule | ✅ one `validateDecisionContracts`, reached from every loader — and the leaf's signal type is now checked there too, so the routing-fragment and canonical paths stop accepting a type only the full-config path rejected |
| Errors carry the decision name and rule-tree location | ✅ unified across the tree, leaf and signal-reference checks, which used two other prefixes |
| Tests cover omitted/AND/OR/NOT, invalid values, nested nodes, arity | ✅ 35-case corpus in `decision_rule_tree_corpus_test.go`, asserted through `ParseYAMLBytes` — the entry point the router, dashboard and canonical/CRD loader share — plus `ParseRoutingYAMLBytes` for the DSL-owned routing fragment, which skips the signal-reference validators |
| Runtime and validation share the operator set | ✅ the set lives in `routing_surface_catalog.go` with the other routing-surface enums, behind `RuleTreeOperators()` / `IsRuleTreeOperator()`, and both `evalNode` and `evalNodeWithTrace` switch on its constants. `TestRuleTreeOperatorsAgreeWithEvaluator` drives every operator through **both** evaluators and fails if one reaches a default branch, or if an operator is added without a case in either |
| Maintained examples and DSL output still validate | ✅ the scan above, plus the `pkg/dsl` round-trip tests |
| *Expected behavior:* an omitted operator keeps the AND default | ⚠️ deliberately not implemented — the router evaluates it as `OR`, so asserting `AND` would document a behaviour that does not exist. Deferred to #2937 |

## Follow-ups

- **#2937** — rule-tree fields are not normalized consistently across producers: omitted means `OR` in the router but `AND` in the CLI, DSL and docs, and the DSL decompilers switch on the raw string, so a valid lowercase `and` decompiles into a route with no `WHEN` clause at all. The same root cause covers the other identity fields: a padded `name` loads clean and then never matches, because the validator trims it for the declared-signal lookup while the evaluator compares the raw value. Choosing the omitted-operator default is a maintainer call, so this PR documents the ambiguity instead of picking one.
- **#2938** — no depth or node-count limit on any surface. Measured ceiling across the same scan: depth 6, 49 nodes.
- ~~**#2939**~~ — fixed in main by #2945, which unblocked the "a leaf requires a `name`" check now included here.

Tested: `pkg/config` (and `-race`), `pkg/decision`, `pkg/dsl`, `pkg/k8s`, dashboard `recipe`/`configprojection`/`handlers`, operator `controllers`/`api`. `make agent-lint` clean, markdownlint clean, DCO signed.


theohsiung:
@rootfs @Xunzhuo Hi I wonder if (XOR, NOR etc) are requiered, if so I can open a new PR to add this new feature

Xunzhuo (changes requested):
Blocking on issue intake: #2122 is still needs-acceptance, so this implementation cannot be approved yet. Please wait for the intake decision; no additional code changes are requested in this review.

Xunzhuo (commented):
Thanks—the linked issue is accepted now, and the rule-tree validation itself looks sound. This head no longer merges with main in the validator and both evaluator paths; main also added `on_unknown` handling in the same walker. Please rebase, preserve that current behavior while resolving the conflicts, and rerun the exact-head checks.

FAUST-BENCHOU (commented):
Thanks for addressing this validation gap. Enforcing the decision rule-tree contract at load time helps prevent silent routing behavior changes caused by invalid configurations.

The changes look good overall. Please resolve the current conflicts with main first so we can review and merge against the latest codebase.

PR #2125, opened 2026-06-10
[Router] Resolve reasoning family by routing model, not rewritten upstream id

### LINKED ISSUE #2124: [Router] use_reasoning silently dropped when provider_model_id differs from model name
## Summary

With a **valid** reasoning config, `use_reasoning: true` is silently dropped — no reasoning parameter is sent upstream — whenever a model's `provider_model_id` (or `external_model_ids`) differs from its router model name. That is the normal case, and it matches the shipped `config/config.yaml` (model `qwen3-8b`, `provider_model_id: qwen3-8b-instruct`, `reasoning_family: qwen3`).

## Reproduction (controlled, live router)

Valid config:
```yaml
providers:
  defaults:
    reasoning_families:
      qwen3: {type: chat_template_kwargs, parameter: enable_thinking}
  models:
    - name: qwen3-122b
      provider_model_id: Qwen/Qwen3.5-122B-A10B-GPTQ-Int4   # differs from name
      reasoning_family: qwen3
routing:
  decisions:
    - name: reason-on-route
      rules: {operator: AND, conditions: []}
      modelRefs:
        - model: qwen3-122b
          use_reasoning: true
```

Send any request (auto/`MoM`). The decision engine logs `reasoning_enabled=true`, but the upstream body has **no** `chat_template_kwargs`:
```
req_filter_reason.go: "No reasoning support for model: Qwen/Qwen3.5-122B-A10B-GPTQ-Int4 (no reasoning family configured)"
body_mutation: {"max_tokens":...,"messages":[...],"model":"Qwen/Qwen3.5-122B-A10B-GPTQ-Int4"}   # no reasoning param
```

Setting `provider_model_id == name` (no rewrite) is the **only** change needed to make it work — reasoning is then injected:
```
body_mutation: {"chat_template_kwargs":{"enable_thinking":true},...,"model":"qwen3-122b"}
```
This isolates the cause to the model-name dimension.

## Root cause

In the auto-routing path:
- `handleAutoModelRouting` rewrites the body model to the upstream id (`resolveModelNameForBackend`) and calls `modifyRequestBodyForAutoRouting`, which sets `openAIRequest.Model = upstreamModel`.
- `setReasoningModeToRequestBodyForProvider` → `parseReasoningRequestMutation` reads `mutation.model` from the (now-rewritten) request body, then `getModelReasoningFamily(mutation.model)` looks up the family by the **upstream id**.
- `reasoning_family` is configured on the **router model name**, so the lookup misses and reasoning is skipped. The same applies to `getReasoningEffort` (it matches the decision's `modelRefs`, which use router names), so the `reasoning_effort` family is affected too.

## Fix

Thread the selected router model name into `setReasoningModeToRequestBodyForProvider` and use it for reasoning-family / reasoning-effort / metric resolution. The request body's `model` field is unchanged (still the upstream id). PR attached.

theohsiung:
## What

With a valid reasoning config, `use_reasoning: true` was silently dropped (no reasoning parameter sent upstream) whenever a model's `provider_model_id` / `external_model_ids` differ from its router model name — the normal case, including the shipped `config/config.yaml`.

Fixes #2124.

## Root cause

In the auto-routing path the body model is rewritten to the upstream id before reasoning runs; `setReasoningModeToRequestBodyForProvider` then resolved the reasoning family from that rewritten body model (`getModelReasoningFamily(mutation.model)`), but `reasoning_family` is keyed by the router model name → no match → reasoning skipped (only an info log "no reasoning family configured"). `getReasoningEffort` is affected the same way.

## Fix

Thread the selected router model name into `setReasoningModeToRequestBodyForProvider` and use it for reasoning-family / reasoning-effort / metric resolution. The request body's `model` field is unchanged (still the upstream id). The looper path passes `""` and keeps the prior body-model behavior.

## Test plan

- [x] Added `pkg/extproc/reasoning_family_lookup_test.go`: router model carries `reasoning_family`, body holds the rewritten upstream id → asserts `enable_thinking:true` is injected. RED before fix (compile/behavior), GREEN after.
- [x] Verified end-to-end on a live router: differ → reasoning dropped; `provider_model_id == name` → reasoning injected (controlled, single-variable).

## Notes
- DCO signed-off.


Xunzhuo (commented):
Current main already preserves the logical model through `providerDispatch.logicalModel` and uses it for reasoning-mode adaptation. This branch targets the pre-unified protocol path and now conflicts with main, so the reported lookup bug is already superseded. Please close this PR unless a rebased branch has a distinct remaining case.

FAUST-BENCHOU (commented):
Thanks for fixing this reasoning-family lookup issue. Resolving the router model name vs provider model id mismatch addresses a real case where reasoning configuration could be silently ignored.

Issue #2126, opened 2026-06-10
[Router] system_prompt insert mode drops structured system-message content (and duplicates non-first system messages)


theohsiung:
## Summary

With a valid `system_prompt` plugin (`mode: insert`), the existing system message's instructions are **silently dropped** when its content is structured (an array of content parts) — valid OpenAI input. The injected prompt remains but the original system content is lost.

A closely related defect (B, below) comes from the same `firstSystemMessage` helper assuming a string system message at index 0.

## (A) Structured system content dropped — fixed in the attached PR

Request (valid):
```json
{"messages":[{"role":"system","content":[{"type":"text","text":"ORIG_SYS"}]},{"role":"user","content":"hi"}]}
```
With an `insert`-mode `system_prompt` plugin, the upstream body becomes:
```json
{"messages":[{"role":"system","content":"INJECTED\n\n"},{"role":"user","content":"hi"}], ...}
```
`ORIG_SYS` is gone. (A string system message is preserved correctly — only the structured form is dropped.)

Root cause: `firstSystemMessage` (`pkg/extproc/req_filter_sys_prompt.go`) reads `firstMsg["content"].(string)`; structured content fails the assertion and yields `""`, so `systemPromptContent` merges `injected + "\n\n" + ""`.

## (B) Non-first system message → duplicate system message → backend 400 (related, not in this PR)

Request:
```json
{"messages":[{"role":"user","content":"hi"},{"role":"system","content":"ORIG"}]}
```
`firstSystemMessage` only inspects `messages[0]` (role `user`), concludes "no existing system message", and **prepends** a new system message, yielding two system messages:
```json
{"messages":[{"role":"system","content":"INJECTED"},{"role":"user","content":"hi"},{"role":"system","content":"ORIG"}]}
```
Backends reject this (`"System message must be at the beginning."` → 400).

This shares the same index-0 assumption, but the correct behavior (relocate/merge the existing system message vs. reject) is a design decision, so it is intentionally left for a follow-up.

## Fix (A)

Extract text from structured system content (`text`/`input_text` parts) in `firstSystemMessage` so insert mode preserves the original instructions. PR attached.

PR #2127, opened 2026-06-10
[Router] Preserve structured system-message content in system_prompt insert mode

theohsiung:
## What

With a valid `system_prompt` plugin (`mode: insert`), the existing system message's instructions were **silently dropped** when its content is structured (an array of content parts — valid OpenAI input). The injected prompt remained but the original system content was lost (`"INJECTED\n\n"`).

Fixes part (A) of #2126.

## Root cause

`firstSystemMessage` (`pkg/extproc/req_filter_sys_prompt.go`) read `firstMsg["content"].(string)`; structured content fails the assertion → `""`, so insert mode merged `injected + "\n\n" + ""` and dropped the original.

## Fix

Extract text from structured system content (`text`/`input_text` parts) in `firstSystemMessage`, so insert mode preserves the original system instructions while prepending the configured prompt. String content is unchanged.

## Test plan

- [x] Added `pkg/extproc/system_prompt_structured_content_test.go`: structured-content insert preserves the original (RED before fix → GREEN), and a string-content regression test.
- [x] Reproduced/verified on a live router via `SR_LOG_LEVEL=debug` body-mutation logs.

## Out of scope (deferred in #2126)

Part (B): a system message **not** at index 0 makes `firstSystemMessage` report "no system message", so insert prepends a **duplicate** system message → two system messages → backend 400. Same index-0 assumption, but the correct fix (relocate/merge vs. reject) is a design decision.

## Notes
- DCO signed-off.


src/semantic-router/pkg/extproc/req_filter_sys_prompt.go
```diff
@@ -121,8 +122,35 @@ func firstSystemMessage(messages []interface{}) (string, bool) {
 	if !ok || role != "system" {
 		return "", false
 	}
-	content, _ := firstMsg["content"].(string)
-	return content, true
+	return systemMessageContentText(firstMsg["content"]), true
```
AayushSaini101, line 125: Good catch thanks : )

AayushSaini101 (approved):
LGTM thanks

PR #2129, opened 2026-06-10
[Router] request_params strip_unknown: keep valid OpenAI fields and router-injected chat_template_kwargs

### LINKED ISSUE #2128: [Router] request_params strip_unknown drops valid OpenAI params (max_completion_tokens, stream_options, ...) and the router-injected chat_template_kwargs
## Summary

The `request_params` plugin's `strip_unknown: true` removes any top-level field not in `allowedTopLevelFields`, but that allowlist is stale. As a result it silently drops **valid** OpenAI request parameters and even the reasoning parameter the router itself injects.

`strip_unknown: true` is the value used in the request-params tutorial (`website/docs/tutorials/plugin/request-params.md`).

## Reproduction (live router, observed via the upstream body)

Decision with `use_reasoning: true` and a `request_params` plugin (`strip_unknown: true`). Request:
```json
{"model":"MoM","messages":[...],"max_completion_tokens":64,"top_k":20,"stream_options":{"include_usage":true},"temperature":0.5}
```
Upstream body actually sent:
```json
{"messages":[...],"model":"qwen3-122b","temperature":0.5}
```
Stripped: `max_completion_tokens`, `stream_options`, `top_k`, and `chat_template_kwargs` (the reasoning param the router injected for the qwen3 family). Only allowlisted fields (`temperature`, `messages`, `model`) survived.

## Root cause

`allowedTopLevelFields` (`pkg/extproc/req_filter_request_params.go`) omits:
- current OpenAI fields: `max_completion_tokens` (replaces `max_tokens`), `stream_options`, `parallel_tool_calls`, `metadata`, `store`, `prediction`, `service_tier`, `modalities`, `audio`
- legacy: `functions`, `function_call`
- **`chat_template_kwargs`** — injected by the router's own reasoning mutation (`req_filter_reason.go`), and `request_params` runs *after* reasoning in the auto-routing pipeline, so strip_unknown removes it.

## Impact

With `strip_unknown: true`:
- `max_completion_tokens` (the current standard token limit) is dropped → the model runs with its default limit (wrong/truncated output).
- Enabling reasoning + strip_unknown on the same decision silently disables reasoning (its `chat_template_kwargs` is stripped).
- `stream_options.include_usage` is dropped → streaming usage accounting breaks.

## Fix

Extend `allowedTopLevelFields` with the current OpenAI fields, the legacy function fields, and `chat_template_kwargs`. Genuinely unknown fields are still stripped. PR attached.

## Related (not in this PR)

vLLM-specific sampling params (`top_k`, `min_p`, `repetition_penalty`) are also stripped. Since the router proxies to vLLM-compatible backends these are often valid, but whether to allowlist backend-specific params under `strip_unknown` is a policy decision left for maintainers.

theohsiung:
## What

`request_params` `strip_unknown: true` removed any top-level field absent from a stale `allowedTopLevelFields`, silently dropping valid OpenAI request parameters — and the reasoning parameter the router itself injects.

Fixes #2128.

## Reproduction

Decision with `use_reasoning: true` + `request_params { strip_unknown: true }`. A request with `max_completion_tokens`, `stream_options`, `top_k`, `temperature` reached the backend as just `{messages, model, temperature}` — `max_completion_tokens`, `stream_options`, and the injected `chat_template_kwargs` were all stripped.

## Fix

Extend `allowedTopLevelFields` with current OpenAI fields (`max_completion_tokens`, `stream_options`, `parallel_tool_calls`, `metadata`, `store`, `prediction`, `service_tier`, `modalities`, `audio`), legacy function fields, and `chat_template_kwargs` (router-injected). Genuinely unknown fields are still stripped.

## Test plan

- [x] Added `pkg/extproc/request_params_strip_allowlist_test.go`: valid/injected fields survive strip_unknown; an unknown field is still removed. RED before fix → GREEN.
- [x] Reproduced/verified live via `SR_LOG_LEVEL=debug` body-mutation logs.

## Related (deferred, see #2128)

vLLM sampling params (`top_k`, `min_p`, `repetition_penalty`) are also stripped; allowlisting backend-specific params is a maintainer policy decision.

## Notes
- DCO signed-off.

## Follow-up commit: close the cap-bypass this allowlist change would open

Allowlisting `max_completion_tokens` exposed a governance gap: `max_tokens_limit` only capped `max_tokens`, so a client could bypass the limit by renaming the field. Second commit applies the same `max_tokens_limit` cap to `max_completion_tokens` (same metric), with a dedicated unit test (RED→GREEN).

### DIFF
--- a/src/semantic-router/pkg/extproc/req_filter_request_params.go
+++ b/src/semantic-router/pkg/extproc/req_filter_request_params.go
@@ -73,6 +73,9 @@ func (r *OpenAIRouter) buildRequestParamsMutations(
 
 	modified := applyBlockedParams(body, paramsConfig.BlockedParams, decision.Name)
 	modified = capIntField(body, "max_tokens", paramsConfig.MaxTokensLimit, decision.Name, metrics.RecordMaxTokensCapped) || modified
+	// max_completion_tokens is the current OpenAI name for the same knob; cap it
+	// with the same limit so clients cannot bypass max_tokens_limit by renaming.
+	modified = capIntField(body, "max_completion_tokens", paramsConfig.MaxTokensLimit, decision.Name, metrics.RecordMaxTokensCapped) || modified
 	modified = capIntField(body, "n", paramsConfig.MaxN, decision.Name, metrics.RecordMaxNCapped) || modified
 	modified = stripUnknownFields(body, paramsConfig.StripUnknown, decision.Name) || modified
 
--- a/src/semantic-router/pkg/extproc/req_filter_request_params.go
+++ b/src/semantic-router/pkg/extproc/req_filter_request_params.go
@@ -32,6 +32,22 @@ var allowedTopLevelFields = map[string]bool{
 	"tool_choice":       true,
 	"user":              true,
 	"reasoning_effort":  true,
+	// Current OpenAI Chat Completions fields that postdate the original list.
+	"max_completion_tokens": true,
+	"stream_options":        true,
+	"parallel_tool_calls":   true,
+	"metadata":              true,
+	"store":                 true,
+	"prediction":            true,
+	"service_tier":          true,
+	"modalities":            true,
+	"audio":                 true,
+	// Legacy function-calling fields still accepted by OpenAI-compatible backends.
+	"functions":     true,
+	"function_call": true,
+	// Injected by the router's own reasoning mutation; must not be stripped here,
+	// otherwise enabling reasoning together with strip_unknown silently removes it.
+	"chat_template_kwargs": true,
 }
 
 func (r *OpenAIRouter) buildRequestParamsMutations(
--- a/src/semantic-router/pkg/extproc/req_filter_request_params_test.go
+++ b/src/semantic-router/pkg/extproc/req_filter_request_params_test.go
@@ -62,3 +62,35 @@ func TestBuildRequestParamsMutationsBlockedAndCaps(t *testing.T) {
 		t.Fatalf("n cap: got %v", body["n"])
 	}
 }
+
+// max_tokens_limit must also cap max_completion_tokens (the current OpenAI
+// name for the same knob). Now that the strip_unknown allowlist passes
+// max_completion_tokens through, leaving it uncapped would let clients bypass
+// the governance limit by simply renaming the field.
+func TestBuildRequestParamsMutationsCapsMaxCompletionTokens(t *testing.T) {
+	r := &OpenAIRouter{}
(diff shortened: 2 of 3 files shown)


src/semantic-router/pkg/extproc/req_filter_request_params.go
```diff
@@ -32,6 +32,22 @@ var allowedTopLevelFields = map[string]bool{
 	"tool_choice":       true,
 	"user":              true,
 	"reasoning_effort":  true,
+	// Current OpenAI Chat Completions fields that postdate the original list.
```
drivebyer, line 35: Non-blocking idea: rather than hand-maintaining `allowedTopLevelFields`, it could be reflected from `openai.ChatCompletionNewParams` at startup so it tracks the SDK automatically.

theohsiung: Good idea — reflecting from `ChatCompletionNewParams` would cut drift for SDK-tracked fields and auto-pick up newer ones (`prompt_cache_key`, `safety_identifier`, `web_search_options`).

One caveat though: pure reflection over the struct misses two fields we must keep allowed:

- `chat_template_kwargs` — router-injected, not part of the OpenAI SDK at all (this is the field the PR exists to protect).
- `stream` — openai-go drives streaming via `NewStreaming()`, so `Stream` isn't a field on `ChatCompletionNewParams`; reflection would omit it and strip_unknown would drop `stream` from streaming requests.

So it'd need to be `reflect(ChatCompletionNewParams) ∪ {stream, chat_template_kwargs}` (a small explicit router-injected/not-in-SDK supplement). Happy to do that as a follow-up rather than widen this PR's scope.

drivebyer: It would be great if you could create an issue to track this. :)

drivebyer (approved):

Xunzhuo (commented):
Current main removed this raw allowlist: strict ingress codecs reject unknown fields, and neutral `MaxOutputTokens` now covers and caps the supported token fields. This branch targets deleted implementation and conflicts with main. Please close it as superseded by the unified protocol work unless a distinct gap remains.

FAUST-BENCHOU (commented):
Thanks for fixing this request parameter filtering issue. Keeping valid OpenAI fields and router-injected reasoning parameters from being stripped is an important correctness fix.

wilsonwu (approved):
LGTM

drivebyer:
@theohsiung Please resolve the conflict.

PR #2131, opened 2026-06-10
[Router] Do not cache non-2xx upstream responses (cache poisoning fix)

### LINKED ISSUE #2130: [Router] Semantic cache stores non-2xx upstream responses and replays them as HTTP 200 (cache poisoning)
## Summary

The semantic cache stores **non-2xx upstream responses** and replays them as HTTP **200** cache hits. A transient upstream error (429/502/503/504/timeout) or a permanent one (400/500) is frozen into the cache for the whole TTL and served as a "success" to every semantically-similar request — even after the upstream recovers. Clients that retry on 5xx/429 will not retry a fake 200.

## Reproduction (live, in-memory cache)

Decision with a `semantic-cache` plugin (`enabled: true`, `similarity_threshold: 0.8`) and a model whose upstream rejects it (here `provider_model_id` is unknown, so the backend returns HTTP 400 + a JSON error body).

- **Request 1** (`{"model":"cachebug","messages":[{"role":"user","content":"<prompt>"}]}`): client correctly receives **HTTP 400** with the upstream error body. ✅
- **Request 2** (identical prompt): client receives **HTTP 200 OK** with:
  - header `x-vsr-cache-hit: true`
  - header `x-vsr-cache-similarity: 1.0000`
  - body = the upstream **400** error JSON: `{"error":{"message":"...Invalid model name...","code":"400"}}`

So an error body is served as a 200 success. A real client parsing `choices[0].message.content` gets nothing while the status line falsely says success.

## Root cause

`pkg/extproc/processor_res_cache.go`:

- `updateResponseCache` (non-streaming) calls `r.Cache.UpdateWithResponse(...)` **unconditionally** — it never inspects the upstream HTTP status.
- The status *is* computed at the response-header phase (`evaluateResponseHeaderOutcome` → `statusCode`/`isSuccessful` in `processor_res_header_runtime.go`) but is **discarded** — used only for metrics, the upstream span, and the optional replay recorder. It is **not stored on `RequestContext`**, so the response-body phase has no way to gate on it.
- Cache hits are returned via `CreateCacheHitResponse` as HTTP 200 (`req_filter_cache.go`), so whatever was stored is replayed as a success.

The streaming path is partly protected by `validateStreamingCachePreconditions` (an error response is usually not SSE → no completion/content → skipped), but the non-streaming path has no such guard — the asymmetry is the giveaway that the missing gate is an oversight.

## Impact

- **Correctness:** error bodies served as HTTP 200 → clients cannot distinguish success from failure; downstream parsing breaks.
- **Reliability:** a transient upstream blip (429/5xx/timeout) is cached and amplified — similar queries keep getting the stale error for the full TTL even after recovery.
- **Retry/UX:** clients retry on 5xx/429 but not on 200, so the cached error short-circuits their recovery path.

## Fix (PR attached)

Capture the upstream status on `RequestContext.UpstreamStatusCode` at the response-header phase, and gate both the non-streaming and streaming cache writes: skip the write unless the status is 2xx. An unobserved status (`0`, e.g. paths that never process response headers) does not block the write, preserving prior behavior. Skips are recorded with the existing `RecordCacheWriteSkipped` metric + `cache.write_skipped_reason` trace attribute.

theohsiung:
## What

The semantic cache wrote the upstream response body **unconditionally** on the non-streaming path, so a non-2xx error body got cached and replayed as an HTTP **200** cache hit (cache poisoning).

Fixes #2130.

## Reproduction (live, in-memory cache)

Decision with a `semantic-cache` plugin + a model the upstream rejects (HTTP 400 + JSON error body):

- Request 1 (identical prompt): client gets **HTTP 400** + upstream error body. ✅
- Request 2 (identical prompt): client gets **HTTP 200 OK**, `x-vsr-cache-hit: true`, `x-vsr-cache-similarity: 1.0000`, and the **400 error JSON** as the body. ❌

## Root cause

- `updateResponseCache` (non-streaming) calls `Cache.UpdateWithResponse` without inspecting the upstream status.
- The status computed at the response-header phase (`evaluateResponseHeaderOutcome`) was discarded — not stored on `RequestContext` — so the body phase could not gate on it.
- Cache hits are served as HTTP 200 via `CreateCacheHitResponse`.

## Fix

- Add `RequestContext.UpstreamStatusCode`; set it in `handleResponseHeaders`.
- New `shouldSkipCacheWriteForStatus`: skip the write unless 2xx. Status `0` (never observed) does **not** block — preserves prior behavior for paths that don't process response headers.
- Gate both `updateResponseCache` and `cacheStreamingResponse`; record the skip with the existing `RecordCacheWriteSkipped` metric + `cache.write_skipped_reason` trace attribute.

## Test plan

- [x] `processor_res_cache_status_test.go`: gate truth table (0/200/204/299 → keep; 100/301/400/429/500/503 → skip); `updateResponseCache` skips on 400, writes on 200, writes on unknown (0); `cacheStreamingResponse` skips on 502. RED before fix → GREEN.
- [x] `processor_res_header_test.go`: `handleResponseHeaders` persists 200/400/429/503 onto `ctx.UpstreamStatusCode`.
- [x] `go test ./pkg/extproc/` full suite green (no regression; existing retention/TTL cache tests unaffected — they use status 0).
- [x] Reproduced the bug live before the fix (see above).

## Notes
- DCO signed-off.
- Scope: cache write gating only. Reporting the *client-facing* error status on request 1 is already correct (passthrough); this PR only stops the error from being cached.


AayushSaini101 (approved):
LGTM thanks :) cc: @rootfs @Xunzhuo can you take a look as well on this thanks

AayushSaini101:
QQ @theohsiung Do we need to cache all the 200-299 status codes response?

theohsiung:
@AayushSaini101  Good question, thanks! 
Short answer: in practice only 200 shows up here: vLLM / OpenAI-compatible backends return 200 for a successful completion and 4xx/5xx on error, so for this path 2xx and == 200 are equivalent today. I went with the full 2xx success class to stay consistent with how the router already defines a successful response — evaluateResponseHeaderOutcome uses statusCode >= 200 && statusCode < 300 (in processor_res_header_runtime.go) to decide whether to emit the VSR decision headers. Reusing the same range keeps the cache-write gate aligned with that existing convention.

theohsiung:
@rootfs @Xunzhuo — friendly nudge 🙏 Approved (by @AayushSaini101, who cc'd you both) and `MERGEABLE` — only `BEHIND` main, no conflicts (Mergify should update it at merge time). Could you do the final merge when you have a moment? It's a small, self-contained cache-poisoning fix (non-2xx upstream responses no longer cached + replayed as 200). Thanks!

PR #2133, opened 2026-06-10
[Router] Redact upstream credentials from ext_proc debug logs (CWE-532)

### LINKED ISSUE #2132: [Router] Upstream provider API key leaked in plaintext to debug logs (CWE-532)
## Summary

At debug log level the router writes the **upstream provider API key in plaintext** to its logs. `sendResponse` dumps every ext_proc `ProcessingResponse` with `%+v`, and the credential resolver injects the provider key as a set-header — so the dump includes `Authorization: Bearer <key>` (or `x-api-key`, etc.). This is CWE-532 (Insertion of Sensitive Information into Log File).

## Reproduction (live)

Any routed chat request with `SR_LOG_LEVEL=debug` produces a log line like:

```
{"level":"debug","caller":"utils.go:31","msg":"Processing at stage [request body]: request_body:{response:{header_mutation:{set_headers:{header:{key:\"Authorization\"  raw_value:\"Bearer sk-3pn5Mmz90jlDK7perGMMow\"}} ... }}}"}
```

The `Bearer sk-...` value is the real upstream LLM credential (here injected via `api_key_env`).

## Root cause

`pkg/extproc/utils.go` `sendResponse`:
```go
logging.Debugf("Processing at stage [%s]: %+v", msgType, response)
```
`response` carries the `HeaderMutation.SetHeaders` that the routing path populated with the credential (`processor_req_body_routing.go` injects `authHeader: <prefix> <accessKey>` as `RawValue`; the Anthropic path injects `x-api-key`). Nothing redacts the value before it is logged.

## Impact

- Debug logging is routinely enabled in dev/staging and during prod incident troubleshooting, and logs are commonly shipped to centralized stores (Loki/ELK/CloudWatch) with broad read access and long retention.
- The leaked key is the **backend LLM provider credential** — it can incur cost and has whatever scope the provider key grants. Anyone with log access obtains it.
- Same credential-hygiene class the project already addressed for the `/info/classifier` endpoint (adding `json:"-"` to credential fields).

## Fix (PR attached)

Redact the values of sensitive headers (`Authorization`, `x-api-key`, `proxy-authorization`, and by-convention `*-key` / `*token*` / `*secret*` / `*api-key*` names — covering custom provider auth headers) in a **deep copy** of the response before logging; the response actually sent to Envoy is untouched. The clone+redact work is gated behind a debug-level check (`logging.DebugEnabled()`) so it costs nothing when debug logging is off.

## Note

On a shared host the router API port often binds `0.0.0.0`; combined with this log leak, an exposed key should be rotated.

theohsiung:
## What

At debug level the router logged the **upstream provider API key in plaintext**. `sendResponse` dumped every ext_proc `ProcessingResponse` with `%+v`, and the credential resolver injects the provider key as a set-header, so the dump included `Authorization: Bearer <key>` / `x-api-key: <key>` (CWE-532).

Fixes #2132.

## Reproduction (live)

A routed request with `SR_LOG_LEVEL=debug` produced:
```
caller":"utils.go:31" ... set_headers:{header:{key:"Authorization" raw_value:"Bearer sk-..."}}
```
i.e. the real upstream credential.

## Fix

- `pkg/extproc/log_redaction.go`: `redactProcessingResponseForLog` deep-copies the response and masks the values of sensitive headers (`Authorization`, `x-api-key`, `proxy-authorization`, `x-goog-api-key`, plus by-convention `*-key`/`*token*`/`*secret*`/`*api-key*` to cover custom provider auth headers). The response sent to Envoy is untouched.
- `pkg/extproc/utils.go`: log the redacted copy, gated on `logging.DebugEnabled()` so the clone+redact cost is only paid when debug logging is on.
- `pkg/observability/logging`: add `DebugEnabled()`.

## Test plan

- [x] `log_redaction_test.go`: `isSensitiveHeaderKey` truth table; redaction masks Authorization (RawValue) + x-api-key (Value), preserves a non-sensitive header, and leaves the **original** response (sent to Envoy) intact; nil-safe.
- [x] `gofmt`/`go vet` clean; `golangci-lint` (repo config, incl. gosec) → 0 issues.
- [x] Reproduced the leak live before the fix.

## Notes
- DCO signed-off.
- Same credential-hygiene class as the `/info/classifier` `json:"-"` fix.
- On a shared host the API port often binds 0.0.0.0; an exposed key should be rotated.


theohsiung:
@AayushSaini101 thx for reviewing

src/semantic-router/pkg/extproc/utils.go
```diff
@@ -28,7 +28,13 @@ func sendResponse(stream ext_proc.ExternalProcessor_ProcessServer, response *ext
 		}
 	}
 
-	logging.Debugf("Processing at stage [%s]: %+v", msgType, response)
+	// Redact credentials (Authorization / x-api-key / ...) before dumping the
+	// mutation: the credential resolver injects the upstream provider key as a
+	// set-header, so a verbatim %+v would leak it to the log (CWE-532). Gate on
+	// the debug level so the clone+redact cost is paid only when it is logged.
+	if logging.DebugEnabled() {
+		logging.Debugf("Processing at stage [%s]: %+v", msgType, redactProcessingResponseForLog(response))
```
AayushSaini101: @theohsiung can we use better naming here? We are filtering the incoming response, so it is better to use proper naming convention in-future if any user want to leverage the same method can easily understand the purpose of the method thanks

src/semantic-router/pkg/extproc/response_log_redaction.go
```diff
@@ -0,0 +1,102 @@
...
+		return v.RequestBody.GetResponse().GetHeaderMutation()
+	case *ext_proc.ProcessingResponse_ResponseBody:
+		return v.ResponseBody.GetResponse().GetHeaderMutation()
+	}
+	return nil
+}
+
+// redactHeaderMutation masks the value of every sensitive set-header in place.
+// It operates on a clone, so the mutation sent to Envoy is unaffected.
+func redactHeaderMutation(hm *ext_proc.HeaderMutation) {
```
AayushSaini101, line 80: can we use here proper naming convention as well for the methods that are calling here thanks

src/semantic-router/pkg/extproc/utils.go
theohsiung: @AayushSaini101 thanks for the feedback! For this one I'm proposing

**redactProcessingResponseForLog → redactSensitiveHeadersForLog** 

clearer that it masks the sensitive (credential) headers before logging. Does that  work, or do you have a preferred name? Will push once you confirm the naming.

src/semantic-router/pkg/extproc/response_log_redaction.go
theohsiung: For the helpers here: `redactHeaderMutation` / `redactHeaderValueOption` already follow a consistent `redact*` convention, so I'd keep those as-is. The one worth renaming for clarity is `commonResponseHeaderMutation` → `responseHeaderMutation` (the `common` just leaks the `CommonResponse` proto type). Full proposal is in the `utils.go:36` thread above — happy to use a different name if you have a preference. Will push once you confirm.

theohsiung:
My apologies, accidentally hit the review button

src/semantic-router/pkg/extproc/response_log_redaction.go
AayushSaini101: thanks, suggested one name above thread

src/semantic-router/pkg/extproc/utils.go
AayushSaini101: Thanks @theohsiung 🙏 , can we use redactResponseForLog(response) naming here, this will tell complete information, thoughts? if this is okay you can use this naming in the other places as well thanks
- redact → remove or mask sensitive information
- Response → the object being processed
- ForLog → the purpose is safe logging

theohsiung: Done 🙏 
 used `redactResponseForLog(response)` as you suggested, and applied the same clarity to the helper: `commonResponseHeaderMutation` → `responseHeaderMutation`. Kept `redactHeaderMutation` / `redactHeaderValueOption` since they already follow the `redact*` convention. Pushed in 670aaf74. Thanks for the review!

AayushSaini101: thanks a lot 🙏 ,  overall looks good to me , QQ: cannot we move the [log_redaction.go](https://github.com/vllm-project/semantic-router/pull/2133/changes#diff-fb699a7afb7b4a261be60726618976f1bc327690d1dc0cf766c0be1cc196209c) and test case to the pkg/observability/logging directory that makes the structure clearer, ?

theohsiung: Thanks! 🙏 I looked into it.

 `log_redaction.go` is built on Envoy ext_proc types — `ext_proc.ProcessingResponse`, `ext_proc.HeaderMutation`, `core.HeaderValueOption` — and it knows which headers the credential resolver injects, so it's really  **ext_proc-domain logic** rather than generic logging.

`pkg/observability/logging` is a foundational, leaf package: no `go-control-plane`/Envoy dependency today (just `zap`), and it's imported across the codebase (`config`, `services`, `cache`, …). Moving the redaction there would pull `go-control-plane` into that base package and **invert the layering** — a low-level utility depending on high-level Envoy types.

The generic, protocol-agnostic piece — `DebugEnabled()` — already lives in `logging`; the ext_proc-specific redaction is intentionally kept next to the code that produces the `ProcessingResponse`.

 So I'd lean toward keeping it in `extproc`. Happy to rename the file for clarity if that helps — e.g. `` `response_log_redaction.go` ``. WDYT?

AayushSaini101: thanks for the explanation @theohsiung 🙏 yes, I guess we can change the name to _response_log_redaction.go for better understanding thanks

theohsiung: Agree!

theohsiung: Done 🙏 — renamed to `response_log_redaction.go` (and the test to `response_log_redaction_test.go`), kept in `extproc` as discussed. Pushed in 17b659bb. Thanks for the review!

AayushSaini101 (approved):
LGTM thanks 👍

PR #2135, opened 2026-06-10
[Router] Enforce hard user-scope isolation in semantic cache (cross-tenant leak fix)

### LINKED ISSUE #2134: [Router] Semantic cache user-scoping is soft (embedding-distance) → cross-tenant response leak (CWE-524)
## Summary

The semantic cache's per-user isolation is **soft** — it relies on embedding *distance* from a short namespace prefix, not a hard partition. As a result **one user can receive another user's cached response** when their queries are semantically similar. Reproduced live: a different user got a byte-identical cached answer at similarity **0.9085** (threshold 0.8).

This is CWE-524 (Use of Cache Containing Sensitive Information) / cross-tenant data leak. LLM responses can contain personalized or confidential content, so cross-user cache hits leak data between tenants.

## How scoping works today

`ScopeQueryToUser(query, userID)` (`pkg/cache/cache.go`) prepends an HMAC-derived namespace token, repeated 3×, to the query text:
```
cache-scope <ns> <ns> <ns> <original query>
```
The cache then embeds this string and matches by cosine similarity ≥ threshold. The intent is that the prefix shifts user A's and user B's embeddings far enough apart to avoid a cross-user match.

**The flaw:** for anything but very short queries, a 16-hex-char prefix (×3) is a tiny fraction of the embedded text, so two users asking the *same* question embed almost identically. Isolation degrades to "hope the embedding distance exceeds the threshold," which it does not.

The existing tests (`cache_user_scope_test.go`) only assert that different users get different *scoped strings* — they never check that the resulting cache *lookup* is isolated. That gap hides the bug.

## Reproduction (live, in-memory cache, threshold 0.8, mmbert)

Router started with `SEMANTIC_CACHE_E2E_USER_FROM_BODY=true` (body `user` field → cache scope); decision has a `semantic-cache` plugin.

A long, identical question sent as three requests:
| # | body `user` | result |
|---|------------|--------|
| 1 | `alice` | MISS → upstream (cached) |
| 2 | `alice` | HIT, similarity **1.0000** (same user — expected) |
| 3 | `bob`   | **HIT, similarity 0.9085** — bob received alice's byte-identical cached answer |

Logs confirm all three resolved a distinct scope user (`cache_scope_user_resolved`, source `body_user_field`); bob's hit fired at 0.9085 > 0.8.

## Fix (PR attached)

Add a **hard scope gate**, independent of embedding similarity. The scoped query already carries the namespace, and each cache entry stores its scoped `Query`, so the backend can compare namespaces exactly and refuse any cross-scope candidate during search — a different user's entry is never returned even if it is the nearest neighbor. New helpers `CacheScopeNamespaceOf` / `SameCacheScope` in `cache.go`; the in-memory backend (default) enforces the gate in both its linear and HNSW scans.

Scope of this PR: the **in-memory backend** (where this was reproduced). The Redis/Milvus/Qdrant/Valkey backends share the same soft-scoping weakness and should adopt the same gate (server-side filter for the vector DBs); the shared helpers are ready for reuse. Tracking that as follow-up.

theohsiung:
## What

Per-user semantic-cache isolation was **soft**: `ScopeQueryToUser` prepends an HMAC namespace token (×3) to the query and relies on embedding *distance* to separate users. For non-trivial queries the prefix is a tiny fraction of the embedded text, so two users asking the same question embed almost identically and **one user receives another user's cached response** (CWE-524, cross-tenant leak).

Fixes #2134.

## Reproduction (live, in-memory cache, threshold 0.8, mmbert)

Router with `SEMANTIC_CACHE_E2E_USER_FROM_BODY=true`; decision has a `semantic-cache` plugin. Same long question, three requests:

| # | body `user` | result |
|---|---|---|
| 1 | alice | MISS → upstream (cached) |
| 2 | alice | HIT, similarity 1.0000 (same user — expected) |
| 3 | bob | **HIT, similarity 0.9085 — bob got alice's byte-identical answer** |

All three resolved distinct scope users in the logs; bob's hit fired at 0.9085 > 0.8.

## Fix

Hard scope gate, independent of embedding similarity:
- `cache.go`: `CacheScopeNamespaceOf` / `SameCacheScope` recover the namespace from a scoped query (each entry already stores its scoped `Query`).
- `inmemory_cache_search.go`: both the linear and HNSW scans skip any candidate whose namespace differs from the requester's — a different user's entry is never returned even if it is the nearest neighbor. Unscoped queries keep the empty global scope (no change for single-tenant setups).
- The shared per-candidate skip logic (response present → scope gate → expiry) is consolidated into one `entryEligible(entry, scopeNamespace, now)` helper, so the security-critical gate lives in a single place that neither scan path can silently drop. `SameCacheScope` is documented as a helper for other backends; the in-memory hot path recovers the requester's namespace once and compares it per candidate rather than re-parsing both sides.

## Test plan

- [x] `cache_scope_isolation_test.go`: `CacheScopeNamespaceOf`/`SameCacheScope` truth table; `TestSearchEnforcesUserScope` drives the gate through `runFindSimilarEmbeddingSearch` over **both** in-memory search paths (linear scan **and** the HNSW index), proving a different-scope entry with an IDENTICAL injected embedding is dropped, a same-scope entry hits, and an unscoped lookup never matches a scoped entry. Deterministic (embedding injected — the model is a process boundary, mocked here; no model load needed) and verified **load-bearing**: disabling the HNSW gate fails the `hnsw` subtest while `linear` still passes.
- [x] In-memory cache suite green (`TestInMemoryCache*`, incl. HNSW + pending workflow with real embeddings); existing scope tests still pass after the `entryEligible` refactor.
- [x] `gofmt`/`go vet` clean; `golangci-lint` → 0 issues in the changed files (`cache.go`, `inmemory_cache_search.go`, `cache_scope_isolation_test.go`).
- [x] Reproduced the leak live before the fix.
- [ ] Note: full `pkg/cache` suite also contains Milvus/Redis/Valkey integration tests that require external servers (skipped/failing locally without them) — unrelated to this change.

## Scope / follow-up

This PR fixes the **in-memory** backend (where reproduced). Three known limitations are deliberately left for follow-up:

- **HNSW hit-rate under dense multi-tenancy (fails closed — no leak).** `scanHNSWCandidates` retrieves the global top-K (k=10) nearest neighbors *before* applying the scope gate. In a busy multi-tenant cache, a requester's own entry can rank outside the global top-K (pushed out by other tenants' near-identical embeddings) and then get dropped → a false MISS that degrades the legitimate hit rate as tenant density grows. It never returns another user's entry. The real fix pushes the scope predicate *into* `searchKNN` so only same-scope nodes count toward k; the linear scan is unaffected. (Not caught by the gate tests above by design — in a small test the same-scope entry is always within top-K.)
- **First-class scope on the entry.** The gate recovers the namespace by re-parsing `entry.Query` per candidate. A cleaner design threads the scope explicitly through `AddEntry`/`AddPendingRequest` and stores `CacheEntry.ScopeNamespace`, removing the per-entry parse and the implicit coupling to `ScopeQueryToUser`'s string format. Kept as-is here because the parse approach is zero-migration for entries already stored with a scoped query.
- **Other backends.** Redis/Milvus/Qdrant/Valkey share the same soft-scoping weakness and should adopt the gate — for the vector DBs as a **server-side metadata filter** (a client-side post-filter would inherit the top-K limitation above). The shared `SameCacheScope` helper is ready for reuse.

DCO signed-off.


theohsiung:
@WUKUNTAI-0211 @rootfs @Xunzhuo 

would appreciate your eyes on this one when you get a chance 🙏. It's a **cross-tenant data-leak** fix in the semantic cache (CWE-524): per-user scoping was soft (embedding-distance only), so I could reproduce one user receiving **another user's cached response** for a similar query (similarity **0.9085 > 0.8** threshold). This PR adds a hard namespace gate in the in-memory backend so cross-scope entries are never returned. Given it's a tenant-isolation / security concern, your review would be really valuable. Thanks!

theohsiung:
Hi @WUKUNTAI-0211, thanks for reviewing! I think there are still a few follow-ups worth discussing. If you're interested, I'd be happy to chat in the [discussion channel](https://github.com/vllm-project/semantic-router/discussions/2180#discussion-10254025).

WUKUNTAI-0211 (approved):
## Approve with a couple of small suggestions

Nice work on this — the 0.9085 reproducer made it really easy to evaluate, and I walked through the call sites to confirm the fix holds:

- `cacheScopePrefix` reorder is byte-identical to the old format, so no migration needed
- `entryEligible` makes it structurally impossible for one scan path to drop the gate
- The test is actually load-bearing (verified the HNSW subtest fails when its gate is removed)
- For authenticated requests, the caller's namespace is prepended before any user content, so a crafted `cache-scope <ns>` substring in the prompt can't escape the requester's own scope

Two small things worth considering, neither is a blocker:

**1. Anonymous path is still a bit soft.** When `cacheScopeUserID` returns `""`, `ScopeQueryToUser` passes the raw query through, so `CacheScopeNamespaceOf(query)` ends up parsing whatever prefix the user wrote. Not a regression from this PR (soft scoping was worse here), and deployments shouldn't really let anonymous traffic into a multi-tenant cache anyway — but threading scope through `FindSimilarWithThreshold` as an explicit param, which lines up with your "first-class scope" follow-up, would close it cleanly. A crafted-prefix unit test would also be nice to lock the invariant.

**2. `Fixes #2134` will auto-close the issue, but Redis / Valkey / Milvus / Qdrant still leak with your exact reproducer.** Probably worth renaming to something like `[Router][in-memory] ...` and switching `Fixes #2134` → `Part of #2134` so the issue stays open with sub-issues per backend. Otherwise prod deployments on Redis (probably most of them?) will look "fixed" when they aren't.

Happy to see either addressed here or as immediate follow-ups — LGTM either way.

PR #2137, opened 2026-06-10
[Router] Validate /api/v1/similarity input: reject blank text and out-of-range priority

### LINKED ISSUE #2136: [Router] /api/v1/similarity input validation gaps: blank text, out-of-range priority, misleading dimension error
## What

The `/api/v1/similarity` and `/api/v1/similarity/batch` handlers (`src/semantic-router/pkg/apiserver/route_embeddings.go`) under-validate input relative to the documented request contract:

1. **Blank input accepted.** `handleSimilarity` rejects only `text1 == "" || text2 == ""`, so whitespace-only strings (e.g. `"   "`) pass through to the embedding backend instead of returning a 400. The batch endpoint likewise accepts a whitespace-only `query` and whitespace-only `candidates[]` entries.
2. **Priority weights unbounded.** `quality_priority` / `latency_priority` are documented as `0.0-1.0` (`config.go`), but negative or `> 1.0` values are silently accepted and forwarded to the model.
3. **Misleading dimension error.** `isValidDimension` accepts `64`, but the similarity/batch error message lists only `128, 256, 512, 768, 1024` — a valid dimension is described as invalid.

## Impact

Malformed similarity requests surface as late backend errors / 500s (or silently wrong scoring for out-of-range priorities) instead of a clear 400 at the API boundary. Same input-hardening class as #2121.

## Scope

Fix is limited to the similarity and batch-similarity paths. The `/api/v1/embeddings` handler is intentionally left out to avoid conflicting with #2104.

WUKUNTAI-0211:
## What

Hardens input validation for the `/api/v1/similarity` and `/api/v1/similarity/batch` handlers, which under-validated input versus the documented request contract:

- **Blank input** — whitespace-only `text1`/`text2` (similarity) and a whitespace-only `query` or any whitespace-only `candidates[]` entry (batch) now return `400 INVALID_INPUT` instead of reaching the embedding backend.
- **Priority range** — `quality_priority` / `latency_priority` outside the documented `[0.0, 1.0]` range now return `400 INVALID_PARAMETER` instead of being silently forwarded to the model.
- **Dimension error message** — now lists `64`, which `isValidDimension` already accepts (the message previously omitted it, describing a valid dimension as invalid).

`handleSimilarity` is refactored to extract `applySimilarityDefaults` / `validateSimilarityRequest`, mirroring the embeddings and batch paths so the similarity validator is unit-testable.

Same input-hardening class as #2121. The `/api/v1/embeddings` handler is intentionally left untouched to avoid conflicting with #2104.

## Fixes

Fixes #2136

## Test Plan

- New table-driven tests for `validateSimilarityRequest` and `validateBatchSimilarityRequest` covering blank/whitespace input, out-of-range priorities, and the `64` dimension case.
- `CGO_ENABLED=1 go test ./pkg/apiserver/` passes (new + existing tests); `go vet` and `gofmt` clean.


WUKUNTAI-0211:
@AayushSaini101 — pinging since you approved the sibling PR #2121 and
this is the same shape on `/api/v1/similarity` and
`/api/v1/similarity/batch`.

Three things:
1. Blank/whitespace `text1` / `text2` / `query` / `candidates[]` →
   `400 INVALID_INPUT` instead of reaching the embedding backend.
2. `quality_priority` / `latency_priority` outside `[0.0, 1.0]` →
   `400 INVALID_PARAMETER` instead of silent forward to the model.
3. Dimension-error message now lists `64`, which `isValidDimension`
   already accepts (the message was incorrectly excluding a valid
   dimension).

Refactored `handleSimilarity` to extract `applySimilarityDefaults` /
`validateSimilarityRequest` so the validators mirror the embeddings
and batch paths and are unit-testable. CI green, +118/-25 in 2 files.

`/api/v1/embeddings` is intentionally left untouched here to avoid
conflicting with #2104.

theohsiung (approved):
Nice, well-scoped hardening. The blank/whitespace rejection (`text1`/`text2`/`query`/`candidates[]`), the corrected dimension message (adding `64`), and the `applySimilarityDefaults` / `validateSimilarityRequest` extraction all mirror the existing embeddings/batch patterns cleanly and are covered by focused table-driven tests. LGTM on those.

One note on the priority-range validation (not blocking):

`quality_priority` / `latency_priority` are only consumed on the embeddings `auto` path via `GetEmbeddingWithMetadata` → `select_embedding_model` (the weights drive the Qwen3-vs-Gemma choice). The similarity and batch backends don't take them:

```go
CalculateEmbeddingSimilarity(text1, text2, modelType, targetDim)
CalculateSimilarityBatch(query, candidates, topK, modelType, targetDim)
```

So on the two endpoints this PR touches, the priority fields are effectively inert today — the new `[0,1]` check is API-contract hygiene (honoring the documented range with a clear 400) rather than a fix for actual mis-scoring. Fine to keep, but two small asks:

1. **Wording** — "silently forwarded to the model / silently wrong scoring" isn't quite accurate for these two paths, since the value never reaches the backend. Something like "reject values the documented `[0.0, 1.0]` contract declares invalid" is more precise.

2. **Follow-through on #2104** — the one path that *actually* consumes priority is `/api/v1/embeddings`, and it's still unvalidated here (deliberately, to avoid colliding with #2104). Could you make sure the embeddings priority-range validation actually lands in #2104 (or a stated follow-up)? Otherwise we end up validating the two paths where priority is inert while the one path where it matters stays unchecked.

Approving so this isn't blocked — the hardening here is a clear improvement. 🙏

drivebyer (approved):

PR #2139, opened 2026-06-10
[Router] Escape user id in Milvus memory filter expressions (injection fix)

### LINKED ISSUE #2138: [Router] Milvus memory filter injection via unescaped user id (CWE-943, cross-tenant read/delete)
## Summary

The Milvus memory backend builds its boolean-expression filters by interpolating the user id with a raw `%s` inside manual quotes, with no escaping:

```go
filterExpr := fmt.Sprintf("user_id == \"%s\"", userID)
```

A user id containing Milvus expression metacharacters breaks out of the quoted string literal and injects filter logic — **CWE-943 (Improper Neutralization of Special Elements in a Query)**.

## Affected sites (`pkg/semantic-router/pkg/memory/`)

| File | Function | Operation |
|---|---|---|
| `milvus_store_retrieve.go:87` | `retrieveFilterExpr` | semantic retrieve |
| `milvus_store_read.go:70` | `List` | list memories |
| `milvus_store_mutate.go:157` | `ForgetByScope` | **delete** |
| `milvus_store_mutate.go:196` | `forgetByScopeWithQuery` | **delete** |

## Impact

With a crafted user id such as `x" || user_id != "y`, the expression becomes:

```
user_id == "x" || user_id != "y"
```

- **Read paths** (retrieve/list): the `|| user_id != ...` clause matches **other users' memories** → cross-tenant memory read.
- **Forget paths** (delete): the same injection in a `Delete(...)` filter can **erase other users' memories**.

The user id is attacker-influenced in **dev builds** (the untrusted body `user` field / `metadata["user_id"]` fallback in `user_id_dev.go`). Even when it comes from the trusted auth header in production builds, escaping is correct defense-in-depth, and an unescaped `"` in a legitimate id silently corrupts the filter.

## Why this is clearly an oversight

The sibling backends already escape:
- **Valkey** memory backend uses `valkeyEscapeTagValue(opts.UserID)` in every filter.
- The same Milvus file escapes memory-type values via `%q` in `buildTypeFilter` — only the user id was interpolated raw.

## Fix (PR attached)

Route all four sites through a shared `milvusUserScopeFilter` helper that renders the id with `%q`, confining it to a properly escaped string literal. Unit test asserts a break-out payload is neutralized (operators stay inside the quoted literal) and that backslashes are escaped.

## Follow-up (not in this PR)

`project_id` scoping in `forgetByScopeWithQuery` filters in memory after a query rather than in the expression, so it is not an injection sink today; worth auditing if it ever moves into the filter string.

theohsiung:
## What

The Milvus memory backend interpolated the user id into its filter expressions with a raw `%s` inside manual quotes — `fmt.Sprintf("user_id == \"%s\"", userID)` — at four sites (retrieve, list, and two forget/delete paths). A crafted id breaks out of the quoted literal and injects filter logic (CWE-943): cross-tenant memory read on the read paths, cross-tenant **delete** on the forget paths.

Fixes #2138.

## Fix

- New `milvusUserScopeFilter(userID)` helper renders the id with `%q`, confining it to an escaped string literal.
- Applied at all 4 sites (`milvus_store_retrieve.go`, `milvus_store_read.go`, `milvus_store_mutate.go` ×2).
- Matches the `%q` convention `buildTypeFilter` already uses and the `valkeyEscapeTagValue` escaping the Valkey backend applies.

## Test plan

- [x] `milvus_filter_test.go`: benign id → plain clause; break-out payload `x" || user_id != "y` → operators escaped inside the literal (not emitted at expression level); backslash escaped. RED before fix → GREEN.
- [x] `go build ./pkg/memory/` clean; `go vet` clean; `golangci-lint` (repo config, incl. gosec) → 0 issues.
- [ ] No live Milvus in this environment; fix is verified by unit test at the filter-construction layer and by parity with the already-escaping Valkey backend.

## Notes
- DCO signed-off. Threat model detailed in #2138 (dev-build untrusted user id; defense-in-depth in prod).


AayushSaini101:
Thanks for the MR @theohsiung QQ: why we cannot use regexp to verify the userID instead of passing the milvusUserScopeFilter() i guess using regexp is more readable as compared to using the parsefilter ? thoughts @theohsiung 
```
var validUserID = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
if !validUserID.MatchString(userID) {
    return errors.New("invalid user id")
}
```

theohsiung:
Hi @AayushSaini101, thx for the suggestion!

I did consider the regex-allowlist route. The catch is what user_id can legitimately be — it comes from x-authz-user-id, injected by the auth layer, and the supported backends emit ids outside ^[a-zA-Z0-9_-]+$:
   - oauth2-proxy (x-forwarded-user) → emails (@, .)
   - Envoy Gateway JWT (x-jwt-sub) → subs like auth0|abc123 (|)
   - Authorino → K8s Secret names (DNS subdomain, allows .)
   
We can widen the pattern to admit those (and \p{L}\p{N} for non-ASCII emails/subs), but once it covers every char escaping already handles, it's the same charset — just expressed as "reject" instead of "sanitize", plus the ongoing burden of tracking each auth backend's id format. A too-narrow allowlist also turns the injection fix into an availability regression for real users. I also leaned on escaping for consistency: the memory subsystem already neutralizes ids this way — Valkey uses valkeyEscapeTagValue() and buildTypeFilter uses %q — so milvusUserScopeFilter mirrors the existing convention. It's the canonical injection defense too (same idea as a parameterized query): it confines the id to a single literal without constraining what a valid identity looks like.

AayushSaini101:
> Hi @AayushSaini101, thx for the suggestion!
> 
> I did consider the regex-allowlist route. The catch is what user_id can legitimately be — it comes from x-authz-user-id, injected by the auth layer, and the supported backends emit ids outside ^[a-zA-Z0-9_-]+$:
> 
> * oauth2-proxy (x-forwarded-user) → emails (@, .)
> * Envoy Gateway JWT (x-jwt-sub) → subs like auth0|abc123 (|)
> * Authorino → K8s Secret names (DNS subdomain, allows .)
> 
> We can widen the pattern to admit those (and \p{L}\p{N} for non-ASCII emails/subs), but once it covers every char escaping already handles, it's the same charset — just expressed as "reject" instead of "sanitize", plus the ongoing burden of tracking each auth backend's id format. A too-narrow allowlist also turns the injection fix into an availability regression for real users. I also leaned on escaping for consistency: the memory subsystem already neutralizes ids this way — Valkey uses valkeyEscapeTagValue() and buildTypeFilter uses %q — so milvusUserScopeFilter mirrors the existing convention. It's the canonical injection defense too (same idea as a parameterized query): it confines the id to a single literal without constraining what a valid identity looks like.

Got it thanks for explanation @theohsiung Looks good to me overall thanks

AayushSaini101 (approved):
LGTM : )

theohsiung:
@rootfs @Xunzhuo — friendly nudge 🙏 Approved (by @AayushSaini101) and `MERGEABLE` — only `BEHIND` main, no conflicts (Mergify should update it at merge time). Could you do the final merge when you have a moment? It's a small, self-contained injection fix (escapes the user id in Milvus memory filter expressions, CWE-943). Thanks!

PR #2141, opened 2026-06-10
[Router] Preserve reasoning_content when caching streaming responses

### LINKED ISSUE #2140: [Router] Streaming semantic cache drops reasoning_content (cache hit returns less than the live stream)
## Summary

When a **streaming** response from a reasoning model is cached, the router drops `reasoning_content`. A later cache hit for a semantically-similar request then returns a response **without** the reasoning the original live stream delivered — a silent fidelity loss. The non-streaming path does not have this problem (it caches the raw upstream body verbatim).

## Root cause

`pkg/extproc/processor_res_body_streaming.go` `extractStreamingContent` accumulates only `delta.content`:

```go
if content, ok := delta["content"].(string); ok && content != "" {
    ctx.StreamingContent += content
}
// delta.reasoning_content is never read
```

`pkg/extproc/processor_res_cache.go` `buildReconstructedStreamingResponse` then builds the cached `message` from `content` (+ optional `tool_calls`) only — no `reasoning_content`. So the reconstructed `chat.completion` that gets written to the cache has no reasoning.

`reasoning_content` is a recognized field elsewhere in the codebase — `looper/client.go` extracts it from both `choices[].reasoning_content` and `choices[].message.reasoning_content`, and it is handled in memory extraction and the Anthropic outbound path — confirming it is a real, supported field, not an unknown extra.

## Impact

- Reasoning models (qwen3 `enable_thinking`, deepseek, gpt-oss families — the router itself injects the reasoning trigger) stream `delta.reasoning_content`.
- First request streams correctly (the router forwards SSE chunks as-is, so the live client sees the reasoning).
- The reconstructed cache entry omits it, so every subsequent **cache hit** serves a degraded response missing the reasoning — inconsistent with both the live stream and the non-streaming cache.

## Fix (PR attached)

Accumulate `delta.reasoning_content` into a new `ctx.StreamingReasoning` and emit it as `message.reasoning_content` in the reconstructed response when non-empty (omitted when no reasoning was streamed).

theohsiung:
## What

Caching a **streaming** reasoning-model response dropped `reasoning_content`: the streaming accumulator captured only `delta.content`, so the reconstructed `chat.completion` written to the cache had no reasoning. A later cache hit then returned a response missing the reasoning the original live stream delivered (the non-streaming cache preserves it, since it stores the raw upstream body).

Fixes #2140.

## Fix

- `request_context.go`: add `StreamingReasoning`.
- `processor_res_body_streaming.go`: accumulate `delta.reasoning_content` alongside `delta.content`.
- `processor_res_cache.go`: emit `message.reasoning_content` in the reconstructed response when non-empty.

`reasoning_content` is already a recognized field (looper extraction reads `choices[].reasoning_content` and `.message.reasoning_content`; memory + anthropic-outbound handle it).

## Test plan

- [x] `processor_res_cache_reasoning_test.go`: accumulator captures `delta.reasoning_content` across chunks; reconstruction includes `message.reasoning_content`; reconstruction omits the field when no reasoning was streamed. RED before fix → GREEN.

## Notes
- DCO signed-off.
- Out of scope (separate): multi-choice (n>1) streaming is reconstructed as a single merged choice — a distinct fidelity gap to address separately.


theohsiung:
Hi @AayushSaini101, appreciate you spending your weekend time on this!  hahaha 🥹

Xunzhuo (approved):
LGTM

PR #2143, opened 2026-06-10
[Router] Validate semantic_cache similarity_threshold range

### LINKED ISSUE #2142: [Router] semantic_cache similarity_threshold is unvalidated; >1.0 silently disables caching
## Summary

The semantic-cache `similarity_threshold` is a cosine similarity and must be in `[0.0, 1.0]`, but it is never validated. An out-of-range value is silently accepted and reaches the runtime unchanged. A threshold `> 1.0` can never be reached by any cosine similarity, so **the cache never hits and caching is silently disabled despite `enabled: true`**.

## Reproduction

```yaml
global:
  stores:
    semantic_cache:
      enabled: true
      similarity_threshold: 1.5   # accepted at load; cache then never hits
```
or per-decision:
```yaml
decisions:
  - name: route
    plugins:
      - type: semantic-cache
        configuration:
          enabled: true
          similarity_threshold: 2.0   # accepted; this decision never caches
```

## Root cause

- `SemanticCache.SimilarityThreshold` (`pkg/config/runtime_config.go`) and the per-decision `SemanticCachePluginConfig.SimilarityThreshold` (`pkg/config/plugin_config.go`) are parsed with no range check.
- The runtime accessors do not clamp: `GetCacheSimilarityThreshold` / `GetCacheSimilarityThresholdForDecision` return the configured value verbatim, and the in-memory backend compares `bestSimilarity >= threshold` (cosine ≤ 1.0), so `>1.0` → always miss.

## Inconsistency

The RAG plugin already enforces this exact bound via `validateRAGSimilarityThreshold` (`pkg/config/rag_plugin_validation.go`: "similarity threshold must be between 0.0 and 1.0"). The semantic cache simply lacks the equivalent check.

## Fix (PR attached)

Add a `semantic_cache` contract validator that rejects an out-of-range `similarity_threshold` on both the global block and each decision's semantic-cache plugin.

`ttl_seconds` and `max_entries` are intentionally **not** range-checked: `0`/negative carry meaning (no-expiry / unlimited) in the cache backends, so rejecting them would be a false positive.

theohsiung:
## What

The semantic-cache `similarity_threshold` (cosine, must be in [0,1]) was never validated. A value > 1.0 is accepted at load and returned verbatim at runtime (no clamp), so the cache never hits — caching is silently disabled despite `enabled: true`.

Fixes #2142.

## Fix

Add a `semantic_cache` contract validator (wired into `sharedConfigContractValidators`) that rejects an out-of-range `similarity_threshold` on:
- the global `semantic_cache` block (`cfg.SemanticCache.SimilarityThreshold`), and
- each decision's `semantic-cache` plugin (`decision.GetSemanticCacheConfig().SimilarityThreshold`).

Mirrors the bound the RAG plugin already enforces (`validateRAGSimilarityThreshold`). An unset threshold stays valid (runtime falls back to a default).

`ttl_seconds`/`max_entries` are intentionally **not** range-checked — `0`/negative carry meaning (no-expiry / unlimited) in the backends, so rejecting them would be a false positive.

## Test plan

- [x] `validator_semantic_cache_test.go`: global threshold range table (unset/0/0.8/1.0 ok; 1.5/-0.1 rejected); per-decision out-of-range rejected and error names the decision; nil config ok. RED before fix → GREEN.
- [x] Full `pkg/config` suite green — no shipped config trips the new check (verified all repo `similarity_threshold` values are within [0,1]).

## Notes
- DCO signed-off. Same validator-hardening family as #2115 / #2123.

### DIFF
--- a/src/semantic-router/pkg/config/validator.go
+++ b/src/semantic-router/pkg/config/validator.go
@@ -71,6 +71,11 @@ var (
 			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
 		},
 		{
+			name:     "semantic_cache",
+			validate: validateSemanticCacheContracts,
+			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
+		},
+		{
 			name:     "embedding",
 			validate: validateEmbeddingContracts,
 			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
--- a/src/semantic-router/pkg/config/validator_semantic_cache.go
+++ b/src/semantic-router/pkg/config/validator_semantic_cache.go
@@ -1,0 +1,50 @@
+package config
+
+import "fmt"
+
+// validateSemanticCacheContracts validates the semantic-cache similarity
+// threshold wherever it can be configured: the global semantic_cache block and
+// each decision's semantic-cache plugin.
+//
+// The threshold is a cosine similarity, which is bounded by [0.0, 1.0]. A value
+// outside that range is silently accepted today and reaches the runtime
+// unchanged (GetCacheSimilarityThreshold / GetCacheSimilarityThresholdForDecision
+// do not clamp it). A threshold > 1.0 can never be reached by any similarity, so
+// the cache never hits despite enabled: true — caching is silently disabled.
+// This mirrors the bound the RAG plugin already enforces
+// (validateRAGSimilarityThreshold).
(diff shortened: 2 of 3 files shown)


Xunzhuo (approved):
LGTM, can you check if we need to harden the validator somewhere else?

PR #2144, opened 2026-06-10
[Router] Normalize routing signal type usage checks

WUKUNTAI-0211:
## What

Routing-signal usage detection now normalizes direct rule `type` values before comparing them with the requested signal type. This makes direct signal detection consistent with runtime rule evaluation and with the projection usage path, which already compares signal types case-insensitively.

## Why

`UsesSignalTypeInRouting` normalizes the requested signal type, but `collectSignalNames` compared `RuleNode.Type` byte-for-byte. A config using a valid signal type with different casing or surrounding whitespace, such as `type: PII`, could still be evaluated by runtime routing while the preload/usage check returned false. For safety signals, that can skip the PII or jailbreak mapping preload even though routing depends on those signals.

## Fix

- Normalize `Decision.HasSignalType` inputs with `strings.ToLower(strings.TrimSpace(...))`.
- Normalize direct rule node types inside `collectSignalNames`.
- Add a regression test covering mixed-case and whitespace-padded PII/jailbreak rule types.

## Test plan

- RED before fix: `GOCACHE=/tmp/semantic-router-go-build go test ./pkg/config -run TestNeedsCoreMappingsForRouting -count=1`
- GREEN after fix: `GOCACHE=/tmp/semantic-router-go-build go test ./pkg/config -run TestNeedsCoreMappingsForRouting -count=1`
- Package check: `GOCACHE=/tmp/semantic-router-go-build go test ./pkg/config -count=1`

### DIFF
--- a/src/semantic-router/pkg/config/routing_signal_usage.go
+++ b/src/semantic-router/pkg/config/routing_signal_usage.go
@@ -5,7 +5,11 @@ import "strings"
 // HasSignalType returns true if the decision's rules tree references at least
 // one signal of the given type (e.g., "jailbreak", "pii").
 func (d *Decision) HasSignalType(signalType string) bool {
-	return len(collectSignalNames(&d.Rules, signalType)) > 0
+	normalizedType := strings.ToLower(strings.TrimSpace(signalType))
+	if normalizedType == "" {
+		return false
+	}
+	return len(collectSignalNames(&d.Rules, normalizedType)) > 0
 }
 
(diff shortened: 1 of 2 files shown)


WUKUNTAI-0211:
Hi @FAUST-BENCHOU, when you have a moment could you take a look at this PR? CI is green. Thanks!

Xunzhuo (approved):
LGTM Thanks

PR #2145, opened 2026-06-10
[Router] Re-normalize truncated FFI embeddings

WUKUNTAI-0211:
## What

Re-normalize FFI embeddings after target-dimension truncation in the generic Qwen3/Gemma embedding paths and the batched raw embedding path.

## Why

The FFI helpers accepted a `target_dim` and truncated the returned vector with a prefix slice, but did not restore the vector to unit L2 norm. Downstream scoring treats embeddings as already normalized and computes cosine similarity as a dot product, so truncated vectors with norm below 1.0 shrink all similarity scores.

Other matryoshka paths in the repo already document and apply the same rule: after truncation, the L2 norm is no longer 1.0, so the prefix must be re-normalized.

## Fix

- Add a shared `truncate_embedding_to_dimension` helper.
- Use it in the generic single-text helper for Qwen3/Gemma.
- Use it in the generic batch helper for Qwen3/Gemma.
- Use it in the continuous-batching raw embedding path.
- Add a unit test proving a truncated unit-norm embedding remains unit-normalized.

## Test plan

- RED before fix: `cargo test --no-default-features ffi::embedding_test::test_truncate_embedding_renormalizes_prefix`
- GREEN after fix: `cargo test --no-default-features ffi::embedding_test::test_truncate_embedding_renormalizes_prefix`
- Formatting: `cargo fmt --check`
- Diff hygiene: `git diff --check`


WUKUNTAI-0211:
Hi @FAUST-BENCHOU, when you have a moment could you take a look at this PR? CI is green. Thanks!

Xunzhuo:
cc @shraderdm

shraderdm:
Thanks for the ping @Xunzhuo. Happy to take this one. It sits right next to the FFI allocation work in #2163 (boxed-slice free-layout fix), so the area is fresh for me. Reviewing now and will post findings shortly.

shraderdm (commented):
Thanks for this, @WUKUNTAI-0211 - nice catch, and re-normalizing after Matryoshka truncation is the right call. I pulled the branch down, ran it locally, and traced the paths. Notes below.

### What's good
- **Correct root cause.** Truncating a unit-norm vector by a prefix slice leaves `||v|| < 1`, and the embedding-signal scoring path treats vectors as already normalized - `cosineSimilarity` in `embedding_classifier_scoring.go` returns a raw dot product ("Assumes vectors are normalized") - so un-renormalized truncated vectors shrink every score. Re-normalizing at the truncation site fixes it where it originates.
- **Consistent with the existing convention.** This matches the model-level paths: Gemma `matryoshka_forward` re-normalizes after `narrow`, and the mmBERT / multimodal `encode_*` paths call `l2_normalize` after `narrow`. The FFI paths were the odd ones out, and this brings them into line.
- **Good consolidation.** One `truncate_embedding_to_dimension` helper replacing three slightly-divergent inline copies, and dropping the now-unused `embeddings.dim(1)` call in the batch path is a nice touch.
- **The test genuinely falsifies - I confirmed it locally.** Green with the fix; when I temporarily remove the re-normalization the test fails for the right reason (`left: [0.6]`, `right: [1.0]`). A real red/green regression test is the best part of this PR.
- Edge cases read correctly: `None` and `dim >= len` pass through untouched (no spurious renorm), `dim > len` degrades gracefully with a warning, and the `norm > 0.0` guard avoids a divide-by-zero on a zero prefix.

### Findings

#### Minor
- **The over-dimension warning now fires once per row instead of once per batch.** The old batch path computed `actual_dim` once and emitted the warning a single time; the refactor calls the helper per element, so a batch of N with `target_dim > model_dim` now prints N identical warnings. Easy fix: keep the `dim > len` check/warning once at the batch level and pass the clamped dim into the per-row helper, or only `eprintln!` from the single-item path.

### Verdict
Looks correct to me - right root cause, applied to the live Qwen3/Gemma single/batch and `get_embedding_batched` paths, no double-normalization, and the `get_embedding_batched` allocation is unchanged so there's no regression to the FFI free path. I'll defer the formal approval to the maintainers, but from my read this is good to go pending the small per-row warning tidy-up. I also turned up a couple of things while reviewing that are out of scope for this PR (a CI-coverage gap and a parity issue in the ONNX binding) - I'll file those separately and link back here.

shraderdm:
Filed the two out-of-scope items I mentioned while reviewing: #2182 (candle-binding Rust unit tests aren't run in PR CI) and #2183 (onnx-binding `maybe_truncate` missing re-normalization, same class as this fix). Neither blocks this PR.

WUKUNTAI-0211:
Thanks, addressed. The batch path now clamps the requested dimension once before applying per-row truncation, so the over-dimension warning is emitted once per batch while preserving the shared truncate + renormalize helper.

shraderdm (approved):
Thanks for turning this around so quickly, @WUKUNTAI-0211. The per-batch clamp is exactly the right shape, and it resolves the one item from my earlier pass.

I pulled `cd72f81` down, re-ran it locally, and traced the over-dimension path end to end.

### What's good

- **The over-dimension warning now fires exactly once per batch.** Hoisting the `dim > model_dim` check to batch scope and clamping `target_dim` to the model width before the per-row map means the per-row helper's own guard is never reached on the batch path, so a batch of N with an over-sized request logs one line instead of N. That is precisely the fix.
- **No drift in the numeric output.** The in-range path still passes `Some(dim)` straight through, so each row truncates and re-normalizes exactly as before; the over-dimension path clamps to the full width and the per-row helper returns each vector untouched (full vector, already unit-norm, no double-normalization). The embedding callers see byte-identical output.
- **`truncate_embedding_to_dimension` stays the single source of truth** for the truncate-and-renormalize rule, and the renormalization regression test still passes green locally.

### Verdict

Looks correct and complete: the per-row warning is resolved, the numeric path is unchanged, and CI is green. Approving.

Optional and non-blocking, only if you want to harden it later: the batched over-dimension path is not directly unit-tested (the existing test exercises the per-row helper). A small case, a few rows with `target_dim` above the model width asserting each row comes back full-length and unchanged, would lock the equivalence. Not needed for this PR.

PR #2105, opened 2026-06-09
[Dashboard] Surface RAG retrieval metrics in Grafana dashboard + docs

Peterren:
## What

Surfaces the existing RAG (Retrieval-Augmented Generation) observability metrics in the bundled Grafana dashboard and documents them in the RAG plugin tutorial.

This addresses the **"Comprehensive metrics and observability for RAG operations"** acceptance item in #155, which is currently marked _Partial_. The `rag_*` Prometheus metrics and the `semantic_router.rag.retrieval` tracing span already exist (added with #1152 / #1533) but were **charted nowhere and documented nowhere** — so operators had no way to see them. This PR closes that surface gap.

Per the split agreed in the #155 thread, this is the first slice of the "metrics half".

## Design

Scope is intentionally **surface-only**: no new metrics, no tracing changes, no router behavior change. This keeps the blast radius narrow and the change purely additive/reviewable.

- **Dashboard** is generated, not hand-edited. A new `append_rag_section()` (row `panel_id=700`) was added to `grafana_dashboard_sections.py`, following the existing `append_*` section pattern and reusing the existing panel factories (`create_row_panel`, `create_stat_panel`, `create_timeseries_panel`, `create_target`). Every PromQL expression references only metrics/labels that exist in `rag_metrics.go` (`backend`, `decision`, `status`; `_bucket` suffix only on the two histograms). The committed `llm-router-dashboard.serve.json` was regenerated via `generate_dashboard.py` (the only tracked copy).
- **Docs**: an `## Observability` section was added to the current, layer-aligned RAG plugin doc (`website/docs/tutorials/plugin/rag.md`), not to the frozen v0.x versioned snapshots.

### New "RAG Retrieval Metrics" dashboard row (8 panels)

| Panel | Source metric(s) |
| --- | --- |
| RAG Retrieval Rate | `rag_retrieval_attempts_total` |
| RAG Success Rate | `rag_retrieval_attempts_total{status="success"}` / total |
| RAG Cache Hit Rate | `rag_cache_hits_total` / (hits + `rag_cache_misses_total`) |
| RAG Retrieval Latency (P50/P95/P99) by Backend | `rag_retrieval_latency_seconds_bucket` |
| RAG Retrieval Attempts by Status | `rag_retrieval_attempts_total` by `status` |
| RAG Cache Hits/Misses by Backend | `rag_cache_hits_total`, `rag_cache_misses_total` |
| RAG Similarity Score by Backend/Decision | `rag_similarity_score` |
[...]

Issue #155, opened 2025-09-17
RAG-Optimized Routing with On-Demand Retrieval


Xunzhuo:
**Title:** Implement RAG-Optimized Category Routing with On-Demand Knowledge Retrieval

### Overview
Extend the vLLM Semantic Router to include RAG (Retrieval-Augmented Generation) optimization capabilities, enabling intelligent routing decisions based on knowledge retrieval needs and implementing on-demand RAG for queries that would benefit from external knowledge augmentation.

From the latest research via: https://arxiv.org/pdf/2505.10570

When evaluated with several long context models, we observe a performance
+ drop of 7% to 85% as the number of tools increases
+ 7% to 91% degradation in answer retrieval as the tool responses length increases, 
+ 13% and 40% degradation for as multi-turn conversations get longer. 

### Current State
The semantic router currently:
- Routes based on 14 predefined categories using ModernBERT classification
- Uses semantic caching for response reuse
- Supports tool selection based on query similarity
- Has no integrated knowledge retrieval capabilities

#### 1. RAG-Aware Categories
Add new categories specifically designed for knowledge-intensive tasks:

```yaml
categories:
  - name: factual_lookup
    use_reasoning: false
    use_rag: true
    rag_strategy: "always"
    rag_description: "Factual queries requiring up-to-date information"
    knowledge_domains: ["current_events", "statistics", "definitions"]
    
  - name: research_synthesis  
    use_reasoning: true
    use_rag: true
    rag_strategy: "adaptive"
    rag_description: "Complex research requiring multiple sources"
    knowledge_domains: ["academic", "technical", "scientific"]
    
  - name: domain_expertise
    use_reasoning: true
    use_rag: true
    rag_strategy: "conditional"
    rag_description: "Specialized domain knowledge queries"
    knowledge_domains: ["medical", "legal", "financial", "technical"]
```

#### 2. On-Demand RAG Pipeline
Implement intelligent retrieval decision-making:

```go
type RAGDecision struct {
    ShouldRetrieve    bool                   `json:"should_retrieve"`
    RetrievalStrategy string                 `json:"retrieval_strategy"` // "always", "adaptive", "conditional"
    KnowledgeDomains  []string               `json:"knowledge_domains"`
    RetrievalQuery    string                 `json:"retrieval_query"`
    Confidence        float64                `json:"confidence"`
    Reasoning         string                 `json:"reasoning"`
}

type RAGClassificationRequest struct {
    Text                string                 `json:"text"`
[...]

PR #2146, opened 2026-06-10
[Router] Add RAG result-count, truncation, and typed-error metrics

Peterren:
## What

Adds three **additive, non-breaking** Prometheus metrics that give operational depth into RAG retrieval, plus the previously-missing unit tests for the RAG metrics layer. This is the first slice of the "metric depth" follow-up from #2105 (the RAG-observability / metrics half of #155).

| Metric | Type | Labels | Meaning |
| --- | --- | --- | --- |
| `rag_retrieval_results_count` | histogram | `backend`, `decision` | Documents returned per successful retrieval |
| `rag_context_truncations_total` | counter | `backend`, `decision` | Times retrieved context exceeded `max_context_length` and was truncated |
| `rag_retrieval_errors_total` | counter | `backend`, `decision`, `error_type` | Retrieval failures broken down by cause |

No existing metric is changed; `rag_similarity_score` etc. keep their current semantics.

## Design

- **Result count** — the four vector backends (`milvus`, `qdrant`, `vectorstore`, `openai`) set `ctx.RAGResultCount = len(results)` next to where they already set `ctx.RAGSimilarityScore`. `finalizeRAGRetrieval` records it centrally and adds a `rag.result_count` span attribute. It's skipped when 0, so backends that don't report a count (`external_api`, `mcp`) and cache hits don't emit misleading zeros. `hybrid` inherits the count from whichever sub-backend served the request (same pattern as the existing similarity score).
- **Truncation** — recorded centrally in `injectRAGContext` exactly where truncation happens (UTF-8-safe rune trim), so it's consistent across all backends and injection modes.
- **Typed errors** — `classifyRAGError` maps an error to a small, stable `error_type` set (`timeout`, `embedding`, `no_results`, `backend_unavailable`, `config`, `backend_error`, `other`). It's best-effort string inspection used only for the metric label, never for control flow, and it unwraps the orchestrator's `backend %q retrieval failed: %w` wrapping. Recorded alongside the existing coarse `rag_retrieval_attempts_total{status="error"}`.

### Out of scope (intentionally deferred)
- **Dashboard panels + docs** for these metrics — the Grafana generator and `rag.md` are owned by the still-open #2105; panels/docs will follow once it merges to avoid conflicts.
- **Per-stage latency** (embedding vs vector-DB), **collection label** (cardinality), **similarity gauge→histogram** (semantics change) — next slices of metric depth.

## Test Plan

- New `rag_metrics_test.go` covers all RAG recorders (attempts-by-status, cache hit/miss, result-count histogram, truncation counter, typed-error breakdown) using `prometheus/client_golang/testutil` — closes the prior zero-coverage gap.
- New `req_filter_rag_classify_test.go` table-tests `classifyRAGError` across all buckets, including a wrapped-error case.
- `cd src/semantic-router && go build ./pkg/extproc/... ./pkg/observability/metrics/...` — OK; `gofmt` clean.
- `go test ./pkg/extproc/ ./pkg/observability/metrics/` — pass (RAG + classify + metrics suites).
- `make agent-ci-lint` — **exit 0** (Go structural lint, structure checks, full fast-gate test suite).
- Pure observability addition: no routing/startup/config/API behavior change → no E2E updates required.

Refs #155


Peterren:
low risk change, need approval @rootfs

Xunzhuo (commented):
I don't see a separate correctness blocker in the metrics/runtime/test delta. The cited umbrella #155 is closed and was never accepted, and this head conflicts with current main. Please link a current accepted, bounded scope and rebase before another readiness pass.

FAUST-BENCHOU (commented):
Thanks for adding these RAG metrics and improving the observability coverage. The additional retrieval result, truncation, and error metrics provide useful operational visibility without changing existing routing behavior.

PR #2149, opened 2026-06-11
[Router] Preserve structured system content in RAG system_prompt injection

### LINKED ISSUE #2148: [Router] RAG system_prompt injection drops structured system-message content
## Summary

RAG context injection in `system_prompt` mode **silently drops the existing system message's instructions** when that message's content is a structured array of content parts (valid OpenAI input). The user's original system prompt is lost; only the RAG context remains.

This is the same silent-content-loss class as #2127 (the `system_prompt` plugin), but on a **different code path** (`injectAsSystemPrompt` in `pkg/extproc/req_filter_rag.go`) that #2127 did not touch.

## Root cause

`injectAsSystemPrompt` (`pkg/extproc/req_filter_rag.go`):
```go
systemContent, _ = firstMsg["content"].(string)
...
newFirst["content"] = contextPrefix + systemContent
```
When `content` is `[]interface{}` (content parts, e.g. `[{"type":"text","text":"..."}]`), the `.(string)` assertion fails → `systemContent == ""` → the prepend yields `"<rag context>" + ""`, dropping the original instructions.

## Reproduction

RAG plugin with `injection_mode: system_prompt`, request:
```json
{"model":"m","messages":[
  {"role":"system","content":[{"type":"text","text":"ORIGINAL_SYSTEM_INSTRUCTIONS"}]},
  {"role":"user","content":"hi"}
]}
```
Resulting system message content:
```
Context from knowledge base:

RAG_CONTEXT_XYZ

```
`ORIGINAL_SYSTEM_INSTRUCTIONS` is gone. (A plain-string system message is preserved correctly — only the structured form is dropped.)

## Impact

A request that combines a structured system message with RAG `system_prompt` injection loses its system instructions — the model no longer sees the caller's directives, changing behavior silently with no error.

## Fix (PR attached)

Extract text from string **or** structured (`[]content-parts`, types `text`/`input_text`) content via a `ragSystemMessageText` helper before prepending the RAG context. Mirrors the system_prompt plugin's handling (#2127). Plain-string content is unaffected.

theohsiung:
## What

RAG `system_prompt` injection (`injectAsSystemPrompt`) read the existing system message via `firstMsg["content"].(string)`. For a structured (array) system content the assertion fails → `""`, so the prepend dropped the user's original system instructions, keeping only the RAG context.

Fixes #2148.

## Fix

Add `ragSystemMessageText` to extract text from string **or** structured (`[]content-parts`, type `text`/`input_text`) content, and use it before prepending the RAG context. Same silent-content-loss class as #2127 (system_prompt plugin), on the distinct RAG path #2127 didn't touch.

## Test plan

- [x] `req_filter_rag_sysprompt_test.go`: a structured system message + RAG injection preserves both the RAG context and the original instructions. RED before fix (original dropped) → GREEN.
- [x] Plain-string system content unaffected — covered by existing RAG tests; full `pkg/extproc` suite green.

## Notes
- DCO signed-off. Flattening structured content to text mirrors the #2127 approach (system messages are text). Found via an intensive multi-surface audit; this was the one Tier-1 confirmed by reproduction (others triaged out as guarded/by-design/unreachable).


AayushSaini101 (approved):
LGTM thanks : )

Xunzhuo (commented):
Current main's neutral request representation preserves structured system content and prepends the injected text part without flattening it. This branch targets the old map-based implementation and conflicts with main, so the content-loss bug is already superseded by the unified protocol path. Please close it unless a rebased distinct case remains.

FAUST-BENCHOU (commented):
Thanks for fixing this structured system content handling issue. Preserving the original system instructions during RAG injection is an important correctness fix, especially for OpenAI-compatible structured messages.

wilsonwu (approved):
LGTM

drivebyer:
@theohsiung  Please resolve the conflict.

PR #2150, opened 2026-06-11
[Router] Support comma-separated auto model names

### LINKED ISSUE #2116: feature: When Claude Code GUI enables 1M context, it appends `[1M]` to model names (e.g., `MoM[1M]`), but `IsAutoModelName` doesn't recognize the suffix, so requests bypass intelligent routing.
### Describe the feature

# Comma-Separated auto_model_name Support

## Background

Currently `auto_model_name` only accepts a single string value; only an exact match triggers intelligent routing. The default is `"MoM"`.

**Problem**: When Claude Code GUI enables the 1M context window feature, it appends a `[1M]` suffix to the model name, sending `MoM[1M]` to the router backend. However, `IsAutoModelName("MoM[1M]")` returns `false` (because it only matches `"MoM"`), causing the request to bypass intelligent routing and go through direct passthrough instead.

**Existing capability**: `model_config` can already define a model named `MoM[1M]`, but `auto_model_name` is a single value and cannot simultaneously match both `"MoM"` and `"MoM[1M]"`.

## Approach

Keep the `auto_model_name` YAML field as-is (still a `string`), but split on commas internally to match multiple values.

```yaml
# Configuration example
auto_model_name: "MoM, MoM[1M]"
```

This way both `model: "MoM"` and `model: "MoM[1M]"` trigger intelligent routing, while `"auto"` remains permanently valid.

### 1. `src/semantic-router/pkg/config/helper.go` — Core logic

**Delete** the existing `GetEffectiveAutoModelName` and `IsAutoModelName`, **replace with**:

```go
// GetEffectiveAutoModelName returns the primary auto model name for automatic model selection.
// Returns the first configured AutoModelName if set, otherwise defaults to "MoM".
func (c *RouterConfig) GetEffectiveAutoModelName() string {
	if c.AutoModelName != "" {
		return c.AutoModelName
	}
	return "MoM"
}

// GetEffectiveAutoModelNames returns all effective auto model names for automatic model selection.
// Supports comma-separated values in AutoModelName (e.g. "MoM, MoM[1M], router").
// Always includes "auto" for backward compatibility.
func (c *RouterConfig) GetEffectiveAutoModelNames() []string {
	raw := c.AutoModelName
	if raw == "" {
		raw = "MoM"
	}
	parts := strings.Split(raw, ",")
	names := make([]string, 0, len(parts)+1)
	names = append(names, "auto")
	for _, p := range parts {
		p = strings.TrimSpace(p)
		if p != "" && p != "auto" {
			names = append(names, p)
		}
	}
	return names
}

// IsAutoModelName checks if the given model name should trigger automatic model selection.
// Matches against all configured auto model names (comma-separated), plus "auto" for backward compatibility.
func (c *RouterConfig) IsAutoModelName(modelName string) bool {
	for _, name := range c.GetEffectiveAutoModelNames() {
		if modelName == name {
			return true
		}
	}
	return false
}
```

The `strings` package is already imported in helper.go — no new imports needed.

### 2. `src/semantic-router/pkg/extproc/req_filter_models.go` — /v1/models endpoint

**Changes**: Lines ~37 and ~62, replace `r.Config.GetEffectiveAutoModelName()` with `IsAutoModelName` for the dedup check.

Around line 37, when building the auto-model entry (keep as-is, use the first name as ID):

```go
// Current (~line 37):
effectiveAutoModelName := r.Config.GetEffectiveAutoModelName()
models = append(models, OpenAIModel{
    ID:          effectiveAutoModelName,
    ...

// Replace with:
models = append(models, OpenAIModel{
    ID:          r.Config.GetEffectiveAutoModelNames()[1], // index 0 is "auto", index 1 is first configured name
    ...
```

Around line 62, when skipping duplicates:

```go
// Current (~line 62):
if m == r.Config.GetEffectiveAutoModelName() {
    continue
}

// Replace with:
if r.Config.IsAutoModelName(m) {
    continue
}
```

### 3. `src/semantic-router/pkg/apiserver/route_models.go` — API server model list

Same changes as req_filter_models.go.

Around line 24:

```go
// Current:
effectiveAutoModelName := cfg.GetEffectiveAutoModelName()

// Replace with GetEffectiveAutoModelNames()[1]:
models = append(models, OpenAIModel{
    ID: cfg.GetEffectiveAutoModelNames()[1],
    ...
```

Around line 47:

```go
// Current:
if m == cfg.GetEffectiveAutoModelName() {
    continue
}

// Replace with:
if cfg.IsAutoModelName(m) {
    continue
}
```

### 4. `src/semantic-router/pkg/config/config.go` — Add comment

Line 170, the field definition stays the same, just add a clarifying comment:

```go
type RouterOptions struct {
	// AutoModelName supports comma-separated names to trigger automatic routing
	// (e.g. "MoM, MoM[1M]"). The first name is the primary one used in /v1/models.
	AutoModelName             string               `yaml:"auto_model_name,omitempty"`
	...
```

### 5. `src/semantic-router/pkg/config/config_test.go` — Update tests

Add a new Context inside the existing `Describe("AutoModelName Configuration", ...)` block:

```go
Context("comma-separated AutoModelName", func() {
    It("should match all comma-separated names", func() {
        cfg := &RouterConfig{
            RouterOptions: RouterOptions{
                AutoModelName: "MoM, MoM[1M], router",
            },
        }
        Expect(cfg.GetEffectiveAutoModelNames()).To(Equal([]string{
            "auto", "MoM", "MoM[1M]", "router",
        }))
        Expect(cfg.IsAutoModelName("auto")).To(BeTrue())
        Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
        Expect(cfg.IsAutoModelName("MoM[1M]")).To(BeTrue())
        Expect(cfg.IsAutoModelName("router")).To(BeTrue())
        Expect(cfg.IsAutoModelName("gpt-4")).To(BeFalse())
    })

    It("should handle whitespace in comma-separated values", func() {
        cfg := &RouterConfig{
            RouterOptions: RouterOptions{
                AutoModelName: "  MoM  ,  MoM[1M]  ",
            },
        }
        Expect(cfg.GetEffectiveAutoModelNames()).To(Equal([]string{
            "auto", "MoM", "MoM[1M]",
        }))
    })
})
```

### 6. `src/semantic-router/pkg/extproc/processor_req_body_streamed_test.go` — No changes needed

This file sets `cfg.AutoModelName = autoModel` (single value) and works correctly with the existing logic.

### 7. `src/semantic-router/pkg/config/canonical_export.go` / `canonical_global.go` — No changes needed

These files only do string passthrough (`cfg.AutoModelName = global.Router.AutoModelName`). All split logic lives in helper.go.

## Usage

```yaml
# config.yaml (canonical v0.3)
global:
  router:
    auto_model_name: "MoM, MoM[1M]"   # comma-separated
```

Or legacy format:

```yaml
router_options:
  auto_model_name: "MoM, MoM[1M]"
```

After configuration, any of the following model names trigger intelligent routing:
- `"auto"` — permanently built-in
- `"MoM"` — default / first configured name
- `"MoM[1M]"` — second comma-separated name

## Verification

```bash
# Unit tests
cd src/semantic-router && go test ./pkg/config/... -run "AutoModelName" -v

# End-to-end: after configuring auto_model_name: "MoM, MoM[1M]"
# 1. curl with model=MoM → should trigger intelligent routing
# 2. curl with model=MoM[1M] → should trigger intelligent routing (previously failed)
# 3. curl with model=gpt-4o → should go through direct passthrough
```

### Primary layer

signal

### Why this layer?

_No response_

_No response_

### Additional context

_No response_

WUKUNTAI-0211:
## Summary

- support comma-separated `auto_model_name` values while keeping the YAML field as a string
- keep `auto` as a built-in alias and expose the first configured name as the primary `/v1/models` ID
- skip all configured auto-model aliases when appending config models to `/v1/models`

Closes #2116

## Testing

- `GOCACHE=/tmp/semantic-router-go-build go test ./pkg/config -run TestConfig -count=1`
- `git diff --check`

## Notes

I also attempted compile checks for `./pkg/apiserver ./pkg/extproc`, but this local environment is missing native binding artifacts such as `libcandle_semantic_router`, so those packages fail at link time before running tests.

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_models.go
+++ b/src/semantic-router/pkg/apiserver/route_models.go
@@ -21,9 +21,8 @@ func (s *ClassificationAPIServer) handleOpenAIModels(w http.ResponseWriter, _ *h
 
 	// Add the effective auto model name (configured or default "MoM")
 	if cfg != nil {
-		effectiveAutoModelName := cfg.GetEffectiveAutoModelName()
 		models = append(models, OpenAIModel{
-			ID:          effectiveAutoModelName,
+			ID:          cfg.GetEffectiveAutoModelName(),
 			Object:      "model",
 			Created:     now,
 			OwnedBy:     "vllm-semantic-router",
@@ -44,7 +43,7 @@ func (s *ClassificationAPIServer) handleOpenAIModels(w http.ResponseWriter, _ *h
 	if cfg != nil && cfg.IncludeConfigModelsInList {
 		for _, m := range cfg.GetAllModels() {
 			// Skip if already added as the configured auto model name (avoid duplicates)
-			if m == cfg.GetEffectiveAutoModelName() {
+			if cfg.IsAutoModelName(m) {
 				continue
 			}
 			models = append(models, OpenAIModel{
--- a/src/semantic-router/pkg/config/config.go
+++ b/src/semantic-router/pkg/config/config.go
@@ -167,6 +167,8 @@ type LLMObservability struct {
 }
 
 type RouterOptions struct {
+	// AutoModelName supports comma-separated names to trigger automatic routing.
+	// The first name is the primary model ID returned by /v1/models.
 	AutoModelName             string               `yaml:"auto_model_name,omitempty"`
 	IncludeConfigModelsInList bool                 `yaml:"include_config_models_in_list,omitempty"`
 	ClearRouteCache           bool                 `yaml:"clear_route_cache"`
--- a/src/semantic-router/pkg/config/config_test.go
+++ b/src/semantic-router/pkg/config/config_test.go
@@ -2262,6 +2262,15 @@ api:
 				cfg := &RouterConfig{}
 				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("MoM"))
 			})
+
+			It("should return the first comma-separated AutoModelName as primary", func() {
+				cfg := &RouterConfig{
+					RouterOptions: RouterOptions{
+						AutoModelName: "MoM, MoM[1M], router",
+					},
+				}
+				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("MoM"))
+			})
 		})
 
 		Context("IsAutoModelName", func() {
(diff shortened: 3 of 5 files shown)


WUKUNTAI-0211:
Closing this for now because I could verify the router-side exact-match behavior, but I could not verify the upstream premise that the client actually sends `MoM[1M]` in requests. Happy to revisit this once we have a concrete reproduction or request trace.

PR #2038, opened 2026-06-04
[E2E] add request-preservation tests for Anthropic /v1/messages

siloteemu:
## Summary
- Eight strong-assertion e2e tests verifying that request-side fields survive the Anthropic→Anthropic translation path through the router and reach the `anthropic-shim` backend verbatim. Each test inspects what the shim received via its `/debug/last-request` endpoint (introduced in #1940).
- Registered in `AnthropicShimContract`; runs under the `anthropic-shim` profile.
- Why: complements the contract surface that #1940 deployed by exercising field preservation across the request-translation path.
- Risk: low — test-only addition; cannot affect production code paths.

## Scenarios covered
- **image block survival** — base64 PNG block reaches shim with `type`, `source.type`, `source.media_type`, `source.data` all intact.
- **top_k preservation** — `top_k=40` reaches upstream AND response carries no `x-vsr-lossiness-warnings` (Anthropic-shaped backend must not drop it).
- **metadata.user_id preservation** — survives translation verbatim.
- **tool_result.is_error preservation** — `is_error=true` not stripped during routing.
- **tool_result array content** — multi-block content array forwarded; assertion checks the newline-joined form the shim records.
- **stop_sequences round-trip** — mapped into the OpenAI IR `Stop` field inbound and back to Anthropic `StopSequences` outbound; asserted verbatim.
- **temperature / top_p round-trip** — both numeric sampling fields map through the OpenAI IR (`applyNumericSampling` inbound, `applySampling` outbound) with no scaling; each asserted verbatim at the shim.

## Test plan
- `make agent-lint CHANGED_FILES="<paths>"` (vet, fmt, golangci-lint, codespell): pass locally
- CI on push runs the `anthropic-shim` profile end-to-end against a real kind cluster
- Each test is self-contained (opens its own port-forwards, cleans up via `defer`)

## File layout
- `e2e/testcases/anthropic_messages_request_preservation.go` — eight test functions
- `e2e/testcases/anthropic_messages_request_preservation_helpers.go` — shared types, infra/navigation/assertion helpers, split out to keep the main test file under the project's 800-line cap
- `e2e/pkg/testmatrix/testcases.go` — adds eight test names to `AnthropicShimContract`

## Status
Kept in draft to avoid pulling reviewer attention from in-flight #1941 (streaming emitter). Will mark ready for review after #1941 is reviewed/merged.

PR #2153, opened 2026-06-11
[Router] preserve thinking and tool_choice.disable_parallel_tool_use on Anthropic→Anthropic path

siloteemu:
## Summary
Fixes two same-protocol (Anthropic client → Anthropic backend) **transparency drops**: request parameters captured on inbound but never restored on the outbound request body, so they were silently lost with no signal to the client. On the same-protocol path every client-supplied field should reach the backend verbatim.

Both fields have no OpenAI representation, so neither can ride the OpenAI IR the way temperature/top_p do — they were falling through the translation.

### 1. `thinking` request parameter
- `convertThinking` writes `ext.Thinking` on inbound, but the outbound builder had no step reading it back → extended thinking never enabled on the backend.
- Fix: carry `thinking` on the `AnthropicPassthrough` sidecar (same mechanism as `top_k`/`metadata.user_id`); restore via a new `applyThinking` step mapping `enabled` (with `budget_tokens`) and `disabled` through the SDK's `ThinkingConfig` union.

### 2. `tool_choice.disable_parallel_tool_use`
- Inbound mirrors Anthropic `disable_parallel_tool_use:true` onto OpenAI `parallel_tool_calls:false`, but the outbound `openAIToolChoiceToAnthropic` never restored it → a client asking the backend to serialize tool calls got parallel execution.
- Fix: `applyDisableParallelToolUse` reads `parallel_tool_calls` and sets the flag on whichever ToolChoice variant (auto/any/tool) is active, only when explicitly false (parallel is the Anthropic default).

The `tool_choice` *type* mapping (auto/any/tool, named function) and per-tool `cache_control` already round-trip correctly — verified, no change needed.

## Risk
Low. Both fixes are no-ops when the field is unset (nil passthrough / omitted thinking / parallel default), so all existing callers are byte-identical.

## Verification
Each drop was reproduced before fixing (raw body → parse → rebuild produced no field), then confirmed restored. Unit tests cover:
- thinking: enabled / disabled / omitted / full raw-body round-trip
- disable_parallel: any+disabled / tool+disabled / default-omitted

## Test plan
- `make agent-lint CHANGED_FILES="<paths>"` (vet, fmt, golangci-lint, codespell): pass
- `go test ./pkg/anthropic/`: pass, no regressions

## Notes
Pairs with the e2e transparency suite in #2038 (shim-backed assertions for image/top_k/metadata/tool_result/stop_sequences/temperature/top_p). Shim-backed e2e assertions for `thinking` and `disable_parallel_tool_use` are a natural addition to that suite once both land.

### DIFF
--- a/src/semantic-router/pkg/anthropic/request_body.go
+++ b/src/semantic-router/pkg/anthropic/request_body.go
@@ -87,6 +87,25 @@ func applyMetadata(params *anthropic.MessageNewParams, req *openai.ChatCompletio
 	params.Metadata = anthropic.MetadataParam{UserID: anthropic.String(userID)}
 }
 
+// applyThinking restores the Anthropic extended-thinking request parameter from
+// the passthrough. The thinking config has no OpenAI representation, so without
+// this the same-protocol (Anthropic→Anthropic) path would silently drop it and
+// the backend would not enable extended thinking. nil passthrough or unset
+// Thinking is a no-op, leaving the field omitted.
+func applyThinking(params *anthropic.MessageNewParams, pt *AnthropicPassthrough) {
+	if pt == nil || pt.Thinking == nil {
+		return
+	}
+	switch pt.Thinking.Type {
+	case "enabled":
+		params.Thinking = anthropic.ThinkingConfigParamOfEnabled(pt.Thinking.BudgetTokens)
+	case "disabled":
+		params.Thinking = anthropic.ThinkingConfigParamUnion{
+			OfDisabled: &anthropic.ThinkingConfigDisabledParam{},
+		}
+	}
+}
+
 // applyToolsCacheControl attaches cache_control markers to tool definitions
 // based on the `tools[i]` keys in the passthrough.
 func applyToolsCacheControl(params *anthropic.MessageNewParams, pt *AnthropicPassthrough) {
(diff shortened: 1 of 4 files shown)


src/semantic-router/pkg/anthropic/request_body.go
```diff
@@ -87,6 +87,25 @@ func applyMetadata(params *anthropic.MessageNewParams, req *openai.ChatCompletio
...
+func applyThinking(params *anthropic.MessageNewParams, pt *AnthropicPassthrough) {
+	if pt == nil || pt.Thinking == nil {
+		return
+	}
+	switch pt.Thinking.Type {
+	case "enabled":
+		params.Thinking = anthropic.ThinkingConfigParamOfEnabled(pt.Thinking.BudgetTokens)
+	case "disabled":
+		params.Thinking = anthropic.ThinkingConfigParamUnion{
+			OfDisabled: &anthropic.ThinkingConfigDisabledParam{},
```
FAUST-BENCHOU, line 104: https://platform.claude.com/docs/en/api/messages
better alse support `adaptive`

FAUST-BENCHOU (commented):
other lg

Xunzhuo (changes requested):
The change needs an accepted implementation issue before approval. This currently has no linked issue, so the Anthropic passthrough behavior and release scope have no accepted contract to review against. Please open/link the issue and complete intake first.

FAUST-BENCHOU (commented):
Thanks for fixing these Anthropic request passthrough issues. Preserving `thinking` and `tool_choice.disable_parallel_tool_use` across the same-protocol path is important to avoid silently changing user-requested behavior.

FAUST-BENCHOU (commented):
Thanks for fixing these Anthropic transparency drops. Preserving same-protocol request fields like `thinking` and `tool_choice.disable_parallel_tool_use` is important to avoid silently changing user-requested behavior.

Issue #1835, opened 2026-04-27
bug: router container gets killed by OOM due to unbounded tool call cache


cfr-convoso:
### Describe the bug

I have been running vllm-sr following the standard instructions: `vllm server` spins up the containers.
I noticed that `vllm-sr-router-container` was getting killed fairly reliably, so the first think I did was move away from Docker to Orbstack (lighter!)
The container was still getting killed:

```
docker inspect vllm-sr-router-container --format 'oom={{.State.OOMKilled}} exit={{.State.ExitCode}} error={{.State.Error}}'
oom=true exit=137 error=
```

I noticed that orbstack was allocating 16GB RAM to run containers. I upped the limit to 32GB. After a day, the container got OOM'd again.

Not sure what's happening here?

### To Reproduce

In my case: run vllm-sr, allow hermes agent to talk to it.
I can post my config.yaml file if it helps.

### Affected layer

None

rootfs:
@cfr-convoso thanks for reporting, do you happen to have the docker logs too?

cfr-convoso:
It looks like there is no sensitive information in there, please let me know if I am making a mistake sharing it here.

[router.txt](https://github.com/user-attachments/files/27147850/router.txt)

rootfs:
@cfr-convoso thanks for sharing! Here is analysis from AI:



## What could be going wrong (OOM root cause analysis)

The log reveals a strong candidate for the memory leak, corroborated by several patterns:

### 1. Unbounded `tool_trace_flow` / `tool_trace_step_count` growth — the most likely OOM cause

Look at the `tool_trace_step_count` field in the `router_replay_start` / `router_replay_complete` log entries at the end of the log. It starts at 1 and climbs monotonically:

- Line 148: `tool_trace_step_count: 1` (start of session)
- Line 818: `tool_trace_step_count: 1` (new session, resets)
- Lines 730–810: `tool_trace_step_count: 95 → 96 → 97 → 98 → 99 → 100 → 101 → 102 → 103 → 104 → 105 → 106 → 107 → 108 → 109 → 110 → 111 → 112`

And the accompanying `tool_trace_flow` string is literally serialized as a growing chain appended each turn:

```
"User Query -> LLM Tool Call -> Client Tool Result -> LLM Tool Call -> ... (112 steps)"
```

[...]

PR #2155, opened 2026-06-11
[Router] fix router_replay history 413 on large requests (bound list response + safer default)

### LINKED ISSUE #2189: bug: Failed to fetch insight records: 413 Request Entity Too Large
### Describe the bug

Image

### To Reproduce

This vsr.public.router_replay_recordstable seems to throw an error whenever the request data gets a bit too large.

### Expected behavior

_No response_

### Affected layer

None

### Additional context

_No response_

siloteemu:
## Problem
`GET /v1/router_replay` (the replay **history list**) returns **413 Payload Too Large** once a large request has been recorded — reproduced by sending a ~5–10 MB input and then opening the history view. One big record breaks the whole list for everyone.

## Root cause
The list response embeds whole records, and several fields hold free text from the **full** request/response (`prompt`, `tool_definitions`, tool-trace step text, `request_body`/`response_body`). A page of records is serialized into one ext-proc message; past the 4 MB cap it hard-fails with 413. Raw bodies were capped (`max_body_bytes`), but the structured fields were bounded only by `max_tool_trace_bytes`, which **defaulted to 0 (unlimited)** — so one large input, or a long agent tool-trace, produced a multi-MB page.

## Fix
**1. List endpoint never 413s a page (the actual fix).** When a page would exceed the cap, it progressively trims the heaviest text fields on the page records — tool-trace text → bodies → prompt/tool_definitions — and sets the existing `*_truncated` flags so reduced fidelity is visible. Page records are per-request **clones**, so the stored originals are untouched and full fidelity stays available via `GET /v1/router_replay/{id}`. This is what makes the endpoint correct under **any** capture config — including the two cases a storage cap alone can't fix: an operator setting `max_tool_trace_bytes: 0`, and a long agent tool-trace under default config.

**2. Safer default (defense-in-depth).** `max_tool_trace_bytes` now defaults to 4096 — the missing sibling of the existing bounded defaults (`max_body_bytes=4096`, `max_tool_trace_steps=100`, the latter added for OOM-prevention in #1835). It bounds what is *stored* per record (a memory safeguard); `0 = unlimited` stays as an explicit opt-out. Not sufficient alone, which is why layer 1 exists.

## Verification
Reproduced the 413 with a 5 MB input through the real recorder config path, then confirmed a **200** with trimmed + `*_truncated` fields for all three cases: large body, explicit `max_tool_trace_bytes:0`, and a 100-step agent tool-trace under default config.

## Tests
- `pkg/extproc`: oversized page now returns 200 (replaces the old 413-asserting test); plus opt-out-unlimited, long-agent-tool-trace, and default-config large-input cases.
- `pkg/config`: pins the `max_tool_trace_bytes=4096` default.

## Risk
Low. Under extreme sizes the list may return truncated heavy fields (flagged via `*_truncated`; full record via the per-record endpoint). Explicit configs are unaffected; the `0` opt-out is preserved.

## Follow-up
This PR is the immediate, non-breaking fix (bound the list response so it never 413s; full fidelity stays on `GET /v1/router_replay/{id}`). The deeper design issue — the list endpoint embedding full record bodies at all — is tracked in #2157 for maintainers to decide whether to advance (summaries-only list, or a projection param). Happy to own that follow-up if wanted.

---

Closes #2189 (same `413` on the replay history list from a large recorded request; same code path).

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -702,11 +702,13 @@ routing:
             capture_request_body: true
             capture_response_body: true
             max_body_bytes: 4096
-            # 0 = no truncation (the documented default). Set to a positive
-            # byte cap to independently limit structured tool-trace fields
-            # (prompt, tool_definitions, and each tool-call's arguments/output)
-            # without having to shrink max_body_bytes.
-            max_tool_trace_bytes: 0
+            # Caps structured tool-trace fields (prompt, tool_definitions, and
+            # each tool-call's arguments/output) independently of max_body_bytes.
+            # Defaults to 4096; a large request would otherwise produce a
+            # multi-MB prompt that breaks the /v1/router_replay history list with
+            # a 413. Set to 0 for no truncation (full structured fields, at the
+            # risk of oversized history responses).
+            max_tool_trace_bytes: 4096
             # Cap the number of tool-trace steps retained per record. Long
             # agent sessions (many sequential tool calls) can otherwise grow
             # the trace unbounded and OOM the router (#1835). 0 = no cap.
--- a/e2e/profiles/router-replay/profile.go
+++ b/e2e/profiles/router-replay/profile.go
@@ -62,6 +62,7 @@ func (p *Profile) GetTestCases() []string {
 		"router-replay-restart-recovery",
 		"router-replay-session-list-filter",
 		"router-replay-session-turn-progression",
+		"router-replay-large-request-list-ok",
 	}
 }
 
--- a/e2e/testcases/router_replay_large_request.go
+++ b/e2e/testcases/router_replay_large_request.go
@@ -1,0 +1,91 @@
+package testcases
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/vllm-project/semantic-router/e2e/pkg/fixtures"
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+	"k8s.io/client-go/kubernetes"
+)
+
+func init() {
+	pkgtestcases.Register("router-replay-large-request-list-ok", pkgtestcases.TestCase{
+		Description: "GET /v1/router_replay returns 200 (not 413) after a large request is recorded",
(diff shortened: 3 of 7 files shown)


siloteemu:
Added an end-to-end regression test (commit b2faf2fd): `router-replay-large-request-list-ok` records a ~5 MB chat request and asserts `GET /v1/router_replay` returns 200 (not 413). It's wired into the router-replay e2e profile. Note: that profile is manual-only, so it runs when the profile is invoked rather than on every PR's default CI matrix — the pkg/extproc unit tests (which drive the real handler + recorder with a 5 MB body across all three 413 cases) provide the in-CI coverage.

theohsiung:
This also resolves #2189 — same `413` on the replay history list from a large recorded request, traced to this exact path. Linking so it closes when this merges. Nice fix 👍

akshayv:
@Xunzhuo  @rootfs  can you have a look at this one please, we're running into 419 issues with large requests

akshayv:
The other option is to change the insights page so that the list API call only retrieves the summary data for each row and the details view for the insights gets the full data for that request/response

siloteemu:
That summaries-only-list + full-data-on-detail split is already tracked in #2157. Good news for that direction: the detail view already goes through the detail endpoint — clicking a row opens the standalone record page (`InsightsRecordPage`), which fetches `GET /v1/router_replay/{id}` itself; the list table renders only summary columns and client-side search is on `request_id`. So making the list summaries-only looks largely non-breaking for the current dashboard runtime (a couple of e2e mock fixtures embed bodies and would just need aligning).

#2155 is the immediate non-breaking stop-gap (bound the list payload so it stops 413ing); #2157 is the right home for the redesign — happy to own it if maintainers want.

siloteemu:
After discussing with @akshayv, we agreed the cleaner fix is the summaries-only list redesign tracked in #2157 — the list returns lightweight rows and full data stays on the per-record detail endpoint — rather than bounding the list payload as this PR does. @akshayv is opening a PR for that approach, so closing this in favor of it. That redesign also resolves #2189. Thanks!

PR #2159, opened 2026-06-11
[Bindings] Implement causal masking in the chunked-SDPA kernel

Peterren:
## What

Implements **causal masking** in the shared `chunked_sdpa` kernel, which was previously reserved and hard-failing (`cfg.causal == true` returned an error). This is step 0 of the remaining #2014 rollout — a pure kernel capability that unblocks the Gemma3/Gemma embedding migration (#4) and the generative Qwen3-LoRA migration (#5).

**Scope: prefill / same-length** (`q_len == k_len`, no KV-cache offset). A query at absolute index `i` attends only to keys `j <= i`. The KV-cache/decode-offset path (`k_len > q_len`) is deferred to the generative migration; `build_causal_mask` already takes absolute q/k indices so that extension is small.

No model is migrated here, and the Qwen3 F64→F32 question is untouched.

## Design

- **`build_causal_mask(q_start, q_len, k_start, k_len, device)`** — mirrors the existing `build_local_band_mask`: an additive `(q_len, k_len)` triangle (`0` where `j <= i`, `-inf` otherwise). Absolute indices keep it correct under a future decode offset.
- **Block loop** applies the causal mask after the padding and window masks. Masks compose by addition (`-inf + anything = -inf`), so `causal + window` is exactly the intersection of the causal triangle and the sliding-window band.
- **Key-range optimization**: causal blocks cap their upper key bound at the block end `qe` instead of `seq_len` — keys beyond `qe` are always masked for every query in the block, so excluding them is correctness-preserving *and* bounds memory further.
- **dtype-agnostic**: the mask is converted to the score dtype before adding, so F64 works unchanged.
- **Behavior-preserving**: the only existing caller (mmBERT) passes `causal: false`; its output is bit-identical to before.

## Test Plan

- `cd candle-binding && cargo test --no-default-features --lib model_architectures::attention` → **7/7 pass**, including:
  - `test_chunked_sdpa_matches_dense_causal` — global+causal and sliding-window+causal across `seq ∈ {1,5,16,40}`, `block ∈ {0,1,3,8,16,512}`, `max|Δ| < 1e-4` vs a dense causal reference.
  - `test_chunked_sdpa_matches_dense_causal_with_padding` — causal + trailing padding.
  - `test_causal_mask_semantics` — unit-tests `build_causal_mask` on an offset block.
  - Existing global/local/F64/band tests still pass (dense reference extended with a `causal` param; obsolete `rejects_causal` test removed).
- `cargo fmt --check` clean; `cargo clippy --no-default-features` — **0 findings in the changed files** (changed-file-filtered Rust lint exits 0).
- `structure_check.py` on the changed files — passed (file 200 lines, `chunked_sdpa` 78 lines, both within limits).
- Pure kernel capability — no routing/startup/config/API change → no E2E updates.

Refs #2014


Peterren:
@Xunzhuo need a thumb up

Xunzhuo (approved):

PR #2162, opened 2026-06-12
[Router] Dedup query-embedding inference on the cache-miss path (in-memory backend)

### LINKED ISSUE #2290: perf: [Router] In-memory semantic cache embeds the query twice on every cache miss
### Describe the bug

On every **cache-miss** request, the in-memory semantic cache embeds the same
query **twice**:

1. `FindSimilarWithThreshold` (lookup) → `generateEmbedding(query)`
2. `AddPendingRequest` (pending-entry write) → `generateEmbedding(query)` — same
   query, microseconds later

`generateEmbedding` calls the candle binding (BERT/mmBERT/Qwen3/…) directly with
no memoization, and embedding inference dominates the cache-path latency, so the
miss path pays roughly **2× the embedding cost** for no benefit.

### Expected behavior

A cache miss should embed the query at most once; the lookup and the pending
write a moment later should reuse the same vector.

### Affected layer

Router — semantic cache (in-memory backend).

### Additional context

An embedding is a deterministic pure function of (query, configured model), so a
small bounded memo (query text → embedding) on the in-memory backend is safe and
removes the redundant inference. The `CacheBackend` interface splits the miss
path into two calls, so the other backends share the pattern and can adopt the
same memo independently.

theohsiung:
Closes #2290

## What

On every **cache-miss** request the in-memory semantic cache embeds the same query **twice**:
1. `FindSimilarWithThreshold` (lookup) → `generateEmbedding(query)`
2. `AddPendingRequest` (pending-entry write) → `generateEmbedding(query)`  *(same query, microseconds later)*

`generateEmbedding` had no memoization — it calls the candle binding (BERT/mmBERT/Qwen3/…) directly each time. Embedding inference dominates the cache-path latency, so the miss path paid ~2× the embedding cost for no benefit.

## Change

Route `generateEmbedding` through a small **bounded, concurrency-safe memo** (`query text → embedding`); the candle-calling body is unchanged, just renamed to `computeEmbedding`. The lookup populates the memo, and the pending-write a moment later hits it.

- **Correct under concurrency:** an embedding is a deterministic pure function of (query, configured model), so memoizing is always safe — the worst case under a race is a redundant recompute, never a wrong vector. `compute` runs **outside** the memo lock, so concurrent embeddings are never serialized behind one another.
- **Bounded:** default 512 entries (FIFO), a few MB at most even for 1024-dim models.
- **Bonus:** also serves repeated identical queries across requests.

## Scope / limitations

In-memory backend only. The `CacheBackend` interface separates the lookup and pending-write into two calls, so threading the precomputed embedding through explicitly would change the signature for all six backends; a backend-local memo is the contained fix. Redis/Milvus/Qdrant/Valkey/Hybrid can adopt the same approach independently.

## Test plan

- [x] `inmemory_embedding_memo_test.go`: computes once per key; distinct keys compute independently; evicts oldest past `maxSize`; propagates and does **not** memoize compute errors (retries); race-clean under concurrent access (`-race`).
- [x] `go test ./pkg/cache/` in-memory suite green (no regression).

## Notes
- DCO signed-off. Pure performance change — no behavior/contract change (cache hit/miss decisions are unaffected).


Xunzhuo (approved):
Nice catch.

1. Does the issue appear in other cache backend?
2. Can you help check the lifecycle of cache in each semantic cache backend? I mean lets make sure there is no memory leak

theohsiung:
Thanks for the review @Xunzhuo. I audited the lifecycle of all six cache backends — summary below.

**1. Does the duplicate-embedding-on-miss appear in other backends?**

Structurally yes — it's interface-level. `CacheBackend` splits the miss path into two calls (lookup `FindSimilarWithThreshold` + `AddPendingRequest`) and each generates the embedding, so every backend pays it. This PR scopes the memo fix to the in-memory backend (as noted in the PR's Scope section); Redis/Milvus/Qdrant/Valkey/Hybrid can adopt the same backend-local memo independently. For the external backends the redundant compute is dwarfed by the network round-trip, so in-memory is where it actually hurts.

**2. Lifecycle / memory-leak check across backends — no leak, with one finding:**

| Backend | Verdict |
|---|---|
| in-memory | **Bounded.** All structures capped by `maxEntries`; HNSW fully rebuilt (`markStale`) on every eviction/TTL cleanup so no orphan nodes; embedding memo fixed 512 FIFO; background cleanup goroutine closed in `Close()`. |
| Redis / Milvus / Qdrant / Valkey | **Clean.** Single client created once and reused, closed in `Close()`; no per-request goroutines/tickers; embeddings & buffers are per-call locals. |
| hybrid | **Bounded, but retains TTL-expired entries** in the in-memory HNSW until the `maxMemoryEntries` cap (no TTL-aware reclamation; `RebuildFromMilvus` only runs at startup). Not unbounded and not a correctness bug (a stale node resolves to a Milvus `GetByID` miss). |

I filed the hybrid finding as #2288 and the fix as #2289 (a `Close()`-managed periodic re-sync) to keep this PR tightly scoped to the in-memory dedup. This PR itself is unchanged.

(For context: this is unrelated to the memory growth in #2222 — that report has `semantic_cache` disabled, and all backends are bounded regardless.)

theohsiung:
@WUKUNTAI-0211 can u help merge this one so I can move on to #2289

PR #2175, opened 2026-06-13
[Router] fix deepseek official thinking mode

### LINKED ISSUE #2147: feature: deepseek 4 family thinking support
### Describe the feature

Currently, if i understood correctly, there is no way to properly set reasoning effort and thinking mode for deepseek 4 series, because it have [different format](https://api-docs.deepseek.com/guides/thinking_mode):

```
{"thinking": {"type": "enabled/disabled"}}
```

compared to much more common

```
{"thinking": false}
```
(which was also present in previous deepseeks)

It also supports thinking_effort, but there is only 2 toggles - high and max, so "none" is not possible as well

### Primary layer

global level

### Why this layer?

part of model settings config / ui

to support deepseek official endpoint properly

### Additional context

Maybe instead of fixing one model (if it's possible) reasoning configuration needs to be more flexible with some kind of values mapping, because reasoning_effort might have different values per model too.

It's also impossible to use both reasoning toggle and reasoning effort combined.

brelance:
Closes #2147

## Purpose

DeepSeek's official OpenAI-compatible API uses a different thinking-mode
request shape for the DeepSeek v4 family than the local vLLM chat-template
path. Before this change, Semantic Router could express DeepSeek-style
reasoning as `chat_template_kwargs.thinking`, but the official DeepSeek
endpoint expects a top-level `thinking` object and top-level
`reasoning_effort`.

This PR fixes that provider-specific behavior and factors the provider
classification into a shared OpenAI backend dialect helper. The dialect helper
keeps official OpenAI, official DeepSeek, local vLLM, and generic
OpenAI-compatible backends explicit so request mutation and unknown-field
stripping use the same provider capability model.

This PR updates the request handling so:

- Official DeepSeek provider profiles receive top-level
  `thinking: {"type": "enabled"}` when reasoning is enabled.
- Official DeepSeek provider profiles receive top-level
  `thinking: {"type": "disabled"}` when reasoning is disabled, with
  `reasoning_effort` removed for that disabled path.
- Configured per-decision `reasoning_effort` is still applied for enabled
  official DeepSeek requests, including aliased model references.
- Official OpenAI provider profiles keep top-level `reasoning_effort`.
- Local vLLM and generic OpenAI-compatible providers keep the existing
  `chat_template_kwargs` reasoning syntax.
- `request_params.strip_unknown` now preserves top-level `thinking` only for
  official DeepSeek API profiles and still strips it for other providers.
- Provider capability detection is covered by focused dialect classification
  tests.
- Affected module: `Router`.

For example, an enabled official DeepSeek request is now mutated toward:

```json
{
  "model": "deepseek-v4-pro",
  "messages": [],
  "thinking": {"type": "enabled"},
  "reasoning_effort": "max"
}
```

The equivalent local vLLM-compatible path still uses the configured chat
template field instead:

```json
{
  "model": "deepseek-v4-pro",
  "messages": [],
  "chat_template_kwargs": {"thinking": true}
}
```

## Test Plan

- Reviewers can validate the changed-file scope with:

  ```bash
  GO_FILES="src/semantic-router/pkg/extproc/req_filter_reason.go,src/semantic-router/pkg/extproc/req_filter_request_params.go,src/semantic-router/pkg/extproc/req_filter_provider_dialect.go,src/semantic-router/pkg/extproc/req_filter_provider_dialect_test.go"

  make agent-report ENV=cpu CHANGED_FILES="$GO_FILES"
  make agent-lint CHANGED_FILES="$GO_FILES"
  ```

- Run Go-specific checks for the affected router/extproc code:

  ```bash
  make check-go-mod-tidy
  make go-lint
  make test-semantic-router
  ```

- For behavior validation, exercise an auto-routed request that selects an
  OpenAI-compatible provider profile with `base_url: https://api.deepseek.com`
  and a DeepSeek reasoning-capable model. Confirm the upstream JSON body uses
  top-level `thinking` plus top-level `reasoning_effort`.
- Confirm a disabled reasoning path for the official DeepSeek profile sends
  `thinking: {"type": "disabled"}` and does not keep `reasoning_effort`.
- Confirm the same DeepSeek reasoning family on a local vLLM-compatible
  provider still sends reasoning through `chat_template_kwargs.thinking`.
- Confirm the official OpenAI profile keeps `reasoning_effort` as a top-level
  field.
- Confirm `request_params.strip_unknown` preserves `thinking` for the official
  DeepSeek provider profile and strips it for non-DeepSeek providers.

## Test Result

- `make agent-report ENV=cpu CHANGED_FILES="$GO_FILES"` passed and resolved
  this as a Router/plugin runtime change.
- `make check-go-mod-tidy` passed.
- `make go-lint` passed.
- `make agent-lint CHANGED_FILES="$GO_FILES"` passed.


rlex:
BTW, it's also affecting providers like opencode zen and probably tons of others which pipe requests to official endpoints as a proxy - so it will only fix official api endpoint. Can it be configurable?

brelance:
> BTW, it's also affecting providers like opencode zen and probably tons of others which pipe requests to official endpoints as a proxy - so it will only fix official api endpoint. Can it be configurable?

You’re right that proxy providers can expose different OpenAI-compatible reasoning semantics, and the current host-based detection is not sufficient for all cases.

However, I don’t think we should expose low-level provider-specific fields directly in config. The router should own known provider/model request formats in code, because users should not need to know whether a backend expects top-level `thinking`, `enable_thinking` inside `chat_template_kwargs`, top-level `reasoning_effort`, or another provider-specific reasoning shape.

For this PR, I’d prefer to keep the DeepSeek official fix scoped to the official DeepSeek endpoint. For proxy providers like OpenCode Zen, we should first verify which reasoning request format they actually support. If needed, we can add a higher-level named override later, such as `reasoning_format`, as an escape hatch for proxy/gateway endpoints, while avoiding raw field-level controls.

FAUST-BENCHOU:
pls resolve the conflict

drivebyer (approved):
/lgtm

PR #2186, opened 2026-06-14
[CI/Build] Run CI-safe Rust lib tests

### LINKED ISSUE #2182: [CI/Build] candle-binding Rust unit tests (cargo test --lib) are not run in PR CI
## What

The Rust unit tests in `candle-binding` (`cargo test --lib`, via the `test-rust` make target) are not executed by PR CI.

Under `CI=true`, `ifeq ($(CI),true)` in `tools/make/build-run-test.mk` selects a `test:` recipe (line 97) that runs the Go binding tests (`test-binding-minimal` / `test-binding-lora`) and never invokes `test-rust`. `test-rust` (the `cargo test --lib` target, `tools/make/rust.mk:14`, and the only `--lib` run in the `make test` chain) appears only in the non-CI `test:` recipe (line 99). Separately, `publish-crate.yml` is tags-only and runs selected integration tests (`cargo test --test ...`), not `--lib`. Net effect: the `candle-binding` `--lib` unit tests run only via `make test-rust` locally - they are invisible to PR CI, and the release workflow does not cover them either.

Found while reviewing #2145 (FFI embedding re-normalization), which adds a good red/green unit test (`test_truncate_embedding_renormalizes_prefix`). It passes locally via `make test-rust`, but does not run in PR CI, so it cannot guard against a future regression at PR time.

## Why it matters

Correctness-critical pure-Rust logic (FFI helpers, pooling, similarity, Matryoshka truncation) is unit-tested, but those tests do not gate PRs. A regression in pure-Rust logic can merge green.

## Complication (why this needs a design decision)

A blanket `cargo test --lib` cannot simply be switched on: model-dependent unit tests gate inconsistently. Some `panic!` when models are absent (e.g. `core/similarity_test.rs:23`), some `return`/skip (e.g. `core/tokenization_test.rs:291`), some fall back to mock paths, and some are `#[ignore]`d. Enabling the full suite cleanly would mean standardizing that behavior first.

## Options

- **(a)** Run a curated no-model subset of pure-logic unit tests in PR CI (smallest, safest, fastest).
- **(b)** Standardize model-dependent tests to skip-not-panic when models are absent, then run `cargo test --lib` (cleaner long-term, larger change).
- **(c)** Run `cargo test --lib` in the existing model-download phase, reusing the models CI already pulls for the Go tests (no extra download, but slower and dependent on which models are present per phase).

Happy to prototype (a) as a starting point.

brelance:
Closes #2182

## Purpose

- Add a CI-safe Rust `--lib` unit test target for `candle-binding` that runs without downloaded model assets or GPU/CUDA requirements.
- Wire that target into the `CI=true` `make test` path so PR CI covers the curated Rust unit-test subset before model-dependent binding tests.
- Run the same CI-safe Rust unit tests in `publish-crate.yml` before the crate release workflow moves on to API smoke tests.
- Affected module(s): `Bindings`, `CI/Build`

### DIFF
--- a/tools/make/rust.mk
+++ b/tools/make/rust.mk
@@ -5,6 +5,31 @@
 # Default GPU device for testing (can be overridden: TEST_GPU_DEVICE=3 make test-rust)
 TEST_GPU_DEVICE ?= 2
 
+# Rust lib unit tests that are safe for PR CI:
+# - no downloaded model assets
+# - no GPU/CUDA requirement
+# - deterministic pure Rust/FFI helper logic
+#
+# Keep this list explicit. Do not include tests whose fixtures initialize
+# models from ../models unless those tests have been converted to skip cleanly.
+RUST_CI_LIB_TESTS ?= \
+	core::tokenization_test::test_tokenization_config_default \
+	core::tokenization_test::test_tokenization_config_custom
+
+test-rust-ci:
+	@$(LOG_TARGET)
+	@echo "Running CI-safe Rust lib unit tests (CPU-only, no model assets)"
+	@cd candle-binding && \
+	test_list="$$(cargo test --release --no-default-features --lib -- --list)" && \
+	for test_filter in $(RUST_CI_LIB_TESTS); do \
+		echo "$$test_list" | grep -F "$${test_filter}:" >/dev/null || { \
+			echo "Configured Rust CI test not found: $$test_filter"; \
+			exit 1; \
+		}; \
+		echo "Running $$test_filter"; \
+		cargo test --release --no-default-features --lib "$$test_filter" -- --exact --test-threads=1 --nocapture; \
+	done
+
 # Test Rust unit tests (with release optimization for performance)
 # Note: Uses TEST_GPU_DEVICE env var (default: 2) to avoid GPU 0/1 which may be busy
 # Override with: TEST_GPU_DEVICE=3 make test-rust
--- a/tools/make/rust.mk
+++ b/tools/make/rust.mk
@@ -27,7 +27,7 @@ test-rust-ci:
 			exit 1; \
 		}; \
 		echo "Running $$test_filter"; \
-		cargo test --release --no-default-features --lib "$$test_filter" -- --exact --test-threads=1 --nocapture; \
+		cargo test --release --no-default-features --lib "$$test_filter" -- --exact --test-threads=1 --nocapture || exit 1; \
 	done
 
 # Test Rust unit tests (with release optimization for performance)
--- a/tools/make/rust.mk
+++ b/tools/make/rust.mk
@@ -14,7 +14,8 @@ TEST_GPU_DEVICE ?= 2
 # models from ../models unless those tests have been converted to skip cleanly.
 RUST_CI_LIB_TESTS ?= \
 	core::tokenization_test::test_tokenization_config_default \
-	core::tokenization_test::test_tokenization_config_custom
+	core::tokenization_test::test_tokenization_config_custom \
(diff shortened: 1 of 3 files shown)


tools/make/rust.mk
```diff
@@ -5,6 +5,31 @@
...
+# Rust lib unit tests that are safe for PR CI:
+# - no downloaded model assets
+# - no GPU/CUDA requirement
+# - deterministic pure Rust/FFI helper logic
+#
+# Keep this list explicit. Do not include tests whose fixtures initialize
+# models from ../models unless those tests have been converted to skip cleanly.
+RUST_CI_LIB_TESTS ?= \
+	core::tokenization_test::test_tokenization_config_default \
+	core::tokenization_test::test_tokenization_config_custom
```
drivebyer: Consider add `test_truncate_embedding_renormalizes_prefix ` as it metioned in #2182

drivebyer (approved):

shraderdm (commented):
Thanks for picking this one up, @brelance, and for turning #2182 (the candle `--lib` CI-coverage gap) into such a tidy, surgical change. I reviewed the wiring end to end, and pulled it down and ran it locally; it does exactly what it says.

### What's good
- The PR-gating path is real, not just plausible. `test-rust-ci` lands in the `CI=true` branch of `test:` (`tools/make/build-run-test.mk:97`), and that recipe is what `test-and-build.yml` runs on pull requests (`make test` with `CI: true`). That is the gap the issue described, actually closed.
- `--exact` is load-bearing and the right call. `candle-binding/src/ffi/embedding_test.rs` mixes the model-free `test_truncate_embedding_renormalizes_prefix` with model-dependent siblings whose fixture `panic!`s when assets are absent. Pinning to exact names keeps the CI subset from ever pulling those in.
- The `grep -F "${test_filter}:"` existence check fails loud (`exit 1`) if a configured test is renamed or removed, so the curated list cannot silently rot.
- `--no-default-features` matches the repo's CPU-only CI convention (`default = ["cuda"]`), and running before `download-models` means the subset fails fast without paying the model-download cost.
- I ran `make test-rust-ci` locally to confirm the behavior: all three tests execute (`1 passed` each, 617 filtered out), a bogus entry in the list fails the target loudly (`Configured Rust CI test not found`, exit 2), and breaking the renorm assertion turns the subset red. The lane works end to end.

### Minor / nits (optional, non-blocking)
- nit: the PR body has `## Purpose` but not the template's `## Test Plan` / `## Test Result`. For a test-infra change it would be nice to capture how it was checked (for example, "ran `make test-rust-ci`; confirmed the guard fails on a bad name"). Purely for the record.
- nit: `RUST_CI_LIB_TESTS` is a hand-maintained allowlist, so a newly added pure-Rust test will not gate PRs until someone appends it here. That is the accepted tradeoff of the curated-subset approach, and the issue stays open to track the longer-term skip-not-panic path that would let the full `--lib` suite run. A one-line breadcrumb in the comment would help the next reader, something like:

  ```makefile
  # Keep this list explicit. Do not include tests whose fixtures initialize
  # models from ../models unless those tests have been converted to skip cleanly.
  # Interim curated subset; the durable follow-up is to standardize
  # model-dependent tests to skip-not-panic, then run the full --lib suite.
  ```

### Verdict
Looks correct, verified locally (red/green plus the guard) and against the base. Both nits are optional. drivebyer has already approved, so I am leaving this as a comment rather than a second formal sign-off, but from my read it is ready to merge.

shraderdm (approved):
Approving - verified the curated subset locally (all three run, guard fails loud, renorm goes red when broken) and the CI wiring against the base. Thanks @brelance!

PR #2188, opened 2026-06-14
[Bindings] Re-normalize onnx multimodal embeddings after MRL truncation

### LINKED ISSUE #2183: [Bindings] onnx-binding multimodal maybe_truncate does not re-normalize after MRL dimension truncation
## What

`onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs` `maybe_truncate` (around L413) L2-normalizes the full embedding, then slices the prefix to `target_dim` (L424) without re-normalizing the truncated vector. Downstream similarity scoring treats embeddings as unit-norm (e.g. `cosineSimilarity` in `src/semantic-router/pkg/classification/embedding_classifier_scoring.go` returns a raw dot product, "Assumes vectors are normalized"), so a truncated vector with `||v|| < 1` shrinks every similarity score.

It is a live path: `maybe_truncate` is called from the text / image / audio encode paths (L289 / L322 / L371).

This is the same bug class as #2145 (FFI embedding re-normalization), which fixes it on the candle-binding side. Found while reviewing that PR.

## Fix

Re-normalize the prefix after slicing, mirroring the candle-binding fix and the existing model-level `l2_normalize`-after-`narrow` convention. Reusing the function's existing `1e-12` guard idiom:

```rust
if let Some(dim) = target_dim {
    if dim > 0 && dim < normalized.len() {
        let sliced = normalized.slice(ndarray::s![..dim]).to_owned();
        let n = sliced.iter().map(|x| x * x).sum::<f32>().sqrt();
        let n_safe = if n > 1e-12 { n } else { 1e-12 };
        return Ok(sliced.mapv(|x| x / n_safe));
    }
}
```

## Related (latent, candle-binding)

The candle-binding `extract_embeddings` trait impls (`qwen3_embedding.rs`, `mmbert_embedding.rs`, `multimodal_embedding.rs`) also `narrow` without re-normalizing, unlike their `encode_*` siblings. They appear dormant today (the only caller is a unit test, and it passes `None` for `target_dim`, so the truncation branch is never exercised), so this is latent parity debt rather than a live bug - but a comment or the same renorm would prevent a future caller from reintroducing the issue.

WUKUNTAI-0211:
## What

`maybe_truncate` in `onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs` L2-normalized the full embedding, then sliced the prefix to `target_dim` **without re-normalizing** the truncated vector.

## Why

Downstream similarity scoring treats embeddings as already unit-normalized and computes cosine similarity as a plain dot product (`cosineSimilarity` in `src/semantic-router/pkg/classification/embedding_classifier_scoring.go`, "Assumes vectors are normalized"). A truncated vector with `||v|| < 1` therefore shrinks every similarity score.

It is a live path: `maybe_truncate` is called from the text / image / audio encode paths.

Same bug class as #2145 (FFI embedding re-normalization on the candle-binding side); found while reviewing that PR.

## Fix

- Extract a shared `truncate_embedding_to_dimension` helper that re-normalizes the prefix after slicing, plus an `l2_normalize` helper.
- `maybe_truncate` now delegates to it.
- Mirrors the candle-binding fix and the existing model-level `l2_normalize`-after-`narrow` convention.

## Test plan

- Add model-free unit tests in `multimodal_embedding.rs`:
  - `test_truncate_embedding_renormalizes_prefix` — truncated unit-norm embedding stays unit-normalized
  - `test_truncate_general_prefix_is_unit_norm`
  - `test_no_truncation_still_normalizes`
  - `test_out_of_range_dim_is_ignored`
- `rustfmt --check` clean on the changed file.

Note: per #2182, candle-binding/onnx `cargo test --lib` does not run in PR CI today; the renormalization logic was additionally verified against reference arithmetic.

Fixes #2183

### DIFF
--- a/onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs
+++ b/onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs
@@ -383,28 +383,33 @@ impl MultiModalEmbeddingModel {
             "last_hidden_state",
         ];
         for name in &names {
-            if let Some(output_value) = outputs.get(*name) {
-                if let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() {
-                    let dims: Vec<usize> = shape.iter().map(|&d| d as usize).collect();
-                    if dims.len() == 2 && dims[0] >= 1 {
-                        let dim = dims[1];
-                        let vec: Vec<f32> = data.iter().take(dim).copied().collect();
-                        return Ok(Array1::from_vec(vec));
-                    } else if dims.len() == 1 {
-                        return Ok(Array1::from_vec(data.to_vec()));
-                    }
-                }
+            let Some(output_value) = outputs.get(*name) else {
+                continue;
+            };
+            let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() else {
+                continue;
+            };
+
+            let dims: Vec<usize> = shape.iter().map(|&d| d as usize).collect();
+            if dims.len() == 2 && dims[0] >= 1 {
+                let dim = dims[1];
+                let vec: Vec<f32> = data.iter().take(dim).copied().collect();
+                return Ok(Array1::from_vec(vec));
+            } else if dims.len() == 1 {
+                return Ok(Array1::from_vec(data.to_vec()));
             }
         }
         // Fallback: first output
         if let Some((_, output_value)) = outputs.iter().next() {
-            if let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() {
-                let dims: Vec<usize> = shape.iter().map(|&d| d as usize).collect();
-                let dim = *dims.last().unwrap_or(&0);
-                if dim > 0 {
-                    let vec: Vec<f32> = data.iter().take(dim).copied().collect();
-                    return Ok(Array1::from_vec(vec));
-                }
+            let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() else {
+                return Err(errors::ort_error("no valid embedding output found"));
+            };
+
+            let dims: Vec<usize> = shape.iter().map(|&d| d as usize).collect();
+            let dim = *dims.last().unwrap_or(&0);
+            if dim > 0 {
+                let vec: Vec<f32> = data.iter().take(dim).copied().collect();
+                return Ok(Array1::from_vec(vec));
             }
         }
         Err(errors::ort_error("no valid embedding output found"))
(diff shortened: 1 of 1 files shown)


shraderdm (commented):
Thanks for picking this up so fast, @WUKUNTAI-0211, and for carrying the fix across to the onnx side. I filed #2183 (onnx renorm) and reviewed the candle-binding sister fix in #2145 (FFI renorm), so this was a quick one to pick up. I pulled the branch down and ran it locally.

### What's good

- **The re-normalization is correct, and I verified it red/green.** `truncate_embedding_to_dimension` re-normalizes the prefix after the Matryoshka slice, which is exactly what #2183 (onnx renorm) was about: downstream cosine similarity is a plain dot product that assumes unit-norm vectors (`cosineSimilarity` in `embedding_classifier_scoring.go`), so a truncated vector with `||v|| < 1` was shrinking every score. With the fix all four unit tests pass; when I temporarily drop the prefix re-normalization, `test_truncate_embedding_renormalizes_prefix` fails with `left: 0.6, right: 1.0`, the right reason. A genuinely falsifying regression test is the best part of this.
- **Nice consolidation.** The shared `l2_normalize` + `truncate_embedding_to_dimension` helpers mirror the candle-binding fix, and re-normalizing after truncation matches the existing pattern in the mmbert path (`mmbert_embedding.rs` truncates, then `l2_normalize`s), so the bindings now read consistently.

### Findings

#### Important (currently blocking CI)

- **`data.as_ref()` trips `clippy::useless_asref`, which fails the pre-commit lint under `-D warnings`** (`onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs:417`). `try_extract_tensor::<f32>()` already hands back a `&[f32]`, so the `.as_ref()` is a no-op and clippy rejects it. I reproduced it locally, and it is the current red check on the PR. One-character fix, dropping `.as_ref()`:

  ```rust
      embedding_from_tensor_data(&dims, data)
  ```

#### Scope (a code-owner call, but I have a lean)

- **This PR bundles two independent changes, and only one is described.** The title and body cover the #2183 (onnx renorm) fix (commit 1), but commit 2 ("fix onnx multimodal output extraction") is a separate, behavior-preserving refactor of the output-extraction path - it hoists the inline parsing into `extract_embedding_from_value` / `extract_embedding_from_fallback_value` / `embedding_from_tensor_data` - that the description doesn't mention. I'd lean toward splitting it into its own PR, for two concrete reasons: the project asks contributors to keep unrelated refactors out of a feature PR, and that refactor is also what's failing CI here (the `as_ref` above lives entirely in commit 2; the renorm fix in commit 1 doesn't touch it). Split, and the #2183 (onnx renorm) fix can be reviewed and land on its own merits while the refactor gets its own review. If you'd prefer to keep them together, a line in the body describing the refactor and retitling that commit to `refactor ...` would at least close the disclosure gap. The final scoping call is the code-owners' (rootfs / Xunzhuo).

#### Minor / nit (optional)

- **The extracted `embedding_from_tensor_data` helper has no unit coverage.** It is the dispatch of the output-extraction refactor (2D first-row take / 1D passthrough / unsupported-rank `None`) and is pure and fixture-free testable (`(&[usize], &[f32]) -> Option<Array1<f32>>`, no ONNX session needed), but the four new tests only exercise `truncate_embedding_to_dimension`. A couple of direct cases (2D, 1D, unsupported rank) would guard that dispatch cheaply. Since this is the same code the scope note is about, the test should travel with the helper: add it here if the refactor stays, or in the split-out PR if it moves.
- nit: **On the truncation path, `truncate_embedding_to_dimension` normalizes the full vector, slices the prefix, then re-normalizes the prefix** (`multimodal_embedding.rs:461`). The trailing re-normalization makes the prefix independent of the first pass, so that full-vector normalization is redundant on this branch (output is identical to sub-ULP). Not a correctness issue and not a hot path. If you want to trim it, normalize the prefix inside the `Some(dim)` branch and keep the existing `l2_normalize` for the no-truncation and out-of-range-`dim` returns, which still need it. Entirely optional.

### Verdict

The renorm fix itself is correct, well-tested, and closes #2183 (onnx renorm). Two things before it is ready: the one-line `as_ref` fix to get CI green, and the scoping question above (my lean is to split the refactor out). I will defer the formal approval to the code-owners (rootfs / Xunzhuo).

WUKUNTAI-0211:
Split the output-extraction refactor out to #2190. This PR is now rebased on latest `main` and contains only the #2183 ONNX MRL truncation re-normalization fix plus its unit tests.

shraderdm (approved):
Thanks for splitting this out so quickly, @WUKUNTAI-0211. The rebased PR is exactly right: just the #2183 (onnx renorm) fix and its tests, with the output-extraction refactor moved cleanly to #2190 (refactor). I re-checked the rebased diff - the re-normalization logic and the falsifying `test_truncate_embedding_renormalizes_prefix` are unchanged from what I verified red/green earlier, and CI is green. Approving. Nice work turning this around.

shraderdm:
Quick follow-up, and a correction to my own approval above: this PR and #2190 (refactor) both modify `extract_embedding`, from the same point on `main` - here it is converted to `let-else`, and in #2190 the same method is refactored into helpers (`extract_embedding_from_value` and friends). Because both rewrite the same lines from the same base, they will conflict on merge whichever lands second.

#2190's helper-based version of `extract_embedding` is the cleaner of the two, so could you please drop the `let-else` change from this PR and leave `extract_embedding` as it is on `main`? That keeps this one scoped to just the #2183 (onnx renorm) fix, lets the two land in any order, and `extract_embedding` still ends up in the nicer #2190 form once that merges. The re-normalization change itself looks correct - this is only about separating the two cleanly.

FAUST-BENCHOU (approved):
LG

Issue #2173, opened 2026-06-12
bug: outer_dc algorithm fails to initialize with "failed to generate batched embedding (status: -1)" on ARM64 Linux (Apple Silicon/Docker)


TwilightTechie:
When running vllm-sr serve on an Apple Silicon Mac (using Docker/Colima which runs a linux/arm64 VM), the semantic router fails to initialize the router_dc model selection algorithm. It crashes when attempting to generate batched embeddings for the model descriptions during startup.

The status: -1 error originates from the ONNX Runtime Go bindings, suggesting that the onnxruntime-linux-aarch64 C library within the Docker container is failing/crashing when processing batched text inputs with the mmbert-embed-32k-2d-matryoshka model.

### To Reproduce

1. Run on an Apple Silicon Mac using Docker Desktop or Colima.
2. Create a config.yaml that uses the router_dc (or hybrid) algorithm:
```
global:
  model_catalog:
    embeddings:
      semantic:
        embedding_model: mmbert
routing:
  listeners:
    - name: http-8899
      port: 8899
  decisions:
    - name: MoM
      modelRefs:
        - model: qwen-coder
        - model: gpt-4o
      algorithm:
        type: router_dc
```
3. Run the container via the CLI:
`vllm-sr serve --config config.yaml`

4. Observe the router startup logs.
Error Logs
```
{"level":"info","ts":"2026-06-12T20:14:20.192","caller":"logging.go:259","msg":"redis_store_initialized","event":"redis_store_initialized","component":"responsestore"}
{"level":"warn","ts":"2026-06-12T20:14:20.213","caller":"router_dc.go:176","msg":"[RouterDC] Failed to embed model qwen-coder: failed to generate batched embedding (status: -1)"}
{"level":"warn","ts":"2026-06-12T20:14:20.214","caller":"router_dc.go:176","msg":"[RouterDC] Failed to embed model gpt-4o: failed to generate batched embedding (status: -1)"}
{"level":"info","ts":"2026-06-12T20:14:20.216","caller":"logging.go:259","msg":"startup_complete","embedding_ready":false,"model_selection":true,"component":"router"}
```

### Expected behavior

Environment
Host OS: macOS (Apple Silicon / M-Series)
Container Runtime: Docker / Colima (linux/arm64)
Semantic Router Version: ghcr.io/vllm-project/semantic-router/vllm-sr:latest
Expected Behavior
[...]

Issue #2172, opened 2026-06-12
bug: Docker image ships ONNX-only mmBERT model — missing model.safetensors causes embedding_ready=false


TwilightTechie:
### Describe the bug

Description
The vLLM Semantic Router Docker image (v0.3.0) ships the mmbert-embed-32k-2d-matryoshka model with only ONNX files, but the Rust/Candle embedding runtime requires model.safetensors + tokenizer.json to initialize. This causes the embedding runtime to fail silently on every startup.

Error from docker logs:
```
ERROR: Failed to register mmBERT model: Failed to load mmBERT model:
  External { library: "candle-core", operation: "failed to load safetensors from
  models/mmbert-embed-32k-2d-matryoshka/model.safetensors",
  error: "path: \"models/mmbert-embed-32k-2d-matryoshka/model.safetensors\" No such file or directory (os error 2)" }
```
Additionally, the ONNX fallback also fails with:

Error: unsupported model type 'mmbert' for batched embeddings (only 'qwen3' supported)
Files present in container (/app/models/mmbert-embed-32k-2d-matryoshka/):

```
config.json ✅
README.md ✅
onnx/layer-{6,11,16,22}/model.onnx ✅
model.safetensors ❌ MISSING
tokenizer.json ❌ MISSING
special_tokens_map.json ❌ MISSING
```
Impact:

embedding_ready: false in startup log
router_dc algorithm cannot embed model descriptions → falls back to auto_routing (which just picks the default model)
Semantic caching disabled
Tool database loading skipped
RouterDC model card embeddings all fail: models_with_descriptions: 0
Expected behavior: The Docker image should include model.safetensors, tokenizer.json, and special_tokens_map.json from the HuggingFace repo llm-semantic-router/mmbert-embed-32k-2d-matryoshka.

Workaround: Manually download the missing files into the running container:

bash

```
docker exec vllm-sr-router-container /opt/vllm-sr-venv/bin/hf download \
  llm-semantic-router/mmbert-embed-32k-2d-matryoshka \
  model.safetensors tokenizer.json special_tokens_map.json \
  --local-dir /app/models/mmbert-embed-32k-2d-matryoshka
```
Then restart the SR.

Environment:

vllm-sr CLI: v0.3.0
Platform: macOS (Apple Silicon M4 / aarch64)
Docker: ARM64 container
HF_TOKEN: set and valid

### To Reproduce

try it on apple mac mini m4

### Affected layer

None

[...]

PR #2190, opened 2026-06-15
[Bindings] Refactor onnx multimodal output extraction

WUKUNTAI-0211:
## What

Split out the ONNX multimodal output-extraction refactor from #2188 so the #2183 renormalization bugfix can stay narrowly scoped.

This PR keeps the existing extraction behavior but hoists the tensor parsing logic into small helpers:

- `extract_embedding_from_value`
- `extract_embedding_from_fallback_value`
- `embedding_from_tensor_data`

## Why

The extraction cleanup is independent from the MRL truncation renormalization fix in #2188. Keeping it separate makes the bugfix easier to review and gives this helper-level refactor its own tests.

## Test plan

- `rustfmt --check onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs`
- `cargo test --manifest-path onnx-binding/Cargo.toml --lib test_embedding_from_tensor_data -- --nocapture`

### DIFF
--- a/onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs
+++ b/onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs
@@ -382,35 +382,20 @@ impl MultiModalEmbeddingModel {
             "pooler_output",
             "last_hidden_state",
         ];
-        for name in &names {
-            let Some(output_value) = outputs.get(*name) else {
-                continue;
-            };
-            let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() else {
-                continue;
-            };
-
-            let dims: Vec<usize> = shape.iter().map(|&d| d as usize).collect();
-            if dims.len() == 2 && dims[0] >= 1 {
-                let dim = dims[1];
-                let vec: Vec<f32> = data.iter().take(dim).copied().collect();
-                return Ok(Array1::from_vec(vec));
-            } else if dims.len() == 1 {
-                return Ok(Array1::from_vec(data.to_vec()));
-            }
+        if let Some(embedding) = names
+            .iter()
+            .find_map(|name| outputs.get(*name).and_then(extract_embedding_from_value))
+        {
+            return Ok(embedding);
         }
-        // Fallback: first output
-        if let Some((_, output_value)) = outputs.iter().next() {
-            let Ok((shape, data)) = output_value.try_extract_tensor::<f32>() else {
-                return Err(errors::ort_error("no valid embedding output found"));
-            };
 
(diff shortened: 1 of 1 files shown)


theohsiung:
Hi @WUKUNTAI-0211 QQ: Is this the same problem list in #2172 ?

TwilightTechie:
Hi Team, This PR  only fixed the embedding extraction crash in the Rust bindings (onnx-binding/src/model_architectures/embedding/multimodal_embedding.rs).

However, the Semantic Router's core daemon is written in Go, and router_dc.go uses onnxruntime_go (the Go bindings). So we need to port their fix over to the Go side of the codebase, the Go router still crashes with the exact same status: -1 error: "msg":"[RouterDC] Failed to embed model gpt-4o: failed to generate batched embedding (status: -1)"
cc: @WUKUNTAI-0211 @theohsiung @aayushRedHat

WUKUNTAI-0211:
Good question — no, this one is unrelated to #2172. #2190 is a pure refactor of the `onnx-binding` multimodal output-extraction helpers (split out from #2188), no behavior change.

#2172 is the mmbert embedding-readiness bug, which has two real causes being fixed separately: the download-completeness half in #2195, and the selection-dispatch (`status: -1`) half in #2192. This PR doesn't touch either path.

WUKUNTAI-0211:
Thanks @TwilightTechie. One clarification on the codepath: that `status: -1` isn't coming from the ONNX Runtime Go bindings. `router_dc.go` uses an injected embedding func that calls `candle_binding.GetEmbeddingBatched` (CGO → the **candle-binding** crate), and the candle batched FFI returns `-1` for any non-qwen3 `model_type` (the default is `mmbert`). So it's architecture-independent — not an `onnxruntime-linux-aarch64` issue.

This PR (#2190) is an unrelated `onnx-binding` refactor on a different codepath. The crash you filed in #2173 is fixed by #2192 (routes mmbert through the single-text FFI `GetEmbeddingWithModelType`); the missing-model-files side of #2172 is #2195. cc @theohsiung @aayushRedHat

WUKUNTAI-0211:
Closing this. It was split out of #2188 to keep that renormalization bugfix narrowly scoped — but #2188 has since merged (with the extraction logic inline), so the original reason for a separate refactor PR no longer applies.

The remaining diff is a pure no-behavior-change refactor; the only added value was unit tests for the tensor-shape dispatch, which cover the trivial path and not the actual `ort::Value` extraction. Not worth the churn on its own. The real fixes for the mmbert/router_dc issues live in #2192 (#2173) and #2195 (#2172).

PR #1907, opened 2026-05-14
feat(openvino-binding): Add OpenVINO backend for ModernBERT embedding and classifier inference

EmonLu:
Add openvino-binding Go module wrapping OpenVINO C API for tokenizer and inference, enabling the semantic router to run ModernBERT embedding and classifier models via Intel OpenVINO Runtime with zero external Python/ONNX dependencies.

Key changes:
- CGO bindings for OpenVINO C API (tokenizer + inference)
- InferRequest pool with configurable concurrency (OPENVINO_EMBEDDING_INFER_POOL_SIZE)
- ModernBERT embedding and classifier
- Build-tagged integration: openvino_embedding_cgo.go / stub.go
- Auto-detect libopenvino.so version, conditional CUDA, NUMA env vars
- Go runtime tuning (GOMAXPROCS, GOGC) for multi-backend throughput

## Purpose

- **What does this PR change?**
  Adds a new `openvino-binding/` Go module that wraps the OpenVINO C API via CGO, providing native tokenizer and inference capabilities for ModernBERT embedding and classifier models. Includes a connection pool, build-tag-based integration with the router, and automatic runtime environment detection.

- **Why is this change needed?**
  Eliminates external Python and ONNX Runtime dependencies for Intel-platform inference, reducing container image size and startup latency. Enables the semantic router to leverage Intel OpenVINO Runtime directly for ModernBERT models with better throughput on Intel CPU.

- **Which module(s) does this affect?**
  `Bindings` / `Router`

## Test Plan

- Build the C++ library and Go binding:
  ```bash
  cd openvino-binding && mkdir -p build && cd build && cmake .. && cmake --build . -j$(nproc)
  cd openvino-binding && go build -v .
  ```
- Run unit tests for embedding and classifier:
  ```bash
  cd openvino-binding && LD_LIBRARY_PATH=./build go test -v -count=1 -timeout 120s .
  ```
- Validate concurrency under load:
  ```bash
  cd openvino-binding && LD_LIBRARY_PATH=./build go test -v -run TestConcurrency -count=1
  ```
- Run embedding and classifier benchmarks:
  ```bash
  bash openvino-binding/scripts/build_and_run_mmbert_embedding_bench.sh --run-only
  bash openvino-binding/scripts/build_and_run_mmbert_classifier_bench.sh --run-only
  ```
- Run pre-commit on changed files:
  ```bash
  pre-commit run --files $(git diff --name-only upstream/main..HEAD | tr '\n' ' ')
  ```

[...]

PR #2192, opened 2026-06-15
[Router] Fix selection embeddings for non-qwen3 model_type

### LINKED ISSUE #2173: bug: outer_dc algorithm fails to initialize with "failed to generate batched embedding (status: -1)" on ARM64 Linux (Apple Silicon/Docker)

WUKUNTAI-0211:
## Problem

`resolveSelectionEmbeddingFunc` (introduced in #1907) routes every candle-backend
selection embedding through `GetEmbeddingBatched`, which the candle FFI implements
**only for the qwen3 continuous-batching model** — it rejects any other `model_type`
with `unsupported model type ... (only 'qwen3' supported)`.

The default embedding `model_type` is `mmbert` (`pkg/config/canonical_defaults.go`),
so with the default config **every model-description embedding fails**:

```
[RouterDC] Failed to embed model <m>: failed to generate batched embedding (status: -1)
router_dc_config_embeddings_initialized  models_with_descriptions=0
```

RouterDC — and hybrid's `router_dc` component, and `session_aware` when its base
selector uses them — are then left with no model embeddings and silently fall back
to undifferentiated scores.

## Root cause

Two issues in the candle (default) branch:

1. **Wrong FFI entrypoint.** `mmbert` / `gemma` are supported by the single-text FFI
   `GetEmbeddingWithModelType` (used everywhere else in the router), not by the
   qwen3-only `GetEmbeddingBatched`.
2. **Wrong target dimension.** The branch passed `target_dim=1024` (qwen3's native
   dimension). `mmbert` is a 768-dim 2D-Matryoshka model and cannot be truncated
   upward, so 1024 fails the forward pass even on the single-text path.

## Fix

Dispatch by batched-capability: only `qwen3` uses `GetEmbeddingBatched`; every other
`model_type` uses `GetEmbeddingWithModelType` with `target_dim=0` (native dimension,
model-agnostic). Adds a `candleEmbeddingSupportsBatched` helper and a unit test for
the dispatch decision.

## Test Plan

- New unit test `TestCandleEmbeddingSupportsBatched` covers qwen3 (batched) vs
  mmbert/gemma/bert/empty (single-text).
- Built the router image with the fix and ran against a config using the default
  `mmbert` embedding `model_type`:
  - **Before:** `[RouterDC] Failed to embed model ... (status: -1)` for every model,
    `router_dc_config_embeddings_initialized models_with_descriptions=0`.
  - **After:** `models_with_descriptions=4`, no embed errors, and with a `router_dc`
    selector `[RouterDC] Selected model ... (similarity=1.0000, confidence=0.95)` —
    embeddings are computed and consumed.

Fixes #2173
Refs #2172 — this fixes the selection-dispatch half (the `status: -1` on the
qwen3-only batched path). The packaging half (Docker mmbert dir missing
`model.safetensors` / `tokenizer.json`) is #2195.

### DIFF
--- a/src/semantic-router/pkg/extproc/router_selection.go
+++ b/src/semantic-router/pkg/extproc/router_selection.go
@@ -66,22 +66,9 @@ func resolveSelectionEmbeddingFunc(cfg *config.RouterConfig) func(string) ([]flo
 	case "openvino":
 		return openvinoEmbeddingFunc(modelType)
 	default:
-		// The candle batched embedding path (GetEmbeddingBatched) is only
-		// implemented for the qwen3 continuous-batching model; it rejects any
-		// other model_type with "unsupported model type ... (only 'qwen3'
-		// supported)". The default embedding model_type is "mmbert", so
-		// unconditionally using the batched path makes every selection
-		// embedding fail for the default config — RouterDC/hybrid then fall
-		// back to flat scores. Route non-qwen3 model types through the
-		// single-text FFI (GetEmbeddingWithModelType), which supports mmbert,
-		// gemma, etc. and is what the rest of the router already uses.
-		//
-		// target_dim differs by path: the qwen3 batched model emits 1024-dim
-		// vectors, but mmbert is a 2D-Matryoshka model whose native dimension
-		// is 768 and cannot be truncated upward — passing 1024 fails the
-		// forward pass. Use 0 (native dimension) for the single-text path so
-		// each model returns its own dimension; the selector only needs the
-		// per-model embeddings to be mutually consistent, which they are.
+		// Batched FFI only supports qwen3 (1024-dim). Other types (default
+		// mmbert, gemma, ...) use the single-text FFI with target_dim 0 so
+		// each model returns its native dimension.
 		if candleEmbeddingSupportsBatched(modelType) {
 			return func(text string) ([]float32, error) {
 				output, err := candle_binding.GetEmbeddingBatched(text, modelType, 1024)
@@ -101,10 +88,8 @@ func resolveSelectionEmbeddingFunc(cfg *config.RouterConfig) func(string) ([]flo
 	}
 }
 
-// candleEmbeddingSupportsBatched reports whether the candle continuous-batching
-// embedding FFI (GetEmbeddingBatched) supports the given model type. Only qwen3
-// is implemented for batching; all other model types must use the single-text
-// FFI (GetEmbeddingWithModelType).
+// candleEmbeddingSupportsBatched reports whether the batched embedding FFI
+// supports the model type. Only qwen3 does; others use the single-text FFI.
 func candleEmbeddingSupportsBatched(modelType string) bool {
 	return modelType == "qwen3"
 }
--- a/src/semantic-router/pkg/extproc/router_selection.go
+++ b/src/semantic-router/pkg/extproc/router_selection.go
@@ -66,14 +66,47 @@ func resolveSelectionEmbeddingFunc(cfg *config.RouterConfig) func(string) ([]flo
 	case "openvino":
 		return openvinoEmbeddingFunc(modelType)
 	default:
+		// The candle batched embedding path (GetEmbeddingBatched) is only
+		// implemented for the qwen3 continuous-batching model; it rejects any
+		// other model_type with "unsupported model type ... (only 'qwen3'
+		// supported)". The default embedding model_type is "mmbert", so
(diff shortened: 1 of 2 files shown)


WUKUNTAI-0211:
Just a quick note: to address the hardcoded `"qwen3"` check, I've opened a feature request here: #2194 to propose adding a proper capability check in the FFI. 

I'm more than happy to work on that FFI implementation in a follow-up PR if the maintainers agree with the approach. For now, this PR should safely resolve the immediate crash for default configs! 🚀

theohsiung:
@WUKUNTAI-0211 is this the same problem list in #2173

src/semantic-router/pkg/extproc/router_selection.go
```diff
@@ -66,8 +66,33 @@ func resolveSelectionEmbeddingFunc(cfg *config.RouterConfig) func(string) ([]flo
...
+		// back to flat scores. Route non-qwen3 model types through the
+		// single-text FFI (GetEmbeddingWithModelType), which supports mmbert,
+		// gemma, etc. and is what the rest of the router already uses.
+		//
+		// target_dim differs by path: the qwen3 batched model emits 1024-dim
+		// vectors, but mmbert is a 2D-Matryoshka model whose native dimension
+		// is 768 and cannot be truncated upward — passing 1024 fails the
+		// forward pass. Use 0 (native dimension) for the single-text path so
+		// each model returns its own dimension; the selector only needs the
+		// per-model embeddings to be mutually consistent, which they are.
```
AayushSaini101: @WUKUNTAI-0211 can you write the comment in short so that it would be easy to understand :😃 same for all the places in the changes thanks

WUKUNTAI-0211:
> @WUKUNTAI-0211 is this the same problem list in #2173

Yes — this PR is the router-side dispatch half of the #2173 issue you identified. The separate model completeness / missing-files half is in your #2195.

AayushSaini101 (approved):
LGTM thanks : )

PR #2195, opened 2026-06-15
[Router] Require embedding model weights+tokenizer for download completeness

### LINKED ISSUE #2172: bug: Docker image ships ONNX-only mmBERT model — missing model.safetensors causes embedding_ready=false

theohsiung:
## Problem

On the default (candle) image, the `mmbert-embed-32k-2d-matryoshka` embedding model can ship as an **ONNX-only** directory (`config.json` + `onnx/layer-*/model.onnx` + `README.md`, no `model.safetensors` / `tokenizer.json`). The candle embedding runtime loads from `model.safetensors` + `tokenizer.json`, so it fails:

```
ERROR: Failed to register mmBERT model: ... failed to load safetensors from
  models/mmbert-embed-32k-2d-matryoshka/model.safetensors ... No such file or directory
```

…and the router comes up with `embedding_ready=false`, RouterDC model-card embeddings all fail, semantic cache + tool DB are skipped. (#2172)

## Root cause

It is **not** a download-filter bug — `DownloadModelWithProgress` runs `hf download <repo>` with no `--include`, so a real download pulls the full snapshot (safetensors + tokenizer included). The problem is the **completeness check** that decides whether to download at all:

- `BuildModelSpecs` gives the embedding model only `DefaultRequiredFiles = ["config.json"]`.
- `IsModelComplete` then accepts the model as soon as `config.json` exists **and** `hasModelWeights` finds any weight file — and `hasModelWeights` walks recursively and matches `*.onnx`, so the nested `onnx/layer-*/model.onnx` shards satisfy it.

So a partial ONNX-only directory is treated as "ready", `EnsureModels` skips it, and the snapshot download that would back-fill `model.safetensors` / `tokenizer.json` never runs. (This is also why the manual `hf download model.safetensors tokenizer.json …` workaround fixes it.)

## Fix

Mark the configured semantic embedding model (`EmbeddingModels.MmBertModelPath`) as requiring `model.safetensors` + `tokenizer.json` for completeness. A partial ONNX-only directory is then detected as incomplete and the full snapshot is re-downloaded.

Scoped deliberately:
- Only the **embedding model** path is affected — classifier/LoRA-adapter dirs (which legitimately ship without a top-level `tokenizer.json` / safetensors) are untouched.
- The downloader already fetches the whole repo, so this only changes **which partial states count as complete**; it does not narrow what gets downloaded. ONNX/ROCm deployments still get a correct (full) snapshot.

The upstream repo `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` contains both `model.safetensors` and `tokenizer.json`, so this is a zero-false-positive requirement for the default config.

## Test Plan

New `pkg/modeldownload/embedding_completeness_test.go` (model-free, pure Go):

- `TestBuildModelSpecsRequiresEmbeddingModelWeightsAndTokenizer` — the embedding spec's `RequiredFiles` now includes `model.safetensors` + `tokenizer.json`.
- `TestOnnxOnlyEmbeddingDirReportedIncomplete` — reproduces #2172: a dir with `config.json` + nested `onnx/layer-6/model.onnx` (no safetensors/tokenizer) is reported **incomplete** (RED before the fix — it was reported complete via the nested-onnx heuristic).
- `TestFullEmbeddingDirReportedComplete` — control: a dir with `config.json` + `model.safetensors` + `tokenizer.json` is complete.

Verified:
- `go test ./pkg/modeldownload/` (incl. `-race`) — green, no regressions.
- `gofmt`, `go vet`, `golangci-lint` — 0 issues.
- `make agent-lint CHANGED_FILES=...` — all gates Passed (structural lint, structure check, reference config contract, AST scan).

## Related

This is one half of getting `router_dc` + mmbert working out of the box. The other half — the selection path routing mmbert through the qwen3-only batched embedding API — is #2173, addressed by #2192.

Fixes #2172

### DIFF
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -98,6 +98,7 @@ func BuildModelSpecs(cfg *config.RouterConfig) ([]ModelSpec, error) {
 	// Extract all model paths from config
 	paths := filterDisabledOptionalModelPaths(cfg, ExtractModelPaths(cfg))
 	requiredFilesByModel := ExtractRequiredFilesByModel(cfg)
+	addEmbeddingModelRequiredFiles(cfg, requiredFilesByModel)
 
 	// Allow empty paths for API-only configurations
 	if len(paths) == 0 {
@@ -134,6 +135,32 @@ func BuildModelSpecs(cfg *config.RouterConfig) ([]ModelSpec, error) {
 	}
 
 	return specs, nil
+}
+
+// embeddingModelWeightFiles are the files the candle embedding runtime loads to bring a
+// semantic embedding model up. They are deliberately stricter than the nested-weight
+// heuristic in IsModelComplete: a directory holding only config.json + onnx/ (the layout
+// shipped in the image) otherwise satisfies that heuristic via the nested *.onnx files and
+// the safetensors/tokenizer download is never triggered, leaving embedding_ready=false (#2172).
+var embeddingModelWeightFiles = []string{"model.safetensors", "tokenizer.json"}
+
+// addEmbeddingModelRequiredFiles marks the configured semantic embedding model as requiring
+// its safetensors weights and tokenizer so a partial (ONNX-only) directory is detected as
+// incomplete and the full snapshot is re-downloaded.
+func addEmbeddingModelRequiredFiles(cfg *config.RouterConfig, requiredFilesByModel map[string][]string) {
+	// MmBertModelPath holds the configured semantic embedding model directory.
+	path := cfg.MmBertModelPath
+	if path == "" || !strings.HasPrefix(path, "models/") {
+		return
+	}
+
+	existing := requiredFilesByModel[path]
+	for _, fileName := range embeddingModelWeightFiles {
+		if !slices.Contains(existing, fileName) {
+			existing = append(existing, fileName)
+		}
+	}
+	requiredFilesByModel[path] = existing
 }
 
 // ExtractRequiredFilesByModel derives per-model completeness requirements from
--- a/src/semantic-router/pkg/modeldownload/embedding_completeness_test.go
+++ b/src/semantic-router/pkg/modeldownload/embedding_completeness_test.go
@@ -1,0 +1,122 @@
+package modeldownload
+
+import (
+	"os"
+	"path/filepath"
+	"slices"
+	"testing"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+const (
+	testEmbeddingModelPath = "models/mmbert-embed-32k-2d-matryoshka"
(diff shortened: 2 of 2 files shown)


WUKUNTAI-0211 (approved):
LGTM — nice, tightly-scoped fix.

Root-cause analysis is spot on: `hasModelWeights` walks recursively and matches `*.onnx`, so an ONNX-only directory satisfies the weight heuristic and the snapshot back-fill never runs. Adding `model.safetensors` + `tokenizer.json` to the embedding model's completeness contract is the right lever — and since `RequiredFiles` only feeds `IsModelComplete` (not the download include-filter), this changes which partial states count as complete without narrowing what gets downloaded. Scoping it to `MmBertModelPath` keeps classifier/LoRA-adapter dirs untouched, and the tests are RED-first with proper reproduction + control cases.

Approving.

Issue #2193, opened 2026-06-15
Support Fusion API for multi-model deliberation


Xunzhuo:
## Summary

Add a Fusion-style API to vLLM Semantic Router for multi-model deliberation, similar to OpenRouter Fusion.

The target user experience should support two entry points:

```json
{
  "model": "vllm-sr/fusion",
  "messages": [
    {"role": "user", "content": "What are the strongest arguments for and against ...?"}
  ]
}
```

and request-level customization through a plugin payload:

```json
{
  "model": "vllm-sr/fusion",
  "messages": [
    {"role": "user", "content": "..."}
  ],
  "plugins": [{
    "id": "fusion",
    "model": "judge-model-name",
    "analysis_models": [
      "analysis-model-a",
      "analysis-model-b",
      "analysis-model-c"
    ]
  }]
}
```

We may also want an OpenRouter-compatible alias, for example `openrouter/fusion`, if that fits the project API policy.

## Motivation

Fusion is different from the current router selection path. It is not just selecting one best model. A Fusion-style request should:

1. fan out the prompt to multiple analysis models,
2. run those calls in parallel,
3. ask a judge/calling model to produce structured analysis across the panel outputs,
4. use that analysis to produce the final response,
5. degrade gracefully when some panel models fail.

This gives users a higher-quality deliberation mode while still exposing it through an OpenAI-compatible request shape.

Reference: https://openrouter.ai/blog/announcements/fusion-beats-frontier/

## Current state

vLLM Semantic Router already has related building blocks, but not the Fusion API or exact semantics:

- The `selection` package currently chooses one selected model for a request.
- The `looper` package is the closest execution substrate for multi-model workflows.
- `ReMoM` already implements multi-round parallel reasoning plus synthesis, so it is likely the closest starting point.
- The existing looper algorithms (`confidence`, `ratings`, `remom`, etc.) do not expose a request-level Fusion plugin API.
- A direct model slug such as `vllm-sr/fusion` or `openrouter/fusion` would currently need special routing behavior rather than being treated as a normal explicit model name.

Relevant areas to inspect:

- `src/semantic-router/pkg/looper/`
- `src/semantic-router/pkg/extproc/processor_req_body.go`
- `src/semantic-router/pkg/extproc/req_filter_looper.go`
- `src/semantic-router/pkg/config/decision_config.go`
- `src/semantic-router/pkg/config/routing_surface_catalog.go`
- `src/vllm-sr/cli/algorithms.py`
[...]

PR #2198, opened 2026-06-15
[Router] Report real token usage across looper algorithms

Peterren:
## What

Make every looper algorithm report **real aggregated token usage** instead of the hardcoded `usage: {0, 0, 0}` block.

Today the looper substrate parses each backend completion into `ModelResponse` but discards the OpenAI `usage` block, so `confidence`, `ratings`, `remom`, `rl_driven`, and the base looper all emit zeros. This change captures per-call usage and sums it across every model call an execution makes.

## Why

Multi-model loopers are exactly where token accounting matters most (they make N calls per request), yet they were the only paths reporting zero usage — breaking cost/usage visibility for callers, dashboards, and metrics. This is also a prerequisite for the Fusion API (#2193): a Fusion looper fans out to a panel + judge + synthesis and needs to report the summed cost. This PR lands the reusable substrate so Fusion (and every other looper) gets correct usage for free.

## Changes

- **`looper/usage.go` (new):** `TokenUsage{Prompt,Completion,Total}` with nil-safe `SumUsage(...)` / `Add(...)` (non-mutating) and a `Map()` helper that renders the OpenAI-compatible usage block — the single seam loopers use in place of the `{0,0,0}` literal.
- **`looper/client.go`:** add `ModelResponse.Usage`, populated from `completion.Usage` in `parseNonStreamingResponse` (was parsed, then dropped).
- **`looper/looper.go`:** add `Response.Usage` so extproc/dashboard/metrics can read totals without re-parsing the body.
- **Wire all algorithms:** `base` and `ratings` sum the responses they hold; `remom` accumulates across rounds; `rl_driven` accumulates over both the selector and fallback paths; `confidence` includes **every cascade attempt** (not just the accepted answer) so usage reflects the full cost. Both JSON and streaming return paths set `Response.Usage`.

Substrate-only — no API/config/behavior change beyond `usage` going from zeros to real numbers. Streaming sub-calls (no `stream_options.include_usage`) still report zero for that call; non-streaming sub-calls (which all looper internal calls are) report real counts.

## Test plan

**Unit + integration (`pkg/looper/usage_test.go`, `usage_integration_test.go`):**
- `SumUsage`/`Add` summation, nil-safety, no-receiver-mutation; `Map()` JSON shape.
- `parseNonStreamingResponse` lifts `usage` into `ModelResponse.Usage`.
- Base-looper body + `Response.Usage` aggregation.
- **Integration:** `BaseLooper.Execute` against a real in-process `httptest` backend that returns a `usage` block → asserts the wrapped completion sums per-call usage (20/40/60 from two 10/20/30 calls). Self-contained — no external services, config, or auth.

```
make test-semantic-router      # all looper tests green
make agent-ci-lint             # full gate (pre-commit, ruff, go-lint, structure_check) → exit 0
```

**Full-stack e2e (manual, macOS dockerless):** ran curl → envoy → router extproc → `ratings` looper → two `llm-katan` echo backends (:8000/:8001) → aggregated response. A decision configured with `algorithm.type: ratings` over two models returned:

```json
"model": "Model-A,Model-B",
"choices": [ /* Model-A */, /* Model-B */ ],
"usage": { "prompt_tokens": 18, "completion_tokens": 34, "total_tokens": 52 }
```

Each backend reported `{9, 17, 26}`, so **18/34/52 = exactly 2× a single call** — confirming real cross-call aggregation through the live stack (previously this block was `{0,0,0}`). Router logs confirmed `looper_execution_started algorithm: ratings` → `execution_completed successful_models: 2`.

## Related

Substrate for #2193 (Fusion API): when the Fusion looper lands, it reports usage by summing panel + judge + synthesis `ModelResponse.Usage` via the same `SumUsage(...).Map()` seam.

### DIFF
--- a/src/semantic-router/pkg/looper/base.go
+++ b/src/semantic-router/pkg/looper/base.go
@@ -176,6 +176,8 @@ type AggregatedResponse struct {
 // When the final response contains tool_calls, the original raw response
 // is preserved (with metadata patched) to avoid dropping tool_calls.
 func (l *BaseLooper) formatJSONResponse(agg *AggregatedResponse, modelsUsed []string, iterations int) (*Response, error) {
+	usage := SumUsage(agg.Responses...)
+
 	// If the final response has tool_calls, use the original raw response
 	// but patch the model name and id to reflect the looper wrapper.
 	if agg.HasToolCalls && len(agg.Responses) > 0 {
@@ -194,6 +196,7 @@ func (l *BaseLooper) formatJSONResponse(agg *AggregatedResponse, modelsUsed []st
 						ModelsUsed:    modelsUsed,
 						Iterations:    iterations,
 						AlgorithmType: "simple",
+						Usage:         usage,
 					}, nil
 				}
 			}
@@ -214,6 +217,7 @@ func (l *BaseLooper) formatJSONResponse(agg *AggregatedResponse, modelsUsed []st
 				ModelsUsed:    modelsUsed,
 				Iterations:    iterations,
 				AlgorithmType: "simple",
+				Usage:         usage,
 			}, nil
 		}
 	}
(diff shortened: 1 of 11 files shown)


Xunzhuo (approved):
LGTM

PR #2197, opened 2026-06-15
Feat: Support Fusion API for multi-model deliberation

Xunzhuo:
## Summary

- add Fusion looper support for multi-model deliberation with concurrent panel calls, judge analysis, final synthesis, partial-failure handling, recursion protection, and aggregated token usage
- support direct Fusion slugs with `vllm-sr/fusion` enabled by default and `openrouter/fusion` available only as an opt-in compatibility alias
- implement Fusion-only decision matching for `model: "vllm-sr/fusion"`: requests still run signal extraction, but only decisions with `algorithm.type: fusion` are eligible; if none match, the router returns a clear no-match error instead of falling back to a normal single-model route
- add namespaced Auto aliases with `global.router.auto_model_names`; defaults are `vllm-sr/auto`, `auto`, and `MoM`, and operators can provide an explicit allow-list
- preserve the separation of concerns: `global` registers API entry aliases and looper runtime integration only; judge model, analysis panel, concurrency, token budget, templates, and error behavior remain per-decision policy under `routing.decisions[].algorithm.fusion`
- support request-level `plugins[].id == "fusion"` overrides for judge model, analysis models, concurrency, sampling, and error behavior; if no Fusion decision matches and `analysis_models` is provided, the router builds a request-scoped `fusion_direct` decision
- wire the new Fusion/Auto behavior through router config, canonical export/import, `/v1/models`, Python CLI schema/runtime mutation, dashboard config UI, docs, examples, and agent plan metadata
- fix looper internal alias routing so panel aliases route via `x-selected-model` and rewrite to provider model IDs without stale `content-length`
- harden the Kubernetes integration workflow against GitHub runner Chrome apt source mirror-sync failures

Closes #2193.

## API behavior

| Request entry | Routing behavior |
| --- | --- |
| `model: "vllm-sr/auto"` | Run full vLLM-SR signal + decision policy across all decisions. If the selected decision uses `algorithm.type: fusion`, execute Fusion; otherwise run the normal selected route. |
| `model: "auto"` / `model: "MoM"` | Legacy Auto aliases, configurable through `global.router.auto_model_names`. |
| `model: "vllm-sr/fusion"` | Run signal extraction, then evaluate Fusion-capable decisions only. No fallback to non-Fusion decisions. |
| `plugins[].id: "fusion"` | Request-scoped Fusion override for judge/panel/runtime knobs. Can provide `analysis_models` to build `fusion_direct` when no Fusion decision matches. |

## Validation

- `go test ./pkg/looper`
[...]

PR #2202, opened 2026-06-16
feat: recognize comma-separated auto_model_name aliases in IsAutoModelName

### LINKED ISSUE #2116: feature: When Claude Code GUI enables 1M context, it appends `[1M]` to model names (e.g., `MoM[1M]`), but `IsAutoModelName` doesn't recognize the suffix, so requests bypass intelligent routing.

mvanhorn:
Closes #2116

## Purpose

This makes `auto_model_name` accept a comma-separated list of aliases so a single router config such as `auto_model_name: "MoM, MoM[1M]"` recognizes every client variant of the same auto-routing entrypoint. This affects the `Router` module.

Different Claude clients send different model IDs for the same auto entrypoint, which is the heart of the reported problem. Claude CLI requires the `[1M]` suffix when the 1M context window is on and sends `MoM[1M]`, while the cc GUI strips that suffix and sends `MoM`. Because `IsAutoModelName` previously matched only the single configured name, one of those two variants always failed the auto check, silently bypassed intelligent routing, and fell through to direct passthrough. As the reporter and collaborators converged on in the issue thread, this is really an aliasing and normalization concern rather than a new routing semantic, so handling it inside the existing config field avoids forcing operators to stand up an extra NGINX or proxy rewrite layer just to reconcile the two client behaviors.

The change keeps the YAML field a plain string with no schema break: it is split on commas internally, whitespace is trimmed, and blank entries are ignored. `IsAutoModelName` returns true for `"auto"` (unchanged) or any alias in the set, while `GetEffectiveAutoModelName` still returns the primary (first) alias so the `/v1/models` listing and single-name display paths are unchanged. The `/v1/models` duplicate filters now skip any model that is itself an auto alias, so when `include_config_models_in_list` is enabled an alias that also exists in `model_config` is no longer advertised as a regular upstream backend even though selecting it would trigger automatic routing. Single-value and unset configurations behave exactly as they did before.

## Test Plan

From `src/semantic-router`, reviewers can build, vet, and test the three touched packages with `go build ./pkg/config/... ./pkg/extproc/... ./pkg/apiserver/...`, `go vet ./pkg/config/... ./pkg/extproc/... ./pkg/apiserver/...`, and `go test ./pkg/config/... ./pkg/extproc/... ./pkg/apiserver/...`. New Ginkgo specs in `pkg/config/config_test.go` cover multi-alias matching, whitespace trimming, backward compatibility for both unset and single-value configs, the negative case for a non-alias model, and blank-entry handling. This is sufficient because the alias logic is pure config behavior fully exercisable by unit tests, and the listing-filter change is covered by the existing `/v1/models` tests in `pkg/apiserver`.

## Test Result

All three packages build, vet clean, and pass their tests. The new and existing `AutoModelName Configuration` specs pass and the `/v1/models` listing tests in `pkg/apiserver` continue to pass. There are no follow-up risks or blockers since the change is additive and backward compatible.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_models.go
+++ b/src/semantic-router/pkg/apiserver/route_models.go
@@ -43,8 +43,9 @@ func (s *ClassificationAPIServer) handleOpenAIModels(w http.ResponseWriter, _ *h
 	// Append underlying models from config (if available and configured to include them)
 	if cfg != nil && cfg.IncludeConfigModelsInList {
 		for _, m := range cfg.GetAllModels() {
-			// Skip if already added as the configured auto model name (avoid duplicates)
-			if m == cfg.GetEffectiveAutoModelName() {
+			// Skip any model that is itself an auto model alias (avoid duplicates and
+			// avoid advertising an auto alias as a regular upstream backend)
+			if cfg.IsAutoModelName(m) {
 				continue
 			}
 			models = append(models, OpenAIModel{
--- a/src/semantic-router/pkg/config/config_test.go
+++ b/src/semantic-router/pkg/config/config_test.go
@@ -2249,6 +2249,15 @@ api:
 				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("CustomAuto"))
 			})
 
+			It("should return primary alias when AutoModelName contains multiple aliases", func() {
+				cfg := &RouterConfig{
+					RouterOptions: RouterOptions{
+						AutoModelName: "MoM, MoM[1M]",
+					},
+				}
+				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("MoM"))
+			})
+
 			It("should return default 'MoM' when AutoModelName is not set", func() {
 				cfg := &RouterConfig{
 					RouterOptions: RouterOptions{
@@ -2289,7 +2298,30 @@ api:
 						AutoModelName: "",
 					},
 				}
+				Expect(cfg.IsAutoModelName("auto")).To(BeTrue())
 				Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
+			})
+
+			It("should recognize comma-separated AutoModelName aliases", func() {
+				cfg := &RouterConfig{
+					RouterOptions: RouterOptions{
+						AutoModelName: "MoM, MoM[1M]",
+					},
+				}
+				Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
+				Expect(cfg.IsAutoModelName("MoM[1M]")).To(BeTrue())
+				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("MoM"))
+			})
+
+			It("should trim whitespace around comma-separated AutoModelName aliases", func() {
+				cfg := &RouterConfig{
+					RouterOptions: RouterOptions{
+						AutoModelName: " MoM , MoM[1M] ",
+					},
+				}
+				Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
+				Expect(cfg.IsAutoModelName("MoM[1M]")).To(BeTrue())
+				Expect(cfg.GetEffectiveAutoModelName()).To(Equal("MoM"))
 			})
 
 			It("should not recognize other model names as auto", func() {
@@ -2299,6 +2331,7 @@ api:
 					},
 				}
 				Expect(cfg.IsAutoModelName("gpt-4")).To(BeFalse())
+				Expect(cfg.IsAutoModelName("gpt-4o")).To(BeFalse())
 				Expect(cfg.IsAutoModelName("claude")).To(BeFalse())
 			})
 
@@ -2311,6 +2344,16 @@ api:
 				Expect(cfg.IsAutoModelName("auto")).To(BeTrue())
 				Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
 				Expect(cfg.IsAutoModelName("other")).To(BeFalse())
+			})
+
+			It("should ignore blank entries in comma-separated AutoModelName aliases", func() {
+				cfg := &RouterConfig{
+					RouterOptions: RouterOptions{
+						AutoModelName: "MoM,,",
+					},
+				}
+				Expect(cfg.IsAutoModelName("")).To(BeFalse())
+				Expect(cfg.IsAutoModelName("MoM")).To(BeTrue())
 			})
 		})
 
--- a/src/semantic-router/pkg/config/helper.go
+++ b/src/semantic-router/pkg/config/helper.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 )
 
+const defaultAutoModelName = "MoM"
+
 // GetModelReasoningFamily returns the reasoning family configuration for a given model name
 func (rc *RouterConfig) GetModelReasoningFamily(modelName string) *ReasoningFamilyConfig {
 	familyName, ok := rc.GetModelReasoningFamilyName(modelName)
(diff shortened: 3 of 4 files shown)


Xunzhuo:
@mvanhorn can you rebase? a new PR looks like covered this area

Xunzhuo:
@mvanhorn thanks for the contribution btw, new version we supported a slice for auto name aliases

Xunzhuo (commented):
Thanks for the contribution. This branch no longer needs a rebase: current `main` already has the same capability through the typed `auto_model_names: []` contract merged in #2197.

The mainline implementation includes normalization/deduplication, explicit allow-list behavior, default `vllm-sr/auto` / `auto` / `MoM` compatibility, YAML tests, and a single `IsAutoModelName` path. Keeping comma-separated aliases inside the legacy scalar would create a second, less-typed configuration format and the branch now conflicts with that implementation.

My recommendation is to close this PR as superseded by #2197; no conflict-resolution work is needed from you. If you see a concrete alias case that the list-based contract still misses, please point it out or send a small focused test/fix against current `main`.

mvanhorn:
Thanks @Xunzhuo, that makes sense — #2197's typed `auto_model_names` contract is the cleaner path, and I don't want to introduce a competing scalar-alias format. Closing this in favor of the mainline implementation. Appreciate the detailed context!

PR #2214, opened 2026-06-17
[Router] Grounding-aware synthesis for Fusion (faithfulness-scored panel)

Peterren:
## What

Add **grounding-aware synthesis** to the Fusion looper: an optional stage that scores each panel response for *faithfulness* before the judge runs, then ranks/filters the panel so synthesis works from the most-grounded responses. Off by default — when disabled, Fusion behaves exactly as today.

This is the first of the "new deliberation algorithms" direction (see the proposal doc added here), and it's a differentiator OpenRouter's Fusion structurally can't match: its judge is a bare LLM with no grounding oracle, whereas vSR is a classifying gateway with a built-in hallucination/groundedness detector and an NLI entailment model.

## Why

OpenRouter's Fusion benchmark (DRACO) explicitly penalizes confident-but-wrong answers, and judge choice alone swings scores 10–25 points. vSR can score panel faithfulness with local encoder models (no extra LLM calls) and keep ungrounded/contradicted responses out of synthesis.

> Honest framing: grounding measures faithfulness/consistency, not truth. With no authoritative source we down-weight the least-supported responses, not certify correctness.

## Changes

- **`config/fusion_config.go`** — new `FusionGroundingConfig` (`enabled`, `reference`, `min_score`, `min_keep`, `nli_contradiction_penalty`, `on_error`) + `ValidateFusionGroundingConfig`. Python CLI mirror (`cli/algorithms.py`) with **identical bounds** (avoids the Go-allows-0 vs Python-ge=1 drift). Reference config + Go contract test + Python tests.
- **`looper/grounding.go` (new)** — hybrid reference scoring: hallucination **detector** against provided context when present, else cross-model **NLI** (the panel as its own mutual reference); rank + filter; `FusionGroundingTrace`. Kept hermetic — candle/CGO backends are injected via `SetGroundingBackends`, so the looper package stays unit-testable without the Rust lib.
- **`looper/fusion.go`** — run grounding between panel and judge; usage still summed from the **full** panel (real cost paid); grounding notes added to the judge prompt; grounding recorded in `FusionTrace` and surfaced in responses.
- **`classification`** — wire the candle `ClassifyNLI` + detector into the looper at hallucination-detector init (the single deterministic init point; no extproc changes).
- **Docs** — Fusion tutorial grounding section + `proposals/deliberation-algorithms.md`.

Substrate reuse: the hallucination detector (`pkg/classification`), the NLI binding (`candle-binding`), and the `SumUsage` seam from #2198 — all already present.

## Test plan

**Unit + integration (Go, hermetic — stubbed backends, no CGO):** `pkg/looper/grounding_test.go` covers panel/context scoring, `min_score`/`min_keep` filtering, reference resolution, context extraction, `on_error` fallback, and an `Execute`-level test asserting a contradicted panel response never reaches the judge while usage reflects the full panel. **Python:** `test_algorithm_config.py` covers grounding defaults + bounds (mirrors the Go validator).

```
go test ./pkg/looper/... ./pkg/config/... ./pkg/classification/...   # green
make agent-ci-lint                                                   # exit 0
```

**Full-stack e2e (manual, macOS dockerless):** ran curl → envoy → router extproc → fusion looper → real `models/mom-halugate-detector` over two `llm-katan` echo backends, with a `reference: context` grounding config:

- Startup: `hallucination_detector_initialized` (confirms the grounding backends wire at init).
- Per request: `fusion_execution_started` → `fusion_grounding_applied reference_mode: context panel_in:2 panel_kept:2`; response `fusion.grounding.scores[]` carried real per-model scores and `flagged_spans` (the detector correctly flagged content unsupported by the provided context).
- Drop path (`min_score: 0.6`): `panel_in:2 panel_kept:1`, trace showed the lower-scoring response `dropped: true` and the judge synthesized from the survivor.

(The cross-model NLI `panel`/`hybrid` path needs an NLI model that isn't bundled in the dev checkout; the `context` path was used for the live e2e. Unit tests cover the panel path with stubs.)

## Related

- Builds on #2197 (Fusion API, merged) and #2198 (looper usage accounting, merged).
- Part of the response to #2193 (Fusion API for multi-model deliberation).

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -806,6 +806,16 @@ routing:
           analysis_template: ""
           synthesis_template: ""
           judge_prompt_version: fusion-v1
+          # Grounding-aware synthesis: score panel responses for faithfulness
+          # (hallucination detector against context, else cross-model NLI) and
+          # rank/filter them before the judge. Disabled by default.
+          grounding:
+            enabled: false
+            reference: hybrid          # hybrid | context | panel
+            min_score: 0.0             # drop responses scoring below this (0-1)
+            min_keep: 1                # always keep at least this many
+            nli_contradiction_penalty: 1.0
+            on_error: skip             # skip (fall back to plain fusion) | fail
 
     - name: terse_elo_route
       description: Personalized terse-answer route using Elo selection.
--- a/src/semantic-router/pkg/classification/classifier_fact_hallucination_lifecycle.go
+++ b/src/semantic-router/pkg/classification/classifier_fact_hallucination_lifecycle.go
@@ -3,6 +3,8 @@ package classification
 import (
 	"fmt"
 
+	candle "github.com/vllm-project/semantic-router/candle-binding"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/looper"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
 )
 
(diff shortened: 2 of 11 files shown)


Xunzhuo (approved):
LGTM!

PR #2228, opened 2026-06-17
[Dashboard] Helm chart: fix Jaeger proxy target to existing service name

drivebyer:
## Purpose

The Helm chart wires the dashboard's `TARGET_JAEGER_URL` to
`<release>-jaeger-query:16686`, but the bundled Jaeger subchart (all-in-one
mode) only renders a single Service `<release>-jaeger` — the `16686` query
port is already exposed on it, and there is no `<release>-jaeger-query`
Service. As a result the dashboard's `/embedded/jaeger/` reverse proxy can't
resolve the host and the **Tracing page returns 502 Bad Gateway**.

## Test Plan

- `helm template t deploy/helm/semantic-router --set dependencies.observability.jaeger.enabled=true`:
  `TARGET_JAEGER_URL` now renders `http://t-jaeger:16686`, matching the only
  rendered jaeger Service (`t-jaeger`, port `16686`).
- kind end-to-end (dashboard + jaeger enabled): `GET /embedded/jaeger/`
  returns **502** before this change and **200** after, with the Jaeger UI
  loading.


AayushSaini101 (approved):
LGTM thanks

PR #2218, opened 2026-06-17
[Router] v0.4 protocol headers: rename + cross-protocol-only + x-vsr-debug trigger

theohsiung:
## Summary

Closes #2206. Closes #2216. Part of #2200 — v0.4 VSR response header contract.

Reworks the three protocol-translation response headers for v0.4, and introduces the shared debug-mode trigger they (and #2205) key off.

## What changed

**Rename** (no behavior change on its own):
- `x-vsr-inbound-protocol` → `x-vsr-client-protocol`
- `x-vsr-outbound-protocol` → `x-vsr-upstream-protocol`
- `x-vsr-lossiness-warnings` → `x-vsr-protocol-warnings` (already emits only when warnings exist)

**Conditional emission** (#2206): `client` / `upstream` markers now ride only on **cross-protocol** responses (client protocol ≠ upstream) — same-protocol calls omit them to keep the contract lean. Centralised in an `addProtocolMarkers` builder helper.

**Debug trigger** (#2216): a new `x-vsr-debug: true` request header opts a request into verbose headers. The shared `debugHeadersRequested(ctx)` predicate gates it; when set, the protocol markers emit even on same-protocol responses. #2205 (demote) will reuse the same predicate to re-surface its demoted intermediates.

> Scope note: the contract's "debug/replay mode" clause had no usable trigger today (replay defaults on, so replay-id can't gate). #2216 settled on the per-request `x-vsr-debug` header — lightest, request-scoped, no config-schema change.

## Files

- `pkg/headers/headers.go` — rename the 3 constants, add `VSRDebug`.
- `pkg/extproc/processor_res_header_mutation.go` — `addProtocolMarkers` (cross-protocol + debug gate).
- `pkg/extproc/debug_mode.go` (new) — shared `debugHeadersRequested` predicate.
- `pkg/extproc/debug_mode_test.go` (new), `processor_res_header_mutation_test.go`, `extproc_test.go` — tests.
- `pkg/ir/warning_reasons.go` — comment.
- `e2e/testcases/anthropic_messages_headers.go` — renamed header reads (cross-protocol case, markers still present).
- `website/docs/troubleshooting/vsr-headers.md` — document `x-vsr-debug` + conditional emission.

## Test plan

- [x] `go build` + `go vet` — pkg/headers, pkg/extproc, pkg/ir, pkg/utils/http, e2e
- [x] `go test ./pkg/headers/` (incl. headers-are-documented) and targeted `pkg/extproc` protocol/debug/header tests — pass
[...]

Issue #2216, opened 2026-06-17
[Router] v0.4 headers: define a debug/replay mode trigger for verbose headers


theohsiung:
Part of #2200.

Several v0.4 headers are specced to emit "only in debug/replay mode" — `x-vsr-client-protocol` / `x-vsr-upstream-protocol` (#2206) and the demoted intermediate headers (#2205: `matched-*`, `cache-*`, `tools-*`, …). But there's **no debug-mode trigger in the router today** (replay is on by default, so gating on replay-id is a no-op).

This issue defines that shared primitive — one "verbose/debug headers" switch the other slices key off.

Trigger options considered:

1. **Per-request header** (`x-vsr-debug: true`) — request-scoped, no config-schema change, opt-in per call. Lightest.
2. **Config setting** (`global.observability.debug_headers: true`) — router-wide, operator-controlled.
3. **Log level** (`SR_LOG_LEVEL=debug`) — reuse an existing knob, no new surface.

Planning to go with **option 1 (`x-vsr-debug` request header)** unless there's a reason to prefer router-wide control — it's the lightest and lets a caller pull full detail per request without changing global behavior. #2205 (demote) and #2206 (protocol) will gate their detail headers on it. Feedback welcome.

Issue #2204, opened 2026-06-16
[Router] v0.4 headers: merge response warnings into x-vsr-response-warnings


theohsiung:
Part of #2200 — v0.4 VSR response header contract redesign.

Fold the individual response-warning booleans into a single comma-separated `x-vsr-response-warnings`, emitted only when warnings exist:

- `x-vsr-hallucination-detected` → `hallucination`
- `x-vsr-unverified-factual-response` → `unverified_factual`
- `x-vsr-response-jailbreak-detected` → `response_jailbreak`

Detail headers (`fact-check-needed`, `verification-context-missing`, `hallucination-spans`, `response-jailbreak-type` / `-confidence`) move to replay/debug rather than the default surface.

**Heads-up:** widest blast radius of the set — reaches the replay stores, metrics, and canonical export, not just `res_filter_hallucination.go` / `res_filter_jailbreak.go`.

Issue #2200, opened 2026-06-16
[Feature] Redesign VSR response header contract for v0.4


Xunzhuo:
## Describe the feature

Redesign the public `x-vsr-*` response header contract for v0.4 as a breaking change.

Today the response header surface exposes many intermediate pipeline details from before or inside the `signal -> decision -> algorithm -> plugin` flow, including domain/category classification, matched signal families, jailbreak/PII signal hits, cache internals, tools internals, looper internals, and plugin execution details. This makes the response contract noisy and hard to reason about.

For v0.4, response headers should only expose final response facts:

| Header | Emit condition | Purpose |
| --- | --- | --- |
| `x-vsr-schema-version` | Every VSR-processed response | Header contract version, for example `2` |
| `x-vsr-response-path` | Every VSR-processed response | Final response path: `upstream`, `cache`, `fast_response`, `looper`, `image_generation`, `blocked`, `rate_limited`, or `error` |
| `x-vsr-replay-id` | When replay id exists | Entry point for complete routing/debug trace |
| `x-vsr-selected-decision` | When a final decision exists | Final selected decision |
| `x-vsr-selected-model` | When a final model exists | Final selected model |
| `x-vsr-selected-confidence` | When confidence exists | Final decision confidence |
| `x-vsr-response-warnings` | When final response warnings exist | Comma-separated warning codes such as `hallucination`, `unverified_factual`, `response_jailbreak` |
| `x-vsr-client-protocol` | Only for cross-protocol handling or debug/replay mode | Inbound client protocol |
| `x-vsr-upstream-protocol` | Only for cross-protocol handling or debug/replay mode | Outbound upstream protocol |
| `x-vsr-protocol-warnings` | Only when protocol compatibility/conversion warnings exist | Protocol warning codes/details |

## Primary layer

global level

## Why this layer?

This is a cross-cutting response contract change. It touches router response header emission, immediate responses, response warning plugins, dashboard display allowlists, docs, and tests. The goal is not to add a new signal or plugin behavior, but to define the public wire contract for final routing/response outcomes.

## What should be removed or changed

| Current header | v0.4 handling |
| --- | --- |
| `x-vsr-cache-hit` | Replace with `x-vsr-response-path: cache` |
| `x-vsr-fast-response` | Replace with `x-vsr-response-path: fast_response` |
| `x-vsr-looper-*` | Replace public exposure with `x-vsr-response-path: looper`; details go to replay |
[...]

PR #2229, opened 2026-06-17
[Router] v0.4 headers: demote intermediate signal/decision/tool/looper/cache/retention headers to x-vsr-debug

### LINKED ISSUE #2205: [Router] v0.4 headers: demote intermediate signal/decision headers to replay/debug
Part of #2200 — v0.4 VSR response header contract redesign.

Drop intermediate signal/decision/algorithm/plugin headers from the default response surface, keeping them available in replay/debug:

`x-vsr-matched-*`, `x-vsr-selected-category`, `x-vsr-selected-modality`, `x-vsr-selected-reasoning`, `x-vsr-cache-similarity`, `x-vsr-tools-*`, `x-vsr-context-token-count`, `x-vsr-injected-system-prompt`, `x-vsr-looper-*`, `x-vsr-retention-*`.

**Ordering:** depends on the keystone slice (`x-vsr-schema-version` + `x-vsr-response-path`) — these removals only make sense once `x-vsr-response-path` replaces the path-specific signals, otherwise the dashboard allowlists and header assertions break mid-sequence.

**Where the details go (not lost, just off the wire):** every response keeps `x-vsr-replay-id` as the entry point. The demoted fields live in the router replay record, inspected via:
- Replay API — `GET /v1/router_replay/{replay-id}` (single record), plus `/v1/router_replay` (list/filter), `/aggregate`, `/trajectory?session_id=…`
- the dashboard replay view (proxies the same API, with redaction)

Replay is enabled by default (postgres store, 30-day TTL), so this slice moves detail *off the default headers*, it does not remove observability.

**Before removing each header, confirm it has a home in the replay record** (`store.Record` in `pkg/routerreplay/store/store.go`) — otherwise demoting it silently drops observability. Most are already covered (`Signals`/`SignalConfidences`/`SignalValues` ← `matched-*`, `Category`, `Modality`, `ReasoningMode`, `ToolTrace`/`ToolDefinitions` ← `tools-*`, `FromCache`). The following have **no dedicated field yet** and need one added to `store.Record` + the recorder as part of this slice (or a documented reason they don't need it):

- [ ] `x-vsr-cache-similarity` — only `RAGSimilarityScore` exists (RAG, not cache)
- [ ] `x-vsr-context-token-count` — usage tokens are stored, but not this context count
- [ ] `x-vsr-injected-system-prompt` — recoverable from `RequestBody`, but no explicit field
- [ ] `x-vsr-retention-*` — issue scopes this as an operator signal; decide whether it belongs in replay at all

theohsiung:
## What

Implements **#2205** (Part of **#2200** — v0.4 response header contract): demote the intermediate signal/decision/classification, matched-signal, tool-observability, looper-trace, cache-detail and retention headers off the default response surface, behind the `x-vsr-debug` trigger.

The v0.4 contract splits response headers into two surfaces:

- **Default surface** (every non-cache-hit response): keystone (`x-vsr-schema-version`, `x-vsr-response-path`) + final routing facts (`x-vsr-selected-decision`, `x-vsr-selected-confidence`, `x-vsr-selected-model`) + `x-vsr-replay-id`. Protocol markers ride only on cross-protocol handling; `x-vsr-protocol-warnings` only when present.
- **Debug surface** (only when the request sets `x-vsr-debug: true`): the intermediate details, matched signals, tool observability, looper trace, cache detail and retention directives. All remain recoverable from the replay record via `x-vsr-replay-id`.

This reuses the `debugHeadersRequested(ctx)` predicate introduced in #2218 (#2216).

## Changes (per path)

- **Main upstream path** (`processor_res_header_mutation.go`): split `addStandardDecisionHeaders` into `addFinalDecisionHeaders` (default) and `addDecisionDetailHeaders` (debug); `addMatchedSignalHeaders` debug-gated.
- **Tool selection** (`req_filter_tools.go`): `emitToolObservability` gates the three `x-vsr-tools-*` headers on debug; the Prometheus tools-retrieval metric stays unconditional.
- **Looper** (`req_filter_looper_response.go`): final routing facts stay on the default surface; the looper execution trace, matched signals, category and session-phase are demoted. `content-type` always rides.
- **Semantic cache** (`utils/http/response.go` + `req_filter_cache.go`): `x-vsr-cache-hit` and the selected-decision fact stay; category, cache-similarity and matched-keywords are passed through only under debug (`CreateCacheHitResponse` keeps its signature; the caller blanks the detail unless debug).
- **Retention** (`processor_res_header_mutation.go`): `x-vsr-retention-*` demoted to debug per #2200. The directive's runtime effects (cache write skip, TTL override, model-switch-gate stay) are applied internally and are **not** driven by the wire header, so demoting the header is observability-only and changes no behavior.

## E2E + docs

- 10 routing/classification e2e testcases opt into `x-vsr-debug` (routing logic unchanged); the retention e2e test, which reads only the final `x-vsr-selected-decision`, stays on the lean surface.
- `website/docs/troubleshooting/vsr-headers.md`: documents the default vs debug surfaces, flags which headers are debug-only, and shows a lean + a debug example.

## Out of scope (follow-ups)

- Removing the now-redundant `x-vsr-cache-hit` / `x-vsr-fast-response` markers (superseded by `x-vsr-response-path`) — separate cleanup.
- An operator config-gate for `x-vsr-debug` (currently client-controlled) — wide blast radius, dedicated PR.
- The warning-merge axis (`x-vsr-response-warnings`) is **#2204**, independent of this change.

## Test Plan

- [x] `make go-lint` (CI-equivalent, `src/semantic-router`) → 0 issues
- [x] `go test ./pkg/extproc/ ./pkg/utils/http/ ./pkg/headers/` → pass (new lean-default + debug-surface tests added per path)
- [x] e2e module `go build ./...` + `go vet ./testcases/` → clean
- [x] `make agent-lint` per changed file set → clean; `markdownlint` on the doc → clean
- [ ] Full e2e run on a kind cluster (CI / reviewer environment)

Closes #2205
Part of #2200


WUKUNTAI-0211:
Thanks for the clean split — the demotion itself looks complete and consistent across the upstream / looper / tools / cache paths, and the unit tests pin each surface well.

One blocker against the #2205 acceptance criteria, though. #2205 has **two** requirements, not one: demote the headers **and** "*confirm each has a home in the replay record before removing it — otherwise demoting silently drops observability*", with an explicit checklist for the four fields that have no dedicated `store.Record` field yet.

This PR doesn't touch `pkg/routerreplay/store/store.go`, and `Record` still only carries `RAGSimilarityScore` (RAG, not cache). So the PR description's "*all remain recoverable from the replay record via `x-vsr-replay-id`*" doesn't actually hold for those four. Going through the checklist:

- `x-vsr-retention-*` — ✅ justified here (runtime effects applied internally, observability-only). Good.
- `x-vsr-injected-system-prompt` — borderline: #2205 notes it's recoverable from `RequestBody`, which is a valid "documented reason", but that reasoning isn't stated anywhere in the PR. Please call it out explicitly.
- `x-vsr-cache-similarity` — ❌ now off the non-debug upstream surface with **no** replay field. Genuinely dropped.
- `x-vsr-context-token-count` — ❌ demoted on the looper path (the new `...DefaultSurfaceIsLean` test asserts its absence) with **no** replay field. Genuinely dropped.

Could we either (a) add `store.Record` fields + recorder wiring for `cache-similarity` and `context-token-count`, or (b) document a per-field exemption and tick the boxes on #2205? As written, these two regress observability rather than relocate it, which is the exact failure mode #2205 calls out.

Minor, separate from #2205: three e2e cases build raw requests and read now-demoted headers without setting `x-vsr-debug` — `embedding_signal_image_routing.go` (`x-vsr-matched-embeddings` → `SignalTriggered` always false), `mcp_common.go` (`x-vsr-selected-category` assertion), and `routing_fallback.go` (method inference). The `sendLocalChatCompletion` helper covers the rest; these three just need the same one-line header.

WUKUNTAI-0211 (dismissed):
Approving — the demote itself is clean and consistent across the upstream / looper / tools / cache paths, and the per-surface unit tests pin each surface well.

On my earlier #2205 blocker: agreed to **defer** the two genuinely-dropped replay fields rather than block this PR, since the demote is additive-safe and the values are still live under `x-vsr-debug`. Tracking the replay-record wiring in **#2254**, to land before v0.4 ships so this doesn't become a silent observability regression:

- `x-vsr-cache-similarity` → needs a `store.Record` field (only `RAGSimilarityScore` exists today).
- `x-vsr-context-token-count` → needs a `store.Record` field (usage tokens stored, but not this context count).

The other two boxes are resolved without new fields — please tick them on #2205 with the reasons:
- `x-vsr-injected-system-prompt` — recoverable from `RequestBody`.
- `x-vsr-retention-*` — runtime effects applied internally; observability-only.

**One thing worth fixing in this PR before merge** (cheap, one line each): three e2e cases read now-demoted headers without setting `x-vsr-debug`, so their assertions silently degrade to false rather than failing —
- `embedding_signal_image_routing.go` (`x-vsr-matched-embeddings` → `SignalTriggered` always false)
- `mcp_common.go` (`x-vsr-selected-category` assertion)
- `routing_fallback.go` (method inference)

Not blocking the approval, but please add the same `x-vsr-debug` header these three need so we don't merge dead assertions.

---

theohsiung:
Rebased onto latest `main` — the merge conflict is resolved, so this is `MERGEABLE` again and CI is re-running.

The only conflict was in `website/docs/troubleshooting/vsr-headers.md` (Decision Headers table). `main` had added the new `x-vsr-learning-*` headers to that table while this PR restructured it with the `Surface` column (#2205). Resolution keeps the `Surface` column and places all five `x-vsr-learning-*` headers on the **debug** surface — matching `addDecisionDetailHeaders` in `processor_res_header_mutation.go`, which only emits them under `x-vsr-debug`.

@WUKUNTAI-0211 your prior review was dismissed by the merge commit — would you mind taking another look?

WUKUNTAI-0211 (dismissed):
Re-approving — the DISMISSED state was the post-rebase auto-invalidation, not a substantive change.

Two commits since my 6/20 approval, both green:
- `b50082a` rebase: only conflict was vsr-headers.md (resolved correctly, x-vsr-learning-* under debug).
- `b21941d`: opts the two x-vsr-learning-* unit tests into x-vsr-debug — same dead-assertion fix I flagged on 6/20, extended consistently.

The three e2e cases from my 6/20 review were already handled via the shared debug helper in 220ff8d / 1127b4b.

WUKUNTAI-0211 (approved):
LGTM

PR #2232, opened 2026-06-17
feat(cli): add shell completion command

### LINKED ISSUE #2185: feature: enable autocomplete feature in the vllm-sr CLI
### Describe the feature

Currently, the VLLM-SR CLI doesn't support the autocomplete feature of the commands and arguments. Implement a proper way for the autocomplete feature in all the commands and arguments so that it would be easy for newcomers to understand the flow of the design properly and increase user experience.

Current Status: On clicking the tab the CLI must suggest the commands and arguments if the user selects a suitable command for the use case. 

Image

Important: To increase user experience across the semantic router

TwilightTechie:
Add 'vllm-sr completion <shell>' command that generates shell completion scripts for bash, zsh, and fish. Uses Click's built-in shell_completion module to produce the scripts.

The command auto-detects the current shell from the SHELL env var when no argument is given, and prints setup instructions as comments.

Closes #2185


TwilightTechie:
cc: @AayushSaini101

AayushSaini101:
> cc: @AayushSaini101

thanks @TwilightTechie for the approach but seems user need to execute the script to add the appropriate variable in the env, it possible we allow CLI completion direclty without executing any script, you can check glab command gitlab it supports the automate feature by default https://gitlab.com/gitlab-org/cli thanks

TwilightTechie:
Hi @AayushSaini101 got it, Should vllm-sr completion install modify ~/.zshrc / ~/.bashrc directly? Some users may prefer not to have tools auto-modify their shell configs. An alternative is to only write a file to a completions directory and print instructions. What's should be our preference? based on that I will share the updated commit.

TwilightTechie:
Thanks for the feedback @AayushSaini101! I've updated the PR to address the concern.
The completion command is now a group with two subcommands:

vllm-sr completion install — automatically writes the completion setup to the user's shell config (zero manual steps):
```$ vllm-sr completion install
✓ Shell completions for zsh installed to /Users/anup/.zshrc
  Restart your shell or run: source /Users/anup/.zshrc```
```
It's idempotent: safe to run multiple times:

```
$ vllm-sr completion install
✓ Shell completions for zsh are already configured in /Users/anup/.zshrc
```
Automatic setup via install.sh
install.sh now calls vllm-sr completion install automatically after CLI installation, so users get tab-completion out of the box:
```
[step] Setting up shell completions
[done] Shell completions installed
```
Supported shells
```
bash → appends to ~/.bashrc
zsh → appends to ~/.zshrc
fish → writes to ~/.config/fish/completions/vllm-sr.fish
```
Shell is auto-detected from $SHELL when not specified.

AayushSaini101:
> Thanks for the feedback @AayushSaini101! I've updated the PR to address the concern. The completion command is now a group with two subcommands:
> 
> vllm-sr completion install — automatically writes the completion setup to the user's shell config (zero manual steps):
> 
> ```
> ✓ Shell completions for zsh installed to /Users/anup/.zshrc
>   Restart your shell or run: source /Users/anup/.zshrc```
> ```
> 
> It's idempotent: safe to run multiple times:
> 
> ```
> $ vllm-sr completion install
> ✓ Shell completions for zsh are already configured in /Users/anup/.zshrc
> ```
> 
> Automatic setup via install.sh install.sh now calls vllm-sr completion install automatically after CLI installation, so users get tab-completion out of the box:
> 
> ```
> [step] Setting up shell completions
> [done] Shell completions installed
> ```
> 
> Supported shells
> 
> ```
> bash → appends to ~/.bashrc
> zsh → appends to ~/.zshrc
> fish → writes to ~/.config/fish/completions/vllm-sr.fish
> ```
> 
> Shell is auto-detected from $SHELL when not specified.

shall we consider this flow @Xunzhuo @FAUST-BENCHOU @shraderdm : )

Xunzhuo (approved):
LGTM

PR #2236, opened 2026-06-18
[Feature] Add a rerank endpoint with bi-encoder and cross-encoder backends

arun-elr:
Part of #2247

## Purpose

This PR adds first-class reranking to Semantic Router.

The long-term shape is a **model-managed cross-encoder rerank signal**: a bi-encoder shortlist re-scored by a cross-encoder, exposing relevance scores/margins that routing, RAG, semantic cache, and memory can consume. The `POST /v1/rerank` endpoint added here exposes that same primitive as a Cohere/vLLM-style API for RAG clients, rather than only an external inference endpoint.

**Problem.** Semantic Router can embed and route queries, but it has no built-in model-based way to *rerank* candidate documents for a query. In a RAG pipeline the first-stage retriever (bi-encoder / vector search) ranks by topical overlap, so a distractor that shares words with the query can outrank the passage that actually answers it. The standard fix is a second-stage cross-encoder reranker, which today means standing up a *separate* service (Cohere, vLLM, a `sentence-transformers` / TEI server) just for that step.

**This PR** adds the first version of that reranking capability with two interchangeable backends, so rerank can become a first-class scoring signal inside Semantic Router while also serving external RAG clients through `/v1/rerank`.

## Alignment With Review Feedback

This PR started as a `/v1/rerank` endpoint plus a cross-encoder backend. Based on review feedback, the intended direction is:

- Cross-encoder loading now runs in the existing `modelruntime` lifecycle instead of the API server startup path (done in this PR).
- Keep the full public `model_catalog.rerank` config contract, docs, and reference-config updates for a follow-up PR.
- Treat rerank as a signal-layer scoring primitive, not only an external inference endpoint.
- Support two-stage matching: bi-encoder shortlist first, then optional cross-encoder rerank over top-K.
- Keep `/v1/rerank` as a compatibility surface over that engine for RAG clients and for the broader RAG orchestration work.

> **Scope (stated up front so the boundaries are clear):** this is one feature (a rerank endpoint with two backends), kept to a reviewable scope:
> - **BERT-family cross-encoder first.** The cross-encoder loader reuses the repo's existing candle `BertModel` stack, so the new model code is minimal (text-pair input + pooler + 1-logit head). Non-BERT rerankers (XLM-RoBERTa/bge, ModernBERT) are a deliberate follow-up behind architecture dispatch, not a hidden gap.
> - **Model selection by name.** A request selects the cross-encoder by naming the served model (or the `"cross-encoder"` alias); the response echoes the served name. Any other `model` uses the bi-encoder backend.
> - **Loading runs on the `modelruntime` lifecycle path**, but the model path/name are still sourced from env vars as an interim hook; the larger public `model_catalog.rerank` schema/docs land separately.
>
> Each constraint is a scoping choice with a stated follow-up (see *Follow-up PRs* below), not undocumented behavior.

**What this PR changes**

- New endpoint `POST /v1/rerank` (+ native alias `POST /api/v1/rerank`) with a Cohere/vLLM-inspired request/response shape (`query`, `documents`, `top_n`, `return_documents` -> `results[].{index, relevance_score, document}`). Note: this is the common rerank field shape, not a byte-for-byte Cohere/vLLM response; it does not (yet) emit envelope fields like `id` or `usage`.
- **Bi-encoder backend (Go only, no new ML):** the handler maps the request onto the existing `candle_binding.CalculateSimilarityBatch` (the same bi-encoder behind `/api/v1/similarity/batch`) and reshapes the matches into the rerank schema. Gives a usable rerank API with zero new ML; scores are bi-encoder cosine (documented). This is the default backend.
- **Cross-encoder backend (true cross-encoder, BERT-family):** a new `BertCrossEncoder` in `candle-binding` that jointly encodes each `(query, document)` pair and emits a sigmoid relevance score (pipeline diagram below). Pairs are scored in **batched** forward passes, and inputs are **truncated to the model's `max_position_embeddings`** so long RAG documents can't overflow BERT's position limit. Reusing the repo's existing candle `BertModel` + tokenizer machinery keeps the new model code to just the text-pair input, pooler, and scoring head, which is why BERT-family is the first supported architecture.
- **Model selection:** a request reaches the cross-encoder by naming it, via either the configured served model id (set via `SR_CROSS_ENCODER_MODEL_NAME`, e.g. `cross-encoder/ms-marco-MiniLM-L-6-v2`) or the convenience alias `"cross-encoder"`. The response echoes the **actual served model name**, so `model` names the served reranker rather than being a magic string. Any other `model` value uses the bi-encoder backend. (Model loading runs on the `modelruntime` lifecycle path; the served-name mapping and model path are still resolved from env vars as an interim hook, with the public `model_catalog.rerank` contract as follow-up work.)

**Design / request flow**

A single endpoint dispatches to one of two backends based on the requested `model`. The bi-encoder backend (default) reuses the existing similarity model; the cross-encoder backend handles requests that name it, when that model is loaded.

```mermaid
flowchart TD
    Client["RAG client / retriever"] -->|"POST /v1/rerank<br/>{query, documents, top_n}"| H["handleRerank"]
    H --> P["parse + validate + defaults<br/>(top_n, dimension)"]
    P --> Dispatch{"model targets<br/>the cross-encoder?"}
    Dispatch -->|"'cross-encoder' alias<br/>or configured served id"| CE["handleRerankCrossEncoder"]
    Dispatch -->|"any other model (default)"| BE["Bi-encoder backend<br/>CalculateSimilarityBatch (existing similarity binding)"]
    CE --> Init{"cross-encoder<br/>initialized?"}
    Init -->|no| Err["503 RERANK_MODEL_UNAVAILABLE"]
    Init -->|yes| RR["Cross-encoder backend<br/>BertCrossEncoder.rerank() via Rust FFI"]
    BE --> Resp["RerankResponse<br/>results[] sorted by score (desc)"]
    RR --> Resp
```

The cross-encoder backend scores each `(query, document)` pair jointly (batched up to 32 pairs per forward), which is what gives it sharper relevance separation than a bi-encoder:

```mermaid
flowchart LR
    Q["query"] --> Tok["tokenize pair<br/>[CLS] query [SEP] document [SEP]<br/>+ token_type_ids<br/>truncate to max_position_embeddings"]
    D["document"] --> Tok
    Tok --> Bert["BERT encoder"]
    Bert --> Cls["take [CLS] hidden state"]
    Cls --> Pool["pooler: dense + tanh"]
    Pool --> Clf["classifier head -> logit(s)"]
    Clf --> Score["sigmoid (1 label) /<br/>softmax positive class<br/>-> relevance in [0, 1]"]
```

**Why this change is needed**

- **Ecosystem compatibility.** Clients (LangChain/LlamaIndex retrievers, etc.) already speak the Cohere/vLLM-style rerank shape; this lets Semantic Router slot into existing RAG stacks as the rerank step, one service instead of a separate reranker deployment. (Field-shape compatibility, not a certified drop-in for the full Cohere/vLLM response envelope.)
- **Architecture fit.** The Semantic Router docs already describe cross-encoder reranking as part of the encoder-intelligence layer; this PR turns that into a concrete scoring primitive plus a widely-adopted `/v1/rerank` request/response surface. The intended router integration is two-stage matching: bi-encoder shortlist first, then optional cross-encoder rerank over top-K candidates so downstream decisions can consume a sharper relevance score or margin.

**Modules affected:** `Router` (apiserver endpoint + `modelruntime` cross-encoder load) and `Bindings` (candle-binding Rust model + FFI + Go wrappers).

**Files**

- Bindings (Rust): `candle-binding/src/model_architectures/traditional/cross_encoder.rs` (new), `candle-binding/src/ffi/rerank.rs` (new), `candle-binding/src/ffi/types.rs`, `candle-binding/src/ffi/mod.rs`, `candle-binding/src/model_architectures/traditional/mod.rs`
- Bindings (Go): `candle-binding/semantic-router.go`, `candle-binding/semantic-router_mock.go`, `candle-binding/cross_encoder_live_test.go` (new)
- Bindings (verification): `candle-binding/tools/reranker-parity/verify_parity.py` (new, transformers oracle), `candle-binding/testdata/reranker_parity_golden.json` (new, committed ground-truth scores)
- Router: `src/semantic-router/pkg/apiserver/config.go`, `.../route_rerank.go`, `.../routes.go`, `.../route_rerank_test.go`, `.../server.go` (resolves the served reranker name; model loading now runs in `modelruntime`), `.../rerank_live_integration_test.go` (new, env-gated live test)
- Router (lifecycle): `src/semantic-router/pkg/modelruntime/router_runtime.go` + `.../cross_encoder_task_test.go` (best-effort cross-encoder load task from `SR_CROSS_ENCODER_MODEL_PATH`)

## Test Plan

All CPU-only; no GPU required.

1. **Build / static gates**
   - `make rust-ci` (or `cd candle-binding && cargo build --release --no-default-features`)
   - `cargo fmt` + `cargo check --no-default-features` (candle-binding)
   - `make vet` (go vet, both modules) and `make go-lint` (golangci-lint v2 on `src/semantic-router`)
2. **Handler + lifecycle unit tests (Router):** `cd src/semantic-router && go test ./pkg/apiserver/ -run 'Rerank|CrossEncoder' -v` (defaults, validation, `top_n` normalization, ordering, invalid index, `return_documents`, the cross-encoder result builder, model dispatch `alias`/served-name routing, and served-name env resolution `TestConfigureCrossEncoderServedNameFromEnv`) plus `go test ./pkg/modelruntime/ -run CrossEncoder -v` (the `router.rerank.cross_encoder` lifecycle task is built when configured and skipped when not).
3. **Cross-encoder binding tests (Bindings), real model, CPU:** download a BERT cross-encoder (e.g. `cross-encoder/ms-marco-MiniLM-L-6-v2`), then `CROSS_ENCODER_MODEL_PATH=<dir> go test -run TestCrossEncoderRerank -v ./candle-binding` (env-gated; skipped when the var is unset). Covers ranking accuracy (`TestCrossEncoderRerankLive`), **transformers parity** (`TestCrossEncoderRerankMatchesTransformersGolden`, asserts each candle score is within 0.02 of the committed HuggingFace `transformers` golden in `testdata/reranker_parity_golden.json`), and **long-input truncation** (`TestCrossEncoderRerankLongInputTruncates`, feeds ~4k-token pairs and asserts no inference error + finite scores). The golden file is regenerated with `tools/reranker-parity/verify_parity.py` (needs torch/transformers; not run in CI).
4. **Live router-path integration (Router), real model, CPU:** `CROSS_ENCODER_MODEL_PATH=<dir> go test ./pkg/apiserver/ -run TestRerankLiveThroughRuntimeAndAPI -v` (env-gated; skipped when unset). Loads the cross-encoder through the real `modelruntime.PrepareRouterRuntime` lifecycle (not API-server-local init) and drives the real `/v1/rerank` route over HTTP, asserting the documented startup behavior (`503 RERANK_MODEL_UNAVAILABLE` before load, `200` with ranked results after) and that the served model id is echoed.
5. **End-to-end (Router + Bindings):** run the router with `SR_CROSS_ENCODER_MODEL_PATH=<dir>` (optionally `SR_CROSS_ENCODER_MODEL_NAME=<id>`) and `curl POST /v1/rerank` naming the served model (or the `"cross-encoder"` alias); compare to `model:"qwen3"` (bi-encoder) and to `/api/v1/similarity/batch`.

Why sufficient: the binding tests exercise the exact FFI path the endpoint uses with a real model; `TestCrossEncoderRerankMatchesTransformersGolden` asserts numeric parity with HuggingFace `transformers` against a committed golden file (so "matches transformers" is reproducible, not a local-only claim); the e2e curl confirms the shape, alias, `top_n`, and validation behavior in the running binary; unit tests cover the Go handler edges.

## Test Result

- **Handler + lifecycle unit tests:** pass (rerank suite incl. defaults/validation/ordering, model-dispatch alias + configured served-name routing, served-name echo, served-name env resolution, and the `router.rerank.cross_encoder` lifecycle task build/skip in `modelruntime`).
- **Long-input truncation:** `TestCrossEncoderRerankLongInputTruncates` passes: ~4k-token query/document pairs are truncated to 512 and return finite scores (correctly ranking the short relevant doc above the long filler) instead of failing at inference.
- **Cross-encoder correctness (5 RAG cases):** top-1 **5/5**, avg answer rank **1.00** with `ms-marco-MiniLM-L-6-v2` (committed `TestCrossEncoderRerankLive`).
- **Transformers parity:** `TestCrossEncoderRerankMatchesTransformersGolden` passes: every candle per-document score is within **0.02** of the committed HuggingFace `transformers` golden (in practice they agree to ~4 decimals, e.g. answer `0.9998`, refund case `0.8590` vs distractor `0.0858`), confirming the candle loader/pooler/head are numerically faithful, not just correctly ordered.
- **Multi-model (loader robustness across hidden size / depth):**

  | model | arch | hidden | layers | candle top-1 | transformers |
  |-------|------|--------|--------|--------------|--------------|
  | ms-marco-MiniLM-L-6-v2 | BERT | 384 | 6 | 5/5 | 5/5 |
  | ms-marco-MiniLM-L-12-v2 | BERT | 384 | 12 | 5/5 | n/a |
  | ms-marco-TinyBERT-L-2-v2 | BERT | 128 | 2 | 3/5 | 3/5 (same 2 misses) |

  TinyBERT's 3/5 is the genuine quality of a 2-layer model (its own `transformers` output misses the same cases), confirming the loader is correct, not lossy. Batched scoring is byte-identical to per-pair scoring.

- **End-to-end live `/v1/rerank`** (CPU, 4 docs, refund-for-cancelled-flight case):

  | rank | doc | cross-encoder | qwen3 bi-encoder |
  |------|-----|---------------|-------------------|
  | 1 | answer | **0.8590** | 0.6881 |
  | 2 | distractor | 0.0858 | 0.6068 |

  Same #1, but the cross-encoder separates answer-vs-distractor by a wide margin (0.86 vs 0.09) where the bi-encoder is tight (0.69 vs 0.61); that is the calibration benefit a RAG keep/drop threshold needs. `top_n`, the `/api/v1/rerank` alias, and 400 validation all behave; bi-encoder scores match `/api/v1/similarity/batch` exactly.

- **Live router-path integration:** `TestRerankLiveThroughRuntimeAndAPI` passes: the cross-encoder is loaded through the real `modelruntime` lifecycle, then `POST /v1/rerank` returns `503 RERANK_MODEL_UNAVAILABLE` before the model is ready and `200` after, echoing the served model id and ranking the true answer (doc 0) at #1 (`0.8590` > `0.0858` > `~0.00`). Confirms the lifecycle move and served-name dispatch end to end over HTTP, not just at the FFI layer.

- **Latency:** endpoint-only use is opt-in, so existing routing traffic does not pay rerank inference cost. For signal-routing integration, rerank should run only as an optional second stage over the top-K bi-encoder shortlist, which bounds the `O(N_docs)` cross-encoder cost. Measured CPU (4 short docs): MiniLM-L6 cross-encoder ~97-123 ms vs Qwen3-0.6B bi-encoder ~745-877 ms. Recommended usage is two-stage retrieve-then-rerank (shortlist with the bi-encoder, cross-encode the top-K).
- **Static gates:** `cargo fmt` (no churn) + `cargo check` pass; `cargo clippy` clean on the changed files (the rerank C structs were moved into `ffi/rerank.rs` so `ffi/types.rs` is unchanged and its pre-existing clippy debt stays out of the changed-file lint); `gofmt`/`go vet` clean; `golangci-lint v2` clean (0 issues) on the changed `apiserver` + `modelruntime` packages; trailing-whitespace / final-newline / `codespell` clean.

**Known behavior and limitations (this PR):**

- **Single cross-encoder instance.** One global `OnceLock<Arc<BertCrossEncoder>>`. The API contract (`model` names the reranker, response echoes it) is already multi-model-ready, so a name->model registry (+ optional lazy-load/LRU) is a follow-up with no API change.
- **Unknown `model` -> bi-encoder.** Same fallthrough as `/api/v1/similarity/batch`: an unconfigured reranker name gets bi-encoder scores instead of a 4xx. Happy to add a strict unknown-model error here if maintainers prefer.
- **Named cross-encoder, not yet loaded -> 503.** Requesting the cross-encoder (the `"cross-encoder"` alias or the configured served name) before it has loaded, or when its load failed, returns `503 RERANK_MODEL_UNAVAILABLE` rather than falling back to bi-encoder scores under a cross-encoder name. Because loading runs on the model-runtime path while the API server starts early (for `/startup-status`), there is a brief startup window where this 503 applies; this matches how the rest of the router brings models up asynchronously.
- **Pre-existing similarity limitation (not introduced here).** `model=auto`/`mmbert` hit an existing `CalculateSimilarityBatch` limitation (`model=qwen3` works).

**Follow-up PRs:**

- **`model_catalog.rerank` config contract** (after this merges): declare rerankers in config like other models + canonical/reference config, docs, and contract tests; removes the `INTERIM` env loading (`SR_CROSS_ENCODER_MODEL_PATH` / `SR_CROSS_ENCODER_MODEL_NAME`).
- **Signal-routing integration** (after this merges, design agreed in #2247): optional second stage (bi-encoder shortlist -> cross-encoder rerank top-K) feeding the decision engine.
- **Non-BERT reranker support** (after this merges): XLM-RoBERTa/bge + ModernBERT behind architecture dispatch; no API change. The current `BertForSequenceClassification` boundary is documented on the `BertCrossEncoder` type.
- **BERT pooler fix** (independent, pre-existing): correct the `traditional/bert.rs` pooler weight orientation; it changes existing classifier confidence, so it lands as its own PR.

## Checklist

- [x] Code builds (`cargo` CPU + `go build`) and is formatted (`cargo fmt`, `gofmt`).
- [x] Lints pass for changed files (`go vet`, `golangci-lint`, `codespell`).
- [x] New unit tests + an env-gated real-model binding test added.
- [x] End-to-end verified against a locally-built router.
- [ ] Maintainer guidance requested on: endpoint paths, accepting the bi-encoder backend as an interim default, and the preferred default cross-encoder checkpoint (see #2247).


src/semantic-router/pkg/apiserver/server.go
```diff
@@ -105,6 +114,47 @@ func InitWithRuntime(configPath string, port int, runtimeRegistry *routerruntime
...
+// INTERIM: env-based loading is a deliberate stopgap. The intended end state is
+// declaring rerankers in the model catalog/config (alongside other models) and
+// loading them through the same path; this env hook keeps the Phase 2 backend
+// usable until that config plumbing lands. See the PR/issue for the plan.
+func initCrossEncoderFromEnv() {
+	// Clear any prior served-name mapping first so a no-op/failed (re)init can never
+	// leave dispatch pointing at a previously loaded model.
+	crossEncoderServedName = ""
+
+	modelPath := strings.TrimSpace(os.Getenv("SR_CROSS_ENCODER_MODEL_PATH"))
```
Xunzhuo: Looks like it is not aligned with the current model lifecycle management. can you check it?

AayushSaini101:
Thanks for the feature request @arun-elr can you please first create an issue and link to the PR and also the lint CI is failing, reference, you can test this via locally and try to fix this thanks
```
 go test ./pkg/config/... -run TestReferenceConfig -count=1 (cwd=/home/runner/work/semantic-router/semantic-router/src/semantic-router)
Running Rust lint... 
```

Xunzhuo (commented):
one more concern, it treats vllm-sr as an external inference server, this is acceptable but we should explore how to adopt it and align with the signal decision routing, any ideas for the alignment? maybe for embeddings related signals, maybe we can explore the optional 2 stage bi-encoder first then cross-encoder matching to improve the accuracy?

src/semantic-router/pkg/apiserver/server.go
arun-elr: Good catch, thanks. Fixed: cross-encoder loading no longer happens in the API server startup path. It now runs on the existing model-runtime lifecycle.

- Added a best-effort task `router.rerank.cross_encoder` in `modelruntime.PrepareRouterRuntime`, alongside the other model loads, so the reranker is initialized on the same runtime lifecycle path and reports failures through the existing lifecycle event logging.
- `server.go` no longer loads the model. It only resolves the public served-name used to dispatch `/v1/rerank` requests (`configureCrossEncoderServedName`).

One thing I kept interim to keep this PR reviewable: the model path/name still come from env vars (`SR_CROSS_ENCODER_MODEL_PATH` / `SR_CROSS_ENCODER_MODEL_NAME`), marked `INTERIM` in code. Declaring rerankers in the public `model_catalog` schema is a larger change (config contract + canonical/reference config + docs + contract tests, per `pkg/config/AGENTS.md`), so I'd prefer to land that as a focused follow-up rather than bundle it here.

One behavior consequence worth flagging: because loading is now async on the model-runtime path (and the API server starts early for `/startup-status`), a `/v1/rerank` request that names the cross-encoder before it has finished loading returns `503 RERANK_MODEL_UNAVAILABLE` rather than silently serving bi-encoder scores under a cross-encoder name. That matches how the rest of the router brings models up and is consistent with the existing `"cross-encoder"` alias path.

For the broader "how does this align with routing" question, I opened #2247 with the architecture: bi-encoder shortlist -> cross-encoder rerank over top-K -> relevance score/margin consumed by routing / RAG / cache / memory. I'm keeping the direct signal-routing integration out of this PR until we agree on the layer/API shape in #2247.

arun-elr:
> Thanks for the feature request @arun-elr can you please first create an issue and link to the PR and also the lint CI is failing, reference, you can test this via locally and try to fix this thanks
> 
> ```
>  go test ./pkg/config/... -run TestReferenceConfig -count=1 (cwd=/home/runner/work/semantic-router/semantic-router/src/semantic-router)
> Running Rust lint... 
> ```

Thanks, both addressed:

- Issue: opened #2247 (problem statement + architecture proposal) and linked it; the PR references it with `Part of #2247` (not `Closes`, since #2247 also tracks the `model_catalog` config and signal-routing follow-ups).
- Rust lint CI: fixed. Root cause was not new code. This PR added the rerank C structs to `candle-binding/src/ffi/types.rs`, which put `types.rs` into the changed-file set; the changed-file clippy gate then surfaced two pre-existing warnings already living in `types.rs` (`unsafe fn missing a # Safety section` and a manual `is_multiple_of`) that are unrelated to this change. I moved the rerank structs into `ffi/rerank.rs` (where they're used), so `types.rs` returns to its original contents and drops out of the changed set. Clippy on the actually-changed files is clean.
- For reference: `go test ./pkg/config/... -run TestReferenceConfig` passes locally; the failing step in that log was the Rust lint that runs right after it, not the Go test.

arun-elr:
> one more concern, it treats vllm-sr as an external inference server, this is acceptable but we should explore how to adopt it and align with the signal decision routing, any ideas for the alignment? maybe for embeddings related signals, maybe we can explore the optional 2 stage bi-encoder first then cross-encoder matching to improve the accuracy?

Agree, and that is the direction I would like to take it. The `/v1/rerank` endpoint is just the external API surface; the cross-encoder underneath should become an internal scoring component that signal decision routing can consume, not a separate server. That is the two-stage design I proposed in #2247:

**Stage 1 (recall, cheap):** reuse the existing bi-encoder / embedding similarity path to produce a top-K candidate set.

**Stage 2 (precision, optional):** cross-encode the query against only those top-K candidates. Because this runs on top-K rather than the full candidate/doc set, the added latency is bounded and config-gated.

The cross-encoder relevance score, and possibly the score margin between top candidates, can become a signal the decision engine uses: routing confidence, semantic-cache hit/miss gating, or KB/RAG reranking. So it feeds the existing signal decision path instead of living only as a side endpoint.

One question back on the layer, since it drives the config surface: should rerank live inside the embedding similarity signal itself, as a refinement step before it emits a score, or as a separate rerank stage that the decision engine composes over embedding-backed signals? That choice shapes whether `model_catalog.rerank` is wired into the similarity signal or exposed as its own stage.

For this PR, I'd keep the standalone endpoint plus the model-runtime-managed cross-encoder engine, and land signal-routing integration as a dedicated follow-up once we agree on the layer/config shape. I'm happy to prototype the two-stage path behind a flag so we can measure the accuracy lift on a routing/eval set.

Xunzhuo (commented):
This contains substantial and useful cross-encoder work, especially the parity fixture and real-model lifecycle test. The blocker is that the PR currently combines a valuable internal primitive with an interim public API/config contract.

The smallest mergeable path I recommend is:

1. Keep the Candle cross-encoder implementation, FFI, model-runtime lifecycle integration, parity fixture, and focused tests.
2. Move the public `/v1/rerank` endpoint and env-var configuration to a follow-up that uses the canonical `model_catalog`/CLI/docs/E2E contract.

If you prefer to keep the endpoint in this PR, please first complete all of the following:

- Replace `SR_CROSS_ENCODER_*` as the runtime contract with canonical config plus Python CLI schema/validation and user docs.
- Move `crossEncoderServedName` out of package-global mutable state into server/runtime-owned state.
- Reject an unknown requested `model` instead of silently returning bi-encoder scores under different semantics.
- Add request bounds (at least document count/aggregate input) and a repository-registered E2E profile; env-gated Go tests and manual curl are not the durable API acceptance gate.
- Resolve the architecture choice in #2247: internal signal/refinement stage versus standalone API surface.
- Fix the linked-issue gate without over-closing the umbrella issue. The clean option is a scoped issue that this PR fully closes, while #2247 remains the umbrella.

Please choose the split or full-contract route explicitly in a reply; once that boundary is agreed, the remaining review can focus on the implementation rather than repeatedly revisiting scope.

WUKUNTAI-0211:
Committer note for whoever carries this forward: this is not a blocker on the
relevance-rerank scope of this PR. It is a capability boundary for #2247 and
the semantic-cache follow-up in #2691.

The cross-encoder path here exposes a scalar relevance score. That is useful
for ranking, but it is not by itself a verified polarity/contradiction signal
for a semantic-cache hit: cache reuse must distinguish `enable` / `turn on`
from `enable` / `disable`, not merely rank topical overlap.

NevIR motivates measuring this explicitly, rather than assuming a
relevance-trained checkpoint handles it. Its published paired-ranking results
top out at 50.6% for MS MARCO MonoT5-3B, but the paper does not evaluate this
PR's MiniLM checkpoint or BGE. Therefore it is evidence to add a cache-negation
evaluation, not a model-specific verdict on this implementation.

If the semantic-cache verifier is added later, model the NLI-style
three-label relation (`contradiction`, `entailment`, `neutral`) as a distinct,
label-aware pair-scoring capability. Its query formulation, directional policy,
thresholds, and false-miss rate need cache-specific validation. #2247 should
own the base typed contract; #2252 should make the resulting capability
architecture-dispatchable. The relevance-ranking scope of this PR is
unaffected.

**Terminology update (2026-08-05):** After the task taxonomy proposed in #2779
and the scoped relevance primitive in #2783, pair-shaped NLI input should not be
described as a `pair_score` task. NLI belongs to `sequence_classify` /
`label_distribution`, while #2783 is limited to scalar cross-encoder relevance
scoring. The separation conclusion above remains unchanged.

WUKUNTAI-0211:
I opened #2783 as the scoped child issue under #2247 for the internal
cross-encoder primitive described in the July 14 review.

The intended scope is limited to the existing `BertCrossEncoder`, Candle FFI,
typed internal scoring contract, model-runtime lifecycle/readiness, bounds, the
parity fixture, real-model lifecycle test, and focused tests. I do not plan to
carry over the `/v1/rerank` endpoint, the bi-encoder fallback, served-name
globals, or the `SR_CROSS_ENCODER_*` configuration.

The implementation in this PR already contains substantial reusable work. Before
opening a fresh implementation PR, would you prefer to split those pieces from
#2236, or are you comfortable with me porting them from this branch while
preserving attribution?

Xunzhuo:
Thanks for pushing this forward. After the Router Models tree cleanup, please rebase on current `main` and narrow this PR to #2783: the internal typed BERT cross-encoder scoring primitive. The current diff also adds a public `/v1/rerank` endpoint, interim environment-based loading, and bi-encoder fallback; those should not land as part of that primitive. Catalog/config work belongs in #2250, additional architectures in #2252, and recipe-level invocation in #2251. Once the split and rebase are ready, we can review the smaller contract and its parity tests directly.

FAUST-BENCHOU (commented):
Thanks for adding first-class rerank support to Semantic Router. Turning reranking into a reusable scoring primitive is a valuable direction, and the separation between the endpoint, runtime lifecycle, and backend implementation looks reasonable.

PR #2248, opened 2026-06-18
[Router] Invalidate memory cache via per-user index set instead of ke…

### LINKED ISSUE #2277: [Router] Memory cache invalidation uses O(keyspace) SCAN — switch to a per-user index set
# Summary

`RedisCache.InvalidateByUser` in `redis_cache.go` currently locates keys to delete by performing a Redis `SCAN` over the entire keyspace using the pattern `{prefix}{userID}:*`. As a result, the cost of invalidation scales with the total number of keys in Redis rather than the number of cache entries belonging to the target user.

Since cache invalidation is triggered on every store, update, or forget operation for a user, this places an **O(keyspace)** operation on a hot path. On Redis instances shared across multiple workloads—or simply containing a large number of cached entries—each invalidation traverses significantly more keys than it ultimately deletes, causing latency and load to increase as the deployment grows.

## Current Behavior

```go
pattern := c.prefix + userID + ":*"
iter := c.client.Scan(ctx, 0, pattern, 100).Iterator()
// collect matching keys, then DEL them
```

* Performs a Redis `SCAN` over the entire keyspace using a pattern match.
* Work is proportional to the total number of keys in Redis, not the number of cached entries for the user.
* Every invalidation incurs an **O(keyspace)** operation.

## Proposed Change

Replace keyspace scanning with a per-user index set that tracks all cache entries for a user.

### Key Structure

* **Value keys:** `{prefix}v:{userID}:{queryHash}`
* **Per-user index set:** `{prefix}u:{userID}`

The `v:` (value) and `u:` (index) namespaces are intentionally disjoint, preventing collisions regardless of the contents of `userID`.

### Cache Write (`Set`)

When storing a user-scoped cache entry:

1. Store the value at `{prefix}v:{userID}:{queryHash}`.
2. Add the value key to the user's index set using `SADD`.
3. Refresh the index set's TTL to match the value TTL so the index always outlives the entries it tracks.

### Cache Invalidation (`InvalidateByUser`)

1. Read all tracked value keys using `SMEMBERS` on the user's index set.
2. Issue a single `DEL` containing:

   * All tracked value keys
   * The index set itself

If some value keys have already expired via TTL, `DEL` is a harmless no-op, so stale members remaining in the index set do not cause errors.

## Benefits

* Eliminates keyspace-wide `SCAN` operations.
* Reduces invalidation complexity from **O(keyspace)** to **O(user's cached queries)**.
* Keeps invalidation cost proportional to the amount of data actually owned by the user.
* Improves scalability and reduces latency as the Redis keyspace grows.

## Affected Files

* `src/semantic-router/pkg/memory/redis_cache.go`
* `src/semantic-router/pkg/memory/redis_cache_test.go`

gagandhakrey:
Closes #2277

[Router] Invalidate memory cache via per-user index set instead of keyspace SCAN

## Problem

`RedisCache.InvalidateByUser` ran `SCAN ... MATCH {prefix}{userID}:*` on every memory
write. Redis applies `MATCH` *after* fetching keys, so the scan walks the **entire
keyspace** regardless of how few keys belong to the user — cost is O(total keys in the
Redis DB), not O(the user's keys).

This invalidation fires on the production write path whenever the Redis hot cache is
enabled: memory extraction on conversations (`extractor.go`), consolidation
(`consolidation.go`), and the memory API (`apiserver/route_memory.go`), all routed
through `CachingStore` → `InvalidateByUser`.

## Impact (before)

- Each memory write triggers a full-keyspace scan. With keyspace size `N` and write
  rate `W`, server-side work ≈ O(N·W).
- On a dedicated, small cache the effect is negligible (why it hasn't visibly broken).
- On a **shared or large** Redis instance (10⁵–10⁷ keys), every extracted memory drives
  a multi-pass scan that competes with serving traffic — added write-path latency and
  Redis CPU pressure. The blast radius grows with *other* tenants' data, so unrelated
  keyspace growth silently slows memory invalidation.

## Fix

Track each user's cache keys in a per-user Redis set and invalidate by deleting that
set's members — no scan:

- `cacheKey` namespaces value keys as `{prefix}v:{userID}:{hash}`.
- New `userIndexKey` returns `{prefix}u:{userID}`, the set of that user's value keys.
- `Set` pipelines `SET` + `SADD` + `EXPIRE` (one round-trip); the index set's TTL is
  refreshed on every `Set`, so it always outlives every live key it tracks.
- `InvalidateByUser` does `SMEMBERS` + a single `DEL` of the members plus the set.

Safety properties verified:
- **No key collisions:** the `v:` (value) and `u:` (index) namespaces are disjoint for
  any `userID`, so an index set can never collide with a cached value (avoids WRONGTYPE).
- **No stale reads from a missing index entry:** value key and index set share the same
  TTL set at the same instant, so the set never expires before a key it tracks.
- **Stale set members are harmless:** `DEL` on an already-expired key is a no-op.
- **No new race:** the populate-vs-invalidate race is identical to the prior `SCAN`
  behavior and remains bounded by TTL.

## Impact (after)

- Invalidation cost drops to **O(user's cached queries)** (typically single digits),
  independent of total keyspace size — effectively constant-time per write.
- Eliminates Redis CPU contention from repeated full scans; write-path latency is flat
  regardless of how large or shared the Redis instance is.
- Added cost is one `SADD`+`EXPIRE` per cache `Set` (pipelined, ~free) and small bounded
  memory for the index sets (TTL-expiring).
- Invalidation semantics are unchanged — same keys cleared, located directly instead of
  by scanning.

**Rollout note:** value keys written by the previous key format aren't tracked in any
index, so on first deploy they persist until TTL (≤5 min default) rather than being
invalidated on the next write. Self-heals within one TTL window.

## Test Plan

- `go build ./pkg/memory/... ./pkg/extproc/...`, `go vet`, and `gofmt` — clean.
- Existing memory unit tests pass; `cacheKey` substring assertions still hold under the
  new namespace.
- Added integration tests (run against the CI Redis service via `make start-redis`):
  - `TestRedisCache_InvalidateByUser_DeletesTrackedKeys` — `Set` registers keys in the
    index set; `InvalidateByUser` deletes all tracked keys and the set itself.
  - `TestRedisCache_InvalidateByUser_ScopedToUser` — invalidating one user leaves another
    user's cached entries intact.


AayushSaini101:
Thanks for the fix @gagandhakrey can you please create an issue for this, and link to this PR, we will start reviewing that thanks 🙏

gagandhakrey:
> Thanks for the fix @gagandhakrey can you please create an issue for this, and link to this PR, we will start reviewing that thanks 🙏

@AayushSaini101  created and linked the issue for the same

AayushSaini101 (approved):
LGTM thanks for the optimisation

PR #2261, opened 2026-06-22
[Router] Validate memory similarity_threshold range

theohsiung:
## What

Follow-up to #2143 (per @Xunzhuo's review: _"can you check if we need to harden the validator somewhere else?"_).

The long-term-memory `similarity_threshold` is a cosine similarity and must be in `[0.0, 1.0]`, but it is **never validated** — the structural twin of the gap #2143 just fixed for the semantic cache. An out-of-range value is silently accepted at load and reaches vector-store retrieval unchanged. A threshold `> 1.0` can never be reached by any cosine similarity, so **memory retrieval never matches and long-term memory is silently disabled despite `enabled: true`**.

## Reproduction

```yaml
# global memory block
memory:
  enabled: true
  default_similarity_threshold: 1.5   # accepted at load; memory then never retrieves
```
or per-decision:
```yaml
decisions:
  - name: route
    plugins:
      - type: memory
        configuration:
          enabled: true
          similarity_threshold: 2.0   # accepted; this decision never retrieves memory
```

## Root cause

- `MemoryConfig.DefaultSimilarityThreshold` (`runtime_config.go`) and the per-decision `MemoryPluginConfig.SimilarityThreshold` (`plugin_config.go`) are parsed with no range check.
- The vector-store backends (`memory/milvus_store_retrieve.go`, `qdrant_store.go`, `valkey_store.go`) fall back to / compare against the configured value without clamping (`bestSimilarity >= threshold`, cosine ≤ 1.0), so `> 1.0` → always miss.

## Fix

Add a `memory` contract validator (wired into `sharedConfigContractValidators`, right after `semantic_cache`) that rejects an out-of-range `similarity_threshold` on:
- the global `memory` block (`cfg.Memory.DefaultSimilarityThreshold`), and
- each decision's `memory` plugin (`decision.GetMemoryConfig().SimilarityThreshold`).

This mirrors the bound the semantic cache and RAG plugins already enforce (`validateSemanticCacheContracts` / `validateRAGSimilarityThreshold`). The global field is a plain `float32` whose zero value means "unset" (runtime falls back to a default); `0` is within range, so unset configs pass without special-casing. An unset per-decision pointer stays valid.

## Test plan

- [x] `validator_memory_test.go`: global threshold range table (unset/0/0.74/1.0 ok; 1.5/-0.1 rejected); per-decision out-of-range rejected and error names the decision; nil config ok. RED before fix → GREEN.
- [x] Full `pkg/config` suite green — no shipped config trips the new check (all repo `similarity_threshold` values are within `[0,1]`, range 0.2–0.95).
- [x] `gofmt`/`go vet` clean; `golangci-lint` (repo config, `tools/linter/go/.golangci.yml`) → 0 issues.

## Remaining candidates (deferred, noted for visibility)

Other cosine-`similarity_threshold` / confidence-`threshold` fields that still lack the same `[0,1]` guard and could be a further follow-up in this validator-hardening family (#2115 / #2123 / #2143): `PrototypeScoringConfig.ClusterSimilarityThreshold`, `ToolsConfig.SimilarityThreshold` / `ToolSelectionPluginConfig.SimilarityThreshold`, and the various classifier confidence `threshold` fields. Each needs an individual semantics check (some siblings like `*_weight` / `temperature` are **not** `[0,1]` and must not be range-checked), so they're intentionally out of scope here to keep this PR tightly mirrored on #2143.

## Notes
- DCO signed-off. Same validator-hardening family as #2115 / #2123 / #2143.


theohsiung:
@AayushSaini101 would you like to review this pr, thx~

theohsiung:
Hi @WUKUNTAI-0211 can you also review this one thx!

drivebyer (approved):

PR #2269, opened 2026-06-24
[Router] Grounding-aware Fusion: servability (metrics, early-exit, adaptive escalation) + cached-panel A/B/C/D eval harness

Peterren:
## Summary

Builds out **grounding-aware Fusion** along two axes:

1. **Cached-panel A/B/C/D eval harness** (`bench/grounded_fusion`, `cmd/fusioneval`): generate the panel once per item and synthesize every arm from the byte-identical panel, so deltas isolate the intervention. Arms: `A` judge-solo, `B` plain fusion, `C` `weight` (shipped default), `D` random-weight placebo. A pre-registered KEEP/KILL rule (`compare_multiarm.py`) turns the paired bootstrap CIs into a decision. Reuses the existing `CachedPanel` looper seam.

2. **Servability for fusion traffic** (this stack adds):
   - **Prometheus metrics** (`vsr_fusion_*`): request count/status, end-to-end and per-stage duration, per panel-model success/failure, grounding score distribution and drop count, early-exit and escalation counters, per-request token cost. Fusion previously emitted **zero** metrics.
   - **Panel-agreement early-exit** (`grounding.early_exit_*`, default off): skip the analysis judge call when the panel is unanimous under panel-mode NLI. Requires unanimity, so a dissenter keeps the full pipeline — it never drops the minority view.
   - **Adaptive escalation** (`fusion.escalation`, default off): run the full panel only for queries matching a configured hard-complexity rule (reusing the existing binary complexity signal); answer easy queries with a single judge-model call instead of N+2.

## Honest result

This does **not** demonstrate a quality win for grounding. A small cached-panel smoke (N=4, Medicine/Law; judge `qwen3:14b`, grader `qwen3:32b`) ran the full chain to a verdict: the machinery is correct (panels byte-identical across arms; rule returns `INCONCLUSIVE` as designed at N=4), and every directional signal matches `FINDINGS.md` — solo ≈ placebo > plain fusion ≈ weight. **Do not read science into N=4.** The harness + servability are the shippable contribution; a powered verdict needs ≥20–25 items/domain (overnight) and ideally a grader from a different model family than the judge. Documented in `bench/grounded_fusion/README.md`.

Streaming remains buffered (every looper returns one Envoy ExtProc `ImmediateResponse`); true token streaming is a cross-cutting looper↔ExtProc change, noted as future work in the tutorial.

## Test Plan

- `go test ./pkg/config/ ./pkg/looper/ ./pkg/observability/metrics/` — pass (new unit tests for metric helpers, early-exit/escalation gate logic, and validation bounds).
- Full `golangci-lint` (agent config) on every changed package — 0 issues on changed files (complexity-driven refactor: servability helpers split into `fusion_servability.go`).
- `pre-commit run --files <changed>` — all hooks pass (go fmt/lint, black, ruff agent config, md/yaml fmt, codespell, AST scan, structure_check no errors).
- End-to-end smoke validated locally: `fusioneval --arms A,B,C,D` → `grade_only` → `compare_multiarm` produces `verdict.json` with byte-identical `panel_sha256` across arms.

Defaults are off for both new levers, so behavior is unchanged unless explicitly enabled.


Xunzhuo:
@Peterren can you rebase?

src/semantic-router/cmd/fusioneval/main.go
Xunzhuo, line 1: can we put the eval outside of cmd/, maybe into eval or test package

Xunzhuo (commented):
Thanks for the servability work. Three blockers remain: the documented `early_exit_min_consistency: 0.0` makes every scored panel—including dissenting ones—skip analysis; the single-model path bypasses the decision output-contract normalization used by full Fusion; and `hard_complexity_rules` is not checked against declared complexity rules, so a typo silently sends all traffic down the cheap path. Please fix those cases with end-to-end tests, link an accepted tracking issue, and rebase onto main.

FAUST-BENCHOU (commented):
Thanks for extending Fusion with grounding-aware observability and evaluation tooling. Adding metrics and controllable optimization paths while keeping the new behaviors disabled by default is a good approach.

PR #2272, opened 2026-06-24
[Operator] fix updateStatus dropping reconcile-derived status fields

### LINKED ISSUE #2271: [Operator] updateStatus drops reconcile-derived status fields (GatewayMode, OpenShiftFeatures, ObservedGeneration)
`updateStatus` re-fetches a fresh object (`current`) and only copies `Replicas`, `ReadyReplicas`, `Phase`, and `Conditions` before calling `Status().Update()`. Status fields written during reconciliation — `GatewayMode` (set in `reconcileOwnedResources`) and `OpenShiftFeatures` (set in `reconcileRoute`) — are silently dropped because they are never copied to `current`.

Additionally, `ObservedGeneration` is set to `current.Generation` (the re-fetched object's generation, which may have been bumped by a concurrent spec update) instead of the generation actually reconciled.

drivebyer:
Fix `updateStatus` silently dropping `GatewayMode`, `OpenShiftFeatures`, and using wrong `ObservedGeneration`.

Switch from re-fetch + selective copy + `Status().Update()` to `Status().Patch()` with `client.MergeFrom`, preserving all status fields set during reconciliation.

Fixes https://github.com/vllm-project/semantic-router/issues/2271

### DIFF
--- a/deploy/operator/controllers/semanticrouter_controller.go
+++ b/deploy/operator/controllers/semanticrouter_controller.go
@@ -81,11 +81,13 @@ func (r *SemanticRouterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	baseSR := semanticrouter.DeepCopy()
+
 	if err := r.reconcileOwnedResources(ctx, semanticrouter, logger); err != nil {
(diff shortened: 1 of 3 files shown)


AayushSaini101 (approved):
LGTM thanks : )

PR #2279, opened 2026-06-28
[Feature] Extend RL-driven routing with LLM router prompting

### LINKED ISSUE #2310: [Feature] Extend RL-driven LLM router naming and prompt flexibility
### Goal

Extend the existing RL-driven LLM router support so it better matches the intended routing abstraction and is easier to customize.

  The current RL-driven LLM router path uses Router-R1-style naming, and the prompt is effectively fixed. The data exposed to the prompt is also limited, which makes it hard to adapt the router behavior for
  different scenarios.

  I would like the RL-driven LLM router path to support:
  - clearer naming that reflects the RL-driven routing concept
  - configurable prompt templates instead of a hard-coded prompt
  - more flexible prompt inputs
  - additional request/context fields beyond just decision name and model reference names
  - optional template file loading with relative path resolution from the router config base directory

### Proposed Workgroup

MoM & Routing

### Scope

##### Primary layer

unsure

##### Why this layer?

This change is specifically about the interface of the RL-driven LLM router path.

  It does not fit signal extraction or plugin-style post-processing. It is also not clearly a global router capability, because the scope is limited to the RL-driven routing path and its prompt/config
  contract. For that reason, the ownership is not obvious enough to force a more specific layer classification.

##### Why do you need this feature?

The current implementation is too rigid for routing experimentation and future extension.

  We need the router prompt to be customizable because different routing scenarios may require different request fields, different decision context, or different model metadata. Without template flexibility,
  every prompt change requires code changes, and the router cannot be adapted cleanly to new use cases.

##### Additional context

This change should refine the existing RL-driven routing integration into something more flexible and easier to extend, without changing the underlying routing goal.

### Non-goals

### Completion signal

georgefs:
Closes #2310

  ## Purpose

  - This PR updates the `rl-driven` decision algorithm to render a configurable LLM router query and send it to a router server.
  - The new flow stays compatible with the original Router-R1 contract: `POST /route` with `query`, and a response containing `selected_model`, `thinking`, and `full_response`.
  - It replaces the older `router_r1_server_url` naming with `llm_router_server_url` for the `rl-driven` path.
  - It adds the runtime, test, E2E, and docs updates needed to validate the flow end to end.
  - This affects `Router`, `CLI`, `E2E`, and `Docs`.
 
  
    ### Query template supported parameters
  The rendered query template can use these fields:

  - `decision_name`
  - `decision_description`
  - `query`
  - `request_id`
  - `user_id`
  - `session_id`
  - `category_name`
  - `matched_domains`
  - `matched_keywords`
  - `labels`
  - `tags`
  - `conversation_history`
  - `available_models`
  - `available_model_names`
  - `available_model_details`
  - `metadata`

    ## Example

  ```yaml
  routing:
    decisions:
      - name: llm-router-decision
        description: Decision powered by rl_driven
        algorithm:
          type: rl_driven
          rl_driven:
            storage_path: /var/lib/vsr/rl_state.json
            enable_llm_routing: true
            llm_router_server_url: http://host.docker.internal:9001
            llm_router_query_template: |
              Decision: {{.decision_name}}
              Query: {{.query}}
              Domains: {{join .matched_domains ", "}}
              Labels: {{join .labels ", "}}
 ````

  ## Test Plan

  - `make agent-dev ENV=cpu`
  - `make agent-serve-local ENV=cpu AGENT_STACK_NAME=llm-router-smoke AGENT_SERVE_CONFIG=/home/george/work2/semantic-router/e2e/config/config.llm-router-mock.yaml`
  - `./e2e/testing/run_llm_router_smoke.sh`
  - `make test-semantic-router`

  Why this validation is sufficient:
  - `agent-dev` verifies the local image build path.
  - `agent-serve-local` exercises the repo-native runtime startup and config translation path.
  - The smoke script verifies the deployed router actually sends the rendered query to the mock LLM router server using the Router-R1-compatible contract.
  - `test-semantic-router` covers the Go runtime and selection path affected by the change.

  ## Test Result

  - `make agent-dev ENV=cpu` passed.
  - `make agent-serve-local ENV=cpu AGENT_STACK_NAME=llm-router-smoke AGENT_SERVE_CONFIG=/home/george/work2/semantic-router/e2e/config/config.llm-router-mock.yaml` passed.
  - The mock LLM router server received the rendered query and returned a Router-R1-compatible response.
  - `make vllm-sr-test` still fails on `test_status_when_not_running` in `e2e/testing/vllm-sr-cli/test_unit_lifecycle.py`.

  Follow-up:
  - The remaining `vllm-sr-test` failure is a CLI lifecycle status assertion mismatch and is not caused by the rl-driven LLM router flow itself.
  - Manual smoke assets are included to validate the routing contract end to end.

  ---
  <details>

  </details>


theohsiung:
The direction here makes sense for RL / LLM-router experimentation. My suggestion would be to hold the configurable query template (and its template-variable set) for now, and let its design be driven by concrete training/experimentation requirements once we have them.

Two reasons:

- The template effectively promotes the internal selection-context shape into a *public config contract*. Freezing which fields are exposed (and their names/semantics) before there's a real consumer risks locking in a shape we'll want to change once actual RL training/routing experiments tell us which signals matter.
- Adding these config fields later is purely additive (non-breaking), so there's no cost to waiting — we lose nothing by designing it once the requirements are concrete.

Concretely: could we gather the training-side needs first — which request/context fields the router prompt actually benefits from — and then design the template contract around that? Happy to help spec it once those needs surface.

The de-branding rename (`router_r1_server_url` → `llm_router_server_url`) is a separate call; I'll leave that to maintainer preference for the v0.4 window.

georgefs:
> The direction here makes sense for RL / LLM-router experimentation. My suggestion would be to hold the configurable query template (and its template-variable set) for now, and let its design be driven by concrete training/experimentation requirements once we have them.
> 
> Two reasons:
> 
> * The template effectively promotes the internal selection-context shape into a _public config contract_. Freezing which fields are exposed (and their names/semantics) before there's a real consumer risks locking in a shape we'll want to change once actual RL training/routing experiments tell us which signals matter.
> * Adding these config fields later is purely additive (non-breaking), so there's no cost to waiting — we lose nothing by designing it once the requirements are concrete.
> 
> Concretely: could we gather the training-side needs first — which request/context fields the router prompt actually benefits from — and then design the template contract around that? Happy to help spec it once those needs surface.
> 
> The de-branding rename (`router_r1_server_url` → `llm_router_server_url`) is a separate call; I'll leave that to maintainer preference for the v0.4 window.

Thanks for the detailed feedback! I think we may be looking at this from slightly different perspectives.

My intention with the configurable template isn't to define or freeze the eventual training signals. Instead, it's to define what context is available to a router implementation while keeping the implementation itself agnostic.

For router training, I actually expect the workflow to be somewhat reversed. We often don't know upfront which signals will be useful. Instead, we first define which request/context fields are reasonable for a router to access, collect and accumulate that information over time, and then use it to train or evaluate different router implementations. In that sense, the question is less "which fields will the training require?" and more "which fields should a router be allowed to see?"

The motivation behind this change is also to support different router implementations rather than Router-R1 specifically. Different routers may consume different subsets of the available context, so I was hoping to make that configurable instead of baking a Router-R1-specific prompt into the interface.

That said, I do agree that we should avoid committing to an overly broad or unstable public contract too early. If the current proposal still feels like it exposes too much of the internal context shape, I'm happy to narrow the scope or split the changes so we can iterate on the template contract separately.

Xunzhuo (commented):
Thanks for the thoughtful response. The implementation direction is useful, but the current PR still freezes a broad public template contract before we have an experiment that needs it. To get this moving, please narrow the current PR to a non-breaking generic LLM-router integration:

- Remove `llm_router_query_template`, `llm_router_query_template_file`, and the 15-field public template-variable surface from this PR. Use a small internal prompt/request shape for now (for example, query plus available model names).
- Preserve `router_r1_server_url` as a deprecated compatibility alias. A direct rename to `llm_router_server_url` breaks existing v0.3 configs; validate the case where both names are present.
- Keep the configurable template as a follow-up issue/PR driven by a concrete routing or training experiment and its required fields.
- Register the mock flow in the repository E2E profile system instead of relying only on `run_llm_router_smoke.sh`.
- Update the current English documentation. Please do not rewrite the versioned v0.3 contract for a new mainline feature.
- Rebase onto current `main`, rerun `make vllm-sr-test`, `make test-semantic-router`, the CLI integration gate, and the affected E2E profile, then update the recorded results.

This scope preserves the useful de-branding/integration work without committing the project to an unstable template API.

FAUST-BENCHOU (commented):
Thanks for extending the rl-driven algorithm with configurable LLM router prompting. Keeping the integration behind a feature flag and validating the Router-R1-compatible flow end to end is a good approach.

PR #2283, opened 2026-06-29
[Router] persist demoted cache-similarity and context-token-count in replay record (#2254)

### LINKED ISSUE #2254: [Router] v0.4 headers: add replay record fields for cache-similarity and context-token-count
Follow-up to #2205 / #2229 — Part of #2200 (v0.4 response header contract).

## Background

#2229 demotes the intermediate signal/decision/tool/looper/cache/retention headers off the default response surface, behind the `x-vsr-debug` trigger. Two of the demoted fields have **no dedicated home in the replay record** (`store.Record` in `pkg/routerreplay/store/store.go`), so for any request that does *not* set `x-vsr-debug`, these values are not recoverable after the fact via `x-vsr-replay-id`:

- [ ] `x-vsr-cache-similarity` — only `RAGSimilarityScore` exists today (RAG, not the semantic-cache similarity).
- [ ] `x-vsr-context-token-count` — usage tokens (`PromptTokens`/`TotalTokens`/…) are stored, but not this context-token count.

These are the two unchecked boxes from #2205's "confirm each demoted header has a home in the replay record" acceptance checklist. #2229 was approved with this gap explicitly deferred (not silently dropped); this issue tracks closing it **before v0.4 ships**.

The other two boxes from #2205 are resolved without new fields and just need documenting on #2205:
- `x-vsr-injected-system-prompt` — recoverable from `RequestBody`.
- `x-vsr-retention-*` — runtime effects applied internally; observability-only, no replay home needed.

## Scope

1. Add `CacheSimilarity` (`float32`) and `ContextTokenCount` (`int`) fields to `store.Record`.
2. Wire the recorder so both are populated on the paths where the headers were previously emitted (semantic-cache hit path; looper/context path).
3. Confirm the values surface through the replay API (`GET /v1/router_replay/{replay-id}`) and the dashboard replay view.

## Note

The demote is purely additive-safe to extend — this work does not conflict with #2229's header logic and can land as its own clean PR.

Part of #2200
Follows #2205, #2229

theohsiung:
## Summary

Follow-up to #2229 (merged) — part of #2200 (v0.4 response header contract).

#2229 demoted the intermediate `x-vsr-cache-similarity` and `x-vsr-context-token-count` headers off the default response surface (behind `x-vsr-debug`). These two values had **no dedicated home in the replay record** (`store.Record`), so for any request that did *not* set `x-vsr-debug`, they were unrecoverable after the fact via `x-vsr-replay-id`. This closes that gap — the last two unchecked boxes from #2205's "every demoted header has a home in the replay record" acceptance checklist.

## Changes

- **`store.Record`**: add `CacheSimilarity float32` (`cache_similarity`) and `ContextTokenCount int` (`context_token_count`), both `omitempty`.
- **Recorder** (`pkg/extproc/recorder.go`): populate both in `buildReplayRoutingRecord` from `ctx.VSRCacheSimilarity` / `ctx.VSRContextTokenCount`, so every replay record carries them regardless of the `x-vsr-debug` trigger.
- **Dashboard replay view** (`insightsPageSupport.tsx` / `insightsPageTypes.ts`): surface the two fields in the replay/insights view.
- **Tests**: `record_demoted_header_fields_test.go` (store) + `recorder_test.go` additions.

The other two #2205 boxes need no new field (to be documented on #2205):
- `x-vsr-injected-system-prompt` — recoverable from `RequestBody`.
- `x-vsr-retention-*` — runtime effects applied internally; observability-only.

## Test Plan

- `go test ./pkg/routerreplay/... ./pkg/extproc/` — green (store record fields + recorder wiring).
- `tsc --noEmit` (dashboard/frontend) — clean.
- Rebased onto current `main` (includes #2229); no conflicts.

Purely additive to the replay record — does not change #2229's header logic.

Closes #2254
Part of #2200. Follows #2205, #2229.

### DIFF
--- a/src/semantic-router/pkg/routerreplay/store/store.go
+++ b/src/semantic-router/pkg/routerreplay/store/store.go
@@ -195,6 +195,18 @@ type Record struct {
 	RAGContextLength   int     `json:"rag_context_length,omitempty"`
 	RAGSimilarityScore float32 `json:"rag_similarity_score,omitempty"`
 
+	// v0.4 demoted-header replay homes (#2200, #2254): both values were removed
+	// from the default response header surface and now live only in the replay
+	// record, so they stay recoverable via x-vsr-replay-id when a request does
+	// not set x-vsr-debug.
+	//
+	// CacheSimilarity is the semantic-cache lookup similarity (0 = no lookup),
+	// formerly the x-vsr-cache-similarity header.
+	CacheSimilarity float32 `json:"cache_similarity,omitempty"`
+	// ContextTokenCount is the request context token count used for
+	// context-based routing, formerly the x-vsr-context-token-count header.
+	ContextTokenCount int `json:"context_token_count,omitempty"`
+
 	// Hallucination Detection
 	HallucinationEnabled    bool     `json:"hallucination_enabled,omitempty"`
 	HallucinationDetected   bool     `json:"hallucination_detected,omitempty"`
--- a/dashboard/frontend/src/pages/insightsPageSupport.tsx
+++ b/dashboard/frontend/src/pages/insightsPageSupport.tsx
@@ -346,6 +346,7 @@ export function buildInsightsRecordSections(
   sections.push({
     title: 'Usage & Cost',
     fields: [
+      { label: 'Context tokens', value: formatTokenValue(record.context_token_count) },
       { label: 'Prompt tokens', value: formatTokenValue(record.prompt_tokens) },
       { label: 'Completion tokens', value: formatTokenValue(record.completion_tokens) },
       { label: 'Total tokens', value: formatTokenValue(record.total_tokens) },
@@ -368,6 +369,7 @@ export function buildInsightsRecordSections(
     title: 'Plugin Status',
     fields: [
       { label: 'Cache', value: record.from_cache ? 'Hit' : 'Miss' },
+      { label: 'Cache similarity', value: formatSimilarityValue(record.cache_similarity) },
       { label: 'Streaming', value: record.streaming ? 'On' : 'Off' },
       { label: 'Guardrails', value: buildGuardrailsValue(record) },
       { label: 'RAG', value: buildRagValue(record) },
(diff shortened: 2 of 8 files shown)


src/semantic-router/pkg/routerreplay/store/store.go
```diff
@@ -195,6 +195,18 @@ type Record struct {
...
 	RAGSimilarityScore float32 `json:"rag_similarity_score,omitempty"`
 
+	// v0.4 demoted-header replay homes (#2200, #2254): both values were removed
+	// from the default response header surface and now live only in the replay
+	// record, so they stay recoverable via x-vsr-replay-id when a request does
+	// not set x-vsr-debug.
+	//
+	// CacheSimilarity is the semantic-cache lookup similarity (0 = no lookup),
+	// formerly the x-vsr-cache-similarity header.
+	CacheSimilarity float32 `json:"cache_similarity,omitempty"`
```
drivebyer, line 205: the Postgres backend silently drops both new fields. Redis/Milvus/Qdrant/memory json.Marshal the whole Record, but Postgres maps columns explicitly and this PR doesn't add them — so under store_backend: postgres (the documented default), cache_similarity/context_token_count are never written or read back

drivebyer (changes requested):

src/semantic-router/pkg/routerreplay/store/store.go
theohsiung: @drivebyer Thx for reviewing.

Fixed in f52311c5. Wired `cache_similarity` (REAL) and `context_token_count` (INTEGER) through the Postgres path — CREATE TABLE migration (`ADD COLUMN IF NOT EXISTS`), the INSERT column + $N placeholder list, the insert args, the SELECT column list, and the scan destinations — mirroring `rag_similarity_score` / `decision_tier`. Added `TestPostgresBackendWiresDemotedHeaderFields` pinning all four sites plus SELECT↔scan symmetry (the existing alignment test only covered the insert side).

drivebyer (approved):

PR #2289, opened 2026-06-30
[Router] Reclaim TTL-expired entries from the hybrid semantic cache

### LINKED ISSUE #2288: bug: [Router] Hybrid semantic cache retains TTL-expired entries in the in-memory HNSW index until capacity
### Describe the bug

Surfaced during the per-backend cache lifecycle audit requested in the review of
#2162.

The **hybrid** semantic cache keeps an in-process HNSW index (`embeddings`,
`idMap`, HNSW nodes) alongside the persistent Milvus store. Milvus expires
entries via its own TTL and silently deletes them, but the in-memory side is
**only ever evicted when it hits `maxMemoryEntries`** (default 100,000) via FIFO
— there is **no TTL-aware reclamation**:

- eviction triggers only on capacity:
  `src/semantic-router/pkg/cache/hybrid_cache.go:381` / `:454` / `:516`
  → `evictOneUnsafe()` (FIFO at index 0, `:602`)
- no background reaper: no goroutine / `time.Ticker` in `hybrid_cache*.go`
  (contrast `InMemoryCache.backgroundCleanup`, `inmemory_cache_lifecycle.go`)
- `RebuildFromMilvus()` re-syncs the index but is **only called once at startup**
  (`hybrid_cache.go:185`), never periodically

**This is bounded, not an unbounded leak** (capped at `maxMemoryEntries`), and it
is **not a correctness bug**: a stale HNSW node resolves to a Milvus `GetByID`
miss and returns `(nil, false)`, never a stale answer
(`hybrid_cache_lookup.go:107-120`).

Impact:

1. In-process memory is held far past TTL — up to the full `maxMemoryEntries`
   high-water mark regardless of how short the configured TTL is.
2. Wasted work: HNSW search traverses dead nodes and each stale candidate
   triggers a futile Milvus `GetByID` round-trip before falling through to miss.
3. FIFO eviction can drop a **live** entry while **expired** entries keep their
   slots → effective cache capacity (and hit rate) erodes under churn.

### To Reproduce

1. Enable the hybrid cache backend with a short TTL (e.g. `ttl_seconds: 60`).
2. Drive sustained cache writes at a rate below `maxMemoryEntries`.
3. Watch the in-memory entry gauge `metrics.UpdateCacheEntries("hybrid", ...)`.
4. After TTL elapses, the Milvus collection count drops but the in-memory entry
   count stays at its high-water mark and never decreases until capacity FIFO
   kicks in.

### Expected behavior

After an entry's TTL elapses, its in-memory HNSW node / `embeddings` / `idMap`
slot should be reclaimed (or periodically re-synced against Milvus's live set),
so in-process memory tracks the live working set rather than the all-time
high-water mark.

### Affected layer

Router — semantic cache (hybrid backend only).

### Additional context

Found while answering the lifecycle/leak question in the #2162 review. The other
five backends were audited and are clean: in-memory is bounded (HNSW rebuilt on
every eviction/TTL cleanup, background cleanup goroutine `Close()`-managed);
Redis/Milvus/Qdrant/Valkey hold a single reused client closed on `Close()` with
no per-request goroutines. Unrelated to the memory growth reported in #2222 (that
report has `semantic_cache` disabled).

Proposed fix: a `Close()`-managed background re-sync that reuses the existing
`RebuildFromMilvus()` on a configurable interval (gated on `ttl > 0`), mirroring
the in-memory backend's `backgroundCleanup` pattern; optionally lazy mark-dead on
a stale `GetByID` miss to amortize cost.

theohsiung:
Closes #2288

## Purpose

- **What:** Add a `Close()`-managed background goroutine to the **hybrid** semantic cache that periodically reclaims entries Milvus has already expired from the in-process HNSW index.
- **Why:** The hybrid backend keeps an in-memory HNSW index (`embeddings` / `idMap` / HNSW nodes) alongside the persistent Milvus store. Milvus expires entries via its own TTL and silently deletes them, but the in-memory side was only ever trimmed when it hit `maxMemoryEntries` (default 100,000) via FIFO — there was **no TTL-aware reclamation**. As a result, entries Milvus had already expired kept occupying in-process memory and HNSW search slots until the capacity cap forced eviction, so memory tracked the all-time high-water mark instead of the live working set (and FIFO could drop a *live* entry while keeping expired ones). Surfaced during the per-backend cache lifecycle audit requested in #2162's review.
- **Module:** `Router` (semantic cache, hybrid backend only).

### Approach

Reclamation reuses the existing `RebuildFromMilvus`: re-querying Milvus returns only the live (non-expired) entries, so the rebuilt index naturally drops the expired ones. The cadence is derived from the TTL (`reclaimIntervalForTTL`, floored at 60s so a short TTL can't trigger constant full rebuilds). The goroutine starts only when the cache is enabled, `ttl > 0`, and the startup rebuild mechanism is not disabled (so the existing `DisableRebuildOnStartup` opt-out also suppresses the recurring Milvus re-query). It is stopped before `Close()` takes the write lock, so an in-flight reclaim pass can finish and release the lock (no deadlock).

A `reclaimFn` seam (defaulting to `RebuildFromMilvus`) lets the white-box tests drive the loop without a live Milvus. Scope is intentionally narrow: in-process structures and the non-cgo stub are unchanged; the other five backends were audited as clean (no leak).

This is not an unbounded leak (it was already capped at `maxMemoryEntries`) and not a correctness bug (a stale HNSW node resolves to a Milvus `GetByID` miss, never a stale answer) — it is a memory-retention / efficiency fix. A lighter, fully-incremental local reclamation is possible future work.

## Test Plan

New white-box tests in `hybrid_cache_reclaim_test.go`, run under `-race`:

- `reclaimIntervalForTTL` floor/passthrough table.
- start/stop gating: no goroutine when the cache is disabled or `ttl <= 0`; `reclaimFn` defaults to `RebuildFromMilvus`.
- background loop lifecycle: ticks invoke `reclaimFn`, the loop survives a failing pass, exits cleanly on stop, and stop is idempotent (no double-close panic).

Commands (cgo + milvus tag, candle lib on `LD_LIBRARY_PATH`):

```
go vet -tags=milvus ./pkg/cache/
go test -tags=milvus -race -run 'Reclaim|BackgroundReclaim' ./pkg/cache/
go test -tags=milvus -race -count=1 ./pkg/cache/   # full suite, integration backends skipped
golangci-lint run --config tools/linter/go/.golangci.yml --build-tags=milvus ./pkg/cache/...
```

## Test Result

- New reclaim tests: 7/7 pass under `-race`.
- Full `pkg/cache` suite (integration backends skipped): green under `-race` (~34s), no regression.
- `gofmt`/`go vet` clean; `golangci-lint` (repo config) → 0 issues.
- Pure in-process change; no behavior change to routing/startup/config/API, so no E2E update required. Follow-up to the lifecycle question in #2162.

### DIFF
--- a/src/semantic-router/pkg/cache/hybrid_cache.go
+++ b/src/semantic-router/pkg/cache/hybrid_cache.go
@@ -94,6 +94,7 @@ type HybridCache struct {
 	// all-time high-water mark. See issue #2288.
 	reclaimTicker *time.Ticker
 	stopReclaim   chan struct{}
+	reclaimDone   chan struct{} // closed by the reclaim goroutine on exit; lets stop block until it has returned
 	closeOnce     sync.Once
 	reclaimFn     func(context.Context) error // defaults to RebuildFromMilvus; overridable in tests
 }
@@ -567,8 +568,12 @@ func (h *HybridCache) Close() error {
 		return nil
 	}
 
-	// Stop the background reclaim goroutine before taking the write lock so a
-	// reclaim pass in flight can finish and release the lock (avoids deadlock).
+	// Stop the background reclaim goroutine before taking the write lock. This
+	// is ordered (not locked) on purpose: stopBackgroundReclaim blocks until the
+	// goroutine has exited, and an in-flight reclaim pass needs h.mu to finish,
+	// so we must release the lock to it first. By the time we acquire the write
+	// lock below, no reclaim pass can still be running — which is what makes the
+	// subsequent nil-out of embeddings/idMap/hnswIndex safe (no use-after-close).
 	h.stopBackgroundReclaim()
 
 	h.mu.Lock()
--- a/src/semantic-router/pkg/cache/hybrid_cache.go
+++ b/src/semantic-router/pkg/cache/hybrid_cache.go
@@ -88,6 +88,14 @@ type HybridCache struct {
 
 	// Concurrency control
 	mu sync.RWMutex
+
+	// Background TTL reclamation: drops in-memory entries that Milvus has
+	// already expired so the index tracks the live working set rather than the
+	// all-time high-water mark. See issue #2288.
+	reclaimTicker *time.Ticker
+	stopReclaim   chan struct{}
+	closeOnce     sync.Once
+	reclaimFn     func(context.Context) error // defaults to RebuildFromMilvus; overridable in tests
 }
 
 // HybridCacheOptions contains configuration for the hybrid cache
@@ -190,6 +198,11 @@ func NewHybridCache(options HybridCacheOptions) (*HybridCache, error) {
 			})
 			// Don't fail initialization, just log warning and continue with empty index
 		}
+
+		// Start periodic TTL reclamation (reuses the same rebuild path). Gated
+		// on the rebuild mechanism being enabled so the DisableRebuildOnStartup
+		// opt-out also suppresses the recurring Milvus re-query.
+		cache.startBackgroundReclaim()
 	} else {
 		logging.ComponentWarnEvent("cache", "hybrid_cache_rebuild_skipped", map[string]interface{}{
 			"source":      "startup",
@@ -553,6 +566,10 @@ func (h *HybridCache) Close() error {
 	if !h.enabled {
 		return nil
 	}
+
+	// Stop the background reclaim goroutine before taking the write lock so a
+	// reclaim pass in flight can finish and release the lock (avoids deadlock).
+	h.stopBackgroundReclaim()
 
 	h.mu.Lock()
 	defer h.mu.Unlock()
--- a/src/semantic-router/pkg/cache/hybrid_cache_reclaim.go
+++ b/src/semantic-router/pkg/cache/hybrid_cache_reclaim.go
@@ -1,0 +1,98 @@
+//go:build !windows && cgo
+
+package cache
+
+import (
+	"context"
+	"time"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
+)
+
+const (
+	// minReclaimInterval is the floor for the background TTL-reclaim cadence.
+	// A reclaim pass performs a full RebuildFromMilvus, so we never run it more
+	// often than this regardless of how short the configured TTL is.
+	minReclaimInterval = 60 * time.Second
+
+	// reclaimTimeout bounds a single background reclaim pass.
+	reclaimTimeout = 5 * time.Minute
+)
+
+// reclaimIntervalForTTL derives the background reclaim cadence from the TTL.
+// We re-sync roughly once per TTL window (so an expired entry is dropped within
+// about one extra TTL period), floored at minReclaimInterval to avoid hammering
+// Milvus with full rebuilds when the TTL is very short.
+func reclaimIntervalForTTL(ttlSeconds int) time.Duration {
+	d := time.Duration(ttlSeconds) * time.Second
+	if d < minReclaimInterval {
+		return minReclaimInterval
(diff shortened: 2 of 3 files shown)


theohsiung:
@WUKUNTAI-0211 can you help review this PR

WUKUNTAI-0211 (dismissed):
Dismissed — superseded by the comment below.

WUKUNTAI-0211 (commented):
Thanks @theohsiung — pulled it down and ran the new tests under `-race` (cgo + `milvus` tag); the 7 reclaim tests and the rest of the non-integration `pkg/cache` suite are green, and `go vet -tags=milvus` is clean. Scope, gating on the existing `DisableRebuildOnStartup` switch, the 60s floor, and the `reclaimFn` seam all look right.

One thing worth flagging as a follow-up (not blocking):

`stopBackgroundReclaim` only sends the stop signal — it doesn't wait for the goroutine to exit (`hybrid_cache_reclaim.go:88-97`). So if `Close()` lands while a reclaim pass is mid-`RebuildFromMilvus` (specifically: after `GetAllEntries` started but before it takes the write lock at `hybrid_cache.go:293`), the goroutine can grab the write lock *after* `Close()` has nil'd `embeddings`/`idMap`/`hnswIndex` and then call `addNodeHybrid` on a nil `hnswIndex`. Very hard to hit in practice (shutdown-timing × ≥60s ticker × the narrow window where `GetAllEntries` returns success after the Milvus client was closed), but the comment at `hybrid_cache.go:570-572` reads as if ordering alone fixes it, when the missing piece is a `WaitGroup` / `reclaimDone` channel so stop actually blocks until the goroutine returns. Worth a small follow-up since once that's in, `Close()`'s ordering becomes correct for the right reason and a fast-ticker unit test can lock it in.

WUKUNTAI-0211 (dismissed):

theohsiung:
Thanks for the re-review @WUKUNTAI-0211 (and for updating the branch). I went ahead and addressed the `Close`-vs-reclaim follow-up in this PR rather than deferring it — it's a small, self-contained change and your write-up made the fix obvious:

- Added a `reclaimDone` channel that the reclaim goroutine closes on exit; `stopBackgroundReclaim` now **blocks** on it until the goroutine has actually returned (`hybrid_cache_reclaim.go`).
- Because `stopBackgroundReclaim` is called from `Close()` **before** the write lock is taken, an in-flight `RebuildFromMilvus` pass can still acquire `h.mu`, finish, and exit — so by the time `Close()` nils `embeddings`/`idMap`/`hnswIndex`, no pass is running (no use-after-close). Rewrote the `Close()` comment so the ordering reads as load-bearing for *that* reason, not just deadlock avoidance.
- New regression test `TestStopBackgroundReclaimWaitsForInFlightReclaim`: drives a 5ms ticker, lets a pass enter `reclaimFn`, then asserts `stopBackgroundReclaim` does **not** return until the in-flight pass completes (and returns promptly once it does). Fails without the wait.
- Minor: symmetric `hybrid_cache_reclaim_stopped` log on exit, and a comment on `reclaimTimeout` noting the lock-hold window.

Validated locally with cgo + `milvus` tag, candle lib on `LD_LIBRARY_PATH`: `go build`/`go vet` clean and all 8 reclaim tests green under `-race`. PTAL.

WUKUNTAI-0211 (approved):
LGTM 🙏 The `reclaimDone` wait + `TestStopBackgroundReclaimWaitsForInFlightReclaim` cover the lifecycle gap cleanly, and the `reclaimTimeout` comment now matches the actual lock-hold window. Thanks for turning the follow-up around so quickly.

PR #2291, opened 2026-06-30
[CLI] add Podman support for local vllm-sr serve

### LINKED ISSUE #2296: feat: support Podman as a first-class local container runtime
## Motivation

`vllm-sr serve` and `make vllm-sr-dev` currently hard-require Docker. On hosts where Docker is unavailable or undesirable (rootful container management on shared servers, distros that ship Podman by default, environments that prohibit the Docker daemon), users hit fail-fast errors like:

> Podman is not supported for local vLLM Semantic Router deployment.

even though every container operation the CLI performs (`run`/`exec`/`network`/`logs`/`inspect`) has an exact Podman counterpart.

## Proposed Behavior

Treat Podman as a first-class peer of Docker for the local target:

- `--runtime {docker,podman}` flag on `serve` / `status` / `logs` / `stop` / `dashboard`, equivalent to `CONTAINER_RUNTIME=<runtime>`.
- Auto-detection: prefer `docker` when present, fall back to `podman`.
- Image references, FROM directives, GPU passthrough, host-gateway flag, and dashboard container lifecycle all work transparently for either runtime.
- Existing Docker workflows must remain unchanged.

## Notes

- Podman exposes a Docker-Engine-API-compatible socket at `/run/podman/podman.sock`, so the dashboard's in-image `docker` CLI can drive container lifecycle through Podman with zero Go-side changes if the socket is mounted at the canonical `/var/run/docker.sock` path.
- NVIDIA GPU passthrough on Podman uses CDI (`--device nvidia.com/gpu=all`) instead of `--gpus all --runtime nvidia`.
- Podman's default short-name policy rejects unqualified images, so all base images and storage/observability image references should be fully qualified with `docker.io/...` (a no-op on Docker).

## Validated Scenarios

- `make vllm-sr-dev CONTAINER_RUNTIME=podman` (CPU)
- `make vllm-sr-dev VLLM_SR_PLATFORM=amd CONTAINER_RUNTIME=podman` (ROCm)
- First-run wizard → Activate → router transitions Created → Up via the mounted podman socket
- mmBERT ONNX embedding model loads with ROCm execution provider on AMD GPU
- Container-to-container DNS, L4 TCP, and L7 HTTP across all stack pairs

haowu1234:
Closes #2296

## Summary

Add Podman as a first-class local runtime alongside Docker for `vllm-sr serve` and `make vllm-sr-dev`. Existing Docker workflows are unchanged.

## Changes

- **`--runtime {docker,podman}`** flag on `serve` / `status` / `logs` / `stop` / `dashboard`, plumbed via the existing `CONTAINER_RUNTIME` env var
- **`cli/docker_runtime.py`** accepts Podman, falls back to it when Docker is absent or resolves to a Podman shim
- **`cli/docker_run_command.py`** extends `--add-host=...:host-gateway` and NVIDIA GPU passthrough (CDI on Podman, `--gpus all --runtime nvidia` on Docker) to both runtimes
- **All Dockerfile FROM images and storage/observability image references fully qualified with `docker.io/...`** — Podman's strict short-name policy rejects unqualified images by default; Docker silently rewrites to the same canonical path so this is a no-op there
- **Dashboard container lifecycle on Podman**: `_attach_docker_socket` now mounts the Podman API-compatible socket at the canonical `/var/run/docker.sock`, so the dashboard's in-image `docker` CLI drives container lifecycle through Podman with **zero Go-side changes**

## Validation

- **419 unit tests pass** (existing 386 + new 33 covering podman detection, host-gateway, NVIDIA CDI, ROCm Dockerfile FROM qualification, and podman socket mount)
- Verified end-to-end on a remote 9-GPU AMD MI300X host (Ubuntu 24.04, podman 4.9.3):
  - `make vllm-sr-dev VLLM_SR_PLATFORM=amd CONTAINER_RUNTIME=podman` builds the 15.6 GB ROCm router image
  - `vllm-sr serve --runtime podman --platform amd` starts the full stack
  - First-run wizard → Activate → router transitions Created → Up automatically through the mounted podman socket
  - `mmBERT` ONNX embedding model loads with **ROCm execution provider** on AMD GPU (3 FP16 layers)
  - CPU mode (`make vllm-sr-router-build CONTAINER_RUNTIME=podman` then `vllm-sr serve --runtime podman`) verified through the same wizard → Activate → `startup_complete` flow
- Container-to-container DNS, L4 TCP, and L7 HTTP all verified across 6 service pairs (router↔milvus/redis/postgres/sim, envoy↔router, dashboard↔router) plus a freshly-attached external mock backend

## Known limitations (none introduced by this PR)

- ROCm image build needs ≥ 40 GB free disk (build pipe failures otherwise — runtime-agnostic, also affects `docker build`)
- On Linux with `ufw` enabled: must `ufw allow from <podman-bridge-subnet>` so containers can reach `aardvark-dns` on the host gateway

## Test plan

- [x] `pytest src/vllm-sr/tests/` — 419 passing
- [x] Manual: `make vllm-sr-dev CONTAINER_RUNTIME=podman` then `vllm-sr serve --runtime podman` then First-run wizard Activate (CPU)
- [x] Manual: `make vllm-sr-dev VLLM_SR_PLATFORM=amd CONTAINER_RUNTIME=podman` then `vllm-sr serve --runtime podman --platform amd` then First-run wizard Activate (ROCm + AMD GPU passthrough)
- [x] Existing Docker path unchanged: full pytest suite includes the docker fixtures


src/vllm-sr/cli/container_openclaw_support.py
```diff
@@ -42,26 +42,34 @@ def configure_openclaw_support(
...
+        # image already ships a real `docker` CLI; mounting podman.sock at the
+        # canonical /var/run/docker.sock path lets the in-image docker CLI
+        # drive container lifecycle (start/stop/inspect/logs) through podman
+        # transparently — no Go-side changes needed.
+        _attach_docker_socket(mount_specs, runtime)
+        env_vars["OPENCLAW_CONTAINER_RUNTIME"] = "docker"
+        log.info(
+            "Podman runtime: dashboard will use the in-image Docker CLI against "
+            "the mounted podman.sock for container lifecycle"
+        )
```
AayushSaini101, line 62: @haowu1234 i guess we need to change the naming of the file names of the command because the currently it is docker_openclaw_support.py and clearly mentioned related to the docker service, now we are adding the podman support in this, this is not good flow if we thing based on the user experience thoughts ? same applies for the rest of the commands and other places in the MR thanks : )

haowu1234: Thanks for the review, you're right — the naming was inconsistent. I've renamed all CLI modules, functions, classes, constants, and env vars from `docker_*` to `container_*` throughout the PR, so the naming now clearly reflects that these modules support both Docker and Podman as first-class runtimes. PTAL, thanks!

dashboard/backend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Stage 0: Build DSL compiler WASM assets
-FROM golang:1.24-alpine AS wasm-builder
+FROM docker.io/library/golang:1.24-alpine AS wasm-builder
```
AayushSaini101: @haowu1234 i guess we shouldn't hardcore the registry while fetching the dependencies if the there are multiple registeries are present, like if one registry is failed it will automatically search for the second for the second, and now consider if we are running the project in the restricted docker env, many companies are using podman only, that's why i guess we can make it default for all the places thoughts ? 
```
unqualified-search-registries = [
    "registry.company.com",
    "docker.io"
]
```

AayushSaini101:
Overall looks Good to me, thank you for the implementation : )

dashboard/backend/Dockerfile
haowu1234: I've replaced the hardcoded docker.io/ with a runtime probe: when the build target is Podman, we test whether unqualified short names are resolvable by pulling library/alpine:latest. If the probe succeeds (search chain is configured), IMAGE_REGISTRY defaults to "" so your registries.conf priority order is honoured. If the probe fails (no search chain), it falls back to docker.io/ for out-of-the-box compatibility. Docker builds always use docker.io/. Still overridable via IMAGE_REGISTRY=... on the command line. PTAL, thanks!

AayushSaini101 (approved):
LGTM thanks : )

PR #2302, opened 2026-06-30
Fix auto-discovery for mmbert32k merged models

### LINKED ISSUE #2301: Unified Classifier Auto-Discovery Fails for mmbert32k Merged Models
## Description
The initialization of the new `mmbert32k-*-merged` models is failing across three different layers of the unified classifier stack:

1. **Go Auto-Discovery Engine**: The `AutoDiscoverModels` function fails to discover the new `mmbert32k-*-merged` models because the scanning logic has hardcoded checks that expect `"lora_"` prefixes or `"category"` strings in the directory names.
2. **Rust Core Architecture (candle-binding)**: Even when properly discovered and passed the `modernbert` architecture flag, the PII token classifier in `pii_lora.rs` unconditionally instantiates a standard BERT token classifier. This crashes immediately when it encounters the `sans_pos` positional embeddings used by the `mmbert32k` architecture.
3. **Benchmark Test Suite**: The `classification_bench_test.go` script successfully initializes the engine but then discards the returned instance and attempts to run the benchmarks against an uninitialized global singleton.

As a result, the unified classifier cannot be initialized during benchmarks (like `make perf-bench-quick`), preventing us from testing the performance improvements.

## Steps to Reproduce
1. Download the `mmbert32k-*-merged` models into the `models/` directory.
2. Run `make perf-bench-quick`.
3. The initialization will fail and crash the classification benchmarks.

## Expected Behavior
- The auto-discovery engine should recognize `mmbert32k` directories and correctly classify them.
- The Rust bindings should seamlessly route to the `TraditionalModernBertTokenClassifier` for PII detection when `is_modernbert` is true.
- The benchmark suite should test against the initialized instance.

## Actual Behavior
The engine skips directories, the Rust bindings crash on `sans_pos` embeddings, and the test suite tests a nil singleton.

## H2 execution scope

Fix unified classifier auto-discovery for mmBERT merged models and align the fix with signal model lifecycle/capability metadata from #2360/#2358.

Relevant code surfaces: `src/semantic-router/pkg/modelruntime`, `src/semantic-router/pkg/classification`, model catalog/config, and native bindings.

## Acceptance criteria

- The mmBERT merged-model failure is reproduced in a test or documented validation receipt.
- Auto-discovery resolves the correct model family/capability metadata.
- Unsupported merged models fail with actionable diagnostics.
- Multimodal and classifier readiness reporting stays consistent with #2318.

TwilightTechie:
Fixes #2301

### DIFF
--- a/candle-binding/src/classifiers/lora/pii_lora.rs
+++ b/candle-binding/src/classifiers/lora/pii_lora.rs
@@ -4,14 +4,22 @@
 
 use crate::core::{ModelErrorType, UnifiedError};
 use crate::model_architectures::lora::bert_lora::HighPerformanceBertTokenClassifier;
+use crate::model_architectures::traditional::modernbert::TraditionalModernBertTokenClassifier;
 use crate::model_error;
 use candle_core::Result;
+use std::path::Path;
 use std::time::Instant;
+
+/// Classifier backend enum to avoid Box<dyn Trait>
+enum TokenClassifierBackend {
+    Bert(HighPerformanceBertTokenClassifier),
+    ModernBert(TraditionalModernBertTokenClassifier),
+}
 
 /// PII detector with real token classification model inference (merged LoRA models)
 pub struct PIILoRAClassifier {
-    /// High-performance BERT token classifier for PII detection
-    bert_token_classifier: HighPerformanceBertTokenClassifier,
+    /// Classifier backend (either BERT or ModernBERT/mmBERT)
+    backend: TokenClassifierBackend,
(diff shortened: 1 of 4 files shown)


TwilightTechie:
Hi @AayushSaini101 the build_pr (dashboard) failure is the transient pytorch-CDN SSL flake (passing on all other runs today), could you re-run failed jobs? Everything else is green.

AayushSaini101 (approved):
LGTM thanks for the fix : )

Issue #2222, opened 2026-06-17
[Bug] Investigate unbounded request-path memory retention under repeated load


poggioxk:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

# Possible memory leak or unbounded memory retention under load in vLLM Semantic Router v0.3

### To Reproduce

1. Start vLLM Semantic Router with the attached configuration.
2. Ensure these features are disabled:

   * `response_api.enabled: false`
   * `observability.tracing.enabled: false`
   * `router_replay.enabled: false`
   * `semantic_cache.enabled: false`
3. Run a non-streaming chat completion load test:

   * 400 users
   * 30 users/second spawn rate
   * About 25 RPS
   * Average latency around 4 seconds
4. Stop the load test.
5. Observe memory usage after CPU returns to 0%.
6. Repeat the load test again without restarting the router process.

### Expected behavior

See **Expected result** above: memory should return toward an idle baseline or stabilize at a repeatable warm-state plateau.

### Impact

Not separately specified in the original report.

### Additional context

This report describes v0.3. A current-main reproduction with heap/RSS profiles and repeated load/idle cycles is required before treating the listed possible causes as findings.

> [!NOTE]
> Current maintainer scope (2026-09-01): keep this in needs-acceptance until it is reproduced on current main. Run repeated load/idle/soak cycles; capture RSS, heap, goroutine, allocator, and bounded high-cardinality state; and show whether memory reaches a stable plateau. The historical v0.3 report below remains useful evidence, not a current-main diagnosis.

##### Summary

I am seeing continuous memory growth in vLLM Semantic Router under load. After stopping the load test, CPU usage returns to 0%, but memory usage does not decrease. Re-running the load test causes memory usage to continue increasing.

This still happens after disabling `response_api`, OpenTelemetry tracing, `router_replay`, and `semantic_cache`, so the issue does not appear to be caused by those optional storage/tracing/cache features.

##### Environment

* vLLM Semantic Router config version: `v0.3`
* Service version in config: `v0.3.0`
* CPU: 16 cores
* Memory: 29.9 GB
* Backend protocol: OpenAI-compatible HTTP backend
* Request type: non-streaming chat completion
* Tested endpoint/name: `llm_chat_no_stream`
* Load test users: 400
* Spawn rate: 30 users/second
* Observed RPS: about 25.7 RPS
* Average latency: about 3946 ms
* Estimated in-flight requests: around 100

##### What I expected

[...]

PR #2305, opened 2026-07-01
[Router] Bound the uncapped session-telemetry stores (size cap)

### LINKED ISSUE #2304: bug: [Router] session-telemetry stores lack a size cap → unbounded retention under high session cardinality
### Describe the bug

The `sessiontelemetry` package keeps three in-memory stores keyed by a session
ID that is **derived from message content** (`deriveSessionIDFromMessages` /
`DeriveChatCompletionsSessionID`) when no explicit conversation/session ID is
supplied. Distinct conversations therefore create distinct entries, so the key
space scales with traffic cardinality, not with a fixed client count.

Three of these stores grow their session map with **no size cap**:

| Store | File | TTL | Size cap |
|---|---|---|---|
| `routerSessionMemoryStore` | `router_memory.go` | 24h | **none** |
| turn store (`store`) | `telemetry.go` | 1h | **none** |
| `transitionStore` | `transition.go` | **none** | **none** (per-session events capped at 200, but the session map never evicts) |

Only `lastModelStore` (`last_model.go`) is bounded — it has both a 1h TTL **and**
`maxLastModelSessions = 50_000` + `evictOldestLocked`. Its own doc comment even
notes the telemetry and transition stores as the in-memory siblings, so the
missing caps look like an oversight rather than intent.

All three are written on the hot path (`RecordSessionDecision` /
`RecordSessionUsage` / `RecordTurn` per request, `RecordTransition` per model
switch — see `pkg/extproc/session_telemetry.go`).

### Impact

Under sustained traffic with high session cardinality (e.g. a load test with
many distinct prompts, or normal production traffic with many users/
conversations), these maps accumulate one entry per distinct session. Because
entries persist for the TTL window — **24h** for `routerSessionMemoryStore`, and
**forever** for `transitionStore` (no TTL at all) — the process memory:

- stays elevated after load stops / CPU returns to idle (entries are not
  eligible for eviction yet), and
- keeps increasing across repeated load rounds (each round's new distinct
  sessions are added and not drained between rounds).

This matches the "memory stays high after the load test stops" and "repeated
runs keep increasing memory" behaviour reported in #2222. (It is a slow,
bounded-by-TTL growth on the Go heap; it is a distinct contributor from the
large warm-state working-set RSS held by the CGO/candle model side, which
plateaus.)

### Expected behavior

Each session store should bound its total number of tracked sessions, mirroring
`lastModelStore`, so memory stays bounded regardless of session cardinality.

### Affected layer

Router — `pkg/sessiontelemetry`.

### Additional context

Surfaced while investigating #2222 (heap/RSS profiling under load showed
`routerSessionMemoryStore.sessionLocked` and the telemetry store growing while
the Go heap otherwise stayed tiny). Proposed fix: add a `maxSessions` cap +
`evictOldestLocked` (least-recently-seen) to each of the three stores, mirroring
`last_model.go`. TTL eviction is unchanged.

theohsiung:
Closes #2304

## Purpose

- **What:** Add a per-store session-count cap to the three `sessiontelemetry` stores that had none — `routerSessionMemoryStore` (`router_memory.go`), the cumulative turn store (`telemetry.go`), and `transitionStore` (`transition.go`).
- **Why:** These stores key on a **message-derived session ID**, so distinct conversations create distinct entries and the key space scales with traffic cardinality. Only `lastModelStore` was bounded (`maxLastModelSessions = 50_000` + `evictOldestLocked`); its own comment already named the telemetry and transition stores as the uncapped siblings. Because entries persist for the TTL window — **24h** for `routerSessionMemoryStore`, and **forever** for `transitionStore` (no TTL at all) — the maps accumulate under sustained high-cardinality load, so memory stays elevated after load stops and grows across repeated rounds. This is the Go-side contributor to the retention behaviour in #2222 (distinct from the CGO/candle model working-set RSS, which plateaus).
- **Module:** `Router` — `pkg/sessiontelemetry`.

## Change

Each store now evicts the least-recently-seen session when a new session would exceed a `50_000` cap, mirroring `last_model.go` exactly:

- `routerSessionMemoryStore`: cap check in `sessionLocked` + `evictOldestLocked` (by `lastSeen`).
- telemetry turn store: cap check in `recordTurnState` + package-level `evictOldestLocked` (by `turnState.lastSeen`).
- `transitionStore`: cap check in `RecordTransition` + `evictOldestLocked` (by each session's most-recent event `Timestamp`); the per-session 200-event cap is unchanged.

Existing TTL eviction is untouched. The caps are `var`s so tests exercise the eviction path without inserting 50k entries. No behaviour change for any realistic session count.

## Test Plan

New `store_size_cap_test.go` (pure Go), run under `-race`:

- Each store: lower the cap, insert more distinct sessions than the cap with distinct timestamps, assert the session map stays `<= cap` and that the oldest entry is evicted while the newest is retained.

Commands:

```
CGO_ENABLED=1 go test -race -count=1 ./pkg/sessiontelemetry/
gofmt -l / go vet ./pkg/sessiontelemetry/
golangci-lint run --config tools/linter/go/.golangci.yml ./pkg/sessiontelemetry/...
```

## Test Result

- Full `pkg/sessiontelemetry` suite green under `-race` (new caps + existing tests, incl. `last_model`'s cap test).
- `gofmt`/`go vet` clean; `golangci-lint` (repo config) → 0 issues.
- Pure in-memory bookkeeping change; no routing/startup/config/API surface affected, so no E2E update required.

## Notes

Scoped to the Go-side unbounded-retention facet of #2222 only. The bulk of the reported multi-GB growth is warm-state working-set RSS on the CGO/candle model side (which plateaus at a high-water mark and does not return to the OS) — that is a separate, largely-expected behaviour addressed by tuning/guidance, not by this PR. #2222 stays open as the umbrella; see the discussion there.

### DIFF
--- a/src/semantic-router/pkg/sessiontelemetry/last_model.go
+++ b/src/semantic-router/pkg/sessiontelemetry/last_model.go
@@ -10,6 +10,17 @@ import (
 // eviction path without inserting tens of thousands of entries.
 var maxLastModelSessions = 50_000
 
+// evictionSampleSize bounds evictOldestLocked to an approximate (sampled) LRU:
+// rather than scanning the whole map for the true least-recently-seen entry
+// (O(N) under the write lock — which bites exactly at the cap this guards, since
+// once the map is full every new session would trigger a full scan), we sample
+// this many entries (Go map iteration order is randomized) and evict the oldest
+// of the sample. O(1), the same approach Redis uses for approximate LRU.
+// Intentionally approximate: bounding the map is what matters, evicting the
+// exact oldest does not. Shared by the evictOldestLocked of all four session
+// stores in this package.
+const evictionSampleSize = 10
+
 type lastModelState struct {
 	model    string
 	lastSeen time.Time
@@ -98,20 +109,23 @@ func (s *lastModelStore) evictExpiredLocked(t time.Time) {
 	}
 }
 
-// evictOldestLocked removes the single least-recently-seen entry. It is a
-// best-effort safety valve for the size cap when TTL eviction did not free room.
+// evictOldestLocked evicts the oldest entry among a bounded random sample
+// (approximate LRU — see evictionSampleSize). It is a best-effort safety valve
+// for the size cap when TTL eviction did not free room. Callers must hold s.mu.
 func (s *lastModelStore) evictOldestLocked() {
 	var oldestKey string
 	var oldestSeen time.Time
-	first := true
+	sampled := 0
 	for k, v := range s.sessions {
-		if first || v.lastSeen.Before(oldestSeen) {
-			oldestKey = k
-			oldestSeen = v.lastSeen
-			first = false
+		if sampled == 0 || v.lastSeen.Before(oldestSeen) {
+			oldestKey, oldestSeen = k, v.lastSeen
+		}
+		sampled++
+		if sampled >= evictionSampleSize {
+			break
 		}
 	}
-	if oldestKey != "" {
+	if sampled > 0 {
 		delete(s.sessions, oldestKey)
 	}
 }
--- a/src/semantic-router/pkg/sessiontelemetry/router_memory.go
+++ b/src/semantic-router/pkg/sessiontelemetry/router_memory.go
@@ -8,6 +8,14 @@ import (
 
 const routerMemoryTTL = 24 * time.Hour
 
+// maxRouterSessions caps the number of sessions tracked by the router-owned
+// memory store so memory stays bounded under high session cardinality (session
+// IDs are derived from message content, so distinct conversations create
+// distinct entries). Mirrors last_model.go's maxLastModelSessions. It is a var
+// (not a const) so tests can exercise the eviction path without inserting tens
+// of thousands of entries.
+var maxRouterSessions = 50_000
+
 // RouterSessionSnapshot is the router-owned, model-independent memory for a
 // session. It is intentionally about routing state, not prompt-visible user
 // memory.
@@ -229,6 +237,9 @@ func (s *routerSessionMemoryStore) sessionLocked(sessionID string) *routerSessio
 	if st != nil {
 		return st
 	}
+	if len(s.sessions) >= maxRouterSessions {
+		s.evictOldestLocked()
+	}
 	st = &routerSessionState{
 		sessionID:  sessionID,
 		modelTurns: make(map[string]int),
@@ -242,6 +253,25 @@ func (s *routerSessionMemoryStore) evictExpiredLocked(now time.Time) {
 		if now.Sub(v.lastSeen) > routerMemoryTTL {
 			delete(s.sessions, k)
 		}
+	}
+}
+
+// evictOldestLocked removes the single least-recently-seen session. It is a
+// best-effort safety valve for the size cap when TTL eviction did not free room.
+// Callers must hold s.mu.
+func (s *routerSessionMemoryStore) evictOldestLocked() {
+	var oldestKey string
+	var oldestSeen time.Time
+	first := true
+	for k, v := range s.sessions {
+		if first || v.lastSeen.Before(oldestSeen) {
+			oldestKey = k
+			oldestSeen = v.lastSeen
+			first = false
+		}
+	}
+	if oldestKey != "" {
+		delete(s.sessions, oldestKey)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


theohsiung:
@WUKUNTAI-0211 need your review

WUKUNTAI-0211 (commented):
Thanks for tackling this — the direction is right. `lastModelStore` was the only bounded store and its own comment already flagged these three as the uncapped siblings, so mirroring that cap here is exactly the right move, and keeping the caps as `var`s so the tests can drive the eviction path without inserting 50k entries is a nice touch. I pulled the branch down and read through all three stores.

### What's good

- Caps mirror `last_model.go` consistently across all three stores, and the TTL eviction paths are left untouched.
- The transition store's guard (`!exists && len(...) >= max`) correctly only evicts when a genuinely new session key is added, so re-recording an existing session at the cap boundary doesn't trigger a needless eviction.
- The cap tests are deterministic and assert both the evict-oldest and retain-newest halves.

### Findings

#### Important (perf under the exact scenario this PR targets — not a CI blocker)

`evictOldestLocked()` in all three stores (`router_memory.go`, `telemetry.go`, `transition.go`) does a full `O(N)` scan of the session map to find the single oldest entry, and it runs **while holding the exclusive write lock**. Once the map reaches the cap — which is precisely the high-cardinality state this PR exists to defend against, not a rare edge case — *every* subsequent new-session write triggers one full-map scan under the lock.

I ran a quick local benchmark on the eviction alone (50k entries, values are pointers as in the real stores):

```
BenchmarkEvictFull50k-20        1555416 ns/op   (~1.5 ms)
BenchmarkEvictSampled50k-20        6193 ns/op   (~0.006 ms)
```

At ~1.5 ms per eviction under the write lock, the new-session write path is serialized to roughly a few hundred inserts/sec once the map is full. Under a burst of distinct sessions (the same burst that fills the map), goroutines queue on the lock and latency climbs — so this effectively converts the unbounded-memory problem into a lock-contention / tail-latency one under load.

Suggested fix — random-sampled LRU (same approach Redis uses for approximate LRU): since Go map iteration is randomized, sample a small fixed number of entries and evict the oldest of that sample. That drops the eviction from `O(N)` to `O(1)`:

```go
func (s *routerSessionMemoryStore) evictOldestLocked() {
    var oldestKey string
    var oldestSeen time.Time
    first := true
    count := 0
    for k, v := range s.sessions {
        if first || v.lastSeen.Before(oldestSeen) {
            oldestKey = k
            oldestSeen = v.lastSeen
            first = false
        }
        count++
        if count >= 10 { // sample size, not full scan
            break
        }
    }
    if oldestKey != "" {
        delete(s.sessions, oldestKey)
    }
}
```

Worth noting this makes eviction *approximate* LRU rather than exact — but that's fine for a best-effort safety valve, and it doesn't touch your current tests: with the test cap at `3`, a sample of `10` still walks the whole (3-entry) map, so the victim is identical and the tests stay deterministic. (Only caveat for future test authors: a test cap between the sample size and the insert count would lose that determinism — that's inherent to sampled LRU, just something to keep in mind.)

#### Minor / nit (optional)

- `routerSessionCount()` and `telemetrySessionCount()` take the exclusive `mu.Lock()` for a read-only length, whereas `transitionSessionCount()` uses `mu.RLock()`. These are all test-only helpers so there's no production impact, but matching the `RLock()` form would be more consistent.
- In `transition.go`'s `evictOldestLocked()`, the `if len(list) == 0` branch is effectively dead — `RecordTransition` always appends at least one event before storing, so no zero-length slice is ever persisted. Harmless as defensive code, just flagging it.

### Verdict

`comment` — the cap itself is correct and I'd be happy to see it land; the `O(N)`-under-lock eviction is the one thing I'd like addressed first since it bites in exactly the scenario this PR is meant to protect, and the sampled-LRU fix is ~3 lines with no test changes.

theohsiung:
Addressed in b2c6876b.

**Sampled eviction (the O(N)-under-lock finding).** Switched `evictOldestLocked` to approximate LRU: sample `evictionSampleSize` (10) entries — Go map iteration is randomized — and evict the oldest of the sample. O(1), same approach you referenced. The cap still bounds the map exactly. I also applied it to `last_model.go`: it's the store these caps were mirrored from and carried the identical full scan, so making all four consistent seemed better than leaving three sampled and one O(N). Sample size is a shared package const with the rationale in a comment. Existing tests stay deterministic — their cap (3) is below the sample (10), so eviction still walks the whole map and picks the exact oldest.

**Dead `len(list) == 0` branch in `transition.go`.** Removed — `RecordTransition` always appends before storing, so it was unreachable.

**The `RLock()` consistency nit — left as-is, and I believe that's correct.** `routerSessionMemoryStore.mu` and `telemetry.go`'s package-level `mu` are `sync.Mutex`, which has no `RLock()` method — so `Lock()` is the only option in `routerSessionCount()` / `telemetrySessionCount()`. Only `transitionStore` uses `sync.RWMutex`, which is why `transitionSessionCount()` can `RLock()`. Making the three uniform would mean converting those two stores to `RWMutex`, adding read-lock overhead to single-writer hot paths for a test-only counter — didn't seem worth it. Happy to convert if you'd prefer uniformity.

All four stores + the cap tests pass under `-race`; gofmt/vet/golangci clean.

Xunzhuo (approved):

WUKUNTAI-0211 (approved):

PR #2258, opened 2026-06-21
Complete Router Learning follow-up adaptations

Xunzhuo:
## Summary

Completes the first Router Learning implementation scope from `website/docs/proposals/router-learning-memory-and-adaptations.md`.

This PR replaces the old learning-family public API with a smaller self-improving router surface:

- `global.router.learning.adaptation` with the day-0 `routing_sampling` strategy.
- `global.router.learning.protection` with conversation/session switch protection.
- `routing.decisions[].adaptations` for per-decision apply/observe/bypass control.
- Replay-linked `POST /v1/router/outcomes` for typed outcome ingestion.
- Compact learning headers plus typed replay diagnostics.
- Offline `vllm-sr eval recipe-learning` artifacts for metrics, findings, candidate recipes/patches, and experience seed packs.
- Updated dashboard/header UX, docs, AMD recipe, and AMD guide/blog coverage.

## Proposal Completion Audit

Completed first-scope items:

- Public config: `global.router.learning`, `learning.adaptation`, `learning.protection`, and decision `adaptations` are implemented and validated.
- Config cleanup: old public learning names such as `session_aware`, `lookup_tables`, `bandit`, `elo`, `rl_driven`, `gmtrouter`, and `personalization` are rejected rather than translated.
- Runtime composition: routing now uses a typed pipeline of recipe decision -> protection preflight -> adaptation proposal -> protection switch decision.
- Strategy registration: `routing_sampling` is registered through a typed adaptation strategy seam.
- Adaptation: `routing_sampling` supports `decision`, `tier`, and `global` candidate sets.
- Protection: conversation and session scopes support sampling preflight, switch guard, and deterministic rescue.
- State: local typed protection state and local typed model experience are implemented.
- Outcomes: `POST /v1/router/outcomes` links typed outcomes to replay and updates model-targeted experience.
- Observability: compact `x-vsr-learning-*` headers and typed replay diagnostics are emitted.
- Offline loop: `vllm-sr eval recipe-learning` produces metrics, findings, candidate recipes/patches, and optional seed-pack artifacts.
- Validation: unit/config/replay/dashboard/CLI gates and AMD recipe validation have been run.

Explicitly still out of scope, matching the proposal:

- automatic recipe promotion
- multi-replica state sharing
- synchronous external hot-path state lookup
- dashboard UX for recipe-learning artifacts
- contextual feature-model training on the request path
- additional adaptation strategies beyond `routing_sampling`
- compatibility aliases for old learning config names

## Validation

Local:

[...]

Issue #2244, opened 2026-06-18
Expand Router Learning architecture evals


Xunzhuo:
Parent: #2238
Deterministic local baseline: #2258

## Goal

Expand the #2258 deterministic Router Learning eval into reusable CI/release profiles that measure routing quality, stability, cache behavior, and cost.

## What #2258 Adds

- `bench/agentic_routing_experiment.py --learning-architecture`.
- Deterministic scenarios covering session-aware, bandit, Elo, and personalization.
- Metrics for route correctness, replay explainability, cost savings, cache token delta, unnecessary switches, and overhead.
- Unit tests for the eval harness.

## Remaining Work

- Turn the deterministic script into named reusable profiles for PR, release, and recipe validation.
- Add live-router replay mode that reads real Router Replay records instead of only synthetic scenarios.
- Add fixtures for session-scope and conversation-scope protection, tool-loop pinning, identity missing, idle reset, privacy/domain boundaries, and cache evidence.
- Produce stable JSON and concise human-readable summaries that can be attached to PRs or release notes.
- Define pass/fail thresholds for route correctness, explainability coverage, switch rate, cache/cost regression, and latency overhead.
- Coordinate Inferoa display separately; Semantic Router owns routing/replay semantics and eval outputs.

## Metrics

Reports should include:

- expected route family versus actual route family,
- base model, final model, and whether learning changed the route,
- adaptation method, mode, action, reason, and scope,
- state and experience status where relevant,
- unnecessary switch rate within a conversation,
- session-scope versus conversation-scope protection behavior,
- prompt tokens, completion tokens, cached tokens, cache hit, and cache gap,
- estimated base-only cost versus Router Learning cost,
- estimated cost saving or regression including prefix-cache effects,
- learning overhead on routing latency,
- replay explainability coverage for action, reason, scope, base model, final model, state, experience, and cache evidence.

## Acceptance Criteria

- Eval profiles can run deterministic and replay-backed multi-run agent sessions.
- Reports identify reasonable and unreasonable routing choices.
- Reports quantify cost/cache impact and learning latency overhead.
- Results are usable in PR review and release notes without manual log inspection.

Issue #2243, opened 2026-06-18
[Feature] Define distributed Router Learning state semantics


Xunzhuo:
### Goal

Define Router Learning state behavior for multi-replica deployments after the single-replica local-state model in #2258.

### Proposed Workgroup

MoM & Routing

### Scope

Implement the bounded behavior described by this issue while preserving its stated constraints.

#### Preserved proposal context

Parent: #2238
Single-replica baseline: #2258

##### What #2258 Assumes

- Request-time adaptation state is local to one router process.
- Replay is an event log, not a synchronous request-time state lookup.
- Experience should be snapshot-style and loaded locally before it is used.
- Missing state/experience degrades safely instead of failing the request.

##### Modes To Design

1. **Single replica**: all live state is local; simplest and lowest latency.
2. **Sticky multi-replica**: clients or infrastructure route the same session/conversation/user to the same router replica.
3. **Shared experience + local state**: replicas periodically load shared snapshots while hot state remains local.
4. **Shared hot state**: optional future mode only if latency and failure behavior are acceptable.

##### Questions To Resolve

- Which identities should sticky routing use: session, conversation, tenant, user, workspace, provider, or configured identity keys?
- Which state is local-only, eventually consistent, or strongly consistent?
- How should replicas recover useful state after restart?
- How should Replay events reconcile with state and experience snapshots?
- What diagnostics should show state misses, stale experience, non-sticky routing, or degraded mode?
- What Kubernetes/Envoy guidance is needed for sticky routing?

### Non-goals

- No required synchronous external storage call on every routed request.
- Replay writes can be durable and asynchronous.
- Raw identity values should not leak into public diagnostics.
- Missing or stale distributed state should degrade to base routing, local state, or observe mode.

### Completion signal

- A design covers single-replica, sticky multi-replica, shared-experience/local-state, and shared-hot-state modes.
- Latency and failure-mode tradeoffs are explicit.
- Kubernetes deployment guidance explains when sticky routing is required.
- Diagnostics explain missing, stale, unavailable, or non-sticky state behavior.

Sanchar127:
Hi @Xunzhuo, I’d like to work on this issue.
I’ll review #2258 and the related snapshot/replay work, then propose a design covering the four deployment modes before implementing it. Please let me know if this approach aligns with your direction for #2243.

Sanchar127:
Thanks for the clarification, and sorry for the delayed follow-up.

[...]

Issue #2242, opened 2026-06-18
[Cleanup] Finish Router Learning migration and reject legacy learning config paths


Xunzhuo:
Parent: #2238
Related: #2239, #2241, #2308, #2346
Day0 implementation baseline: #2258

## Goal

Finish the migration from legacy stateful learning-style selector paths into the Router Learning adaptation model, and make legacy public config paths fail with clear actionable errors.

This issue is intentionally scoped to migration cleanup and compatibility boundaries. New day2 learning algorithms are tracked separately by #2308 and #2346. Adaptation composition is tracked by #2241. Runtime contract hardening is tracked by #2239.

## Audited current-main baseline (2026-07-15)

- Current `main` accepts the clean `global.router.learning` contract and rejects legacy public method names through config validation.
- Legacy selector implementation files such as `pkg/selection/elo.go`, `rl_driven.go`, `gmtrouter.go`, and lookup-table code still exist for cleanup/internal migration analysis.
- Their presence is not evidence that those names remain supported public Router Learning APIs.

## Motivation

Router Learning should own stateful cross-request behavior through one documented adaptation runtime and config surface. Leaving older public paths around under `decision.algorithm`, legacy RL router names, or private per-algorithm state stores would make the current roadmap harder to reason about and harder to validate.

## Scope

- Audit docs, examples, recipes, CLI schema, dashboard surfaces, and tests for legacy public learning paths.
- Reject legacy public config paths with actionable validation errors instead of silently rewriting them.
- Confirm stateful learning behavior is configured under Router Learning adaptations.
- Decide whether old GMTRouter/Elo/RL-driven implementation pieces should be deleted, wrapped as internal helpers, or documented as legacy internals.
- Keep migration diagnostics clear in replay/debug output.

## Non-goals

- Do not add new learning algorithms here; use #2308 or follow-up algorithm issues.
- Do not redesign adaptation composition here; use #2241.
- Do not change the accepted Router Learning runtime contract without coordinating with #2239.

## Acceptance criteria

- Stateful learning behavior is owned by Router Learning state/adaptations, not public base algorithms.
- Legacy public config paths are rejected with clear validation errors.
- Docs and examples no longer imply stateful learning belongs under `decision.algorithm`.
- Any remaining legacy implementation code is either removed, wrapped as internal-only, or documented with a removal plan.
- Tests cover at least one legacy config rejection path.

## Validation entrypoint

```bash
[...]

PR #2307, opened 2026-07-01
[Feature] Add LinUCB and Linear Thompson routing strategies

### LINKED ISSUE #2308: [Research] Benchmark contextual Thompson and LinUCB Router Learning strategies
## Goal

Benchmark LinUCB and contextual Thompson Sampling as possible future values of the existing Router Learning `adaptation.strategy` contract, and graduate one only if it demonstrates a reliable benefit within production limits.

## Proposed Workgroup

MoM & Routing

## Background

On current main, `routing_sampling` is the supported day-zero strategy and the typed strategy registry is the intended extension seam. There is no maintained LinUCB or contextual Thompson implementation. This issue is therefore an evidence task first, not an assumption that either method should ship.

Related benchmark protocol: #2346. Related implementation exploration may be coordinated with #2307.

## Scope and approach

1. Define a shared contextual observation, action, propensity, and outcome schema aligned with #2239 and #2240.
2. Compare static/base selection, `routing_sampling`, and a simple non-contextual Beta-Bernoulli Thompson baseline before adding contextual complexity.
3. Evaluate LinUCB and contextual Thompson on byte-identical stationary, drifting, cold-start, sparse/delayed/censored feedback, low-trust/poisoned outcome, restart, and bounded-state scenarios.
4. Start with typed, content-minimized features and justify any learned embedding or hashing choice.
5. Report task quality, cost, latency, switches, protection suppressions, calibration, uncertainty, state size, update cost, and hot-path overhead.
6. Test conditioning, regularization, numerical stability, determinism, and failure recovery before choosing a matrix implementation.
7. If a strategy graduates, add it through the existing candidate, protection, outcome, Replay, and `adaptation.strategy` contracts, beginning in observe mode.

## Non-goals

- Add top-level method-specific public configuration or revive legacy selector state.
- Require raw prompts in the default feature projection.
- Let missing or invalid learning state fail the request or bypass hard authorization, safety, residency, context, candidate, or protection constraints.
- Treat benchmark improvement without uncertainty and matched budgets as production evidence.

## Completion criteria

- [ ] Data, configs, random seeds, and simple baselines are reproducible.
- [ ] Feature trust, feedback coverage, privacy, delay, drift, and missing-data semantics are explicit.
- [ ] Numerical stability and bounded state growth are tested.
- [ ] Results report uncertainty, negative results, and conditions where contextual methods provide no benefit.
- [ ] A production candidate meets latency/allocation limits and emits typed Replay diagnostics.
- [ ] Observe/shadow evidence precedes any canary or apply proposal.

## Research-to-production boundary

This issue owns reproducible matched-budget evidence and a graduate, iterate, or stop decision. It must not be expanded to make an existing production implementation fit. Any LinUCB or contextual Thompson runtime, public configuration, or apply-mode behavior requires a separate accepted implementation leaf after the evidence gate. An active pull request may relate to this research issue, but it should not close it while mixing ungraduated production code with the benchmark deliverable.

haowu1234:
Closes #2308

## Purpose

This PR does two related things and asks for review as one unit because the second is only meaningful with the first:

1. **Router**: register the two contextual bandit strategies that #2238 §\"Confirmed API Direction\" commits to as day-0 defaults but that are not actually implemented on `main` today — `linucb` and `linear_thompson` — as parallel entries under `global.router.learning.adaptation.strategy`.
2. **Bench**: add a synthetic, pure-Python MAB comparison harness (`bench/mab_comparison/`) that produces the academic MAB metrics the codebase currently lacks — cumulative regret, optimal-arm rate, drift-recovery time — and applies a pre-registered KEEP/KILL rule using paired bootstrap CIs, so \"is algorithm X better than routing_sampling?\" has an evidence-backed answer rather than a fixture pass/fail.

Complementary to (not overlapping with) #2244 / #2275 fixture-based Router Learning eval.

Related issues:
- **Partially addresses #1140** (P1, good-first-issue) — implements the phase 3 \"contextual bandit with UCB exploration\" part; phases 1, 2, 4, 5 (reward extractor, context builder, signal optimizer, monitoring) remain out of scope for this PR.
- **Partially addresses #2242** (day2 algorithms) — adds `linucb` and `linear_thompson`. The remaining day2 variants (discounted / sliding-window / budgeted / dueling / EXP3) are separate follow-ups.
- **Complementary to #2244 / PR #2275** — that PR gates the fixture-based deterministic eval; this harness covers the orthogonal academic-metric axis.

Note on #2238: the roadmap body says #2258 shipped these as the day-0 default. On the merged code (`src/semantic-router/pkg/extproc/router_learning_strategy.go:53` — the strategy registry contains only `routingSamplingAdaptationStrategy{}`) that is not the case yet. This PR closes the gap.

### Router (PR-2 half)

A shared matrix backend (`router_learning_matrix.go`) with:
- ridge-regularised (A, b) state per arm
- Sherman-Morrison rank-1 inverse update so per-request scoring stays O(d²)
- Cholesky factor + triangular solve for multivariate-normal sampling and as a numerical fallback when Sherman-Morrison declines

A shared feature extractor (`router_learning_features.go`): 8 hand-rolled signals (length, question form, code/math hints, decision/session presence, lowercase ratio) plus hashed token-bigram bag-of-tokens with L2 normalisation. Deliberately does not go through the mmBERT embedding subsystem — that would expand the PR scope across two subsystems for marginal benefit at d=64.

Two strategies share both, differing only in the score formula:
- **LinUCB** (Li et al. 2010): `score = θ·x + α·sqrt(x^T A^{-1} x)`
- **Linear Thompson** (Agrawal & Goyal 2013): `score = θ_tilde·x`, `θ_tilde ~ N(θ, σ² A^{-1})`

Both keep the routing_sampling penalty/bonus block (cost, overuse, reliability, latency, cache) so cost behaviour stays comparable across strategies.

Chosen as a **custom matrix backend rather than pulling Gonum in** because `pkg/selection/` already hand-rolls linear algebra everywhere (elo.go, automix.go, gmtrouter.go), the operations required are ~200 lines total, and adding a scientific-computing dependency for two adapters expands the review surface disproportionately at d=64.

Config surface follows #2238 §\"Confirmed API Direction\" — `dim` / `alpha` / `lambda` / `sigma` all typed with `EffectiveX` accessors; validator bounds them; decision-level overrides restricted to `mode` / `candidate_set` per policy.

### Bench (PR-1 half)

Follows the multi-arm eval pattern PR #2269 introduced (`bench/grounded_fusion/compare_multiarm.py`): pre-registered KEEP/KILL decision rule + paired bootstrap CI + byte-identical inputs across arms (via SHA-256 hashing of the reward + context sequence).

Six workloads: `stationary_3arm`, `stationary_5arm_close`, `drift_at_t5000`, `gradual_drift`, `contextual_2cluster`, `cold_start_with_prior`.

Verdict rule (pre-registered before any data was collected):

| Verdict | Condition |
| --- | --- |
| KEEP_X | X wins on every workload (CI excludes 0) |
| KEEP_X_CONTEXTUAL_ONLY | X wins only on contextual_2cluster; ties elsewhere |
| KILL_X | X significantly worse on any workload |
| INCONCLUSIVE | otherwise |

## Evidence

Ran the harness at N=30 seeds on all 5 registered algorithms (routing_sampling_py, ucb1, epsilon_greedy, linucb_py, linear_thompson_py). Full JSON verdict lives in the bench output; the decisive comparisons are:

**routing_sampling_py vs linucb_py**:
- `contextual_2cluster`: `regret_delta = +3478.67`, CI 95% = [+3451.5, +3505.0] (excludes 0), optimal-arm rate 0.34 → 0.99
- Non-contextual workloads (stationary / gradual_drift / cold_start): CI includes 0 in every case (ties)
- Overall verdict: **KILL_ROUTING_SAMPLING_PY**

**routing_sampling_py vs linear_thompson_py**: same pattern (KILL, decisive on contextual, tie elsewhere).

**linucb_py vs linear_thompson_py**: **INCONCLUSIVE** — the expected result for two algorithms that share the same posterior and differ only in the exploration mechanism.

The tie on non-contextual workloads is the important negative evidence: LinUCB *should not* dominate routing_sampling in a context-free setting; if it did, the implementation would be adding a bias rather than genuinely using context.

## Test Plan

Local:
- `python3 -m pytest bench/test_mab_comparison.py -q` → 20/20 pass
- `CGO_ENABLED=1 go vet ./pkg/extproc/ ./pkg/config/` → clean
- `CGO_ENABLED=1 go build ./pkg/extproc/ ./pkg/config/` → clean
- `make bench-mab-comparison-quick` → smoke run + verdict JSON emitted
- `make bench-mab-comparison SEEDS=30` → full 5-algorithm comparison, ~14 min pure-Python

The matrix-only tests (`router_learning_matrix_test.go`) were verified in an isolated Go module (extracted to bypass the cgo cluster locally) — all 5 pass, including the 50-iteration Sherman-Morrison vs full-Cholesky-recompute parity check (drift < 1e-7) and the 5000-sample multivariate-normal mean check.

Not run locally: `make test-semantic-router` and the fixture-based `make bench-router-learning PROFILE=pr` gate — both require the Rust bindings artefacts (`make build`), which take 20-40 min on this machine. Requesting CI to cover those.

## Test Result

- Bench harness (Python): 20/20 tests pass, verdict JSON reproducibly matches on rerun (workload SHA-256 identical across seeds).
- Router matrix backend (isolated Go module): 5/5 pass, no numerical drift over 50 rank-1 updates.
- `go vet` and `go build` for the two packages that this PR modifies: clean.
- 5-algorithm comparison at N=30 seeds produces the pre-registered KEEP/KILL verdict pattern above; the contextual gap is decisive (CI excludes 0 by an order of magnitude relative to the effect size on non-contextual workloads).

## What is deliberately not in this PR

- Additional day2 bandit variants from #2242 (`discounted_linucb`, `sliding_window_linucb`, `budgeted_linucb`, `exp3`, dueling) — separate follow-up PRs.
- Wiring the LinUCB feature extractor to the mmBERT embedding subsystem — a separate PR once the base algorithms are in.
- Distributed state semantics — tracked in #2243.
- E2E `e2e/testcases/router_learning_linucb.go` — needs a real cluster and can be added incrementally after the algorithm is landed.
- Adding this harness to any CI gate — the fixture-based `make bench-router-learning` gate remains the production-critical path (per PR #2275); this harness is complementary academic evidence, not a release blocker.

<details>

- [x] PR title uses module-aligned prefixes (`[Router]`, `[Bench]`)
- [x] The PR spans two modules; the title includes both

</details>


Xunzhuo (changes requested):
The learning state is not safely synchronized: `arm()` returns the mutable matrix after releasing its map lock, and request scoring reads that matrix outside the outcome-update lock. Concurrent scoring and outcome updates can race on theta/inverse data. Please keep snapshot and update operations behind state-owned locking and add a race regression.

FAUST-BENCHOU (commented):
Thanks for adding LinUCB and Linear Thompson strategies along with the MAB evaluation harness. Adding algorithm-level evaluation metrics provides useful evidence for comparing routing strategies beyond fixture-based tests.

PR #2312, opened 2026-07-02
fix(guard): use kv_cache_snapshot/restore instead of re-prefilling prefix per request

### LINKED ISSUE #2311: Qwen3Guard fast path re-prefills fixed prefix on every request (~640ms wasted/req on CPU)
## Problem

`generate_with_prefix_cache` calls `clear_kv_cache()` + `process_prefix()` on every single request, re-computing a full transformer forward pass over the fixed prefix template (~200 tokens). Since the prefix is a fixed string and `temperature=0.0`, every `process_prefix()` call produces byte-identical KV cache state — the computation is pure waste.

## Evidence

The `kv_cache_snapshot()` / `kv_cache_restore()` methods already exist on `ModelForCausalLM` (`qwen3_with_lora.rs:509,514`) and are used by `qwen3_multi_lora_classifier.rs:1088,1118` for the exact same optimization (prefill once, restore per-category). They are simply not wired into the guard path.

Benchmark on real Qwen3Guard-Gen-0.6B (CPU, F32):

| | Before | After |
|---|---|---|
| Prefill per request | 640ms | ~0.05ms |
| End-to-end per request | 2620ms | 2182ms |
| Speedup | 1x | 1.2x |

Determinism verified: 3 consecutive `process_prefix()` calls produce byte-identical logits (151,936 values match exactly).

## Fix

Pre-compute the KV state once in `initialize_prefix_caches()` via `process_prefix()` + `kv_cache_snapshot()`, then restore per-request via `kv_cache_restore()`. Memory cost: ~109MB (GPU) / ~218MB (CPU), ~9% of model weights.

haowu1234:
## Summary

Fixes #2311

The Qwen3Guard fast path re-prefills the fixed prefix template on every request. Since the prefix is a fixed string with `temperature=0.0`, each `process_prefix()` call produces byte-identical KV cache state. This PR pre-computes the KV state once at init and restores it per-request.

## Changes

- **`qwen3_guard.rs`**: In `initialize_prefix_caches()`, after tokenizing each prefix, run `process_prefix()` once and save the result via `kv_cache_snapshot()`.
- **`qwen3_guard_generation.rs`**: Replace `clear_kv_cache()` + `process_prefix()` with `kv_cache_restore()` in `generate_with_prefix_cache()`.

## Why

`kv_cache_snapshot()` / `kv_cache_restore()` already exist on `ModelForCausalLM` (`qwen3_with_lora.rs:509,514`) and are already used by `qwen3_multi_lora_classifier.rs:1088,1118` for the same optimization pattern. This PR wires them into the guard path.

## Benchmark

Real Qwen3Guard-Gen-0.6B, CPU F32:

| | Before | After |
|---|---|---|
| Prefill per request | 640ms | ~0.05ms |
| End-to-end per request | 2620ms | 2182ms |
| Speedup | 1x | 1.2x |

Memory cost: ~109MB (GPU BF16) / ~218MB (CPU F32), ~9% of model weights, allocated once at init.

## Determinism

Verified: 3 consecutive `process_prefix()` calls produce byte-identical logits (151,936 values match exactly). `kv_cache_restore()` is mathematically equivalent to `clear_kv_cache()` + `process_prefix()`.


WUKUNTAI-0211 (approved):
Implementation is correct — mirrors the `kv_cache_snapshot`/`restore` pattern already shipping in `qwen3_multi_lora_classifier.rs:1088,1118`; fixed prefix at `temperature=0` makes restore equivalent to `clear + process_prefix`, and the PR's byte-identical logits check confirms it. CI green.

Note: the snapshot is default-on (`prefix_cache.enabled` defaults `true`) and costs ~218MB CPU F32 / ~109MB GPU BF16 — ~9% on top of already-loaded model weights, a fair trade for dropping the 640ms/req prefill.

LGTM.

PR #2314, opened 2026-07-02
[Router] Rebuild the hybrid semantic cache HNSW index after eviction

### LINKED ISSUE #2313: bug: [Router] HybridCache eviction wipes the HNSW index without rebuilding — semantic cache silently degrades to ~100% miss once full
### Describe the bug

When the hybrid semantic cache (`cache_backend: hybrid`, in-memory HNSW + Milvus) reaches `max_memory_entries`, it silently stops returning cache hits.

`HybridCache.evictOneUnsafe()` (`src/semantic-router/pkg/cache/hybrid_cache.go`) does a FIFO eviction: it drops entry `0`, shifts every remaining entry index down by one, and then calls `h.hnswIndex.markStale()`, which **clears the entire HNSW graph** (`nodes=[]`, `nodeIndex={}`, `entryPoint=-1`).

The problem: unlike `InMemoryCache`, `HybridCache` has **no "needs rebuild" flag and no rebuild step**. `markStale()` is documented as *"needs rebuild with new indices,"* but nothing ever rebuilds. After an eviction, the add path only appends the single new node, so the in-memory graph is left holding **~1 node** while `embeddings`/`idMap` still hold up to `max_memory_entries` vectors.

Because `searchKNNHybridWithThreshold` traverses only the HNSW graph, every previously-cached entry becomes unreachable. And since the cache is now permanently at capacity, every subsequent insert re-triggers eviction, so the graph perpetually oscillates near a single node.

For comparison, `InMemoryCache` handles this correctly: `evictOne()` sets `hnswNeedsRebuild = true` and the next `addEntryToHNSWIndex`/search lazily calls `rebuildHNSWIndex()`. `HybridCache` is missing that half.

Note the background reclaim (`RebuildFromMilvus`) only re-queries Milvus on a ticker and is gated on `ttl_seconds > 0`, so with TTL disabled the in-memory index stays degraded indefinitely.

### To Reproduce

1. Configure the `hybrid` cache backend with a small `max_memory_entries` (e.g. `5`).
2. Insert more distinct entries than `max_memory_entries` — each `AddEntry`/`AddPendingRequest` past capacity triggers `evictOneUnsafe`.
3. Query for any entry that is still in memory but was **not** the most-recently inserted one.
4. It misses, even though the entry is present in both `embeddings` and Milvus.

Verified with a package-level reproduction that drives the exact in-memory insert path: after inserting 20 distinct entries into a cap-5 cache, the HNSW graph holds **only 1 node** for **5** retained embeddings, and a search for the oldest surviving entry returns **no match**.

### Expected behavior

After eviction, the HNSW graph should continue to index every retained in-memory embedding, so previously-cached (not yet evicted) entries stay findable and the hybrid cache keeps serving hits while at capacity.

### Affected layer

global level

### Additional context

Root cause is `evictOneUnsafe` marking the graph stale without a deferred rebuild. The fix mirrors the existing `InMemoryCache` pattern (a `hnswNeedsRebuild` flag set on eviction + a lazy rebuild from the in-memory embeddings on the next add).

Separately, rebuilding the whole graph on each insert-at-capacity is O(n) — this is an existing characteristic of `InMemoryCache` too; true incremental HNSW node deletion is a larger, separate optimization worth its own issue.

I have a fix + regression test ready and will open a linked PR.

glitch-ux:
Closes #2313

## Purpose

- **What:** Fixes a correctness bug where `HybridCache` eviction cleared the in-memory HNSW graph but never rebuilt it, so the hybrid semantic cache silently degraded to a ~100% miss rate once it reached `max_memory_entries`.
- **Why:** `evictOneUnsafe()` called `h.hnswIndex.markStale()` (which wipes the graph) with no rebuild, so only the most-recently-added node remained. Retained entries became unreachable via `searchKNNHybridWithThreshold` even though they were still in `embeddings`/Milvus, and because the cache stays at capacity, every subsequent insert re-triggered eviction and kept the graph near a single node.
- **How:** Mirror the existing `InMemoryCache` pattern — add an `hnswNeedsRebuild` flag set on eviction, and rebuild the graph from all in-memory `embeddings` on the next add (`addNodeHybridOrRebuild` → `rebuildHNSWFromEmbeddings`). An evict+append pair rebuilds once. `RebuildFromMilvus` clears the flag since it builds a fresh index from scratch.
- **Module(s):** `Router`

## Test Plan

- New regression test `src/semantic-router/pkg/cache/hybrid_cache_eviction_test.go`:
  - `TestHybridCacheEvictionRebuildsHNSWIndex` — inserts 20 distinct entries into a cap-5 cache and asserts the HNSW node count equals the retained embedding count and that every retained entry is still findable via the HNSW search path.
  - `TestHybridCacheDeferredRebuildMechanics` — pins the evict → mark-stale → rebuild-on-next-add handshake and the flag lifecycle.
- Commands (CGO + candle lib):
  ```
  go test ./pkg/cache/ -run 'Hybrid|HNSW|InMemory|Evict|Reclaim'
  go test ./pkg/cache/ -run 'Hybrid|Evict' -race
  go vet ./pkg/cache/ && gofmt -l pkg/cache/hybrid_cache.go pkg/cache/hybrid_cache_eviction_test.go
  ```
- The tests drive the real in-memory insert/evict/search path (no live Milvus needed, since the eviction and HNSW code touch only in-memory structures).

## Test Result

- New tests pass; existing hybrid / HNSW / in-memory / reclaim tests pass, including under `-race`. `go vet` and `gofmt` are clean.
- Confirmed the regression test fails against the pre-fix behavior: 5 retained embeddings → **1** HNSW node, and a search for the oldest surviving entry returns **no match**. After the fix: 5 embeddings → 5 nodes, entry found.
- **Follow-up (out of scope):** the graph rebuild is O(n) per insert once the cache is at capacity — this is an existing characteristic of `InMemoryCache` as well; true incremental HNSW node deletion would remove that cost and is worth a separate issue.

---
<details>

</details>


WUKUNTAI-0211 (approved):
Confirmed the bug and fix. `evictOneUnsafe` clears the whole HNSW graph via `markStale()` but nothing rebuilt it, leaving only the last node — and since an HNSW miss doesn't fall back to a Milvus brute-force scan (`hybrid_cache_lookup.go:75`→`:109` only does `GetByID` on an existing hit), the cache degraded to ~100% misses at capacity. The deferred-rebuild flag faithfully mirrors the `InMemoryCache` pattern, lock discipline is correct (evict+rebuild under write lock, search under RLock — no visible stale window), and the reclaim path rebuilds fully. Both regression tests are on point.

One nit (non-blocking): `hybrid_cache_eviction_test.go` hand-rolls `itoaForTest` to avoid importing `strconv` — just use `strconv.Itoa`, it's idiomatic in tests.

LGTM.

PR #2320, opened 2026-07-03
[Feature] Forward per-request authorization headers to backends

### LINKED ISSUE #2286: [Feature] Forward a deployment-injected credential only to trusted upstream gateways
### Goal

Allow a deployment authentication layer to inject a dedicated credential for an explicitly trusted upstream gateway, while preventing the Router from forwarding the client's raw Authorization header or owning tenant, quota, budget, or spend policy.

Parent Epic: #1138

### Proposed Workgroup

Data Plane & Networking

### Current-main boundary

Current main can consume credentials produced by an external authorization service and strips dedicated credential headers to prevent leakage. It does not expose a safe generic contract for copying the inbound Authorization value to an upstream.

The credential used to authenticate the client to the deployment and the credential intended for a trusted downstream gateway are separate trust objects.

### Scope

- Define a dedicated authentication-context field or header for the downstream gateway credential.
- Accept it only from a trusted deployment authentication result; sanitize any client-supplied copy before policy evaluation.
- Forward it only when the selected backend profile explicitly identifies an allowlisted trusted gateway.
- Apply the same resolution to direct requests and bounded internal model calls.
- Fail closed when a route requires the dedicated credential and trusted authentication did not provide it.
- Redact the credential from logs, traces, metrics, Router Replay, errors, and projected configuration.
- Leave caller attribution and any tenant, quota, budget, spend, or rate-limit policy to the trusted gateway.

### Non-goals

- Forward the raw client Authorization header.
- Build organization, virtual-key, quota, budget, billing, or rate-limit products in Semantic Router.
- Let request input select the trust target or enable credential forwarding.
- Forward deployment credentials to arbitrary model-provider endpoints.
- Fall back silently to a shared credential when a dedicated credential is required.

### Completion signal

- [ ] Client-controlled credential copies are removed before trust evaluation.
- [ ] Only a trusted authentication result can provide the dedicated gateway credential.
- [ ] Only explicitly trusted gateway profiles can receive it.
- [ ] Direct and bounded internal calls resolve the same credential contract.
- [ ] Missing or invalid credentials fail before upstream dispatch.
- [ ] Redirects, logs, diagnostics, Replay, and errors cannot leak the credential.
- [ ] E2E coverage proves trusted forwarding, untrusted rejection, sanitization, and default-disabled behavior.

blouargant:
Closes #2286

## Purpose

Adds an opt-in, per-backend `forward_authorization_header` field on `providers.models[].backend_refs[]`. When set to `true`, the router forwards the caller's inbound `Authorization` header verbatim to that upstream instead of injecting a static service key, so per-user identity (for example LiteLLM virtual keys) reaches the gateway for spend attribution and budget enforcement.

## Behavior

- Forwards the caller's `Authorization` verbatim to the selected backend; `api_key`/`api_key_env` are ignored for that backend.
- When a forward backend is selected without an inbound `Authorization`, the router returns `401` at that leg. Enforcement is per selected backend, so a decision that mixes forward and static candidates is not rejected before backend selection.
- Default `false`; existing static-key behavior is unchanged.
- Applies to the direct routing path and to the looper (internal multi-model re-dispatch). Caller identity is carried across looper legs on a dedicated `x-vsr-inbound-authorization` header, kept separate from any static key.

## Internal-leg trust boundary

The looper re-dispatches requests back through Envoy into the same ExtProc. Internal markers and the caller-identity carrier are authenticated with a per-process secret (`x-vsr-looper-authorization`) verified with a constant-time compare. Requests that fail verification have all reserved internal headers stripped, so a client cannot spoof looper state or the caller-identity carrier. Reserved internal headers are removed at the header phase, so they never reach an upstream. `VSR_LOOPER_INTERNAL_AUTH_SECRET` can be set to a shared value for multi-replica topologies where a re-dispatch may land on a different process. `LooperConfig` rejects configuring any reserved internal header name.

## Config contract

- `forward_authorization_header` is added to the Go canonical loader and the Python CLI `BackendRef` schema, with matching validation on both sides.
- Combining `forward_authorization_header: true` with `api_format: anthropic` is rejected (the Anthropic-native path does not forward the caller Authorization), as is combining it with a static `api_key`/`api_key_env` on the same backend.
- Adds `global.router.strip_inbound_authorization` (default `false`): when enabled, strips the caller's inbound `Authorization` on the passthrough path for backends that neither inject a static key nor opt into forwarding. Off by default, so existing behavior is unchanged.

## Test Plan and Result

- Unit: `make test-semantic-router` and the Python CLI config tests. Covers the config contract (including the anthropic and static-key rejections), the direct forward path (overwrite, case-insensitive lookup, `401` on absent/blank), per-leg enforcement, the looper caller-identity carrier, and internal-leg trust (valid proof honored, spoofed headers stripped).
- E2E: a registered `forward-auth` profile (direct forward `200`/`401`, per-backend static-key `200`, spoofed reserved-header `401`), wired into the CI matrix and the agent harness inventory.
- Manual end-to-end through Envoy ExtProc to a live LiteLLM gateway, exercising a forward backend and a static-key backend.
- CI on this branch is green.

---
<details>

</details>


blouargant:
Hello,
I probably should have started with this, but as explained in feature request [#2286](https://github.com/vllm-project/semantic-router/issues/2286), I need to track LLM costs precisely for each user, hence the need for key passthrough using the users' own keys.
Since this is my first contribution to the project, I tried to keep the scope of the changes limited.

If anything needs to be revised or reworked, that’s no problem; I am open to suggestions.

Xunzhuo (commented):
Thanks for the thorough implementation and test notes. The per-user cost/accounting use case is valid, but after a fresh pass against current `main`, this PR still needs security and config-contract work before review can converge.

Please address these in order:

1. Rebase onto current `main` and resolve the existing conflicts.
2. Establish a trusted boundary for the internal looper credential carrier. Today `x-vsr-looper-request` and `x-vsr-inbound-authorization` are ordinary headers, and `LooperConfig.Headers` is copied before the carrier is conditionally set. A client or configured looper header can therefore spoof the internal path/caller identity. Please reject reserved headers in config, sanitize client-supplied internal headers at a trusted ingress (or authenticate the internal leg), and add spoofing tests.
3. Remove the decision-wide early 401. A mixed decision with one static-key backend and one forward-auth backend must not reject before the selected backend is known. Enforce the requirement at the selected backend/leg.
4. Complete the config contract: add `forward_authorization_header` to the Python CLI `BackendRef` schema and validation, and reject unsupported `api_format: anthropic` combinations until that path is implemented.
5. Split the unconditional “strip inbound Authorization by default” hardening into its own change (or explicitly scope, document, and E2E-test it). It changes behavior even when `forward_authorization_header` is false.
6. Add a repo-registered E2E case for direct routing, mixed static/forward candidates, looper redispatch, and spoofed reserved headers; the manual LiteLLM run is useful but is not the durable acceptance gate.
7. Remove the public tool-attribution footer from the PR body and update the Test Result after the rebase.

The key blocker is item 2; please settle that security boundary before spending time on conflict cleanup around it.

blouargant:
@Xunzhuo thanks for the review. Summary of the changes made in response:

- **Internal-leg trust boundary**: the looper now authenticates every internal request with a per-process secret (`x-vsr-looper-authorization`, constant-time compare). ExtProc honors looper markers and the caller-identity carrier only on an authenticated leg; otherwise all reserved internal headers are stripped. Reserved headers are removed at the header phase so they never reach an upstream, and `LooperConfig` rejects configuring any reserved header name. `VSR_LOOPER_INTERNAL_AUTH_SECRET` covers multi-replica topologies.
- **Enforcement moved per-leg**: removed the decision-wide early `401`. The requirement is now enforced at the selected backend, so a decision mixing forward and static candidates is not rejected before backend selection.
- **Config contract completed**: `forward_authorization_header` is now in the Python CLI `BackendRef` schema with matching validation, and `api_format: anthropic` combined with forwarding is rejected on both the Go and Python sides.
- **Strip-inbound-Authorization is now opt-in**: gated behind `global.router.strip_inbound_authorization` (default `false`), so default behavior is unchanged.
- **E2E**: added a registered `forward-auth` profile (direct forward `200`/`401`, per-backend static-key `200`, spoofed reserved-header `401`), wired into the CI matrix and the agent harness inventory.
- **PR body**: refreshed and the tool-attribution footer removed.

Looper-leg enforcement is covered by Go unit tests rather than the in-cluster E2E; the profile asserts the router-side behavior that a simulated backend can exercise.

e2e/testcases/forward_authorization.go
```diff
@@ -0,0 +1,199 @@
...
+		return false, fmt.Sprintf("send request: %v", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	respBody, _ := io.ReadAll(resp.Body)
+
+	if tc.expectRejected {
+		if resp.StatusCode == http.StatusUnauthorized {
+			return true, ""
+		}
+		return false, fmt.Sprintf("expected 401, got %d: %s", resp.StatusCode, truncate(respBody))
```
FAUST-BENCHOU, line 204: The E2E catches the missing-Authorization and spoofed-header cases, but the success case only checks that the simulated backend returns 200. Since the sim does not assert the upstream `Authorization` value, this would still pass if the router dropped the caller token or forwarded a static key on the 200 path. Could we make the acceptance gate verify the received `Authorization` value, or add a small echo/asserting backend for this profile, so the test covers the spend-attribution behavior from #2286 directly?

blouargant: @FAUST-BENCHOU indeed, there was a gap in the tests, fixed in 75c499f6.

The forward-auth backends now run the `mock-vllm` image, which echoes the received `Authorization` back on an `x-echo-authorization` response header. The success cases assert that value: the forward case checks the caller's token reached the upstream verbatim, and the static case checks the router-injected static key did. So the 200 path now fails if the caller token is dropped or a static key is forwarded instead.

The mock is built locally and loaded into kind via the profile's LocalImages capability, so no new image or CI wiring was added. **Integration Test [Kubernetes]** passes with the value assertions.

Xunzhuo (changes requested):
The earlier implementation work is substantial, and the per-leg branch, CLI field, Anthropic rejection, and reserved-header authentication are now present. A fresh security pass still finds merge blockers.

Full-duplex/header-only paths bypass the strip and selected-backend enforcement; the default HTTP client can redirect the custom caller-token and looper-proof headers to another host; weak shared secrets are accepted; and raw inbound Authorization remains the credential source while non-opted passthrough is still allowed by default. The E2E also does not exercise a real mixed decision or looper redispatch and checks only a substring. Current dashboard and Operator editing cannot preserve the new field.

Please rebase, close these trust-boundary gaps, require a trusted credential source and fail-closed secret policy, add exact-value redirect/full-duplex/mixed-decision/redispatch coverage, and complete dashboard/Operator config parity before another review.

Xunzhuo (commented):
Thanks for the substantial trust-boundary follow-up. One security blocker is still present: the looper uses the default redirect-following `http.Client` and adds both `x-vsr-inbound-authorization` and `x-vsr-internal-auth` before sending. A 307/308 redirect to another host therefore carries these custom headers to that host, exposing the caller credential and the process-local proof. Please fail closed on redirects (or explicitly scrub every sensitive/internal header before any allowed redirect) and add a cross-origin redirect regression. The branch also needs a rebase to resolve its current conflicts with `main`.

FAUST-BENCHOU (commented):
Thanks for adding per-request authorization forwarding support. The opt-in design and additional internal header validation help preserve existing behavior while enabling user identity propagation to upstream gateways.

PR #2321, opened 2026-07-04
[Bug] Return 503 when classification models are not ready

### LINKED ISSUE #2120: [Bug] Return typed client and not-ready errors from classification APIs

p172913:
Closes #2120 

## Purpose

This change improves error handling for classification endpoints when the required model is not available.

Added a new regression tests in src\semantic-router\pkg\apiserver\classify_empty_input_test.go to verify the new error handling behavior.

model not ready -- > returns 503 Service Unavailable
wrapped model not ready --> returns 503 Service Unavailable
generic runtime error --> 500 Internal Server Error

## Test Plan

Command to run: go test ./pkg/apiserver/ -run "TestClassify|TestWrite" -v
and
Command to run: go test -v -count=1 -run 'TestEmbeddingEndpoints' ./pkg/apiserver/ -tags=cgo

## Test Result

running Suite: VectorStore Suite - /mnt/e/Open-Source/vLLM/semantic-router/src/semantic-router/pkg/vectorstore

==============================================================================================================
Random Seed: 1784033959

Will run 310 of 310 specs
••••••••••SS•••••••••••••SSS•••SS••••••SS••SSSSSS•••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••2026-07-14T12:59:23.301282Z  WARN logger_core: connection creation - Failed connecting to 192.0.2.1:6379, due to timed out
••SSSS••••••••••••••••••••••••••••••••SS••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••SSSSSSSSSSSSS•••••••••••••••••••••

Ran 276 of 310 Specs in 4.547 seconds
SUCCESS! -- 276 Passed | 0 Failed | 0 Pending | 34 Skipped
--- PASS: TestVectorStore (4.56s)
PASS
ok      github.com/vllm-project/semantic-router/src/semantic-router/pkg/vectorstore     5.095s
make[1]: Leaving directory '/mnt/e/Open-Source/vLLM/semantic-router'

---
<details>

</details>


Xunzhuo:
PTAL /cc @theohsiung

theohsiung (commented):
Nice work on part (B) — this nails the part the issue was actually blocked on. A couple of notes before it can close #2120:

**What's solid**
- `ErrModelNotReady` as a typed sentinel + `errors.Is` mapping is exactly the "typed not-ready error" the issue asked for.
- Importantly, the readiness flag is checked *before* the inference call, so genuine `ClassifyPIIWithDetails` / `CheckForJailbreak` failures still map to 500. That avoids the "blanket-map to 503 masks real 500s" trap the issue explicitly called out. 👍
- `CLASSIFIER_NOT_READY` / 503 also lines up with the existing conventions (`NLI_MODEL_NOT_READY`, `VECTOR_STORE_DISABLED`, `MEMORY_NOT_AVAILABLE`).

**Gap vs. the #2120 scope**
Part (B) explicitly lists three endpoints, and embeddings is one of them:
```
POST /api/v1/embeddings  {"texts":["hi"]}  -> 500 EMBEDDING_GENERATION_FAILED ("status: -1")
```
This PR covers `/classify/pii` and `/classify/security`, but `pkg/apiserver/route_embeddings.go:27` still maps any `buildEmbeddingResults` error — including a not-loaded model — to 500 `EMBEDDING_GENERATION_FAILED`. So `Closes #2120` currently over-claims.

Could you either:
1. fold embeddings not-ready → 503 into this PR (same pattern: a typed not-ready error surfaced from the embedding layer, checked around generation), **or**
2. keep this PR scoped to the two classify endpoints, switch `Closes #2120` → `Part of #2120`, and we track embeddings as a follow-up?

Either works for me — I just want the issue's close-state to be accurate.

**Minor / optional**
- Worth a sanity-check on the disabled-but-configured case: readiness is set inside `initialize{PII,Jailbreak}Classifier`, which only runs when the signal is used in routing (`classifier_lifecycle.go:103-104`). For a classify-API-only deployment that enables PII purely for the API (no routing signal), the model won't init and callers get 503. That's arguably correct for "feature disabled" — just flagging it so it's a deliberate choice rather than a surprise.

Thanks for picking this up — the core design is right.

p172913:
> Nice work on part (B) — this nails the part the issue was actually blocked on. A couple of notes before it can close #2120:
> 
> **What's solid**
> 
> * `ErrModelNotReady` as a typed sentinel + `errors.Is` mapping is exactly the "typed not-ready error" the issue asked for.
> * Importantly, the readiness flag is checked _before_ the inference call, so genuine `ClassifyPIIWithDetails` / `CheckForJailbreak` failures still map to 500. That avoids the "blanket-map to 503 masks real 500s" trap the issue explicitly called out. 👍
> * `CLASSIFIER_NOT_READY` / 503 also lines up with the existing conventions (`NLI_MODEL_NOT_READY`, `VECTOR_STORE_DISABLED`, `MEMORY_NOT_AVAILABLE`).
> 
> **Gap vs. the #2120 scope** Part (B) explicitly lists three endpoints, and embeddings is one of them:
> 
> ```
> POST /api/v1/embeddings  {"texts":["hi"]}  -> 500 EMBEDDING_GENERATION_FAILED ("status: -1")
> ```
> 
> This PR covers `/classify/pii` and `/classify/security`, but `pkg/apiserver/route_embeddings.go:27` still maps any `buildEmbeddingResults` error — including a not-loaded model — to 500 `EMBEDDING_GENERATION_FAILED`. So `Closes #2120` currently over-claims.
> 
> Could you either:
> 
> 1. fold embeddings not-ready → 503 into this PR (same pattern: a typed not-ready error surfaced from the embedding layer, checked around generation), **or**
> 2. keep this PR scoped to the two classify endpoints, switch `Closes #2120` → `Part of #2120`, and we track embeddings as a follow-up?
> 
> Either works for me — I just want the issue's close-state to be accurate.
> 
> **Minor / optional**
> 
> * Worth a sanity-check on the disabled-but-configured case: readiness is set inside `initialize{PII,Jailbreak}Classifier`, which only runs when the signal is used in routing (`classifier_lifecycle.go:103-104`). For a classify-API-only deployment that enables PII purely for the API (no routing signal), the model won't init and callers get 503. That's arguably correct for "feature disabled" — just flagging it so it's a deliberate choice rather than a surprise.
> 
> Thanks for picking this up — the core design is right.

@theohsiung Thanks for the review.
The latest commit will make sure embedding endpoints has similar safety pattern as remaining two. And added a testcase in src\semantic-router\pkg\apiserver\route_embeddings_test.go

Xunzhuo (commented):
The typed `ErrModelNotReady` plus `errors.Is` mapping is the right shape, and this PR looks close once the branch is repaired.

Please do the following:

- Fix the malformed `route_embeddings_test.go`. Current CI reports missing structure before the tests at lines 268, 299, and 311, followed by `expected '}', found EOF` at line 324.
- Rebase onto current `main`; this test file has moved and should be resolved carefully rather than only adding the missing brace to the stale version.
- Add/extend a behavior-level E2E case covering 503 for PII, security, and the embedding endpoints. These are externally visible HTTP status/code changes, so package tests alone are not the completion boundary.
- Rename the PR to a module-aligned title such as `[Router][Bindings] Return 503 when classification models are not ready`.
- Rerun `make precommit-branch-gate`, `make test-semantic-router`, the minimal binding gate, and the affected E2E profile; record the actual results in the PR body.

I do not see a larger architecture blocker here. Once the syntax/rebase/E2E items are green, this should be ready for another review.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -62,6 +62,12 @@ func (s *ClassificationAPIServer) handleEmbeddings(w http.ResponseWriter, r *htt
 		return
 	}
 
+	if !candle_binding.IsEmbeddingReady() {
```
Xunzhuo: [P1] Readiness must be checked for the model and modalities requested, not through one global unified-text flag. `IsEmbeddingReady` becomes true only in `InitEmbeddingModels`, while image requests use the separately initialized multimodal model. A valid multimodal-only deployment is therefore rejected with 503; conversely, if a unified text model is ready but the multimodal model is not, an image request passes this gate and later returns 500. The same issue applies when a specific requested text model is absent but another one initialized. Expose per-family readiness and validate `req.Model`, `req.Texts`, and `req.Images` before inference.

e2e/testcases/service_readiness_503.go
```diff
@@ -0,0 +1,124 @@
...
+	"net/http"
+	"time"
+
+	"github.com/vllm-project/semantic-router/e2e/pkg/fixtures"
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+	"k8s.io/client-go/kubernetes"
+)
+
+func init() {
+	pkgtestcases.Register("model-not-ready-503", pkgtestcases.TestCase{
```
Xunzhuo, line 16: [P1] This testcase is registered but never selected: `model-not-ready-503` is absent from `e2e/pkg/testmatrix` and every profile testcase list, so the affected CI run cannot execute it. Wire it into a profile with a deterministic no-model configuration, or have the testcase establish and restore that configuration before asserting 503. Please also add the positive/partial-readiness cases so the new gate cannot regress working multimodal or model-specific requests.

Xunzhuo (changes requested):
The previous syntax/rebase/title work is fixed, the typed classifier error path is sound, and `make build-e2e` passes locally. However, this is still not safe to merge: the new embedding gate conflates readiness across model families and the added E2E testcase is not selected by any profile, so CI does not cover the new behavior. Please fix the two inline blockers and add request/model-aware tests for text-only, multimodal-only, and partially initialized configurations. Current required CI jobs are also still running; please rerun/confirm the affected profile after wiring the testcase.

theohsiung:
+1 to both of @Xunzhuo's P1s. Two small supplementary notes, non-blocking:

- **Follow-through on P1-a**: once embedding readiness becomes per-family, consider
  surfacing it the way classify already does — a typed error from the service layer
  mapped via `errors.Is` — instead of an inline `IsEmbeddingReady()` check in the
  handler. That keeps `handleEmbeddings` / `handleSimilarity` / `handleBatchSimilarity`
  uniform with `writeClassificationError` and avoids a second "not ready" shape.
- **Nit** — `route_embeddings_test.go` labels the batch case `/api/v1/batch-similarity`,
  but the registered route is `/api/v1/similarity/batch` (`routes.go:94`). Harmless today
  (the handler is invoked directly), just a misleading label.
- **Nit** — the PR now also adds `e2e/testcases/…`, so `[E2E]` could join
  `[Router][Bindings]` per the multi-module title checklist item.

p172913:
Hi @Xunzhuo Is there any update on this PR..??

theohsiung (commented):
Thanks for pushing on the two items from @Xunzhuo's review. Checked the current diff directly (not just the commit messages):

**Model-family readiness split (Xunzhuo's first blocker) — looks solid.** `checkEmbeddingReadiness` in `route_embeddings.go` correctly gates on request shape (`len(req.Texts) > 0` → text readiness, `len(req.Images) > 0` → multimodal readiness), and the new tests in `route_embeddings_test.go` cover all four partial-init combinations. This resolves the conflation issue.

**E2E coverage (Xunzhuo's second blocker) — not actually resolved.** The `no-model` profile and `model-not-ready-503` testcase are registered in Go and in `ci-changes.yml` / `tools/agent/e2e-profile-map.yaml`, but the file that actually decides which e2e profiles run per-PR — `.github/workflows/integration-test-k8s.yml` (`determine-profiles` job) — was never updated. Every other profile has an explicit line there, e.g.:
```bash
[[ "${{ needs.changes.outputs.e2e_kubernetes }}" == "true" ]] && profiles+=("kubernetes")
```
There's no `e2e_no_model` line, so this profile can never be selected by CI — the new readiness test will never run, regardless of how correct it is. `tools/agent/e2e-profile-map.yaml`'s `full_ci_profiles` is only consumed by the local agent-harness scripts (`agent_ci_validation.py` / `agent_context_resolution.py`), not by GitHub Actions, so adding `no-model` there doesn't wire it into CI either.

Could you add:
```bash
[[ "${{ needs.changes.outputs.e2e_no_model }}" == "true" ]] && profiles+=("no-model")
```
to `integration-test-k8s.yml`'s `determine-profiles` step? Until that's there, this is functionally the same gap Xunzhuo flagged on 7/15.

**Minor / optional:**
- The onnx-binding readiness changes (`ensureEmbeddingModelReady`, `embeddingReadyOverride`, `multiModalEmbeddingReady` in `onnx-binding/semantic-router.go`) don't appear to have any caller in `src/semantic-router/pkg/` — the embeddings API always uses `candle_binding`. Worth dropping that part to keep the diff scoped to what's actually exercised, unless there's a planned caller I'm missing.
- Branch has merge conflicts with `main` — will need a rebase before this can be merged regardless of the above.

p172913:
> Thanks for pushing on the two items from @Xunzhuo's review. Checked the current diff directly (not just the commit messages):
> 
> **Model-family readiness split (Xunzhuo's first blocker) — looks solid.** `checkEmbeddingReadiness` in `route_embeddings.go` correctly gates on request shape (`len(req.Texts) > 0` → text readiness, `len(req.Images) > 0` → multimodal readiness), and the new tests in `route_embeddings_test.go` cover all four partial-init combinations. This resolves the conflation issue.
> 
> **E2E coverage (Xunzhuo's second blocker) — not actually resolved.** The `no-model` profile and `model-not-ready-503` testcase are registered in Go and in `ci-changes.yml` / `tools/agent/e2e-profile-map.yaml`, but the file that actually decides which e2e profiles run per-PR — `.github/workflows/integration-test-k8s.yml` (`determine-profiles` job) — was never updated. Every other profile has an explicit line there, e.g.:
> ```bash
> [[ "${{ needs.changes.outputs.e2e_kubernetes }}" == "true" ]] && profiles+=("kubernetes")
> ```
> There's no `e2e_no_model` line, so this profile can never be selected by CI — the new readiness test will never run, regardless of how correct it is. `tools/agent/e2e-profile-map.yaml`'s `full_ci_profiles` is only consumed by the local agent-harness scripts (`agent_ci_validation.py` / `agent_context_resolution.py`), not by GitHub Actions, so adding `no-model` there doesn't wire it into CI either.
> 
> Could you add:
> ```bash
> [[ "${{ needs.changes.outputs.e2e_no_model }}" == "true" ]] && profiles+=("no-model")
> ```
> to `integration-test-k8s.yml`'s `determine-profiles` step? Until that's there, this is functionally the same gap Xunzhuo flagged on 7/15.
> 
> **Minor / optional:**
> - The onnx-binding readiness changes (`ensureEmbeddingModelReady`, `embeddingReadyOverride`, `multiModalEmbeddingReady` in `onnx-binding/semantic-router.go`) don't appear to have any caller in `src/semantic-router/pkg/` — the embeddings API always uses `candle_binding`. Worth dropping that part to keep the diff scoped to what's actually exercised, unless there's a planned caller I'm missing.
> - Branch has merge conflicts with `main` — will need a rebase before this can be merged regardless of the above.

I made changes as per your review. Please validate those changes...

theohsiung (commented):
Verified the latest commits — thanks for rebasing and adding the `e2e_no_model` line to `integration-test-k8s.yml`, and for resolving the merge conflicts.

**The line is correct but currently unreachable for this exact PR.** `determine-profiles` has an early-exit branch above the per-profile checks:
```bash
if [[ "${{ needs.changes.outputs.e2e_common }}" == "true" ]] || \
   [[ "${{ needs.changes.outputs.core }}" == "true" ]] || \
   ... ; then
  profiles=("kubernetes" "dashboard" "remote-embedding")
  ...
  exit 0
fi
# only reached if none of the above matched
[[ "${{ needs.changes.outputs.e2e_no_model }}" == "true" ]] && profiles+=("no-model")
```
`core` is defined as `src/**` + `candle-binding/**` (`ci-changes.yml`), and this PR's own changes (`route_classify.go`, `route_embeddings.go`, `classifier.go`, `candle-binding/semantic-router.go`) all fall under that. I pulled the actual `determine-profiles` job log from this PR's latest run and confirmed it:
```
Running default baseline profiles due to common/core changes or push/schedule/manual trigger
[[ "true" == "true" ]] && profiles+=("no-model")   # <- e2e_no_model was true, but never reached
```
Only `kubernetes`, `dashboard`, and `remote-embedding` ran — no `no-model`. So the readiness regression test this PR adds still doesn't execute in CI, and structurally it never will for any future PR that touches the classify/embeddings code the test is meant to guard, since that code is itself `core`.

To actually close this gap, `no-model` needs to be part of the **default baseline** set, not just the per-profile branch:
```bash
profiles=("kubernetes" "dashboard" "remote-embedding" "no-model")
```
(or an equivalent restructuring so `no-model` isn't gated behind the early exit). The `e2e_no_model`-only line can stay for the non-core-change case, but it won't be what actually protects this PR's own code path.

.github/workflows/integration-test-k8s.yml
```diff
@@ -66,6 +66,7 @@ jobs:
           [[ "${{ needs.changes.outputs.e2e_multi_endpoint }}" == "true" ]] && profiles+=("multi-endpoint")
           [[ "${{ needs.changes.outputs.e2e_authz_rbac }}" == "true" ]] && profiles+=("authz-rbac")
           [[ "${{ needs.changes.outputs.e2e_streaming }}" == "true" ]] && profiles+=("streaming")
+          [[ "${{ needs.changes.outputs.e2e_no_model }}" == "true" ]] && profiles+=("no-model")
```
theohsiung: This line is correct but unreachable for this PR (and for any future PR touching classify/embeddings core code). `determine-profiles` has an early-exit branch above this section: if `core` (`src/**`, `candle-binding/**` per `ci-changes.yml`), `docker`, `make`, or `ci` changed, it sets `profiles=("kubernetes" "dashboard" "remote-embedding")` and `exit 0` *before* reaching this line. This PR's own changes (`route_classify.go`, `route_embeddings.go`, `classifier.go`, `candle-binding/semantic-router.go`) are all under `src/**`, so `core == true` here.

I pulled the actual `determine-profiles` job log from this PR's latest run:
```
Running default baseline profiles due to common/core changes or push/schedule/manual trigger
[[ "true" == "true" ]] && profiles+=("no-model")   # e2e_no_model was true, but never reached
```
Only `kubernetes`/`dashboard`/`remote-embedding` ran — no `no-model`. The new readiness regression test still does not execute in CI.

To actually close this gap, `no-model` needs to be in the default baseline set above (line ~45: `profiles=("kubernetes" "dashboard" "remote-embedding")`), not gated behind this per-profile branch.

p172913: @theohsiung made changes as per your review.

Xunzhuo (changes requested):
The classifier sentinel path and no-model E2E reachability are now correct, and the no-model profile ran successfully. I still cannot approve this: embedding readiness remains global rather than per requested model. A factory initialized with only one family—or with a non-fatal Gemma load failure—passes readiness, after which an explicitly requested absent model returns generic 500 instead of typed 503.

Please rebase onto current main, derive readiness from actual per-family loaded state, add explicit-model tests, make the multimodal E2E request image-only, and document the 503 response in OpenAPI. The branch is currently conflicting, so all affected checks must be rerun after resolution.

theohsiung (changes requested):
Thanks for pushing through the last two rounds — the E2E CI-wiring gap from 8/1 is confirmed fixed (`e2e_no_model` is now wired into `integration-test-k8s.yml`'s `determine-profiles`, and `E2E / integration-test (no-model)` is green on this PR). Did a fuller pass (standards + spec-conformance) on top of that, a few things still need attention before this can merge:

**Not addressed from earlier rounds**

- **onnx-binding — still unresolved, and grew.** I flagged on 8/1 that `ensureEmbeddingModelReady`/`embeddingReadyOverride`/`multiModalEmbeddingReady` in `onnx-binding/semantic-router.go` have no caller in `src/semantic-router/pkg/` (the embeddings API only ever uses `candle_binding`). This round adds a `github.com/vllm-project/semantic-router/onnx-binding => ../../onnx-binding` replace directive to `src/semantic-router/go.mod` and an `onnx-binding` stub-module block to `dashboard/backend/Dockerfile`'s WASM stage — neither has a consumer. The real candle↔onnx swap already lives in the separate `src/semantic-router/go.onnx.mod` (which replaces `candle-binding => ../../onnx-binding` directly), so this new replace line in the main `go.mod` doesn't participate in that mechanism at all. Please drop the onnx-binding readiness code + the go.mod/Dockerfile additions, or explain the intended caller if there's a follow-up planned.
- **7/15 ask only half-done.** I asked for the embedding-readiness check to move fully into a typed error surfaced via `errors.Is` (matching `writeClassificationError`'s pattern), replacing the inline handler check. `classifyEmbeddingError` now does have the typed-error mapping, but `checkEmbeddingReadiness` (in `handleEmbeddings`) and the bare `candle_binding.IsEmbeddingFamilyReady(...)` calls in `handleSimilarity`/`handleBatchSimilarity` still short-circuit before the typed-error path runs. Two "not ready" shapes for the same condition — could you collapse to one?

**New finding: OpenAPI 503 contract is over-broad and contradicts the issue itself**

`routes_catalog.go` now marks `classify/intent`, `fact-check`, `user-feedback`, `combined`, `batch`, `eval`, and `nli` as `ServiceUnavailable: true`, and `openapi_spec_test.go`'s new `TestOpenAPISpecDocumentsModelNotReady503` locks in a 503 response for all of them. But this PR only added real readiness checks for `pii`/`security` (`IsPIIModelReady`/`IsJailbreakModelReady`) and the embeddings family — none of the other handlers changed. I checked `classification_handlers_misc.go`: `ClassifyFactCheck` still returns `200` with `label: "fact_check_disabled"` when the classifier isn't initialized — which is exactly what #2120 itself says ("fact-check/user-feedback degrade to 200"). So the new test enforces a documented contract that's false for at least those two endpoints, and unverified for the rest (`intent`/`combined`/`batch`/`eval`/`nli`). Could you scope `ServiceUnavailable: true` to only the endpoints that actually gained a readiness check in this PR?

**Minor / non-blocking**

- `e2e/profiles/no-model/profile.go`: `func init() { _ = fmt.Sprintf("no-model profile registered") }` formats a string with no verbs and discards it — does nothing, safe to delete.
- `ensureEmbeddingModelReady` / `IsEmbeddingFamilyReady` / `SetEmbeddingFamilyReady` / `normalizeFamily` are near-identically reimplemented across `candle-binding/semantic-router.go`, `semantic-router_mock.go`, and `onnx-binding/semantic-router.go`, with inconsistent naming (`embeddingModelsReady` vs `embeddingReadyOverride`) for the same concept — not asking for a shared abstraction across modules, just flagging for awareness.
- No unit tests for the new binding-layer readiness logic itself (`ensureEmbeddingModelReady`, `syncEmbeddingFamilyReadiness`, `familyLoaded`, the Rust `is_embedding_family_ready`) — only the `apiserver` consumer is tested.

Once the OpenAPI over-claim and the onnx-binding question are resolved, I think this is close.

src/semantic-router/go.mod
```diff
@@ -6,6 +6,7 @@ replace (
 	github.com/vllm-project/semantic-router/candle-binding => ../../candle-binding
 	github.com/vllm-project/semantic-router/ml-binding => ../../ml-binding
 	github.com/vllm-project/semantic-router/nlp-binding => ../../nlp-binding
+	github.com/vllm-project/semantic-router/onnx-binding => ../../onnx-binding
```
theohsiung, line 9: This replace directive has no consumer — nothing in `src/semantic-router/pkg/`` imports `onnx-binding` (the embeddings API only ever uses `candle_binding`). The real candle↔onnx swap already lives in the separate `src/semantic-router/go.onnx.mod`, which replaces `candle-binding => ../../onnx-binding` directly — this line in the main go.mod doesn't participate in that mechanism at all. Please drop it unless there's a planned caller.

dashboard/backend/Dockerfile
```diff
@@ -10,10 +10,11 @@ COPY src/semantic-router/ ./
 
 # Local replace directives point at sibling bindings that are not needed for the
 # WASM build. Stub them under /app so `../../*-binding` resolves correctly.
-RUN mkdir -p /app/candle-binding /app/ml-binding /app/nlp-binding /app/openvino-binding && \
+RUN mkdir -p /app/candle-binding /app/ml-binding /app/nlp-binding /app/onnx-binding /app/openvino-binding && \
```
theohsiung, line 16: Follow-on from the go.mod comment: this onnx-binding stub is only needed because of the unused replace directive added in `src/semantic-router/go.mod`. If that replace line is dropped, this stub (and the matching `echo 'module ...'` line below) can go too.

src/semantic-router/pkg/apiserver/routes_catalog.go
```diff
@@ -40,73 +40,73 @@ func apiHealthRoutes() []apiRoute {
...
 		managedRoute(
-			EndpointMetadata{Path: "/api/v1/classify/security", Method: "POST", Description: "Detect jailbreak attempts and security threats"},
+			EndpointMetadata{Path: "/api/v1/classify/security", Method: "POST", Description: "Detect jailbreak attempts and security threats", ServiceUnavailable: true},
 			routePolicy{Permission: PermClassifyInvoke, Sensitivity: SensitivityOperational},
 			(*ClassificationAPIServer).handleSecurityDetection,
 			jsonBody(),
 		),
 		managedRoute(
-			EndpointMetadata{Path: "/api/v1/classify/fact-check", Method: "POST", Description: "Classify if text needs fact-checking"},
+			EndpointMetadata{Path: "/api/v1/classify/fact-check", Method: "POST", Description: "Classify if text needs fact-checking", ServiceUnavailable: true},
```
theohsiung: `ServiceUnavailable: true` here (and on `user-feedback` two lines down) doesn't match actual behavior. `ClassifyFactCheck` (`classification_handlers_misc.go`) returns 200 with `label: "fact_check_disabled"` when the classifier isn't initialized — no 503 path exists. That's exactly what #2120 itself says ("fact-check/user-feedback degrade to 200"). The new `TestOpenAPISpecDocumentsModelNotReady503` test locks in a 503 contract that's false for this endpoint. Same question applies to `intent`/`combined`/`batch`/`eval`/`nli` below — none of those handlers gained a readiness check in this PR, so marking them `ServiceUnavailable: true` overclaims. Suggest scoping this field to only `pii`/`security`/`embeddings`/`similarity`/`similarity/batch`, which actually got readiness checks.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -34,6 +35,28 @@ func (e *imageEncodeError) Error() string {
...
+}
+
+// checkEmbeddingReadiness validates that the models required for the request
+// are initialized. Text inputs require readiness of the requested model family
+// (an explicitly requested absent model is not ready, even when another text
+// family is loaded); image inputs require multimodal-model readiness. This
+// prevents a text-ready-only deployment from attempting image inference (which
+// would 500) and a multimodal-only deployment from being rejected for
+// text-only requests.
+func checkEmbeddingReadiness(req EmbeddingRequest) error {
```
theohsiung, line 48: This inline pre-check and the typed-error path in `classifyEmbeddingError` (`isEmbeddingModelNotReady`) are now two shapes for the same "not ready" condition — this is the follow-through I asked for on 7/15 that didn't fully land. Could this collapse into just the typed-error path (return the error from the generation call, map it via `errors.Is` like `writeClassificationError` does), so `handleEmbeddings`/`handleSimilarity`/`handleBatchSimilarity` don't need a separate pre-flight readiness check?

e2e/profiles/no-model/profile.go
```diff
@@ -0,0 +1,71 @@
...
+	return testmatrix.Combine(testmatrix.ReadinessContract)
+}
+
+// GetServiceConfig returns the service configuration.
+func (p *Profile) GetServiceConfig() framework.ServiceConfig {
+	return p.stack.ServiceConfig()
+}
+
+func init() {
+	_ = fmt.Sprintf("no-model profile registered")
```
theohsiung, line 70: This formats a string with no verbs and discards the result — it's a no-op. Safe to delete the whole `init()` block; the profile is already registered via `register("no-model", ...)` in `e2e/profiles/all/imports.go`.

Xunzhuo (commented):
The August update resolves the two code blockers from my last review: readiness is now tracked per requested embedding family, and the `no-model` testcase is selected and passed in CI.

The branch is now conflicting with `main`, and the required Core quality job still fails. Please rebase and adapt the profile/CI registration to the current agent E2E registry instead of carrying the old workflow edits. Ping me after the rebase for the final approval pass.

theohsiung:
@p172913 Could you fix the CI error, thx~

Xunzhuo (commented):
Thanks for rebasing. This head still removes the C declarations used later in `candle-binding/semantic-router.go` (`init_generic_classifier`, `classify_modernbert_jailbreak_text_with_probabilities`, and `classify_mmbert_32k_jailbreak_with_probabilities`), so the exact-head build cannot compile. The repository validation also fails because `no-model` was added to the test registry without updating `e2e/README.md`. Please restore the declarations, sync the supported-profile inventory, and rerun the failed checks.

Xunzhuo (changes requested):
Thanks for rebasing. The two exact blockers are still present on this head: the cgo preamble still lacks declarations for `init_generic_classifier`, `classify_modernbert_jailbreak_text_with_probabilities`, and `classify_mmbert_32k_jailbreak_with_probabilities`, while the supported-profile list in `e2e/README.md` still omits `no-model`. Please restore those declarations, synchronize the profile inventory, and rerun the exact-head build and registry checks.

Xunzhuo (changes requested):
Thanks—the earlier cgo declarations and no-model inventory are now in place. The current update also removes SupportsBatchedEmbedding from the real Candle binding while the router and its binding test still call it, so the exact-head core and E2E lanes fail to compile with undefined: candle_binding.SupportsBatchedEmbedding. Please restore the real helper, including its qwen3-only contract, and rerun the required checks.

FAUST-BENCHOU (dismissed):
LGTM. Thanks for fixing the classification model readiness error handling and adding regression coverage.

Xunzhuo (commented):
Thanks—the qwen3-only batched-capability helper is restored, and the exact-head core and E2E checks are green, so my latest compile blocker is resolved. One current review thread is still open; please resolve that thread, then this is ready for the final approval pass against current main.

p172913:
Hi @Xunzhuo could you please point me to the exact thread you're referring to here. I will try to address it as quick as possible.

Xunzhuo (commented):
Thanks for checking. The remaining open item is [this `go.mod` thread](https://github.com/vllm-project/semantic-router/pull/2321#discussion_r3763224228). It became outdated after the rebases, but GitHub still marks it unresolved. Please resolve it, then I can run the final approval pass.

p172913:
> Thanks for checking. The remaining open item is [this `go.mod` thread](https://github.com/vllm-project/semantic-router/pull/2321#discussion_r3763224228). It became outdated after the rebases, but GitHub still marks it unresolved. Please resolve it, then I can run the final approval pass.

@Xunzhuo . I resolved this issue as per your review.

candle-binding/semantic-router.go
```diff
@@ -168,14 +170,12 @@ typedef struct {
 typedef struct {
     int class;
     float confidence;
-    char* label;
 } ClassificationResult;
 
 // Classification result with full probability distribution structure
 typedef struct {
```
Xunzhuo, line 184: **Restore the Rust FFI layouts**

Impact: Rust still returns the generic probability result as `{float confidence; int predicted_class; char *label; float *probabilities; int num_classes}`. This declaration is smaller and reorders/removes fields, so `ClassifyTextWithProbabilities` reads the label as the probability pointer and a pointer fragment as the count; on common ABIs the Rust return also writes 32 bytes into a 24-byte C return slot before Go may free the wrong pointer. `classify_jailbreak_text_with_probabilities` has the opposite problem: Rust returns `ModernBertClassificationResultWithProbs`, but line 272 now declares the incompatible generic type.

Required change: Restore the generic C structs exactly to `ffi/types.rs`, and declare the jailbreak probability function with `ModernBertClassificationResultWithProbs`; add a binding-level ABI regression that exercises both calls.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -34,6 +35,26 @@ func (e *imageEncodeError) Error() string {
...
+		errors.Is(err, candle_binding.ErrEmbeddingModelNotReady)
+}
+
+// checkEmbeddingReadiness validates that the models required for the request
+// are initialized. Text inputs require text-model readiness; image inputs
+// require multimodal-model readiness. This prevents a text-ready-only
+// deployment from attempting image inference (which would 500) and a
+// multimodal-only deployment from being rejected for text-only requests.
+func checkEmbeddingReadiness(req EmbeddingRequest) error {
+	if len(req.Texts) > 0 && !candle_binding.IsEmbeddingReady() {
```
Xunzhuo: **Gate the backend selected by the request**

Impact: `texts` does not imply the unified text backend: `model: multimodal` sends text to `MultiModalEncodeText`, but this check rejects a valid multimodal-only deployment because `IsEmbeddingReady()` is false. Conversely, one aggregate bit cannot tell whether the explicitly requested qwen3, gemma, or mmBERT model loaded, and standalone `InitMmBertEmbeddingModel` never sets that bit. This regresses the earlier per-family readiness fix.

Required change: Derive readiness from `req.Model` and the actual per-family loaded state, including standalone initialization, and cover multimodal-text plus partial-family configurations.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -105,8 +123,8 @@ func (s *ClassificationAPIServer) parseEmbeddingRequest(w http.ResponseWriter, r
 
 	applyEmbeddingDefaults(&req)
 	mmbertPath := ""
-	if cfg := s.currentConfig(); cfg != nil {
-		mmbertPath = cfg.EmbeddingModels.MmBertModelPath
+	if s.config != nil {
```
Xunzhuo: **Read the current config snapshot**

Impact: Config publication replaces `s.config` under `configMu` and updates `runtimeConfig`, but this direct read takes neither path. Concurrent config updates race with request validation, and `target_layer` can be validated against stale startup model metadata.

Required change: Use `s.currentConfig()` here, as current main does, and retain the concurrent-mutation coverage.

src/semantic-router/pkg/services/classification_pii.go
```diff
@@ -62,6 +62,10 @@ func (s *ClassificationService) DetectPII(req PIIRequest) (*PIIResponse, error)
 		}, nil
 	}
 
+	if !s.classifier.IsPIIModelReady() {
```
Xunzhuo: **Use the locked classifier snapshot consistently**

Impact: The method captures `classifier` under `configMutex` above, then reads `s.classifier` again without that lock for readiness. A live reload can replace the pointer between those operations, causing a data race and checking one generation before inferring with another; `classification_security.go:58` repeats the same pattern.

Required change: Call the readiness method on the captured `classifier` in both handlers and add reload/race coverage.

candle-binding/semantic-router.go
```diff
@@ -1939,63 +1986,54 @@ func IsModelInitialized() (rustState bool, goState bool) {
...
-		return nil
-	}
-	if modelPath == "" {
-		modelPath = "bert-base-uncased"
-	}
-	if numClasses < 2 {
-		return fmt.Errorf("number of classes must be at least 2, got %d", numClasses)
-	}
+	var err error
+	classifierInitOnce.Do(func() {
```
Xunzhuo, line 2129: **Do not consume initialization after a failure**

Impact: `err` is local to each call, but `sync.Once` is permanently consumed even when validation or native initialization fails. A later retry with corrected inputs skips the closure and returns nil, allowing an uninitialized classifier to be published as ready.

Required change: Preserve and return the initialization error on every later call, or retain retry-until-success semantics, with a failed-first-attempt regression.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -339,8 +371,8 @@ func (s *ClassificationAPIServer) handleBatchSimilarity(w http.ResponseWriter, r
 		ProcessingTimeMs: result.ProcessingTimeMs,
 	}
 
-	logging.Infof("Calculated batch similarity: query=%s, %d candidates, top-%d matches (model: %s, took: %.2fms)",
-		logging.ContentDescriptor(req.Query), len(req.Candidates), len(matches), result.ModelType, result.ProcessingTimeMs)
+	logging.Infof("Calculated batch similarity: query='%s', %d candidates, top-%d matches (model: %s, took: %.2fms)",
```
Xunzhuo: **Keep request content out of logs**

Impact: This reintroduces the raw user query into application logs. Batch-similarity inputs may contain private prompts or PII; current main deliberately logs `logging.ContentDescriptor(req.Query)` instead.

Required change: Restore the content descriptor/redaction rather than formatting `req.Query` directly.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -379,28 +411,17 @@ func applyBatchSimilarityDefaults(req *BatchSimilarityRequest) {
 }
 
 func validateBatchSimilarityRequest(req BatchSimilarityRequest) (string, string, bool) {
-	if strings.TrimSpace(req.Query) == "" {
+	if req.Query == "" {
```
Xunzhuo: **Preserve the existing similarity input contract**

Impact: This rewrite accepts whitespace-only queries/candidates and removes the documented `[0,1]` priority validation; the simple-similarity parser above loses the same checks. Its dimension message also rejects the wording for accepted dimension 64. Invalid requests can now reach the binding instead of receiving deterministic 400 responses.

Required change: Restore the shared whitespace, candidate, priority, and dimension validation from current main, together with the deleted table tests.

candle-binding/semantic-router.go
```diff
@@ -1249,19 +1290,10 @@ func MultiModalEncodeImageFromURL(url string, targetDim int) (*MultiModalEmbeddi
...
 	)
 
 	client := &http.Client{Timeout: time.Duration(httpTimeout) * time.Second}
-	req, err := http.NewRequest(http.MethodGet, url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("failed to build HTTP request: %w", err)
-	}
-	req.Header.Set("User-Agent", userAgent)
-	resp, err := client.Do(req)
+	resp, err := client.Get(url)
```
Xunzhuo: **Retain the project User-Agent for image downloads**

Impact: `client.Get` sends Go's generic User-Agent. The prior request construction was added because hosts with User-Agent policies, including Wikimedia, return 403 for that client, so valid configured image candidates regress to download failures.

Required change: Restore the explicit request and project User-Agent before calling `client.Do`.

Xunzhuo (commented):
Thanks for the rebase work. This head introduces several regressions; the FFI layout mismatch is the most urgent.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -232,33 +253,71 @@ func embeddingOutput(req EmbeddingRequest, text string) (*candle_binding.Embeddi
 		return candle_binding.GetEmbeddingWithMetadata(text, req.QualityPriority, req.LatencyPriority, req.Dimension)
 	case "mmbert":
 		return candle_binding.GetEmbedding2DMatryoshka(text, req.Model, req.TargetLayer, req.Dimension)
+	case "multimodal":
+		return candle_binding.MultiModalEncodeText(text, req.Dimension)
```
Xunzhuo: **Adapt the multimodal result and dimension contract**

Impact: The exact-head Core build fails here because `MultiModalEncodeText` returns `*MultiModalEmbeddingOutput`, while this function promises `*EmbeddingOutput`. Even after wrapping that value, an omitted dimension is still defaulted to 768, but the multimodal encoder's full width is 384 and rejects larger targets; the HTTP validator also rejects 384. A normal `{model: "multimodal", texts: [...]}` request therefore cannot compile today and would fail at runtime after the type fix.

Required change: Convert the multimodal result into the API's embedding result shape (or safely generalize the helper), apply a model-aware 384 default and supported-dimension validation, and add a loaded-model endpoint regression for the omitted-dimension request.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -34,6 +35,29 @@ func (e *imageEncodeError) Error() string {
...
+// checkEmbeddingReadiness validates that the models required for the request
+// are initialized. Text inputs require text-model readiness; image inputs
+// require multimodal-model readiness. This prevents a text-ready-only
+// deployment from attempting image inference (which would 500) and a
+// multimodal-only deployment from being rejected for text-only requests.
+func checkEmbeddingReadiness(req EmbeddingRequest) error {
+	if (len(req.Texts) > 0 || len(req.Images) > 0) && req.Model == "multimodal" && !candle_binding.IsMultiModalReady() {
+		return candle_binding.ErrEmbeddingModelNotReady
+	}
+	if len(req.Texts) > 0 && !candle_binding.IsEmbeddingModelReady(req.Model) {
```
Xunzhuo, line 52: **Validate model names before readiness**

Impact: `IsEmbeddingModelReady` returns false for every unknown model string, so this new preflight maps typos and unsupported values to retryable 503 responses even when the server is healthy. The same ordering is repeated for similarity and batch similarity.

Required change: Validate and normalize each endpoint's supported model enum before checking readiness, return a permanent 400 for invalid models, and cover an unknown model on a ready server.

Xunzhuo (commented):
Thanks for the careful update. The earlier blockers are addressed, but two embedding API issues still block this head.

Xunzhuo (commented):
Thanks for the substantial update—three readiness regressions still need fixes on this head.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -36,6 +37,29 @@ func (e *imageEncodeError) Error() string {
...
```
Xunzhuo, line 54: **Check the acquired embedding generation**

Impact: This request already holds the exact prepared generation used by the owned embedding path, but this preflight ignores it and reads legacy process-global Candle flags. Owned Candle/ORT and remote OpenAI-compatible providers are prepared directly into `embedding.Set` without setting those flags, so a healthy provider is rejected with 503; a stale flag from another generation can also let a missing provider fall through as 500. Similarity and batch similarity repeat the same split-brain check.

Required change: Validate the supported model name first, then derive readiness and model availability from the acquired `embedding.Set` while its release lease is held, including multimodal; cover owned local and remote providers plus a generation swap.

src/semantic-router/pkg/classification/classifier.go
```diff
@@ -64,6 +64,12 @@ type Classifier struct {
 	// Preference classifier for route matching via external LLM
 	preferenceClassifier *PreferenceClassifier
 
+	// Model readiness flags — set true after successful model initialization.
+	// Handlers use these to distinguish "model not loaded" (503) from runtime
+	// failures (500).
+	piiModelReady       bool
```
Xunzhuo, line 70: **Mark constructed remote classifiers ready**

Impact: These flags become true only in the protocol/local initializer paths. The `PromptGuard.Backend` and `PIIModel.Backend` branches both return success without setting them, so `CheckSecurity` and `DetectPII` now reject every successfully constructed remote backend with `ErrModelNotReady`/503 before inference.

Required change: Derive readiness from the successfully constructed inference backend, or mark both backend branches ready after successful assembly, and add endpoint regressions for remote PII and PromptGuard backends.

e2e/profiles/no-model/values.yaml
```diff
@@ -0,0 +1,54 @@
...
+    model_catalog:
+      embeddings:
+        semantic:
+          # Clear every local embedding path. Otherwise the default catalog
+          # injects models/mmbert-embed-32k-2d-matryoshka, which makes the
+          # embedding runtime report ready and triggers model download plus
+          # tools-database warmup even though this profile intends no models.
+          qwen3_model_path: ""
+          gemma_model_path: ""
+          mmbert_model_path: ""
```
Xunzhuo, line 50: **Make the no-model profile boot**

Impact: The exact-head no-model job never reaches `model-not-ready-503`: the router crash-loops with `required embedding model "mmbert" has no artifact path` and setup times out. Clearing this artifact while the effective canonical config still requires the default primary means this PR currently has no end-to-end evidence for its public 503 contract.

Required change: Make the profile's effective config contain no embedding consumers, or otherwise provide a bootable classifier-free configuration, then rerun the profile and verify the testcase itself executes.

candle-binding/semantic-router.go
```diff
@@ -455,6 +455,22 @@ extern void candle_mlp_free_string(char* ptr);
 */
 import "C"
 
+var ErrEmbeddingModelNotReady = errors.New("embedding model is not initialized")
```
Xunzhuo, line 475: **[P1] Keep all imports before the new readiness declarations**

These declarations now sit between `import "C"` and the ordinary import block. Go requires imports before package declarations, and the exact-head Core job fails at line 474 with `imports must appear before other declarations`, preventing the router and no-model tests from building. Move the declarations below the imports and add the `errors` and `sync/atomic` imports used by the new code, then rerun Core and no-model E2E.

Xunzhuo (changes requested):
This head cannot build because the new Candle readiness declarations precede the regular imports. Please fix the import block and rerun Core and the no-model profile. Existing unresolved findings remain in their original threads; runtime readiness was not verified on this head.

Xunzhuo (dismissed):
The remote-classifier readiness and embedding error-envelope fixes address the remaining regressions. Current-head Core readiness tests pass, and the no-model deployment now boots and executes model-not-ready-503 successfully.

src/semantic-router/pkg/classification/classifier_builtin_models.go
```diff
@@ -63,6 +63,16 @@ func (c *Classifier) initializeJailbreakClassifier() error {
 	}
 
 	if c.Config.PromptGuard.Backend != nil {
+		c.jailbreakModelReady = true
+		return nil
+	}
+	if c.Config.PromptGuard.Protocol != "" {
```
Xunzhuo: [P1] Remove the retired Protocol branch after merging main

Main has removed `PromptGuardConfig.Protocol`, but this initialization branch and `IsJailbreakModelReady` in `classifier.go` still reference it. The current [ExtProc image build](https://github.com/vllm-project/semantic-router/actions/runs/35512433942/job/106082931718) fails with `Protocol undefined`; line 70 also refers to `config` after this file's import was removed. This blocks the router from compiling. Remove the legacy Protocol initialization and readiness checks, retain `jailbreakModelReady = true` for the supported `Backend` path, and cover that path in the readiness regression.

Xunzhuo (changes requested):
The latest main merge removes the textual conflict but leaves obsolete PromptGuard.Protocol references in readiness and initialization. The exact-head ExtProc CI build fails on these references, so the router cannot currently compile. Please remove the retired branch while preserving readiness on the canonical Backend path, then rerun the router build and readiness/no-model regressions. Runtime checks on this new head have not run successfully.

src/semantic-router/pkg/apiserver/route_embeddings.go
```diff
@@ -38,6 +38,51 @@ func (e *imageEncodeError) Error() string {
...
+
+// textEmbeddingReady reports whether the text portion's selected model family
+// is prepared for the generation. An unspecified ("") or "auto" request needs
+// any prepared text model; a multimodal-only generation must not satisfy text
+// inputs. An explicit family (qwen3, gemma, mmbert, ...) must be present, with
+// model == "multimodal" allowed to serve text through its own provider.
+func textEmbeddingReady(set *embedding.Set, model string) bool {
+	switch strings.ToLower(strings.TrimSpace(model)) {
+	case "", "auto":
+		for _, info := range set.Models() {
```
Xunzhuo: Thanks for using the prepared set here. A BERT-only RAG generation passes this `auto` check, but `embedding.Set.Select` only tries qwen3/gemma/mmbert and the configured primary (default qwen3). In a focused HTTP test, `model:auto` returned 500 `EMBEDDING_GENERATION_FAILED`; explicit `model:bert` returned 200. Please make readiness match the models `auto` can select, or allow it to select BERT, and cover this case.

Xunzhuo (changes requested):
One remaining blocker: the new `model:auto` readiness check accepts a BERT-only generation that auto-selection cannot use, so the embeddings endpoint returns 500. A focused HTTP regression reproduced this on the current head; Core and no-model CI checks pass. Please address the inline case and rerun the relevant checks. Native model inference was not run.

PR #2323, opened 2026-07-05
[Router] Reassemble split SSE frames in double-Anthropic streaming (#2316)

### LINKED ISSUE #2316: bug: <bug report anthropic sse>
## Describe the bug

When all models in the router config use `api_format: anthropic` with upstream `provider: anthropic` (double-Anthropic cell), SSE streaming responses produce **zero output**. The client receives no data on the stream at all.

Steps:
1. Start the router with the config below (all models use `api_format: anthropic` + `provider: anthropic`).
2. Point Claude Code (latest) at the router's `/v1/messages` endpoint.
3. Send a simple message like "你好" (hello).

The bug is 100% reproducible — every streaming request is silent.

## To Reproduce

**Minimal config** (sensitive info replaced with placeholders):

```yaml
version: v0.3
listeners:
    - name: http-8899
      address: 0.0.0.0
      port: 8899
      timeout: 300s
providers:
    defaults:
        default_model: test-model
    models:
        - name: test-model
          reasoning_family: deepseek
          provider_model_id: test-model
          backend_refs:
            - name: primary
              protocol: https
              weight: 100
              base_url: https://<UPSTREAM_ENDPOINT>
              provider: anthropic
              api_key: <API_KEY>
          api_format: anthropic
routing:
    modelCards:
        - name: test-model
          modality: text
    decisions:
        - name: default-route
          priority: 100
          rules:
            operator: AND
          modelRefs:
            - model: test-model
              use_reasoning: true
global:
    router:
        config_source: file
        auto_model_name: MoM
        include_config_models_in_list: false
```

With Claude Code (latest version):

```
claude --endpoint http://localhost:8899/v1/messages --api-key test "你好"
```

## Expected behavior

The SSE stream should return proper Anthropic events (`message_start` → `content_block_start` → `content_block_delta` → … → `message_stop`) and the client should display the assistant's response.

## Actual behavior

The SSE stream produces **no data at all**. The client hangs waiting for chunks on the wire.

## Affected layer

Unsure — likely in the response-body streaming pipeline (double-Anthropic translation path).

### Code path analysis

1. **Request routing** (`processor_req_body.go`): `GetModelAPIFormat(targetModel) == APIFormatAnthropic` → `handleAnthropicRouting()` → sets `ctx.APIFormat = APIFormatAnthropic`.

2. **Response dispatch** (`processor_res_body.go`): `IsStreamingResponse && ClientProtocol == ClientProtocolAnthropic` → `handleAnthropicClientStreamingResponseBody()`.

3. **Stream handler** (`processor_res_body_streaming_anthropic_client.go`):
   - `translateUpstreamToOpenAI()` translates upstream Anthropic SSE → OpenAI SSE via `TransformSSEChunkToOpenAI()`.
   - Then `EmitAnthropicSSEChunk()` translates OpenAI SSE back → Anthropic SSE for the client.

The symptom suggests that either the Anthropic→OpenAI translation, the OpenAI→Anthropic re-emission, or the shared `StreamState` lifecycle across both stages is dropping data silently.

### Notes

- Non-streaming requests on the same path may work; only streaming is broken.
- OpenAI-native clients hitting Anthropic backends (single-direction translation) appear unaffected.
- `stream: true` is correctly propagated to the upstream request body via `WithStreamingRequestBody()`.
- Environment: commit `263c3625`, Windows, Go router.

### To Reproduce

config like that
use claude code

### Expected behavior

_No response_

### Affected layer

None

### Additional context

_No response_

theohsiung:
## Summary

Fixes #2316 — double-Anthropic streaming (Anthropic client + Anthropic backend, e.g. a reasoning model routed with `api_format: anthropic` + `provider: anthropic`) produced **zero output** and the client hung waiting for events.

## Root cause

Envoy STREAMED mode delivers the upstream response body split at **arbitrary byte offsets**, with no guarantee that a chunk boundary aligns to an SSE frame boundary — a single event can straddle two body chunks.

`handleAnthropicClientStreamingResponseBody` parsed each chunk independently with **no cross-chunk buffering**, so any event split across two chunks failed `json.Unmarshal` in `extractSSEDataLines` and was silently dropped. Because this handler **suppresses the raw upstream bytes** to keep ownership of the wire (it synthesizes its own Anthropic SSE via `emptyAnthropicBodyMutation`), a dropped frame *vanishes entirely* rather than being forwarded. Reasoning models emit large, verbose streams whose frames split routinely, so the client received nothing and hung.

The sibling OpenAI-client handler (`handleAnthropicStreamingResponseBody`) masks the same underlying split because it returns a **nil** `BodyMutation` on empty translation, letting Envoy forward the raw bytes — which is exactly why only the double-Anthropic cell exhibited the failure while the OpenAI-client-to-Anthropic-backend cell appeared unaffected.

The translation logic itself is correct: happy-path delivery (per-event, batched, whole-response, reasoning round-trip) all translate fine. Only byte-split delivery breaks it.

## Fix

Reassemble complete SSE frames across response-body chunks:

- New `reassembleSSEFrames` helper (`sse_frame_buffer.go`) merges the trailing partial frame carried on `RequestContext.PendingSSEBytes` with the next chunk, splits off only complete frames (delimited by a blank line, `\n\n`) for translation, and holds the remainder for the following chunk.
- `handleAnthropicClientStreamingResponseBody` now reassembles before translating; the keepalive ping still flows when only a partial frame arrived, so a long split does not stall the connection.

Scope is intentionally limited to the reported client handler. The sibling handler has the same latent split behavior but is masked by its nil-mutation/forward-raw contract; fixing it changes that contract and is left as a follow-up.

## Test Plan

- [x] `TestReassembleSSEFrames` — unit tests for the helper: complete frame pass-through, mid-JSON split held as remainder, remainder completed on next chunk, trailing partial after a complete frame, empty chunk, split exactly between the two newlines.
- [x] `TestReassembleSSEFrames_RemainderDoesNotAliasChunk` — remainder is copied, not aliasing Envoy's buffer.
- [x] `TestDoubleAnthropicStream_FrameAlignedDelivery` — baseline: full reasoning stream (message_start → thinking → text → message_delta → message_stop) translates correctly one-frame-per-chunk.
- [x] `TestDoubleAnthropicStream_ByteSplitDelivery` — regression: same stream sliced at spans of 1, 7, 32, 64, 200 bytes (span=1 = one byte per chunk); client always receives message_start, thinking/text deltas, the content, and message_stop.
- [x] `go test ./pkg/extproc/ ./pkg/anthropic/` pass; `-race` clean on streaming tests.
- [x] `golangci-lint` with the repo's strict agent config → 0 issues; `gofmt`/`go vet` clean.

## Note for the reporter

I could reproduce the exact symptom (zero output + hang) deterministically via byte-split delivery, which matches every observed characteristic (reasoning model, only the double-Anthropic cell affected). If you can `git checkout` this branch and confirm against your upstream — or share an SSE wire capture — that would fully close the loop.

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
+++ b/src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go
@@ -51,6 +51,19 @@ func (r *OpenAIRouter) handleAnthropicClientStreamingResponseBody(
 		ctx.AnthropicStream = anthropic.NewStreamState()
 	}
 
+	// Envoy STREAMED mode delivers the response body split at arbitrary
+	// byte offsets, so an upstream SSE frame may straddle two chunks.
+	// Reassemble complete frames before translation and hold any trailing
+	// partial frame for the next chunk. Without this, a split frame fails
+	// to parse, the translated output is empty, and — because this handler
+	// suppresses the raw upstream bytes to keep ownership of the wire (see
+	// emptyAnthropicBodyMutation) — the frame vanishes entirely, hanging
+	// the client (issue #2316). The keepalive ping still flows below even
+	// when only a partial frame arrived, so a long split does not stall the
+	// connection.
+	frames, remainder := reassembleSSEFrames(ctx.PendingSSEBytes, responseBody)
+	ctx.PendingSSEBytes = remainder
+
 	// Inject a keepalive ping if the gap since the last chunk crossed
 	// the cadence threshold. Anthropic clients (including the SDK
 	// MessageStream accumulator) treat pings as no-ops; the value is
@@ -59,7 +72,7 @@ func (r *OpenAIRouter) handleAnthropicClientStreamingResponseBody(
 	// reaches the client before this chunk's content events.
 	pingBytes := maybeEmitPing(ctx.AnthropicStream)
 
-	openAIBytes, err := r.translateUpstreamToOpenAI(responseBody, ctx)
+	openAIBytes, err := r.translateUpstreamToOpenAI(frames, ctx)
 	if err != nil {
 		logging.Errorf("Failed to translate upstream streaming chunk to OpenAI: %v", err)
 		return emptyAnthropicBodyMutation()
--- a/src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client_split_test.go
+++ b/src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client_split_test.go
@@ -1,0 +1,100 @@
+package extproc
+
+import (
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/anthropic"
+)
+
+// reasoningStreamFrames returns a realistic reasoning-model upstream
+// Anthropic SSE sequence (message_start → thinking block → text block →
+// message_delta → message_stop), one complete frame per element.
+func reasoningStreamFrames() []string {
+	frame := func(event, data string) string {
+		return "event: " + event + "\n" + "data: " + data + "\n\n"
+	}
+	return []string{
+		frame("message_start", `{"type":"message_start","message":{"id":"msg_2316","type":"message","role":"assistant","model":"test-model","content":[],"stop_reason":null,"usage":{"input_tokens":10,"output_tokens":1}}}`),
+		frame("content_block_start", `{"type":"content_block_start","index":0,"content_block":{"type":"thinking","thinking":""}}`),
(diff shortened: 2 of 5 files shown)


theohsiung:
@WUKUNTAI-0211 nice to have your review

WUKUNTAI-0211 (approved):
Nice catch on the root cause. Traced the handler against the diff — reassembly is correct and the byte-split test genuinely reproduces #2316. `LastIndex("\n\n")` handles multi-frame chunks, only complete frames forwarded so no mid-rune UTF-8 split (nice `span=1` + `你好`), and the no-alias copy is tested. LGTM 

Two optional nits:
- `assertUsableAnthropicStream` only uses `Contains` — add `strings.Count(wire, "event: message_stop") == 1` so a duplicated frame can't slip through.
- `reassembleSSEFrames` allocates a fresh buffer every chunk; when `len(pending)==0` you could `LastIndex(chunk)` directly and only copy the remainder, saving a copy per streaming call.

Both fine as follow-ups.

PR #2325, opened 2026-07-05
[Router] Reject complexity decision conditions without a difficulty suffix (#2324)

### LINKED ISSUE #2324: [Router] complexity decision condition with bare rule name silently never matches (validation vs runtime mismatch)
## Describe the bug

A `decision` condition of `type: complexity` that references a complexity rule by its **bare name** (e.g. `name: needs_reasoning`) passes config validation but **never matches at runtime**, silently disabling that route. To fire, the condition must use the `rule:difficulty` form (e.g. `name: needs_reasoning:hard`).

The shipped example `config/decision/composite/priority-safe-escalation.yaml` uses the bare form, so its complexity branch is dead as written — and anyone copying it inherits a silently-broken route.

## Root cause

There is a mismatch between how complexity signal names are **validated** and how they are **matched at runtime**:

- **Emit** — the classifier emits `"<rule>:<difficulty>"` into `MatchedComplexityRules`, where difficulty ∈ {`hard`, `easy`, `medium`}:
  - `src/semantic-router/pkg/classification/classifier_signal_complexity.go:34` → `matchName := fmt.Sprintf("%s:%s", result.RuleName, result.Difficulty)`
  - `src/semantic-router/pkg/classification/complexity_rule_scoring.go:73` (`classifyComplexityDifficulty`)
- **Handoff** — passed straight through to the decision engine:
  - `src/semantic-router/pkg/classification/classifier_signal_decision.go:66` → `ComplexityRules: signals.MatchedComplexityRules`
- **Runtime match** — literal membership test:
  - `src/semantic-router/pkg/decision/engine.go:220,235` → `slices.Contains(signals.ComplexityRules, name)`
  - So `slices.Contains(["needs_reasoning:hard"], "needs_reasoning")` → **false**.
- **Validation** — strips the `:` suffix before checking the base name, so the bare form passes validation:
  - `src/semantic-router/pkg/config/validator_projection.go:433` (`complexityNameDeclared`)

Net effect: validation accepts `needs_reasoning`, runtime only matches `needs_reasoning:hard|easy|medium` → bare-name complexity conditions are silently inert.

## To reproduce

Minimal decision-engine check (bare vs suffixed) against `SignalMatches{ComplexityRules: ["needs_reasoning:hard"]}`:

```
bare   name "needs_reasoning"      -> matched=false
suffix name "needs_reasoning:hard" -> matched=true
```

(Confirmed with a throwaway `EvaluateDecisionsWithSignals` unit test in `pkg/decision`.)

## Expected behavior

Either the bare name should match (any difficulty), or config validation should reject a bare complexity name and require the `rule:difficulty` form — validation and runtime should agree, and the shipped example should be a working reference.

## Suggested fixes (pick one — needs a maintainer call on the intended contract)

1. **Make runtime matching accept a bare rule name** (match any difficulty) so validation and runtime agree — most user-friendly; also lets configs route on "rule fired at all" vs a specific level.
2. **Tighten validation** to reject bare complexity names and require `rule:hard|easy|medium`, with a clear error.
3. At minimum, **fix the example** `config/decision/composite/priority-safe-escalation.yaml` to use `name: needs_reasoning:hard`.

## Affected layer

Router — decision engine / config validation (complexity signal).

## Additional context

Found while investigating how to use semantic difficulty as a routing signal; the `complexity` signal itself works, only the decision-condition name contract is inconsistent.

theohsiung:
## Summary

Fixes #2324 — a `decision` condition of `type: complexity` that references a complexity rule by its **bare name** (e.g. `name: needs_reasoning`) passed config validation but **never matched at runtime**, silently disabling that route.

## Root cause

The decision engine matches complexity conditions by literal membership against the classifier's emitted `"<rule>:<difficulty>"` strings:

- **Emit**: `classifier_signal_complexity.go` → `matchName := fmt.Sprintf("%s:%s", RuleName, Difficulty)` (difficulty ∈ `hard|easy|medium`).
- **Runtime match**: `pkg/decision/engine.go` → `slices.Contains(signals.ComplexityRules, name)`.
- **Validation**: `validator_projection.go` `complexityNameDeclared` strips the `:` suffix before checking the base name, so the bare form passed validation.

So `slices.Contains(["needs_reasoning:hard"], "needs_reasoning")` → `false`. A complexity rule always classifies into one of `hard`/`easy`/`medium`, so a bare name can never be a discriminating match — the route is silently inert.

## Fix (fail-fast validation)

Add a dedicated `validateComplexityContracts` validator (sibling file per `pkg/config/AGENTS.md`, registered in the validator table) that walks each decision's rule tree and requires every `type: complexity` condition to name a **declared** rule in the `"<rule>:<hard|easy|medium>"` form. Bare names, unknown/missing difficulties, and undeclared rules now fail at config load with a clear error, mirroring the existing domain reference validator.

Also fixes the one shipped example that used the bare form and documents the required form.

## Why this contract (vs. making runtime lenient)

A complexity rule emits a difficulty for **every** request, so treating a bare name as "match any difficulty" would make it always-true (useless). Every maintained config (`config/config.yaml`, all `deploy/recipes/*`, kserve) already uses the explicit `:difficulty` form and declares its rules — so this enforces the de-facto contract with **zero breakage**; only the one broken example is corrected.

## Changes

- `pkg/config/validator_complexity.go` (new) + registration in `pkg/config/validator.go`
- `pkg/config/validator_complexity_test.go` (new)
- `config/decision/composite/priority-safe-escalation.yaml`: `needs_reasoning` → `needs_reasoning:hard`
- `website/docs/tutorials/decision/composite.md`: same fix
- `website/docs/tutorials/signal/learned/complexity.md`: document the `<rule>:<hard|easy|medium>` routing form

## Test Plan

- [x] `TestValidateComplexityContracts` (table-driven): valid `hard`/`easy`/`medium`, nested under OR/NOT, bare name rejected (incl. nested), unknown difficulty rejected, undeclared rule rejected, leading/trailing colon rejected.
- [x] `TestValidateComplexityContracts_NoComplexityConditions`: decisions without complexity conditions unaffected.
- [x] `go test ./pkg/config/` passes — including `maintained_asset_contract_test` (validates all `deploy/` recipes through the new validator) and the reference-config enforcement.
- [x] `gofmt`/`go vet` clean; `golangci-lint` (strict agent config) → 0 issues; `go build ./pkg/...` clean.

### DIFF
--- a/config/decision/composite/priority-safe-escalation.yaml
+++ b/config/decision/composite/priority-safe-escalation.yaml
@@ -13,7 +13,7 @@ routing:
               - type: keyword
                 name: urgent_keywords
               - type: complexity
-                name: needs_reasoning
+                name: needs_reasoning:hard
           - operator: NOT
             conditions:
               - type: jailbreak
--- a/src/semantic-router/pkg/config/validator.go
+++ b/src/semantic-router/pkg/config/validator.go
@@ -91,6 +91,11 @@ var (
 			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
 		},
 		{
+			name:     "complexity",
+			validate: validateComplexityContracts,
+			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
+		},
+		{
 			name:     "model_selection",
 			validate: validateModelSelectionConfig,
 			scopes:   configValidationScopeFile | configValidationScopeKubernetes,
--- a/src/semantic-router/pkg/config/validator_complexity.go
+++ b/src/semantic-router/pkg/config/validator_complexity.go
@@ -1,0 +1,101 @@
+package config
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// complexityDifficultyLevels enumerates the difficulty levels a complexity
+// rule can emit at runtime. It mirrors classifyComplexityDifficulty in
+// pkg/classification/complexity_rule_scoring.go, which always tags a rule
+// result as "<rule>:<level>" (e.g. "needs_reasoning:hard"). The config
+// package cannot import the classification package (import cycle), so the set
+// is duplicated here; keep the two in sync.
+var complexityDifficultyLevels = map[string]struct{}{
+	"hard":   {},
+	"easy":   {},
+	"medium": {},
+}
+
+// validateComplexityContracts checks that every decision condition of type
+// "complexity" references a declared complexity rule using the required
+// "<rule>:<difficulty>" form.
+//
+// The decision engine matches complexity conditions by literal membership
+// against the classifier's emitted "<rule>:<difficulty>" strings
+// (pkg/decision/engine.go uses slices.Contains), so a bare rule name — or one
+// with an unknown difficulty — never matches at runtime and becomes a
+// silently inert route. Config validation previously accepted the bare form
+// because the projection-input validator strips the ":" suffix before
+// checking the base name, leaving the mismatch undetected. Enforcing the
+// suffix here turns that latent misconfiguration into a clear load-time error
+// (issue #2324).
+func validateComplexityContracts(cfg *RouterConfig) error {
+	declared := collectComplexityRuleNames(cfg.ComplexityRules)
+	for _, decision := range cfg.Decisions {
+		if err := validateDecisionComplexityReferences(decision.Name, &decision.Rules, declared); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func validateDecisionComplexityReferences(decisionName string, node *RuleNode, declared map[string]struct{}) error {
+	if node == nil {
+		return nil
+	}
+
+	if node.Type == SignalTypeComplexity {
+		if err := validateComplexityConditionName(decisionName, node.Name, declared); err != nil {
+			return err
+		}
+	}
+
+	for i := range node.Conditions {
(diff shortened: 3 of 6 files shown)


Xunzhuo (approved):
LGTM for the fix

Xunzhuo:
Also i am thinking about if this kind of nested naming logics is a good pattern, any general ideas to make it sync with other signals?

theohsiung:
yeah, looked into this a bit — a few thoughts, curious what you think.

I don't think the nesting itself is the problem; a rule + a graded label is genuinely two pieces of info. The smell is cramming both into one `name` with a colon (`needs_reasoning:hard`) — that's the #2324 footgun (bare name validates but never matches).

On syncing: there isn't really one convention to align to — we already have `name` = the rule that fired (jailbreak, pii…) vs `name` = the output label (domain→category, modality→AR/BOTH…). complexity is a third shape (rule + label). Cleanest unification I can see is splitting them: `type` + `name` (the instance) + optional `label` (the output). Boolean signals omit `label`, graded ones set it, and `name` means the same thing everywhere.

But it's only complexity today, so I'd lean toward not generalizing yet (rule of three) — keep #2325 as the guardrail, just write the direction down so the next multi-label signal doesn't repeat the pattern, and revisit when a second graded signal actually shows up.

Does that match how you're seeing it? Happy to draft a short RFC either way.

PR #2328, opened 2026-07-05
[Benchmark] Add an NVIDIA CUDA CPU-versus-GPU benchmark harness

### LINKED ISSUE #2348: [Feature] CUDA perf benchmark harness + published CPU-vs-GPU numbers (follow-up to #2281)
### Goal

Follow-up to #2281 (CUDA `Dockerfile.cuda` + NVIDIA deploy guide). During review, @Xunzhuo asked for CUDA perf numbers (latency / throughput / batch size / context-length pressure) to land as a follow-up.

Tracked by a single consolidated PR:

- **#2328** `[Bench][Docs]` — the bench harness (CUDA counterpart of `bench/cpu-vs-gpu/bench-long-context.sh`) plus the measured CPU-vs-GPU numbers (RTX 4090 baseline) in `deploy/nvidia/README.md`.

Current status: #2328 is open and remains the implementation PR. Keeping the harness and measured documentation together avoids dangling references if it lands.

### Proposed Workgroup

Evaluation & Quality

### Scope

- CUDA signal-extraction latency across 500 / 2K / 8K / 16K tokens, all three classifiers (domain / jailbreak / PII).
- Concurrency/throughput sweep (ext_proc has no batch knob, so concurrency is the analog).
- Honest documentation of the 512-token classifier input cap and the embedding-vs-classifier parity distinction.

##### History

Originally split as #2327 (docs) + #2328 (bench) but consolidated into #2328 — the two halves cross-reference each other, so shipping them atomically avoids a window where `main` has dangling references. #2327 has been closed with a pointer to #2328.

### Non-goals

### Completion signal

WUKUNTAI-0211:
Related #2348

## What

Adds the NVIDIA CUDA counterpart of the ROCm `bench/cpu-vs-gpu/` harness: the
same Prometheus metric and the same 500 / 2K / 8K / 16K sweep, with `--gpus all`
instead of the ROCm device flags, plus a concurrency harness because ext_proc
classifies one request at a time and has no batch knob.

Rebased onto current `main` and reduced to the harness. The July RTX 4090
numbers are not carried over: they were produced by the legacy `use_cpu`
classifier path against a config that no longer parses. Re-measuring on this
branch produced a CPU baseline but no GPU column, because classifier inference
on an NVIDIA GPU is currently unreachable on `main` — see "Measured on an RTX
4090" below.

## Bench harness (`bench/cpu-vs-gpu/`)

- `bench-cuda-long-context.sh` — CPU vs GPU signal-extraction latency across
  500 / 2K / 8K / 16K token prompts, driving the domain / jailbreak / PII
  classifiers through Envoy ext_proc and reading the Prometheus
  `llm_signal_extraction_latency_seconds` histogram.
- `bench-cuda-throughput.sh` + `load_test.py` — concurrency sweep (N clients
  over a fixed duration).
- `config-bench-cuda.yaml` — canonical v0.3 config declaring all three signals
  so every classifier runs per request (`USE_CPU_PLACEHOLDER` toggles the
  CPU/GPU phase). Registered in `maintainedTemplatedConfigAssets`, so canonical
  schema drift now fails CI instead of silently breaking the benchmark.
- `signal_samples.py` — the sample gate described below.
- `bench/cpu-vs-gpu/README.md` gains the CUDA setup and run section.

Self-contained: reuses the shared `envoy-bench.yaml`, pulls public models from
`llm-semantic-router/mmbert32k-*-merged`, uses the image built from
`src/vllm-sr/Dockerfile.cuda`, ships a built-in stub upstream, and makes all
ports env-overridable for shared hosts. Percentiles use the Python stdlib.

## Validation gate

Addresses @Xunzhuo's review: the harness previously recorded every request
regardless of outcome, so a broken setup could publish plausible latency and
QPS with zero valid signal samples.

Both scripts now refuse to publish numbers from a run that did not classify:

- **HTTP outcome is load-bearing.** Only HTTP 200 contributes to the latency
  tables and to QPS. `load_test.py` reports `ok` / `http_err` / `conn_err`
  separately and exits non-zero when nothing succeeded; the long-context script
  drops non-200 rows from the e2e statistics and aborts the phase when any
  benchmark request failed. Warmup failures abort the run instead of being
  swallowed.
- **Metric deltas must match the traffic.** Each script scrapes
  `llm_signal_extraction_latency_seconds_count` before and after every
  measurement batch and requires each of `domain`, `jailbreak` and `pii` to
  record at least one new sample per successful request. `signal_samples.py`
  owns that comparison and exits non-zero on divergence.
- **The report is gated too.** `verify_results()` re-checks every table cell
  before a report is written; missing samples abort with a non-zero exit
  instead of emitting a table with silently skipped rows.
- **The GPU phase must actually be on a GPU.** `--gpus all` only exposes the
  device; whether the router uses it depends on the resolved model bindings.
  The GPU phase parses the router's `model_binding_ready` events and requires
  every prepared classifier to report a non-CPU device, so a phase that fell
  back to the CPU cannot be published as a speedup.

The metric floor is one sample per successful request rather than an exact
match, because a single request emits more than one sample for some signals.
Measured on this branch: 12 requests produce `domain 12`, `jailbreak 12`,
`pii 24` — the PII and jailbreak paths record a per-rule sample on top of the
per-request `*_evaluated` one. Exact equality would fail healthy runs.

Failure-path evidence, driven against local stub upstreams:

```
1. load_test.py against a router that 503s every request
   row : 4 0 8495 0 0 0 0 0
   err : ERROR: no successful responses (8495 HTTP errors, 0 transport errors)
   exit: 1

2. load_test.py against a healthy upstream
   row : 4 8752 0 0 4374.9 1 1 2
   exit: 0

3. signal_samples.py, 12 successful requests
   after-ok      -> exit 0   domain 12 / jailbreak 12 / pii 12
   after-none    -> exit 1   ERROR: signal samples below the expected floor of 12: domain=0, jailbreak=0, pii=0
   after-partial -> exit 1   ERROR: signal samples below the expected floor of 12: jailbreak=0

4. send_requests + e2e statistics against a 503 router
   SEND_OK=0 SEND_FAILED=3
   csv: 1,17,503 / 2,5,503 / 3,6,503
   compute_e2e_stats -> 0 0 0 0 0 0      (previously: 3 samples, ~9 ms average)
```

## Rebase

The branch was 631 commits behind, so it is rebased rather than merged:

- `config-bench-cuda.yaml` no longer parsed on `main`. Fixed against the
  loader: `providers.defaults.default_model` → `model`, `backend_refs[].type`
  removed, `prompt_guard.use_mmbert_32k` → `variant: mmbert32k`, and the
  jailbreak rule's retired `method: hybrid` → `classifier`. Verified by loading
  the file through `config.ParseYAMLBytes` in both the CPU and GPU phase.
- `docs/agent/` was removed from the repository, so the deploy-guide commits
  that targeted `docs/agent/nvidia-local.md` are dropped and the remaining
  references point at `website/docs/installation/nvidia-cuda.md`.
- The CUDA section of `bench/cpu-vs-gpu/README.md` is rewritten against the
  current README.

The rebase also surfaced a latent bug in the earlier revision: under
`set -euo pipefail` the warmup batch returned a non-zero status, so the script
exited before reaching the measurement phase. Verified against the previous
head and this one:

```
previous head : exit 1   (never reached the benchmark phase)
this branch   : exit 0   "reached the benchmark phase"
```

## Measured on an RTX 4090

Run end to end on a host where `docker run --gpus all nvidia-smi` works.

**With the image built from `src/vllm-sr/Dockerfile.cuda` as-is, the GPU phase
cannot start:**

```
model capability mismatch: candle capability: requested Candle device is
unavailable: cuda:0
```

The classifier signals resolve to the Candle provider, and `Dockerfile.cuda`
builds `candle-binding` with `--no-default-features`, which drops its default
`cuda` feature — the shipped `libcandle_semantic_router.so` is compiled against
candle's `dummy_cuda_backend.rs` and contains no `cudarc` references at all.
The ONNX Runtime provider cannot take up the slack: `ortOptions` accepts only
`cpu`, `rocm:index` and `migraphx:index`, the owned-instance `Provider` enum is
`Cpu | Migraphx | Rocm`, and `model_deployments.go` rejects `provider: ort`
with a CUDA device before that. Filed as #4004.

**Rebuilding `candle-binding` with its default `cuda` feature (CUDA devel
builder, everything else unchanged) makes the same harness run end to end**,
with all three classifiers reporting `candle cuda:0`:

| Signal | Tokens | CPU avg | GPU avg | Speedup |
|--------|-------:|--------:|--------:|--------:|
| domain | ~500 | 696 ms | 30.5 ms | 22.8x |
| domain | ~2,000 | 723 ms | 29.5 ms | 24.5x |
| domain | ~8,000 | 734 ms | 28.1 ms | 26.1x |
| domain | ~16,000 | 725 ms | 26.4 ms | 27.5x |
| jailbreak | ~500 | 814 ms | 37.4 ms | 21.8x |
| jailbreak | ~2,000 | 1921 ms | 61.2 ms | 31.4x |
| jailbreak | ~8,000 | 1765 ms | 63.4 ms | 27.8x |
| jailbreak | ~16,000 | 1953 ms | 62.8 ms | 31.1x |
| pii | ~500 | 1197 ms | 321 ms | 3.7x |
| pii | ~2,000 | 1952 ms | 462 ms | 4.2x |
| pii | ~8,000 | 1967 ms | 478 ms | 4.1x |
| pii | ~16,000 | 1982 ms | 450 ms | 4.4x |

Client-side end to end, same run: 1206 / 1962 / 1979 / 1996 ms on CPU versus
330 / 472 / 489 / 463 ms on GPU for 500 / 2K / 8K / 16K tokens. Latency is
flat across prompt sizes in both columns because the classifier caps input at
512 tokens; the sweep measures bounded classification plus full-prompt
tokenization, which is the intended behavior.

The concurrency sweep, same image, 15 s per level at ~1K tokens:

| Concurrency | CPU classified QPS | CPU p50 | GPU classified QPS | GPU p50 |
|---:|---:|---:|---:|---:|
| 1 | 0.6 | 1714 ms | 2.3 | 403 ms |
| 8 | 1.0 | 8000 ms | 4.3 | 1831 ms |
| 16 | 1.0 | 16278 ms | 4.2 | 3808 ms |
| 32 | 1.0 | 30697 ms | 3.9 | 8063 ms |

CPU throughput is flat at about 1 req/s — adding clients only queues them —
while the GPU sustains roughly 4 req/s. Note that this is far below the "80
req/s, 400x" figure in the branch's earlier revision; that number came from the
legacy ONNX path and is not reproducible.

Running this sweep end to end also surfaced something worth flagging on its
own: once the router saturates, a few requests complete successfully without
any signal extraction — 3 of 19 at concurrency 16 on CPU — and the router emits
no shed or error metric for them. Counting those as throughput would overstate
what the classifiers did, so each level now reports `ok`, `classified` and
`unclassified` and computes QPS over classified requests only. The generated
Envoy config also sets `failure_mode_allow: false`, so a request can never
reach the upstream by skipping ExtProc, and a level still fails outright on any
error or if nothing was classified.

Two caveats on reading these: averages come from the histogram `sum/count` and
are exact, but the percentile columns are interpolated inside the default
Prometheus buckets, which are coarse below 100 ms. And the numbers stay out of
the documentation until the CUDA image itself can produce them — a published
table that nobody can reproduce from `main` would be the same defect this PR's
gates exist to prevent.

## Test plan

- `go test ./src/semantic-router/pkg/config/` — includes the maintained-asset
  and docs contract tests that now cover `config-bench-cuda.yaml`.
- `shellcheck` on both scripts, `ruff` and `black --check` on both Python
  files, `yamllint` on the config, `markdownlint -c
  tools/linter/markdown/markdownlint.yaml` on the README.
- The failure-path runs quoted above.
- The full RTX 4090 run above: CPU phase and GPU phase, every gate satisfied,
  report written only after `verify_results()` passed.

### DIFF
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -102,6 +102,13 @@ versus ~1–3 s on CPU — a ~1–2 order-of-magnitude speedup. See
 [`deploy/nvidia/README.md`](../../deploy/nvidia/README.md) for the full table
 and caveats.
 
+Note on the "long-context" sweep: the classifier hard-caps input at 512 tokens
+(`MAX_CLASSIFICATION_SEQ_LEN` in `onnx-binding/src/model_architectures/classification/mmbert_classifier.rs`), so classifier latency
+is roughly flat across the 500→16K prompt sizes — the larger prompts exercise
+tokenization and the ext_proc path, not deeper inference. This is intended
+behavior, not a benchmark artifact; the sweep confirms latency stays bounded as
+prompts grow.
+
 ### Throughput / concurrency (NVIDIA CUDA)
 
 ext_proc classifies one request at a time (no batch knob), so throughput is
@@ -177,7 +184,11 @@ Reports are written to `results/`.
 | File | Description |
 |------|-------------|
 | `bench-3way.sh` | ONNX GPU vs ONNX CPU vs Candle CPU latency comparison |
-| `bench-long-context.sh` | CPU vs GPU, multi token-size, Prometheus metrics |
+| `bench-long-context.sh` | CPU vs GPU (AMD ROCm), multi token-size, Prometheus metrics |
+| `bench-cuda-long-context.sh` | CPU vs GPU (NVIDIA CUDA), multi token-size, Prometheus + e2e latency + speedup table |
+| `bench-cuda-throughput.sh` | CPU vs GPU (NVIDIA CUDA) throughput via concurrency sweep |
+| `load_test.py` | Concurrent request driver for `bench-cuda-throughput.sh` (stdlib only) |
+| `config-bench-cuda.yaml` | v0.3 router config for the CUDA benchmarks (all three signals) |
 | `bench-sdpa-vs-fa.sh` | SDPA vs FA on GPU, Prometheus metrics |
 | `bench-buffered-vs-streamed.sh` | BUFFERED vs STREAMED body mode, builds patched binary inside container |
 | `config-bench.yaml` | Canonical v0.3 router config template for ONNX benchmarks |
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -48,6 +48,55 @@ Compares ONNX CPU vs ROCm GPU across 500/2K/8K/16K token prompts:
 BENCH_IMAGE=semantic-router:rocm REQUESTS_PER_SIZE=10 ./bench-long-context.sh
 ```
 
+### CPU vs GPU (NVIDIA CUDA)
+
+NVIDIA counterpart of the above, using ONNX Runtime's CUDA Execution Provider.
+Same metric (`llm_signal_extraction_latency_seconds`), same 500/2K/8K/16K sizes,
+across the domain / jailbreak / PII classifiers.
+
+Prerequisites: NVIDIA driver + `nvidia-container-toolkit` (so `docker run
+--gpus all` works) and the CUDA router image built from
+[`src/vllm-sr/Dockerfile.cuda`](../../src/vllm-sr/Dockerfile.cuda) (see
+[`deploy/nvidia/README.md`](../../deploy/nvidia/README.md); default tag
+`vllm-sr-cuda:local`).
+
+Download the classifiers **and** the embedding model into `models/` using the
+directory names the bench mounts (note: no `-onnx` suffix, unlike the ROCm
+setup above):
+
+```bash
+python3 -c "
+from huggingface_hub import snapshot_download
+for repo in [
+    'mmbert32k-intent-classifier-merged',
+    'mmbert32k-jailbreak-detector-merged',
+    'mmbert32k-pii-detector-merged',
+    'mmbert-embed-32k-2d-matryoshka',
+]:
+    snapshot_download(f'llm-semantic-router/{repo}',
+        local_dir=f'bench/cpu-vs-gpu/models/{repo}',
+        allow_patterns=['onnx/*', '*.json'],
+        ignore_patterns=['*.safetensors', '*.bin', '*.pt'])
+"
+```
+
+Run it:
+
+```bash
+BENCH_IMAGE=vllm-sr-cuda:local REQUESTS_PER_SIZE=10 ./bench-cuda-long-context.sh
+```
+
+All ports are env-overridable (`EXTPROC_PORT`, `API_PORT`, `METRICS_PORT`,
+`ENVOY_PORT`, `STUB_PORT`) for hosts where the defaults (50051/8080/9190/8801/
+8091) are already bound. A built-in stub upstream answers on `STUB_PORT` so
+requests return 200; the signal-extraction metric is recorded regardless.
+
+Reference numbers (RTX 4090, `vllm-sr-cuda`): GPU signal extraction stays in the
+single-digit-to-low-tens of ms across 500–16K tokens for all three classifiers,
+versus ~1–3 s on CPU — a ~1–2 order-of-magnitude speedup. See
+[`deploy/nvidia/README.md`](../../deploy/nvidia/README.md) for the full table
+and caveats.
+
 ### SDPA vs Flash Attention
 
 Compares standard attention vs CK Flash Attention on GPU:
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -62,7 +62,8 @@ Prerequisites: NVIDIA driver + `nvidia-container-toolkit` (so `docker run
 
 Download the classifiers **and** the embedding model into `models/` using the
 directory names the bench mounts (note: no `-onnx` suffix, unlike the ROCm
-setup above):
+setup above). Run this from the repo root (the `local_dir` paths are
+repo-root-relative):
 
 ```bash
 python3 -c "
@@ -80,7 +81,7 @@ for repo in [
 "
 ```
 
-Run it:
+Run it (from `bench/cpu-vs-gpu/`):
 
 ```bash
 BENCH_IMAGE=vllm-sr-cuda:local REQUESTS_PER_SIZE=10 ./bench-cuda-long-context.sh
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -92,6 +92,10 @@ All ports are env-overridable (`EXTPROC_PORT`, `API_PORT`, `METRICS_PORT`,
 8091) are already bound. A built-in stub upstream answers on `STUB_PORT` so
 requests return 200; the signal-extraction metric is recorded regardless.
 
+The report (`results/report-cuda-*.md`) tables, per prompt size: client-side
+end-to-end latency (avg/P50/P95/min/max), the per-signal histogram latency
+(avg/P50/P95/P99) CPU vs GPU, and a CPU-vs-GPU speedup ratio per signal.
+
 Reference numbers (RTX 4090, `vllm-sr-cuda`): GPU signal extraction stays in the
 single-digit-to-low-tens of ms across 500–16K tokens for all three classifiers,
 versus ~1–3 s on CPU — a ~1–2 order-of-magnitude speedup. See
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -97,6 +97,22 @@ versus ~1–3 s on CPU — a ~1–2 order-of-magnitude speedup. See
 [`deploy/nvidia/README.md`](../../deploy/nvidia/README.md) for the full table
 and caveats.
 
+### Throughput / concurrency (NVIDIA CUDA)
+
+ext_proc classifies one request at a time (no batch knob), so throughput is
+measured via concurrency — N clients over a fixed duration, fixed prompt size,
+CPU vs GPU:
+
+```bash
+BENCH_IMAGE=vllm-sr-cuda:local CONCURRENCIES="1 8 16 32" ./bench-cuda-throughput.sh
+```
+
+`load_test.py` is the concurrent driver (stdlib only). Reference (RTX 4090):
+CPU throughput caps at ~0.2 req/s (added clients just queue — latency climbs
+from ~5 s to ~114 s), while GPU sustains ~80 req/s with sub-second P50 even at
+32-way concurrency (~400× throughput). Optional `CPUSET="10-19"` pins the router
+to specific cores on a busy host.
+
 ### SDPA vs Flash Attention
 
 Compares standard attention vs CK Flash Attention on GPU:
--- a/bench/cpu-vs-gpu/README.md
+++ b/bench/cpu-vs-gpu/README.md
@@ -57,7 +57,7 @@ across the domain / jailbreak / PII classifiers.
 Prerequisites: NVIDIA driver + `nvidia-container-toolkit` (so `docker run
 --gpus all` works) and the CUDA router image built from
 [`src/vllm-sr/Dockerfile.cuda`](../../src/vllm-sr/Dockerfile.cuda) (see
-[`deploy/nvidia/README.md`](../../deploy/nvidia/README.md); default tag
+[`docs/agent/nvidia-local.md`](../../docs/agent/nvidia-local.md); default tag
 `vllm-sr-cuda:local`).
 
 Download the classifiers **and** the embedding model into `models/` using the
@@ -99,7 +99,7 @@ end-to-end latency (avg/P50/P95/min/max), the per-signal histogram latency
 Reference numbers (RTX 4090, `vllm-sr-cuda`): GPU signal extraction stays in the
 single-digit-to-low-tens of ms across 500–16K tokens for all three classifiers,
 versus ~1–3 s on CPU — a ~1–2 order-of-magnitude speedup. See
-[`deploy/nvidia/README.md`](../../deploy/nvidia/README.md) for the full table
+[`docs/agent/nvidia-local.md`](../../docs/agent/nvidia-local.md) for the full table
 and caveats.
 
 Note on the "long-context" sweep: the classifier hard-caps input at 512 tokens
--- a/bench/cpu-vs-gpu/bench-cuda-long-context.sh
+++ b/bench/cpu-vs-gpu/bench-cuda-long-context.sh
@@ -86,14 +86,18 @@ generate_config() {
 
 generate_envoy() {
     # Reuse the shared envoy-bench.yaml, moving ext_proc/listener to our ports.
+    # The backend cluster is STATIC, so point it at the stub (STUB_PORT); the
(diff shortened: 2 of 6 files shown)


Xunzhuo:
Thanks for adding the CPU-vs-GPU benchmark path. The branch currently conflicts with main, so please rebase before we spend time on a deep review. After resolving the benchmark/deployment-doc conflicts, verify the scripts and load-test harness against current main, refresh any documented commands or numbers affected by the rebase, and post the validation commands/results here.

Xunzhuo (changes requested):
The benchmark records every long-context curl attempt even on transport or non-2xx failure, while the throughput driver counts arbitrary HTTP errors as processed without proving ext_proc ran. A broken setup can therefore publish plausible latency/QPS with zero valid signal samples. Please require expected HTTP success plus matching metric/request-count deltas and fail the run when they diverge.

FAUST-BENCHOU (commented):
Thanks for adding the NVIDIA CUDA benchmark coverage and the corresponding performance measurements. Having a CUDA counterpart to the existing CPU-vs-GPU harness helps provide a more complete performance evaluation path.

bench/cpu-vs-gpu/README.md
```diff
@@ -83,6 +85,67 @@ The scripts use the container names `sr-bench` and `envoy-bench` and remove
...
+for name in (
+    "mmbert32k-intent-classifier-merged",
+    "mmbert32k-jailbreak-detector-merged",
+    "mmbert32k-pii-detector-merged",
+    "mmbert-embed-32k-2d-matryoshka",
+):
+    snapshot_download(
+        repo_id=f"llm-semantic-router/{name}",
+        local_dir=f"bench/cpu-vs-gpu/models/{name}",
+        allow_patterns=["onnx/*", "*.json"],
```
Xunzhuo: [P2] Download weights required by the CUDA image

The documented CUDA image still builds the router without the `onnx` tag or a default-provider override, so these module defaults select Candle; the scripts' `AI_BINDING=onnx` does not change that selection. On this commit, parsing this template and calling the actual default execution policy with that environment produces `candle/cpu` and `candle/cuda:0`. The ModernBERT Candle loader opens `model.safetensors`, but this download explicitly excludes it. A fresh setup following this command therefore cannot initialize the classifiers in either phase. Please make the downloaded artifacts and configured execution provider agree (for example, include the required Candle weights, or explicitly configure a supported ORT/CUDA deployment); the device-only gate also accepts `provider=candle, device=cuda:0`.

Xunzhuo (changes requested):
Reviewed commit 12721e251132 against current main 54458c539cb1. The previous failed-request/signal-sample findings are addressed: current scripts reject failed HTTP requests, require signal metric deltas, and the late GPU-device gate accepts GPU events and rejects CPU/missing events. Isolated validation passed 18 Bash/Python fixture checks, the maintained canonical-config asset suite (55 subcases), and `make impact`. An executable check of the actual configuration/default-provider code confirms that `AI_BINDING=onnx` still selects Candle in both benchmark modes.

Requesting changes for the inline model-artifact/provider mismatch: the documented fresh setup excludes the weights required by that default backend. Independent review confirmed the same contract mismatch. No real NVIDIA GPU run, CUDA image build, model loading, or performance measurements were executed here; this review does not establish hardware benchmark results.

WUKUNTAI-0211:
@Xunzhuo @FAUST-BENCHOU rebased onto current `main` (the branch was 631 commits behind) and implemented the validation you asked for. Details are in the updated description; the short version:

**The gate.** Only HTTP 200 contributes to latency tables and QPS. `load_test.py` now reports `ok` / `http_err` / `conn_err` separately and exits non-zero when nothing succeeded; the long-context script drops non-200 rows and aborts the phase on any failure. Both scripts scrape `llm_signal_extraction_latency_seconds_count` around every measurement batch and require each of `domain` / `jailbreak` / `pii` to record at least one new sample per successful request, and the report is only written after `verify_results()` re-checks every cell. Failure-path evidence against stub upstreams:

```
load_test.py against a router that 503s every request
  row : 4 0 8495 0 0 0 0 0
  err : ERROR: no successful responses (8495 HTTP errors, 0 transport errors)
  exit: 1

send_requests + e2e statistics against the same router
  SEND_OK=0 SEND_FAILED=3
  csv: 1,17,503 / 2,5,503 / 3,6,503
  compute_e2e_stats -> 0 0 0 0 0 0      (previously: 3 samples, ~9 ms average)
```

The sample floor is one-per-successful-request rather than an exact match: 12 requests really do produce `domain 12`, `jailbreak 12`, `pii 24`, because the jailbreak and PII paths record a per-rule sample on top of the per-request one. Exact equality would fail healthy runs.

**What the gate then caught.** I added one more check — the GPU phase must see every classifier prepared on a non-CPU device — and re-measured on an RTX 4090. The CPU phase passes all gates; the GPU phase cannot start at all:

```
model capability mismatch: candle capability: requested Candle device is unavailable: cuda:0
```

That is not a harness problem. Classifier signals resolve to the Candle provider, and `src/vllm-sr/Dockerfile.cuda` builds `candle-binding` with `--no-default-features`, dropping its default `cuda` feature — the shipped library is compiled against candle's `dummy_cuda_backend.rs` and contains no `cudarc` references. ORT cannot cover for it either: `ortOptions` accepts only `cpu`, `rocm:index`, `migraphx:index`, and the owned-instance `Provider` enum is `Cpu | Migraphx | Rocm`, so `provider: ort` with a CUDA device is rejected at config validation — even though the ONNX binding itself implements a CUDA EP and the image ships it.

Rebuilding just `candle-binding` with its default `cuda` feature, everything else unchanged, makes the same harness run end to end with all three classifiers reporting `candle cuda:0`:

| Signal | CPU avg (500 / 2K / 8K / 16K) | GPU avg | Speedup |
|---|---|---|---|
| domain | 696 / 723 / 734 / 725 ms | 30.5 / 29.5 / 28.1 / 26.4 ms | 22.8–27.5x |
| jailbreak | 814 / 1921 / 1765 / 1953 ms | 37.4 / 61.2 / 63.4 / 62.8 ms | 21.8–31.4x |
| pii | 1197 / 1952 / 1967 / 1982 ms | 321 / 462 / 478 / 450 ms | 3.7–4.4x |

I am keeping those numbers out of the documentation until the CUDA image itself can produce them — a table nobody can reproduce from `main` is the same defect these gates exist to prevent — and filed #4004 for the CUDA gap with the verified fix. The old July numbers are removed rather than carried over: they came from the legacy `use_cpu` classifier path against a config that no longer parses.

Also fixed while rebasing: `config-bench-cuda.yaml` no longer loaded on `main` (`providers.defaults.default_model` → `model`, `backend_refs[].type` removed, `prompt_guard.use_mmbert_32k` → `variant: mmbert32k`, jailbreak `method: hybrid` → `classifier`), and it is now registered in `maintainedTemplatedConfigAssets` so future schema drift breaks CI instead of the benchmark. The earlier revision also had a `set -euo pipefail` bug that exited during warmup before any measurement ran.

WUKUNTAI-0211:
Follow-up: the concurrency sweep now runs end to end too, so the PR has both halves of the harness measured on the RTX 4090.

| Concurrency | CPU classified QPS | CPU p50 | GPU classified QPS | GPU p50 |
|---:|---:|---:|---:|---:|
| 1 | 0.6 | 1714 ms | 2.3 | 403 ms |
| 8 | 1.0 | 8000 ms | 4.3 | 1831 ms |
| 16 | 1.0 | 16278 ms | 4.2 | 3808 ms |
| 32 | 1.0 | 30697 ms | 3.9 | 8063 ms |

Running it surfaced one thing worth a second look, @Xunzhuo, because it is the failure mode your review described and it is real rather than hypothetical: once the router saturates, a few requests complete successfully without any signal extraction, and no shed or error metric is emitted for them. At concurrency 16 on CPU, 3 of 19 successful requests were never classified — `llm_request_outcomes_total{outcome="success"}` moved by 19 while `llm_signal_extraction_latency_seconds_count{signal_type="domain"}` moved by 16, with `llm_model_admission_outcomes_total` showing 16 admissions and no shed.

The harness now treats that as a measurement rather than trusting it: every level reports `ok`, `classified` and `unclassified`, QPS is computed over classified requests only, the generated Envoy config sets `failure_mode_allow: false` so ExtProc can never be skipped on the way upstream, and a phase refuses to start measuring until five consecutive requests are confirmed classified. A level still fails outright on any HTTP or transport error, or if nothing was classified.

This is not load shedding: the bench config sets no `max_concurrency`, so `resourceAdmission` hands back `admission.Noop`, and the level's `llm_model_admission_outcomes_total` shows 16 `admitted` and no `shed`. I also ruled out the harness itself — the generated Envoy config has `failure_mode_allow: false`, the closing snapshot waits for the counters to settle, and a warm stack driven at 8, 16 and 32 concurrency matches one sample per request exactly. The candidate I am left with is the classifier fail-open path (`on_error` defaults to allow), but I could not find a matching error counter moving, so I would rather not guess. Happy to open a separate issue if you consider it a router bug — you will recognise the path faster than I will.

bench/cpu-vs-gpu/README.md
WUKUNTAI-0211: Correct. I hit the same failure while re-measuring: with only `onnx/*` downloaded the router exits with

```
runtime_initializer_failed :: prepare default/pii_classifier: candle load:
path: "models/mmbert32k-pii-detector-merged/model.safetensors" No such file or directory (os error 2)
```

I worked around it by downloading the weights locally and never carried the fix back into the README, which is what your review caught. Fixed in 2104e748e. The three classifiers now fetch `*.safetensors` plus their mapping JSON, and only the embedding directory keeps `onnx/*`. Resolving the documented pattern against the repositories gives the loader everything it opens — `model.safetensors`, `config.json`, `tokenizer.json`, `tokenizer_config.json`, `special_tokens_map.json`, and the `category_mapping.json` / `jailbreak_type_mapping.json` / `pii_type_mapping.json` the config points at.

I also dropped `-e AI_BINDING=onnx` from both scripts in the same commit. You are right that it selects nothing: these scripts start `/usr/local/bin/router` directly, so the entrypoint that reads the variable never runs, and leaving it in implied a backend choice that does not happen.

On the provider itself: the CUDA image cannot run these classifiers on the GPU at all, because `Dockerfile.cuda` builds `candle-binding` with `--no-default-features` and so ships candle's dummy CUDA backend. I filed #4004 for that with a verified fix, and the measurements in this PR come from an image rebuilt that way. The device gate is what stopped the run from quietly reporting a CPU-versus-CPU comparison.

WUKUNTAI-0211: For whenever you come back to this: the re-review base has moved to `2c95c3e0c`, three commits past the `12721e251` you reviewed.

- `2104e748e` — the artifact fix above.
- `3c05b9ba8` — the long-context script now also generates Envoy with `failure_mode_allow: false`; the throughput script already did. A bypassed request would otherwise return a fast 200 whose wall clock is not a classification measurement. Re-ran both phases after the change: 12/12 per size, every sample accounted for, speedup unchanged within run-to-run noise.
- `2c95c3e0c` — rebase onto current `main`.

Product Verification is green on that head.

PR #2327, opened 2026-07-05
[Docs] Add CUDA CPU-vs-GPU perf benchmark to NVIDIA deploy guide

WUKUNTAI-0211:
Closes #2348

## What

Follow-up to #2281 (CUDA `Dockerfile.cuda` + NVIDIA deployment guide). During
review, @Xunzhuo asked for CUDA perf numbers (latency / throughput / batch size
/ context-length pressure) to land in the README as a follow-up. This adds a
**Performance (CPU vs GPU)** section to `deploy/nvidia/README.md`.

## Details

Measured with the same harness and Prometheus metric as the existing ROCm
`bench/cpu-vs-gpu/` benchmark (`llm_signal_extraction_latency_seconds`), ported
to CUDA (`--gpus all`), on an RTX 4090 — a consumer-GPU baseline.

- **Latency** — signal-extraction, CPU vs GPU, all three classifiers
  (domain / jailbreak / PII) across 500 / 2K / 8K / 16K tokens. GPU stays
  single-digit-to-low-tens of ms; CPU sits in the ~1–2.7 s band (~65–240×).
- **Throughput** — ext_proc classifies one request at a time (no batch knob),
  so the analog is concurrency: CPU caps at ~0.2 req/s (requests queue, latency
  explodes) while GPU sustains ~80 req/s sub-second (~400×).
- **Context-length pressure** — bounded by design: the classifier caps input at
  512 tokens (`MAX_CLASSIFICATION_SEQ_LEN`, the `547adc6e` OOM fix), so
  classifier latency is decoupled from prompt length. Documented honestly.
- Reconciles the existing "CPU ≈ GPU" note: that parity is the **embedding**
  model with 2D-Matryoshka layer early-exit; the full-depth **classifier** is a
  large GPU win.

Docs-only change.

## Test Plan

- `markdownlint -c tools/linter/markdown/markdownlint.yaml deploy/nvidia/README.md`
- Verified all in-page anchors and relative links (`bench/cpu-vs-gpu/`,
  `onnx-binding/README.md`, `mmbert_classifier.rs`) resolve.
- Numbers cross-checked against raw Prometheus histogram dumps (per-request
  N=12), the router `[Perf] classifier inference` log line, and a direct ONNX
  Runtime CPU-vs-CUDA EP micro-benchmark — three independent paths agree.

PR #2329, opened 2026-07-05
[Docs] Clarify NVIDIA Step 5 uses the Envoy front-door port

WUKUNTAI-0211:
## What

Docs-only clarity fix for `deploy/nvidia/README.md`. The Step 5 sanity-check
curl used a bare `http://localhost:8999/...` port that is never defined anywhere
else in the playbook. Readers hit two questions with no answer:

- Where does `8999` come from? (It is not one of the host ports remapped in
  Step 3c: `50151` / `8180` / `9290`.)
- Why does it differ from the router API port used later in Troubleshooting
  (`localhost:8180/api/v1/classify/intent`)?

The answer is that chat completions enter through the **Envoy** front door of
the existing `vllm-sr` deployment, not the router's own API port — a distinction
the playbook already relies on but never states here.

## Change

- Replace the magic `8999` with an `${ENVOY_PORT}` placeholder.
- Add a one-line note that Step 5 targets the deployment's Envoy listener (not
  the Step 3c router API port), matching the env-var port convention already
  used in [`deploy/amd/README.md`](../amd/README.md).

No behavior change; documentation only.

## Test Plan

- `markdownlint -c tools/linter/markdown/markdownlint.yaml deploy/nvidia/README.md` — clean.
- Verified the `${ENVOY_PORT}` reference and the Step 3c cross-reference read
  correctly against the surrounding steps.

### DIFF
--- a/deploy/nvidia/README.md
+++ b/deploy/nvidia/README.md
@@ -330,10 +330,13 @@ the active config.
 
 ## Step 5: Sanity-check a real query
 
-Send any prompt through the router and confirm it routes successfully:
+Send any prompt through the router and confirm it routes successfully. Chat
+completions enter through the **Envoy** front door of your existing `vllm-sr`
+deployment — not the router's own API port from Step 3c — so use that
+deployment's Envoy listener port, shown here as `${ENVOY_PORT}`:
 
 ```bash
-curl -sS -X POST http://localhost:8999/v1/chat/completions \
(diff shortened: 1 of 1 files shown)


WUKUNTAI-0211:
Folding this into #2327 — the Step 5 Envoy-port clarity fix now rides along with the perf-section PR rather than as a separate PR touching the same file.

Issue #2377, opened 2026-07-05
[Feature] Add TensorRT-LLM and Triton TensorRT-LLM inference-aware backend adapter


Xunzhuo:
### Goal

Add a TensorRT-LLM / Triton TensorRT-LLM adapter for inference-aware backend routing once the engine-neutral backend telemetry and replica identity contract is defined.

Depends on: #2349
Related: #2350, #2351, #2352, #2353, #2359, #2372, #2376
Related external docs: https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/getting_started/trtllm_user_guide.html, https://nvidia.github.io/TensorRT-LLM/advanced/kv-cache-reuse.html, https://nvidia.github.io/TensorRT-LLM/blogs/tech_blog/blog5_Disaggregated_Serving_in_TensorRT-LLM.html
Related code: `src/semantic-router/pkg/config`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/latency`, `src/semantic-router/pkg/inflight`, `src/semantic-router/pkg/extproc`, `src/vllm-sr/cli/templates/envoy.template.yaml`, `deploy/operator`, `deploy/kubernetes`, `deploy/recipes`, `src/fleet-sim`, observability and replay diagnostics

### Proposed Workgroup

Data Plane & Networking

### Scope

Define and implement, or first design if telemetry surfaces need confirmation, TensorRT-LLM support for:

- backend and replica identity mapping for direct TensorRT-LLM and Triton TensorRT-LLM deployments;
- endpoint discovery and config representation using canonical `providers.models[].backend_refs[]`;
- telemetry ingestion for health, queue depth or scheduler backlog, active requests, TTFT/TPOT/E2E latency, GPU/memory pressure, and Triton/TensorRT-LLM metrics where available;
- KV cache reuse, block reuse, prefix/cache affinity, and disaggregated prefill/decode signals where exposed by the deployment;
- stale or missing telemetry fallback to existing model-level routing behavior;
- diagnostics that distinguish TensorRT-LLM telemetry freshness, selected backend/replica, and backend policy reason;
- examples or docs for direct TensorRT-LLM, Triton TensorRT-LLM, and how this differs from Dynamo-managed serving.

##### Motivation

[...]

Issue #2353, opened 2026-07-05
[Feature] Add Envoy endpoint-selection integration for backend-aware routing


Xunzhuo:
### Goal

Define how backend-aware routing decisions are enforced at the Envoy/upstream endpoint boundary after the router selects a logical model and backend/replica candidate.

Related code: `src/semantic-router/pkg/extproc`, `src/vllm-sr/cli/templates/envoy.template.yaml`, `deploy/operator/controllers/semanticrouter_envoy.go`, `deploy/operator/controllers/backend_discovery.go`

### Proposed Workgroup

Data Plane & Networking

### Scope

Evaluate and implement the least invasive endpoint-control mechanism:

- router-selected backend ref / endpoint metadata emitted to Envoy;
- Envoy cluster or subset routing configuration generated by CLI/operator templates;
- optional feedback path where Envoy reports the chosen upstream endpoint back to the router;
- behavior for unsupported gateway profiles;
- replay/trace fields for requested backend, actual backend, and fallback reason.

##### Motivation

Current Envoy integration lets the router mutate request metadata and logical model routing, while Envoy generally performs endpoint load balancing inside the selected upstream cluster. Backend-aware routing needs an explicit control boundary; otherwise the router cannot guarantee queue-depth, GPU-load, or KV-cache-affinity placement.

### Non-goals

- Do not require all gateways to support deterministic endpoint pinning on day one.
- Do not make endpoint pinning bypass health checks or load-shedding.
- Do not hide endpoint-selection fallback from diagnostics.

### Completion signal

- The router/Envoy contract for backend-aware routing is explicit and testable.
- Envoy config generation can support backend refs or subset routing without breaking existing model-level routing.
- Replay can show requested backend, actual backend when known, and fallback reason.

Parent Epic: #2332

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #2352, opened 2026-07-05
[Feature] Add ROCm ATOM inference-aware backend adapter


Xunzhuo:
### Goal

Add a ROCm ATOM adapter for inference-aware backend routing so AMD self-hosted deployments can expose ATOM runtime, cache, and profiling signals through the same backend policy used by other engines.

Depends on: engine-neutral backend telemetry contract
Related external project: https://github.com/ROCm/ATOM
Related code: `src/semantic-router/pkg/config`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/latency`, `deploy/amd`, `deploy/operator`, `src/fleet-sim`

### Proposed Workgroup

Data Plane & Networking

### Scope

Define and implement ATOM support for:

- ATOM backend/replica identity under a logical model;
- OpenAI-compatible serving endpoint integration;
- queue, latency, GPU, health, prefix-cache, and disaggregation-related telemetry where available;
- ROCm/AMD deployment examples under `deploy/amd` or related Kubernetes profiles;
- diagnostics that distinguish ATOM telemetry freshness and fallback from vLLM/SGLang.

##### Motivation

ATOM is a ROCm-focused lightweight vLLM-like implementation with OpenAI-compatible serving, prefix caching, P/D disaggregation, profiling, and benchmark surfaces. Those capabilities are directly relevant to self-hosted AMD routing, but they should enter vLLM Semantic Router through a normalized engine adapter rather than through ATOM-specific routing branches.

### Non-goals

- Do not require ATOM for default CPU CI.
- Do not duplicate ATOM's benchmark dashboard inside the router.
- Do not make AMD-specific telemetry fields part of the generic backend policy unless they are normalized.

### Completion signal

- ATOM backends can be modeled as backend-aware candidates without changing logical recipe/decision behavior.
- ATOM telemetry maps to the normalized backend telemetry contract.
- Missing telemetry falls back to existing routing behavior.
- AMD/ATOM examples document opt-in setup and validation expectations.

Parent task: #2349
Parent Epic: #2332

Issue #2351, opened 2026-07-05
[Feature] Add SGLang inference-aware backend adapter


Xunzhuo:
### Goal

Add an SGLang-specific adapter for inference-aware backend routing once the engine-neutral backend telemetry contract is defined.

Depends on: engine-neutral backend telemetry contract
Related code: `src/semantic-router/pkg/config`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/latency`, `deploy/operator`, `deploy/kubernetes`

### Proposed Workgroup

Data Plane & Networking

### Scope

Define and implement SGLang support for:

- backend/replica identity mapping;
- queue, active request, latency, health, and memory/cache telemetry ingestion;
- optional cache-affinity signals if exposed by the deployment;
- config and operator integration points that do not rename existing vLLM-only fields prematurely;
- diagnostics for adapter availability and telemetry freshness.

##### Motivation

Self-hosted deployments often mix serving engines. SGLang can expose different runtime and scheduling signals from vLLM, so the router needs an adapter boundary that maps SGLang metrics into the same normalized backend policy instead of adding one-off model-selection branches.

### Non-goals

- Do not require SGLang in default smoke tests.
- Do not fork the logical model-selection contract for SGLang.
- Do not add SGLang-specific fields to core selection structures if normalized fields are sufficient.

### Completion signal

- SGLang backends can participate in backend-aware routing through the normalized telemetry contract.
- Stale/missing SGLang telemetry degrades to existing model-level routing behavior.
- Docs/examples show how an SGLang deployment opts into the adapter.

Parent task: #2349
Parent Epic: #2332

theohsiung:
hi @haowu1234 I'm also interested in this one, maybe we can work together?

haowu1234:
> hi [@haowu1234](https://github.com/haowu1234) I'm also interested in this one, maybe we can work together?

Shall I assign this sglang issue to you?

theohsiung:
thx alot, glad to discuss this further

PR #2391, opened 2026-07-07
[Inference] Add backend telemetry contract

### LINKED ISSUE #2349: [Feature] Define engine-neutral backend telemetry and replica identity
### Goal

Define the engine-neutral contract that inference-aware backend routing will consume before adding engine-specific integrations for vLLM, SGLang, ATOM, or future serving engines.

Related code: `src/semantic-router/pkg/config`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/inflight`, `src/semantic-router/pkg/latency`, `src/semantic-router/pkg/extproc`, `deploy/operator/controllers/backend_discovery.go`

### Proposed Workgroup

Data Plane & Networking

### Scope

Define a normalized backend observation contract covering:

- logical model name, backend ref, endpoint, and replica identity;
- serving engine kind and version when available;
- queue depth, active requests, GPU utilization, memory pressure, and health;
- TTFT/TPOT/E2E latency samples keyed by backend/replica;
- KV cache, prefix cache, session, or prefix affinity hints;
- freshness, TTL, confidence, and fail-open behavior;
- diagnostics emitted to replay, response headers or traces where appropriate.

##### Motivation

The current router can select a logical model and already has model-level latency, in-flight, multi-factor, and cache-affinity signals. It does not yet have a normalized backend/replica identity model or a telemetry schema that can represent queue depth, GPU load, KV/cache affinity, health, and freshness across different inference engines.

Without this layer, engine-specific work will either duplicate logic or leak vLLM/SGLang/ATOM details into the router hot path.

##### Validation

```bash
make agent-report ENV=cpu CHANGED_FILES="src/semantic-router/pkg/config src/semantic-router/pkg/selection src/semantic-router/pkg/inflight src/semantic-router/pkg/latency src/semantic-router/pkg/extproc deploy/operator website/docs"
make agent-lint CHANGED_FILES="src/semantic-router/pkg/config src/semantic-router/pkg/selection src/semantic-router/pkg/inflight src/semantic-router/pkg/latency src/semantic-router/pkg/extproc deploy/operator website/docs"
make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/config src/semantic-router/pkg/selection src/semantic-router/pkg/inflight src/semantic-router/pkg/latency src/semantic-router/pkg/extproc deploy/operator website/docs"
```

Parent Epic: #2332

### Non-goals

- Do not implement an engine-specific collector here.
- Do not require synchronous telemetry fetches on the request path.
- Do not replace logical model selection or provider config.

### Completion signal

- A backend telemetry schema and identity model can represent vLLM, SGLang, and ATOM without engine-specific fields in core selection logic.
- Missing/stale telemetry behavior is explicit and testable.
- Selection code can consume backend telemetry as a second-stage backend policy after logical model selection.
- Replay/diagnostics can explain selected model, selected backend/replica, telemetry freshness, and policy reason.

haowu1234:
## Summary

This PR adds the engine-neutral backend telemetry foundation for backend-aware routing. It deliberately stops at the shared contract layer: future adapter PRs can publish normalized telemetry into `pkg/backend`, and future endpoint-selection PRs can consume it as a second-stage policy, without changing today's extproc routing behavior.

Closes #2349.
Refs #2377.

## Design At A Glance

| Layer | This PR Adds | Purpose |
| --- | --- | --- |
| Backend identity | `BackendIdentity`, `EngineKind`, optional `backend_id` / `engine_kind` config fields | Give each physical backend or replica a stable engine-neutral name. |
| Telemetry contract | `BackendTelemetry`, `BackendLatency`, `BackendAffinity`, `HealthState` | Normalize queue, load, GPU/memory pressure, latency, cache affinity, health, confidence, and TTL signals. |
| TTL store | `Store`, package-level helpers, 5s default TTL | Keep the latest telemetry sample per model/backend/replica and reject stale data during policy evaluation. |
| Adapter seam | `TelemetryAdapter`, `AdapterConfig`, `AdapterTarget`, `Registry` | Let vLLM/SGLang/ATOM/TensorRT-style collectors register behind one factory shape. |
| Policy helper | `SelectBackendCandidate` and `BackendPolicyResult` diagnostics | Provide the minimal second-stage selection primitive while preserving fail-open behavior. |
| Config propagation | `CanonicalBackendRef.backend_id`, `CanonicalBackendRef.engine_kind`, runtime `VLLMEndpoint` fields, CLI/dashboard/docs updates | Preserve backend identity hints through canonical load, runtime normalization, export, migration, and UI surfaces. |

## Data Flow

```mermaid
flowchart LR
    A[Engine adapter in follow-up PR] -->|Collect normalized BackendTelemetry| B[pkg/backend Store]
    C[Existing model/backend candidates] --> D[SelectBackendCandidate]
    B -->|fresh healthy samples only| D
    D -->|BackendPolicyResult + diagnostics| E[Future endpoint-selection integration]
    D -->|FailOpen=true when telemetry unusable| F[Existing weight / Envoy LB fallback]
```

Important: this PR only adds the contract, store, registry, and helper. The final integration box is intentionally deferred.

## Backend Contract

The main public contract lives in `src/semantic-router/pkg/backend`:

- `types.go` defines the normalized identity and telemetry model:
  - `EngineKind`: `vllm`, `sglang`, `atom`, `tensorrt-llm`
  - `BackendIdentity`: backend ID, model name, replica ID, endpoint, engine kind/version
  - `BackendTelemetry`: queue depth, active requests, GPU/memory/KV pressure, latency, affinity, health, confidence, collection time, TTL
  - `BackendPolicyDiagnostics`: replay/debug-ready explanation of selected or fail-open behavior
- `store.go` provides a thread-safe latest-sample TTL store:
  - key shape is model/backend/replica
  - missing `CollectedAt` defaults to the store clock
  - per-sample `TTL` overrides the default; otherwise the default is 5s
  - stale data is retained as raw telemetry but filtered by `GetFresh` / policy evaluation
- `registry.go` provides the adapter factory seam:
  - adapters implement `EngineKind()` and `Collect(ctx)`
  - constructors are registered per `EngineKind`
  - follow-up PRs can add concrete collectors without changing policy consumers

## Minimal Policy Semantics

`SelectBackendCandidate(modelName, candidates, store)` is intentionally conservative. It only selects when at least one candidate has fresh, selectable telemetry.

| Case | Result | Reason |
| --- | --- | --- |
| No backend candidates | Fail open | `no_backend_candidates` |
| Candidate has no telemetry | Fail open if nothing else is usable | `missing_telemetry` |
| Candidate telemetry is expired | Fail open if nothing else is usable | `stale_telemetry` |
| Candidate telemetry is unhealthy | Reject for selection | `all_unhealthy` if no healthy candidate remains |
| At least one fresh healthy/degraded/unknown sample | Select one candidate | `fresh_selectable_telemetry` |

Current scoring order:

1. lowest `QueueDepth`
2. lowest `ActiveRequests`
3. highest backend weight
4. stable backend ID ordering
5. stable replica ID ordering

The helper returns diagnostics for both selected and fail-open paths. This gives later endpoint-selection work a stable result shape without forcing extproc or Envoy changes into this PR.

## Config And Compatibility

This PR adds optional backend identity hints to canonical backend refs:

```yaml
backend_refs:
  - name: local-primary
    backend_id: qwen3-8b-local-primary
    engine_kind: vllm
    endpoint: 127.0.0.1:8000
    protocol: http
    weight: 80
```

Those fields are preserved through:

- canonical YAML parsing;
- runtime `VLLMEndpoint` normalization;
- canonical export / round-trip;
- Python CLI Pydantic schema and legacy migration helpers;
- dashboard config normalization and model backend display;
- docs and sample config.

Existing configs remain valid because both fields are optional and omitted from export when empty.

## Non-Goals In This PR

This PR intentionally does not:

- add a concrete vLLM/SGLang/ATOM/TensorRT collector;
- start polling metrics endpoints;
- wire backend telemetry into extproc request routing;
- emit `x-vsr-selected-backend` or any Envoy endpoint pinning header;
- change `ResolvePrimaryBackendForModel` semantics;
- change current weighted backend fallback or Envoy LB behavior.

Missing, stale, or unhealthy telemetry remains fail-open by design.

## Reviewer Guide

Suggested review order:

1. `src/semantic-router/pkg/backend/types.go` for the public contract and diagnostics shape.
2. `src/semantic-router/pkg/backend/store.go` for TTL/freshness behavior and keying.
3. `src/semantic-router/pkg/backend/registry.go` for the adapter seam used by future engine collectors.
4. `src/semantic-router/pkg/backend/policy.go` for conservative selection and fail-open behavior.
5. `src/semantic-router/pkg/config/*` for canonical load/export preservation.
6. `src/vllm-sr/cli/*` and `dashboard/frontend/src/*` for schema/UI propagation.
7. Tests under `src/semantic-router/pkg/backend`, `src/semantic-router/pkg/config`, and `src/vllm-sr/tests`.

## Follow-Up Path

This PR is meant to unblock:

- #2350: vLLM adapter maps `/metrics` into `BackendTelemetry` and writes to the store.
- #2351 / #2352: SGLang and ATOM adapters reuse the same registry/store/policy shapes.
- #2377: TensorRT/Triton-style telemetry can map into the same normalized contract instead of introducing TensorRT-specific policy logic.
- #2353: Envoy endpoint selection can consume `BackendPolicyResult` once the adapter shape is stable.

## Validation

- `cd src/semantic-router && go test ./pkg/backend ./pkg/config`
- `python3 -m pytest src/vllm-sr/tests/test_config_file_loaders.py::test_parse_user_config_preserves_backend_identity_hints src/vllm-sr/tests/test_config_migrate.py::test_migrate_config_data_promotes_legacy_lora_catalog_and_backend_refs`
- `make vllm-sr-test`
- `make test-semantic-router`
- `make dashboard-check`
- `env CGO_ENABLED=1 HOME=/tmp/vsr-agent-home CARGO_HOME=/tmp/vsr-agent-cargo RUSTUP_HOME=/tmp/vsr-agent-rustup PATH=/tmp/vsr-agent-cargo/bin:$PATH make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/backend src/semantic-router/pkg/config src/vllm-sr/cli src/vllm-sr/tests dashboard/frontend/src/pages dashboard/frontend/src/types config website/docs website/versioned_docs website/i18n/zh-Hans/docusaurus-plugin-content-docs"`
- `make agent-serve-local ENV=cpu AGENT_STACK_NAME=codex-backend-telemetry AGENT_PORT_OFFSET=49`
- `make agent-smoke-local ENV=cpu AGENT_STACK_NAME=codex-backend-telemetry AGENT_PORT_OFFSET=49`
- `make agent-stop-local ENV=cpu AGENT_STACK_NAME=codex-backend-telemetry AGENT_PORT_OFFSET=49`

### DIFF
--- a/src/semantic-router/pkg/backend/policy.go
+++ b/src/semantic-router/pkg/backend/policy.go
@@ -1,0 +1,275 @@
+package backend
+
+import (
+	"math"
+	"sort"
+	"strings"
+)
+
+const (
+	PolicyReasonSelectedFreshTelemetry = "fresh_selectable_telemetry"
+
+	FallbackReasonMissingTelemetry    = "missing_telemetry"
+	FallbackReasonStaleTelemetry      = "stale_telemetry"
+	FallbackReasonAllUnhealthy        = "all_unhealthy"
+	FallbackReasonNoBackendCandidates = "no_backend_candidates"
+)
+
+type policyOption struct {
+	candidate BackendCandidate
+	telemetry BackendTelemetry
+}
+
+type candidateEvaluation struct {
+	option         policyOption
+	selectable     bool
+	missing        bool
+	stale          bool
+	fresh          bool
+	telemetryAgeMS int64
+	unhealthyCount int
+}
+
+// SelectBackendCandidate applies a minimal second-stage backend policy. It only
+// selects when at least one candidate has fresh, selectable telemetry; otherwise
+// it fails open so existing model/backend routing can continue unchanged.
+func SelectBackendCandidate(modelName string, candidates []BackendCandidate, store *Store) BackendPolicyResult {
+	if store == nil {
+		store = defaultStore
+	}
+
+	diag := BackendPolicyDiagnostics{
+		SelectedModel:  modelName,
+		CandidateCount: len(candidates),
+	}
+	if len(candidates) == 0 {
+		return failOpenResult(diag, FallbackReasonNoBackendCandidates)
+	}
+
+	options := []policyOption{}
+	missingCount := 0
+	staleCount := 0
+	unhealthyCount := 0
+
+	for _, candidate := range candidates {
+		evaluation := evaluateCandidate(modelName, candidate, store)
+		if evaluation.missing {
+			missingCount++
+			continue
+		}
+		if evaluation.stale {
+			staleCount++
+			diag.TelemetryAgeMS = maxInt64(diag.TelemetryAgeMS, evaluation.telemetryAgeMS)
+			continue
+		}
+		if evaluation.fresh {
+			diag.FreshCandidateCount++
+			diag.TelemetryAgeMS = maxInt64(diag.TelemetryAgeMS, evaluation.telemetryAgeMS)
+		}
+		unhealthyCount += evaluation.unhealthyCount
+		if !evaluation.selectable {
+			continue
+		}
+		options = append(options, evaluation.option)
+	}
+
+	diag.UnhealthyCount = unhealthyCount
+	diag.TelemetryFresh = diag.FreshCandidateCount > 0
+	if len(options) == 0 {
+		return failOpenResult(diag, fallbackReasonForEmptyPolicy(candidates, missingCount, staleCount, unhealthyCount))
+	}
+
+	sort.Slice(options, func(i, j int) bool {
+		return policyOptionLess(options[i], options[j])
+	})
+	selected := options[0]
+	age := store.Age(selected.telemetry)
+	diag.SelectedBackendID = selected.candidate.BackendID
+	diag.SelectedReplicaID = selected.telemetry.Identity.ReplicaID
+	diag.TelemetryFresh = true
+	diag.TelemetryAgeMS = age.Milliseconds()
+	diag.PolicyReason = PolicyReasonSelectedFreshTelemetry
+
+	return BackendPolicyResult{
+		SelectedBackendID: selected.candidate.BackendID,
+		SelectedReplicaID: selected.telemetry.Identity.ReplicaID,
+		FailOpen:          false,
+		Reason:            PolicyReasonSelectedFreshTelemetry,
+		Diagnostics:       diag,
+	}
+}
+
+func evaluateCandidate(modelName string, candidate BackendCandidate, store *Store) candidateEvaluation {
+	candidate = normalizeCandidate(modelName, candidate)
+	if candidate.BackendID == "" || candidate.ModelName == "" {
+		return candidateEvaluation{missing: true}
+	}
+
+	rawTelemetry := store.ListByBackend(candidate.ModelName, candidate.BackendID)
+	if candidate.ReplicaID != "" {
+		rawTelemetry = filterTelemetryByReplica(rawTelemetry, candidate.ReplicaID)
+	}
+	if len(rawTelemetry) == 0 {
+		return candidateEvaluation{missing: true}
+	}
+
+	freshTelemetry := filterFreshTelemetry(rawTelemetry, store)
+	if len(freshTelemetry) == 0 {
+		return candidateEvaluation{stale: true, telemetryAgeMS: telemetryAgeMS(rawTelemetry, store)}
+	}
+
+	selectableTelemetry, unhealthyCount := filterSelectableTelemetry(freshTelemetry)
+	if len(selectableTelemetry) == 0 {
+		return candidateEvaluation{fresh: true, telemetryAgeMS: telemetryAgeMS(freshTelemetry, store), unhealthyCount: unhealthyCount}
+	}
+
+	sort.Slice(selectableTelemetry, func(i, j int) bool {
+		return telemetryLess(selectableTelemetry[i], selectableTelemetry[j])
+	})
+	return candidateEvaluation{
+		option:         policyOption{candidate: candidate, telemetry: selectableTelemetry[0]},
+		selectable:     true,
+		fresh:          true,
+		telemetryAgeMS: telemetryAgeMS(selectableTelemetry, store),
+		unhealthyCount: unhealthyCount,
+	}
+}
+
+func filterFreshTelemetry(items []BackendTelemetry, store *Store) []BackendTelemetry {
+	fresh := make([]BackendTelemetry, 0, len(items))
+	for _, telemetry := range items {
+		if store.IsFresh(telemetry) {
+			fresh = append(fresh, telemetry)
+		}
+	}
+	return fresh
+}
+
+func filterSelectableTelemetry(items []BackendTelemetry) ([]BackendTelemetry, int) {
+	selectable := make([]BackendTelemetry, 0, len(items))
+	unhealthyCount := 0
+	for _, telemetry := range items {
+		if isSelectableHealth(telemetry.Health) {
+			selectable = append(selectable, telemetry)
+		} else {
+			unhealthyCount++
+		}
+	}
+	return selectable, unhealthyCount
+}
+
+func failOpenResult(diag BackendPolicyDiagnostics, reason string) BackendPolicyResult {
+	diag.PolicyReason = reason
+	diag.FallbackReason = reason
+	return BackendPolicyResult{
+		FailOpen:    true,
+		Reason:      reason,
+		Diagnostics: diag,
+	}
+}
+
+func normalizeCandidate(modelName string, candidate BackendCandidate) BackendCandidate {
+	candidate.BackendID = strings.TrimSpace(candidate.BackendID)
+	candidate.ReplicaID = strings.TrimSpace(candidate.ReplicaID)
+	candidate.ModelName = strings.TrimSpace(candidate.ModelName)
+	candidate.EndpointName = strings.TrimSpace(candidate.EndpointName)
+	if candidate.ModelName == "" {
+		candidate.ModelName = strings.TrimSpace(modelName)
+	}
+	if candidate.BackendID == "" {
+		candidate.BackendID = candidate.EndpointName
+	}
+	return candidate
+}
+
+func filterTelemetryByReplica(items []BackendTelemetry, replicaID string) []BackendTelemetry {
+	filtered := make([]BackendTelemetry, 0, len(items))
+	for _, telemetry := range items {
+		if telemetry.Identity.ReplicaID == replicaID {
+			filtered = append(filtered, telemetry)
+		}
+	}
+	return filtered
+}
+
+func fallbackReasonForEmptyPolicy(candidates []BackendCandidate, missingCount, staleCount, unhealthyCount int) string {
+	if unhealthyCount > 0 && unhealthyCount == len(candidates) {
+		return FallbackReasonAllUnhealthy
+	}
+	if staleCount > 0 {
+		return FallbackReasonStaleTelemetry
+	}
+	if missingCount > 0 {
+		return FallbackReasonMissingTelemetry
+	}
+	if unhealthyCount > 0 {
+		return FallbackReasonAllUnhealthy
+	}
+	return FallbackReasonMissingTelemetry
+}
+
+func policyOptionLess(a, b policyOption) bool {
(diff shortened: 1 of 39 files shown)


src/semantic-router/pkg/backend/policy.go
```diff
@@ -0,0 +1,275 @@
...
+	selectable := make([]BackendTelemetry, 0, len(items))
+	unhealthyCount := 0
+	for _, telemetry := range items {
+		if isSelectableHealth(telemetry.Health) {
+			selectable = append(selectable, telemetry)
+		} else {
+			unhealthyCount++
+		}
+	}
+	return selectable, unhealthyCount
```
FAUST-BENCHOU: The `all_unhealthy` statistical approach treats "telemetry entry count" as "candidate backend count," which can mislead the fallback reason in multi-replica scenarios. I feel it's more reasonable to determine whether at least one replica under this backend is selectable, and to count by the candidate dimension.

haowu1234: Thanks, good catch. Addressed in 422e50ad.

I changed the policy evaluation to classify each candidate once as missing/stale/unhealthy/selectable before computing the fallback reason. A backend candidate is now considered selectable if any fresh replica telemetry under it is selectable; `unhealthy_count` only counts candidates with fresh telemetry but no selectable replica, so replica-entry count no longer affects `all_unhealthy`.

I also added regression coverage for:
- multiple unhealthy replicas under one backend plus missing telemetry for another backend -> `missing_telemetry`, not `all_unhealthy`
- all unhealthy backends -> `all_unhealthy` counted by candidate/backend dimension
- mixed replicas under one backend (`unhealthy`, `healthy`) -> selects the healthy replica

Validated with:
- `go test ./pkg/backend -count=1`
- `go test ./pkg/backend ./pkg/config -count=1`
- isolated `make test-semantic-router`

FAUST-BENCHOU:
i get no other comments,just solve newest conflicts

Xunzhuo:
Thanks for pushing this forward. This PR is the foundational backend/config contract that #2402 and the other backend adapters build on, but the branch is currently conflicting with main. Please rebase onto the latest main, resolve the config/dashboard/runtime conflicts deliberately, rerun the affected gates, and add a short summary of any contract/API changes introduced by the rebase. We will review and land this contract before investing in the dependent adapters.

haowu1234:
Rebased this backend telemetry contract branch onto latest upstream main (116087dc). Conflict resolution preserved main's current config/dashboard/runtime contract updates, including provider pricing/cache-write docs, remote embedding/model catalog fields, and the structured Dashboard model editor, while keeping this PR's additive backend telemetry identity hints on `providers.models[].backend_refs[]`: `backend_id` and `engine_kind`.

Validation run after rebase:
- `make dashboard-check` passed.
- `VLLM_SR_STACK_NAME=codex-pr2391-rebase-test make vllm-sr-test` passed (33 executed, 6 skipped), isolated from existing local runtime containers.
- Targeted router contract tests passed from `src/semantic-router`: `go test ./pkg/backend ./pkg/config` with the local Rust binding libs on the library path.
- `make agent-ci-gate ...` reached global Go lint/typecheck, but the current workspace binding surface fails outside this PR scope on missing `MultiModalEmbeddingOutput` in `candle-binding/semantic-router_mock.go` (and full `make test-semantic-router` hits the same binding/typecheck surface).

Xunzhuo:
backend_refs:
  - name: local-primary
    backend_id: qwen3-8b-local-primary
    engine_kind: vllm
    endpoint: 127.0.0.1:8000
    protocol: http
    weight: 80


Rename engine_kind -> runtime

And I think backend_id is a bad abstraction, i think the backend ref should support the endpoint slice, how can we express:

Static multiple endpoints per model per inference engine 

And

Dynamic endpoints per model per inference engine?

haowu1234:
@Xunzhuo Thanks, I agree with this direction. The current `backend_id + engine_kind + endpoint` shape feels too much like it is describing a single physical endpoint. It does not naturally express multiple replicas, and it does not leave enough room for dynamic endpoint sources.

I suggest adjusting the contract as follows:

- Rename `engine_kind` to `runtime`.
- Treat each `providers.models[].backend_refs[]` entry as a runtime backend pool / binding, not as a single physical endpoint.
- Do not make `backend_id` the public primary abstraction. Endpoint / replica identity should be derived from the endpoint source and then normalized internally.
- Add two mutually exclusive endpoint sources:
  - `endpoints`: a static endpoint list, for local development, bare metal, CI, Docker Compose, and simple non-Kubernetes deployments.
  - `discovery`: dynamic endpoint discovery. It should use a typed block structure, with Kubernetes Service-based discovery as the first supported source.

Static multiple-endpoint example:

```yaml
providers:
  models:
    - name: qwen3-8b
      backend_refs:
        - name: local-vllm
          runtime: vllm
          protocol: http
          endpoints:
            - name: primary-a
              endpoint: 127.0.0.1:8000
              metrics_endpoint: 127.0.0.1:8000
              weight: 80
            - name: primary-b
              endpoint: 127.0.0.1:8001
              metrics_endpoint: 127.0.0.1:8001
              weight: 20
```

Dynamic Kubernetes discovery example:

```yaml
providers:
  models:
    - name: qwen3-8b
      backend_refs:
        - name: local-vllm
          runtime: vllm
          protocol: http
          discovery:
            type: kubernetes
            kubernetes:
              service: qwen3-vllm
              namespace: inference
              endpoint_port: http
              metrics_port: metrics
```

Semantics:

- `backend_refs[].name` names the backend pool, for example `local-vllm`.
- `runtime` identifies the serving runtime family, for example `vllm`, `sglang`, or `tensorrt-llm`. It drives telemetry adapter selection and runtime-specific metrics normalization.
- `endpoints[]` explicitly lists the static members of the pool.
- `discovery` describes how the pool members are discovered dynamically.
- `discovery.type` is the resolver discriminator; the matching typed block, for example `discovery.kubernetes`, carries resolver-specific configuration.

For Kubernetes discovery, the router/operator should not treat the Service VIP as one opaque backend. Instead, the Service should be the stable user-facing pool name, while the resolver reads the backing EndpointSlices and normalizes ready endpoints into the same internal shape as static endpoints.

Internally, both static YAML and Kubernetes discovery should normalize to one common structure, roughly:

```go
type BackendEndpoint struct {
    ModelName        string
    BackendRefName   string
    Runtime          string
    Name             string
    ReplicaID        string
    InferenceAddress string
    MetricsAddress   string
    Weight           int
    Labels           map[string]string
    Source           string
}
```

With this shape, the telemetry store and backend policy do not need to know whether a target came from YAML or Kubernetes. They only consume normalized endpoint / replica candidates.

Suggested validation rules:

- `runtime` is required for backend-aware self-hosted refs.
- Exactly one of `endpoints` or `discovery` must be set.
- Endpoint names must be unique within one backend ref.
- When `discovery.type = kubernetes`, `discovery.kubernetes` must be set.
- `discovery.kubernetes` requires `service`, `namespace`, and `endpoint_port`; `metrics_port` can only be omitted when the runtime has a safe default.
- Under one `discovery` block, only the typed block matching `type` may be set, so resolver-specific fields do not get mixed.
- Missing or stale telemetry must fail open to the existing model-level routing / LB behavior.

This contract can express both cases you called out:

1. static multiple endpoints per model per inference runtime;
2. dynamic endpoints per model per inference runtime.

It also avoids freezing user-provided `backend_id` as the long-term primary abstraction before the endpoint selection and discovery layers are settled.

Xunzhuo (changes requested):
The telemetry store copies `BackendTelemetry` only shallowly, while `SessionAffinity` and `ExtraHints` remain mutable maps. Callers can mutate stored state after `Upsert` and mutate the store through values returned by `Get` or `List`. Please deep-clone the nested maps at both boundaries and add regressions for both directions.

FAUST-BENCHOU (commented):
Thanks for introducing the backend telemetry contract and keeping the first step focused on the shared abstraction layer. Separating telemetry collection from future routing integration makes the design easier to extend.

PR #2399, opened 2026-07-07
[Bench] Add ml-binding to LD_LIBRARY_PATH in all performance targets

TwilightTechie:
Fixes #2398

(Fix 1 of 2 for the broken perf gate — this PR restores the benchmarks by fixing the library-load failure. The follow-up #2400 makes future failures visible in CI; the reporting half is tracked separately in its linked issue.)

## What

The perf benchmark binary fails to start in CI because `libml_semantic_router.so` is not on the loader path:

```
benchmarks.test: error while loading shared libraries: libml_semantic_router.so:
cannot open shared object file: No such file or directory
exit status 127
FAIL    github.com/vllm-project/semantic-router/perf/benchmarks    0.001s
```

`perf/benchmarks` transitively imports `pkg/modelselection`, which links `ml-binding` (`#cgo LDFLAGS: -lml_semantic_router`, `ml-binding/ml_binding.go:15`), but every perf target in `tools/make/performance.mk` exported only `candle-binding` and `nlp-binding`. `make rust-ci` does build `ml-binding`, so the library exists in CI — it just isn't findable at load time.

## Before / After (same one-line change in all 7 perf targets)

```make
# before
@export LD_LIBRARY_PATH=${PWD}/candle-binding/target/release:${PWD}/nlp-binding/target/release && \
# after
@export LD_LIBRARY_PATH=${PWD}/candle-binding/target/release:${PWD}/ml-binding/target/release:${PWD}/nlp-binding/target/release && \
```

This matches the canonical three-binding form already used throughout `tools/make/rust.mk`.

## Test Plan

- `make perf-bench-decision` passes end-to-end with the updated loader path (5 decision benchmarks, exit 0).
- `make agent-lint CHANGED_FILES="tools/make/performance.mk"` passes.
- The rendered recipe (`make -n perf-bench-quick`) shows the three-path export.

Issue #2398, opened 2026-07-07
[Bench] Performance benchmark CI gate is silently broken: benchmarks fail to load libml_semantic_router but the job reports success


TwilightTechie:
## Description

While working on PR [#2389](https://github.com/vllm-project/semantic-router/pull/2389) I noticed, The component performance benchmark job (`performance-test.yml` → `make perf-bench-quick`) is silently broken on `main`. The benchmark binary fails to start in CI, but the workflow still concludes `success` and posts a "Component benchmarks completed successfully ✅" comment on PRs — so the performance gate that runs on every PR currently protects nothing. Observed in CI logs of green runs on `main` (e.g. run 28784602526, 2026-07-06).

## Steps to Reproduce

1. Open the logs of any recent green `performance-test` workflow run on `main` (e.g. `gh run view 28784602526 --log`) and search for `FAIL`.
2. Or locally on Linux: `make rust-ci && make perf-bench-quick` → exit 127 loading `libml_semantic_router.so`.

## Error / Evidence

From the "Run component benchmarks" step of a run whose job concluded **success**:

```
/tmp/go-build…/benchmarks.test: error while loading shared libraries: libml_semantic_router.so:
cannot open shared object file: No such file or directory
exit status 127
FAIL    github.com/vllm-project/semantic-router/perf/benchmarks    0.001s
FAIL
```

The same run then posted to the PR:

```
## Performance Benchmark Results
Component benchmarks completed successfully.
- Classification benchmarks: ✅
- Decision engine benchmarks: ✅
- Cache benchmarks: ✅
```

## Code Path

Three independent causes:

1. **Missing `ml-binding` on the loader path** — every perf target in `tools/make/performance.mk` (lines 17, 28, 36, 42, 48, 95, 131) exports `LD_LIBRARY_PATH=${PWD}/candle-binding/target/release:${PWD}/nlp-binding/target/release`. `perf/benchmarks` transitively imports `pkg/modelselection`, which links `ml-binding` (`#cgo LDFLAGS: -lml_semantic_router`, `ml-binding/ml_binding.go:15`). The canonical three-path export already exists in `tools/make/rust.mk:83-116`, and `make rust-ci` does build `ml-binding` — the library is present in CI, just not on the loader path.
[...]

PR #2400, opened 2026-07-07
[CI/Build] Fail performance-test on benchmark failure and report honest per-suite results

### LINKED ISSUE #2401: [CI/Build] performance-test workflow masks benchmark failures and posts a hardcoded success comment
## Description

Split from #2398. Even after the benchmark library-load failure is fixed (#2398 / #2399), the `performance-test` workflow cannot catch the next regression: the `component-benchmarks` job passes regardless of whether the benchmarks succeed, and the PR comment always claims success. This is how the failure documented in #2398 stayed invisible on `main`.

## Steps to Reproduce

1. Open the logs of any green `performance-test` run on `main` prior to the #2399 fix (e.g. run 28784602526, 2026-07-06) — the "Run component benchmarks" step contains `FAIL ... exit status 127`, yet the job concluded `success`.
2. The same run posted "Component benchmarks completed successfully ✅ / Classification ✅ / Decision ✅ / Cache ✅" to the PR.

## Error / Evidence

Inside the green job:

```
/tmp/go-build…/benchmarks.test: error while loading shared libraries: libml_semantic_router.so:
cannot open shared object file: No such file or directory
exit status 127
FAIL    github.com/vllm-project/semantic-router/perf/benchmarks    0.001s
FAIL
```

PR comment posted by the same run:

```
## Performance Benchmark Results
Component benchmarks completed successfully.
- Classification benchmarks: ✅
- Decision engine benchmarks: ✅
- Cache benchmarks: ✅
```

## Code Path

- `.github/workflows/performance-test.yml:86` — `make perf-bench-quick 2>&1 | tee reports/bench-output.txt` runs in the default GitHub Actions shell (no `pipefail`), so `tee`'s exit 0 masks a non-zero exit from `make`.
- `.github/workflows/performance-test.yml:96-114` — the "Generate performance summary" step writes hardcoded "✅" text for all three suites regardless of results; the "Comment PR with results" step publishes it.

- Expected: a benchmark failure fails the `component-benchmarks` job, and the PR comment reflects actual per-suite results.
- Actual: the job stays green on failure and the comment always claims success.

## Proposed Fix

`set -o pipefail` in the benchmark step; derive the summary headline from the bench step outcome and per-suite ✅/❌/"not run" status by parsing `bench-output.txt` (`--- FAIL:` markers vs result lines); run summary/comment steps with `if: always()` so a failed run updates the PR comment to ❌ (the job conclusion still fails). Fix is ready in PR #2400.

TwilightTechie:
Fixes #2401

Part of the perf-gate problem tracked in #2398 (fix 2 of 2 — merge after #2399, which fixes the underlying library-load failure; merging this first would turn the perf job red on all open PRs until #2399 lands)

## What

The `component-benchmarks` job has been green on `main` while the benchmark suite fails at binary load (see #2398 for evidence from a green run). Two masking layers in `.github/workflows/performance-test.yml`:

1. `make perf-bench-quick 2>&1 | tee reports/bench-output.txt` runs without `pipefail`, so `tee`'s exit 0 wins and the step passes even when `make` fails.
2. The summary step writes hardcoded `Classification/Decision/Cache: ✅` text regardless of results, and the PR comment publishes it.

## Fix

- `set -o pipefail` in the benchmark step, so benchmark failures fail the job.
- Align the step's `LD_LIBRARY_PATH` with the canonical three-binding form (candle, ml, nlp) used by `tools/make/rust.mk`.
- Derive the summary from reality: headline from the bench step outcome, per-suite ✅ / ❌ / "⚠️ not run" by grepping `bench-output.txt` for result lines vs `--- FAIL:` markers.
- Run the summary/comment steps with `if: always()` so a failed run updates the PR comment to ❌ instead of leaving a stale success comment. **This does not soften the gate** — the job conclusion still fails via the bench step; `always()` only ensures the failure is reported visibly in the PR conversation.

## Before / After

```yaml
# before — failure masked, comment hardcoded
- name: Run component benchmarks
  run: |
    mkdir -p reports
    export LD_LIBRARY_PATH=${PWD}/candle-binding/target/release
    make perf-bench-quick 2>&1 | tee reports/bench-output.txt
# summary step: static text
#   - Classification benchmarks: ✅   (always, even on FAIL)

# after — failure fails the job, comment derived from output
- name: Run component benchmarks
  id: bench
  run: |
    set -o pipefail
    mkdir -p reports
    export LD_LIBRARY_PATH=${PWD}/candle-binding/target/release:${PWD}/ml-binding/target/release:${PWD}/nlp-binding/target/release
    make perf-bench-quick 2>&1 | tee reports/bench-output.txt
# summary step: per-suite status from bench-output.txt
#   "--- FAIL: BenchmarkCache…" → ❌ · result line present → ✅ · absent → ⚠️ not run
```

## Test Plan

- YAML parses; `yamllint` with the repo config (`tools/linter/yaml/.yamllint`) is clean.
- `make agent-lint CHANGED_FILES=".github/workflows/performance-test.yml"` passes.
- `suite_status` grep logic exercised against real benchmark logs for all three outcomes: a passing decision-bench log → ✅, a failing cache-bench log (`--- FAIL:` marker) → ❌, absent suite → "⚠️ not run".


TwilightTechie:
The failing component-benchmarks check here is expected and is this PR working as intended: set -o pipefail now surfaces the pre-existing libml_semantic_router.so load failure (#2398) that previous runs reported as green. It will pass once [#2399](https://github.com/vllm-project/semantic-router/pull/2399) (the library-path fix) merges — please review/merge that one first, then re-run this job. This run is live proof of the masking bug: identical FAIL output was present in every green run of this job on main.
cc: @Xunzhuo

drivebyer (approved):

PR #2418, opened 2026-07-08
[Router] add headless Service so the bundled HPA balances the Envoy ext_proc backend

### LINKED ISSUE #2417: feature: headless Service option so the bundled HPA balances the Envoy ext_proc (gRPC) backend
### Describe the feature

The Helm chart (`deploy/helm/semantic-router`) bundles an HPA for the router Deployment
(`templates/hpa.yaml`, `autoscaling.*`) but exposes the router only through a **ClusterIP** Service
(`templates/service.yaml` → `spec.type: {{ .Values.service.type }}`, default `ClusterIP`; there is
no `clusterIP`/headless knob).

Please add a chart option to expose the router as a **headless** Service (`clusterIP: None`) — or
support `publishNotReadyAddresses` / a `service.headless` flag — so a gRPC `ext_proc` client can
resolve per-pod endpoints (`STRICT_DNS`) and client-side load-balance across replicas.

Minimal shape:

```yaml
# values.yaml
service:
  type: ClusterIP
  clusterIP: ""   # set to "None" for a headless Service (per-pod endpoints for gRPC ext_proc clients)
```

```yaml
# templates/service.yaml
spec:
  type: {{ .Values.service.type }}
  {{- with .Values.service.clusterIP }}
  clusterIP: {{ . | quote }}
  {{- end }}
```

### Primary layer

unsure

### Why this layer?

This isn't a signal/plugin/global-level (classification-core) feature — it's a Helm-chart /
deployment-packaging concern: how the router's Service is exposed so the bundled HPA can actually
distribute load. Marking "unsure" because the layer options describe the router core, not the chart.

For the router's primary use — an Envoy `ext_proc` gRPC backend (AI Gateway / EnvoyPatchPolicy) — a
single ClusterIP VIP plus HTTP/2 stream multiplexing pins **every** ext_proc stream onto **one**
backing pod. The other HPA-scaled replicas receive no traffic, so the bundled HPA can't distribute
load: it scales up on the hot pod's CPU, but the new pods never get streams.

Verified on a live cluster: with the ClusterIP Service the HPA scaled to its max but only one pod
served (the rest stayed near-idle); pointing the ext_proc cluster at a headless Service (per-pod
endpoints via `STRICT_DNS`) immediately spread the streams across all replicas.

### Additional context

**Why it happens:** an ext_proc client (Envoy) holds long-lived HTTP/2 connections and multiplexes
many concurrent streams over them. Against a ClusterIP VIP, kube-proxy balances at *connection*
setup and H2 keeps reusing the same connection → the same pod. Client-side round-robin/least-request
can't help, because the VIP resolves to a single address.

**Workaround:** downstream we add a sibling headless Service in our umbrella chart that selects the
router's pod labels, and point the Envoy ext_proc cluster (`STRICT_DNS` + `LEAST_REQUEST`) at it. It
works but couples us to the subchart's selector labels; a first-class chart option would let us drop
the sibling.

**Environment:** chart `deploy/helm/semantic-router` (main) — `service.type: ClusterIP`,
`autoscaling.enabled` present; client Envoy AI Gateway `ext_proc` (gRPC, port `50051`).

xiongzubiao:
Closes #2417

## Purpose

- **What does this PR change?** Adds an unconditional `<fullname>-headless` Service (`clusterIP: None`) to the router Helm chart. It mirrors the primary Service's selector and exposes the gRPC port (`grpc` 50051), hardcodes `type: ClusterIP`, and carries a `service: headless` label. The primary ClusterIP Service is left untouched.
- **Why is this change needed?** For the router's primary use — an Envoy `ext_proc` gRPC backend — the client holds long-lived HTTP/2 connections and multiplexes every stream over them. Against the single ClusterIP VIP, kube-proxy balances only at connection setup, so all ext_proc streams pin to one pod: the bundled HPA (`autoscaling.*`) scales up but the new replicas stay idle. Pointing the `ext_proc` cluster at a headless Service (`STRICT_DNS` + `LEAST_REQUEST`) resolves per-pod endpoints so Envoy client-side load-balances across replicas. (Verified on a live cluster in #2417.) The `type: ClusterIP` is hardcoded rather than inheriting `service.type`, so a `LoadBalancer`/`NodePort` primary still renders a valid headless spec.
- **Which module(s) does this affect?** `Router` — Helm chart / deployment packaging (`deploy/helm/semantic-router/templates/service.yaml`).

## Test Plan

- `helm lint deploy/helm/semantic-router` — the CI gate in `helm-publish.yml`.
- `helm template … --show-only templates/service.yaml` on default values — assert the headless Service renders with `clusterIP: None`, `type: ClusterIP`, the `grpc` port only, the router selector, and the `service: headless` label; the primary and metrics Services are unchanged.
- `helm template … --set service.type=LoadBalancer` — assert the primary renders as `LoadBalancer` while the headless stays `ClusterIP` + `clusterIP: None` (proves the hardcoded type keeps a `LoadBalancer`/`NodePort` primary valid).
- **Why sufficient:** the change is purely additive and default-inert (the new Service resolves per-pod DNS but carries no traffic until a client targets it); no routing/startup/config/API behavior changes. `helm lint` plus the render assertions cover template correctness, and the existing Helm-based E2E continues to deploy the additive Service.

## Test Result

- `helm lint`: `1 chart(s) linted, 0 chart(s) failed`.
- Default render: `Service/<name>-headless` renders with `clusterIP: None`, `type: ClusterIP`, the `grpc`(50051) port, the router selector, and the `service: headless` label; the primary and metrics Services are byte-identical to before.
- `--set service.type=LoadBalancer`: primary `type: LoadBalancer`, headless still `type: ClusterIP` + `clusterIP: None`.
- Not exercised on a live cluster in this PR (no kube context); the load-spreading behavior was confirmed on a live cluster by the reporter in #2417.
- Follow-up/risk: `<name>-headless` becomes a stable part of the chart's surface. No new E2E load-distribution assertion (would require a multi-replica load generator). Local `helm lint`/`template` were run against a dependency-stripped copy of the chart since subchart deps are not vendored; the change does not touch dependencies.

---
<details>

</details>


deploy/helm/semantic-router/templates/service.yaml
```diff
@@ -20,6 +20,32 @@ spec:
     {{- include "semantic-router.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/component: router
 ---
+# Headless per-pod DNS for gRPC ext_proc client-side load balancing (issue #2417).
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "semantic-router.fullname" . }}-headless
```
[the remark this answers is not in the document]
xiongzubiao: Intentional — the `-headless` name follows the chart's existing suffix convention (`-metrics`, `-config`, `-dashboard`), none of which reserve room for the suffix; `-dashboard` (10 chars) is in fact longer than `-headless` (9). Truncating only this name would be inconsistent, and the 63-char DNS-label limit affects all suffixed resource names equally, so guarding it belongs in a separate chart-wide change rather than this PR. Leaving as-is here for consistency.

deploy/helm/semantic-router/templates/service.yaml
```diff
@@ -20,6 +20,32 @@ spec:
...
+  clusterIP: None
+  ports:
+  - port: {{ .Values.service.grpc.port }}
+    targetPort: {{ .Values.service.grpc.targetPort }}
+    protocol: {{ .Values.service.grpc.protocol }}
+    name: grpc
+  - port: {{ .Values.service.api.port }}
+    targetPort: {{ .Values.service.api.targetPort }}
+    protocol: {{ .Values.service.api.protocol }}
+    name: classify-api
```
FAUST-BENCHOU: Why we need to export this?Seems ext_proc only uses gRPC 50051

xiongzubiao: Good call — dropped it. The headless Service now exposes only the gRPC port (50051). The classify-api (8080) is plain request/response REST that the ClusterIP VIP already load-balances well, and per-pod DNS wouldn't help it anyway (an ordinary HTTP client resolves the name and picks one address; only a client doing client-side LB over the records, like Envoy, benefits — which is the gRPC ext_proc case). Happy to add the port back if a concrete per-pod/fan-out need for the REST API comes up; it's a non-breaking additive change. Pushed as d31c3c65.

drivebyer (approved):
Non-blocking: the repository’s Envoy manifests still use the original ClusterIP, so users following the official installation guide will not automatically benefit from this Service. Please consider updating the relevant manifests and documentation in this PR or a follow-up.

PR #2428, opened 2026-07-09
[observability]: Improve semantic router defaults

### LINKED ISSUE #2427: Improve Semantic Router production observability defaults
## Summary

Semantic Router already exposes useful Prometheus metrics, OpenTelemetry traces, and Grafana assets, but the default observability surface has a few gaps that make production triage harder than it needs to be.

This issue tracks consolidating the existing observability into a reliable operator-facing surface for routing, model selection, inflight load, latency/error windows, session transitions, and cache warmth.

## Problems observed

- Tracing sampling config accepts values that look valid, but `traceidratio` style names can fall through to the unknown sampler fallback instead of a ratio sampler.
- Some documentation and examples refer to stale `semantic_router_*` metric names instead of the metrics actually exported by the router.
- The default Grafana dashboard does not make key router health signals easy to inspect, especially routing latency, model selection duration, inflight requests, windowed model health, request errors, session transitions, and cache warmth.
- Kubernetes/OpenShift/Helm users do not have a default PrometheusRule set for common Semantic Router symptoms such as high error rate, high TTFT/TPOT/completion latency, high routing latency, high inflight requests, or low cache hit rate.
- There is limited contract testing around the key metrics that dashboards and alerts depend on.

## Desired outcome

- Sampling config supports the canonical `probabilistic` value and compatibility aliases without surprising fallback behavior.
- Docs use the current `llm_*`, `rag_*`, and `sr_*` metric families.
- Default Grafana dashboards can show routing, selection, inflight, request errors, windowed latency/error/queue/utilization, session transition, and cache warmth signals.
- Kubernetes/OpenShift/Helm deployments can opt into baseline Prometheus alert rules.
- Metric and dashboard contracts have lightweight regression coverage.

## Validation notes

A CPU smoke deployment was also used to validate that Prometheus and Jaeger can receive data and that the added Grafana panels have live series when the relevant router features are enabled.

haowu1234:
## Summary

Closes #2427.

This PR strengthens the default Semantic Router observability surface without removing or renaming existing metrics. It focuses on making the current Prometheus, OpenTelemetry, Grafana, Kubernetes/OpenShift, and Helm assets more useful for production triage.

## What changed

- Add tracing sampling compatibility for `traceidratio` and `trace_id_ratio`, both mapped to `TraceIDRatioBased(rate)`, while keeping `probabilistic` as the recommended value.
- Emit a structured warning when an unknown tracing sampling type falls back to always-on sampling.
- Add metric contract tests for key router/model Prometheus metrics used by dashboards and alerts.
- Extend the generated Grafana dashboard with panels for routing latency, model selection duration/counts, inflight requests, request errors, top failing models, windowed model health, session transitions, and cache warmth.
- Add Kubernetes and OpenShift PrometheusRule manifests.
- Add Helm values/schema/template support for the same alert rules.
- Update observability docs to use the actual metric families and add runbook-style guidance for the new alerts.

## Operator impact

Operators should be able to answer the common first questions directly from the default assets:

- Is routing latency high?
- Is model selection slow or skewed by method/tier?
- Which model has inflight pressure?
- Are requests failing, and why?
- Are model latency/error/utilization signals degrading over short windows?
- Are session transitions or cache warmth contributing to behavior?

## Validation

- `python3 -m pytest src/vllm-sr/tests/test_grafana_dashboard_template.py` passed.
- `git diff --check upstream/main...HEAD` passed.
- `go test ./pkg/observability/tracing ./pkg/observability/metrics ./pkg/selection` partially passed locally:
  - `pkg/observability/tracing` passed.
  - `pkg/observability/metrics` passed.
  - `pkg/selection` did not build in this local environment because native binding packages are unavailable/mismatched (`ml-binding: build constraints exclude all Go files`, `candle-binding` missing `MultiModalEmbeddingOutput`).
- CPU smoke deployment validated live Prometheus and Jaeger data with the branch image:
  - Prometheus had `llm_model_completion_latency_seconds`, `llm_model_ttft_seconds`, `llm_model_tpot_seconds`, `llm_model_routing_latency_seconds`, and `llm_model_selection_duration_seconds` series.
  - Jaeger exposed the `vllm-sr` service and router spans.
  - Grafana panels rendered live data after generating request, error, and inflight traffic.


deploy/kubernetes/observability/prometheus/rules.yaml
```diff
@@ -0,0 +1,90 @@
...
+  semantic-router-alerts.yml: |
+    groups:
+      - name: semantic-router-observability
+        rules:
+          - alert: HighRequestErrorRate
+            expr: |
+              (
+                sum(rate(llm_request_errors_total[5m]))
+                /
+                clamp_min(sum(rate(llm_model_requests_total[5m])) + sum(rate(llm_request_errors_total[5m])), 1)
```
FAUST-BENCHOU: why add errors here?I think llm_model_requests_total is enough in fact

src/vllm-sr/cli/templates/grafana_observability_sections.py
```diff
@@ -0,0 +1,317 @@
...
+    ctx.panels.append(
+        create_row_panel("Windowed Model Health", y=ctx.y_pos, panel_id=575)
+    )
+    ctx.y_pos += 1
+
+    append_windowed_latency_error_panels(ctx)
+    append_windowed_queue_utilization_panels(ctx)
+
+
+def append_windowed_latency_error_panels(ctx) -> None:
```
FAUST-BENCHOU, line 171: seems dashboard's config miss 
```
      metrics:
        enabled: true
        windowed_metrics:
          enabled: true
```
maybe we need to add in  deploy/kubernetes/observability/dashboard/config.yaml?

Xunzhuo (approved):

deploy/kubernetes/observability/prometheus/rules.yaml
haowu1234: Good catch, agreed. `llm_model_requests_total` represents request attempts rather than successful-only requests, so adding `llm_request_errors_total` to the denominator can double-count failed requests and under-report the error rate.

src/vllm-sr/cli/templates/grafana_observability_sections.py
haowu1234: Agreed. The dashboard includes windowed metrics panels, so the bundled observability config should enable `global.services.observability.metrics.windowed_metrics` by default.

PR #2432, opened 2026-07-10
[CI/Build][Bindings] Wire MULTIMODAL_MODEL_PATH so model-gated multimodal tests execute in Test And Build

### LINKED ISSUE #2319: [CI/Build] Model-gated multimodal tests never execute in CI: MULTIMODAL_MODEL_PATH is set nowhere (Go and Rust)
## What

Every multimodal test that needs the real model is gated on `MULTIMODAL_MODEL_PATH` and skips: on the Go side, `TestEmbeddingClassifier_Integration*` (`pkg/classification/embedding_classifier_integration_test.go:54-56`); on the candle-binding side, the Go binding tests (`TestMultiModalEncodeImageFromBase64` and siblings in `candle-binding/semantic-router_test.go`) skip on the same variable, and the Rust unit tests in `multimodal_embedding.rs` are `#[ignore]`-gated and read it too. No workflow or makefile sets the variable, so these tests skip on every PR run and on main. #2186 added the CI-safe Rust lib lane (closing #2182) - that lane excludes model-dependent tests by design, so this gap is the complement of that fix, not a regression of it.

## Why it matters

During the #2104 review, the FFI image-encode path had zero executed coverage anywhere - verification had to be manual against a local model. Green CI on image-touching PRs currently proves the validation and extraction layers only; an encode-path regression would ship invisibly. The tests exist and would fail for the right reasons - nothing runs them.

## Proposal

Options by cost: (a) a nightly job that fetches `multi-modal-embed-small` via the router's `--download-only` path (`make download-models` with a config referencing it) and exports `MULTIMODAL_MODEL_PATH`; (b) a small fixture checkpoint for PR CI (fastest feedback, requires producing one); (c) at minimum, a documented manual-receipt convention for PRs touching the multimodal FFI. Happy to take this once there's a preference on lane cost.

## H2 execution scope

Make multimodal model-gated tests either run in CI with a supported fixture or produce an explicit manual receipt policy. This is part of #2347 release validation.

Relevant code surfaces: Go/Rust multimodal tests, CI workflows, model artifact setup, and repo harness validation docs.

## Acceptance criteria

- CI no longer silently skips all multimodal tests because `MULTIMODAL_MODEL_PATH` is unset.
- The project either supplies a lightweight fixture path or documents manual receipt requirements.
- Skipped multimodal tests report why they are skipped and what validation replaces them.
- Release validation can prove at least one meaningful multimodal encode path works.

shraderdm:
## Purpose

Closes #2319 (multimodal CI lane).

Every multimodal test that needs the real model gates on `MULTIMODAL_MODEL_PATH` and skips today because nothing sets it. The gap turned out to be cheaper to close than the issue assumed: the `test-and-build` job already downloads `multi-modal-embed-small` on every run. `make test` runs `make download-models`, and `config/config.yaml` references `models/mom-embedding-multimodal` under `model_catalog`, so the router's `--download-only` pass fetches it. The log of a regular run (28784602794) shows the model downloading successfully and the gated tests skipping two steps later in the same job. The model was already there; only the environment variable was missing.

What this PR does:

- Exports `MULTIMODAL_MODEL_PATH` in the `test-and-build` job. With that, two suites execute on every `test-and-build` run (PRs matching the job's path filters - core, make, ci, helm, e2e, docker - non-draft only, pushes to main, and the nightly schedule):
  - `TestEmbeddingClassifier_Integration*` in `src/semantic-router/pkg/classification/`, via `test-semantic-router`. Hermetic: a synthetic in-source PNG through the real candle-binding FFI image-encode path.
  - The hermetic multimodal binding tests (`TestMultiModalEmbeddingInit`, `TestMultiModalEncodeText`, `TestMultiModalInputValidation`), added to the `test-binding-minimal` run regex. `TestMultiModalInputValidation` additionally gains the same model gate as its siblings plus an init-then-probe (the `TestMultiModalEmbeddingInit` pattern): its rejections are pure-Go argument checks, but without a working model a regressed check would fall through to the FFI and error anyway because the model is not loaded, still satisfying the err != nil assertions. With the verified model, a neutered validation check falls through to a working encode and fails the test (verified by mutation: disabling the empty-text check turns the EmptyText subtest red), and a broken model path fails the test at the probe instead of letting the negative subtests pass vacuously (verified: a nonexistent path fails with "multi-modal model not functional").
- Fixes the latent fixture defect that would have turned the lane red on arrival: `multimodalHNSWConfig` never set `TargetDimension`, `WithDefaults()` filled in 768, and the 384-dim model rejects that at encode time. These gated tests had never passed anywhere, including at authoring time.
- Repairs the Wikimedia fixtures so network-dependent tests fail only for real reasons: a descriptive User-Agent in the Go and Rust test download helpers and the example (Wikimedia returns 403 for default library User-Agent strings), and a direct URL for the cat image (its 512px thumbnail rendition now returns 400 from the thumbor service).
- Applies the same User-Agent fix to the exported `MultiModalEncodeImageFromURL` paths in BOTH bindings (candle and onnx), keeping the two exported URL-encode implementations aligned on User-Agent behavior (their pre-existing differences on redirects and size caps are untouched). The router's own request path is unaffected (it encodes images from base64, not URLs), but the exported URL variants had the same 403 problem for any consumer pointing them at a User-Agent-policy host, and the candle one is what the URL-encode tests exercise. Intentionally distinct strings identify test traffic (`semantic-router-tests/1.0`) versus library traffic (`vllm-semantic-router/candle-binding` and `vllm-semantic-router/onnx-binding`).
- Adds `make test-binding-multimodal`, the manual pass/fail receipt for PRs touching the multimodal FFI: all multimodal binding Go tests (including the network-dependent ones) plus the router integration tests, against a local model copy. The ignored Rust unit tests live in a separate `make test-binding-multimodal-rust-baseline` target, kept out of the receipt on purpose: that suite has a known-red baseline (below), and a receipt whose failure is expected cannot tell "you broke something" apart from "the baseline is red".
- Documents the lane split and the known-failing Rust baseline in `docs/agent/testing-strategy.md`.

What stays out of the per-PR CI lane, and why:

- The image-encode binding tests (`TestMultiModalEncodeImageFrom*`, `TestMultiModalCrossModalRetrieval`) download fixture images from Wikimedia at test time; keeping external-network fetches out of the merge gate avoids flakiness. The same image-encode FFI path runs hermetically in CI via the pkg/classification integration tests.
- The `#[ignore]` Rust unit tests: no automated lane runs model-dependent `cargo test --lib`, several fetch Wikimedia fixtures, and actually running them surfaced 7 pre-existing failures out of 32 (below).

Pre-existing findings surfaced while doing this work (out of scope here; happy to file follow-up issues):

- `go vet ./...` on `onnx-binding` fails on a pre-existing test-file error present on the base commit (`semantic-router_test.go:39` uses the two-value `IsModelInitialized` in single-value context), meaning the onnx test file does not currently compile; the package itself builds clean. Related gap: `onnx-binding/**` appears in none of the `test-and-build` path filters and `make vet` covers only candle-binding and the router, so no CI lane ever compiles the onnx test file - which is why this went undetected.
- The 5 audio-path tests fail with "Audio encoder not loaded" even though the model's `model.safetensors` header lists 67 `audio_encoder.encoder.*` tensors: `WhisperEncoder::load` fails and the loader's `.ok()` fallback silently disables audio.
- `test_text_batch_encoding` fails with a broadcast shape mismatch (`[4, 128, 384]` vs `[1, 128, 384]`) in the batched text-encode path.
- `test_text_semantic_similarity` hard-asserts a semantic ranking the model does not reliably produce (cat-dog 0.2172 vs cat-car 0.2224); the Go equivalent already downgrades this check to a log line.

Cost: no new job, no new model download (the job fetches the model today). Measured locally, the newly executing tests add about 3 seconds: the two integration tests run 1.37s + 1.29s and the three binding tests about 0.5s combined.

Risk note, stated as a tradeoff: the model is public, so the download needs no repository secrets. But the downloader soft-skips ANY failure when `HF_TOKEN` is absent (the #2107 (gated-model soft-skip) class), including transient network failures on fork PRs. On such a run the gated tests would fail against a missing model directory rather than skip, and the failure surfaces far from its root cause. That hard failure is intentional (silent skip forever is exactly what this issue is about), but it does mean a flaky-network day can red a fork PR until re-run; the alternative (a directory-existence skip naming the soft-skip) trades that noise for a lane that can silently stop running again. This PR picks fail-loud; happy to flip if the team prefers.

## Test Plan

- [ ] `test-and-build` on this PR shows `TestEmbeddingClassifier_Integration*` as RUN/PASS (not SKIP) in the `test-semantic-router` step
- [ ] `test-and-build` on this PR shows `TestMultiModalEmbeddingInit`, `TestMultiModalEncodeText`, `TestMultiModalInputValidation` as RUN/PASS in the `test-binding-minimal` step
- [ ] Path filters verified: this PR touches `src/**`, `candle-binding/**`, `tools/make/**`, `.github/workflows/**`, so the `changes` filter (`core`, `make`, `ci`) fires `test-and-build`
- [ ] Local: gated Go tests skip without `MULTIMODAL_MODEL_PATH`, pass with it, and both the integration tests and the input-validation test fail against a broken path
- [ ] Local: `make test-binding-multimodal` (the pass/fail receipt: both Go suites) exits zero against a local model copy; `make test-binding-multimodal-rust-baseline` compared against the recorded known-red baseline

(The first two boxes tick from this PR's own CI run; CI execution is not claimed as demonstrated until then.)

## Test Result

Local receipts (CPU build, `cargo build --release --no-default-features`, matching the CI `rust-ci` path).

Without `MULTIMODAL_MODEL_PATH` (current CI behavior): both `TestEmbeddingClassifier_Integration*` tests SKIP, and all three binding tests SKIP.

With `MULTIMODAL_MODEL_PATH` pointing at a downloaded `multi-modal-embed-small` copy:

```
--- PASS: TestEmbeddingClassifier_IntegrationImageQueryEndToEnd (1.37s)
--- PASS: TestEmbeddingClassifier_IntegrationTextRulesIgnoredOnImagePath (1.29s)
--- PASS: TestMultiModalEmbeddingInit (0.07s)
--- PASS: TestMultiModalEncodeText (0.38s)
--- PASS: TestMultiModalInputValidation (0.00s)
```

Falsifiability: with `MULTIMODAL_MODEL_PATH` pointing at a nonexistent directory, both integration tests FAIL rather than skip, and `TestMultiModalInputValidation` fails at its model probe ("multi-modal model not functional") rather than passing vacuously.

`make test-binding-multimodal` full run: all 7 multimodal binding Go tests pass (including the Wikimedia-dependent ones after the fixture repairs) and both router integration tests pass - the receipt target exits zero. `make test-binding-multimodal-rust-baseline` reports 25 passed / 7 failed, matching the recorded known-red baseline (the pre-existing failures listed above).

Gates: `make agent-lint CHANGED_FILES=...` passes (the changed Rust files are clippy-clean; crate-wide clippy debt outside the changed set is scoped out by the gate). `yamllint` with the repo config passes on the workflow file, `markdownlint` on the docs change, `go vet` passes on `candle-binding` and `src/semantic-router` (the packages `make vet` covers); `onnx-binding`'s pre-existing test-file vet failure is noted above. The stubbed multimodal unit tests in pkg/classification still pass after the fixture change.


Xunzhuo (approved):
No blocking findings on this head. The workflow wiring turns the model-gated multimodal tests into real CI execution, the hermetic/network split is sensible, and the fixture/model-dimension fixes make failures meaningful. Local checks: `git diff --check`, `make agent-validate`, and changed-file `make agent-lint` passed (the lint wrapper reported existing crate-wide clippy failures as out-of-scope baseline). The branch is currently behind main and overlaps the now-merged #2431 in Rust/make surfaces, so please update/rebase and let the affected CI rerun before merge; re-review is warranted if conflict resolution changes this logic.

PR #2435, opened 2026-07-10
[CI/Build] pre-download benchmark models in performance-test workflow

### LINKED ISSUE #2434: [Bench] performance-test CI has no model files; classification and cache benchmarks cannot run
The performance-test workflow never downloads models. Its comment assumes the router downloads them at startup, but the component benchmarks initialize classifiers/embeddings directly and never start the router, so:

- classification benchmarks fail: `AutoInitializeUnifiedClassifier` needs the intent+pii+jailbreak merged classifiers under `models/`
- cache benchmarks fail: `models/mom-embedding-pro` (Qwen/Qwen3-Embedding-0.6B) is missing

All 15 benchmarks in those two suites fail, and the "Cache Models" step restores a cache that no step populates. Currently masked by the hardcoded-success issue; visible in https://github.com/vllm-project/semantic-router/actions/runs/29076060503.

Fourth layer of https://github.com/vllm-project/semantic-router/issues/2398.

drivebyer:
Fixes https://github.com/vllm-project/semantic-router/issues/2434 (fourth layer of https://github.com/vllm-project/semantic-router/issues/2398): the component benchmarks initialize classifiers/embeddings directly and never start the router, so nothing downloads models in CI and all classification/cache benchmarks fail.

## How

- New `download-models-perf` target: the 3 mmbert32k merged classifiers (`AutoInitializeUnifiedClassifier` needs all three) + `Qwen/Qwen3-Embedding-0.6B` at `models/mom-embedding-pro`, with `onnx/` excluded (~5GB instead of ~14GB)
- Workflow runs it before the benchmarks; models land in `models/`, so the existing "Cache Models" step (key: `hashFiles('tools/make/models.mk')`) finally gets populated and later runs hit the cache

Should land before or together with https://github.com/vllm-project/semantic-router/pull/2400, otherwise perf-path PRs turn red once that PR unmasks failures.

## Test plan

- `yamllint` (repo config) and `make agent-lint CHANGED_FILES=".github/workflows/performance-test.yml,tools/make/models.mk"` pass
- This PR touches the workflow itself, so its own performance-test run exercises the download step end to end

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -76,8 +76,12 @@ jobs:
         run: |
           pip install -U "huggingface_hub[cli]" hf_transfer
 
-      # Models are now automatically downloaded by the router at startup
-      # No need to pre-download models - the router will download them on first run
+      # The router downloads models at startup, but the component benchmarks
+      # initialize classifiers and embedding models directly, so the minimal
+      # benchmark model set must be pre-downloaded. Populates the "Cache Models"
+      # cache above; the cache key rotates when tools/make/models.mk changes.
+      - name: Download benchmark models
+        run: make download-models-perf
 
       - name: Run component benchmarks
         run: |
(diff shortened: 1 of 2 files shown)


AayushSaini101 (approved):
LGTM thanks

PR #2442, opened 2026-07-10
[Router] Fix nil-receiver panic in classification API during startup

### LINKED ISSUE #2441: [Router] Classification API panics with nil-pointer dereference when called during startup
## Description

While the router is still initializing (models downloading, classification runtime not yet registered), the API server is already listening and `/health` returns `{"status": "healthy"}` — but any request to the classification endpoints (e.g. `POST /api/v1/classify/intent`) crashes the HTTP handler with a nil-pointer panic instead of returning the designed placeholder response. Observed on `main` @ `eadeb1d0` (2026-07-10), macOS arm64, and the code path is platform-independent.

## Steps to Reproduce

```bash
make build-router
LD_LIBRARY_PATH=$PWD/candle-binding/target/release:$PWD/ml-binding/target/release:$PWD/nlp-binding/target/release \
  ./bin/router -config=config/config.yaml &
sleep 20   # while startup/model download is still in progress
curl -s http://localhost:8080/health
# -> {"status": "healthy", "service": "classification-api"}
curl -s -X POST http://localhost:8080/api/v1/classify/intent \
  -H 'Content-Type: application/json' -d '{"text":"What is 2+2?"}'
# -> connection reset; router log shows the panic below
```

## Error / Evidence

```
http: panic serving [::1]:58597: runtime error: invalid memory address or nil pointer dereference
goroutine 203 [running]:
...
github.com/vllm-project/semantic-router/src/semantic-router/pkg/services.(*ClassificationService).ClassifyIntent(0x0, ...)
    src/semantic-router/pkg/services/classification.go:123
github.com/vllm-project/semantic-router/src/semantic-router/pkg/apiserver.(*liveClassificationService).ClassifyIntent(...)
    src/semantic-router/pkg/apiserver/runtime_service.go:78
github.com/vllm-project/semantic-router/src/semantic-router/pkg/apiserver.(*ClassificationAPIServer).handleIntentClassification(...)
    src/semantic-router/pkg/apiserver/route_classify.go:38
```

Note the receiver: `ClassifyIntent(0x0, …)` — the method runs on a nil `*ClassificationService`.

## Code Path

- `src/semantic-router/pkg/apiserver/server.go:70-78` — the resolver closure passed to `newLiveClassificationService` returns `runtimeRegistry.ClassificationService()` / `services.GetGlobalClassificationService()` directly. Both return a concrete `*services.ClassificationService` (`routerruntime/registry.go:53`, `services/classification.go:94`), which is nil until initialization completes.
- `src/semantic-router/pkg/apiserver/runtime_service.go:65-75` — `current()` checks `svc != nil`, but a nil concrete pointer wrapped in the `classificationService` interface is a **non-nil interface value** (typed-nil interface), so the guard passes and the fallback/placeholder path is never taken.
- `src/semantic-router/pkg/services/classification.go:123` — the method dereferences `s.classifier` on the nil receiver and panics. (Ironically the method already handles `s.classifier == nil` with a placeholder response — it never gets the chance.)

- Expected: classification requests during startup return the placeholder response (`"routing_decision": "placeholder_response"`), as `current()`'s fallback design intends.
- Actual: nil-pointer panic in the HTTP handler; the connection is reset. `/health` claims healthy throughout.

## Proposed Fix

Nil-check the concrete pointers before they are wrapped into the interface: extract the resolver closure into a `buildClassificationResolver` function (mirroring the existing `buildConfigResolver`) that returns an untyped nil when no service is available, so `current()` falls back to the placeholder. Fix + regression tests ready in branch `fix-apiserver-typed-nil-classify` (verified end-to-end: the panicking request now returns the placeholder JSON with zero panics logged).

TwilightTechie:
Fixes #2441

## What

While the router is still initializing (models downloading, classification runtime not yet registered), the API server already listens and `/health` reports healthy — but any classification request panics the HTTP handler:

```
http: panic serving ...: runtime error: invalid memory address or nil pointer dereference
services.(*ClassificationService).ClassifyIntent(0x0, ...)      ← nil receiver
    src/semantic-router/pkg/services/classification.go:123
apiserver.(*liveClassificationService).ClassifyIntent
    src/semantic-router/pkg/apiserver/runtime_service.go:78
```

## Why

Classic typed-nil interface bug. The resolver closure passed to `newLiveClassificationService` returns `runtimeRegistry.ClassificationService()` / `services.GetGlobalClassificationService()` directly. Both return a concrete `*services.ClassificationService`, which is nil until initialization completes. Wrapping that nil pointer in the `classificationService` interface produces a **non-nil interface value**, so the `svc != nil` guard in `liveClassificationService.current()` passes and the method runs on a nil receiver — the placeholder fallback that `current()` was designed to provide never gets a chance.

## Before / After

```go
// before — nil *ClassificationService escapes into the interface
func() classificationService {
    if runtimeRegistry != nil {
        return runtimeRegistry.ClassificationService()
    }
    return services.GetGlobalClassificationService()
}

// after — untyped nil lets current() fall back to the placeholder as designed
func buildClassificationResolver(runtimeRegistry *routerruntime.Registry) func() classificationService {
    return func() classificationService {
        if runtimeRegistry != nil {
            if svc := runtimeRegistry.ClassificationService(); svc != nil {
                return svc
            }
            return nil
        }
        if svc := services.GetGlobalClassificationService(); svc != nil {
            return svc
        }
        return nil
    }
}
```

The extraction mirrors the existing `buildConfigResolver` pattern in the same file.

## Behavior change (user-visible)

`POST /api/v1/classify/intent` during startup:
- before: connection reset, panic in router log
- after: `{"classification":{"category":"general","confidence":0.5,...},"routing_decision":"placeholder_response"}` — the placeholder response the code already intended for this window.

## Test Plan

- New regression tests (`pkg/apiserver/runtime_service_test.go`):
  - `TestBuildClassificationResolverReturnsUntypedNil` — the resolver with an empty registry must return an untyped nil, not a typed-nil interface.
  - `TestLiveClassificationServiceFallsBackDuringStartup` — the full request path that panicked now serves the placeholder without panicking.
- End-to-end on macOS arm64: rebuilt `bin/router`, issued the previously panicking request during startup — placeholder JSON returned, `grep -c panic` on the router log = 0.
- `make agent-lint CHANGED_FILES="src/semantic-router/pkg/apiserver/server.go,src/semantic-router/pkg/apiserver/runtime_service_test.go"` passes.


FAUST-BENCHOU (approved):
LGTM.Thanks

PR #2449, opened 2026-07-11
[Router] default router_replay list to summary records

### LINKED ISSUE #2157: router_replay list endpoint returns full record bodies (list should return summaries)
## Problem
`GET /v1/router_replay` (the history list) returns **full record objects** — including the heavy free-text fields (`request_body`, `response_body`, `prompt`, `tool_definitions`, full `tool_trace`). A list endpoint embedding full per-record payloads is an anti-pattern: response size scales with `page_size × per-record-content`, which the list query can't bound.

This is the root cause behind the history-list **413 Payload Too Large** that #2155 fixes pragmatically. A single large request — or a long agent `tool_trace` — inflates one record enough to push a normal page (`limit=25`) past the ext-proc size cap. Capture-side caps (`max_body_bytes`, `max_tool_trace_bytes`, `max_tool_trace_steps`) bound per-field/per-step size but not the page-level total (`records × steps × fields`), so the list can still exceed the cap.

## Current state
- A per-record detail endpoint already exists and is the natural home for full fidelity: `GET /v1/router_replay/{id}`.
- The dashboard frontend already relies on it: the Insights list table renders only summary columns, client-side search is on `request_id`, and the per-record view (`InsightsRecordPage`) fetches `GET /v1/router_replay/{id}` directly rather than reading bodies from list rows. So dropping heavy fields from the list looks largely non-breaking for the dashboard's runtime — the main follow-on would be aligning a couple of e2e mock fixtures that currently embed bodies.

## Options
1. **Summaries-only list** — list returns lightweight rows; the detail view already loads `/{id}`. Cleanest; low-risk for the current frontend (runtime renders bodies only from the detail endpoint).
2. **Projection param** — e.g. `?expand=bodies`; list defaults to summaries, callers opt into heavy fields. More API surface, less breaking.
3. **Bounded full records** (the #2155 stop-gap) — list trims heavy fields for transport and never 413s, full fidelity on `/{id}`. Safe, but the list still carries more than it should.

If other list endpoints embed full bodies too, the same "list = summaries, detail = full" convention may be worth applying consistently.

## Next step
#2155 is the immediate, non-breaking fix; this issue tracks the redesign (options 1–2) for maintainers to decide on. No strong preference from me — happy to own the implementation once a direction is chosen. One scheduling note: I have ~4 working days before a long absence, so if you'd like me to drive it I'd need a direction soon; otherwise this issue captures everything for whoever picks it up.

## H2 execution scope

Change router replay list APIs to return summary records while preserving full request/response details on detail endpoints. This directly supports #2345 dashboard analytics and safer replay browsing.

Relevant code surfaces: `src/semantic-router/pkg/routerreplay`, `src/semantic-router/pkg/extproc/recorder.go`, dashboard replay proxy/redaction, and dashboard analytics backend.

## Acceptance criteria

- List endpoints return bounded summaries without full raw bodies.
- Detail endpoints still return complete records according to existing access/redaction policy.
- Dashboard/replay consumers are updated to use summaries for list views.
- Tests cover list summary shape, detail behavior, and redaction expectations.

## Validation

```bash
make agent-report ENV=cpu CHANGED_FILES="src/semantic-router/pkg/routerreplay src/semantic-router/pkg/extproc dashboard/backend dashboard/frontend website/docs"
make agent-lint CHANGED_FILES="src/semantic-router/pkg/routerreplay src/semantic-router/pkg/extproc dashboard/backend dashboard/frontend website/docs"
make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/routerreplay src/semantic-router/pkg/extproc dashboard/backend dashboard/frontend website/docs"
```

AayushSaini101:
## Summary

Fixes [#2157](https://github.com/vllm-project/semantic-router/issues/2157) by implementing 

**Option 1**: `GET /v1/router_replay` now returns **summary records by default**, preventing list responses from embedding full request/response bodies and tool traces that can push paginated pages past the ext-proc ~4 MiB cap (413 Payload Too Large).

Full fidelity remains on `GET /v1/router_replay/{id}`. `showDetails=true` is kept as an explicit opt-in for legacy callers that need full bodies in list rows.

## Problem

The router replay list endpoint returned full `RoutingRecord` objects by default, including heavy fields (`request_body`, `response_body`, `prompt`, `tool_definitions`, full `tool_trace` steps, `projection_trace`). A single large agent session could inflate one row enough that a normal page (`limit=25`) exceeded the ext-proc message size limit.

The dashboard Insights page already used `showDetails=false` and fetched detail via `GET /v1/router_replay/{id}` — so summaries-by-default is non-breaking for the primary UI consumer.

### Router (`src/semantic-router/pkg/extproc/`)
- Flip default `showDetails` from `true` → `false` in `parseRouterReplayListQuery`
- Default list rows pass through existing `ListSummaryRecord()` helper (strips bodies, prompt, tool definitions, projection trace, tool trace steps; keeps routing metadata, signals, costs, tool names)
- `showDetails=true` opt-in preserved for backward compatibility (can still 413 on oversized pages)

### Tests
- Default list with 5MB body → **200 OK**, empty `request_body`
- New test: `showDetails=true` returns full bodies
- 413 test scoped to `showDetails=true` only
- Dashboard: list-payload redaction test for `router_replay.list` responses

### CLI / docs (`src/vllm-sr/`)
- `recipe_learning` no longer forces `showDetails=true` (summaries include the metadata it needs)
- README updated to document summaries-by-default contract

### CI tooling (`tools/make/`, `tools/linter/markdown/`)
- Pin `markdownlint-cli@0.43.0` in repo-local `tools/linter/markdown/package.json`
- `agent-bootstrap` installs via `npm install --prefix tools/linter/markdown`
- `make markdown-lint` uses local binary with clear install instructions if missing

## API contract after this PR

| Endpoint | Behavior |
|----------|----------|
| `GET /v1/router_replay` | Summary rows (metadata only) — **new default** |
| `GET /v1/router_replay/{id}` | Full record (unchanged) |
| `GET /v1/router_replay?showDetails=true` | Full bodies in list (legacy opt-in; can 413) |

## Test plan

- [x] `go test ./pkg/routerreplay/` — passed locally
- [x] `go test ./router/ -run TestRedact` (dashboard backend) — passed locally
- [x] `make markdown-lint` — passed locally
- [x] `pre-commit run md-fmt --all-files` — passed locally
- [x] `pre-commit run` on changed files (go fmt, black, md-fmt) — passed; go-lint blocked locally by golangci-lint Go 1.25 toolchain mismatch (CI uses correct toolchain)
- [ ] CI `test-semantic-router` / extproc tests (requires bindings in CI environment)

## Commits

1. `[Router] default router_replay list to summary records` — Fixes #2157
2. `[CI/Build] install markdownlint locally for pre-commit` — dev tooling fix encountered during validation


FAUST-BENCHOU (approved):
make sense

PR #2461, opened 2026-07-12
[CI/Build] Add optional maintainer board apply mode

### LINKED ISSUE #2448: feature: optional apply mode for scheduled maintainer board actions
### Describe the feature

Add an **optional, explicitly gated apply path** to the maintainer board CI workflow so maintainers can periodically apply **reviewable** actions from `proposed-actions.json` (e.g. `label_issue`, `label_pr`) without running local scripts by hand.

This is a follow-up to the read-only scheduled workflow delivered in #2246 / PR #2416. The daily cron should remain visibility-only by default; apply mode should be opt-in and manual.

### Primary layer

global level

### Why this layer?

This is cross-cutting maintainer infrastructure (GitHub issue/PR workflow, scheduled CI, maintainer policy) rather than a runtime router subsystem.

#2246 intentionally shipped read-only CI: sync state, publish `today.md` + JSON artifacts, no GitHub mutations on the scheduled path.

After merge, @AayushSaini101 suggested on PR #2416:

> Can we allow to automate update the MR, not frequently but on a suitable interval, i guess this is out of the scope of the issue, feel free to create an issue for this is discussion

The local tooling already supports apply (`maintainer_board.py apply --actions ... --confirm`), but CI does not expose it yet. Maintainers still need a local step to turn `proposed-actions.json` into labels.

A gated CI apply mode would close that gap while preserving the safety model in `docs/agent/maintainer-ops.md`.

### Proposed scope

- Extend `.github/workflows/maintainer-board.yml` with an **explicit `workflow_dispatch` input** such as `apply_actions` (default: `false`).
- When `apply_actions=true`:
  - Run sync first (same as today).
  - Apply only actions from the generated `proposed-actions.json` using existing maintainer board apply logic.
  - Require maintainer-only trigger (no scheduled apply).
- **Never** auto-close issues/PRs from this workflow — keep that in `.github/workflows/stale.yml`.
- Respect `tools/agent/maintainer-policy.yaml` labels/thresholds; no divergent taxonomy.
- Document apply mode, safety gates, and coexistence with `stale.yml` in `docs/agent/maintainer-ops.md`.
- Consider a conservative action allowlist initially: `label_issue`, `label_pr` only (defer comments/closes).

### Acceptance criteria

- Scheduled cron path remains read-only (no mutations).
- Apply mode runs **only** via manual `workflow_dispatch` with an explicit boolean input.
- Apply uses the reviewable `proposed-actions.json` payload produced by sync.
- Failed/partial apply is reported clearly in the Actions job summary.
- Docs explain when to use visibility-only vs apply mode.
- Validation includes harness gates for touched workflow/docs files.

### Additional context

**Related work (already merged):**
- Issue: #2246
- PR: #2416 — adds `.github/workflows/maintainer-board.yml` and `tools/agent/scripts/run_maintainer_board_ci.sh`
- Local apply seam: `tools/agent/scripts/maintainer_board.py apply --actions ... --confirm`
- Policy: `tools/agent/maintainer-policy.yaml`
- Ops docs: `docs/agent/maintainer-ops.md` (Apply Policy + stale.yml coexistence)

**Non-goals for v1 of this follow-up:**
- Scheduled/automatic apply without human dispatch
- Replacing `stale.yml`
- Creating/closing issues or PRs without explicit maintainer review

Happy to work on this follow-up if maintainers agree on the gating model and allowlist.

haritha3320:
Fixes #2448

## Purpose

- Add an optional `apply_actions` input to the maintainer board workflow.
- Keep the daily cron path read-only.
- Let maintainers apply labels from `proposed-actions.json` via manual `workflow_dispatch`.
- **Module:** `CI/Build`

## Test Plan

- Run local read-only sync:
  ```bash
  MAINTAINER_BOARD_OUTPUT_DIR=/tmp/mb-safe-test \
  MAINTAINER_BOARD_MILESTONE=v0.4 \
  MAINTAINER_BOARD_ISSUE_LIMIT=5 \
  MAINTAINER_BOARD_PR_LIMIT=5 \
  bash tools/agent/scripts/run_maintainer_board_ci.sh
  ```
- Confirm apply script refuses local runs:
  ```bash
  MAINTAINER_BOARD_OUTPUT_DIR=/tmp/mb-safe-test \
  bash tools/agent/scripts/run_maintainer_board_apply_ci.sh
  ```
- Lint changed workflow/docs/shell files.
- After merge, maintainer can test read-only dispatch first, then `apply_actions=true` if needed.

## Test Result

- Local sync succeeded and wrote `today.md`, `current.json`, and `proposed-actions.json` under `/tmp/mb-safe-test`.
- Local apply script exited with a guard error outside `workflow_dispatch` (expected).
- Shell and YAML lint passed locally.
- Apply mode was not run against GitHub in this test.

---
<details>

</details>


FAUST-BENCHOU (commented):
Acceptable for me

haritha3320:
@AayushSaini101 can you please review

.github/workflows/maintainer-board.yml
```diff
@@ -66,3 +74,37 @@ jobs:
             ${{ runner.temp }}/maintainer-board/current.json
             ${{ runner.temp }}/maintainer-board/proposed-actions.json
             ${{ runner.temp }}/maintainer-board/milestone-*.md
+
+  apply-actions:
+    needs: maintainer-board
+    if: github.event_name == 'workflow_dispatch' && inputs.apply_actions == true
```
Xunzhuo: [P1] Add a real review boundary before granting write permissions. `apply_actions` is selected before sync, and this same run immediately downloads and applies the newly generated payload, so nobody can inspect the exact `proposed-actions.json` before mutation. This contradicts the documented apply policy. Please consume a previously reviewed artifact identified by run ID/digest, or place the apply job behind a protected-environment approval after the artifact is uploaded.

tools/agent/scripts/run_maintainer_board_apply_ci.sh
```diff
@@ -0,0 +1,46 @@
...
+PY
+)"
+
+if [[ "${action_count}" == "0" ]]; then
+  echo "No proposed actions to apply."
+  exit 0
+fi
+
+echo "Applying ${action_count} proposed maintainer board action(s) from ${ACTIONS_FILE}"
+python3 maintainer_board.py apply --actions "${ACTIONS_FILE}" --confirm
```
Xunzhuo: [P1] Enforce the advertised label-only allowlist before invoking the generic apply command. This wrapper only counts arbitrary JSON, while `maintainer_board.py apply` also accepts `create_issue`; a synthetic `create_issue` payload reaches `gh issue create` through this CI wrapper. Validate the root/list schema, allow only `label_issue`/`label_pr`, require repo-local numeric targets, and restrict labels to maintainer policy, with negative tests.

tools/agent/scripts/run_maintainer_board_apply_ci.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
```
Xunzhuo, line 50: [P1] Always report partial application before failing the job. Actions mutate sequentially and `set -e` exits on the first failed `gh` command before this summary block, so earlier successful mutations become invisible. A two-action reproduction applied the first, failed the second, returned non-zero, and produced no summary. Track attempted/succeeded/failed actions through a trap/finally path, write the summary, then preserve the failing status.

Xunzhuo (changes requested):
The scheduled path remains read-only, but the new manual apply path does not yet preserve the repository's explicit reviewed-payload boundary or its documented label-only contract. I reproduced both key failures with a fake `gh`: a `create_issue` payload reaches `gh issue create`, and a two-action partial failure mutates the first action then exits with no job summary. Please add behavior tests for the allowlist, malformed/empty payloads, review boundary, and partial failure. `make agent-validate`, `bash -n`, the schedule guard, and `git diff --check` passed.

haritha3320:
@Xunzhuo Thanks for the review — I've addressed all three P1 items in the latest commits:

1. **Review boundary** — sync and apply are now separate `workflow_dispatch` runs via `source_run_id`; apply downloads a previously reviewed artifact instead of mutating in the same run.
2. **Label-only allowlist** — added `maintainer_board_ci_apply.py` to validate payloads (only `label_issue`/`label_pr`, `#<number>` targets, maintainer-policy labels; rejects `create_issue` and malformed input).
3. **Partial failure reporting** — apply tracks attempted/succeeded/failed actions and writes the full summary even when some `gh` commands fail.

Also added behavior tests in `test_maintainer_board_ci_apply.py` and fixed a shellcheck issue in the apply wrapper.

Could you please take another look when you have a moment?

Xunzhuo (commented):
Thanks for addressing the allowlist, reviewed-artifact, malformed/empty payload, and partial-failure summary points; I don't see a new code blocker in this head. It now conflicts with main, exact-head lint is failing, and #2448 is closed. Please rebase and link a current accepted scope before the final readiness pass.

FAUST-BENCHOU (commented):
Thanks for adding the optional maintainer board apply mode. Keeping the scheduled workflow read-only while requiring an explicit reviewed artifact for mutations is a good approach.

PR #2488, opened 2026-07-13
[Router] cache: per-lookup similarity and context propagation

### LINKED ISSUE #2473: cache: propagate cancellation and return similarity per lookup
## Summary

Make semantic-cache operations context-aware and return similarity as part of each lookup result. Owner surfaces: `src/semantic-router/pkg/cache/`, `src/semantic-router/pkg/extproc/req_filter_cache.go`, and `src/semantic-router/pkg/extproc/recorder.go`.

## Current behavior

Cache interfaces discard request and runtime contexts, several backends use background contexts, and similarity is published through cache-global mutable state. A concurrent lookup or a miss can therefore expose another request's score in headers, telemetry, or Replay.

## Expected behavior

Cancellation reaches storage and embedding work, each lookup returns its own body/found/similarity value, misses and errors have explicit absent similarity, and partial construction cleans up clients.

## Acceptance

- Pass context through health, read, write, and search interfaces and all backends.
- Return a value object containing body, found, and similarity; remove global last-similarity state.
- Return zero or absent similarity on misses and errors.
- Preserve request-owned similarity through debug and Replay paths.
- Close every partially constructed client on failure.

## Validation

- Add barrier-controlled concurrent lookups proving result isolation.
- Add miss, error, cancellation, native/backend interruption, and constructor-failure tests.
- Run `cd src/semantic-router && go test -race ./pkg/cache/... ./pkg/extproc`.
- Include a cancellation and per-request-score assertion in the affected cache E2E profile.

## Related

- Parent audit: #2375
- Memory growth tracker: #2222

WUKUNTAI-0211:
Closes #2473

## Summary

Make semantic-cache operations context-aware and return similarity as part of each lookup result.

This removes backend-global similarity state that could leak scores between concurrent requests.

## Changes

- Add `LookupResult{Body, Found, Similarity}` across all cache backends.
- Remove `SimilarityTracker`, `StoreSimilarity`, and `LastSimilarity`.
- Pass `context.Context` through health, read, write, and search operations.
- Define a strict contract: only hits carry similarity; every miss returns zero.
- Propagate cancellation and deadline errors from remote backend searches.
- Keep ordinary backend outages fail-open as cache misses.
- Prevent canceled in-memory writes from publishing orphaned entries.
- Close partially constructed Redis and Valkey clients on constructor failure.
- Distinguish Hybrid cross-model misses from actual Milvus fetch failures.
- Preserve the shared Redis/Valkey IP distance conversion introduced by #2716.
- Reject zero-hit E2E runs and invalid similarity values, including `NaN` and infinities.
- Replace the cache E2E file-wide lint exclusion with function-local directives.

## Test Plan

- Barrier-controlled concurrent lookups verify request-owned similarity.
- Cancellation tests cover embedding and storage-operation boundaries.
- Mutation checks prove all four remote backend cancellation branches are exercised.
- Hybrid regression tests cover cross-model misses and actual Milvus fetch failures.
- E2E helper tests cover zero-hit runs, hit/miss similarity, and non-finite values.

Local verification:

```bash
go test -count=1 -race ./pkg/cache/... ./pkg/extproc
cd e2e && go test -count=1 ./testcases
make agent-ci-gate ENV=cpu CHANGED_FILES="<changed files>"
```

Redis and Valkey example programs also compile successfully.

## Notes

Embedding uses synchronous CGO and cannot be interrupted mid-call. Cancellation is checked before embedding and again before publishing in-memory state.

Remote storage operations receive the request context directly. A canceled operation returns its context error, while ordinary storage failures retain the existing fail-open behavior.

Cluster-backed E2E was not run locally and will be covered by the repository CI profile matrix.


e2e/testcases/cache.go
```diff
@@ -151,9 +153,18 @@ func testSingleCacheRequest(ctx context.Context, testCase CacheTestCase, questio
 	cacheHitHeader := resp.Header.Get("x-vsr-cache-hit")
 	result.CacheHit = (cacheHitHeader == "true")
 
+	// Per-request similarity assertion (#2473): validate the request-owned
+	// x-vsr-cache-similarity header (see parseCacheSimilarity for the contract).
+	sim, simErr := parseCacheSimilarity(resp.Header.Get("x-vsr-cache-similarity"), result.CacheHit)
```
Xunzhuo, line 256: [P1] Make this assertion affect the testcase result. A missing or malformed similarity header is only copied into `CacheResult.Error`; `testCache` later prints all results and unconditionally returns `nil`, so this new assertion still passes CI even when every request fails or no request executes. Please aggregate result errors, fail on zero executed requests, and encode an explicit per-request isolation/threshold acceptance condition rather than only a broad `[0,1]` range.

src/semantic-router/pkg/cache/inmemory_cache_stub.go
```diff
@@ -34,35 +35,28 @@ func (c *InMemoryCache) IsEnabled() bool {
 }
 
 // AddPendingRequest stores a request awaiting its response
-func (c *InMemoryCache) AddPendingRequest(requestID string, model string, query string, requestBody []byte) error {
+func (c *InMemoryCache) AddPendingRequest(_ context.Context, requestID string, model string, query string, requestBody []byte) error {
```
Xunzhuo: [P1] Keep the non-cgo method set compatible with `CacheBackend`. The interface requires `ttlSeconds` on all three write methods, but this stub and `hybrid_cache_stub.go` add context while still omitting TTL, so `windows || !cgo` builds cannot return these types as `CacheBackend`. Please update both stubs and add compile-time interface assertions so future contract migrations fail at the definition site.

src/semantic-router/pkg/cache/redis_cache.go
```diff
@@ -98,7 +97,7 @@ func NewRedisCache(options RedisCacheOptions) (*RedisCache, error) {
 	}
 
 	// Test connection using the new CheckConnection method
-	if err := cache.CheckConnection(); err != nil {
+	if err := cache.CheckConnection(context.Background()); err != nil {
```
Xunzhuo: [P1] Close the partially constructed client on this failure path. If `CheckConnection` fails, `redisClient` is returned without `Close`; the equivalent Valkey path has the same leak. This is an explicit #2473 acceptance item, and the PR currently claims every constructor error path cleans up. Please fix both backends and add constructor-failure cleanup tests.

src/semantic-router/pkg/cache/inmemory_cache.go
```diff
@@ -332,7 +340,7 @@ func (c *InMemoryCache) AddPendingRequest(
 }
 
 // UpdateWithResponse completes a pending request by adding the response
-func (c *InMemoryCache) UpdateWithResponse(requestID string, responseBody []byte, ttlSeconds int) error {
+func (c *InMemoryCache) UpdateWithResponse(_ context.Context, requestID string, responseBody []byte, ttlSeconds int) error {
```
Xunzhuo: [P1] Do not discard the propagated context on the in-memory write path. A canceled request still completes and publishes the pending entry here; `AddPendingRequest`/`AddEntry` also need to re-check after synchronous embedding so cancellation during CGO does not create orphaned cache state. Please check `ctxErr` immediately before mutation and cover cancellation-before-write in tests.

src/semantic-router/pkg/cache/hybrid_cache_lookup.go
```diff
@@ -19,55 +19,58 @@ type candidateWithID struct {
...
-		return responseBody, true, nil
+		return LookupResult{Body: responseBody, Found: true, Similarity: candidate.similarity}, nil
 	}
 
 	h.recordLookupMiss(start, metricOp, logPrefix, threshold, model, totalCandidates, len(candidatesWithIDs))
-	return nil, false, nil
+	// Best HNSW candidate was above threshold but Milvus fetch failed;
+	// surface its score as the per-request best so callers can distinguish this
+	// "found in index, unfetchable" miss from a no-candidate miss.
+	return LookupResult{Similarity: candidatesWithIDs[0].similarity}, nil
```
Xunzhuo: [P1] Do not publish a hit-level score after the storage fetch fails. Every `GetByID` failure reaches this return with `Found=false` and a similarity already above the hit threshold, contradicting the shared `LookupResult` contract (`upstream error` → zero) and exposing/persisting an error-path score as an ordinary miss. Preserve the fetch-error distinction and return zero/an error here.

Xunzhuo (changes requested):
The move from backend-global similarity state to a request-owned `LookupResult` is the right direction, but the migration is incomplete across build variants, failure paths, and the E2E acceptance signal. Please address the inline blockers and add tests for constructor cleanup/cancellation. Also align the documented below-threshold miss contract across Qdrant/Hybrid versus memory/Redis/Valkey/Milvus, and update the stale Valkey addon example to the new interface. `make agent-validate`, `make build-e2e`, and `git diff --check` passed; native cache tests could not link locally without the Candle library, while project CI is green.

WUKUNTAI-0211:
Thanks @Xunzhuo — all five inline blockers and both cross-cutting items from the summary are addressed, rebased onto latest `main`. One commit per point:

**Inline blockers**

- **[P1] E2E assertion now affects the result** (`e2e/testcases/cache.go`) — `30f391f5`. `testCache` no longer returns `nil` unconditionally: priming-request failures are collected, per-request `CacheResult.Error` values are aggregated into the verdict, a run that executed **zero** similar-question requests now fails, and there is an explicit per-request isolation check (a hit with non-positive similarity fails). `parseCacheSimilarity` enforces the contract — a hit with an absent header, or a score outside `(0,1]`, is rejected — so it is no longer a broad `[0,1]` pass.

- **[P1] non-cgo method set kept compatible with `CacheBackend`** (`inmemory_cache_stub.go`, `hybrid_cache_stub.go`) — `f8ed67ec`. `ttlSeconds` added to all three write methods on both stubs, and compile-time `var _ CacheBackend = (*T)(nil)` assertions for all five backends now live at the interface definition site (no build tag), so a future contract change fails to compile there — including the `windows || !cgo` stubs.

- **[P1] constructor closes the partially built client on connect failure** (`redis_cache.go`, `valkey_cache.go`) — `2d957d2e`. Both backends now `Close()` the client when `CheckConnection` fails; constructor-failure tests added in `constructor_cleanup_test.go`. One honest caveat: a *refused* dial leaks no goroutines (go-redis/glide build pools lazily), so those unit tests assert the constructor returns `(nil, err)` rather than proving `Close()` reclaims a live connection — the stranded-connection case (CheckConnection failing *after* a successful dial, e.g. auth) needs a live-server seam I left out of the unit test. Happy to add that as a follow-up if you'd like it in-tree.

- **[P1] in-memory write path honors cancellation** (`inmemory_cache.go`) — `cefb96df`. `AddPendingRequest` and `AddEntry` re-check `ctxErr(ctx)` *after* the synchronous (CGO) embedding and *before* mutating cache state; `UpdateWithResponse` now takes and checks the context too. Covered in `cache_cancellation_test.go` with a context that trips only on the post-embed guard, asserting the context error is returned and no entry is published.

- **[P1] hybrid fetch-failure returns zero + error** (`hybrid_cache_lookup.go`) — `8b9d79f1`. The "above threshold but storage fetch failed" path now returns `LookupResult{}, error` instead of a hit-level score with `Found=false`.

**Summary items**

- **Below-threshold miss contract aligned across backends** — `110a4779`. The same leak fixed in hybrid also existed in milvus, valkey, redis and qdrant: a match qualifying above threshold (server-side filtered, for qdrant) with a missing/empty/pending body returned that hit-level score with `Found=false`. All four now return zero similarity on the data-error path. Genuine below-threshold misses still report their own best-observed score, unchanged.

- **Stale Valkey addon example migrated to the new interface** — `a0a2801c`. `deploy/addons/valkey/valkey-cache.go` now threads `context.Background()` through the write calls and reads the hit body/flag from the returned `LookupResult`; `make run-valkey-cache-example` compiles again.

**Local verification:** `gofmt` clean · `go vet ./pkg/cache/` clean · valkey example compiles · cache suite green (~22.9s) · `git diff --check` clean.

WUKUNTAI-0211:
Force-pushed: `3a8c84df` → `5f0845f2`. History was rewritten, so a re-review should start from the new tip rather than the old range.

### Correction to my 7/16 reply

I wrote then that "genuine below-threshold misses still report their own best-observed score, unchanged". **That is no longer true, and the PR body has been updated to match.** All six backends now return zero similarity on *every* `Found=false` path — below-threshold misses included — so `x-vsr-cache-similarity` and the Replay record only ever carry a hit's own score. The best-observed value is still emitted in each backend's `cache_miss` log event, where it stays useful for threshold tuning. Please review against that contract, not my earlier description.

### `3a8c84df` split into four commits

That commit was a 17-file grab bag with a one-line message. It is now four, each with its own rationale, and the tree is byte-identical to what it replaced:

| | |
|---|---|
| `f85729a2` | propagate request cancellation out of remote backend searches |
| `39c35ec5` | tell a Hybrid cross-model miss from a Milvus fetch failure |
| `fb3b757a` | return zero similarity on every miss |
| `d0836bef` | harden the E2E similarity assertion, drop its lint exclusion |

### Self-review follow-ups on top

Re-reviewing my own branch turned up two things I had shipped in a state that could not fail, plus some scope and hygiene issues:

- **`4dd2efae` — the constructor-cleanup tests were vacuous.** They asserted `(nil, err)`, which both constructors already returned before the cleanup existed; deleting both `Close()` calls kept the suite green. Worth recording that the live-server follow-up I offered earlier would not have worked either: against a server that accepts the dial but never answers, go-redis drops the connection on command timeout, so the peer sees EOF with or without our `Close`. The cleanup now runs through a `releaseOnFailure` seam with an injectable closer, and removing either release call fails `TestNewRedisCacheClosesClientOnConnectFailure`. A refused dial fails inside `glide.NewClient` before any Valkey client exists, so that path now asserts the absence of a client rather than pretending to cover the release branch.
- **`a007f182` — the E2E cancellation assertion #2473 asks for was missing.** A probe now abandons a request mid-flight and re-asks the same question; a cancelled request must publish no cache state, including the pending entry written before the upstream answers. A request that beats the cancel window is reported and skipped, not failed, so this adds no new flake class.
- **`5772bd94` — reverted an over-tightening of mine.** My verdict failed the testcase on zero cache *hits*, which would have promoted the existing `semantic-cache` acceptance-contract hit-rate floor into a matrix-wide gate for every profile running this testcase. Zero executed requests and per-request assertion failures still fail; hit rate goes back to the acceptance contract.
- **`b4596bfd`** narrows the `benchmark.go` golangci entry to function-local directives. The `structure-rules.yaml` exception has to stay — `structure_check.py` enforces its own per-function limit and has no inline directive.
- **`d077ca64`** pins the cancellation specs to the guard under test by asserting the bare context error, so a future `ctxErr` call added before the embed fails them instead of silently moving them to another branch.
- **`20c2f74b`, `5f0845f2`** correct three comments that overstated what the code does, and cut the comment weight this branch added (18% of added lines vs. a 9%/5% baseline for these packages) back to 13%.

### Known gaps, stated rather than papered over

- The `windows || !cgo` stubs are review-verified, not build-verified: nothing in the repo compiles that variant (`CGO_ENABLED=0` fails inside valkey-glide, which needs cgo, and there is no windows job). The compile-time assertions are still worth having; their doc comment now says this.
- A request cancelled between `AddPendingRequest` and `UpdateWithResponse` leaves its pending entry behind. It is unmatchable (`entryEligible` rejects entries with no response) and TTL-bounded; noted at the call site.
- `AddEntriesBatch` still uses `context.Background()`. It is not on `CacheBackend` and has no request context available.
- `recorder.go` is untouched. It reads `ctx.VSRCacheSimilarity`, which is now request-owned; the change belonged on the assignment side.
- `redis_cache.go` is at 797 lines against the 800-line structure limit.

### Verification

```
go build ./pkg/...                                  ok
go vet ./pkg/cache/... ./pkg/extproc                ok
go test -race ./pkg/cache/...                       55/55 specs pass
go test -race ./pkg/extproc                         ok
cd e2e && go test ./testcases                       ok
golangci-lint (agent config, changed files)         0 issues
structure_check.py (changed files)                  0 errors
every commit in the branch builds and vets          10/10
```

Mutation-checked rather than assumed: removing the redis release call, the in-memory post-embed guards, or the hybrid fetch-error return each turns the corresponding tests red.

Eight tests in `pkg/cache` still fail locally — the Hybrid/Milvus set needs a live Milvus, the Valkey integration set needs the valkey-search module. Both reproduce identically on the merge base, so they are environmental, not regressions. Cluster-backed E2E was not run locally and is left to the CI profile matrix.

WUKUNTAI-0211:
Superseded by #2813 (merged) and #2919.

This PR carried two separable changes. The request-owned `LookupResult` half landed in #2813. The context-propagation half — plus all five inline blockers and both cross-cutting items from the 7/15 review — is now #2919, opened against current `main` instead of being rebased through this branch's 26 commits.

One deliberate difference from this PR's final state: below-threshold misses report their rejected candidate's score again, rather than the zero-on-every-miss contract I force-pushed here on 8/7 and corrected in the comment above. `main` already ships that behaviour after #2813, and the reason for zeroing — a score that could be overwritten by a concurrent lookup — no longer exists now that the score rides `LookupResult` instead of shared backend state. The contract is aligned across all six backends either way, which is what the review asked for.

The `structure-rules.yaml` exception for `benchmark.go` is not carried over: #2919 splits `runBenchmarkScenario` into two helpers so the structure gate passes without an exception.

Closing this one; review continues on #2919.

Issue #2375, opened 2026-07-05
[Quality] Audit and harden router code quality, security, and API boundaries


Xunzhuo:
## Summary

Create a bounded roadmap track for a deep router code-quality, security, resource-safety, and API-boundary audit, followed by prioritized hardening work.

Parent roadmap: #2287
Related: #2357, #2326, #2330, #2331, #2355, #2200, #1510
Related code: `src/semantic-router/pkg/config`, `src/semantic-router/pkg/extproc`, `src/semantic-router/pkg/routerruntime`, `src/semantic-router/pkg/services`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/modelselection`, `src/semantic-router/pkg/modelruntime`, `src/semantic-router/pkg/looper`, `src/semantic-router/pkg/memory`, `src/semantic-router/pkg/vectorstore`, `src/semantic-router/pkg/openai`, `src/semantic-router/pkg/anthropic`, `src/semantic-router/pkg/headers`, `src/semantic-router/pkg/responseapi`, `src/vllm-sr/cli`, `deploy/operator/api/v1alpha1`, `deploy/operator/controllers`, `dashboard/backend/router`

## Motivation

The roadmap adds larger config contracts, multiple recipes, custom signals, inference-aware backends, Router Memory/Learning, Looper-family algorithms, dashboard/control-plane surfaces, and edge deployments. These features increase the number of runtime seams and security-sensitive paths.

Before the router grows further, we need a focused quality pass that identifies and prioritizes risks in:

- security-sensitive request, response, feedback, identity, header, plugin, Looper, and memory paths;
- potential resource leaks around goroutines, contexts, timers, streams, HTTP clients, native bindings, model runtimes, vector stores, replay stores, and cache stores;
- weak or duplicated structs, untyped payloads, and ad hoc maps where versioned contracts or SDK/public types should own the API;
- inconsistent API ownership across router core, CLI, dashboard backend, operator, config schemas, and response headers;
- design-pattern drift in hotspots such as config loading/validation, extproc request and response processing, classification runtime, model selection, and service lifecycle wiring.

This is broader than #2357. #2357 tracks known security hardening items across auth, feedback, and Looper paths. This issue should produce a systematic audit and remediation plan across router code quality, API contracts, resource safety, and module boundaries.

## Scope

Audit and prioritize:

- Security analysis
  - request/response header trust boundaries;
  - identity/authz/tenant-sensitive assumptions;
[...]

PR #2496, opened 2026-07-13
[Bench] Make perftest --compare-baseline functional and gate PRs on allocation regressions

### LINKED ISSUE #2455: [Bench] perftest --compare-baseline is non-functional repo-wide: reads a baseline.json nothing produces
## Description

The numeric performance-regression comparison — `perftest --compare-baseline`, exposed via `make perf-compare` and `make perf-check` — cannot run at all, for any suite. It reads a single `baseline.json` that no tooling produces, needs a `current.json` that is never generated, maps every benchmark to one global threshold, and would fail to parse the decimal `ns/op` that `update-baseline.sh` emits for sub-microsecond benchmarks. None of this is exercised in CI (`performance-test.yml` never invokes it), so the breakage has stayed invisible.

This is the numeric-comparison analogue of #2398 — that issue covered the *run + report* half of the perf gate (fixed in #2399/#2400); this covers the *compare-against-baseline* half, which has never worked.

## Steps to Reproduce

```bash
make rust-ci
make perf-compare
# Error comparing with baseline: failed to load baseline:
#   baseline file not found: testdata/baselines/baseline.json
# exit status 1
```

`make perf-check` fails the same way (it runs `perf-compare` under the hood).

## Root Causes (independent)

1. **Baseline filename mismatch.** `perftest` reads `<dir>/baseline.json` (`perf/cmd/perftest/main.go:74`), but `perf/scripts/update-baseline.sh` only ever writes *per-suite* files — `testdata/baselines/{classification,decision,cache,extproc}.json`. There is no `baseline.json`, so `LoadBaseline` errors out immediately.

2. **`current.json` is never produced.** With no `--current` passed (and `make perf-compare` passes none), perftest falls back to `<dir>/current.json` (`main.go:93`) — which nothing writes either.

3. **Thresholds ignore the benchmark name.** `getThresholdForBenchmark` (`perf/pkg/benchmark/baseline.go:145`) never inspects `benchName`; it returns the first non-zero `max_regression_percent` it finds (Classification), so every benchmark effectively shares one threshold and the per-scenario entries in `config/thresholds.yaml` are dead config.

4. **Decimal `ns/op` won't unmarshal.** `BenchmarkMetric.NsPerOp` is `int64` (`baseline.go:21`), but `update-baseline.sh` captures `ns/op` as `[0-9.]+` and writes it raw. Any sub-microsecond benchmark (several cache and `pkg/looper` micro-benchmarks) produces a decimal like `173.1`, which fails `json.Unmarshal` into `int64`. Latent behind (1) today, but it would break the moment (1) is fixed.

5. **Never run in CI.** `.github/workflows/performance-test.yml` only runs `make perf-bench-quick` + the honest per-suite reporting; it never calls `perf-compare`/`perf-check`. So the "catch regressions" gate has never actually gated anything.

## Expected vs Actual

- **Expected:** `make perf-compare` diffs current benchmark results against a recorded baseline and flags per-benchmark regressions beyond the configured threshold; CI fails on regression.
- **Actual:** it exits 1 at baseline load. Even past that, thresholds are global and decimal metrics won't parse, and CI never invokes it.

## Proposed Fix (sketch)

- Align producer and consumer on the same file(s): either write a single combined `baseline.json` + a `current.json` (produced by a "run current benchmarks" step), or teach `perftest` to load every `*.json` in the baseline dir and diff a produced current set.
- Round `ns/op` to an integer in `update-baseline.sh` **or** make `NsPerOp` a `float64`, so metrics round-trip.
- Make `getThresholdForBenchmark` map benchmark name → scenario key (or fall back to a documented default) so `thresholds.yaml` is honored.
- Once functional, invoke it from `performance-test.yml` — warn-only first, then gating — so regressions are actually caught.

## Context

Surfaced while adding the Looper benchmark suite (#2371). That PR deliberately does **not** fix this — it plugs Looper into the run + honest-report gate, which does work post-#2400. Filing separately because this is repo-wide perf tooling, not Looper-specific.

theohsiung:
## Summary

`perftest --compare-baseline` — the numeric performance-regression comparison behind `make perf-compare` / `perf-check` — has never worked repo-wide (#2455). It read a `baseline.json` no tool produces, needed a `current.json` nothing generates, mapped every benchmark to one global threshold, couldn't parse the decimal `ns/op` the committed baselines actually contain, and was never invoked in CI. This PR makes it work end-to-end and turns it into a PR-visible regression gate.

This is the compare-against-baseline half of #2398 (the run + report half was fixed in #2399/#2400).

## What was broken (all fixed)

1. **Baseline filename mismatch** — the consumer read `<dir>/baseline.json`, but `update-baseline.sh` only writes per-suite files (`decision.json`, `looper.json`, …). → `LoadBaselineDir` merges every `*.json` in the directory.
2. **`current.json` never produced** — nothing generated the current result set. → `ParseBenchOutput` + a `perftest --parse-bench` mode build it from the raw bench output.
3. **Thresholds ignored the benchmark name** — `getThresholdForBenchmark` always returned the first non-zero entry (Classification's), and the struct silently dropped the `looper:` block. → per-benchmark regex → threshold, first-match-wins, with a `default`.
4. **Decimal `ns/op` wouldn't unmarshal** — the field was `int64`, so every baseline with a sub-microsecond timing (e.g. `628.5`) failed to parse — including the committed decision/extproc/looper baselines. → `float64`.
5. **Never run in CI** — `performance-test.yml` now generates the current set and runs the comparison with `--fail-on-regression`.

## Design note: what the gate blocks on

The gate **blocks on `allocs/op` and `B/op`** and treats **`ns/op` as advisory** (reported in the PR comment, never fails the gate).

Absolute `ns/op` depends on the runner's CPU and noisy neighbours, so comparing it across machines — even between two `ubuntu-latest` runs — produces false positives, and biasing the baseline slow to dodge them only hides real regressions. `allocs/op` and `B/op` are determined by the code path, not the hardware, so they compare cleanly across machines with **no baseline seeding**, and still catch the most common, GC-heavy Go regressions (extra allocations, heap escapes, per-call buffers, O(n²)-with-allocation). Introducing allocations where the baseline had none (0 → N) always counts.

What this does *not* hard-block: a pure compute-bound slowdown that doesn't change allocations (only `ns/op` moves) — those surface as advisory. Reliably gating wall-clock would need same-runner self-baselining (run main + PR back-to-back), proposed as a follow-up.

Per-benchmark thresholds live in `perf/config/thresholds.yaml`; a metric an entry omits inherits the `default` (never a silent 0% bound).

## Scope

- `performance-nightly.yml` is intentionally left **disabled** — enabling an auto-commit/push baseline job is a maintainer call, and with allocation-gating it is not required for the gate to be trustworthy (baselines refresh via `make perf-baseline-update`).
- The cross-hardware / cross-backend matrix (Candle/ONNX × NVIDIA × AMD) stays with #1510; this is the single-runner mechanism.
- Surfaced by the router-quality audit in #2375.

## Test Plan

- `cd perf && go test -race ./pkg/benchmark/...` — new unit tests cover: decimal `ns/op` round-trip; baseline-dir merge; bench-output parsing (decimals, subtests with `/`, no-`benchmem` lines, `-N` suffix stripping); per-benchmark threshold selection including the shipped `thresholds.yaml`; alloc/bytes gating with `ns/op` advisory; `0 → N` allocations; omitted-threshold inheritance; ungated-benchmark reporting.
- End-to-end against the real committed baselines: merges 37 benchmarks; a **+100% `ns/op`-only** change **passes** (exit 0, advisory) while a **+100% `allocs/op`** change **fails** (exit 1).
- `go build ./...`, `go vet ./...`, `gofmt` clean.
- Workflow YAML validated (parse); the workflow itself runs first on this PR.

## Known limitations / follow-ups

- Pure compute-bound (non-allocating) regressions are advisory, not blocked — reliable wall-clock gating needs self-baselining (follow-up).
- Threshold values (especially the Looper family, which involves concurrency) are initial estimates and may need tuning from real CI runs.
- `classification`/`cache` baselines are currently empty, so those suites are reported as *ungated* until populated (skipped, never falsely failed).
- **`decision.json` was reseeded to current `main`** (`afdf1159`): its committed numbers came from a Dec-2025 dev branch and had drifted substantially over ~7 months of code changes (`ns_per_op` ~20×, `allocs_per_op` ~8×), so the gate was flagging that accumulated drift on every PR. This refresh sets the reference to what `main` produces today — it is **not** a real regression being "refreshed away" (and `ns/op` is advisory / ignored cross-machine regardless). Future baseline refreshes should be reviewed the same way.

Closes #2455.


theohsiung:
@Xunzhuo one heads-up on how the gate behaves, since it slightly reinterprets the acceptance criteria:

Right now the gate **only hard-blocks on `allocs/op` and `B/op`**, and **`ns/op` is advisory** — reported in the PR comment, but it never fails the check.

Why: absolute `ns/op` depends on the CI runner's CPU and noisy neighbours, so comparing it across machines (even between two `ubuntu-latest` runs) produces false positives — and biasing the baseline slow to avoid that just hides real regressions. `allocs/op` and `B/op` are determined by the code path, not the hardware, so they compare cleanly across machines with no baseline seeding, and still catch the most common, GC-heavy Go regressions (extra allocations, heap escapes, per-call buffers, O(n²)-with-allocation).

The trade-off: a pure compute-bound slowdown that doesn't change allocations won't fail the gate (it only shows as advisory). Reliably gating wall-clock would need same-runner self-baselining (run `main` + PR back-to-back in one job) — happy to add that as a follow-up if you'd like time to be a hard gate too. Let me know which you'd prefer.

theohsiung:
@WUKUNTAI-0211 need help on reviewing this PR

TwilightTechie:
#### Verified this end-to-end on macOS arm64 (Apple Silicon, CPU-only), so a completely different machine from the one that recorded the baselines :) :

- cd perf && go test -race ./pkg/benchmark/... — all pass.
- Ran the decision + extproc + looper suites through the new pipeline (--parse-bench → --compare-baseline --fail-on-regression): 40 benchmarks parsed, 37 baseline entries merged from the per-suite files, the 5 extproc benchmarks without baselines were surfaced as ungated, exit 0.
- The hardware-independence design holds up empirically: the committed alloc baselines matched my machine exactly (e.g. BenchmarkHeaderManipulation 616 B/op / 3 allocs, decision suite 40 allocs / 18888 B) while ns/op differed 2–5× and correctly stayed advisory. Gating on allocs/B-op instead of wall-clock is the right call.

#### Two things I found while testing, one worth addressing:
- A crashed benchmark suite still passes make perf-check locally. If go test -bench crashes partway, the | tee pipe hides the failure and the comparison just runs on whatever output exists. Benchmarks missing from the current run are silently skipped — I tested this by deleting the looper lines from bench-output.txt, and the gate printed No regressions detected with exit 0. CI isn't affected (it has pipefail), only local runs. Suggested fix: print a warning when baseline entries have no matching current result (the reverse of your UngatedBenchmarks check).

- Minor: perf-compare's guard message says "run 'make perf-bench' (tee its output there)", but perf-bench/perf-bench-quick don't tee to reports/bench-output.txt, so the natural make perf-bench && make perf-compare sequence fails with a hint implying it should have worked. Either add the tee to those targets or point the hint at perf-check.
Neither blocks the PR for me,  the core fix is solid and well-tested. LGTM....

theohsiung:
Thanks for the thorough review @TwilightTechie — and especially for verifying the hardware-independence empirically on macOS arm64. That the committed alloc/B baselines matched your machine exactly while ns/op differed 2–5× is exactly the property the design leans on, so it's great to see it confirmed on a different arch. 🙏

Both points addressed:

- **Missing benchmarks silently skipped** (the one worth fixing): added `MissingBenchmarks` — the reverse of the `UngatedBenchmarks` check — so `perftest` now prints a warning listing any *baseline* benchmark that produced no current result (crashed suite, rename, or filtered run). A disappeared benchmark is now visible instead of passing quietly. (`ca6eb8b1`)
- **Misleading `perf-compare` hint**: it now points at `make perf-check` (which actually produces `reports/bench-output.txt`) and shows the explicit `... 2>&1 | tee reports/bench-output.txt` alternative, instead of implying `make perf-bench` tees. (`560bf1cb`)

I left the local `| tee`-hides-a-crash behavior itself as-is (CI is protected by `pipefail`), since the new warning makes a partial run obvious locally too. Thanks again!

WUKUNTAI-0211 (commented):
Thanks for taking this on — `--compare-baseline` has been dead weight for a while and this makes it a real gate. I read through the diff and pulled the workflow + committed baselines at HEAD to check the wiring.

The core design is the right one: **blocking on `allocs/op` + `B/op` and treating `ns/op` as advisory.** Those are code-path-determined, not CPU-determined, so comparing a PR run against a baseline recorded on another machine is sound; wall-clock across shared runners isn't. The implementation matches — `RegressionDetected = allocsRegressed || bytesRegressed`, `NsAdvisory` is a separate flag that never feeds it, and `TestCompareWithBaseline_NsAdvisoryFlaggedNotBlocking` locks it in. Tests are thorough.

Two small things worth folding in:

**1. A `0 → N` new allocation is correctly blocked, but renders as "0%" in the PR comment.** `metricRegression(0, N)` returns `change=0, regressed=true`, and the summary prints `allocs/op \(.AllocsPerOpChange|round)%` — so the one line that actually failed the gate reads as a no-op. Suggest printing the raw counts instead:

```yaml
            blocked=$(jq -r '.results[]? | select(.RegressionDetected==true) | "- \(.BenchmarkName): allocs/op \(.Baseline.allocs_per_op // 0)→\(.Current.allocs_per_op // 0), B/op \(.Baseline.bytes_per_op // 0)→\(.Current.bytes_per_op // 0)"' reports/comparison.json 2>/dev/null || true)
```

`0→3` is unambiguous where `0%` misleads, and the normal case still reads fine (`100→108`). (`allocs_per_op`/`bytes_per_op` are `omitempty`, so the `// 0` fallback covers a value of 0 being dropped from the JSON.)

**2. A malformed `pattern:` silently relaxes the gate.** `getThresholdsForBenchmark` calls `regexp.MatchString` and skips the entry on `err != nil`, so a typo'd pattern drops that benchmark to `default` — loosening a tight bound (decision-engine 5% → 10%) with no signal. Since the point is a trustworthy gate, worth failing fast at config load:

```go
// in LoadThresholds, before `return &thresholds, nil` (add "regexp" to imports)
for _, b := range thresholds.ComponentBenchmarks.Benchmarks {
    if b.Pattern == "" {
        continue
    }
    if _, err := regexp.Compile(b.Pattern); err != nil {
        return nil, fmt.Errorf("invalid regexp for benchmark %q: %w", b.Name, err)
    }
}
```

Both are small; happy to approve once they're in. Nice, well-tested cleanup.

.github/workflows/performance-test.yml
```diff
@@ -123,11 +150,26 @@ jobs:
...
+            headline="Component benchmarks completed with no regressions beyond thresholds."
           else
             headline="❌ Component benchmarks **FAILED** — see the [workflow logs](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})."
           fi
+          # Blocking (allocs/op or B/op over threshold) and advisory (ns/op
+          # slower — reported, never blocks) lists, pulled from the comparison.
+          blocked=""
+          advisory=""
+          if [ -f reports/comparison.json ]; then
+            blocked=$(jq -r '.results[]? | select(.RegressionDetected==true) | "- \(.BenchmarkName): allocs/op \(.AllocsPerOpChange|round)%, B/op \(.BytesPerOpChange|round)%"' reports/comparison.json 2>/dev/null || true)
```
WUKUNTAI-0211: A `0 → N` new allocation is correctly blocked, but `metricRegression(0, N)` returns `change=0`, so this line prints it as `allocs/op 0%` — the one entry that actually failed the gate reads as a no-op. Suggest raw counts instead:

```yaml
            blocked=$(jq -r '.results[]? | select(.RegressionDetected==true) | "- \(.BenchmarkName): allocs/op \(.Baseline.allocs_per_op // 0)→\(.Current.allocs_per_op // 0), B/op \(.Baseline.bytes_per_op // 0)→\(.Current.bytes_per_op // 0)"' reports/comparison.json 2>/dev/null || true)
```

`0→3` is unambiguous where `0%` misleads; the normal case still reads fine (`100→108`). (`allocs_per_op`/`bytes_per_op` are `omitempty`, hence the `// 0` fallback.)

perf/pkg/benchmark/config.go
```diff
@@ -94,11 +94,32 @@ type ThresholdsConfig struct {
...
+	MaxAllocsRegressionPercent float64 `yaml:"max_allocs_regression_percent"`
+	MaxBytesRegressionPercent  float64 `yaml:"max_bytes_regression_percent"`
+	MaxNsRegressionPercent     float64 `yaml:"max_ns_regression_percent,omitempty"`
+}
+
+// BenchmarkRegressionThreshold maps a benchmark-name regexp to its thresholds.
+// Name is a human label for readability only.
+type BenchmarkRegressionThreshold struct {
+	Name                string `yaml:"name"`
+	Pattern             string `yaml:"pattern"`
```
WUKUNTAI-0211, line 122: A typo in this `pattern` is silently absorbed: `getThresholdsForBenchmark` uses `regexp.MatchString` and skips the entry on error, so an invalid regexp drops the benchmark to `default` and quietly loosens the gate (e.g. decision-engine 5% → 10%) with no signal. Worth failing fast in `LoadThresholds`:

```go
// before `return &thresholds, nil` (add "regexp" to imports)
for _, b := range thresholds.ComponentBenchmarks.Benchmarks {
    if b.Pattern == "" {
        continue
    }
    if _, err := regexp.Compile(b.Pattern); err != nil {
        return nil, fmt.Errorf("invalid regexp for benchmark %q: %w", b.Name, err)
    }
}
```

WUKUNTAI-0211 (commented):
Dropping my two earlier points inline on the exact lines for easy reference.

theohsiung:
Thanks @WUKUNTAI-0211 — both folded in:

1. **Invalid `pattern` silently loosening the gate**: `LoadThresholds` now compiles every `pattern` and fails fast on an uncompilable regexp (`invalid regexp for benchmark %q`), with a test covering it. So a typo can't quietly drop a benchmark to `default` anymore. (`e9c195b0`)
2. **`0 → N` rendering as `0%`**: the blocked-regression list now prints raw baseline→current counts with the percent in parentheses, and `(new)` when the baseline is 0 — e.g. `allocs/op 0→3 (new)` and `allocs/op 100→108 (+8%)`. The line that actually failed the gate now reads unambiguously. (`d5e7b211`)

Verified locally (golangci-lint agent config clean, `-race` green, jq output checked against a crafted 0→N case). Thanks for the careful read! 🙏

perf/pkg/benchmark/config.go
theohsiung: Fixed in `e9c195b0` — `LoadThresholds` now `regexp.Compile`s every `pattern` and fails fast on an uncompilable one (`invalid regexp for benchmark %q`), with a test (`TestLoadThresholds_InvalidPatternFails`). A typo can no longer quietly drop a benchmark to `default`.

.github/workflows/performance-test.yml
theohsiung: Fixed in `d5e7b211` — the blocked list now prints raw baseline→current counts with the percent in parentheses, and `(new)` when the baseline is 0: e.g. `allocs/op 0→3 (new)` and `allocs/op 100→108 (+8%)`. So the line that actually failed the gate reads unambiguously. Verified the jq locally against a crafted 0→N comparison.json.

theohsiung:
@WUKUNTAI-0211 @TwilightTechie follow-up: since the review surfaced a couple of edge cases we'd missed, I did a deliberate adversarial positive/negative pass over the gate before asking for another look. No new bugs — added `adversarial_test.go` (34 tests total, `-race` clean):

- **metricRegression boundaries**: exactly-at-threshold passes (strict `>`), a decrease never regresses, `0→0` isn't "new" but `0→N` regresses even at a 0% threshold.
- **Mixed direction**: a benchmark that allocates fewer times but far more bytes is still blocked (on bytes), and the allocs delta reads as the improvement it is.
- **ParseBenchOutput on hostile input**: a crashed `--- FAIL` benchmark is not parsed as a result (it becomes a MissingBenchmarks warning instead), a no-`benchmem` line parses with 0 allocs, and a leading custom metric (`MB/s`) doesn't shift the ns/B/allocs column mapping.

Three behaviors are intentional, but worth calling out in case anyone disagrees:
1. A missing/vanished baseline benchmark warns but does not fail the gate (exit 0) — legit renames/filters shouldn't turn it red, and CI's `pipefail` catches a real crash at the bench step. Happy to add a `--fail-on-missing` flag if a stricter local gate is preferred.
2. A benchmark whose allocs drop to 0 reads as an improvement (only a risk if `-benchmem` is dropped, which CI hardcodes).
3. Exactly-at-threshold passes (`max_regression_percent` = max allowed).

Thanks again for the careful reviews 🙏

WUKUNTAI-0211 (commented):
Three-site display-precision fix — `%.0f` on the new `float64 NsPerOp` drops the sub-microsecond decimal the model and JSON now round-trip. Suggest `%.1f`. Inline on each site.

perf/pkg/benchmark/baseline.go
```diff
@@ -197,32 +288,33 @@ func PrintComparisonResults(results []ComparisonResult) {
...
-				result.Baseline.ThroughputQPS,
-				result.Current.ThroughputQPS,
-				result.ThroughputChange,
-			)
+		// Advisory metric: ns/op (machine-dependent, never blocks).
+		advisory := ""
+		if result.NsAdvisory {
+			advisory = "  ⚠️ advisory"
 		}
+		fmt.Printf("  └─ ns/op (adv): %-15.0f %-15.0f %+.2f%%%s\n",
```
WUKUNTAI-0211: 🟡 `%-15.0f` rounds decimal `ns_per_op` to an integer in the stdout table. `NsPerOp` was changed to `float64` specifically to preserve sub-microsecond decimals (baseline JSON keeps `628.5` verbatim), and `calculatePercentChange` computes the % change from the true float — but the displayed columns are rounded, so a reader sees e.g. `628 → 629  +0.08%` and the arithmetic doesn't line up. Sub-us benchmarks (Looper family, some cache paths) are exactly where the decimal matters.

Suggest `%-15.1f` so the column keeps its width and the precision that rc#4 explicitly round-trips shows in the table too.

perf/pkg/benchmark/report.go
```diff
@@ -117,7 +117,7 @@ func (r *Report) SaveMarkdown(path string) error {
 		}
 
 		// ns/op row
-		md.WriteString(fmt.Sprintf("| %s | ns/op | %d | %d | %+.2f%% | %s |\n",
+		md.WriteString(fmt.Sprintf("| %s | ns/op | %.0f | %.0f | %+.2f%% | %s |\n",
```
WUKUNTAI-0211: 🟡 Same precision loss as `baseline.go:312` — the Markdown `ns/op` row rounds `628.5 → 629`. Since rc#4's whole point is that decimal `ns/op` round-trips through the model and JSON, the report should show it too. Suggest `%.1f` (or `%g` for shortest form, though `%.1f` reads cleaner in an aligned column).

perf/pkg/benchmark/report.go
```diff
@@ -222,7 +222,7 @@ func (r *Report) SaveHTML(path string) error {
 			statusText = "IMPROVED"
 		}
 
-		html.WriteString(fmt.Sprintf(`            <tr><td>%s</td><td>ns/op</td><td>%d</td><td>%d</td><td>%+.2f%%</td><td class="%s">%s</td></tr>`,
+		html.WriteString(fmt.Sprintf(`            <tr><td>%s</td><td>ns/op</td><td>%.0f</td><td>%.0f</td><td>%+.2f%%</td><td class="%s">%s</td></tr>`,
```
WUKUNTAI-0211: 🟡 Same in the HTML report — `%.0f` drops the decimal that `float64 NsPerOp` and the JSON baseline both preserve. Suggest `%.1f` for consistency with the Markdown row above.

perf/pkg/benchmark/baseline.go
theohsiung: Fixed in `cb639d3f` — `%-15.1f` here (and `%.1f` at the two report.go sites), so the sub-microsecond decimal that `float64 NsPerOp` round-trips now shows in the table and lines up with the computed %.

perf/pkg/benchmark/report.go
theohsiung: Fixed in `cb639d3f` — Markdown `ns/op` row now uses `%.1f`.

perf/pkg/benchmark/report.go
theohsiung: Fixed in `cb639d3f` — HTML `ns/op` cell now uses `%.1f`, consistent with the Markdown row.

TwilightTechie:
LGTM now 👍

The adversarial pass covers exactly the failure modes I was worried about for the gate wiring, especially `ParseBenchOutput` treating a crashed `--- FAIL` benchmark as a `MissingBenchmarks` warning instead of a parsed result, and the custom-metric column-shift case (that one is a classic silent corrupter of ns/B/allocs mappings).

WUKUNTAI-0211 (dismissed):
Precision fix in — decimal ns/op flows through stdout table + Markdown + HTML. LGTM.

WUKUNTAI-0211 (approved):
LGTM

PR #2507, opened 2026-07-13
[Security] Bound external RAG API payloads

### LINKED ISSUE #2478: [Bug] Bound and type external RAG request templates and responses
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Make external RAG request construction typed and enforce an exact maximum successful-response size. Owner surface: `src/semantic-router/pkg/extproc/req_filter_rag_external.go`.

### To Reproduce

### Expected behavior

User values cannot change the configured request structure, final requests are valid typed JSON, and successful response decoding stops at a configured hard byte limit with an explicit error.

### Impact

##### Current behavior

Configured request bodies use direct string replacement for user values, so ordinary escaping characters can invalidate or alter the outbound JSON structure. Successful responses are decoded without a byte limit even though error text and transport time are bounded.

##### Acceptance

- Prefer a typed request builder; if templates remain, replace typed JSON nodes and validate the final document.
- Marshal every user substitution rather than performing raw string replacement.
- Enforce a configurable exact maximum before successful-response decoding.
- Return a clear limit error without accepting a truncated prefix.

##### Validation

- Add focused tests for quotes, backslashes, line breaks, Unicode, marker substrings, invalid templates, and field-boundary invariance.
- Add exact-limit success, one-byte-over rejection, chunked response, and cancellation tests.
- Run `cd src/semantic-router && go test -race ./pkg/extproc`.
- Run the affected external RAG E2E profile.

##### Related

- Parent audit: #2375
- This is a P2 API-boundary follow-up and is separate from dashboard outbound-fetch policy.

iroiro147:
Closes #2478

## Purpose

- Parse custom external-RAG request templates as JSON before substitution so user content is encoded safely while `top_k` and `threshold` remain typed values.
- Reject malformed or trailing template data and reject placeholders in object keys, while preserving the existing `${...}` and `{{...}}` placeholder forms.
- Add `max_response_body_bytes` for successful external-RAG responses, defaulting to 16 MiB. The complete body is read with a bounded `limit + 1` strategy so an exact-limit response succeeds and a one-byte-over response fails without accepting a truncated JSON prefix.
- Preserve external-RAG runtime placeholders through config environment expansion, validate nested hybrid external-API settings, and document the new contract.
- Add a dedicated Kubernetes E2E profile that exercises typed substitution and both sides of the response-size boundary through the deployed gateway stack.

Affected modules: `Router` / `E2E` / `Docs`

## Test Plan

- `cd src/semantic-router && go test -race ./pkg/config ./pkg/extproc`
- `cd e2e && go test ./profiles/rag-external-api ./testcases`
- `make agent-lint AGENT_BASE_REF=origin/main AGENT_SKIP_PRECOMMIT_BASELINE=1 CHANGED_FILES="$(git diff --name-only origin/main...HEAD)"`
- `python3 tools/security/ast_security_scanner.py diff origin/main --fail-on HIGH`
- `BUILDKIT_PROGRESS=plain make e2e-test E2E_PROFILE=rag-external-api E2E_TESTS=rag-external-api`

The focused unit/race tests cover malformed templates, typed and embedded placeholders, configured-number preservation, exact and one-byte-over limits, chunked reads, cancellation, and nested validation. The Kubernetes profile verifies the same contract through the rendered configuration, Semantic Router, Envoy Gateway, AI Gateway, and an external-RAG mock.

## Test Result

- Focused race-enabled Go tests passed.
- E2E profile and test-case packages compiled successfully.
- Changed-file agent lint, YAML lint, diff hygiene, and the AST PR-diff security threshold passed.
- The real Kind deployment completed with every pod ready and zero restarts. `rag-external-api` passed and reported `typed_request_verified`, `exact_limit_success_verified`, and `one_byte_over_rejected` as true.

No known blocker remains in the changed surface.

---
<details>

</details>


src/semantic-router/pkg/extproc/req_filter_rag_external.go
```diff
@@ -208,6 +222,28 @@ func (r *OpenAIRouter) retrieveFromExternalAPI(traceCtx context.Context, ctx *Re
...
 
+func externalAPIResponseBodyLimit(apiConfig *config.ExternalAPIRAGConfig) int64 {
+	if apiConfig.MaxResponseBodyBytes != nil {
+		return *apiConfig.MaxResponseBodyBytes
+	}
+	return defaultExternalAPIMaxResponseBodyBytes
+}
+
+func readExternalAPIResponseBody(reader io.Reader, maxBytes int64) ([]byte, bool, error) {
+	readLimit := maxBytes
```
Xunzhuo: [P2] Do not accept `MaxInt64` as an exact hard limit. This branch cannot read the sentinel byte, so an over-limit stream is indistinguishable from an exact-limit stream; in practice `io.ReadAll` will exhaust memory first. Validation currently explicitly accepts this value. Reject values for which `limit+1` cannot be represented (preferably cap to a sane deployment maximum) and test one-byte-over at the largest allowed value.

src/semantic-router/pkg/config/rag_plugin_validation.go
```diff
@@ -79,6 +81,11 @@ func validateExternalAPIRAGBackend(c *RAGPluginConfig) error {
 	if apiConfig.RequestFormat == "" {
 		return fmt.Errorf("request format is required for external API")
 	}
+	if apiConfig.MaxResponseBodyBytes != nil {
```
Xunzhuo: [P2] Validate the full external/custom request contract at config load. `request_format: custom` with an empty, malformed/trailing-JSON template or placeholders in object keys still passes `Validate` and fails every request; arbitrary unsupported non-empty formats also pass until runtime. Under `skip`/`warn`, that makes RAG silently inert. Enforce the allowed formats and compile/validate custom templates here (including nested hybrid configs), ideally sharing the runtime parser representation.

Xunzhuo (changes requested):
The normal response-size boundary and typed custom-template substitution are well designed, and the new real E2E covers hostile text plus exact/one-byte-over limits. Two edge contracts still need to fail at config load rather than degrading into runtime failure or an effectively unbounded read. `make agent-report`, `make build-e2e`, config tests, `git diff --check`, and project CI passed; the branch is behind main.

src/semantic-router/pkg/extproc/req_filter_rag_external.go
iroiro147: Addressed in f03d5b8. Config validation now rejects max_response_body_bytes values at or below zero and above the 64 MiB deployment cap, and the runtime guard enforces the same bound before reading the sentinel byte. Tests cover the largest permitted value, one above it, MaxInt64 rejection, exact-limit acceptance, and one-byte-over rejection.

src/semantic-router/pkg/config/rag_plugin_validation.go
iroiro147: Addressed in f03d5b8. Load-time validation now restricts external_api request_format to pinecone, weaviate, elasticsearch, or custom and compiles custom templates through the same typed parser used at runtime, including hybrid primary and fallback configs. The parser rejects empty, malformed or trailing JSON, unsupported placeholders, and placeholders in object keys. The Python CLI validation surface now enforces the same format inventory.

src/vllm-sr/cli/models.py
```diff
@@ -801,6 +810,13 @@ class RAGPluginConfig(BaseModel):
         description="Minimum confidence for triggering retrieval",
     )
 
+    @model_validator(mode="after")
+    def validate_external_api_request_format(self):
+        """Validate the router-owned request format without rewriting backend data."""
+        if self.backend == "external_api":
```
FAUST-BENCHOU, line 1254: The Go-side loader now validates nested hybrid `external_api` configs, but the Python config model only checks the top-level `backend == "external_api"` case. A hybrid config with an `external_api` fallback and `request_format: openai` would still pass the CLI validation and only fail later in the router. Could we mirror the nested hybrid check here as well, so the CLI and runtime keep the same contract this PR is trying to enforce?

iroiro147:
Good catch — thank you for spotting the asymmetry between the CLI and runtime contracts.

Addressed in c3726af. The Python `RAGPluginConfig` validator now mirrors the Go-side `validateHybridRAGChildBackend`: when `backend == "hybrid"`, it checks each child (`primary` and `fallback`) and validates any child typed as `external_api` through the same `ExternalAPIRAGBackendConfig` contract as a top-level `external_api` backend.

This means a hybrid config with an `external_api` fallback and `request_format: openai` will now fail at CLI validation time, the same way it would fail in the router.

New test coverage (`tests/test_rag_external_api_hybrid_validator.py`, 14 cases):
- Hybrid primary = external_api: accepts valid, rejects invalid/missing format
- Hybrid fallback = external_api: accepts valid, rejects invalid/missing format
- Both children = external_api: rejects if either is invalid
- Non-external_api hybrid children: unaffected

All 37 existing CLI tests still pass.

src/vllm-sr/cli/models.py
iroiro147: Addressed in c3726af ("fix(rag): validate hybrid external_api children in Python CLI") — the Python model validator now checks `backend == "hybrid"` and runs the same `ExternalAPIRAGBackendConfig` validate against each `primary`/`fallback` child typed as `external_api`, mirroring the Go `validateHybridRAGChildBackend`. Coverage is in `src/vllm-sr/tests/test_rag_external_api_hybrid_validator.py` (hybrid fallback + hybrid primary + bad `request_format`). Also pushed 45d194f to keep ruff green on that new test so the `Run pre-commit hooks check file lint` job is happy. PTAL.

iroiro147:
CI note: the single failing check `test-and-build` fails only on `TestHybridCachePendingRequest` in `pkg/cache` ("Expected to find cached entry after update / Response mismatch: got , want {\"answer\": \"Quantum computing uses qubits...\"}"). That test is outside this PR's diff, which touches `pkg/extproc/req_filter_rag_external.go`, `pkg/config/rag_plugin_validation.go`, `pkg/config/rag_external_api_template.go` and the Python CLI RAG validator only. The failure reads as an upstream flake (the same `test-and-build` job has been red on `main` repeatedly over the last day). Could a maintainer rerun `test-and-build`? I do not have the admin rights to re-run it myself. All three open review threads (P2 response-size limit and both custom-template validation points) are addressed at the current head `c2c67c8c`; if the rerun still fails on the same unrelated test I'll rebase onto current main and push a fresh roll.

iroiro147:
Rebased onto current main (78 commits). Resolved one config/README.md merge conflict by keeping the external-API-RAG fragment docs paragraph and adopting upstream's rewritten repository-enforcement paragraph.

Verified the two edge-contract asks against the rebased tree:

1. **`max_response_body_bytes` MaxInt64 rejection** — already enforced at `src/semantic-router/pkg/config/rag_plugin_validation.go:96-105`: `MaxResponseBodyBytes` values > `MaximumExternalAPIResponseBodyBytes` (64 MiB cap, defined in `rag_plugin_backends.go:12`) fail config validation. The MaxInt64 case is explicitly tested at `rag_external_api_template_test.go:127` (`wantErr: "must not exceed"`).

2. **Custom template + format validation at config load** — already enforced at `src/semantic-router/pkg/config/rag_plugin_validation.go:82-95`: `ExternalAPIRequestFormatPinecone/Weaviate/Elasticsearch` are the only accepted non-custom formats; `ExternalAPIRequestFormatCustom` requires `ParseExternalAPICustomRequestTemplate` to succeed (empty/malformed/trailing-JSON/placeholder-in-keys all fail); any other format (incl. empty string or unknown) returns an explicit error. Hybrid primary/fallback children typed as `external_api` go through the same validator via `validateHybridRAGChildBackend` (Go side) and the mirrored Python `RAGPluginConfig.validate_external_api_request_format` added in c3726af.

Please let me know if you'd like any adjustments.

iroiro147:
@Xunzhuo this is now conflict-free again — refreshed onto current main (`9ee68c43`), adopting the `config/fragments/` layout for the external-api fragment and docs, with the `rag-external-api` e2e profile entry moved under the existing `manual_profile_rules` header rather than a duplicate one.

Both review asks remain addressed and covered by tests:
1. `max_response_body_bytes` — load-time validation rejects ≤0 and >64 MiB (so `MaxInt64` is rejected), and the runtime guard enforces the same bound before the sentinel byte read.
2. `request_format` contract — load-time validation restricts formats to `pinecone`/`weaviate`/`elasticsearch`/`custom` and compiles custom templates (including hybrid children, mirrored in the Python CLI per @FAUST-BENCHOU's ask).

Ready for re-review whenever suits you.

iroiro147:
✅ @Xunzhuo — addressed both edge-contract asks in f989efc. The two contracts now fail at config load:

1. **response-size boundary** — `validateExternalAPIRAGBackend` now rejects `max_response_body_bytes > MaximumExternalAPIResponseBodyBytes` (config load, not runtime).
2. **typed custom-template substitution** — `ParseExternalAPICustomRequestTemplate` is enforced on `RequestFormat: custom` at config load, rejecting malformed/unsupported/empty templates.

Plus the hybrid-recursion guard: `validateHybridRAGChildBackend` both (a) prevents hybrid-hybrid recursion and (b) validates each child against the full backend contract. All three units now PASS on local Go tests (`go test ./pkg/config/... -run 'Validate|Custom'`).

iroiro147:
Rebased onto current `main`, and the scope of this PR has genuinely shrunk — worth flagging rather than burying in a merge commit.

**Half of this PR is now redundant: `main` has implemented external-RAG response bounds independently.** `decodeExternalRAGResponse` + `httputil.ReadLimitedBody`, configured through `ExternalAPIRAGConfig.MaxResponseBytes`. I checked it against the property this PR was protecting rather than assuming overlap: `ReadLimitedBody` reads `maxBytes+1` and **errors** when the body is longer, so a truncated body never reaches the decoder and `json.NewDecoder(...).Decode` cannot accept a valid truncated prefix. That is the same sentinel-byte technique this PR used, and it is correct.

So I have dropped my half of it entirely rather than carry a duplicate:

- removed `externalAPIResponseBodyLimit` / `readExternalAPIResponseBody` in favour of `main`'s decode path
- removed `MaxResponseBodyBytes`, `DefaultExternalAPIMaxResponseBodyBytes`, `MaximumExternalAPIResponseBodyBytes` — `main`'s `MaxResponseBytes` is the shipped field name and this PR should not introduce a second one
- removed the three tests that covered the superseded validation

**What remains is the half `main` does not have: request-template typing.** `rag_external_api_template.go` and its tests — supported-format validation (`pinecone` / `weaviate` / `elasticsearch` / `custom`), typed non-null JSON object-or-array template roots, the reserved lowercase runtime tokens (`${user_content}`, `${top_k}`, `${threshold}`) rejected before environment expansion so they cannot collide with uppercase env references like `${RAG_TENANT}`, and full validation of hybrid children at load time. `main`'s `rag_plugin_validation.go` validates the similarity threshold and no template contract, so there is no overlap left.

Also folded in @wilsonwu's merge of `main` into this branch — thank you for that; I merged rather than force-pushed so it is preserved.

`go build ./pkg/config/...` and `go test ./pkg/config/...` green.

@Xunzhuo — given the narrowing, the PR title and description now overstate it. If you would prefer, I will retitle to something like `[Router][Docs] Type and validate external RAG request templates` and rewrite the description around what is actually left, so a reviewer is not looking for response-bounds code that is no longer here. Happy to do that on your word rather than churn the PR again unasked.

Xunzhuo (changes requested):
Thanks for narrowing the branch and preserving main’s response limit. One blocker remains: `ExternalAPICustomRequestTemplate.Render` is never called. `buildCustomRequest` still raw-replaces `ctx.UserContent` into the template, so quoted or escaped input can still break or reshape the JSON even though load-time validation passes. Please use the validated renderer in the runtime path and add an extproc regression that sends quotes and backslashes through `buildCustomRequest`.

iroiro147:
@Xunzhuo Fixed in `d5191d77`, and you were right that this was the blocker — validation that does not gate the runtime path is theatre, and that is exactly what this was.

`buildCustomRequest` string-replaced `ctx.UserContent` into the template *text*, so load-time validation and request execution were validating and performing **different operations**. A query containing a `"` could close a JSON string early and reshape the request while the config still passed validation. The old marker-escaping heuristic above it (`{{`, `}}`, `${` → backslash-prefixed) was guessing at injection rather than preventing it, and it did nothing about a bare quote.

It now parses with `ParseExternalAPICustomRequestTemplate` and calls `Render`, which substitutes into the *decoded document* and marshals — so user content is escaped structurally and cannot alter the request's shape. The heuristic is deleted rather than kept alongside; two mechanisms disagreeing about what is dangerous is how this happened.

`TestBuildCustomRequestEscapesHostileUserContent` sends six inputs through the real `buildCustomRequest` and asserts the body is valid JSON with exactly the template's three keys and a byte-identical `query` round-trip:

| input | what it attacks |
| --- | --- |
| `say "hello"` | plain quote |
| `x","top_k":999,"evil":"` | injected key, the one that reshapes the request |
| `path\` | trailing backslash |
| `a\"b` | pre-escaped quote |
| `line1\nline2\tend` | control characters |
| `${user_content} {{.Query}}` | literal placeholder markers in user input |

**One honest caveat on verification.** `pkg/extproc` does not link in my environment — it needs the `candle_semantic_router` Rust library — so I could not execute that test locally, and it will run for the first time in CI. To avoid asserting an untested property I exercised the same six inputs directly against `ParseExternalAPICustomRequestTemplate` + `Render` in `pkg/config`, which does build here, and all six pass with the shape and round-trip assertions intact. So the escaping behaviour is verified locally; the wiring is verified by CI.

If you would rather that regression live in `pkg/config` where it can run without the native dependency, say the word and I will move it — though I think it belongs in extproc precisely because the bug was the runtime path not using the validated renderer, and a config-level test would not have caught it.

Xunzhuo (dismissed):
Thanks for the update — this resolves the template-injection blocker. The runtime now uses the validated structural renderer, and the regression covers quotes, backslashes, marker text, and injected keys through `buildCustomRequest`. The focused template tests pass. Approved; the branch is conflicting with `main` and this head has no GitHub Actions results yet, so please rebase and let the required CI rerun.

iroiro147:
@Xunzhuo Thank you for the approval — rebased onto current `main` as you asked, and the branch is `MERGEABLE` again. The push dismissed the approval, so it needs your re-review when convenient.

Two conflicts came with the rebase, both resolved keeping `main`'s work:

- `rag_plugin_test.go` — took your new `max_response_bytes must be non-negative` specs for both the external-API and OpenAI backends.
- `rag.md` — kept `main`'s OpenAI `direct_search` limit note *and* this branch's line about the external HTTP API example, since they document different backends.

**One real fix the rebase surfaced.** Your new negative-limit spec failed against this branch, and the cause was ordering rather than a missing check: `validateExternalAPIRAGBackend` parsed the custom request template before it looked at `MaxResponseBytes`, so a config with `request_format: custom`, no template, and `max_response_bytes: -1` reported *"request template is required"* and never reached the limit check your spec asserts.

Moved the scalar check above the structural parse. That is the better order independently of the test — a config with several problems should report the simplest one rather than failing on template compilation first.

`go test ./pkg/config/...` — 315 passed, 0 failed. `ruff` clean on the Python side. The renderer fix and its regression are untouched by the rebase; I checked both are still wired rather than assuming.

Xunzhuo (changes requested):
Thanks for the rebase. I found one delivery blocker: `test_external_api_accepts_router_supported_request_formats` takes `request_format` but has no parameterization or fixture, so the focused pytest run errors before executing that case (21 passed, 1 error). Please parametrize the supported formats and rerun the two split RAG test files.

FAUST-BENCHOU (commented):
LGTM. Thanks for improving external RAG request safety with typed template handling and bounded response processing. The added validation and E2E coverage provide good protection against malformed payloads and oversized responses.

iroiro147:
@Xunzhuo Fixed in `bf9f8e01`, and thank you for running the focused suite — that error was mine and I could not have caught it the way I was verifying.

**The parametrize blocker.** When I split the external-API cases out of `test_plugin_parsing.py` to get under the 800-line limit, I moved the method body but left its `@pytest.mark.parametrize("request_format", [...])` decorator behind, so `request_format` became an unresolvable fixture. Restored with the original four formats. Root cause on my side is that I verified the split with `py_compile` and a line count rather than by executing pytest — the interpreter available to me at the time could not import `pydantic`. I have a working 3.11 environment now and ran it properly.

**One thing your fix surfaced that I want to flag rather than patch quietly.** With the decorator restored, two of `main`'s pre-existing tests failed:

```
test_valid_rag_config_all_fields
test_rag_plugin_in_full_config
```

Both configure an `external_api` backend with `request_format: openai`, and this PR rejects that. **That is intentional and it is a behaviour change**: `openai` is a distinct RAG *backend* with its own config shape, not an external-API request format, and this branch already carries `test_openai_file_search_backend_remains_distinct` asserting the two do not overlap. The old enum was permissive enough to accept a config that could never work at runtime.

I updated both to `pinecone`, which preserves what they actually assert — that every field round-trips — without weakening the new contract. But it does mean a config that validated before will now be rejected, so if you would rather keep `openai` accepted as a format alias, say so and I will widen the enum instead. I did not want to make that call silently inside a test fix.

**Verification, done properly this time:**

- `pytest tests/test_plugin_parsing.py tests/test_plugin_parsing_rag_external.py` — **36 passed**, no errors
- full `pytest tests/` — 31 failures, but the identical 31 fail on pristine `main` in this environment, so they are unrelated
- `black --check` and `ruff check --config tools/linter/python/.ruff.toml` — clean on both files
- `go test ./pkg/config/...` — 315 passed

Also: @FAUST-BENCHOU left an LGTM above. The branch is `BEHIND` again since `main` moved; happy to rebase once more whenever you are ready to take it, rather than churning it now.

Xunzhuo (changes requested):
Thanks for carrying the typed-template path through. One current-contract issue still blocks this: the profile sets max_response_body_bytes, but the Router decodes only max_response_bytes. The unknown field is ignored, so the test falls back to 4 MiB and never exercises its intended 38/39-byte boundary. Please use max_response_bytes and align the stale field, default, and cap in the fragment and docs; split registerRAGValidationSpecs to clear agent-lint; then rebase and run the rag-external-api profile on the exact head.

e2e/profiles/rag-external-api/values.yaml
```diff
@@ -0,0 +1,76 @@
+config:
+  version: v0.3
+  listeners: []
+  providers:
+    defaults:
+      default_model: general-expert
```
Xunzhuo, line 6: **[P2] Use the supported provider contract in the boundary-test profile**

The new profile sets providers.defaults.default_model, providers.defaults.reasoning_families, and providers.models[].reasoning_family. The loader at this head rejects these deprecated provider fields, so the Router cannot start and the new external-API response-size boundary cases never run. Use providers.defaults.model and providers.models[].reasoning (for example reasoning.family), then run this profile against the exact head.

Xunzhuo (changes requested):
The response-limit field is corrected, but the external RAG profile still uses provider fields rejected by the current loader, so it cannot exercise the boundary cases. Please migrate the fixture to the supported schema and run that profile on the updated head. I traced loader rejection; I have not run the profile.

iroiro147:
Fixed, and you were right about the cause — the fixture never reached the boundary code at all.

**The profile was rejected outright.** Three fields were still on the pre-v0.3 shape:

```
deprecated config fields are no longer supported: providers.defaults.default_model,
providers.defaults.reasoning_families, providers.models[0].reasoning_family;
rewrite the file to canonical v0.3 providers/routing/global
```

Migrated to the schema the sibling profiles already use — `defaults.model`,
`models[].reasoning.family`, and an explicit `backend_refs[].provider: vllm`.

**Verified against the real loader, not by inspection.** I ran the profile's own
`config:` block through `config.ParseYAMLBytes`, the same entry point the Router
uses: it fails with the error above on the old fixture and is accepted on the new
one. On the accepted parse the plugin payload comes out intact, so the boundary
cases are actually reachable now:

```json
{"backend":"external_api","backend_config":{"endpoint":"http://external-rag-mock.default.svc.cluster.local:8080/query",
"max_response_bytes":38,"request_format":"custom",
"request_template":"{\"query\":\"${user_content}\",\"top_k\":${top_k},\"threshold\":${threshold}}",
"timeout_seconds":15},"enabled":true,"injection_mode":"tool_role","on_failure":"block",
"similarity_threshold":0.625,"top_k":7}
```

**Also rebased onto main** (139 commits). That surfaced one more break: verification
entries now have to declare a `display_name`, and ours did not, so
`harness.py validate` and `tools/ci/tests/test_domain_registry.py` both failed.
Added.

Head is `88b3339e`. Green locally on the changed files: `harness.py validate`,
`structure_check.py` (exit 0), `go test ./pkg/config/...`, gofmt, ruff + black,
and 56 Python tests.

**On running the profile itself:** I have not been able to run the full
`rag-external-api` profile end to end — no container runtime available on this
machine, and the profile needs the Rust-inclusive image build. So the honest state
is: loader acceptance and payload integrity are verified directly, the in-cluster
run is not. If a maintainer run is easier than waiting on me, this head is ready
for it; otherwise I will report back once I can complete a local cluster run.

e2e/testcases/rag_external_api.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if oversizedResponse.StatusCode != http.StatusServiceUnavailable {
+		return fmt.Errorf(
+			"oversized external RAG response status = %d, want %d: %s",
+			oversizedResponse.StatusCode,
+			http.StatusServiceUnavailable,
+			oversizedBody,
+		)
+	}
+	if !bytes.Contains(oversizedBody, []byte("exceeded configured limit of 38 bytes")) {
```
Xunzhuo, line 109: [P2] Match the byte-limit error emitted by the runtime

This new profile cannot pass its oversized-response case: `ReadLimitedBody` returns `response body exceeds limit of 38 bytes`, which the RAG/ExtProc wrappers preserve, but this assertion requires `exceeded configured limit of 38 bytes`. I reproduced the actual canonical-profile loader → external HTTP retrieval path at this head: the 38-byte response succeeds and the 39-byte response is correctly rejected, then this exact substring check fails; changing only the expected substring makes the control pass. Please align the assertion with the emitted error (or a stable structured rejection) and exercise the profile so correct size enforcement does not fail its required integration check.

Xunzhuo (changes requested):
The previous canonical-profile configuration error and main conflicts are resolved. The new oversized-response assertion still fails when the runtime correctly rejects the response; please align it with the emitted error and run the profile.

At this exact head, credential-free native HTTP checks reproduce the 38/39-byte boundary failure; changing only the assertion makes all 17 checks pass. All 78 affected config checks pass. Earlier checks in this sweep include 56 CLI and 106 E2E framework tests; inspected CI records 454 harness tests on an equivalent source tree. Those scopes are unchanged by the final formatter-only delta.

The deployed Kind/Envoy rag-external-api profile, vllm-sr serve, and model-backed inference were not run. Current-head DCO/Mergify bookkeeping does not supply product-validation evidence.

iroiro147:
Follow-up on running the profile — I found a cheaper path than waiting on me.

`ci.yml` already accepts a verification ID by `workflow_dispatch`:

> `verification`: Verification ID from `tools/ci/verification_catalog.yaml`
> (or `e2e.PROFILE`) to run on the selected branch or tag.

and this PR's registration resolves to a complete record, so there is nothing
left to wire up:

```
e2e.rag-external-api -> {runner: ubuntu-24.04, executor: e2e,
  workflow: .github/workflows/integration-test-k8s.yml,
  profile: rag-external-api, images: [extproc],
  services: [kind, gateway, controlled-backend],
  contract: external-rag-request-and-response-boundary}
```

So a maintainer dispatch of `ci.yml` with `verification: e2e.rag-external-api`
against this head runs exactly the boundary cases you asked about, on your
runners rather than my laptop — which also makes the result reproducible for
anyone reviewing it later.

Current state on `465d8e41`: 12 checks green, nothing red, the rest still
building. That includes `Quality / Checks`, which is the gate that caught the two
things the rebase surfaced — import grouping, then `QF1012` on the placeholder
formatting. Both fixed. The formatting change is byte-identical to what it
replaced, which matters here because the profile asserts an exact payload size.

One note: the branch is `BEHIND` again — `main` moved while CI was running. I am
holding off on another merge so this run finishes and you can look at a settled
head. Say the word and I will bring it up to date, or let the merge queue do it.

iroiro147:
Fixed in `fce00974` — thank you for reproducing it down to the substring.

The assertion expected `exceeded configured limit of 38 bytes`, which is the wording from the response-limit half I dropped in favour of main. `ReadLimitedBody` emits `response body exceeds limit of 38 bytes` (`pkg/utils/http/readlimit.go:23`), so the case failed exactly when the runtime was correct. The assertion now matches that text; nothing else references the old wording.

Also merged main again (80 commits; one `imports.go` conflict with `provider-protocols`, both kept). Head `af700b72`. Re-checked on that head: the profile's `config:` block still parses through `ParseYAMLBytes` with `max_response_bytes: 38` intact, `harness.py validate`, domain-registry tests, `structure_check.py` (exit 0), changed-line golangci on `pkg/config`, `pkg/extproc` and the e2e packages (0 issues), `go test ./pkg/config/...`, ruff/black, and 56 Python tests.

I still have not run the deployed Kind/Envoy profile myself. If a dispatch of `ci.yml` with `verification: e2e.rag-external-api` is possible on your side, that is the fastest way to get the in-cluster result on this head.

Issue #2396, opened 2026-07-07
[Epic] Unify native Router Model adapters and capability discovery


Xunzhuo:
### Goal

Give native Router Model backends one adapter boundary and one capability model so callers do not depend on backend-specific initialization or feature guesses.

### Scope

- Expose typed capabilities for batching, modalities, devices, and result contracts.
- Fix native correctness and validation gaps that violate the shared adapter contract.
- Evaluate bounded adapter extensions such as base plus LoRA serving.
- Add qualified platform backends without changing the public model contract.

### Child issues

- #2194
- #2249
- #2394
- #2440
- #2668
- #2675
- #2926
- #3160

### Non-goals

- Defining routing decisions or signal composition.
- Owning generic outbound provider transport.
- Treating an experimental backend as supported without qualification evidence.

Parent Workgroup: #2966

Xunzhuo:
Audit evidence for #2375 adds classification refresh retirement, eager selector allocation, handle replacement and unload, Candle truthful initialization and bounded batching, and ONNX/NLP lock and lifecycle gaps. Runtime-generation ownership is tracked by #2470, and the required ONNX compile/CI contract is #2477.

PR #2515, opened 2026-07-13
[CI] Restore ONNX Go test compilation and add a model-free lane

### LINKED ISSUE #2477: [CI] Restore ONNX Go test compilation and make it mandatory
## Summary

Restore the ONNX Go binding test contract and add an always-required provider-independent compile and unit lane. Owner surfaces: `onnx-binding/semantic-router_test.go`, `onnx-binding/go.mod`, the ONNX Go wrapper, and `.github/workflows/`.

## Current behavior

The ONNX Go test package no longer compiles against current initialization and function signatures, and normal required CI does not catch the drift. Provider or model availability must not determine whether the base API contract is checked.

## Expected behavior

Provider-independent ONNX Go tests always compile and run in required CI, while explicitly selected provider/model receipts exercise runtime behavior without turning the base gate into a skip-only pass.

## Acceptance

- Update tests to current named model/runtime APIs and remove references to retired globals or signatures.
- Add a required model-free compile and unit job.
- Keep provider/model-dependent receipts explicitly selected and visibly skipped or failed.
- Cover lifecycle, error arity, Matryoshka shape/signature, and ABI/capability contracts.
- Link the required CI contract from #2396.

## Validation

- Run `cd onnx-binding && go test ./...`.
- Run the provider-independent job without model artifacts and prove it executes non-zero tests.
- Run the selected ROCm/provider receipt on supported hardware.
- Verify a deliberate API signature drift fails required CI.

## Related

- Parent audit: #2375
- Native runtime contracts: #2396
- Hardware matrix: #1510

Parent Epic: #3177

TwilightTechie:
Closes #2477

## Purpose

- Restores the ONNX Go binding test package, which no longer compiled against current initialization and function signatures (two-value `IsModelInitialized`, modelType-string Matryoshka/similarity signatures, slice-returning `GetEmbedding`/`GetEmbeddingWithDim`, retired `initOnce`/`modelInitialized` globals, removed `GetEmbeddingsBatch`/`GetMatryoshkaConfig`).
- Adds a **required, provider-independent compile + unit CI lane** (`.github/workflows/onnx-binding-ci.yml`) so this drift can never again land silently: CPU-features cargo build, `go vet`, `go test` with no model artifacts, and an explicit assertion that a non-zero number of tests passed (the gate cannot degrade into a skip-only pass).
- Keeps provider/model-dependent receipts explicitly selected via `workflow_dispatch` — visibly skipped otherwise, so provider or model availability never determines whether the base API contract is checked.
- Adds model-free lifecycle/ABI contract tests: fail-closed empty-path init through the FFI boundary, and rust/go initialization-state agreement.
- Affects: `Bindings` (onnx-binding) and `CI/Build`.
- CI contract linked from #2396 in the workflow header, as requested in the issue.

## Test Plan

- `cd onnx-binding && go vet ./...` — compiles the test package; signature drift fails here.
- `cd onnx-binding && go test -count=1 ./...` with `MMBERT_MODEL_PATH` unset — the provider-independent lane; model-gated tests skip visibly.
- Same command with `MMBERT_MODEL_PATH` pointing at `mmbert-embed-32k-2d-matryoshka` — full model receipts.
- Deliberate-drift check: change any call to a retired signature and confirm `go vet`/the CI lane fails.
- `make agent-lint CHANGED_FILES=".github/workflows/onnx-binding-ci.yml,onnx-binding/semantic-router_test.go"` — repo canonical lint gate.

## Test Result

- Model-free lane: **47 passed, 20 skipped (model-gated), 0 failed** on macOS arm64 against a CPU `cargo build --release` of the binding.
- Deliberate signature drift (single-value use of `IsModelInitialized`) fails `go vet` as intended; restored file is clean.
- `make agent-lint` passes (go fmt, golangci structural lint, yamllint, supply-chain scan).
- Gap/follow-up: the model-receipts job downloads `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` at dispatch time; if the HF repo layout changes, that job (and only that job) needs updating. The lane runs CPU-only; ROCm/provider receipts on supported hardware remain manual per #1510.

---
<details>

</details>


.github/workflows/onnx-binding-ci.yml
```diff
@@ -0,0 +1,167 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Provider-independent compile and unit lane. Requires no model artifacts and
+  # no GPU/provider; must always run and must execute a non-zero number of tests.
+  compile-and-unit:
+    if: ${{ !github.event.pull_request.draft }}
```
Xunzhuo, line 59: [P1] Make this part of an actually required status. The live `main` branch protection currently requires only `Run pre-commit hooks check file lint` and `test-and-build`, and Mergify relies on that protection; `compile-and-unit` is therefore optional and a failed or skipped run can still merge. Please invoke this receipt from an existing required status (for example `test-and-build`) or coordinate adding this exact job to branch protection and verify it with `gh pr checks --required`.

.github/workflows/onnx-binding-ci.yml
```diff
@@ -0,0 +1,167 @@
...
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            onnx-binding/target
+          key: onnx-binding-cargo-${{ runner.os }}-${{ hashFiles('onnx-binding/Cargo.lock') }}
+
+      - name: Build ONNX Rust library (CPU, default features)
+        run: |
```
Xunzhuo, line 84: [P1] Add a real cross-language ABI layout receipt here. Go manually redeclares C structs/functions while Rust separately owns the `#[repr(C)]` layouts; the added Go tests only instantiate Go-facing wrapper values, and `cargo build` checks symbols/signatures but not size, alignment, field offsets, or enum discriminants. A Rust field reorder can still compile and corrupt memory. Please export/assert layout constants across cgo (or generate/use a shared header) and run the relevant Rust ABI tests in this lane.

Xunzhuo (changes requested):
The model-free lane itself is healthy (the exact local run completed 47 tests, skipped 20 model-gated tests, and failed 0), but it does not yet establish the mandatory ABI gate claimed by #2477/#2396. Please make the status actually merge-blocking and add a real cross-language layout receipt. `cargo build --release`, `go vet ./...`, the model-free Go run, and `git diff --check` otherwise passed.

onnx-binding/semantic-router_test.go
```diff
@@ -26,6 +26,46 @@ func getModelPath(t *testing.T) string {
...
+	if err := InitMmBertEmbeddingModel(modelPath, true); err != nil {
+		t.Fatalf("Failed to initialize model: %v", err)
+	}
+}
+
+// TestIsModelInitializedContract is a model-free ABI contract test: the
+// candle_binding compatible IsModelInitialized must report (rust, go) state
+// with both values agreeing, whether or not a model has been loaded.
+func TestIsModelInitializedContract(t *testing.T) {
+	rustState, goState := IsModelInitialized()
```
FAUST-BENCHOU: This test cannot actually verify Rust/Go state agreement because `IsModelInitialized()` reads the Rust state once and returns that same value twice. The equality assertion is therefore guaranteed to pass. If there is no independent Go-side state anymore, could we describe this as an arity/FFI-state contract instead; otherwise the test needs to compare two independently owned states?

TwilightTechie:
Thank you for review working on it

.github/workflows/onnx-binding-ci.yml
TwilightTechie: Done

.github/workflows/onnx-binding-ci.yml
TwilightTechie: Added: `src/ffi/abi_receipt.rs` exports [size, align, field offsets...] for all 10 shared #[repr(C)] structs, and `TestABILayoutAgreement` asserts byte-for-byte agreement against cgo's own view of the preamble declarations.

onnx-binding/semantic-router_test.go
TwilightTechie: You're right, with the Go-side globals retired there's only one state to read, so the equality was tautological. Reframed as TestIsModelInitializedArityContract: it pins the candle_binding two-value arity and wrapper consistency with IsMmBertModelInitialized, and the comment now says exactly that instead of claiming cross-state agreement. Thanks for the precise read and feedback.

TwilightTechie:
Hi @FAUST-BENCHOU would you mind reviewing this again. I have worked on all the changes.

Xunzhuo (commented):
Thanks for making the lane mandatory and adding cross-language size, alignment, and offset receipts for every struct shared by the cgo preamble. I don't see a new blocker and exact-head CI is green, but the branch conflicts with current main. Please rebase it for the final readiness review.

FAUST-BENCHOU (commented):
Thanks for restoring the ONNX binding test coverage and adding a provider-independent contract lane. Ensuring the Go/Rust ABI compatibility is checked continuously helps prevent future signature drift.

Issue #2474, opened 2026-07-12
[Bug] Make vector-store ingestion cancellable, bounded, and transactional


Xunzhuo:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Add a root lifecycle, bounded streaming ingestion, status retention, constructor rollback, and transactional reconciliation to vector storage. Owner surfaces: `src/semantic-router/pkg/vectorstore/`, `src/semantic-router/pkg/routerruntime/vectorstore_runtime.go`, and `src/semantic-router/pkg/config/vectorstore.go`.

### To Reproduce

Use the reproduction or current-code evidence documented in this issue to exercise the affected path.

### Expected behavior

Jobs and backend work cancel within a deadline, ingestion uses bounded streaming batches and workers, status has a retention policy, and every registry/backend/file mutation either commits coherently or reconciles through an explicit saga state.

### Impact

Not separately specified in the original report.

##### Current behavior

Active jobs do not receive a cancellable root context, file processing expands into full in-memory text/chunk/vector sets, completed status can grow indefinitely, and multi-store metadata mutations or constructor failures can leave leaked or inconsistent state.

##### Acceptance

- Derive job, embedding, backend, and count-update contexts from a lifecycle root; implement bounded `Stop(ctx)`.
- Stream parse, chunk, embed, and insert through bounded batches with file, worker, queue, and memory limits.
- Apply TTL, cardinality, and persistence policy to completed and failed status.
- Roll back constructor and metadata mutation side effects or persist a reconcilable pending state.
- Make shutdown idempotent and join all cleanup errors.

##### Validation

- Add permanently blocking embedder/backend, stop-deadline, restart, max-size concurrency, and high-cardinality status tests.
- Fault-inject after every registry, backend, memory, and filesystem side effect, then restart and verify convergence.
- Run `cd src/semantic-router && go test -race ./pkg/vectorstore ./pkg/routerruntime`.
- Add memory and goroutine bounds to the affected vector ingestion E2E profile.

##### Related

- Parent audit: #2375
- Runtime generation issue created from RR-01

Peterren:
## Proposed breakdown into independently shippable PRs

This is large enough that landing it as one PR would be hard to review and risky to merge. Below is a decomposition into 9 core PRs across 4 tracks, ordered by dependency. Each PR compiles, tests, and reviews on its own, and each delivers standalone value — so they can be distributed to different contributors and merged incrementally without ever leaving the tree in a half-broken state.

[...]

PR #2517, opened 2026-07-13
[Router] vectorstore: add lifecycle root context and bounded Stop(ctx)

### LINKED ISSUE #2602: [Feature] Add a lifecycle root and bounded idempotent Stop(ctx) to vector storage
## Summary

## Current behavior

## Expected behavior

## Validation

## Related

- Parent umbrella: #2474

Peterren:
## Purpose

- Give the vector-store ingestion pipeline a cancellable lifecycle root context and replace the unbounded `Stop()` with a bounded `Stop(ctx) error`.
- Today `processJob` runs with a hardcoded `context.Background()` and `Stop()` waits on an unbounded `wg.Wait()`. If an embedder or backend wedges inside a stage, `Stop()` blocks forever and the process cannot shut down. This is the first, foundational slice of the larger vector-store hardening effort (#2474): make ingestion **cancellable** and shutdown **bounded**.
- Module: `Router` (plus `E2E`-style integration tests within the `vectorstore` package).

### What changed

- `IngestionPipeline` gains a lifecycle root context created in `Start()` and cancelled on shutdown. Workers and `processJob` derive their work from it.
- `processJob` checks `ctx.Err()` between stages (read → chunk → per-chunk embed → insert), so a cancelled lifecycle aborts in-flight jobs at the next checkpoint instead of running to completion. Cancelled jobs are marked failed with a `cancelled` error code; count/metadata updates use `context.WithoutCancel` so status stays consistent even under cancellation.
- `Stop()` → `Stop(ctx context.Context) error`: stops accepting jobs, fails queued jobs, then drains in-flight work bounded by `ctx`. If `ctx` elapses first, the root context is cancelled and `Stop` returns `ctx.Err()` without blocking further. A `nil` ctx uses a default 30s bound. `Stop` remains idempotent.
- `VectorStoreRuntime.Shutdown()` now bounds pipeline drain with a 30s deadline and logs if the drain doesn't complete, so a wedged backend cannot hang process shutdown.

This is intentionally scoped as a standalone, foundational slice. It changes the `Stop()` signature to `Stop(ctx context.Context) error` (a source-level change for direct callers, all of which are updated here). Making individual stages (embedder/backend) natively ctx-aware is a deliberate follow-up; `Stop`'s own contract (return within `ctx`) holds regardless.

## Test Plan

```
cd src/semantic-router && go test -race ./pkg/vectorstore ./pkg/routerruntime
```

New tests:
- **Bounded Stop** (`pipeline_lifecycle_test.go`): with a wedged embedder, `Stop(ctx)` returns `context.DeadlineExceeded` promptly (asserted `< 3s`) rather than hanging; `Stop` is idempotent and a second call returns `nil`.
- **Graceful drain + no goroutine leak** (`integration_test.go`): a full ingest-to-search run, then `Stop(context.Background())` returns `nil`, worker goroutines are reclaimed (goroutine count returns to baseline), and completed vectors remain searchable after shutdown.
- **Bounded stop under wedge, then clean unwind** (`integration_test.go`): `Stop(ctx)` honors the deadline while a worker is parked inside `Embed`, and after the embedder is released no goroutines remain.

This validates the two behaviors the change targets — cancellability and bounded shutdown — under both the graceful and the wedged path, with `-race` and goroutine-leak assertions.

## Test Result

- `go test -race ./pkg/vectorstore ./pkg/routerruntime` passes; race detector clean.
- Bounded-Stop and graceful-drain/leak tests ran repeatedly (`-count=1` x3) with no flakiness.
- Follow-up (not in this PR, tracked under #2474): make embedder/backend stages natively ctx-aware so a job wedged *inside* a single stage can be interrupted before the stage returns; streaming/bounded batches; status retention; transactional reconciliation.

---
- [x] PR title uses module-aligned prefix `[Router]`
- [x] Commits signed off with `git commit -s`
- [x] Purpose / Test Plan / Test Result reflect the actual scope, commands, and follow-ups

Closes #2602

Part of the umbrella #2474.

### DIFF
--- a/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
+++ b/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
@@ -2,6 +2,7 @@ package routerruntime
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"time"
@@ -29,8 +30,9 @@ type VectorStoreRuntime struct {
 
 // defaultDrainTimeout is the fallback shutdown drain bound used when a runtime
 // is constructed without a configured drain timeout, so Stop is never
-// unbounded.
-const defaultDrainTimeout = 30 * time.Second
+// unbounded. It sits below the 30s Kubernetes default grace period to leave
+// margin for closing the registry and backend after the drain.
+const defaultDrainTimeout = 25 * time.Second
 
 func NewVectorStoreRuntime(cfg *config.RouterConfig) (*VectorStoreRuntime, error) {
 	if cfg == nil {
@@ -133,6 +135,14 @@ func (r *VectorStoreRuntime) Shutdown() error {
 	if r == nil {
 		return nil
 	}
+
+	// Stop the pipeline first and only tear down shared dependencies once its
+	// workers are actually quiescent. A Stop timeout explicitly means the
+	// pipeline has NOT stopped: a worker may still be inside Embed,
+	// InsertChunks, or count persistence. Closing the registry or backend under
+	// it would risk a use-after-close and race Backend.Close against an active
+	// call. On timeout we therefore leave the dependencies open and let process
+	// exit reclaim them, surfacing the failure to the caller.
 	if r.Pipeline != nil {
 		// Fall back to a bounded default if drainTimeout was not configured
 		// (e.g. a hand-constructed runtime), so Stop is never unbounded.
@@ -143,18 +153,32 @@ func (r *VectorStoreRuntime) Shutdown() error {
 		ctx, cancel := context.WithTimeout(context.Background(), drain)
 		defer cancel()
 		if err := r.Pipeline.Stop(ctx); err != nil {
-			logging.Warnf("Ingestion pipeline did not drain within %s: %v", drain, err)
+			// Pipeline did not drain within the deadline. Do NOT close the
+			// registry/backend — workers may still be using them. Report the
+			// failure so the caller/operator knows shutdown was not clean.
+			logging.Warnf(
+				"Ingestion pipeline did not drain within %s; leaving vector-store "+
+					"dependencies open to avoid use-after-close by in-flight workers: %v",
+				drain, err,
+			)
+			return fmt.Errorf("vector store pipeline did not drain within %s: %w", drain, err)
 		}
 	}
+
+	// Workers are quiescent (or the pipeline was nil): safe to close deps.
+	var errs []error
 	if r.registryCloser != nil {
 		if err := r.registryCloser.Close(); err != nil {
 			logging.Warnf("Failed to close metadata registry: %v", err)
+			errs = append(errs, fmt.Errorf("close metadata registry: %w", err))
 		}
 	}
 	if r.Backend != nil {
-		return r.Backend.Close()
+		if err := r.Backend.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("close backend: %w", err))
+		}
 	}
-	return nil
+	return errors.Join(errs...)
 }
 
 func (r *VectorStoreRuntime) LogInitialized(component string, cfg *config.RouterConfig) {
--- a/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
+++ b/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
@@ -21,6 +22,11 @@ type VectorStoreRuntime struct {
 	Embedder       vectorstore.Embedder
 	registryCloser io.Closer
 }
+
+// vectorStoreShutdownTimeout bounds how long Shutdown waits for the ingestion
+// pipeline to drain in-flight jobs before cancelling them, so a wedged embedder
+// or backend cannot block process shutdown indefinitely.
+const vectorStoreShutdownTimeout = 30 * time.Second
 
 func NewVectorStoreRuntime(cfg *config.RouterConfig) (*VectorStoreRuntime, error) {
 	if cfg == nil {
@@ -123,7 +129,11 @@ func (r *VectorStoreRuntime) Shutdown() error {
 		return nil
 	}
 	if r.Pipeline != nil {
-		r.Pipeline.Stop()
+		ctx, cancel := context.WithTimeout(context.Background(), vectorStoreShutdownTimeout)
+		defer cancel()
+		if err := r.Pipeline.Stop(ctx); err != nil {
+			logging.Warnf("Ingestion pipeline did not drain within %s: %v", vectorStoreShutdownTimeout, err)
+		}
 	}
 	if r.registryCloser != nil {
 		if err := r.registryCloser.Close(); err != nil {
--- a/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
+++ b/src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
@@ -21,12 +21,16 @@ type VectorStoreRuntime struct {
 	Pipeline       *vectorstore.IngestionPipeline
 	Embedder       vectorstore.Embedder
 	registryCloser io.Closer
+	// drainTimeout bounds how long Shutdown waits for in-flight ingestion jobs
+	// to drain before cancelling them. Sourced from
+	// vector_store.ingestion_drain_timeout_seconds.
+	drainTimeout time.Duration
 }
 
-// vectorStoreShutdownTimeout bounds how long Shutdown waits for the ingestion
-// pipeline to drain in-flight jobs before cancelling them, so a wedged embedder
-// or backend cannot block process shutdown indefinitely.
-const vectorStoreShutdownTimeout = 30 * time.Second
+// defaultDrainTimeout is the fallback shutdown drain bound used when a runtime
+// is constructed without a configured drain timeout, so Stop is never
+// unbounded.
+const defaultDrainTimeout = 30 * time.Second
 
 func NewVectorStoreRuntime(cfg *config.RouterConfig) (*VectorStoreRuntime, error) {
 	if cfg == nil {
@@ -77,6 +81,7 @@ func NewVectorStoreRuntime(cfg *config.RouterConfig) (*VectorStoreRuntime, error
 		Pipeline:       pipeline,
 		Embedder:       embedder,
 		registryCloser: regCloser,
+		drainTimeout:   time.Duration(cfg.VectorStore.IngestionDrainTimeoutSeconds) * time.Second,
 	}, nil
 }
 
@@ -129,10 +134,16 @@ func (r *VectorStoreRuntime) Shutdown() error {
 		return nil
 	}
 	if r.Pipeline != nil {
-		ctx, cancel := context.WithTimeout(context.Background(), vectorStoreShutdownTimeout)
+		// Fall back to a bounded default if drainTimeout was not configured
+		// (e.g. a hand-constructed runtime), so Stop is never unbounded.
+		drain := r.drainTimeout
+		if drain <= 0 {
+			drain = defaultDrainTimeout
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), drain)
 		defer cancel()
 		if err := r.Pipeline.Stop(ctx); err != nil {
-			logging.Warnf("Ingestion pipeline did not drain within %s: %v", vectorStoreShutdownTimeout, err)
+			logging.Warnf("Ingestion pipeline did not drain within %s: %v", drain, err)
 		}
 	}
 	if r.registryCloser != nil {
--- a/src/semantic-router/pkg/vectorstore/pipeline.go
+++ b/src/semantic-router/pkg/vectorstore/pipeline.go
@@ -38,6 +38,11 @@ type IngestionJob struct {
 	ChunkingStrategy  *ChunkingStrategy
 }
 
+// defaultStopTimeout bounds a Stop call from a shutdown path that does not
+// supply its own deadline. It keeps process shutdown responsive even when a
+// backend or embedder is wedged.
+const defaultStopTimeout = 30 * time.Second
+
 // IngestionPipeline processes file attachment jobs asynchronously.
 // It reads files, extracts text, chunks, embeds, and stores the
 // resulting vectors in the backend.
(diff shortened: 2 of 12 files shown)


Peterren:
Good catch — you're right, 30s was a made-up constant. I've made it a config param.

To be precise about what it controls: it's **not** a per-file ingest budget (a single job can take much longer or shorter). It's the **graceful-shutdown drain window** — on shutdown the pipeline stops accepting new jobs, waits up to this long for currently-running jobs to finish, then cancels the rest so a wedged embedder/backend can't block process exit.

That means the right value is deployment-specific, bounded on both sides:
- **lower bound:** comfortably above typical single-job latency (file size × chunk count × per-chunk embed + backend insert), so normal in-flight work drains cleanly on rollout/restart;
- **upper bound:** below the platform's shutdown grace period — e.g. Kubernetes `terminationGracePeriodSeconds` (default 30s) — otherwise the platform SIGKILLs before the drain can run, defeating the purpose.

Now exposed as `vector_store.ingestion_drain_timeout_seconds` (default 30), sourced by the runtime, with a bounded fallback so `Stop` is never unbounded even if a runtime is hand-constructed without config. Added the key to the reference config, Helm values, and the RAG e2e profile. Pushed in the follow-up commit.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -97,32 +107,74 @@ func (p *IngestionPipeline) Start() {
...
 	p.running = false
 	p.mu.Unlock()
 
+	if ctx == nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), defaultStopTimeout)
+		defer cancel()
+	}
+
 	p.failQueuedJobs("pipeline_stopped", "ingestion pipeline stopped before processing job")
```
Xunzhuo: [P1] Apply the stop deadline before draining queued status updates. `failQueuedJobs` runs synchronously before the bounded wait and calls `failJob(context.Background())`; `failJob` then uses `WithoutCancel`. A stalled metadata registry can therefore keep `Stop(ctx)` here indefinitely after its deadline. Thread a bounded context through queue failure/persistence, avoid holding `lifecycleMu` around unbounded I/O, and add a blocking-registry plus queued-job test.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -97,32 +107,74 @@ func (p *IngestionPipeline) Start() {
...
+	// Wait for workers to drain, bounded by ctx. On timeout, cancel the root
+	// context so in-flight jobs abort at their next stage checkpoint, then
+	// return without waiting further — Stop must not block past ctx.
+	//
+	// Note: a job wedged *inside* a single stage (e.g. an embedder or backend
+	// call that ignores context) cannot be interrupted here; that worker unwinds
+	// only once the stage returns. Making individual stages ctx-aware is handled
+	// by the follow-up embedder/backend context work. Stop's own contract —
+	// returning within ctx — holds regardless.
+	done := make(chan struct{})
```
Xunzhuo: [P1] Do not reuse this WaitGroup/queue after a timed-out stop. On timeout this waiter and old workers remain live while `running` is already false, so an immediate `Start()` can `Add` to the same WaitGroup and reuse the same queue. That can race the waiter reaching zero, and an old worker can steal/fail a new-generation job with its canceled root context. Track a stopping generation and reject/block restart until joined, or give each generation its own WaitGroup and queue; cover timeout followed by immediate restart.

src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
```diff
@@ -123,7 +134,17 @@ func (r *VectorStoreRuntime) Shutdown() error {
...
-		r.Pipeline.Stop()
+		// Fall back to a bounded default if drainTimeout was not configured
+		// (e.g. a hand-constructed runtime), so Stop is never unbounded.
+		drain := r.drainTimeout
+		if drain <= 0 {
+			drain = defaultDrainTimeout
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), drain)
+		defer cancel()
+		if err := r.Pipeline.Stop(ctx); err != nil {
```
Xunzhuo, line 155: [P1] A timeout here explicitly means the pipeline has not stopped, yet `Shutdown` immediately closes its registry and backend below. A worker can still be inside `Embed`, `InsertChunks`, or count persistence; when it resumes it can use a closed dependency, and `Backend.Close` can race an active call (and is itself unbounded). Make timeout ownership explicit and close only after workers are quiescent/cancellable, or prove these components are safe under concurrent close and bound the close path.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -269,96 +321,132 @@ func (p *IngestionPipeline) DetachFile(ctx context.Context, vectorStoreID, vsfID
...
 	if err := p.backend.InsertChunks(ctx, job.VectorStoreID, embeddedChunks); err != nil {
-		p.failJob(job, "storage_error", fmt.Sprintf("failed to store chunks: %v", err))
+		p.failJob(ctx, job, "storage_error", fmt.Sprintf("failed to store chunks: %v", err))
 		return
 	}
 
 	// Mark as completed.
 	p.setFileStatus(job.VectorStoreFileID, "completed", nil)
-	_ = p.manager.UpdateFileCounts(context.Background(), job.VectorStoreID, func(fc *FileCounts) {
+	_ = p.manager.UpdateFileCounts(ctx, job.VectorStoreID, func(fc *FileCounts) {
```
Xunzhuo: [P1] Do not persist the completed transition with a possibly canceled job context and discard the error. If shutdown cancels the root after `InsertChunks` succeeds, PostgreSQL `SaveStore` can fail while file status is already `completed`; `Manager` has already mutated only its in-memory count, leaving durable counts stale after restart. Commit status/count coherently with a bounded cleanup context and surface or reconcile persistence failure.

Xunzhuo (changes requested):
A lifecycle root and deadline-aware `Stop(ctx)` are the right foundation, but the current implementation still has work outside the deadline and leaves ownership races after a timeout. Please address the four inline lifecycle blockers. Also keep the new config contract in sync with the dashboard/config surface, and either set the default below the Kubernetes termination grace period or configure a larger pod grace period; both are currently 30s, leaving no cleanup margin. CI is otherwise green, but `check-linked-issue` is failing and the branch is behind main. `make agent-report` and `git diff --check` passed; local race tests were blocked by unavailable native Candle dylibs.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -97,32 +107,74 @@ func (p *IngestionPipeline) Start() {
...
+// jobs to drain. If ctx is cancelled or its deadline elapses before draining
+// completes, the pipeline root context is cancelled so in-flight jobs abort at
+// their next checkpoint, and Stop returns ctx.Err() without blocking further.
+// A nil ctx is treated as a bounded shutdown with defaultStopTimeout.
+func (p *IngestionPipeline) Stop(ctx context.Context) error {
 	p.lifecycleMu.Lock()
 	defer p.lifecycleMu.Unlock()
 
 	p.mu.Lock()
 	if !p.running {
```
Xunzhuo: [P1] Preserve the stopping state after a timed-out `Stop`. `running` is cleared before workers exit, so a second `Stop(longerCtx)` returns `nil` here even though the previous generation may still be live. That falsely reports successful shutdown and removes the caller's ability to retry the join. Track `running`/`stopping`/`stopped` separately and have subsequent calls wait on the same generation completion (or return its outstanding state).

src/semantic-router/pkg/vectorstore/pipeline.go
Peterren: Fixed in d7b28da6. `failQueuedJobs` now takes the Stop `ctx` and drains the queue with no unbounded I/O under `lifecycleMu`: it always sets the in-memory failed status (no I/O) so nothing is left stuck in_progress, and bounds only the durable count persist by `ctx` (+ a per-write cap). A stalled registry can no longer push Stop past its deadline.

src/semantic-router/pkg/vectorstore/pipeline.go
Peterren: Fixed in d7b28da6. Each `Start` now creates a generation with its own queue, stopCh, WaitGroup and root context. A timed-out Stop leaves the old generation `stopping`; `Start` waits on that generation's join (without holding `lifecycleMu`) before creating a new one, so a restart never shares a WaitGroup/queue with live workers and an old worker can't steal a new-generation job. Added a timeout-then-immediate-restart test.

src/semantic-router/pkg/routerruntime/vectorstore_runtime.go
Peterren: Fixed in d7b28da6. `Shutdown` now closes the registry and backend only after `Stop` returns nil (workers quiescent). On a Stop timeout it leaves them open to avoid use-after-close by an in-flight worker, logs, and returns the error so the caller sees shutdown wasn't clean.

src/semantic-router/pkg/vectorstore/pipeline.go
Peterren: Fixed in d7b28da6. The completed path now commits status + count with a detached, bounded cleanup context (`WithoutCancel` + timeout), so a shutdown that cancels the root after `InsertChunks` succeeds can't drop the count update, and a persist failure is surfaced (logged) instead of discarded.

src/semantic-router/pkg/vectorstore/pipeline.go
Peterren: Fixed in d7b28da6 via a tri-state lifecycle (stopped/running/stopping). After a timed-out Stop the generation stays `stopping`; a second Stop waits on that same generation's join instead of returning nil, so it no longer falsely reports success and the caller keeps the ability to retry. Covered by a new test.

Peterren:
Thanks for the thorough review — all four lifecycle blockers are addressed in d7b28da6, replies inline.

Summary:
- Per-generation resources (queue/stopCh/WaitGroup/root ctx) + tri-state lifecycle (stopped/running/stopping); a timed-out Stop stays `stopping` and a second Stop retries the join instead of returning nil.
- `failQueuedJobs` and `AttachFile` no longer hold `lifecycleMu` across registry I/O; all cleanup persistence is bounded.
- `Shutdown` closes registry/backend only once workers are quiescent, else leaves them open and returns the error.
- Completed-path status/count committed coherently with a detached bounded context.

Also: synced the dashboard config surface with `ingestion_drain_timeout_seconds`, and lowered the default from 30s → 25s so it sits below the 30s Kubernetes default grace period with cleanup margin.

New regression tests cover timeout-then-restart, second-Stop honesty, and attach racing a restart; `go test -race ./pkg/vectorstore ./pkg/routerruntime ./pkg/config` is clean.

Peterren:
@Xunzhuo this is ready for another look. Since your review:

- **All 5 lifecycle P1s fixed** (`bbd5823a`, previously `d7b28da6`): bounded queue-drain under `Stop(ctx)`, per-generation queue/WaitGroup/root context, timeout-safe `Shutdown` ownership, coherent completed-status/count commit under a detached bounded cleanup context, and a tri-state `stopped`/`running`/`stopping` lifecycle so a second `Stop` after timeout no longer falsely reports success. Each is answered inline on the original threads with a timeout-then-restart and second-`Stop`-after-timeout test.
- **`check-linked-issue`** now passes — added `Closes #2602`, a dedicated slice issue carved from the #2474 acceptance list (lifecycle root + bounded/idempotent `Stop(ctx)`), rather than prematurely closing the umbrella.
- **Removed "backward-compatible"** from the description — `Stop()` → `Stop(ctx) error` is a source-level signature change (all direct callers are updated here).
- **Config↔surface sync** and **grace-period margin**: the drain default is **25s**, below the 30s pod termination grace period, and `ingestion_drain_timeout_seconds` is present across `config/config.yaml`, both helm `values.yaml` blocks, and the dashboard config support/catalog.

Rebased onto latest `main` (was 36 behind). CI is green except the local-only Candle race tests, which link against native dylibs unavailable locally — `test-and-build` exercises those in CI.

drivebyer:
Pushed 62022078 to fix a small classification gap this PR introduces.

`processJob` now passes the cancellable lifecycle root ctx into `backend.InsertChunks` (it was `context.Background()` before). So when a bounded `Stop` times out and cancels the root, a ctx-honoring backend like Milvus returns `context.Canceled` from an in-flight insert. That error was recorded as `storage_error`, which makes a normal shutdown look like a backend fault in status/metrics. The inter-stage checkpoints already record this case as `cancelled`, so the insert path was the one inconsistent spot.

The commit extracts a `storeChunks` helper that records the error as `cancelled` when the job ctx itself is cancelled and the error is `context.Canceled`/`DeadlineExceeded`. The `ctx.Err()` guard keeps a backend-internal timeout (job ctx still live) classified as a real `storage_error`.

Test: a new `ctxHonoringBackend` parks `InsertChunks` until its ctx is cancelled, modelling an insert interrupted by shutdown. The spec asserts a bounded `Stop` leaves the file `failed` with `LastError.Code == "cancelled"`, not `storage_error`. `go test -race ./pkg/vectorstore ./pkg/routerruntime` passes, and the new spec is stable across repeated runs.

drivebyer (dismissed):

FAUST-BENCHOU (dismissed):

Peterren:
Quick status: this now has approvals from @drivebyer and @FAUST-BENCHOU, CI is green (`test-and-build` + pre-commit), and I've rebased onto latest `main`. The child #2535 (embedder-context, stacked on this) is also approved and green.

@Xunzhuo whenever you get a chance — since all five of your lifecycle P1s are fixed and answered inline, could you re-review here so your earlier `CHANGES_REQUESTED` can be cleared? That's the only thing gating the merge of this slice (and the #2535 stack on top).

drivebyer (approved):

Peterren:
Closing this in favor of the work already landed via #2535.

Because this PR (A1) was the base of the stacked #2535 (A2, embedder context), merging #2535 carried A1's foundational lifecycle changes — the cancellable lifecycle **root context**, bounded **Stop(ctx)**, and the **embedChunks** helper — into main. Those acceptance bullets from #2474 are therefore already satisfied on main.

What remains unique to this branch is the tri-state generation lifecycle + shutdown use-after-close hardening. Rather than reshape this stale, now-mostly-redundant PR, I'll fold that hardening into the next focused slice and keep #2474's remaining acceptance items moving:

- bounded streaming parse/chunk/embed/insert batches (file/worker/queue/memory limits)
- status TTL + cardinality retention policy
- constructor / metadata-mutation saga reconciliation

Thanks for the reviews here — the lifecycle-safety feedback will carry forward into those slices.

PR #2526, opened 2026-07-14
[Router] feat: pluggable hallucination detector backend (endpoint)

### LINKED ISSUE #2415: feature: pluggable hallucination detector backend — OpenAI-compatible endpoint for generative span detectors
### Describe the feature

Add a `backend` option to `hallucination_mitigation.detector` selecting between the existing in-process Candle token classifier (`candle`, default, unchanged) and a new `endpoint` backend that calls a generative span detector behind any OpenAI-compatible server (e.g. vLLM):

```yaml
hallucination_mitigation:
  detector:
    backend: endpoint            # default: candle
    endpoint: http://127.0.0.1:8077/v1
    model_id: KRLabsOrg/lettucedect-v2-qwen-2b
    include_explanation: true
```

One call returns typed spans, category/subcategory from a 3×13 hallucination taxonomy plus an optional explanation, replacing the detector + NLI explainer two-pass with a single request (strict `json_schema` structured output, so parsing cannot fail). Results map onto the existing enhanced span structures; response-warning shaping, headers, actions, and metrics are unchanged. The detector becomes just another model served by vLLM.

If the endpoint is unavailable, detection fails gracefully (response passes through with the existing unverified warning path) rather than blocking traffic.

### Primary layer

plugin

### Why this layer?

Hallucination detection is already a plugin (`config/plugin/hallucination/`); this feature swaps the detection backend inside that plugin without touching signal extraction or decision logic.

The HaluGate blog identifies detector recall as the pipeline's bottleneck ("nearly half of hallucinations are missed"). Generative span detectors close that gap (see companion benchmark issue) but cannot run in-process via Candle. An endpoint backend makes detection quality a deployment choice: the fast in-process encoder by default, or a higher accuracy generative detector. It also fits the serving-stack philosophy: the detector itself is served by vLLM. The lettucedetect-qwen-2b model can be efficiently served.

### Additional context

The exact detection prompts (including the explanation variant) and the structured-output schema are documented on the model card: https://huggingface.co/KRLabsOrg/lettucedect-v2-qwen-2b.

pranavthakur0-0:
Closes #2415

## Purpose

  Introduces a new `endpoint` backend for the hallucination mitigation detector, allowing users to call generative span detectors behind any OpenAI-compatible server (e.g. vLLM) instead of being locked into the local in-process Candle classifier. It also includes critical hardening for LLM interactions (response limits, JSON `stream: false` enforcement, and multimodal array support).
  To support large generative models (like `lettucedect-v2-qwen-2b`) which cannot be run in-process via Candle. It replaces the legacy detector + NLI two-pass system with a single strict `json_schema` request.

## Test Plan

- **What commands, checks, or manual steps should reviewers use?**
  1. `make agent-lint`
  2. `go test ./pkg/config/... -run TestReferenceConfig`
  3. Deploy the router with `backend: endpoint` and configure an external vLLM server, verify that hallucinations are correctly detected and that Fusion Grounding (`reference: panel`) gracefully skips NLI without panicking.
- **Why is this validation sufficient for the affected module(s)?**
  Structural linting ensures code quality, while the config test ensures the new YAML fields don't break backward compatibility.

## Test Result

- **What were the actual results?**
  All local Go linters (`golangci-lint`), build checks, and config contract tests pass (`Exit code 0`).
- **Any follow-up risks, gaps, or blockers?**
  Unit tests for `EndpointHallucinationDetector` still need to be written (currently missing). Documentation updates in `website/docs` for the new `backend` fields are also needed in a follow-up PR.

---
<details>

</details>


Xunzhuo:
@pranavthakur0-0 can you sign your commits?

pranavthakur0-0:
> @pranavthakur0-0 can you sign your commits?

[@Xunzhuo](https://github.com/Xunzhuo) Done! 
thanks

src/semantic-router/pkg/classification/hallucination_detector_endpoint.go
```diff
@@ -0,0 +1,296 @@
...
+	req, err := http.NewRequestWithContext(context.Background(), "POST", d.endpoint+"/chat/completions", bytes.NewReader(bodyBytes))
+	if err != nil {
+		return d.fallbackResult(), nil
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := d.client.Do(req)
+	if err != nil {
+		logging.Warnf("Endpoint hallucination detection failed (fallback to safe): %v", err)
+		return d.fallbackResult(), nil
```
Xunzhuo: [P1] Preserve fail-open without converting detector failures into clean verdicts. Every transport, status, read, or parse failure in this method returns `fallbackResult(), nil`, and that fallback declares `hallucination_detected=false` with confidence `1.0`. The response filter therefore records `not_detected` instead of its existing `detection_error`/unverified path, creating false negatives. Please return an error for unavailable or malformed endpoint responses; the caller already fails open without blocking. Reserve the clean result for a successfully parsed empty span list.

src/semantic-router/pkg/classification/hallucination_detector_endpoint.go
```diff
@@ -0,0 +1,296 @@
...
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	return d.initialized
+}
+
+// IsNLIInitialized returns true because the endpoint natively provides explanation when requested
+func (d *EndpointHallucinationDetector) IsNLIInitialized() bool {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	return d.initialized
```
Xunzhuo: [P1] Please do not advertise endpoint initialization as NLI readiness. This returns true even when `include_explanation` is false. `HasHallucinationExplainer` explicitly represents the NLI explainer, while `/classify/nli` still obtains the Candle-only detector, so model/readiness APIs can report NLI ready although the NLI API fails. Keep endpoint NLI readiness false and expose generative explanation as a separate capability if needed.

src/semantic-router/pkg/classification/hallucination_detector_endpoint.go
```diff
@@ -0,0 +1,296 @@
...
+	for _, s := range parsed.HallucinatedSpans {
+		explanation := s.Explanation
+		if explanation == "" {
+			explanation = fmt.Sprintf("Unsupported %s (%s) detected", s.Subcategory, s.Category)
+		}
+
+		spans = append(spans, EnhancedHallucinationSpan{
+			Text:                    s.Text,
+			HallucinationConfidence: 1.0,
+			NLILabel:                NLIContradiction,
```
Xunzhuo: [P1] These endpoint taxonomy values are not NLI labels, but every span is synthesized as `NLIContradiction` with confidence `1.0` and critical severity. Downstream warnings and replay will present fabricated high-confidence NLI evidence. Please use backend-neutral endpoint metadata, or leave NLI fields as UNKNOWN/0. Also validate the returned category/subcategory locally, verify that `text` is an answer substring, and populate deterministic `Start`/`End` offsets.

src/semantic-router/pkg/classification/classifier_fact_hallucination_lifecycle.go
```diff
@@ -43,6 +43,22 @@ func (c *Classifier) initializeHallucinationDetector() error {
 		return nil
 	}
 
+	if c.Config.HallucinationMitigation.HallucinationModel.Backend == "endpoint" {
```
Xunzhuo: [P1] Backend selection needs semantic config validation before this branch. With exact equality here, an unknown value such as `endpont` silently falls through to Candle; endpoint requirements are only checked later during runtime construction. Please add a focused validator that defaults/normalizes Candle, accepts only `candle` or `endpoint`, requires and validates an absolute HTTP(S) endpoint for the endpoint backend, and keeps the Python CLI schema/migration and round-trip tests aligned.

src/semantic-router/pkg/extproc/req_filter_memory_rewrite.go
```diff
@@ -276,16 +280,42 @@ func ExtractConversationHistory(messagesJSON []byte) ([]ConversationMessage, err
 			continue
```
Xunzhuo, line 276: [P2] This multimodal conversation-history change and the response read-limit changes are unrelated to the hallucination endpoint backend. Please move them to a focused PR with their own unit tests. Keeping them here expands the regression surface and makes this endpoint change harder to review and roll back.

Xunzhuo (changes requested):
Thanks for reviving the endpoint-backed hallucination detector; the single structured call is a useful direction. I am requesting changes because the current failure and capability semantics can produce false clean verdicts and advertise NLI support that is not actually available.

I ran `make agent-ci-gate` locally against `d754ec2`; the existing suite passes, but this PR has no endpoint-specific coverage. Before merge, please also add focused tests for the request shape, clean and detected responses, timeout/network errors, non-2xx responses, oversized/malformed responses, taxonomy and span offsets; align the Python CLI config contract; update the required config docs; and add behavior-visible E2E coverage. The unrelated memory rewrite should move to a separate PR.

Once the runtime error semantics, NLI contract, config validation, and coverage are addressed, I will be happy to re-review.

src/semantic-router/pkg/extproc/req_filter_memory_rewrite.go
pranavthakur0-0: removed this for next pr

pranavthakur0-0:
hey @Xunzhuo, thanks for the thorough review and catching all this. just pushed an update that should address everything:

pulled out the memory rewrite changes, i'll put those in a separate pr so we keep this one focused.
fixed the fail-open logic so it actually returns an error on endpoint failure instead of a fake clean result.
stopped advertising NLI  readiness for the endpoint so we don't get false positives.
fixed the span metadata to just leave NLI  as unknown/0, and it now validates the taxonomy/offsets locally.
added the semantic config validation for the backend selection.
added focused unit tests for the endpoint covering the request shape, errors, and offsets.
aligned the python cli config contract and added tests for it.
updated the config docs.
added behavior-visible e2e coverage with a new hallucination profile

drivebyer (changes requested):
Please fix the `Pre-commit` check.

pranavthakur0-0:
sure  @drivebyer 


> Please fix the `Pre-commit` check.

src/semantic-router/pkg/classification/hallucination_detector_endpoint.go
```diff
@@ -0,0 +1,361 @@
...
+	var parsed struct {
+		HallucinatedSpans []struct {
+			Text        string `json:"text"`
+			Category    string `json:"category"`
+			Subcategory string `json:"subcategory"`
+			Explanation string `json:"explanation"`
+		} `json:"hallucinated_spans"`
+	}
+
+	if err := json.Unmarshal([]byte(openaiResp.Choices[0].Message.Content), &parsed); err != nil {
```
Xunzhuo, line 240: [P1] Treat missing or invalid span payloads as detector errors, not clean results. Unmarshalling into a plain slice cannot distinguish an explicit `[]` from a missing field or `null`, so `{}` and `{"hallucinated_spans": null}` both return zero spans. This loop also silently drops empty/non-substring spans, allowing an all-invalid detector response to become `hallucination_detected=false`—the false-clean path this PR otherwise avoids. Please require an explicitly present array and return an error for schema/taxonomy/span validation failures (or at minimum when every returned span is invalid), with tests for missing, null, empty-text, and all-invalid responses.

src/semantic-router/pkg/classification/hallucination_detector_endpoint.go
```diff
@@ -0,0 +1,361 @@
...
+
+		category := normalizeTaxonomyValue(s.Category, endpointCategories)
+		subcategory := normalizeTaxonomyValue(s.Subcategory, endpointSubcategories)
+
+		spans = append(spans, EnhancedHallucinationSpan{
+			Text:                    s.Text,
+			Start:                   start,
+			End:                     start + len(s.Text),
+			HallucinationConfidence: 1.0,
+			NLILabel:                0,
```
Xunzhuo: [P1] `0` is not an unknown NLI label in this type: `NLILabel` aliases `candle.NLILabel`, where `NLIEntailment == 0`. The serialized span therefore says `nli_label: 0` and `nli_label_str: "UNKNOWN"`; consumers that inspect the numeric field will treat endpoint output as entailment. Our earlier shorthand suggestion of “UNKNOWN/0” was imprecise. Please add a real unknown/backend-neutral representation (for example an explicit sentinel or optional NLI fields) and keep the numeric and string forms consistent.

e2e/testcases/hallucination_detection.go
```diff
@@ -0,0 +1,244 @@
...
+			"warning_rate": fmt.Sprintf("%.2f%%", detectionRate(warnedCount, len(testCases))),
+		})
+	}
+
+	printHallucinationResults(results, warnedCount)
+
+	// Fail only if the endpoint detection path never surfaced a warning: that
+	// means the pluggable endpoint backend did not run end-to-end at all. Like
+	// the other classifier e2e tests, per-case model behavior is tolerated.
+	if warnedCount == 0 {
```
Xunzhuo: [P1] This “at least one request passed” gate is explicitly below the repository's acceptance bar (`docs/agent/testing-strategy.md` and the local `e2e/testcases/AGENTS.md`). The mock detector deterministically flags every non-empty answer, so one warning out of five can hide broken routing, context propagation, or warning emission in the other cases. Please require all deterministic cases to warn (and include each case's error in the failure); only use a lower threshold if the behavior is genuinely probabilistic and the floor/rationale is documented.

e2e/pkg/cluster/kind.go
```diff
@@ -267,12 +267,19 @@ func (k *KindCluster) verifyNvidiaRuntime(ctx context.Context) error {
 func (k *KindCluster) getHostMountPath() (string, error) {
 	switch runtime.GOOS {
 	case "linux":
-		// On Linux, use /mnt as it's standard and typically has more space
-		return "/mnt", nil
+		// On Linux under Docker Desktop, /mnt and /tmp are often not shared by default.
+		// Use a directory in the user's home folder for better compatibility.
+		homeDir, _ := os.UserHomeDir()
```
Xunzhuo: [P2] Please split or revert this repository-wide Kind mount-policy change from the hallucination endpoint feature. It changes host paths for every E2E profile and is not needed to exercise this backend. It also discards `os.UserHomeDir()` errors in both this method and `createClusterConfig`; if home lookup fails, `filepath.Join("", ".kind-tmp", ...)` silently becomes a relative path. A focused follow-up can handle the error and validate the Linux/macOS behavior independently.

e2e/config/README.md
```diff
@@ -8,6 +8,14 @@ This directory holds smoke, demo, and harness manifests that the repository uses
 - response API and hallucination test configs
 - ONNX binding test configs
 
+`config.hallucination-endpoint.yaml` exercises the endpoint-backed hallucination
```
Xunzhuo: [P2] These instructions reference artifacts that are not in this PR or repository: there is no `e2e/config/config.hallucination-endpoint.yaml` and no `e2e/testing/mock-hallucination-endpoint.py`. The added config is `config.hallucination.yaml`, while the available mocks/profile are elsewhere. Please make the documented commands and paths match the actual runnable profile (or add the claimed artifacts), so a contributor can reproduce the test.

src/semantic-router/pkg/config/validator_hallucination.go
```diff
@@ -0,0 +1,71 @@
...
+	case HallucinationBackendCandle:
+		return nil
+	case HallucinationBackendEndpoint:
+		// endpoint backend requirements are validated below
+	default:
+		return fmt.Errorf("hallucination detector backend %q is not supported; use %q or %q",
+			cfg.Backend, HallucinationBackendCandle, HallucinationBackendEndpoint)
+	}
+
+	endpoint := strings.TrimSpace(cfg.Endpoint)
```
Xunzhuo: [P2] Validation uses a trimmed endpoint, but this is a pure check and the constructor later stores the raw `cfg.Endpoint` (`strings.TrimRight(cfg.Endpoint, "/")`). A value such as `" https://detector.example "` therefore passes validation and then produces an invalid request URL at runtime. Please either reject surrounding whitespace or normalize the value at the runtime seam before constructing requests, and add a test for this case.

Xunzhuo (changes requested):
Thanks for the substantial follow-up. I re-reviewed the latest head `05c2c68c`: transport/status/JSON failures now propagate correctly, endpoint NLI readiness is no longer advertised, semantic backend validation and focused tests were added, and the unrelated memory rewrite is gone.

I am still requesting changes because malformed-but-decodable endpoint output can become a false clean verdict, the numeric NLI value contradicts its `UNKNOWN` string, and the new E2E only requires one of five deterministic mock cases to pass. The inline comments give concrete fixes.

There are also merge-readiness gaps:
- The latest pre-commit job still fails on `validator_hallucination_test.go:26` (`endpont`); locally, `git diff --check origin/main...HEAD` also reports a blank line at EOF in `deploy/kubernetes/hallucination/gwapi-resources.yaml`.
- This changes the public config contract, but the config-package rules require the corresponding updates to `config/README.md`, `website/docs/installation/configuration.md`, and `website/docs/proposals/unified-config-contract-v0-3.md` (in addition to the tutorial).
- The new `hallucination` profile is manual-only in `e2e-profile-map.yaml`, so the standard green Kubernetes check does not prove this path ran. Please provide a successful affected-profile run after strengthening its acceptance condition.

Focused local checks on this head: `go test ./pkg/classification ./pkg/config` passed, and the new E2E packages compile with `go test ./profiles/hallucination ./testcases`.

adaamko:
Thanks for taking this on so quickly @pranavthakur0-0 

One critical detail as the model author: the system prompt must match the training prompt — these models were SFT-trained on a frozen prompt, and any drift silently degrades detection quality. Comparing this PR against our SFT training data:

- the prompt here merges the first two paragraphs (the training prompt has a blank line before "Quote each hallucinated span…") and reformats the final JSON-contract sentence.
- the user turn in training is User request: {q}\n\nExcerpt 1:\n{context}\n\nAnswer to verify:\n{answer} — there's no Context (trusted evidence): header in the training data

The exact frozen strings are documented on the model card (https://huggingface.co/KRLabsOrg/lettucedect-v2-qwen-2b, including the explanation variant), and the canonical source is https://github.com/KRLabsOrg/LettuceDetect/blob/main/lettucedetect/prompts/generative.py in our repo. I'd also suggest a tiny hash-guard test so future edits can't drift the prompt unnoticed.

cc @Xunzhuo

Xunzhuo:
Thanks @adaamko for the input.

And please address the comments @pranavthakur0-0 Thank you : )

pranavthakur0-0:
hey @Xunzhuo and @adaamko, thanks for the thorough re-review and catching the prompt drift. just pushed an update that should address everything:

fixed the span parsing so a missing/null/empty hallucinated_spans array throws a strict error instead of silently defaulting to a false clean.
swapped the NLI label to use an explicit sentinel instead of the 0 alias so we don't accidentally leak UNKNOWN/0 as an entailment.
tightened the e2e assertions so it now strictly requires all 5 deterministic mock cases to warn, instead of just needing one to pass.
reverted the repository-wide kind.go mount policy change so it doesn't impact other e2e profiles.
fixed the prompt drift so the strings exactly match the frozen SFT training data (added the blank lines back, restored the exact JSON sentence, and swapped back to Excerpt 1:).
updated the canonical config docs, fixed the missing artifact paths in the e2e README, and aligned the python cli contract.
fixed the validator to properly store the whitespace-trimmed endpoint URL so it doesn't break at runtime.
fixed the pre-commit test failure, dropped the blank EOF line in the yaml, and promoted the e2e profile so it runs automatically.
also ran the updated e2e suite locally against the new strict 5/5 requirement and everything passed clean with a 100% warning rate on the endpoint backend. attaching the final semantic-router-logs.txt from the run as proof.

[semantic-router-logs.txt](https://github.com/user-attachments/files/30056033/semantic-router-logs.txt)

src/semantic-router/pkg/classification/hallucination_detector_nli.go
```diff
@@ -19,6 +19,9 @@ const (
 	NLINeutral = candle.NLINeutral
 	// NLIContradiction means the premise contradicts the hypothesis.
 	NLIContradiction = candle.NLIContradiction
+	// NLIUnknown means no NLI judgment is available (e.g. the endpoint backend,
+	// which does not produce NLI labels).
+	NLIUnknown = NLILabel(4)
```
FAUST-BENCHOU: `candle.NLIUnknown` is now defined as `3`, but this alias assigns the same state a different value (`4`). Endpoint spans therefore serialize `nli_label: 4`, and calling `NLILabel.String()` on that value returns `ERROR` instead of `UNKNOWN`. Could this use `candle.NLIUnknown` like the other aliases?

pranavthakur0-0: thanks for the review ! Fixed in the latest commit to use the candle alias.

pranavthakur0-0:
Hey @Xunzhuo, @drivebyer,  @FAUST-BENCHOU 
I've made the requested changes. Could you please take another look and let me know if everything looks good?

FAUST-BENCHOU:
One small non-blocking follow-up: when the same returned span text appears multiple times in an answer, `strings.Index(answer, s.Text)` maps every occurrence to the first matching offset. The upstream LettuceDetect mapper selects the first non-overlapping match, which would preserve later repeated occurrences as well. This does not need to hold up this PR, but it would be worth aligning the offset behavior and adding a repeated-text test in a follow-up PR.

FAUST-BENCHOU (approved):

pranavthakur0-0:
> One small non-blocking follow-up: when the same returned span text appears multiple times in an answer, `strings.Index(answer, s.Text)` maps every occurrence to the first matching offset. The upstream LettuceDetect mapper selects the first non-overlapping match, which would preserve later repeated occurrences as well. This does not need to hold up this PR, but it would be worth aligning the offset behavior and adding a repeated-text test in a follow-up PR.

@FAUST-BENCHOU  Thanks for the review and approval! I'll align the offset handling with LettuceDetect and add a repeated-text test case in a follow-up PR.

adaamko:
hey @pranavthakur0-0 , just interested in the state of the PR, its now ready to be merged, right?

pranavthakur0-0:
> hey @pranavthakur0-0 , just interested in the state of the PR, its now ready to be merged, right?

hey @adaamko , yeah it's good to go, I have addressed all the issues and  review feedback. 
Should be ready to merge.

adaamko:
> > hey @pranavthakur0-0 , just interested in the state of the PR, its now ready to be merged, right?
> 
> hey @adaamko , yeah it's good to go, I have addressed all the issues and review feedback. Should be ready to merge.

I think the threads should be closed for the CI

pranavthakur0-0:
> > > hey @pranavthakur0-0 , just interested in the state of the PR, its now ready to be merged, right?
> > 
> > 
> > hey @adaamko , yeah it's good to go, I have addressed all the issues and review feedback. Should be ready to merge.
> 
> I think the threads should be closed for the CI

@adaamko resolved those. Thanks!

drivebyer (approved):
LGTM

pranavthakur0-0:
hey @Xunzhuo  can you please re-review the changes and let me know if everything looks good ?

Xunzhuo (approved):

Xunzhuo:
Thanks @adaamko and @pranavthakur0-0

PR #2524, opened 2026-07-14
[CI/Build] Auto-queue approved PRs with Mergify

Xunzhuo:
Closes #2525

## Purpose

- Add a repository-level Mergify configuration that automatically queues pull requests targeting `main` after they receive at least one approval.
- Keep the existing GitHub branch protection and required CI checks as the source of truth; Mergify injects those requirements into the merge queue.
- Replace the current manual checkbox/`@mergifyio queue` step with automatic queueing.
- Affected module: `CI/Build`.

## Test Plan

- Validate the file against Mergify's current configuration schema.
- Run YAML lint and the repository's agent validation, lint, CI, and feature gates for the changed file.

## Test Result

- `.venv-agent/bin/mergify config validate` — passed.
- `.venv-agent/bin/yamllint .mergify.yml` — passed.
- `make agent-validate` — passed.
- `make agent-lint CHANGED_FILES=".mergify.yml"` — passed.
- `make agent-ci-gate CHANGED_FILES=".mergify.yml"` — passed.
- `make agent-feature-gate ENV=cpu CHANGED_FILES=".mergify.yml"` — passed; no runtime or E2E commands matched this CI-only change.
- The configuration becomes active only after it lands on the default branch.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title uses module-aligned prefixes such as `[Router]`, `[CLI]`, `[Dashboard]`, `[Operator]`, `[Fleet-Sim]`, `[Bindings]`, `[Training]`, `[E2E]`, `[Docs]`, or `[CI/Build]`
- [x] If the PR spans multiple modules, the title includes all relevant prefixes
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>

Issue #2459, opened 2026-07-12
bug: main does not compile — pkg/extproc calls removed function streamingCachedPromptTokens


theohsiung:
## Summary

`main` currently does not compile — `pkg/extproc` fails typecheck, which breaks `build-router` and cascades to failing CI on **every** open PR (`build_pr/*`, `test-and-build`, `component-benchmarks`, `integration-test`).

## Error

```
pkg/extproc/processor_res_body_streaming_response_api.go:320:19: undefined: streamingCachedPromptTokens
```

## Repro

```bash
cd src/semantic-router && go build ./pkg/extproc/
```

## Root cause

A semantic conflict between two recently-merged PRs. They touch different files, so git merged them with no text conflict, but the resulting tree calls a function that no longer exists — and the combined state was never CI-tested (#2438 merged on stale checks without rebasing onto post-#2430 `main`):

- **#2430** (GPT-5.6 cache-aware pricing, merged 2026-07-11) renamed/expanded
  `streamingCachedPromptTokens(ctx, promptTokens) (int, bool)`
  into
  `streamingPromptTokenDetails(ctx, promptTokens) (cached int, cachedReported bool, cacheWrite int, cacheWriteReported bool)`.
- **#2438** (translate responses api streaming chunks, merged 2026-07-12, ~19h later) was developed against an older base that still had the old function, and added a **new call** to `streamingCachedPromptTokens` in the new file `processor_res_body_streaming_response_api.go`.

## Fix

Update the call in `responseAPIStreamUsage` to use the new `streamingPromptTokenDetails`, taking the `cached` / `cachedReported` return values (behavior-preserving). PR incoming.

## Impact

Blocks all PRs until fixed. This class of break could be avoided by requiring branches to be up to date with `main` before merge (or using a merge queue that re-tests the combined result).

PR #2529, opened 2026-07-14
[CI/Build] Add Mergify queue_rules to auto-update approved PRs

### LINKED ISSUE #2528: bug: Mergify auto_merge_conditions set without queue_rules, so approved PRs merge directly instead of via a merge queue
### Describe the bug

PR #2524 enabled automatic queueing via `merge_protections_settings.auto_merge_conditions`, but the repository's `.mergify.yml` defines **no `queue_rules`**. In Mergify, the merge queue is only enabled once at least one `queue_rules` entry exists. Without it, `auto_merge_conditions` falls back to a **direct merge**: approved PRs are merged straight into `main` without being updated against the latest `main` and without re-running CI.

Consequence: when several approved PRs target `main`, each merge leaves the others stale (their base is now behind). Contributors must manually click "Update branch", and a PR can merge on a CI result that was green against an *older* `main`.

### To Reproduce

1. Open PR A and PR B, both targeting `main`, both approved and CI-green against the same base commit.
2. Merge PR A (auto-queue fires, direct merge — no queue draft is created).
3. Observe PR B: its base is now behind `main`, it is **not** auto-updated, and CI is **not** re-run before it becomes mergeable.

### Expected behavior

Approved PRs should enter a real Mergify merge queue that updates each PR against the latest `main` and re-runs CI before merging. PRs already up-to-date with a green CI merge immediately; only PRs behind `main` are updated + re-checked.

### Affected layer

global level

### Additional context

Fix: add a default `queue_rules` entry so the merge queue is actually created. Branch protection + required CI remain the source of truth (Mergify injects them by default via `branch_protection_injection_mode: queue`).

WUKUNTAI-0211:
Closes #2528

## Purpose

PR #2524 enabled auto-queueing via `merge_protections_settings.auto_merge_conditions` but defined no `queue_rules`. In Mergify the merge queue is only created once a `queue_rules` entry exists, so today `auto_merge_conditions` merges approved PRs **directly** into `main` — without updating them against the latest `main` or re-running CI. Parallel approved PRs go stale after each merge and need manual branch updates.

This adds a default `queue_rules` entry so approved PRs enter a real merge queue that updates each PR against `main` and re-runs CI before merging. PRs already up-to-date with green CI merge immediately; only PRs behind `main` are updated + re-checked.

- `update_method: merge` — avoids force-pushing contributor fork branches (`rebase` fails on forks without maintainer edit access).
- `merge_method: squash` — one clean commit per PR on `main`.
- `batch_size: 1` — each PR updated and CI'd independently; simplest failure isolation.
- Branch protection + required CI remain the source of truth (Mergify injects them via the default `branch_protection_injection_mode: queue`).

Affected module: `CI/Build`.

## Test Plan

- Validate the file against Mergify's current configuration schema.
- YAML lint the changed file.

## Test Result

- `mergify config validate` — schema-valid `queue_rules` block.
- `yamllint .mergify.yml` — passed.
- Behavior becomes active only after the config lands on the default branch.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,4 +1,13 @@
 ---
+queue_rules:
+  - name: default
+    queue_conditions:
+      - base = main
+    merge_conditions: []
+    merge_method: squash
+    update_method: merge
+    batch_size: 1
+
 merge_protections_settings:
   auto_merge_conditions:
     - base = main


theohsiung:
This closes a real gap — one thing before it lands: `merge_conditions: []` means the queue updates each PR against `main` but never waits for the re-run CI to be green before merging.

I hit exactly this recently in #2459: two approved PRs, each green on its own, no textual conflict — they merged and broke `main`'s build (a helper was renamed in one, still called in the other). The queue's "update against main + re-run CI" is what catches that class of breakage, but *only* if the merge is gated on CI success. With empty `merge_conditions` — and no required status checks on `main` today (the `main` ruleset is currently disabled, so `branch_protection_injection_mode: queue` has nothing to inject) — nothing gates it, so the same breakage still slips through.

Heads-up on the fix: `test-and-build` is path-filtered (SKIPPED on non-core PRs), so requiring it directly would deadlock docs/config-only PRs. Either gate on an always-run aggregate check, or use something like:

```yaml
merge_conditions:
  - or:
```

Rest (squash / `update_method: merge` / `batch_size: 1`) looks good.

AayushSaini101 (approved):
thanks

PR #2458, opened 2026-07-12
[Refactor] Modularize native Router Model bindings and runtime contracts

Paramveersingh-S:
## Overview
This PR addresses Issue #2396 by introducing a binding-neutral native runtime contract for all router-owned model assets, completely decoupling the Go semantic router from the legacy CGO/Rust sprawl in `candle-binding`. It establishes the foundation for extensible backends (Candle, ONNX, OpenVINO) without hard-coding provider logic in the critical path.

## Key Changes
- **Native Runtime Contract (`pkg/modelruntime/native`)**
  - Introduced `BackendAdapter`, `ModelHandle`, `LoadRequest`, and `InferenceRequest` to establish a stable, backend-neutral API.
  - Locked down capability, family, modality, and artifact taxonomy in `taxonomy.go` to remove duplicated strings across Rust/Go.
- **Router Call Site Refactoring**
  - **`router_runtime.go`**: Removed all direct legacy `candle_binding.InitModel(...)` calls; now initializes models natively through `native.LoadModel(ctx, backend, request)`.
  - **`classification/unified_classifier.go`**: Removed manual CGO initialization, migrating dynamic labels and thresholding logic to standard `LoadRequest` parameters via the registry.
  - **`embedding/provider.go`**: The unified `NewProvider` now first queries the native `AdapterRegistry` before falling back to remote OpenAI configurations.
- **Centralized Rust Registry (`candle-binding`)**
  - Migrated `state_manager.rs` from sprawling `RwLock<Option<Arc<...>>>` singletons to a centralized `HashMap`-based `ModelRegistry` in `registry.rs` to streamline dynamic model swaps.
- **Strict Configuration Validation**
  - **`config/validator_native.go`**: Added `validateNativeBackendContracts` to explicitly catch invalid backend/capability configurations and reject them before serving API traffic.
- **API Model Info Endpoints**
  - **`apiserver/route_model_info_embeddings.go`**: Refactored the `/models` endpoints to programmatically iterate over `native.Registry.List()` and expose capabilities, matching the normalized discovery schema.

## Related Issues
Fixes #2396 - Modularize native router model bindings and runtime capability contracts.

PR #2530, opened 2026-07-14
[Refactor] Introduce the native model runtime contract

### LINKED ISSUE #2396: [Epic] Unify native Router Model adapters and capability discovery

Paramveersingh-S:
## Overview
This is **Phase 1** of splitting PR #2458 into smaller, staged, and behavior-preserving pull requests to address [Issue #2396](https://github.com/vllm-project/semantic-router/issues/2396). 

As requested in previous reviews, this PR focuses **strictly on introducing the backend-neutral native runtime contracts and taxonomy**. It does *not* touch any existing call sites or Rust logic, ensuring that the legacy `candle-binding` singletons and existing runtime behaviors are 100% preserved.

## Key Additions
- **Taxonomy Lockdown (`taxonomy.go`)**: Established strict `Backend`, `Capability`, `Family`, `ArtifactFormat`, and `Modality` constants to eliminate string duplication and prevent future taxonomy drift.
- **Native Contracts (`adapter.go`, `request.go`)**: Defined the `BackendAdapter` interface that future backend adapters (Candle, ONNX, OpenVINO) will implement. Introduced `CapabilitySet`, `LoadRequest`, and `InferenceRequest` to formalize hardware and feature support.
- **Thread-Safe Registry (`registry.go`)**: Introduced the central `native.Registry` component to track loaded model bindings securely.
- **ABI Tests (`ffi_layout_test.go`)**: Added stub test files for catching layout/taxonomy drifts at the CGO boundary.

This PR only *adds* the new foundational types to `src/semantic-router/pkg/modelruntime/native`. Because no existing `init()` side effects, Rust code, or CGO calls were modified, this guarantees that we bypass the Rust `LazyLock` and Go NUL-byte CI failures encountered previously. 

Part of #2396


src/semantic-router/pkg/modelruntime/native/adapter.go
```diff
@@ -0,0 +1,32 @@
...
+	IsLoaded          bool
+	MaxSequenceLength int
+	DefaultDimension  int
+}
+
+// BackendAdapter defines the neutral contract that all backends must implement.
+type BackendAdapter interface {
+	Name() Backend
+	Capabilities() CapabilitySet
+	LoadModel(ctx context.Context, req LoadRequest) (ModelHandle, error)
```
Xunzhuo, line 41: [P1] The neutral contract has no inference operation. `InferenceRequest` is declared, but neither `BackendAdapter` nor `ModelHandle` can execute it and there is no response type. A migrated caller would still have to import a concrete binding after `LoadModel`, defeating the issue's core goal that request paths ask for a capability/model ref rather than a backend package. Please put inference on the handle or adapter (with a handle argument), and use typed capability-specific request/response shapes instead of an unconsumed `map[string]interface{}` bag.

src/semantic-router/pkg/modelruntime/native/adapter.go
```diff
@@ -0,0 +1,32 @@
...
+type ModelInfo struct {
+	Backend           Backend
+	Capability        Capability
+	Family            Family
+	ModelName         string
+	ModelPath         string
+	IsLoaded          bool
+	MaxSequenceLength int
+	DefaultDimension  int
+}
```
Xunzhuo, line 35: [P1] This is not yet the normalized model-info contract required by #2396 or consumed by Phase 4. It omits artifact format, modality, dimensions/layers as requested-vs-runtime metadata, provider/device, version, runtime feature flags/unsupported reasons, and registry metadata; it also allows only one capability. If this foundational type lands now, the API migration will either lose existing information or require an immediate breaking expansion. Please define the complete normalized discovery shape in Phase 1 and add round-trip/discovery tests before call sites depend on it.

src/semantic-router/pkg/modelruntime/native/registry.go
```diff
@@ -0,0 +1,47 @@
...
+	adapter, ok := r.adapters[name]
+	return adapter, ok
+}
+
+// List returns all registered adapters.
+func (r *AdapterRegistry) List() []BackendAdapter {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	list := make([]BackendAdapter, 0, len(r.adapters))
+	for _, adapter := range r.adapters {
```
Xunzhuo: [P1] Iterating a Go map makes `List()` nondeterministic. Phase 4 emits `/models` in exactly this order, so identical processes can return different response ordering and order-sensitive tests/clients will flake. Return adapters in a stable backend order (or sort the collected keys) and assert it with multiple registrations.

src/semantic-router/pkg/modelruntime/native/registry.go
```diff
@@ -0,0 +1,47 @@
...
+	return &AdapterRegistry{
+		adapters: make(map[Backend]BackendAdapter),
+	}
+}
+
+// Register adds a backend adapter to the registry.
+func (r *AdapterRegistry) Register(adapter BackendAdapter) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.adapters[adapter.Name()] = adapter
```
Xunzhuo: [P2] Registration has no invalid/duplicate contract: a typed-nil adapter can panic at `Name()`, and a second adapter silently replaces the first backend. Since registration becomes the composition boundary, please return an error for nil/empty names and duplicate backends (or document and test an intentional replacement API separately). The current concurrency smoke test deliberately avoids the nil case but does not validate registry composition semantics.

src/semantic-router/pkg/modelruntime/native/ffi_layout_test.go
```diff
@@ -0,0 +1,50 @@
+package native
+
+import (
+	"context"
+	"testing"
+)
+
+// In a real scenario, this would import C and use unsafe.Sizeof to verify structs against C headers.
```
Xunzhuo: [P2] This file is explicitly a placeholder and does not test FFI layout or discriminants, so it cannot be presented as the ABI drift coverage requested by #2396. Either add real Go/Rust/C boundary checks (for example generated discriminants plus `unsafe.Sizeof`/offset assertions against the exported C structs) or rename this to a registry/taxonomy unit test and leave ABI coverage to the Rust-binding phase. A test that only compares a Go string constant to a literal will not catch binding drift.

Xunzhuo (changes requested):
Thanks for splitting the original refactor. Keeping Phase 1 additive is the right boundary, and `go test -race ./pkg/modelruntime/native` passes on this head. I am requesting changes because the foundational contract still cannot support a backend-neutral request path, and downstream phases already expose the gaps.

Please fix this phase before rebasing the later phases:
- Use a module-prefixed title such as `[Router] ...`.
- Replace `Fixes #2396` with `Part of #2396`; Phase 1 does not satisfy the roadmap issue and must not close it when merged.
- Make #2532 base on this branch, #2533 on #2532, and #2534 on #2533 (or land/rebase them sequentially). Today all four target `main` and contain independently recreated Phase 1 commits, so their diffs overlap and cannot be merged as a clean stack.

The inline comments cover the contract, normalized model-info shape, registry semantics, and tests. Once those are addressed, this additive first slice can be reviewed independently of the backend migrations.

Paramveersingh-S:
Thanks for the detailed review! I've updated the PR title and description as requested. I've also pushed a new commit that addresses all your points:

- **[P1] Inference Contract**: Replaced the map bag with strongly-typed request/response interfaces (like `EmbeddingRequest`) and added the `Inference` method to `BackendAdapter`.
- **[P1] Model-Info Shape**: Expanded `ModelInfo` into the complete discovery shape required by #2396 (added dimensions/layers, format, modality, feature flags, etc.).
- **[P1] Registry Determinism**: `List()` now returns adapters in alphabetical order by backend name for stable `/models` responses.
- **[P2] Registration Rules**: `Register()` now strictly returns an `error` for nil adapters, empty names, or duplicate backends.
- **[P2] ABI Tests**: Removed the placeholder `ffi_layout_test.go` and replaced it with `registry_test.go` to test taxonomy, duplicate/nil errors, and ordering. Actual FFI boundary tests will be added in the Rust-binding phases.

The base contracts are fully updated and tested. I'll make sure to sequentially rebase the other phases

Xunzhuo (changes requested):
Thanks for the update. The latest head addresses most of the original feedback: inference is now part of the adapter contract, ModelInfo is expanded, list ordering is deterministic, registration rejects ordinary nil/empty/duplicate inputs, and the placeholder ABI test is gone.

Changes are still required before this contract becomes the base of the remaining phases:

1. EmbeddingRequest must carry the per-inference dimension/layer options supported by current Candle APIs; load-time fields are not equivalent.
2. Info() must return an error so backend discovery failures do not silently become an empty model list.
3. Register must reject typed-nil adapters, with a regression test.
4. Fix the failing gofumpt check.

Please also use a `[Router]` title, change `Fixes #2396` to `Part of #2396`, remove the obsolete `ffi_layout_test.go` description, and make the Phase 2–4 PR bases reflect the actual stack before further review.

Xunzhuo (commented):
Thanks for the update — the earlier request/response, discovery-error, typed-nil, and ordering gaps are addressed. One blocker remains: `BackendAdapter` now has six methods, and the required Go quality check rejects the interface with `interfacebloat`. Please split the contract into smaller consumer-facing interfaces, for example discovery/capability versus lifecycle/inference, then rerun the checks.

FAUST-BENCHOU (commented):
Thanks for introducing the native model runtime contract layer. Keeping this phase limited to interfaces and registry primitives makes the migration easier to review and bisect.

One design consideration: the global registry singleton may become harder to manage when multiple runtime instances or isolated model lifecycles are needed. It may be worth keeping the registry ownership/lifecycle abstraction open for future use cases.

Paramveersingh-S:
what are reviews on this PR as the tests are passing. are their any other changes required ?

FAUST-BENCHOU (approved):
I dont have other comments now.Thanks for your contribution

Xunzhuo:
Closing as superseded by the typed task-binding and generation-owned runtime now on main, including #3769. Remaining native backend and capability work stays tracked in #2396.

PR #2535, opened 2026-07-14
[Router] vectorstore: thread context into Embedder interface

### LINKED ISSUE #2603: vectorstore: thread context into the Embedder interface (embedding-stage cancellation)
## Summary

Follow-on slice of #2474, building directly on the lifecycle root + bounded `Stop(ctx)` from #2602: thread a `context.Context` through the vector-store `Embedder` interface so cancellation reaches *into* the embedding stage.

Covers this acceptance bullet of #2474:
- *Derive job, embedding, backend, and count-update contexts from a lifecycle root* — specifically the **embedding** leg.

## Current behavior

After #2602 the pipeline derives a cancellable lifecycle context and checks `ctx.Err()` between stages, but `Embedder.Embed(text)` takes no context. A job wedged *inside* an embed call therefore cannot be interrupted by a cancelled lifecycle — the per-stage checkpoint only fires between chunks, not during a single embed.

## Expected behavior

- `Embedder.Embed(text)` → `Embed(ctx context.Context, text)`; a cancelled lifecycle/request aborts embedding at the next checkpoint for any embedder that honors the context.
- `CandleEmbedder.Embed` checks `ctx.Err()` before the (synchronous, non-cancellable) FFI call; interrupting a call already in flight remains a deliberate follow-up.
- The pipeline passes the job's lifecycle-derived context into `Embed`; query-side callers pass their existing request/trace context.
- A cooperative embedder that returns `context.Canceled`/`DeadlineExceeded` on shutdown is classified with the `cancelled` status code, not `embedding_error`, so status/metrics distinguish intentional termination from a backend fault.

## Validation

- A ctx-aware embedder parked inside `Embed`: a bounded `Stop(ctx)` elapses and cancels the root; the parked embed then unwinds on its own (no explicit release), no worker goroutines are leaked, and the interrupted job is recorded `failed` with `LastError.Code == "cancelled"`.

## Related

- Parent umbrella: #2474
- Depends on: #2602 (lifecycle root + bounded `Stop(ctx)`)

Peterren:
> **Stacked on #2517 (A1).** This PR targets `main`, so until #2517 merges the diff also shows the A1 commits (lifecycle root context + bounded `Stop(ctx)`). Please review #2517 first; the A2-only change is the final commit, `refactor(vectorstore): thread context into Embedder interface`.

## Purpose

- Thread a `context.Context` through the vector-store `Embedder` interface: `Embed(text)` becomes `Embed(ctx, text)`.
- Today the ingestion pipeline derives a cancellable lifecycle context (added in #2517), but the embedding stage still takes no context, so a job wedged *inside* an embed cannot be interrupted — that gap was explicitly called out as the follow-up in the bounded `Stop(ctx)` work. This PR closes it for cooperative embedders.
- Module: `Router` (with `E2E`-style tests inside the `vectorstore` package).

### What changed

- `Embedder.Embed(text)` → `Embed(ctx context.Context, text)`. The interface doc notes that a cancelled lifecycle/request aborts embedding at the next checkpoint.
- `CandleEmbedder.Embed` checks `ctx.Err()` before the FFI call. The underlying `candle_binding` calls are synchronous and not themselves cancellable; interrupting a call already in flight remains a deliberate follow-up.
- The pipeline passes the job's lifecycle-derived context into `Embed` (the per-chunk `ctx.Err()` checkpoint from A1 is unchanged), so cancelling the root context now propagates *into* an in-flight embed for any embedder that honors it.
- The two query-side callers pass their existing request context: `route_vectorstore_search.go` uses `r.Context()`, and the extproc RAG filter uses its trace context.
- Test embedders and mocks updated to the new signature.

This is a standalone, **source-incompatible** Go interface change: `Embedder.Embed` gains a `ctx` parameter, so every implementation and call site must be updated (all in-tree ones are, in this PR). It does not alter routing, ranking, or ingestion semantics; it only makes cancellation reach the embedding stage.

## Test Plan

```
cd src/semantic-router && go test -race ./pkg/vectorstore ./pkg/routerruntime
```

New test (`pipeline_embedder_ctx_test.go`): a **ctx-aware** embedder that blocks inside `Embed` until either released or its context is cancelled. With a job parked inside the embed, a bounded `Stop(ctx)` elapses and cancels the lifecycle root; the test asserts the parked embed then unwinds **on its own** (no explicit release), no worker goroutines are leaked, and the interrupted job is recorded as `failed` with `LastError.Code == "cancelled"` (a context cancellation is classified as an intentional termination, not an `embedding_error`). The goroutine baseline is captured **before** the fixture starts the worker, so the leak assertion is a real barrier rather than a tautology, and the failed-status read is an `Eventually` because `failJob` runs as the worker unwinds. A ctx-ignoring embedder would stay parked — that contrast is the behavior this PR adds.

## Test Result

- `go test -race ./pkg/vectorstore ./pkg/routerruntime` passes; race detector clean. `pkg/apiserver` and `pkg/extproc` tests also pass with the updated signature.
- `gofmt` and `golangci-lint` (repo config) are clean on the changed files.
- Follow-up (not in this PR, tracked under #2474): make the `candle_binding` embed and backend insert calls natively cancellable so a job wedged *inside* a single stage can be interrupted before the stage returns; streaming/bounded batches (B2) build on this ctx-aware interface.

---
<details>

- [x] PR title uses module-aligned prefix `[Router]`

</details>

Closes #2603

Part of the umbrella #2474 (stacked on #2602 / #2517).


src/semantic-router/pkg/vectorstore/pipeline_embedder_ctx_test.go
```diff
@@ -0,0 +1,124 @@
...
+		// its deadline error and cancels the lifecycle root context.
+		ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+		defer cancel()
+		Expect(f.pipeline.Stop(ctx)).To(MatchError(context.DeadlineExceeded))
+
+		// The key A2 behavior: because the embedder now receives the (now
+		// cancelled) lifecycle context, the parked embed returns on its own —
+		// no explicit release is needed — so the worker unwinds and no
+		// goroutines are leaked. A ctx-ignoring embedder would stay parked here.
+		Eventually(runtime.NumGoroutine, 5*time.Second, 50*time.Millisecond).
```
Xunzhuo, line 133: This condition does not wait for the pipeline worker to exit. `f` starts the worker in `BeforeEach`, then `baseline` is sampled here, so `runtime.NumGoroutine() <= baseline+1` is already true while the worker is still unwinding. The next synchronous status read consequently races with `failJob`; both CI gates reproduced `in_progress` at line 122. Capture the baseline before constructing/starting the fixture (or wait on an explicit worker-done signal), and use `Eventually` for the failed status/count update. The cleanup release should remain only a timeout safety net, not something that can make the assertion pass.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -269,96 +323,132 @@ func (p *IngestionPipeline) DetachFile(ctx context.Context, vectorStoreID, vsfID
...
+// and returns ok=false; the caller should stop processing.
+func (p *IngestionPipeline) embedChunks(ctx context.Context, job IngestionJob, filename string, chunks []TextChunk) ([]EmbeddedChunk, bool) {
+	embeddedChunks := make([]EmbeddedChunk, 0, len(chunks))
+	for _, chunk := range chunks {
+		if err := ctx.Err(); err != nil {
+			p.failJob(ctx, job, "cancelled", fmt.Sprintf("ingestion cancelled before embedding chunk %d", chunk.ChunkIndex))
+			return nil, false
+		}
+
+		embedding, err := p.embedder.Embed(ctx, chunk.Content)
```
Xunzhuo, line 424: A cooperative embedder now returns `context.Canceled` when shutdown cancels the lifecycle root, but this path records it as `embedding_error`. That misclassifies an intentional shutdown as an embedding/backend fault and makes status/metrics misleading. Please check `errors.Is(err, context.Canceled)` / `context.DeadlineExceeded` (or `ctx.Err()`) and use the existing `cancelled` code for context termination; add an assertion on `LastError.Code` in the new test.

Xunzhuo (changes requested):
The A2 direction is sound: the incremental diff threads the lifecycle/request context through every vector-store `Embedder` call site, and it accurately documents that Candle FFI is only pre-call cancellable. I am requesting changes for two correctness/validation issues below. The new test currently fails in both `test-and-build` and the pre-commit fast gate, and cancellation is recorded as an embedding failure. Also, please remove “backward-compatible” from the description (this is a source-incompatible Go interface change) and satisfy `check-linked-issue` with a dedicated closing issue link; `Part of #2474` does not populate GitHub closing references, while closing the umbrella issue from this partial slice would be premature. Keep this PR stacked on #2517 or rebase it after #2517 lands.

Peterren:
@Xunzhuo addressed both review points and rebased on latest `main`:

1. **Test race fixed** (`d2a073ab`): the goroutine baseline is now captured in `BeforeEach` **before** the fixture starts the worker, so `NumGoroutine() <= baseline+1` is a real leak barrier rather than trivially true while the worker unwinds. The failed-status read is now an `Eventually` (since `failJob` runs concurrently as the worker unwinds), so it no longer races and observes `in_progress`. The cleanup release stays only as a timeout safety net.
2. **Cancellation classification fixed**: `embedChunks` now checks `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` and records the job with the existing `cancelled` code instead of `embedding_error`. The new test asserts `status.LastError.Code == "cancelled"`.
3. **Description**: removed "backward-compatible" — this is a source-incompatible Go interface change (`Embed` gains a `ctx` param; all in-tree implementations/callers are updated here).
4. **`check-linked-issue`** now passes — added `Closes #2603`, a dedicated slice issue for the embedding-stage cancellation, rather than `Part of #2474` (which doesn't populate closing references) or prematurely closing the umbrella.

Still stacked on #2517 — the A2-only change is the final commit, `vectorstore: classify embed cancellation as cancelled; fix leak-test baseline race` sitting on top of `refactor(vectorstore): thread context into Embedder interface`. Happy to rebase onto `main` once #2517 lands.

Xunzhuo (approved):

Xunzhuo (approved):

Xunzhuo (approved):

PR #2560, opened 2026-07-15
[Bench] standalone hallucination detector benchmark: generative detectors (LettuceDetect v2) vs the current pipeline

### LINKED ISSUE #2414: feature: benchmark generative hallucination detectors (LettuceDetect v2) against the current pipeline
### Describe the feature

Extend `bench/hallucination/` with a standalone benchmark that evaluates hallucination detectors directly — no router/Envoy stack needed — so candidate detector models can be compared on equal footing:

- support token-classifier encoders *and* generative span detectors served behind an OpenAI-compatible endpoint (e.g. vLLM), starting with the new LettuceDetect v2 family (`KRLabsOrg/lettucedect-v2-qwen-2b`, `lettucedect-v2-mmbert-base`)
- support the current pipeline's public models (halugate-sentinel + modernbert-base-32k-haldetect-combined) under the same protocol, with a threshold/min-span sweep for fair operating points
- report example-level P/R/F1/accuracy, char-level span F1 (when gold spans exist), taxonomy category/subcategory accuracy, and latency (avg/p50/p99)
- reuse the existing `bench/hallucination/datasets.py` loaders (HaluEval, FinanceBench, custom JSONL)

### Primary layer

unsure

### Why this layer?

Evaluation tooling under `bench/`, outside the signal/plugin/global runtime layers; it measures the models that power the `hallucination` plugin.

The HaluGate blog notes detector recall is the pipeline's main limitation. New generative span detectors (LettuceDetect v2, check out our paper arXiv:2607.00895) target exactly that, but the existing `bench/hallucination/evaluate.py` can only measure detection through the router's E2E path; it cannot evaluate a detector model directly, nor one served behind its own OpenAI-compatible endpoint.

Preliminary numbers suggest the comparison is worth making rigorously: on HaluEval (1000 samples, example-level) lettucedect-v2-qwen-2b served by vLLM reaches 0.86 F1 at ~113ms p50, vs 0.65 F1 for the current public detector+sentinel at its best swept threshold (perfect precision, recall-bound, consistent with the blog's analysis). A shared benchmark makes these numbers reproducible by anyone.

### Additional context

**Models**: https://huggingface.co/collections/KRLabsOrg/lettucedetect-v2 
**Paper**: https://arxiv.org/abs/2607.00895
**LettuceDetect v1** is already the documented example detector in `bench/hallucination/README.md`.

adaamko:
Closes #2414

## Purpose

  Adds a standalone detector benchmark to `bench/hallucination/` — no router/Envoy stack needed:
  - `evaluate_detectors.py`: evaluates any detector supported by the `lettucedetect` library — token-classifier encoders or generative span detectors served behind an OpenAI-compatible endpoint (`--base-url`, e.g. vLLM). Reports example-level P/R/F1/accuracy, char-level span F1 (when gold spans exist), taxonomy accuracy, and latency (avg/p50/p99).
  - `evaluate_halugate.py`: evaluates the current pipeline's public models (halugate-sentinel + modernbert-base-32k-haldetect-combined) under the paper serialization (arXiv:2603.23508), with an optional threshold/min-span sweep.
  - README: documents the workflow and full-dataset results (all 10,000 HaluEval QA samples).
  The existing `evaluate.py` measures detection only through the router's E2E path and cannot evaluate a detector model directly, nor one served behind its own endpoint. This benchmark lets candidate detectors (e.g. the LettuceDetect v2 family motivating #2415 / #2526) be compared on equal footing, reproducibly.
- **Which module(s) does this affect?** `Bench`

## Test Plan

- `python3 -m bench.hallucination.evaluate_detectors --self-test` (span-metric unit checks)
- Full runs on an A40 GPU box:
  - `python3 -m bench.hallucination.evaluate_detectors --detector llm:KRLabsOrg/lettucedect-v2-qwen-2b --base-url http://localhost:8077/v1 --dataset halueval --max-samples 10000` (detector served by `vllm serve KRLabsOrg/lettucedect-v2-qwen-2b`)
  - `python3 -m bench.hallucination.evaluate_halugate --max-samples 10000 --sweep`
- New files only (plus a README section) — no router code touched, so no router/E2E impact.

## Test Result

- Self-test passes.
- Full 10,000-sample HaluEval results (in the README): lettucedect-v2-qwen-2b (2B, vLLM-served) 0.855 F1 / 0.869 acc @ ~116ms p50; haldetect-combined (149M) 0.669 F1 at its best swept operating point / 0.566 F1 at production settings @ ~13ms p50. Precision/recall breakdowns, oracle-threshold disclosure, and sentinel-gate behavior documented alongside.
- Both training-time serializations were tested for the baseline (<2 F1 points difference).

---
<details>

</details>


bench/hallucination/evaluate_detectors.py
```diff
@@ -0,0 +1,227 @@
...
+        try:
+            preds = detector.predict(
+                context=[sample.context],
+                question=sample.question,
+                answer=answer,
+                output_format="spans",
+            )
+        except Exception as e:
+            print(f"  {sample.id}: predict failed: {e}")
+            continue
```
FAUST-BENCHOU, line 119: The benchmark is meant to compare detectors on equal footing, but a failed prediction disappears from both the metrics and the reported sample count here. If one backend times out on a few difficult examples, its F1 will be calculated on an easier subset while the final comparison still looks complete. Could we either fail the run by default, or record failures and make sure every detector is compared over the same sample IDs?

bench/hallucination/evaluate_halugate.py
```diff
@@ -0,0 +1,127 @@
...
+                item["question"], return_tensors="pt", truncation=True, max_length=512
+            ).to(device)
+            with torch.no_grad():
+                label = int(s_model(**enc).logits[0].argmax())
+            needs_check = s_model.config.id2label[label] == "FACT_CHECK_NEEDED"
+            skipped += not needs_check
+        probs = answer_token_probs(
+            d_tok, d_model, device, item["knowledge"], item["question"], item["answer"]
+        )
+        rows.append((item["hallucination"] == "no", needs_check, probs))
```
FAUST-BENCHOU, line 127: The issue asks both sides of the comparison to report avg/p50/p99 latency, and the README quotes a ~13 ms p50 for this baseline, but this script never collects or prints timing data. That makes the latency row impossible to reproduce from the documented command. Could we time the sentinel/detector path here and report the same latency fields as `evaluate_detectors.py`? It would also be useful to clarify how sentinel-skipped samples contribute to that number.

FAUST-BENCHOU (commented):
See inline comments.

bench/hallucination/evaluate_detectors.py
adaamko: fixed. Failed predictions now abort the run by default (so a comparison can never silently cover different subsets), and an explicit --skip-failures flag records failed sample IDs in the output

bench/hallucination/evaluate_halugate.py
adaamko: README number originally came from a separate script. evaluate_halugate.py now times the pipeline per sample and reports the same avg/p50/p99 fields as evaluate_detectors.py. Sentinel-skipped samples contribute sentinel-only time (in production the detector never runs for them; the token probabilities are still computed for the sweep but excluded from timing). README updated with the number produced by the documented command.

adaamko:
@FAUST-BENCHOU thanks for the review, replied to your comments and fixed the issues

FAUST-BENCHOU (approved):
LG, Thanks

PR #2561, opened 2026-07-15
[Router] [ Phase 1 ] authenticate and isolate the management API 

### LINKED ISSUE #2463: [Security] Authenticate and isolate the Router management API
## Summary

The router management listener needs a deny-by-default authentication, authorization, exposure, and audit boundary. Owner surfaces: `src/semantic-router/pkg/apiserver/`, `src/semantic-router/pkg/extproc/router_learning_runtime.go`, `deploy/kserve/service.yaml`, `deploy/kubernetes/istio/service.yaml`, and `src/vllm-sr/cli/container_start.py`.

## Current behavior

Non-health management routes expose configuration and mutable router state without route-bound caller authorization. The listener can bind broadly, live configuration reads can include secret-bearing fields, and learning mutations accept caller-provided provenance without binding it to an owned routing event.

## Expected behavior

Every non-health route is explicitly classified and authorized, protected exposure is intentional, secret values never leave the service, and learning mutations are attributable to an authenticated principal and owned decision.

## Acceptance

- Maintain one route inventory containing method, permission, sensitivity, body policy, and audit action; deny any unregistered route.
- Bind local/default management access to loopback or another protected channel, and require an explicit deployment policy for remote exposure.
- Return a typed redacted configuration view; canary secrets are absent from responses, logs, and errors.
- Derive learning source from authenticated credentials and require an owned decision/event and model binding, idempotency, rate limits, and audit attribution.
- Enforce media types, request-size limits, request IDs, and immutable mutation audit events centrally.

## Validation

- Add exhaustive anonymous and role-matrix tests for every registered route, including independent config, data, learning, and secret-view permissions.
- Run `cd src/semantic-router && go test -race ./pkg/apiserver ./pkg/extproc`.
- Run `make agent-report ENV=cpu CHANGED_FILES=src/semantic-router/pkg/apiserver` followed by the reported gates and affected E2E profiles.
- Add deployment checks proving the listener is not remotely exposed without its protection contract.

## Related

- Parent audit: #2375
- Security index: #2357
- Config transactions: #2326
- Feedback provenance: #1452

AayushSaini101:
## Summary

closes #2463 (Phase 1 only — does **not** fully close the issue yet).

The router management API (`:8080`) previously registered ~55 routes with no authentication and listened on all interfaces. This PR adds a deny-by-default auth boundary while keeping local/dev behavior unchanged.

### What changed

| Area | Change |
|------|--------|
| Route inventory | Every management route declares `Permission`, `Sensitivity`, and optional `AuditAction` via `managedRoute()` in `routes_catalog.go` |
| Auth | Bearer auth with `viewer` / `operator` / `admin` roles; `/health` stays anonymous; `auth.mode: disabled` allows all (local default) |
| Middleware | `wrapRouteHandler` gates each bound handler and sets `X-Request-Id` |
| Config | `global.services.management_api` schema (bind, port, remote exposure, tokens/roles) |
| Bind policy | Default bind `127.0.0.1:8080`; wide bind requires `remote_exposure` or `VLLM_SR_MANAGEMENT_INTERNAL_LISTENER=true` |
| CLI | `--api-bind`, `--management-auth-mode`, `--management-remote-exposure` |
| Container | `start-router.sh` enables internal listener + `-api-bind=0.0.0.0` for docker-network peers |

### Default (backward compatible)

```yaml
global:
  services:
    management_api:
      enabled: true
      bind_address: 127.0.0.1
      port: 8080
      remote_exposure: false
      auth:
        mode: disabled

### DIFF
--- a/src/semantic-router/cmd/runtime_bootstrap.go
+++ b/src/semantic-router/cmd/runtime_bootstrap.go
@@ -32,7 +32,7 @@ type runtimeOptions struct {
 	apiPort                int
 	apiBind                string
 	managementAuthMode     string
-	managementRemoteExpose bool
+	managementRemoteExpose *bool
 	metricsPort            int
 	enableAPI              bool
 	secure                 bool
@@ -66,12 +66,28 @@ func parseRuntimeOptions() runtimeOptions {
 		apiPort:                *apiPort,
 		apiBind:                *apiBind,
 		managementAuthMode:     *managementAuthMode,
-		managementRemoteExpose: *managementRemoteExpose,
+		managementRemoteExpose: boolFlagOverride(flag.CommandLine, "management-remote-exposure", *managementRemoteExpose),
 		metricsPort:            *metricsPort,
 		enableAPI:              *enableAPI,
 		secure:                 *secure,
 		downloadOnly:           *downloadOnly,
 	}
+}
+
+// boolFlagOverride returns a pointer to value only when the named flag was
+// explicitly supplied on the command line. Otherwise nil so config defaults
+// are preserved (e.g. management_api.remote_exposure: true).
+func boolFlagOverride(fs *flag.FlagSet, name string, value bool) *bool {
+	set := false
+	fs.Visit(func(f *flag.Flag) {
+		if f.Name == name {
+			set = true
+		}
+	})
+	if !set {
+		return nil
+	}
+	return &value
 }
 
 func initializeRuntimeLogger() {
@@ -438,18 +454,17 @@ func startAPIServerIfEnabled(opts runtimeOptions, runtimeRegistry *routerruntime
 	}
 
 	go func() {
-		remoteExposure := opts.managementRemoteExpose
 		logging.ComponentEvent("router", "api_server_starting", map[string]interface{}{
 			"api_port":                   opts.apiPort,
 			"api_bind":                   opts.apiBind,
 			"management_auth_mode":       opts.managementAuthMode,
-			"management_remote_exposure": remoteExposure,
+			"management_remote_exposure": opts.managementRemoteExpose,
 		})
 		if err := apiserver.InitWithOptions(apiserver.InitOptions{
 			ConfigPath:      opts.configPath,
 			Port:            opts.apiPort,
 			BindAddress:     opts.apiBind,
-			RemoteExposure:  &remoteExposure,
+			RemoteExposure:  opts.managementRemoteExpose,
 			AuthMode:        opts.managementAuthMode,
 			RuntimeRegistry: runtimeRegistry,
 		}); err != nil {
--- a/src/semantic-router/cmd/runtime_bootstrap.go
+++ b/src/semantic-router/cmd/runtime_bootstrap.go
@@ -24,44 +24,53 @@ import (
 )
 
 type runtimeOptions struct {
-	configPath   string
-	certPath     string
-	kubeconfig   string
-	namespace    string
-	port         int
-	apiPort      int
-	metricsPort  int
-	enableAPI    bool
-	secure       bool
-	downloadOnly bool
+	configPath             string
+	certPath               string
+	kubeconfig             string
+	namespace              string
+	port                   int
+	apiPort                int
+	apiBind                string
+	managementAuthMode     string
+	managementRemoteExpose bool
+	metricsPort            int
+	enableAPI              bool
+	secure                 bool
+	downloadOnly           bool
 }
 
 func parseRuntimeOptions() runtimeOptions {
 	var (
-		configPath   = flag.String("config", "config/config.yaml", "Path to the configuration file")
-		port         = flag.Int("port", 50051, "Port to listen on for gRPC ExtProc")
-		apiPort      = flag.Int("api-port", 8080, "Port to listen on for the router apiserver")
-		metricsPort  = flag.Int("metrics-port", 9190, "Port for Prometheus metrics")
-		enableAPI    = flag.Bool("enable-api", true, "Enable the router apiserver")
-		secure       = flag.Bool("secure", false, "Enable secure gRPC server with TLS")
-		certPath     = flag.String("cert-path", "", "Path to TLS certificate directory (containing tls.crt and tls.key)")
-		kubeconfig   = flag.String("kubeconfig", "", "Path to kubeconfig file (optional, uses in-cluster config if not specified)")
-		namespace    = flag.String("namespace", "default", "Kubernetes namespace to watch for CRDs")
-		downloadOnly = flag.Bool("download-only", false, "Download required models and exit (useful for CI/testing)")
+		configPath             = flag.String("config", "config/config.yaml", "Path to the configuration file")
+		port                   = flag.Int("port", 50051, "Port to listen on for gRPC ExtProc")
+		apiPort                = flag.Int("api-port", 8080, "Port to listen on for the router apiserver")
+		apiBind                = flag.String("api-bind", "", "Bind address for the router apiserver (default from config: 127.0.0.1)")
+		managementAuthMode     = flag.String("management-auth-mode", "", "Management API auth mode: bearer or disabled")
+		managementRemoteExpose = flag.Bool("management-remote-exposure", false, "Allow remote exposure of the management API (requires bearer auth tokens in config)")
+		metricsPort            = flag.Int("metrics-port", 9190, "Port for Prometheus metrics")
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

**Comment by AayushSaini101** [comment]
# Management API request flow (Phase 1)

## Flow diagram

```mermaid
flowchart TD
  client[Client HTTP request] --> listen["Listener bind_address:port<br/>default 127.0.0.1:8080"]
  listen --> mux[ServeMux match Method Path]
  mux --> bind["route.bind wraps wrapRouteHandler"]
  bind --> reqId["Resolve X-Request-Id<br/>reuse header or generate UUID"]
  reqId --> setHdr["Set response X-Request-Id"]
  setHdr --> authMode{auth.mode}

  authMode -->|health.read route| anon["Allow anonymous<br/>role=anonymous"]
  authMode -->|disabled or empty| adminEq["Allow as admin-equivalent<br/>AuthEnabled=false"]
  authMode -->|bearer| bearer{Tokens configured?}

  bearer -->|no| e401a["401 MANAGEMENT_AUTH_NOT_CONFIGURED"]
  bearer -->|yes| tok{Authorization Bearer token valid?}
  tok -->|missing or unknown| e401b["401 UNAUTHORIZED"]
  tok -->|known| perm{Role has route.Permission?}
  perm -->|no| e403["403 FORBIDDEN"]
  perm -->|yes| okAuth[Principal OK]

  anon --> bodyLimit
  adminEq --> bodyLimit
  okAuth --> bodyLimit["Optional MaxBytesReader<br/>if RequestBody.LimitBytes set"]

  bodyLimit --> ctx["Attach request_id + principal to context"]
  ctx --> handler["Route handler<br/>e.g. handleConfigGet"]
  handler --> resp[HTTP response]

  e401a --> errJSON["JSON error + request_id"]
  e401b --> errJSON
  e403 --> errJSON
```

## Step-by-step

1. Request hits the management listener (`ListenAddress()` from `management_api` config / CLI).
2. Mux matches a catalog route (`routes_catalog.go`).
3. `bind()` runs `wrapRouteHandler` (not the handler directly).
4. Middleware assigns `X-Request-Id`.
5. Auth:
   - `/health` → always allowed
   - `auth.mode: disabled` → allowed (local default)
   - `auth.mode: bearer` → token → role → permission check
6. On success, principal + request ID go into context, then the real handler runs.
7. On failure, JSON error with `code` + `request_id` is returned before the handler.

Roles (bearer mode): `viewer` / `operator` / `admin` (`*` for admin), from `global.services.management_api.auth`.

AayushSaini101:
@drivebyer @WUKUNTAI-0211 @Xunzhuo need review on this thanks 🙏

src/semantic-router/cmd/runtime_bootstrap.go
```diff
@@ -429,10 +438,21 @@ func startAPIServerIfEnabled(opts runtimeOptions, runtimeRegistry *routerruntime
...
+			"api_bind":                   opts.apiBind,
+			"management_auth_mode":       opts.managementAuthMode,
+			"management_remote_exposure": remoteExposure,
 		})
-		if err := apiserver.InitWithRuntime(opts.configPath, opts.apiPort, runtimeRegistry); err != nil {
+		if err := apiserver.InitWithOptions(apiserver.InitOptions{
+			ConfigPath:      opts.configPath,
+			Port:            opts.apiPort,
+			BindAddress:     opts.apiBind,
+			RemoteExposure:  &remoteExposure,
```
FAUST-BENCHOU: This always passes a non-nil `RemoteExposure` override, even when the flag was not provided. That turns a configured `remote_exposure: true` back into `false` and can make a valid remote listener fail startup. Could we only set this override when the CLI flag was explicitly supplied?

src/semantic-router/pkg/apiserver/server.go
```diff
@@ -18,22 +18,55 @@ import (
...
 	// This ensures we use the same config as the rest of the application
-	cfg := resolveAPIServerConfig(runtimeRegistry)
+	cfg := resolveAPIServerConfig(opts.RuntimeRegistry)
 	if cfg == nil {
 		return fmt.Errorf("configuration not initialized")
 	}
 
-	classificationSvc := resolveClassificationService(cfg, runtimeRegistry)
-	classificationSvc = ensureClassificationService(cfg, runtimeRegistry, classificationSvc)
+	managementCfg, err := cfg.ManagementAPI.ResolvedManagementAPI(config.ManagementAPIRuntimeOptions{
```
FAUST-BENCHOU, line 57: Where is `management_api.enabled` applied during startup? The current startup path only checks the legacy `-enable-api` flag, so setting `management_api.enabled: false` still starts the listener. Since this PR exposes that field as part of the new config contract, I think startup should honor it or the field should be removed for now.

drivebyer (commented):
Thanks for taking this on, and the catalog test that forces every route to declare policy metadata is a nice guard.

One thing worth a look before merge: the auth policy is read from hot-reloadable config on every request. That means a token with config.write (the operator role) can PATCH /config/router to set auth.mode: disabled, or rewrite roles/tokens, and take down the auth boundary at runtime. The hot-reloaded config also never re-runs the exposure validation.

The general issue is that auth policy is mutable through the very API it protects, which lets a lower privilege escalate. Might be worth keeping the resolved auth policy out of the hot-reload path so changing it isn't a runtime config.write operation.

src/semantic-router/pkg/apiserver/auth.go
```diff
@@ -0,0 +1,119 @@
...
+		Roles:  roles,
+		Tokens: cfg.Auth.ResolvedManagementTokens(),
+	}
+}
+
+func (s *ClassificationAPIServer) managementAPIConfig() config.ManagementAPIConfig {
+	if s == nil || s.config == nil {
+		return config.DefaultManagementAPIConfig()
+	}
+	cfg := s.config.ManagementAPI
```
drivebyer: The auth path reads s.config on every request (managementAuthPolicy()), while the config hot-reload writes s.config = newCfg from another request's goroutine, with no mutex guarding the field. That's an unsynchronized read/write on shared state. Since the acceptance criteria run go test -race ./pkg/apiserver, this should surface there. Worth guarding the config field or reading it through a single synchronized accessor.

src/semantic-router/pkg/apiserver/route_policy.go
```diff
@@ -0,0 +1,71 @@
...
+type RoutePermission string
+
+const (
+	PermHealthRead     RoutePermission = "health.read"
+	PermReadyRead      RoutePermission = "ready.read"
+	PermDocsRead       RoutePermission = "docs.read"
+	PermClassifyInvoke RoutePermission = "classify.invoke"
+	PermConfigRead     RoutePermission = "config.read"
+	PermConfigWrite    RoutePermission = "config.write"
+	PermSecretView     RoutePermission = "secret_view"
```
drivebyer, line 15: PermSecretView is defined but bound to no route. GET /config/router is gated on config.read with no redaction, so any viewer reads plaintext secrets, and operator ends up equivalent to admin. Worth wiring secret_view and redacting the config view.

src/semantic-router/pkg/apiserver/auth.go
AayushSaini101: Thanks, good catch. Now I updated the method with a proper Read lock in all the places, and ran the race condition script as well 

```
func (s *ClassificationAPIServer) currentConfig() *config.RouterConfig {
	if s == nil {
		return nil
	}
	if s.runtimeConfig != nil {
		return s.runtimeConfig.Current()
	}
	s.configMu.RLock()
	defer s.configMu.RUnlock()
	return s.config
}
```
Output: 

Image

src/semantic-router/pkg/apiserver/route_policy.go
AayushSaini101: Done, thanks updated  added redact management config secrets without secret_view

src/semantic-router/pkg/apiserver/server.go
AayushSaini101: thanks, management_api.enabled field has been removed

src/semantic-router/cmd/runtime_bootstrap.go
AayushSaini101: thanks, allowed now via the CLI

drivebyer (approved):

AayushSaini101:
@FAUST-BENCHOU need your review as well thanks : )

PR #734, opened 2025-11-26
feat: add RedisVL as new semantic cache storage

rootfs:
FILL IN THE PR DESCRIPTION HERE

FIX #20 

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to semantic-router! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>DCO and Signed-off-by</h3>
[...]

PR #2567, opened 2026-07-16
[Router] cache: fix Redis addon example to compile against current interface

### LINKED ISSUE #2566: [Router] Redis cache addon example fails to compile (stale BackendConfigPath + missing ttlSeconds argument)
### Describe the bug

The Redis cache addon example (`deploy/addons/redis/redis-cache.go`) no longer compiles against `main`, so `make run-redis-example` fails before it can run.

Two independent breakages:

1. **Removed config field.** Line 31 sets `BackendConfigPath` in a `cache.CacheConfig` literal, but that field was dropped in the config refactor. This is the first compile error:
   ```
   deploy/addons/redis/redis-cache.go:31:3: unknown field BackendConfigPath in struct literal of type cache.CacheConfig
   ```
2. **Write calls omit the required `ttlSeconds` argument** (surfaces once #1 is fixed). The `CacheBackend` interface now takes `ttlSeconds int` on the write methods, but the example calls them with the old arity:
   - `AddEntry(requestID, model, query, requestBody, responseBody)` — missing `ttlSeconds`
   - `AddPendingRequest(requestID, model, query, requestBody)` — missing `ttlSeconds`
   - `UpdateWithResponse(requestID, responseBody)` — missing `ttlSeconds`

This is pre-existing debt from the config refactor and the `ttlSeconds` interface change; it is orthogonal to any in-flight cache work. The Valkey sibling example was already updated — Redis was missed.

### To Reproduce

```
make run-redis-example
# or, just the compile check from the router module:
cd src/semantic-router && go vet ../../deploy/addons/redis/redis-cache.go
```

### Expected behavior

`deploy/addons/redis/redis-cache.go` compiles and `make run-redis-example` runs the demo end-to-end, matching the Valkey example.

### Affected layer

Router (`deploy/addons/redis/redis-cache.go`)

### Additional context

Fix mirrors the Valkey addon: drop the removed `BackendConfigPath` field, build a programmatic `*config.RedisConfig` (from `REDIS_HOST`/`REDIS_PORT`) and pass it via the `Redis` field, and pass `ttlSeconds` on every write call.

WUKUNTAI-0211:
Closes #2566

## What

The Redis cache addon example (`deploy/addons/redis/redis-cache.go`) no longer compiled against `main`:

- it set the removed `BackendConfigPath` field on `cache.CacheConfig` (dropped in the config refactor) — `unknown field BackendConfigPath in struct literal of type cache.CacheConfig`;
- the write calls (`AddEntry` / `AddPendingRequest` / `UpdateWithResponse`) omitted the `ttlSeconds` argument the `CacheBackend` interface now requires.

Nothing in CI builds this file (`test-and-build` skips it; it is only reachable via `make run-redis-example`, which needs a live Redis + Rust bindings), so the breakage went unnoticed.

## How

Mirror the already-migrated Valkey example:

- build a programmatic `*config.RedisConfig` in `newRedisConfig()` (from `REDIS_HOST` / `REDIS_PORT`, defaulting to `localhost:6379`) and pass it via the `Redis` field instead of `BackendConfigPath`;
- pass `ttlSeconds` (3600) on every write call;
- keep the example's existing `FindSimilar` / `FindSimilarWithThreshold` `(body, found, err)` usage, which is unchanged on `main`.

Pure example fix — no production code or interface change. Scope limited to the one file.

## Test Plan

- Confirmed the pre-fix file fails to build: `go vet ../../deploy/addons/redis/redis-cache.go` → `unknown field BackendConfigPath in struct literal of type cache.CacheConfig`.
- Post-fix `go vet` on the file — clean.
- `gofmt -l` on the file — clean.
- End-to-end: ran `go run deploy/addons/redis/redis-cache.go` against a live `redis/redis-stack` (RediSearch/HNSW). Completes with exit 0 — `AddEntry` → Cache HIT on a paraphrased query, custom-threshold (0.75) HIT, and the `AddPendingRequest` → `UpdateWithResponse` workflow all succeed.

### DIFF
--- a/deploy/addons/redis/redis-cache.go
+++ b/deploy/addons/redis/redis-cache.go
@@ -3,70 +3,69 @@ package main
 import (
 	"fmt"
 	"log"
+	"os"
 	"time"
 
 	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/cache"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
 )
 
-func main() {
-	// Example: Setting up Redis cache backend
-	fmt.Println("Redis Cache Backend Example")
-	fmt.Println("===========================")
-
-	// Initialize the embedding model
-	fmt.Println("\n0. Initializing embedding model...")
-	err := candle_binding.InitModel("sentence-transformers/all-MiniLM-L6-v2", true)
-	if err != nil {
-		log.Fatalf("Failed to initialize embedding model: %v", err)
+func newRedisConfig() *config.RedisConfig {
+	host := os.Getenv("REDIS_HOST")
+	if host == "" {
+		host = "localhost"
 	}
-	fmt.Println("✓ Embedding model initialized")
-
-	// Configuration for Redis cache
-	config := cache.CacheConfig{
-		BackendType:         cache.RedisCacheType,
-		Enabled:             true,
-		SimilarityThreshold: 0.85,
-		TTLSeconds:          3600, // Entries expire after 1 hour
-		BackendConfigPath:   "../../examples/runtime/semantic-cache/redis.yaml",
+	port := 6379
+	if portEnv := os.Getenv("REDIS_PORT"); portEnv != "" {
+		fmt.Sscanf(portEnv, "%d", &port)
 	}
 
-	// Create cache backend
-	fmt.Println("\n1. Creating Redis cache backend...")
-	cacheBackend, err := cache.NewCacheBackend(config)
-	if err != nil {
-		log.Fatalf("Failed to create cache backend: %v", err)
-	}
-	defer cacheBackend.Close()
+	cfg := &config.RedisConfig{}
+	cfg.Connection.Host = host
+	cfg.Connection.Port = port
+	cfg.Connection.Timeout = 5 // seconds
+	cfg.Connection.TLS.Enabled = false
 
-	fmt.Println("✓ Redis cache backend created successfully")
+	cfg.Index.Name = "semantic_cache_idx"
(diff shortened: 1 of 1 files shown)


AayushSaini101:
QQ: Are we using Redis addon for testing purpose or anything else because in the end calling the methods, and just print ingFinalStates where we are currently utilising the response, just curious to understand thanks 
image

WUKUNTAI-0211:
Good question. The Redis addon (`deploy/addons/redis/redis-cache.go`) is a standalone runnable example, not part of the serving path — it exists to demonstrate the semantic-cache `CacheBackend` API end-to-end (`AddEntry` → `FindSimilar` / `FindSimilarWithThreshold` → the pending-request workflow → stats) and doubles as a manual smoke test via `make run-redis-example`. It only prints results rather than consuming the response because its job is to *show* the API surface working against a live Redis, not to route anything.

It's an existing example (added in #734, "add RedisVL as new semantic cache storage") that had silently bit-rotted: nothing in CI builds it (`test-and-build` skips it, and it's only reachable via `make run-redis-example`, which needs a live Redis + Rust bindings), so the breakage went unnoticed. This PR just realigns the example with the current interface (the removed `BackendConfigPath` field and the new `ttlSeconds` arg on the write methods).

I verified it both ways: pre-fix it fails `go vet` (`unknown field BackendConfigPath in struct literal of type cache.CacheConfig`), and post-fix it runs clean against `redis/redis-stack` (RediSearch/HNSW) — Cache HIT on a paraphrased query, custom-threshold HIT, and the pending → update workflow all succeed.

AayushSaini101:
> known field BackendConfigPath i

thanks for the explanation

AayushSaini101 (approved):
LGTM thanks : )

PR #2569, opened 2026-07-16
[Router] validate periodic ticker intervals and make storage close concurrency-safe

### LINKED ISSUE #2475: config: reject non-positive periodic intervals before ticker creation
## Summary

Validate every ticker-driving duration before runtime construction and make periodic start/close lifecycle idempotent. Owner surfaces: `src/semantic-router/pkg/selection/elo.go`, `src/semantic-router/pkg/selection/rl_driven.go`, `src/semantic-router/pkg/selection/lookuptable/file.go`, `src/semantic-router/pkg/extproc/router_selection.go`, and `src/semantic-router/pkg/config/validator_learning.go`.

## Current behavior

Syntactically valid zero or negative durations can reach ticker construction and panic the process or leave periodic behavior silently inactive. At least one storage close path also uses a check-then-close pattern that is unsafe under concurrent close.

## Public API boundary

The listed Elo, RL-driven, and lookup-table files are legacy/internal cleanup surfaces on current `main`. Fixing their ticker/lifecycle safety must not reintroduce their historical names as supported public Router Learning configuration; #2242 owns that migration boundary.

## Expected behavior

Every periodic duration is positive and bounded at config ingestion and constructor boundaries, and start/close behavior is concurrency-safe and idempotent.

## Acceptance

- Require every ticker-driving duration to be positive and below a documented maximum.
- Return a field-addressed config error before runtime construction.
- Defensively reject invalid intervals in constructors.
- Define and synchronize start, closing, and closed transitions.
- Make repeated and concurrent close safe.

## Validation

- Add table tests for zero, negative, overflow, malformed, and maximum-boundary durations.
- Add repeated-start and high-concurrency close tests under `-race`.
- Run `cd src/semantic-router && go test -race ./pkg/selection ./pkg/selection/lookuptable ./pkg/extproc ./pkg/config`.
- Run the selection changed-surface gates reported by `make agent-report`.

## Related

- Parent audit: #2375

theohsiung:
## Summary

Every periodic ticker interval in the router — Elo/RL rating auto-save, lookup-table auto-save, and lookup-table replay populate — is parsed from a config string and handed straight to `time.NewTicker`, which **panics on a non-positive duration**. A syntactically valid `"0s"` or `"-5s"` parses successfully and reaches the ticker, crashing the auto-save goroutine (and with it the whole process), or — for the replay populator wrapped in `goSafely` — being silently swallowed so periodic re-derivation never runs. Separately, the lookup-table `FileStorage.Close()` used an unsafe check-then-close on its stop channel, so concurrent `Close()` calls double-closed the channel and panicked.

This PR validates every ticker-driving duration before construction and makes the storage start/close lifecycle idempotent and concurrency-safe.

Closes #2475.

## Changes

- **`pkg/config/periodic_interval.go` (new):** `ParsePeriodicInterval(value, def)` + `MaxPeriodicInterval` (24h) — one shared rule: empty → default; reject unparseable, non-positive, or above the documented maximum.
- **`pkg/selection/lookuptable/file.go`:** `Close()` now uses `sync.Once` (idempotent + concurrency-safe, matching `FileEloStorage`); `StartAutoSave` defensively clamps a non-positive interval to a 30s default.
- **`pkg/selection/auto_save_interval.go` (new) + `elo.go` / `rl_driven.go`:** both selectors route `auto_save_interval` through a shared `resolveAutoSaveInterval` helper (validate + warn + fall back) instead of trusting any value that parses.
- **`pkg/selection/storage.go`:** `FileEloStorage.StartAutoSave` defensively clamps a non-positive interval.
- **`pkg/extproc/router_selection.go`:** `buildLookupTableStorage` and `maybePopulateFromReplay` validate intervals via `ParsePeriodicInterval`; `startLookupTablePopulator` defensively clamps a non-positive interval.

### Scope note (Public API boundary)

The Elo / RL-driven / lookup-table interval fields are already hard-rejected as public config on `main` (they are `yaml:"-"` and rejected in `validateModelSelectionConfig` / `validateMigratedLearningAlgorithm`). Per the issue's boundary note, this PR does **not** re-introduce those names as supported configuration in `validator_learning.go`; the migration boundary is owned by #2242. The fix here is defensive: constructor-level validation so a bad duration can never panic or silently disable periodic work, regardless of how it reaches the runtime.

## Test Plan

- **Table tests** (`ParsePeriodicInterval`): empty, valid, zero, bare `0`, negative, malformed, overflow, above-max, and the `24h` max boundary.
- **Lifecycle `-race` tests:** repeated sequential `Close()`, 64-goroutine concurrent `Close()`, `Close()` before `StartAutoSave`, and non-positive-interval `StartAutoSave` (for both `FileStorage` and `FileEloStorage`).
- **Resolver tests:** `resolveAutoSaveInterval` falls back to the default on empty/zero/negative/malformed/above-max and honors valid values.
- Gate: `cd src/semantic-router && go test -race ./pkg/config ./pkg/selection ./pkg/selection/lookuptable ./pkg/extproc` — all green.
- Lint: `golangci-lint --new-from-rev=origin/main` — 0 new issues.

### DIFF
--- a/src/semantic-router/pkg/config/periodic_interval.go
+++ b/src/semantic-router/pkg/config/periodic_interval.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2025 vLLM Semantic Router.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+// MaxPeriodicInterval bounds every ticker-driving interval accepted by the
+// router (Elo/RL auto-save flushes and lookup-table auto-save/replay-populate
+// cadence). Anything larger is almost certainly a misconfiguration: it would
+// make periodic work effectively never run and risks duration overflow. This
+// is the documented maximum referenced by the router configuration contract.
+const MaxPeriodicInterval = 24 * time.Hour
+
+// ParsePeriodicInterval parses a periodic interval string such as "30s" or
+// "5m" into a time.Duration suitable for driving a time.Ticker.
+//
+// An empty (or whitespace-only) value yields def with no error, letting
+// callers keep their documented default. A non-empty value must parse and be
+// strictly positive and no greater than MaxPeriodicInterval; otherwise a
+// descriptive error is returned (with the offending value quoted) so callers
+// can reject the configuration or fall back to a safe default. Because
+// time.NewTicker panics on a non-positive duration, callers must never pass a
+// value that failed this check to a ticker.
+func ParsePeriodicInterval(value string, def time.Duration) (time.Duration, error) {
+	trimmed := strings.TrimSpace(value)
+	if trimmed == "" {
+		return def, nil
+	}
+
+	d, err := time.ParseDuration(trimmed)
+	if err != nil {
+		return 0, fmt.Errorf("invalid duration %q: %w", value, err)
+	}
+	if d <= 0 {
+		return 0, fmt.Errorf("duration must be positive, got %q", value)
+	}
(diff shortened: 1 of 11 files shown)


theohsiung:
@WUKUNTAI-0211 could you take a look when you have a chance? This hardens periodic ticker-interval validation and makes the lookup-table storage close idempotent/concurrency-safe (#2475). Appreciate any feedback 🙏

FAUST-BENCHOU (approved):
LGTM

PR #2573, opened 2026-07-16
[Website] add homepage sponsors section

### LINKED ISSUE #2571: [Website] Add a sponsor section to the homepage
Add Sponser Section in the main website page

AayushSaini101:
closes: #2571


website/src/pages/index.tsx
```diff
@@ -525,11 +525,7 @@ export default function Home(): JSX.Element {
           </ScrollReveal>
         </div>
 
-        <div className={styles.bandBlack}>
-          <ScrollReveal delay={40}>
-            <AcknowledgementsSection />
-          </ScrollReveal>
-        </div>
+        <SponsorsSection />
```
FAUST-BENCHOU, line 528: Was removing `AcknowledgementsSection` intentional here? The linked issue asks to add sponsors, but this replacement also removes the homepage section that credits vLLM, PyTorch, Envoy, Kubernetes, and the other open-source dependencies. Could we keep both sections unless consolidating them is part of the intended redesign?

Xunzhuo:
Thanks for putting this together. I’m closing this with #2571 because the proposed roster is explicitly copied from the broader vLLM sponsor list, and we do not have a verified Semantic Router-specific source for those claims. We should revisit the section only after that ownership and roster are defined.

Issue #2564, opened 2026-07-16
bug: Mergify queue is incompatible with up-to-date branch protection


WUKUNTAI-0211:
## Describe the bug

Mergify reports that the queue configuration is incompatible with the GitHub branch protection setting `Require branches to be up to date before merging`. Approved pull requests are not automatically updated with the latest `main`.

## To Reproduce

1. Enable the `Require branches to be up to date before merging` protection on `main`.
2. Queue an approved pull request through the repository Mergify configuration.
3. Observe the Mergify configuration warning and that the PR is not updated in place.

## Expected behavior

Mergify should update the queued PR branch in place by merging the latest `main`, rerun CI, and continue the queue.

## Affected layer

CI/Build: `.mergify.yml`

## Additional context

Mergify requires `merge_queue.max_parallel_checks: 1`, `batch_size: 1` for every queue rule, and matching `queue_conditions`/`merge_conditions` for in-place checks.

Closes this issue with the follow-up pull request.

PR #2579, opened 2026-07-16
[CI/Build] Gate Mergify merge on test-and-build CI result

### LINKED ISSUE #2578: bug: Mergify merge queue does not gate merges on re-run CI success
## Describe the bug

The Mergify `default` queue rule updates each approved PR against the latest `main` and re-runs CI, but `merge_conditions` does not require that re-run to be **green** before merging. It currently gates only on `base = main`, which is always true for queued PRs, so the merge happens regardless of the re-run CI result.

This defeats the main purpose of the merge queue: catching **semantic conflicts** between two PRs that are each green on their own base and have no textual conflict, but break `main` once combined (e.g. one PR renames a helper, another still calls the old name from a different file). The queue's "update against `main` + re-run CI" step surfaces this breakage, but only if the merge is gated on that re-run passing.

There are no required status checks on `main` today (the branch ruleset is disabled), so `branch_protection_injection_mode` has nothing to inject — `merge_conditions` is the only place the gate can live.

## To Reproduce

1. Open two approved PRs off `main` that are each green in isolation and do not touch the same lines (e.g. PR A renames `foo()` -> `bar()` and updates its own callers; PR B adds a new caller of `foo()` in a different file).
2. Let both enter the Mergify queue.
3. Observe that they merge even though the post-update re-run of CI for the second PR would fail — `main` build breaks.

## Expected behavior

The queue should merge a PR only after the re-run CI on the updated branch reports success (or is legitimately skipped). `test-and-build` is path-filtered (SKIPPED on docs/config-only PRs via its `if:` guard), so the gate must accept the skipped state to avoid deadlocking those PRs.

## Affected layer

CI/Build: `.mergify.yml`

## Additional context

Raised by @theohsiung during review of #2529. Follow-up to #2564/#2565, which fixed in-place-check compatibility but did not add the CI-result gate.

WUKUNTAI-0211:
Closes #2578

## What

Make the Mergify merge queue actually gate merges on `test-and-build` CI success. Two coupled changes:

1. **`.mergify.yml`** — add a `merge_condition` requiring `test-and-build` to pass, OR-ed with its skipped state:
   ```yaml
   merge_conditions:
     - base = main
     - or:
   ```
2. **`.github/workflows/test-and-build.yml`** — run the job on Mergify's merge-queue draft PRs so that gate can actually execute:
   ```yaml
   if: >-
     ${{ (!github.event.pull_request.draft
          || startsWith(github.head_ref, 'mergify/merge-queue/'))
         && (github.event_name == 'schedule' || ...path filters) }}
   ```

## Why

The queue already updates each approved PR against the latest `main` and re-runs CI (from #2564/#2565), but `merge_conditions` only required `base = main` — always true for queued PRs — so the merge happened regardless of the re-run result. That defeats the queue's core purpose.

@theohsiung raised this during review of #2529: two approved PRs, each green in isolation with no textual conflict, can still break `main` once combined (a **semantic conflict** — e.g. one PR renames a helper, another still calls the old name from a different file). The "update against `main` + re-run CI" step catches that class of breakage, but only if the merge is gated on the re-run passing.

**Why the second change is required, not optional:** adding a `merge_condition` that differs from `queue_conditions` makes Mergify validate the queue through a temporary **draft** PR (speculative check) rather than in-place. `test-and-build` previously skipped *all* draft PRs (`if: !github.event.pull_request.draft`), so on that draft the check reported **skipped**, `check-skipped = test-and-build` passed vacuously, and the queue could still merge without ever running CI. Allowing the job on `mergify/merge-queue/*` head refs closes that hole while keeping ordinary WIP drafts skipped.

Net behavior:

- **code PRs** — CI runs on the queued merge result and must be green;
- **docs/config-only PRs** — path-filtered to skipped, merge via the `check-skipped` branch (no deadlock);
- **user WIP drafts** — still skipped (no wasted CI).

## Caveat

This couples to Mergify's default `queue_branch_prefix` (`mergify/merge-queue/`). If a custom `queue_branch_prefix` is ever set in `.mergify.yml`, this guard must be updated in lockstep or the gate silently goes vacuous again.

## Test Plan

- Both files validated as well-formed YAML locally.
- Verified `test-and-build`'s guard is `!github.event.pull_request.draft && (core|helm|e2e|docker|make|ci)` in `.github/workflows/test-and-build.yml`, confirming it reports SKIPPED (not failed) on non-core PRs — hence the `check-skipped` branch.
- Confirmed Mergify's default queue branch prefix is `mergify/merge-queue/`, and that its speculative-check draft PRs evaluate `merge_conditions` against the temporary PR.
- `.mergify.yml` change is validated by Mergify's own `Configuration changed` check on this PR.


WUKUNTAI-0211:
@theohsiung this implements the CI-result gate you flagged in #2529 — `merge_conditions` was only `base = main` (always true), so the queue re-ran CI but merged regardless. Added your suggested `or: check-success / check-skipped = test-and-build` so semantic conflicts (each PR green alone, combined breaks `main` like #2459) get caught, without deadlocking docs/config PRs where `test-and-build` is path-filtered to SKIPPED. Would appreciate your review.

WUKUNTAI-0211:
Follow-up: the first commit alone was **not** an effective gate — noting the fix here for reviewers.

Because \`merge_conditions\` now differs from \`queue_conditions\`, Mergify validates the queue via a temporary **draft** PR (speculative check) rather than in-place. \`test-and-build\` skipped *all* draft PRs (\`if: !github.event.pull_request.draft\`), so on that draft the check reported **skipped**, \`check-skipped = test-and-build\` passed vacuously, and the queue could merge without CI ever running.

Second commit fixes this: \`test-and-build\` now also runs when the PR head ref starts with \`mergify/merge-queue/\` (Mergify's default \`queue_branch_prefix\`), while ordinary WIP drafts stay skipped. Net behavior:

- code PRs → CI runs on the queued merge result, must be green;
- docs/config-only PRs → path-filtered to skipped, merge via the \`check-skipped\` branch;
- user WIP drafts → still skipped (no wasted CI).

Caveat worth flagging: this couples to the default \`queue_branch_prefix\`. If we ever set a custom \`queue_branch_prefix\` in \`.mergify.yml\`, this guard must be updated in lockstep or the gate silently goes vacuous again.

WUKUNTAI-0211:
@theohsiung heads-up on the full picture: the `or:` block alone was **not** enough. Because differing `merge_conditions` push Mergify to a **draft** PR for queue validation, and `test-and-build` skipped all drafts, `check-skipped` passed vacuously and the queue could still merge without CI. The second commit (`test-and-build.yml`) runs the job on `mergify/merge-queue/*` head refs so the gate actually executes — details in the comment above and the updated PR description. Would still value your review on whether coupling to the default `queue_branch_prefix` is acceptable or you'd prefer an always-run aggregate check.

theohsiung (approved):
LGTM.

**One maintainability note (not a bug)**
.github/workflows/test-and-build.yml:32 — the guard hard-codes the string 'mergify/merge-queue/', coupling the CI gate to Mergify's default queue_branch_prefix. If a custom queue_branch_prefix is ever set in .mergify.yml, this guard silently goes vacuous again (skipped draft → check-skipped → merge without CI). The author already documents this in the PR's "Caveat" section, so it's a known, accepted tradeoff — worth a comment cross-linking the two files, but not a blocker.

AayushSaini101:
@WUKUNTAI-0211 I will do review by tomorrw thanks for higlighting the mergify improvement 🙇

AayushSaini101 (approved):
LGTM thanks : )

Issue #2336, opened 2026-07-05
[Feature] Optimize and harden existing Looper, Flow, Fusion, and ReMoM algorithms


Xunzhuo:
### Goal

Harden and optimize the existing Looper, Flow, Fusion, ReMoM, and related multi-model collaboration paths so they are safe, bounded, explainable, and measurable.

### Proposed Workgroup

MoM & Routing

### Scope

- Define hard call, concurrency, token, cost, and wall-time limits for existing algorithms.
- Unify candidate-panel execution, usable-response classification, cancellation, draining, quorum reporting, and per-attempt provenance.
- Separate per-attempt error handling from panel-level quorum and fallback policy.
- Add bounded verifier, selection, synthesis, and aggregate parent/child accounting contracts.
- Preserve replay explanations for algorithm path, candidate roles, intermediate decisions, and final disposition without leaking sensitive content.
- Harden model references, request headers, output contracts, tool-call behavior, and recursive-target validation.
- Keep entrypoint, recipe, virtual-slug, CLI, documentation, and protocol behavior aligned.
- Maintain deterministic evaluation fixtures and quality, latency, token, and cost baselines.

Cost accounting, panel execution, outcome verification, fan-out controls, protocol handling, fallback policy, and evaluation may land independently. Each pull request should state which invariants it closes and must not rely on an unmerged sibling for safety.

##### Proposed Workgroup

MoM & Routing

##### Background

The repository already has a broad `pkg/looper` surface, but algorithm families differ in panel execution, cost accounting, verification, fallback, replay, and protocol behavior. These foundations can be improved in parallel with bounded algorithm research; every change must enforce its own safety invariants and remain independently reviewable.

Related: #1443, #1456, #1815, #2200, #2331, #2333, #2354, #2357. Parent Epic: #3037. Completed benchmark foundation: #2371.

##### Validation

Start with:

```bash
make agent-report ENV=cpu CHANGED_FILES="<space-separated changed files>"
```

Then run the gates and affected E2E profiles reported by the repository harness.

### Non-goals

- Add a new Looper algorithm in this issue.
- Implement arbitrary workflow DAGs or bypass safety and hard-policy boundaries.
- Make research progress contingent on every hardening task landing first.

### Completion signal

- [ ] Existing algorithms enforce clear breadth, amplification, and aggregate budget caps.
- [ ] Candidate execution, quorum, fallback, cancellation, and accounting semantics are explicit and tested.
[...]

PR #2580, opened 2026-07-16
[Router] Bound looper upstream response reads to a byte ceiling

### LINKED ISSUE #2581: feat: bound looper upstream response reads to a byte ceiling
### Describe the feature

The looper HTTP client (`src/semantic-router/pkg/looper/client.go`) reads upstream model responses with an unbounded `io.ReadAll(resp.Body)`. Every Looper-family algorithm (Looper/Confidence, Fusion, ReMoM, Workflows) funnels its upstream calls through `Client.CallModel`, so a single huge or malicious upstream response can exhaust router memory — amplified N-fold by the parallel fan-out algorithms. The looper endpoint is frequently itself another router, so its response is not inherently trusted.

Add a configurable byte ceiling on the upstream response read:

- success bodies read through a helper that returns an explicit error on over-cap input instead of silently truncating (and mis-parsing) it;
- non-2xx error bodies capped to a small diagnostic prefix (marked truncated) while still surfacing the status code;
- configurable via a new `max_response_bytes_mb` knob on the looper config (default 32 MiB), mirroring the existing `grpc_max_msg_size_mb` pattern.

The cap applies to both non-streaming and streaming paths and — because Go's transport decompresses before `resp.Body` — also bounds decompression-bomb responses.

### Primary layer

global level

### Why this layer?

The looper runtime config lives under `global.integrations.looper`; the ceiling is a runtime knob on that integration, applied uniformly to every Looper-family algorithm through the shared `Client.CallModel`.

Sub-task of #2336 (optimize and harden existing Looper/Flow/Fusion/ReMoM algorithms) and addresses the "byte-bounded upstream reads and error bodies" item from the #2375 audit. It closes an OOM/DoS vector that is currently unbounded on the shared upstream-call path.

### Additional context

Smallest, dependency-free first slice of the #2336 hardening track — one seam (`Client.CallModel`) protects all Looper-family algorithms. Implemented in #2580.

theohsiung:
Closes #2581

## Summary

The looper HTTP client (`pkg/looper/client.go`) read upstream model responses with an unbounded `io.ReadAll(resp.Body)`. Every Looper-family algorithm (Looper/Confidence, Fusion, ReMoM, Workflows) funnels its upstream calls through `Client.CallModel`, so a single huge or malicious upstream response could exhaust router memory — amplified N-fold by the parallel fan-out algorithms. The looper endpoint is frequently itself another router, so its response is not inherently trusted.

This PR bounds that read:

- **Success bodies** are read through a new `readLimitedBody` helper that returns an explicit error when the body exceeds the ceiling, rather than silently truncating (and then mis-parsing) it.
- **Non-2xx error bodies** (diagnostic only) are capped to a small 8 KiB prefix and marked `...(truncated)` when clipped, while still surfacing the real status code.
- The ceiling is **configurable** via a new `max_response_bytes_mb` knob on the looper config (default 32 MiB), mirroring the existing `grpc_max_msg_size_mb` pattern.

The cap applies to both the non-streaming and streaming paths, and — because Go's transport decompresses before `resp.Body` is exposed — also bounds decompression-bomb responses.

## Motivation

Part of #2336 (optimize and harden existing Looper/Flow/Fusion/ReMoM algorithms). Addresses the "byte-bounded upstream reads and error bodies" item from the #2375 audit. This is the smallest, dependency-free first slice of that hardening track: one seam (`Client.CallModel`) protects all Looper-family algorithms.

## Changes

- `pkg/config/runtime_config.go`: add `LooperConfig.MaxResponseBytesMB`, `DefaultMaxResponseBytes` (32 MiB), and `GetMaxResponseBytes()`.
- `config/config.yaml`: document `max_response_bytes_mb` on the canonical looper block (keeps the reference-config coverage test green).
- `pkg/looper/client_read_limit.go` (new): `readLimitedBody` (errors on over-cap, rejects a non-positive cap) + `maxErrorBodyBytes`.
- `pkg/looper/client.go`: `CallModel` reads the success body via `readLimitedBody` and bounds error bodies; `Client` stores the resolved ceiling from `NewClient`.

## Test Plan

- [x] `go test ./pkg/looper/ ./pkg/config/` (unit + `httptest` integration) — green
- [x] `go test -race` on both packages — green (changed code is race-clean)
- [x] `gofumpt` / `go vet` / `golangci-lint` (repo config) — clean on changed files
- New tests:
  - `readLimitedBody` boundaries: within / exactly-at-cap / cap+1 / far-over / non-positive cap
  - config getter: default-when-unset / explicit MB / non-positive fallback
  - `CallModel` over HTTP: oversized JSON rejected, oversized **streaming** body rejected, oversized error body truncated with status preserved, within-cap accepted
- [ ] E2E: not required — internal read-bounding with a generous default; no routing/API surface change and the config field is additive + optional. Happy to add an E2E assertion if maintainers prefer.

## Notes

- Commits are split as a reviewable trajectory: (1) config knob, (2) helper + `CallModel` wiring + tests.
- Out of scope: a pre-existing unsynchronized-`seenModels` data race in `TestFusionLooperExecutesPanelJudgeAndFinal` exists on `main` and is not touched here (candidate for a separate small test fix).


theohsiung:
@WUKUNTAI-0211 can you take a look

AayushSaini101:
> @WUKUNTAI-0211 can you take a look

one suggestion: @theohsiung can you keep the writing comments 😅 smaller, : )

AayushSaini101 (approved):
LGTM thanks, feel free to update the docs if needed as we have introduced a new entry in the MaxResponseBytesMB  🙇

PR #2589, opened 2026-07-17
[Docs] Document looper max_response_bytes_mb knob

### LINKED ISSUE #2588: docs: document the looper max_response_bytes_mb config knob
### Describe the feature

#2580 added an optional looper config knob `max_response_bytes_mb` (caps a single upstream response body; default 32 MiB) but merged without docs. Document it where the `global.integrations.looper` block appears in the algorithm/looper tutorials (`website/docs/tutorials/algorithm/looper/*.md`), so users can discover the knob and its default.

Thanks to @AayushSaini101 for flagging this on #2580 🙏

### Primary layer

global level

### Why this layer?

The knob lives under `global.integrations.looper` and applies to every Looper-family algorithm through the shared upstream client.

Right now the knob is only visible in the reference `config/config.yaml`; the tutorials don't mention it (nor the sibling runtime knobs). Documenting it keeps the config surface discoverable. Follow-up to #2580 / #2336.

### Additional context

Minimal doc-only change — add the optional field (with a one-line comment) to the `looper:` example block(s).

theohsiung:
Closes #2588

## Summary

Document the optional `max_response_bytes_mb` looper knob (added in #2580) in the ReMoM, Fusion, and Workflows tutorials, next to the `global.integrations.looper` example.

## Why

The knob was only discoverable in the reference `config/config.yaml`; the tutorials didn't mention it. Thanks @AayushSaini101 for flagging this on #2580 🙏

## Test Plan

- Docs-only, 3-line change. `MD013` (line length) is disabled in the repo markdownlint config, so the inline comment is fine.
- Netlify deploy preview renders the updated tutorial pages.

### DIFF
--- a/website/docs/tutorials/algorithm/looper/fusion.md
+++ b/website/docs/tutorials/algorithm/looper/fusion.md
@@ -142,6 +142,7 @@ global:
   integrations:
     looper:
       endpoint: http://localhost:8899/v1/chat/completions
+      max_response_bytes_mb: 32 # optional; caps a single upstream response body (default 32 MiB)
       fusion:
         model_names:
           - vllm-sr/fusion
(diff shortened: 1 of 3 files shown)


AayushSaini101 (approved):
LGMT thanks: )

PR #2593, opened 2026-07-17
[Router] Populate looper streaming token usage

### LINKED ISSUE #2592: bug: looper streaming responses report zero token usage
### Describe the bug

On the streaming path the looper client reports token usage as `{0,0,0}`. Two causes compound:

1. Streaming requests are sent **without** `stream_options.include_usage`, so vLLM never emits the final usage chunk (`setStreamParam` only sets `stream`).
2. Even if it did, `parseStreamingResponse` never populates `ModelResponse.Usage` (it only extracts content deltas).

Non-streaming records usage correctly. Because `SumUsage` aggregates per-call usage, a multi-model Looper streaming request also totals to zero — so token/cost accounting is wrong for every streamed Looper request (the common case, since UIs stream).

### To Reproduce

1. Configure a Looper decision (e.g. ReMoM/Fusion) behind the router.
2. Send a chat completion with `"stream": true`.
3. Inspect the looper response usage.
4. Observe `prompt_tokens=0, completion_tokens=0, total_tokens=0` despite real token consumption.

### Expected behavior

Streaming looper responses report the real token counts, matching the non-streaming path.

### Affected layer

global level (the `global.integrations.looper` client)

### Additional context

Part of #2336 (harden Looper/Flow/Fusion/ReMoM), scope item "latency & token accounting". Fix is local to `pkg/looper/client.go`: (1) set `stream_options.include_usage: true` on streaming requests; (2) parse the final SSE `usage` chunk into `Usage`. Follow-up to #2580.

theohsiung:
Closes #2592

## Summary

Streaming looper calls reported `{0,0,0}` token usage. Two compounding causes: the client never set `stream_options.include_usage` (so vLLM omitted the usage chunk), and the SSE parser ignored it anyway. This sets `include_usage` on streaming requests and parses the trailing usage chunk into `ModelResponse.Usage`.

## Why

Non-streaming already records usage; streaming — the common case for UIs — reported zero, so token/cost accounting was wrong for every streamed Looper request. Part of #2336 (latency & token accounting).

## Test Plan

- `go test ./pkg/looper/` green; `-race` clean on the new tests.
- New tests: `parseStreamingUsage` (usage present / absent / null-ignored), `parseStreamingResponse` populates `Usage`, `setStreamParam` sets `include_usage` when streaming and still drops `stream_options` when not.

### DIFF
--- a/src/semantic-router/pkg/looper/client.go
+++ b/src/semantic-router/pkg/looper/client.go
@@ -288,16 +288,18 @@ func (c *Client) parseStreamingResponse(body []byte, modelName string) (*ModelRe
 		IsStreaming: true,
 	}
 
-	// Parse SSE chunks to extract content
+	// Parse SSE chunks to extract content and usage
 	content, chunks := parseSSEContent(body)
 	result.Content = content
 	result.StreamingChunks = chunks
+	result.Usage = parseStreamingUsage(body)
 
 	logging.ComponentDebugEvent("looper", "model_call_completed", map[string]interface{}{
-		"decision":    c.decisionName,
-		"model_ref":   modelName,
-		"content_len": len(content),
-		"streaming":   true,
+		"decision":     c.decisionName,
+		"model_ref":    modelName,
+		"content_len":  len(content),
+		"total_tokens": result.Usage.TotalTokens,
+		"streaming":    true,
 	})
 
 	return result, nil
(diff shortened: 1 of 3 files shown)


FAUST-BENCHOU (approved):

WUKUNTAI-0211 (approved):
LGTM

PR #2612, opened 2026-07-20
[Router] add entrypoints and multi-recipe routing config contract

drivebyer:
Closes #2614. Part of #2331.

## What

Part 1 of #2331 — the config contract layer only. The runtime layer (request-time recipe resolution, `/v1/models` listing, E2E) follows in a stacked PR so each review stays focused.

- Canonical `entrypoints` / `recipes` schema and normalization into `RouterConfig`, with the default-recipe bridge: the top-level `routing` block **is** the `default` recipe, `recipes` adds named additional profiles, and existing flat read sites keep working unchanged.
- Validation: duplicate recipe names, a `default`-named recipe conflicting with a non-empty top-level `routing` profile, recipe-owned `modelCards` (the model catalog stays shared), unknown recipe references, empty/duplicate entrypoint model names, and recipe decision modelRef checks.
- Canonical export and config dump emit the normalized entrypoints and recipes and round-trip cleanly; the static (CRD base) export excludes them.
- Reference config `config/config.yaml` gains a `privacy-first` recipe plus two entrypoints, enforced by the existing reference-config coverage ratchet.
- Docs in the same change per the config-package contract: `config/README.md`, `website/docs/installation/configuration.md`, the v0.3 unified-config-contract proposal, and a new tutorial page `tutorials/global/entrypoints-and-recipes.md` with sidebar wiring. A new doc-needle test (`docs_contract_recipes_test.go`) ratchets the terminology.

Zero behavior change for existing configs: entrypoints/recipes are optional additive keys, and single-profile configs normalize exactly as before.

The two open design questions raised in the issue thread (recipe-level `strategy`/`model_selection` defaults, and entrypoint-vs-model-name collision validation) stay pending maintainer confirmation and are intentionally excluded.

## Test Plan

- `go test ./pkg/config/... ./pkg/dsl/...` — includes new normalization/bridge/round-trip tests and a 7-case validation-error table in `canonical_recipes_test.go`, plus the doc-needle contract test
- `make agent-lint CHANGED_FILES=...` passes end to end (pre-commit hooks, golangci structural lint, reference-config contract lint, structure checks)
- `markdownlint` clean on all touched docs

Related: #2331 (config layer), roadmap #2287, follow-up #2354.

Issue #2354, opened 2026-07-05
[Feature] Normalize automatic model names and algorithm virtual slugs


Xunzhuo:
### Goal

Normalize the existing `auto_model_name`, `auto_model_names`, and algorithm virtual slugs into the new entrypoints/recipes model so compatibility behavior is explicit instead of scattered across request handling.

Related code: `src/semantic-router/pkg/config/config.go`, `src/semantic-router/pkg/extproc`, `src/vllm-sr/cli`, `deploy/recipes`, `src/semantic-router/pkg/dsl`

### Proposed Workgroup

MoM & Routing

### Scope

Define compatibility and normalization for:

- `auto_model_name` and `auto_model_names` migration into `entrypoints`;
- algorithm virtual slugs such as MoM/Looper/Fusion-style request model names;
- default working recipe mapping for `vllm-sr/auto`, `auto`, and existing aliases;
- config dump/validate output that explains normalized entrypoints;
- CLI/dashboard/operator handling of compatibility fields.

##### Motivation

Multiple recipe support should not leave old auto-model aliases and algorithm virtual slugs as special cases. The current config has `AutoModelName` and `AutoModelNames`, while Looper/MoM-style virtual names have historically acted like model names. These should map into entrypoints outside `routing`, then select a named recipe/profile.

### Non-goals

- Do not remove legacy fields without migration and validation errors.
- Do not put `entrypoints` inside `routing`; entrypoints select a recipe from above routing.
- Do not couple algorithm slugs to signal definitions.

### Completion signal

- Existing auto model names continue to route through a normalized entrypoint representation.
- Multiple recipe configs can map different virtual model names to different recipes.
- Config validation detects ambiguous aliases and conflicting entrypoint mappings.
- Docs explain migration from old auto-model fields.

Parent Epic: #2971
Related completed recipe and entrypoint work: #2331

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #2331, opened 2026-07-05
feat: support entrypoints and multiple routing recipes


Xunzhuo:
## Summary

Support multiple routing recipes and external entrypoints so different virtual model names can select different routing profiles without duplicating global system configuration.

Parent roadmap: #2287
Child issue: #2354
Related: #2326, #2330, #2333, #2342

## Motivation

The current config has a single inline `IntelligentRouting` surface and compatibility fields such as `auto_model_name` / `auto_model_names`. That forces one working routing profile for aliases like `vllm-sr/auto`, and makes algorithm virtual slugs harder to reason about. H2 needs multiple recipes mapped from entrypoints while preserving the core abstraction: signals, projections, decisions, strategy/model_selection.

## Proposed config direction

Keep shared global/provider/system config outside routing, then let entrypoints select a recipe:

```yaml
entrypoints:
  - model_names: ["vllm-sr/auto", "auto", "MoM"]
    recipe: default
  - model_names: ["vllm-sr/privacy"]
    recipe: privacy

recipes:
  - name: default
    routing:
      signals: []
      projections: {}
      decisions: []
      strategy: priority
      model_selection: {}
```

`routing` keeps the existing profile shape. The new layer above routing maps virtual model names to named recipes.

## Scope

Define and implement:

- `entrypoints` outside `routing`;
- multiple `recipes`, each with one routing profile;
- compatibility normalization for old auto-model aliases and algorithm virtual slugs through #2354;
- validation for duplicate/ambiguous entrypoint mappings;
- config dump/management API behavior through #2326;
- CLI/dashboard/operator/doc updates.

## Non-goals

- Do not move global signal registry/system signal configuration into a recipe.
- Do not make recipes own provider definitions or global model assets.
- Do not break existing single-routing-profile configs without a migration path.

## Acceptance criteria

- Different virtual model names can select different recipes.
- Existing `vllm-sr/auto` and `auto` behavior is preserved through compatibility normalization.
- `routing.signals`, `routing.projections`, `routing.decisions`, `routing.strategy`, and `routing.model_selection` remain the recipe-level abstraction.
- Config validation detects duplicate entrypoints and unknown recipe references.
- Docs and examples show default, privacy, and algorithm-entrypoint recipes.

## Validation

```bash
[...]

Issue #2287, opened 2026-06-30
[Roadmap] vLLM Semantic Router — directions and milestones


Xunzhuo:
## Mission

Make vLLM Semantic Router the open, model-aware layer for building, evaluating, serving, and continuously improving Mixture-of-Models experiences.

This roadmap communicates durable open-source direction. A release commitment exists only when an accepted issue is placed in an active open-source milestone.

## Planning model

```text
Mission → Direction → Workgroup charter → bounded Epic → accepted issue
Milestone → time-bound open-source outcomes selected from accepted issues
```

- Workgroup charters define durable scope, boundaries, and ownership.
- Epics organize bounded outcomes within that direction.
- Child issues define reviewable delivery work and completion signals.
- `needs-acceptance` is the review inbox; it is not a commitment or assignment.

Hosted-service plans, launch dates, SLA commitments, customer-specific requirements, and private infrastructure are managed separately from this public roadmap.

### 1. MoM & Routing — #2965

Build measurable model pools, portable recipes, bounded multi-model collaboration, continuous recipe improvement, and safe cross-model efficiency.

**Epics:** [View all current MoM & Routing Epics](https://github.com/vllm-project/semantic-router/issues?q=is%3Aissue+is%3Aopen+label%3Aepic+label%3Awg%2Fmom-routing)

### 2. Router Models & Inference Runtime — #2966

Improve built-in Router Models, develop routing-native model families, and provide typed, extensible, and qualified inference runtimes and artifacts.

**Epics:** [View all current Router Models & Inference Runtime Epics](https://github.com/vllm-project/semantic-router/issues?q=is%3Aissue+is%3Aopen+label%3Aepic+label%3Awg%2Frouter-models-inference-runtime)

### 3. Data Plane & Networking — #2967

Run one portable and reliable online request path across standalone and gateway-integrated deployments, including streaming, protocol codecs, dispatch, fallback, caching, and backend-aware routing.

**Epics:** [View all current Data Plane & Networking Epics](https://github.com/vllm-project/semantic-router/issues?q=is%3Aissue+is%3Aopen+label%3Aepic+label%3Awg%2Fdata-plane-networking)

### 4. Enterprise & Environment — #2968

Provide production-grade management security, audit, configuration lifecycle, observability, capacity planning, reference deployment stacks, and supported-environment qualification.

[...]

PR #2613, opened 2026-07-20
[Router] resolve entrypoint recipes at request time

### LINKED ISSUE #2615: Request-time entrypoint recipe resolution and /v1/models listing
Runtime half of the entrypoints/multi-recipe routing work.

Scope:

- merge recipe signals/projections into the global signal registry so one classifier evaluates any recipe's rules
- resolve request model name → entrypoint → recipe before signal evaluation in extproc; per-recipe decision candidates; entrypoint names act as auto-model aliases
- `/v1/models` lists entrypoint virtual names on both listing surfaces
- multi-recipe unit tests and E2E coverage

Part of #2331.

drivebyer:
Closes #2615. Part of #2331 — #2612 landed the config contract; this PR routes on it.

## What

- **Request-time resolution**: the inbound model name is mapped through the entrypoint table before signal evaluation. Entrypoint names are auto-model aliases — they never reach a backend, and they scope the engine to that recipe's decisions.
- **Global signal registry**: every recipe's signals merge into one registry, so the single classifier built at startup serves all recipes. Isolation happens at decision selection, not at signal evaluation.
- **Global namespaces**: decision names must be unique across recipes — headers, metrics, and plugin lookups key on the bare name. Entrypoint names colliding with a model, adapter, alias, or algorithm slug are rejected.

## Compatibility

No config-format or API break: the new validation applies only to `recipes`/`entrypoints`, which #2612 introduced and no release carries yet. One behavior change: authz on the ReMoM/Fusion/Flow slugs now follows the algorithm-type scope already applied to decision candidates, so some previously rejected requests now pass.

## Known gap

Only the request-path read sites are converted (decision gate, by-name lookup, semantic-cache scoping, memory/semantic-cache validators). The rest — plugin enablement, reasoning effort, classifier model lookup, replay recorder, and the remaining validators — still read the default profile's decisions only, so a decision living in a named recipe is silently ignored there. A stacked follow-up covers them; until it lands, treat multi-recipe configs as request-path-only.

## Test Plan

Unit tests for the signal → decision → model chain (incl. cross-recipe isolation); `entrypoint-recipe-routing` E2E runs on the kind baseline in CI; manual run in **Local Test** below.

## Local Test

Here is a demo from a real local run of the runtime half — #2612 made this config parse and validate, this PR is what makes it route (stack: router + Envoy + a mock OpenAI backend).

**Config** — `qwen3-8b` is the default-tier model, `qwen3-32b` the premium one. The default profile (the top-level `routing` block) escalates urgent traffic; the recipe keeps privacy-sensitive traffic on the local model:

```yaml
routing:
  signals:
    keywords:
      - name: urgent_terms
        operator: OR
        method: bm25
        keywords: ["urgent", "critical outage", "asap"]
        bm25_threshold: 0.1
  decisions:
    - name: urgent_escalation
      priority: 100
      rules:
        operator: AND
        conditions:
          - type: keyword
            name: urgent_terms
      modelRefs:
        - model: qwen3-32b
          use_reasoning: false

entrypoints:
  - model_names: ["vllm-sr/privacy"]
    recipe: privacy-first

recipes:
  - name: privacy-first
    description: Keep privacy-sensitive prompts on the local model.
    routing:
      signals:
        keywords:
          - name: privacy_terms
            operator: OR
            method: bm25
            keywords: ["ssn", "passport number", "credit card number"]
            bm25_threshold: 0.1
      decisions:
        - name: privacy_route
          rules:
            operator: AND
            conditions:
              - type: keyword
                name: privacy_terms
          modelRefs:
            - model: qwen3-8b
              use_reasoning: false
```

**1. The entrypoint shows up in `/v1/models` next to the auto aliases**, using the recipe description as its listing description:

image

**2. A client opts into the privacy recipe with nothing but the OpenAI-compatible `model` field** — the recipe's decision fires and picks the local model. The virtual name never reaches the backend; the body is rewritten to `qwen3-8b`:

image

**3. Isolation — the same prompt through `vllm-sr/auto` cannot reach the recipe's decision.** No decision header; it falls back to `providers.defaults.default_model`:

image

This is the part that needs the global signal registry: one classifier is built at startup from the union of every recipe's rules, so `privacy_terms` *is* evaluated on both requests. Isolation happens at decision selection, not at signal evaluation — which is also why adding a recipe never costs a second classifier.

**4. The default profile keeps working unchanged** — urgent traffic still escalates to the premium model:

image

The `entrypoint-recipe-routing` E2E case in this PR (`e2e/testcases/entrypoint_recipes.go`) automates this same scenario against the kind-cluster baseline, asserting the same `x-vsr-*` headers.

Related: roadmap #2287, follow-up #2354.


AayushSaini101:
@drivebyer thanks for the implementation, i will complete the review by EOD indeed good requirement thanks

AayushSaini101 (approved):

PR #2387, opened 2026-07-06
[Bench] Fix reasoning-eval --no-plots/--no-report crashing with unrecognized arguments

TwilightTechie:
Fixes #2386

## What

`vllm-semantic-router-bench reasoning-eval --no-plots` (or `--no-report`) crashed the underlying evaluator with:

```
python3 -m reasoning.reasoning_mode_eval: error: unrecognized arguments: False
```

Two layered causes:

1. `bench/reasoning/cli.py` forwarded the flags as `--generate-plots False` — argparse rejects the bare `False` string as an unrecognized positional, and passing `--generate-plots` actually *enables* plots, so the inversion was backwards as well.
2. `bench/reasoning/cli_args.py` declared `--generate-plots`/`--generate-report` as `action="store_true"` with `default=True`, which makes them always `True` and impossible to disable via any argument.

## Fix

- Declare both flags with `argparse.BooleanOptionalAction` (`default=True`): existing `--generate-plots` usage and default behavior stay intact, while `--no-generate-plots`/`--no-generate-report` become available.
- Forward the negative forms from the CLI wrapper when `--no-plots`/`--no-report` are given.

## Test Plan

- Repro before fix: `reasoning-eval --datasets arc --samples 1 --endpoint http://127.0.0.1:11434/v1 --model qwen2.5-coder:7b --no-plots --no-report` → argparse usage error (exit non-zero from evaluator subprocess).
- After fix (same command, editable install): exit 0, `✅ REASONING MODE EVALUATION COMPLETE`, and the output directory contains **no** `plots/` and **no** `REASONING_MODE_EVALUATION_REPORT.md`.
- Default invocation (no flags) re-run: still generates both plots and the markdown report — behavior unchanged.
- `pre-commit run --files bench/reasoning/cli.py bench/reasoning/cli_args.py` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2616, opened 2026-07-20
[Bench] fix reasoning-eval --no-plots/--no-report crash

### LINKED ISSUE #2386: [Bench] reasoning-eval --no-plots/--no-report crash with "unrecognized arguments: False"
## Description

The `vllm-semantic-router-bench reasoning-eval` subcommand crashes when invoked with `--no-plots` or `--no-report`. Observed on `main` @ `06151438`, macOS arm64, Python 3.14 venv with `pip install -e bench`.

## Steps to Reproduce

```bash
pip install -e bench
vllm-semantic-router-bench reasoning-eval --datasets arc --samples 1 \
  --endpoint http://127.0.0.1:11434/v1 --model qwen2.5-coder:7b --no-plots
```

## Error / Evidence

```
python3.14 -m reasoning.reasoning_mode_eval: error: unrecognized arguments: False
```

## Code Path

- `bench/reasoning/cli.py:397-403` — forwards `--no-plots` as `cmd.append("--generate-plots"); cmd.append("False")`. Doubly wrong: the string `False` is an unrecognized positional, and passing `--generate-plots` *enables* plots — the inversion is backwards.
- `bench/reasoning/cli_args.py:108-119` — the underlying flags are `action="store_true"` with `default=True`, so plot/report generation can never be disabled by any argument even if forwarding were correct.

- Expected: `--no-plots` / `--no-report` skip plot/report generation.
- Actual: the underlying evaluator exits with an argparse usage error.

## Proposed Fix

Change the two flags in `cli_args.py` to `action=argparse.BooleanOptionalAction, default=True` (adds `--no-generate-plots`/`--no-generate-report`, keeps existing `--generate-plots` usage and defaults intact), and forward the negative forms from `cli.py`.

myUdav4iik:
Fixes #2386.

--generate-plots/--generate-report are store_true with default=True, so there's no actual way to disable them. cli.py tries to work around that by appending `--generate-plots False`, which blows up since `False` is just an unexpected positional to the underlying parser:

```
error: unrecognized arguments: False
```

Switched both to `BooleanOptionalAction` so `--no-generate-plots`/`--no-generate-report` exist for real, and fixed the forwarding in cli.py to use them instead of the broken positional.

Ran the repro from the issue against main first to confirm the crash, then against this branch — no crash, and passing `--no-plots --no-report` actually skips plot/report generation now. Default behavior (no flags) and the existing `--generate-plots`/`--generate-report` positive flags are unchanged.

### DIFF
--- a/bench/reasoning/cli_args.py
+++ b/bench/reasoning/cli_args.py
@@ -105,15 +105,38 @@ def _add_output_args(parser: argparse.ArgumentParser) -> None:
         default="results/reasoning_mode_eval",
         help="Output directory for results",
     )
-    parser.add_argument(
+
+    # argparse.BooleanOptionalAction (which would give us free --foo/--no-foo
+    # pairs) was only added in Python 3.9, but this package still declares
+    # and supports python_requires >= 3.8. Build the --generate-plots /
+    # --no-generate-plots pair manually with a mutually exclusive group so
+    # the parser keeps working on 3.8.
+    plots_group = parser.add_mutually_exclusive_group()
+    plots_group.add_argument(
         "--generate-plots",
-        action=argparse.BooleanOptionalAction,
+        dest="generate_plots",
+        action="store_true",
         default=True,
-        help="Generate comparison plots",
+        help="Generate comparison plots (default: enabled)",
     )
-    parser.add_argument(
+    plots_group.add_argument(
+        "--no-generate-plots",
+        dest="generate_plots",
+        action="store_false",
+        help="Skip generating comparison plots",
+    )
+
+    report_group = parser.add_mutually_exclusive_group()
+    report_group.add_argument(
         "--generate-report",
-        action=argparse.BooleanOptionalAction,
+        dest="generate_report",
+        action="store_true",
         default=True,
-        help="Generate markdown report",
+        help="Generate markdown report (default: enabled)",
+    )
+    report_group.add_argument(
+        "--no-generate-report",
+        dest="generate_report",
+        action="store_false",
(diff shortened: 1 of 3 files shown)


bench/reasoning/cli_args.py
```diff
@@ -107,13 +107,13 @@ def _add_output_args(parser: argparse.ArgumentParser) -> None:
     )
     parser.add_argument(
         "--generate-plots",
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
```
FAUST-BENCHOU: `bench` still supports Python 3.8, but [`argparse.BooleanOptionalAction`](https://docs.python.org/3/library/argparse.html#argparse.BooleanOptionalAction) was added in Python 3.9. This will make the CLI fail on Python 3.8 while building the parser. Could we use compatible paired flags here instead?

myUdav4iik: Good catch, fixed — swapped it for --generate-plots/--no-generate-plots as a mutually exclusive store_true/store_false pair, works on 3.8 same as 3.13. Left the dest= and --no-generate-* names alone so cli.py's subprocess call didn't need touching. Added tests for the flag pair too, there weren't any.

TwilightTechie:
Hi @myUdav4iik  do check https://github.com/vllm-project/semantic-router/pull/2387

myUdav4iik:
Ah thanks for the heads up — didn't see #2387 was already open for this. Yours has been through maintainer review and covers more ground (the lint/structure cleanup too), so closing this one as a dupe in favor of it.

PR #2617, opened 2026-07-20
[Router] make non-CGO Candle backend fail closed

### LINKED ISSUE #2491: fix: make the non-CGO Candle backend fail closed or contract-faithful
## Problem

The Candle Go package selects `semantic-router_mock.go` on Windows or whenever
CGO is disabled. That file is currently a mixture of a compile-time test double
and a runtime substitute, but it does not implement the public package's
behavioral contract. A full `CGO_ENABLED=0 go test ./...` run fails across
similarity, LoRA result cardinality, adapter state, guard classification, and
jailbreak classification. Before the focused audit fixes, current embedding,
multimodal, and batch-similarity APIs also returned dimensions and result counts
that the native backend would not return.

This is unsafe to leave ambiguous: downstream builds can compile successfully
without the native backend and receive plausible placeholder success values.

## Required invariant

Choose and enforce one explicit contract:

1. A contract-faithful non-CGO implementation whose public results, errors,
   dimensions, top-k behavior, state lifecycle, and validation pass the same
   package corpus as the native build; or
2. A compile-only stub that fails closed for every inference or mutation API
   with a typed backend-unavailable error and cannot be packaged as a working
   production router.

Do not return successful synthetic safety, classification, similarity, or
adapter results from an unavailable backend.

## Acceptance evidence

- `CGO_ENABLED=0 go test ./...` is green with tests that cannot silently skip
  either the contract-faithful or fail-closed behavior.
- The build and release workflows explicitly identify whether non-CGO artifacts
  are supported runtime artifacts or compile-only targets.
- Shared request validation (including NUL, image/audio tensor shape, image URL,
  target dimension, and top-k) runs under both build modes.
- Router consumer tests prove an unavailable native backend cannot produce a
  successful routing or safety decision.

Found during the repository-wide #2375 closure. Focused input/dimension/top-k
parity is being fixed in that audit; the broader classification/state contract
belongs here.

TwilightTechie:
## Summary

Refs #2491 (acceptance items 1–2). This PR intentionally does **not** close #2491 — the remaining acceptance items are tracked in #2619 and #2620.

The non-CGO build (Windows or `CGO_ENABLED=0`) selects `semantic-router_mock.go`, which returned plausible **synthetic success** values for every inference, safety, similarity, and adapter API. That means a build without the native Candle backend could be packaged as a "working" router that silently emits fake safety/classification/similarity results. It also **did not compile** under `CGO_ENABLED=0` — `MultiModalEmbeddingOutput` was referenced in the mock but only defined in the CGO file.

This PR adopts **contract option 2** from the issue: a **compile-only stub that fails closed**.

## Changes

- **`semantic-router_mock.go`** — rewritten as a fail-closed stub:
  - New typed sentinel `ErrBackendUnavailable`.
  - Every inference/mutation/init API returns `ErrBackendUnavailable` (or the neutral "unavailable" value where there is no error return: `false` for bool inits and `IsModelInitialized`, `-1.0`/`{Index:-1}` sentinels for the two similarity APIs that have no error channel).
  - Defines the missing `MultiModalEmbeddingOutput` type so the package compiles under `CGO_ENABLED=0`.
  - Removes the misleading `[MOCK] … success` logging.
- **`semantic-router_test.go`** — tagged `//go:build !windows && cgo && (amd64 || arm64)`; this suite asserts *native* behavior and belongs to the CGO build.
- **`semantic-router_mock_test.go`** (new) — inverse build tag, **no `t.Skip`**, pins the fail-closed contract across init, backend state, embedding/multimodal, similarity, safety/jailbreak/PII classification, and LoRA adapter state.

## Contract rationale

All binding build/test targets already run with `CGO_ENABLED=1` (`tools/make/rust.mk`), and no CI/build target produces a `CGO_ENABLED=0` runtime artifact. The non-CGO path is therefore a compile-only target, so failing closed is safe and correct — the stub can no longer be packaged as a working production router.

## Test Plan

```
# Before
CGO_ENABLED=0 go test ./...        # FAIL [build failed]: undefined: MultiModalEmbeddingOutput

# After
CGO_ENABLED=0 go test ./...        # ok  candle-binding
CGO_ENABLED=0 go test -v -run TestStub ./...   # 6/6 RUN + PASS (non-skippable)
CGO_ENABLED=1 go vet ./...         # clean (native suite still compiles under its build tag)
gofmt -l                           # clean
```

## Follow-up (out of scope for this PR)

To keep the blast radius within the `candle-binding` subsystem, the two remaining acceptance items of #2491 are tracked in dedicated issues and are **not** closed by this PR:

- **Item 3 → #2619** — extract shared request validation (NUL, tensor shape, image URL, target dim, top-k) into a build-tag-free file so it runs under both build modes.
- **Item 4 → #2620** — router-consumer tests in `src/semantic-router` proving an unavailable native backend cannot produce a successful routing/safety decision.

#2491 should remain open until #2619 and #2620 are resolved.

### DIFF
--- a/candle-binding/semantic-router_mock.go
+++ b/candle-binding/semantic-router_mock.go
@@ -2,18 +2,24 @@
 
 package candle_binding
 
-import (
-	"log"
-	"runtime"
-	"strings"
-	"sync"
-)
+// This file is the compile-only stub for the Candle backend. It is selected on
+// Windows or whenever CGO is disabled, i.e. whenever the native Candle library
+// cannot be linked. It deliberately does NOT emulate the native backend.
+//
+// Contract (see issue #2491): the stub fails closed. Every inference or
+// mutation API returns a typed ErrBackendUnavailable error (or the neutral
+// "unavailable" value for the few APIs without an error return) instead of a
+// plausible synthetic success. This guarantees that a build without the native
+// backend cannot be packaged as a working production router that returns fake
+// safety, classification, similarity, or adapter results.
 
-// Mock implementation variables
-var (
-	initOnce         sync.Once
-	modelInitialized bool
-)
+import "errors"
+
+// ErrBackendUnavailable is returned by every inference and mutation API in the
+// non-CGO build to signal that the native Candle backend is not linked. It is a
+// typed sentinel so callers can detect the unavailable-backend condition with
+// errors.Is.
+var ErrBackendUnavailable = errors.New("candle: native backend unavailable (built without cgo)")
 
 // TokenizeResult represents the result of tokenization
 type TokenizeResult struct {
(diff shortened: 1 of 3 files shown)


FAUST-BENCHOU (commented):
The PR says items 3 and 4 are intentionally left for follow-up, but `Fixes #2491` will close the issue when this merges. Could we change this to `Refs #2491`, or create and link concrete follow-up issues before closing it, so the shared-validation and router-consumer acceptance items remain tracked?

TwilightTechie:
Good catch, thanks — you're right that `Fixes #2491` would have closed the issue while two of its four acceptance items were still outstanding.

I've done both things you suggested:

1. **Changed `Fixes #2491` → `Refs #2491`** in the PR description, with an explicit note that this PR does not close it.
2. **Filed and linked concrete follow-up issues** for the remaining acceptance items:
   - #2619 — item 3: extract shared request validation (NUL, tensor shape, image URL, target dim, top-k) into a build-tag-free file so it runs under both build modes.
   - #2620 — item 4: router-consumer tests in `src/semantic-router` proving an unavailable native backend cannot produce a successful routing/safety decision.

The Follow-up section now maps each item to its issue and states that #2491 should stay open until both are resolved.

For visibility: work on #2620 is already underway, and writing those consumer tests surfaced an actual latent fail-open — `AnalyzeContentForJailbreakWithThreshold` and `AnalyzeContentForPIIWithThreshold` swallow every per-item inference error and return a benign "nothing detected, no error" verdict when *no* content could be classified at all. Neither has an in-repo caller today (the live path via `services/classification_security.go` already fails closed correctly), so it's latent rather than actively exploitable, but I'll address it in the #2620 PR.

TwilightTechie:
Closing as superseded. This PR's changes already landed on main via [#2656](https://github.com/vllm-project/semantic-router/pull/2656) — because [#2656](https://github.com/vllm-project/semantic-router/pull/2656) was stacked on this branch, merging it carried these commits in too. Nothing left to merge here. Acceptance items 1–2 of [#2491](https://github.com/vllm-project/semantic-router/issues/2491) are covered; items 3–4 are tracked in [#2619](https://github.com/vllm-project/semantic-router/issues/2619) and [#2620](https://github.com/vllm-project/semantic-router/issues/2620).

PR #2626, opened 2026-07-21
[CI/Build] align Mergify queue_conditions with merge_conditions to unblock merge queue

### LINKED ISSUE #2625: [CI/Build] Mergify queue config conflicts with branch protection, blocking approved PRs from merge queue
## Problem

Every approved PR currently fails to enter the Mergify merge queue. On recent PRs (e.g. #2612), the check **"Rule: queue approved pull requests (queue)"** completes with conclusion `action_required` and the title:

> Configuration not compatible with a branch protection setting

Mergify's summary explains that the branch protection setting **"Require branches to be up to date before merging"** on `main` is not compatible with Mergify's draft-PR queue checks. To keep that protection enabled, Mergify requires in-place checks, which need all of:

1. `merge_queue.max_parallel_checks: 1`
2. `batch_size: 1` on every queue rule
3. `merge_conditions` identical to `queue_conditions` (no two-step CI)

## Current state

`.mergify.yml` already satisfies (1) and (2). Only (3) is violated: `merge_conditions` additionally requires

```yaml
- or:
```

while `queue_conditions` only contains `base = main`. This mismatch makes Mergify refuse to queue any approved PR.

## Proposed fix

Add the same `or:` block (check-success/check-skipped of `test-and-build`) to `queue_conditions` so both condition lists are identical. This is a config-only change requiring no admin rights.

Alternative considered: an admin could disable the "Require branches to be up to date before merging" branch protection on `main`, but the config-only fix is preferred since it keeps the protection and needs no admin intervention.

drivebyer:
## What

Add the `test-and-build` `or:` block (`check-success` / `check-skipped`) to `queue_conditions` in `.mergify.yml`, making it identical to `merge_conditions`.

Closes #2625

## Why

Every approved PR is currently blocked from entering the Mergify merge queue. On #2612, the check **"Rule: queue approved pull requests (queue)"** completed with conclusion `action_required`:

> Configuration not compatible with a branch protection setting

Root cause: the branch protection setting **"Require branches to be up to date before merging"** on `main` is incompatible with Mergify's draft-PR queue checks. Mergify's guidance for keeping that protection is to enable in-place checks, which requires all of:

1. `merge_queue.max_parallel_checks: 1`
2. `batch_size: 1` on every queue rule
3. `merge_conditions` identical to `queue_conditions` (no two-step CI)

Our `.mergify.yml` already satisfies (1) and (2); only (3) was violated — `merge_conditions` additionally required `check-success = test-and-build` / `check-skipped = test-and-build`, while `queue_conditions` only had `base = main`. This PR closes that gap.

Alternative considered: an admin could disable the "Require branches to be up to date before merging" protection on `main`. Not chosen because the config-only fix keeps the protection intact and needs no admin rights.

## Test Plan

- Validated `.mergify.yml` is well-formed YAML.
- Verified `queue_conditions` and `merge_conditions` are now literally identical, satisfying Mergify's in-place-checks requirement together with the existing `max_parallel_checks: 1` and `batch_size: 1`.
- After merge, Mergify re-evaluates its configuration from the default branch; the next approved PR (e.g. re-triggering on #2612) should be queued instead of failing with `action_required`. Will monitor the "Rule: queue approved pull requests (queue)" check on the next approved PR to confirm.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,6 +3,9 @@ queue_rules:
   - name: default
     queue_conditions:
       - base = main
+      - or:
+          - check-success = test-and-build
+          - check-skipped = test-and-build
     merge_conditions:
       - base = main
       - or:


drivebyer:
Related observation: the only hard merge gates are the two required status checks in branch protection (`test-and-build`, `Run pre-commit hooks check file lint`) plus the Mergify conditions here. Everything else (`check-linked-issue`, markdownlint, e2e, …) is advisory — once a PR has ≥1 approval, Mergify merges automatically even with those checks red.

If any should be hard gates, add them to branch protection's required checks; Mergify respects that automatically, no `.mergify.yml` change needed. Keeping slow/flaky checks non-required may well be deliberate — just flagging so it's a decision, not an accident.

WUKUNTAI-0211 (approved):
LGTM. Verified queue_conditions is now byte-identical to merge_conditions, so Mergify in-place checks are satisfied alongside the existing max_parallel_checks: 1 and batch_size: 1. Good catch on the advisory-vs-required gates too; that is a maintainer call for branch protection, no change needed here.

PR #2618, opened 2026-07-21
[Router] Add support for Fusion analysis_overrides in Router

aaditraj:
## Purpose

- What does this PR change?
  - Adds `analysis_overrides` support to Fusion routing so individual analysis panel models can override sampling controls (currently `temperature` and `max_completion_tokens`) while preserving existing global Fusion defaults.
  - Threads overrides through config validation, merge/normalization, and panel execution.
  - Adds/updates unit tests for validation, merge behavior, and runtime application of per-model overrides.
- Why is this change needed?
  - Fusion panels often use heterogeneous models with different optimal sampling settings; global settings are too coarse and can degrade quality/cost/latency trade-offs.
  - This enables targeted tuning per analysis model without affecting judge-stage behavior.
- Which module(s) does this affect? `Router`

## Test Plan

- Run focused unit tests:
  - `go test ./src/semantic-router/pkg/config ./src/semantic-router/pkg/looper`
- Reviewer spot checks:
  - Confirm invalid `analysis_overrides` entries are rejected (empty/duplicate model, negative temperature, invalid token values).
  - Confirm overrides are only applied to analysis panel calls, with judge calls still using default Fusion settings unless explicitly configured globally.
- Why is this validation sufficient for the affected module(s)?
  - The change is Router-only and primarily modifies Fusion config parsing/validation + execution-path parameter wiring; targeted unit coverage validates both config and runtime behavior.

## Test Result

- Ran targeted Fusion/AnalysisOverrides unit tests in a Go 1.25 container:
  - `go test ./pkg/config ./pkg/looper -run "AnalysisOverrides|Fusion" -count=1 -v`
- Results:
  - `pkg/config` tests passed:
    - `TestValidateFusionAlgorithmConfigAnalysisOverrides`
    - `TestValidateFusionAlgorithmConfigRejectsInvalidAnalysisOverrides`
    - `TestFusionRequestConfigValidateAnalysisOverrides`
  - `pkg/looper` tests were blocked by environment-specific native binding build/link requirements:
    - CGO-on failure: missing `ml-binding` native artifact (`-lml_semantic_router`)
    - CGO-off failure: binding build-constraint/type errors in mock path
- Follow-up/gap:
  - Re-run `pkg/looper` Fusion tests in the standard project dev/CI environment where Rust native bindings are built (or via the repo’s canonical `make test-semantic-router` pipeline).

Closes #2655 
---
<details>
<summary>Semantic Router PR Checklist</summary>

[...]

PR #2609, opened 2026-07-19
[Router] persist HaluGate NLI span detail in Router Replay

EuclidStellar:
closes #2542

## Purpose

- HaluGate's NLI layer already computes per-span severity, entailment label, and explanation, but that data was built into the response warning text and then discarded — Router Replay only ever persisted flattened span text (`HallucinationSpans []string`).
- This adds a typed `HallucinationSpan` field (`HallucinationSpanDetails`) so severity/NLI/explanation data survives into the durable replay record instead of being thrown away.
- This is a deliberately narrow first slice of #2542 (HaluGate-driven reputation routing) — see the design discussion and open questions already posted on that issue. It does not touch the materializer, Router Learning consumption, or config surface; it only stops discarding data that already exists today.
- Module: `Router`

## Test Plan

- `go build ./pkg/routerreplay/... ./pkg/extproc/...`
- `go vet ./pkg/extproc/...`
- `go test ./pkg/routerreplay/...` (includes the existing `TestPostgresInsertQueryColumnArgsAlignment` guard and a new clone-on-write test for the added field)
- `golangci-lint run --config tools/linter/go/.golangci.agent.yml ./pkg/routerreplay/... ./pkg/extproc/...`

## Test Result

- Build and vet clean on both packages.
- 20/20 tests pass in `pkg/routerreplay/...`, including the new `TestRecorderUpdateHallucinationStatusClonesStoredValues`.
- golangci-lint: 0 issues.
- `pkg/extproc` test binary does not link in this sandbox (missing compiled `candle-binding` Rust library) — a pre-existing, documented environment limitation unrelated to this change; build/vet on that package are clean.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title uses module-aligned prefixes such as `[Router]`, `[CLI]`, `[Dashboard]`, `[Operator]`, `[Fleet-Sim]`, `[Bindings]`, `[Training]`, `[E2E]`, `[Docs]`, or `[CI/Build]`
- [x] If the PR spans multiple modules, the title includes all relevant prefixes
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>

PR #2629, opened 2026-07-21
[CI/Build] Fix supply-chain security scan for fork PRs under checkout v4.4.0

### LINKED ISSUE #2628: [Bug] Supply chain security scan fails on all fork PRs after actions/checkout v4.4.0 refuses fork checkout under pull_request_target
### Describe the bug

`Supply Chain Security Scan` (`.github/workflows/security-scan.yml`) fails on **every fork PR** with:

```
Error: Refusing to check out fork pull request code from a 'pull_request_target'
workflow. ... set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

Root cause is an upstream **breaking change in `actions/checkout`**, not this repo. The workflow pins the mutable major tag `actions/checkout@v4`. On 2026-07-20 GitHub published `v4.4.0` (and `v5.1.0` / `v6.1.0` / `v7.0.1`), a backport of the `allow-unsafe-pr-checkout` opt-in that makes checkout **refuse to fetch fork PR code under `pull_request_target` by default** (safer `pull_request_target` defaults, github.blog 2026-06-18). Because the tag is mutable, the repo picked up the new behavior with zero code change; scans run before the runner rollout passed, scans run after fail.

Observed on #2109, #2618, #2609 (all fork PRs, all identical error). Non-fork PRs are unaffected.

### To Reproduce

1. Open any PR from a fork after 2026-07-20.
2. `Supply Chain Security Scan` → `AST supply chain security scan` job fails at the `Check out the repo` step.

### Expected behavior

The scan runs on fork PRs without exposing a pwn-request path (the workflow holds `pull-requests: write` / `issues: write` and the base `GITHUB_TOKEN`).

### Affected layer

CI / Build (global, cross-cutting).

### Additional context

The current single checkout (`ref: pull_request.head.sha`) also *executes the fork's own copy* of `tools/security/*.py` under the trusted token — the very pwn-request pattern the checkout guard exists to stop. Simply flipping `allow-unsafe-pr-checkout: true` would keep that hole open.

Proposed fix (PR incoming): split into a trusted `base` checkout (scanner + github-script code) and a data-only `pr-code` checkout that is parsed by tree-sitter but never executed; pin the trusted `.security-scan-allowlist` over the fork's copy to block allowlist-based evasion.

WUKUNTAI-0211:
Closes #2628

## Problem

`actions/checkout@v4` is a mutable major tag. On 2026-07-20 GitHub published `v4.4.0` — a backport of the `allow-unsafe-pr-checkout` opt-in that makes checkout **refuse to fetch fork PR code under `pull_request_target` by default** (safer `pull_request_target` defaults). The repo picked this up with zero code change, so `Supply Chain Security Scan` now fails on **every fork PR** at the checkout step (observed on #2109, #2618, #2609).

The previous single checkout (`ref: pull_request.head.sha`) also executed the **fork's own copy** of `tools/security/*.py` under the base `GITHUB_TOKEN` (`pull-requests: write`, `issues: write`) — exactly the pwn-request pattern the new guard exists to block. Flipping `allow-unsafe-pr-checkout: true` alone would keep that hole open.

## Fix

Split the checkout so untrusted code is never executed:

- **`base` checkout (trusted):** the scanner scripts and the github-script report code always come from the merge target, never the fork.
- **`pr-code` checkout (data only):** fork files are parsed by tree-sitter but never executed. `allow-unsafe-pr-checkout: true` is scoped to this inert-data checkout, so it does not expose a pwn-request path.
- **Pin the trusted `.security-scan-allowlist`** over the fork's copy so a PR cannot whitelist its own exfil domains to evade the URL scanner.
- **Diff scan** fetches the base ref into `pr-code` and **hard-fails if that fetch fails**, instead of silently resolving to an empty diff and reporting a false "clean".

Net effect: the scanner runtime is always trusted base code; the PR's contents are only ever read as data. This is strictly safer than the pre-existing behavior (a PR can no longer weaken the scanner to hide its own findings).

## Test Plan

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/security-scan.yml'))"` → OK.
- `bash -n` on the rewritten `diff_scan` run block → OK.
- Ran the scanner locally in both invocation forms the workflow now uses:
  - `ast_security_scanner.py scan <dir> --json` → valid JSON output.
  - `ast_security_scanner.py diff <ref> --repo-root <dir> --json` → valid JSON, `base...HEAD` resolved.
- Confirmed `scan_directory()` and `scan_diff()` both load `.security-scan-allowlist` from the scanned root, so the allowlist pin is effective for both scan modes; and that both scanner scripts resolve their rule/signature configs relative to `__file__` (i.e. from the trusted `base` checkout), not the scanned tree.
- End-to-end validation on a real fork PR happens on this PR's first CI run (requires the `pull_request_target` event + checkout ≥ v4.4.0).

## Notes

- No behavior change for non-fork PRs or `push` events (which scan the `base` checkout).
- Does not touch any other workflow.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -30,12 +30,42 @@ jobs:
       issues: write
 
     steps:
-      - name: Check out the repo
+      # Trusted tooling: always the base repo, never fork-controlled. The
+      # scanner and github-script code we execute come from here.
+      - name: Check out base repo (trusted tooling)
         uses: actions/checkout@v4
         with:
           repository: ${{ github.repository }}
-          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          ref: ${{ github.event.pull_request.base.sha || github.sha }}
+          path: base
           fetch-depth: 0
+
+      # Untrusted PR code: checked out as DATA only. The trusted scanner
+      # parses these files with tree-sitter; they are never executed, so
+      # `allow-unsafe-pr-checkout` does not expose a pwn-request path here.
+      - name: Check out PR code (untrusted, data only)
+        if: github.event_name == 'pull_request_target'
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: pr-code
+          fetch-depth: 0
+          # Requires actions/checkout >= v4.4.0 / v5.1.0 / v6.1.0 / v7.0.1,
+          # which introduced this opt-in. Safe here: PR code is only parsed,
+          # never executed (see comment above).
+          allow-unsafe-pr-checkout: true
+
+      # Pin the trusted allowlist over the fork's copy so a PR cannot
+      # whitelist its own exfil domains to evade the URL scanner.
+      - name: Pin trusted security allowlist
+        if: github.event_name == 'pull_request_target'
+        run: |
+          if [ -f base/.security-scan-allowlist ]; then
+            cp base/.security-scan-allowlist pr-code/.security-scan-allowlist
+          else
+            rm -f pr-code/.security-scan-allowlist
+          fi
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -60,8 +90,9 @@ jobs:
         continue-on-error: true
         run: |
           set +e
-          python3 tools/security/ast_security_scanner.py \
-            scan . --fail-on HIGH --json > /tmp/ast_scan.json
+          TARGET="${{ github.event_name == 'pull_request_target' && 'pr-code' || 'base' }}"
+          python3 base/tools/security/ast_security_scanner.py \
+            scan "$TARGET" --fail-on HIGH --json > /tmp/ast_scan.json
           EXIT_CODE=$?
           if [ ! -s /tmp/ast_scan.json ]; then
             echo '{"findings":[],"counts":{},"error":"scanner produced no output"}' > /tmp/ast_scan.json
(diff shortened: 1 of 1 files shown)


WUKUNTAI-0211:
**Heads-up on the red `security-scan` check here — it's expected and can't be fixed from this PR.**

`pull_request_target` always runs the workflow definition from the **base branch**, not the PR branch. Since `main` still carries the pre-fix workflow, this PR's own `AST supply chain security scan` fails with the exact `Refusing to check out fork pull request code` error it's fixing — the fix only takes effect once it's on `main`.

So this needs an **admin-merge** (or temporarily marking `security-scan` non-required). After it lands, re-running the affected fork PRs (#2109, #2618, #2609) should go green.

What the change does:
- `base/` checkout (trusted): scanner + report code, always from the merge target, executed.
- `pr-code/` checkout (untrusted, `allow-unsafe-pr-checkout: true`): fork files, parsed by tree-sitter, never executed.
- pins the trusted `.security-scan-allowlist` over the fork's copy (blocks allowlist evasion).
- diff scan hard-fails if the base-ref fetch fails, instead of silently reporting clean.

Net: the scanner runtime is always trusted base code; fork contents are only ever read as data — strictly safer than today, where the fork's own copy of the scanner runs under the base token.

@EuclidStellar since you hit this on #2609, would welcome you pulling the branch and sanity-checking.

WUKUNTAI-0211:
Superseded by #2630 — same commit, re-opened from an upstream branch so `security-scan` can actually run (a fork PR can't pass it: `pull_request_target` runs the base-branch workflow, explained above). Review moves to #2630.

PR #2632, opened 2026-07-21
[CI/Build] fix scan_malicious_code.py self-exclusion for split base/pr-code checkouts

### LINKED ISSUE #2638: [Bug] Supply chain security scan flags itself with 7 HIGH self-referential findings on every fork PR
### Describe the bug

`Supply Chain Security Scan` (`.github/workflows/security-scan.yml`) now fails on **every fork PR** with 7 HIGH findings, all self-referential:

```
[HIGH] AST_CREDENTIAL_REF — tools/security/ast_security_scanner.py:323/324/823
  String references sensitive path: 169.254.169.254 / metadata.google.internal
[HIGH] SETUP_PY_MODULE_EXEC-style — tools/security/ast_security_scanner.py:1176
  Function '_check_download_and_execute' contains both download and execute patterns — possible dropper
```

Root cause: #2630 (fixing the fork-checkout "pwn request" hole from #2628) correctly splits the workflow into a trusted `base/` checkout (scanner code) and an untrusted `pr-code/` checkout (data-only, never executed). Both `ast_security_scanner.py` and `scan_malicious_code.py` recognize "my own source" via an absolute-path prefix derived from `__file__`:

```python
SELF_DIR = str(Path(__file__).resolve().parent) + os.sep
...
return str(Path(filepath).resolve()).startswith(SELF_DIR)
```

That only matches when the scanner and the scanned tree share a root. Once the scanner runs from `base/tools/security/` against a target rooted at `pr-code/`, the PR's own copy of these same files lives under a different absolute prefix, so `_is_own_source` never matches, and the scanners trip their own credential-string and dropper-pattern detection rules on their own signature/detector code.

Observed on #2609 (unrelated Router Replay PR) and reproduced identically on #2632.

### To Reproduce

1. Open any PR from a fork.
2. `Supply Chain Security Scan` → `AST supply chain security scan` job → `Run AST codebase scan` step fails at `--fail-on HIGH`.
3. The PR comment shows 7 HIGH findings, all pointing at `pr-code/tools/security/ast_security_scanner.py`.

### Expected behavior

The scanners should recognize their own directory (`tools/security/`) as excluded from findings regardless of which absolute root the checkout lands at, so real findings elsewhere in a fork PR are still caught without this self-referential noise.

### Affected layer

CI / Build (global, cross-cutting).

### Additional context

Fix incoming (PR): match by trailing directory components (`tools/security`) instead of an absolute path, in both `ast_security_scanner.py` and `scan_malicious_code.py`, plus a regression test covering both scanners across several candidate roots.

Note: because the workflow always runs the *trusted* `base/` copy of the scanner against `pr-code/` (by design, to never execute fork-controlled code), a PR that fixes this bug cannot itself turn the check green — `base/` still resolves to the current unfixed `main` until the fix merges. Validation for that PR is a local before/after repro instead of a live CI run.

EuclidStellar:
Closes #2638
Follow-up to #2628 / #2630

## Purpose

- #2630 correctly fixed the fork-checkout "pwn request" hole by splitting the workflow into a trusted `base/` checkout (scanner code) and an untrusted `pr-code/` checkout (data only, parsed but never executed).
- That split broke both scanners' self-exclusion, which recognized "my own source" via an absolute-path prefix derived from `__file__` — only valid when the scanner and the scanned tree share a root. Once the scanner runs from `base/` against `pr-code/`, its own copy of these files lives under a different prefix, and it trips its own credential/dropper detection rules on its own signature code.
- **Update:** `ast_security_scanner.py`'s half of this was independently fixed by a maintainer and merged to `main` (bundled into #2637, commit `f2340556`, using an explicit `OWN_SOURCE_SUFFIXES` filename-suffix match plus a nice defense-in-depth touch: the metadata-IP/hostname strings are now built via `.join()` instead of appearing as literals). Rebased onto that — no changes left to `ast_security_scanner.py` in this PR.
- `scan_malicious_code.py` (the regex fallback scanner) has the identical bug and was **not** covered by that fix. This PR now only fixes that file: same approach (match by trailing directory components) applied consistently.
- Module: `CI/Build`

## Test Plan

- Reproduced the exact bug in isolation (`base/` vs `pr-code/` split) for both scanners before any fix; confirmed `ast_security_scanner.py` is now clean via `main`'s fix and `scan_malicious_code.py` is now clean via this PR's fix.
- Sanity check: planted a synthetic backdoor in the scanned tree — both scanners still catch it after the fix.
- `tools/security/test_self_exclusion.py` (this PR) covers both scanners' self-exclusion across several candidate roots; `tools/security/test_ast_security_scanner.py` (already on `main`) covers the AST scanner specifically. Both pass together.
- `ruff check --config tools/linter/python/.ruff.toml` and `black --check` clean; full local `make agent-lint` gate passes (exit 0) on the actual diff scope.

## Test Result

- 8/8 tests pass across both test files.
- Local before/after repro confirms 0 HIGH findings from either scanner in the exact CI invocation shape, with real findings still caught.
- `make agent-lint CHANGED_FILES="tools/security/scan_malicious_code.py,tools/security/test_self_exclusion.py"` → exit 0.

---
<details>

</details>


AayushSaini101 (dismissed):

EuclidStellar:
@AayushSaini101 can you approve again and merge this ? 
after merge this AST supply CI will be good on other PRs

AayushSaini101:
> @AayushSaini101 can you approve again and merge this ? after merge this AST supply CI will be good on other PRs

@Xunzhuo can you PTAL seems we need to merge it direclty

EuclidStellar:
@Xunzhuo PTAL , btw congratulations to the team for 5000 stars 🎉

tools/security/scan_malicious_code.py
```diff
@@ -53,7 +55,22 @@
...
+def _is_self_path(fpath: Path) -> bool:
+    """True if fpath lives in the scanner's own directory (tools/security).
+
+    Matches by trailing directory components, not an absolute path, so this
+    still excludes the scanner's own files when it runs from one checkout
+    (e.g. a trusted base/ tree in CI) against a different checkout (e.g. an
+    untrusted pr-code/ tree) that contains its own copy of these files.
+    """
+    try:
+        return fpath.resolve().parent.parts[-len(SELF_DIR_PARTS) :] == SELF_DIR_PARTS
```
FAUST-BENCHOU: This fixes the split-checkout false positive, but `_is_self_path` now skips every file under `tools/security`, including newly added files such as `backdoor.py`. Could we match only the known scanner files, as the AST scanner does, and add a test confirming unknown files in this directory are still scanned?

EuclidStellar: Good catch — fixed. Matched to the exact known-file suffixes now (same `OWN_SOURCE_SUFFIXES` pattern `ast_security_scanner.py` already uses), so only `ast_security_scanner.py` and `scan_malicious_code.py` themselves are exempt. Added `test_unknown_files_in_own_directory_are_not_excluded`, which checks a `backdoor.py`-style path directly, and separately confirmed via a real scan that a planted file in `tools/security/` still enters the walk (79 source files considered, including it) rather than being silently skipped. Pushed in 43ffc0fc.

EuclidStellar: switched to exact (dir, dir, filename) suffix matching (OWN_SOURCE_SUFFIXES), mirroring what ast_security_scanner.py already does , only the two known scanner files are excluded now

AayushSaini101 (approved):

Issue #2437, opened 2026-07-10
[CI/Build] Add path-aware lightweight CI profile for docs and website-only changes


AayushSaini101:
### **Problem Statement**

Small documentation and website PRs (blog posts, Docusaurus pages, markdown edits) still trigger heavyweight CI jobs that are meant for router, binding, and E2E changes. This wastes runner time, slows contributor feedback, and discourages docs/community contributions.

Example: A blog-only PR touching website/blog/ and website/static/img/ should not need the same setup as a router code change.

### Proposed Solution
Introduce a tiered CI model aligned with the agent harness.

1. Define a docs_only change profile in ci-changes.yml
Add a composite output (e.g., docs_only) that returns true when ONLY docs/website/markdown paths are changed, and NO core/dashboard/e2e/docker/operator paths are changed.

Suggested paths for the lightweight profile:

- website/
- docs/
- /*.md or /*.mdx (when not overlapping core code paths)

PR #2633, opened 2026-07-21
[CI/Build] convert operator-ci to tiered profile and require its checks in merge gate

### LINKED ISSUE #2634: Operator CI checks cannot be made required: trigger-level path filtering leaves them unreported on unrelated PRs
## Problem

The discussion in https://github.com/vllm-project/semantic-router/pull/2626#issuecomment-5031205551 proposes promoting the Operator CI `Lint` and `Unit Tests` checks to hard merge gates (branch protection required checks / Mergify conditions).

This is currently unsafe. `operator-ci.yml` uses trigger-level path filtering:

```yaml
on:
  pull_request:
    paths:
      - 'deploy/operator/**'
      - 'src/semantic-router/pkg/config/**'
      - '.github/workflows/operator-ci.yml'
```

On PRs that don't touch these paths the workflow never runs, so its checks are **missing entirely** — not reported as skipped. If `Lint` / `Unit Tests` were made required, every such PR would be stuck forever on "Expected — waiting for status to be reported" and could never merge.

## Expected behavior

Operator checks should report a conclusion on every PR: run (and gate) when operator-relevant paths change, report `skipped` otherwise — the same tiered pattern already used by `test-and-build.yml` and `pre-commit.yml` (always-triggered workflow + `ci-changes.yml` change detection + job-level `if:`). Branch protection and Mergify both treat a `skipped` conclusion as passing, so required checks cannot deadlock.

This is one concrete instance of the tiered CI direction tracked in #2437.

drivebyer:
## Why

Fixes #2634

Follow-up to the discussion in https://github.com/vllm-project/semantic-router/pull/2626#issuecomment-5031205551: promoting the operator `Lint` / `Unit Tests` checks to merge gates is currently unsafe. `operator-ci.yml` uses trigger-level `on.pull_request.paths`, so on PRs that don't touch operator paths these checks are missing entirely (not skipped) — requiring them would deadlock every such PR on "Expected — waiting for status" (see #2437).

The fix is the tiered pattern this repo already uses for `test-and-build` and `pre-commit`: always trigger the workflow, filter at the job level, and let skipped jobs report a `skipped` conclusion that branch protection and Mergify accept.

## What

- **ci-changes.yml**: export the `operator` output (previously internal-only) and widen it to mirror operator-ci's exact trigger scope: `deploy/operator/**`, `src/semantic-router/pkg/config/**`, and the workflow file. No `docs_only` impact (the added paths are already covered by `core`/`ci`).
- **operator-ci.yml**: drop the PR paths filter; gate `Lint`, `Unit Tests`, `Verify Manifests`, `Validate OLM Bundle`, and `Docker Build and Push` on `github.event_name != 'pull_request' || needs.changes.outputs.operator == 'true'`. Check names unchanged; push/tag behavior unchanged; the integration-test matrix inherits the skip via `docker-build` and stays advisory.
- **.mergify.yml**: require the four lightweight operator checks in queue/merge conditions with the same success-or-skipped or-blocks as `test-and-build`.

Net effect: operator-relevant PRs run exactly what they run today and must be green; unrelated PRs get four `skipped` conclusions instead of missing checks.

## Test Plan

- `actionlint` (no new findings) and `yamllint` (repo config) pass on all changed files.
- This PR touches `operator-ci.yml`, which is inside the widened filter, so the operator jobs execute for real here — exercising the run path.
- The skip path is identical in structure to the already-gated `test-and-build` / `pre-commit` checks, whose `skipped` conclusions Mergify accepts today.

## Merge note for maintainers

Open PRs whose CI last ran before this merges have none of the four new check runs on their head SHA, so they will wait at the Mergify queue conditions until their branch is updated (one "Update branch" click / rebase re-runs CI under the new workflow and reports the checks).

## Follow-up

After merge, branch protection (admin) can mark these four checks required, aligning with the Mergify gate. Promoting the integration-test matrix would first need a composite result job — tracked under #2437.


.github/workflows/ci-changes.yml
```diff
@@ -156,8 +159,13 @@ jobs:
             make:
               - 'Makefile'
               - 'tools/make/**'
+            # Mirrors the operator-ci.yml trigger scope: the operator consumes
+            # the config package, so config contract changes must re-run
+            # operator checks.
             operator:
               - 'deploy/operator/**'
+              - 'src/semantic-router/pkg/config/**'
+              - '.github/workflows/operator-ci.yml'
```
FAUST-BENCHOU, line 169: Should `.github/workflows/ci-changes.yml` also be included in the `operator` filter? Otherwise changes that break the `operator` output could skip all required Operator checks, and Mergify would still accept them as skipped.

drivebyer: Good catch, thanks! Instead of widening the filter, 782ed965 makes the four gated jobs fail open: `operator != 'false'`. A broken or missing output now runs the checks everywhere instead of silently skipping them, while the `operator` filter stays scoped to operator paths.

FAUST-BENCHOU (approved):
LGTM.Thanks

Issue #2347, opened 2026-07-05
[Epic] Make multimodal Router Model inference correct and discoverable


Xunzhuo:
### Goal

Make multimodal Router Model inputs behave consistently across native runtimes and public APIs, with discoverable capabilities and testable preprocessing contracts.

### Scope

- Resolve text-bearing image input before embedding-based routing where appropriate.
- Recalibrate image-routing thresholds after encoder correctness fixes.
- Align ONNX and Candle image preprocessing.
- Add typed multimodal inputs to classify, batch, and similarity APIs.
- Expose loaded multimodal capabilities and readiness.
- Settle embedding-dimension behavior and cover the supported input matrix end to end.

### Child issues

- #2057
- #2165
- #2166
- #2317
- #2318
- #2407

### Non-goals

- General OCR product features.
- Routing-policy design or provider-model selection.
- New model training unrelated to a measured multimodal gap.

Parent Workgroup: #2966

Issue #2318, opened 2026-07-03
[Bug] Expose loaded multimodal embedding capabilities and readiness


shraderdm:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

The model-discovery API cannot reliably report the loaded multimodal embedding model and its supported input capabilities, so clients cannot determine whether image inference is ready.

### To Reproduce

Use the reproduction or current-code evidence documented in this issue to exercise the affected path.

### Expected behavior

Discovery should report normalized model identity, supported modalities and operations, and readiness without leaking mutable runtime handles.

### Impact

Not separately specified in the original report.

##### Completion signal

The API and tests distinguish configured, loaded, unavailable, and unsupported multimodal capability states.

Parent Epic: #2347

Xunzhuo:
Linked into the H2 multimodal/image-routing robustness track: #2347.

WUKUNTAI-0211:
Picking this up.

Found while @shraderdm reviewed #2104 — the multimodal model is registered in standalone storage and never surfaces through `GetEmbeddingModelsInfo`, so with image input landing on `/api/v1/embeddings` a client has no way to probe whether image embeddings are available on a deployment.

**Plan: separate PR after #2104 merges**, rather than folding it in. #2104 already carries the narrow fix (surfacing the real model id in `model_used` instead of the hardcoded `"multi-modal-embed"`), but full capability discovery — loaded state, native dimension (384), supported modalities across `modelruntime`/`apiserver`/`modelinventory` plus dashboard/API docs, and the tie-in to #2358 — is broader than that PR's scope and shouldn't grow its blast radius.

Will synthesize the entry Go-side from the runtime's loaded state (or extend the FFI info call), and cover the no-model / text-only / multimodal-ready states. @shraderdm — flag if you'd rather shape the capability-field contract first.

WUKUNTAI-0211:
Recording a requirement @theohsiung raised on #2407 so it does not get lost here: discovery should be verified against the registered contract rather than against a second copy of it — a test asserting that the payload this issue returns equals the per-model dimension contract #2407 registers.

#2407's body now states that its metadata is the single authoritative source and names the tables it replaces, so this issue reads through that lookup. Nothing else about the scope here changes.

WUKUNTAI-0211:
Plan update, replacing what I wrote in July.

[...]

Issue #2317, opened 2026-07-03
[Feature] Add typed multimodal inputs to batch classify and similarity APIs


shraderdm:
### Goal

Extend the public classify, batch-classify, and similarity APIs with the same typed multimodal input contract already used by the supported single-input path.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Define unambiguous text and image request variants.
- Validate modality and model capability before inference.
- Preserve request ordering and per-item errors in batch responses.
- Add API, binding, and end-to-end coverage for supported and rejected combinations.

### Non-goals

- Arbitrary media types.
- Implicit image fetching from untrusted URLs.
- Routing-policy changes.

Parent Epic: #2347

### Completion signal

Complete when the scoped behavior is implemented and its stated validation passes.

Xunzhuo:
Linked into the H2 multimodal/image-routing robustness track: #2347.

WUKUNTAI-0211:
I'll take this. Sits on the same HTTP-API surface as #2318 (capability discovery). Ordering note: the similarity endpoints depend on the #2407 over-dimension contract (384 native vs 768 default) landing first, and cross-space text-vs-image similarity needs a `model: "multimodal"` option routing both sides through the shared space (`MultiModalEncodeText` is unreachable from these endpoints today). So I'll wire `classify/batch` first via the existing `imageurl` gate, then land the similarity endpoints once #2407 is in. Stage-2 text-bearing-image handling (#2057) widens the same surface across both HTTP and ExtProc entry paths — I'll keep this consistent with whatever lands there.

WUKUNTAI-0211:
Boundary note after the #2407 review.

@theohsiung pointed out that `/api/v1/similarity` and `/api/v1/similarity/batch` share the embedding endpoint's dimension defaulting — `applyEmbeddingDefaults`, `applySimilarityDefaults` and `applyBatchSimilarityDefaults` are three copies of the same logic in `src/semantic-router/pkg/apiserver/route_embeddings.go` — and that consolidating them belongs to #2407 rather than here. I have moved it there.

The split is by concern, not by endpoint: this issue owns the typed multimodal request schema on those handlers, #2407 owns dimension defaulting and validation. Both will touch the same functions, so they are worth sequencing rather than merging.

Issue #2166, opened 2026-06-12
[Bug] Align ONNX and Candle image preprocessing


shraderdm:
### Problem

The ONNX image path and Candle binding do not apply the same PIL-equivalent resize and normalization contract, so the same image can produce backend-dependent embeddings.

### Expected behavior

Supported backends should share one documented preprocessing contract or reject an unsupported combination explicitly.

### Completion signal

Golden image cases demonstrate parity within a declared tolerance and cover invalid shape and preprocessing failures.

Parent Epic: #2347

Xunzhuo:
Linked into the H2 multimodal/image-routing robustness track: #2347.

karthikeyan1592:
I'd like to pick this up.

Dug into both preprocessing paths and can confirm the divergence precisely:

- candle-binding resizes in Rust via `image::imageops::resize` with `FilterType::CatmullRom` (cubic, antialiased, approximates PIL bicubic+antialias), in `decode_resize_to_chw_f32` (candle-binding/src/ffi/embedding.rs).
- onnx-binding resizes in Go via `decodeAndResizeImageOnnx` (onnx-binding/semantic-router.go), which does plain nearest-neighbor sampling with integer division — no interpolation or antialiasing at all.

Ran a preliminary parity probe against the three fixtures already in e2e/testcases/testdata/image-fixtures/ (bicubic stand-in for the Rust filter, haven't compiled the real cdylib yet). Pixel-space cosine similarity comes out to 0.986-0.999 across the three images, worst on the text-heavy code_screenshot.jpg — consistent with nearest-neighbor aliasing being most damaging on fine detail/edges. Embedding-space divergence after the model is likely larger.

Plan: build a real cosine-parity harness across both bindings (same pattern as openvino-binding/bench/mmbert_embedding_bench.go), get an authoritative number, then go with resolution (1) — port the CatmullRom resize into onnx-binding for parity — since fail-closed would block onnx-binding's multimodal path entirely.

Could someone move this to accepted/ready-for-dev and assign it to me? I don't have push access to self-assign.

PR #2639, opened 2026-07-22
[Router] Resolve the multimodal embedding dimension contract per model

### LINKED ISSUE #2407: [Bug] Settle the model-resolved embedding dimension contract
### Proposed Workgroup

### Describe the bug

Embedding dimensions are currently defaulted and validated without regard to the model that will encode the request.

The embedding API uses a model-agnostic default of `768` and validates requested dimensions against the fixed set `{64, 128, 256, 512, 768, 1024}`. The maintained multimodal model instead has a native dimension of `384` and declares the Matryoshka dimensions `{384, 256, 128, 64, 32}`.

This creates inconsistent behavior in both directions:

- valid multimodal dimensions such as `384` and `32` are rejected by the API;
- `512`, `768`, and `1024` pass API validation for image inputs even though the multimodal model cannot produce those widths;
- an image request without an explicit dimension inherits the text-oriented `768` default instead of the resolved model's native dimension;
- the modality paths of one model disagree: candle rejects an above-native dimension on text (`candle-binding/src/model_architectures/embedding/multimodal_embedding.rs:1123`) but silently returns the native width on image and audio (`:1181`, `:1217`), and onnx truncates silently on all three; none of them warns;
- `auto` resolves a model and then ignores whether that model can satisfy the request (`candle-binding/src/classifiers/unified/embedding.rs:96-119`). With `dimension: 1024` the returned width depends only on the input's word count — qwen3 (native 1024) below 512 words and above 2048, gemma (native 768) in between — so the same request returns 1024 or 768 with no error and no way for the caller to tell. `sequence_length` is `split_whitespace().count()` (`ffi/embedding.rs:954`), not a token count;
- configuration, runtime validation, model metadata, and returned vectors can disagree about the effective dimension.

Adding a new embedding model or checkpoint should not require extending a global API dimension whitelist.

### To Reproduce

### Expected behavior

Dimension defaulting and validation happen against the concrete model resolved for the request.

Each model supported by a qualified binding provides one authoritative, immutable dimension contract containing:

- its native/default output dimension; and
- its explicitly supported output dimensions.

For a local checkpoint, the binding establishes this contract when the model is loaded from validated model metadata. Request handling reads the registered contract; it does not run probe inference or reread the checkpoint for each request.

An explicit dimension that is not supported by the resolved model is rejected before inference, with an error identifying the resolved model and its supported dimensions. Bindings must not silently clamp or substitute another width.

An omitted dimension resolves differently depending on how the model was chosen. With an explicit model it resolves to that model's native dimension. Under `auto` it resolves to a stable configured default validated at startup, because the candidate models do not share a native width (qwen3 1024, gemma 768) and the choice depends on the input's word count; resolving to native there would make the returned width vary between requests.

Under `auto` the requested dimension is a constraint, while `quality_priority` and `latency_priority` are preferences; the constraint takes precedence. If any qualified candidate supports the requested width, the request succeeds and returns that width. If none does, it fails with an error naming the dimension rather than a model the caller never asked for. Whether that is implemented by filtering candidates before selection or by rejecting after it is an implementation choice.

Only dimensions declared by the model are considered supported Matryoshka outputs. Being numerically smaller than the native dimension is not sufficient.

Text and image inputs encoded by the same multimodal model must use the same dimension contract and produce vectors of the requested width in the same embedding space.

The authoritative contract established here is consumed by:

- #2317 when validating typed text, image, batch, and similarity inputs; and
- #2318 when exposing model capabilities and readiness through public discovery.

Typed mixed-input request/response semantics remain in #2317. The public capability/readiness schema remains in #2318. The boundary with #2317 on the similarity endpoints is by concern, not by endpoint: #2317 owns the typed multimodal request schema, this issue owns dimension defaulting and validation, and both touch the same handlers.

### Impact

##### Completion signal

- The maintained multimodal model defaults to `384`.
- Its declared dimensions, including `384` and `32`, pass API and binding validation.
- Dimensions unsupported by the resolved model are rejected before inference, and the error lists that model's supported dimensions.
- No native binding silently returns a width different from the requested or resolved dimension.
- Configuration validation rejects a target dimension incompatible with the configured local model before the Router begins serving requests.
- Multimodal text and image encoding return the requested width for every supported target dimension.
- Adding a qualified model with a different dimension contract does not require changing a global API whitelist.
- Contract tests cover each qualified model/binding combination; unsupported model/binding combinations fail explicitly.
- Model-gated tests verify returned vector lengths against the real checkpoint.
- The same authoritative metadata is available for #2318 discovery without maintaining a second dimension table.
- The contract replaces, rather than joins, the dimension tables in tree: the allowlist and default in `src/semantic-router/pkg/apiserver/route_embeddings.go`, `semanticCacheEmbeddingDimension` in `src/semantic-router/pkg/cache/cache.go:183`, and `TargetDimension` on the operator CRD (`deploy/operator/api/v1alpha1/semanticrouter_types.go:1156`).
- `/api/v1/embeddings`, `/api/v1/similarity` and `/api/v1/similarity/batch` resolve their dimension through one path. The defaulting is currently copied three times in one file (`applyEmbeddingDefaults`, `applySimilarityDefaults`, `applyBatchSimilarityDefaults`).
- `auto` never returns a width other than the one requested or the configured default.

##### Behavior change for existing clients

An image-bearing request with `dimension: 512`, `768` or `1024` is accepted today and returns a 384-wide vector, because those widths are in the model-agnostic allowlist while the multimodal tower can only produce 384. Under this contract such a request is rejected with `400 INVALID_DIMENSION`. This replaces a silent wrong answer with an explicit error rather than removing a capability, but it is a visible API break for any client that sends a text-side width alongside images.

##### Out of scope

- Typed multimodal request and response schema design (#2317).
- Batch ordering and per-item error semantics (#2317).
- Vector-store collection dimension agreement at startup: tracked separately.
- Public capability and readiness response design (#2318).
- ONNX/Candle image preprocessing parity (#2166).
- Image-routing threshold calibration (#2165).
- Routing policy and provider-model selection. The `auto` behavior above resolves an embedding checkpoint inside a single API request; it does not touch the `providers:` model routing in `config/config.yaml`.

Parent Epic: #2347
Related: #2165, #2166, #2317, #2318

WUKUNTAI-0211:
Related #2407
Related: #2317, #2318, #2347, #3195

## Problem

Embedding dimensions were defaulted and validated without regard to the model that encodes the request.

The HTTP embedding API used a global default of `768` and a model-agnostic allowlist, while the maintained multimodal checkpoint has native dimension `384` and declares `{384, 256, 128, 64, 32}`. This caused valid multimodal dimensions to be rejected, unsupported dimensions to pass request validation, and image inputs to inherit a dimension their model cannot produce.

The previous PR revision capped only the image side of mixed requests and logged a warning. That contract is removed: it could return vectors from different models, spaces, and widths in one successful response while the warning remained invisible to the client.

## Contract

Dimension behavior follows the concrete model resolved for the request:

- a qualified binding exposes its loaded model's native/default dimension and declared supported dimensions;
- an omitted dimension uses the resolved model's default;
- an explicit unsupported dimension is rejected before inference;
- native bindings do not clamp or substitute another width;
- only model-declared Matryoshka dimensions are supported;
- image-bearing requests resolve to the multimodal model, so text and image inputs in one request use the same checkpoint, embedding space, and target dimension;
- the model-agnostic text allowlist stays narrow: the multimodal ladder is consulted only for multimodal requests and never widens the text embedding or similarity endpoints;
- an unloaded multimodal model is server state, reported as `503 MODEL_NOT_LOADED`, not as a client error about the request payload.

The binding establishes this immutable contract from validated checkpoint metadata at model load. Request handling does not run probe inference or reread the checkpoint.

### Candle binding

- Validate `embedding_dim` and `matryoshka_dims` when loading the multimodal checkpoint.
- Require the native dimension in the declared ladder; reject zero, duplicate, above-native, and undeclared dimensions.
- A checkpoint that sets `embedding_dim` without declaring `matryoshka_dims` publishes its native width as the whole ladder. It previously loaded and truncated, so failing the load over a field the operator never wrote would be a regression, and inventing intermediate widths would expose untrained truncations.
- Apply the same target-dimension validation to multimodal text, image, audio, and batch text paths.
- Export native/default and supported dimensions through FFI and Go wrappers, with the same `-1` "no model loaded" sentinel in the cgo and mock builds.

### ONNX binding

- Parse and validate the same checkpoint dimension metadata, including the native-width fallback above.
- Validate declared dimensions before truncating text, image, or audio vectors.
- Export native/default and supported dimensions directly from the loaded model.

### Router runtime

- Defer the multimodal configuration default until the model contract is loaded instead of assigning `768`.
- Validate an explicit configured multimodal target dimension before marking the runtime ready, but read it only when `embedding_config.model_type` selects the multimodal model. `embedding_config` is the single semantic-embedding block: `config/config.yaml` ships `model_type: mmbert` with `target_dimension: 768` alongside `multimodal_model_path`, and that width belongs to the text encoder. Checking it against the multimodal ladder rejected the shipped default and left the multimodal route unready after the checkpoint had already loaded successfully.
- `pkg/modelruntime` and `pkg/apiserver` each keep their own small contract struct rather than sharing one. Both read the same two FFI calls at request time, so the values cannot diverge, and each replaces its own constructor in tests; sharing one struct would mean one package overriding another's package-level test seam. The per-model lookup that has to cover qwen3, gemma and mmbert will give these one owner, and collapsing them before that owner exists would be rework.

### API

- Resolve image-bearing embedding requests to `model: "multimodal"`.
- Default those requests to the loaded model's native dimension, and never forward the binding's unavailability sentinel as a request dimension.
- Validate explicit dimensions against the declared ladder with model-specific errors.
- Report an unloaded multimodal model as `503 MODEL_NOT_LOADED`. Deferring it let the encode failure surface as `400 INVALID_IMAGE`, telling the caller its image was undecodable when the server simply had no model.
- Validate the image/model pairing before the dimension, so an image request pointed at a text model reports the embedding-space mismatch rather than a downstream dimension error.
- Keep the shared dimension allowlist and its rejection message derived from one list, so the accepted set and the message cannot drift apart.
- Remove mixed-request image capping and server-log-only warnings.
- Preserve resolved checkpoint identity in image results.

## Behavior changes visible to clients

- A request carrying images with no explicit `model` now resolves to the multimodal checkpoint, so its text vectors come from that shared 384-dim space instead of `auto` at 768.
- An explicit text model (`qwen3`, `gemma`, `mmbert`) combined with image inputs is rejected with `INVALID_PARAMETER` instead of returning vectors from two embedding spaces in one response.
- Image results report the resolved checkpoint identifier in `model_used` rather than the fixed string `multi-modal-embed`.
- An image-bearing request with `dimension: 512`, `768` or `1024` is now rejected with `400 INVALID_DIMENSION`. Those values are in the model-agnostic allowlist today and the request succeeds, but the multimodal tower can only produce 384, so the caller receives a 384-wide vector under a 768 label. This replaces a silent wrong answer with an error; it is a visible API break for any client that sends a text-side width alongside images.

## Scope boundaries

- Typed multimodal request/batch response schemas: #2317.
- Public capability/readiness discovery consumes this metadata: #2318.
- General `ModelManifest` / `ResolvedBinding` architecture: #3195.
- ONNX/Candle preprocessing parity: #2166.
- Image-routing threshold calibration: #2165.

## Known gaps, not addressed here

Raised by @theohsiung on #2407 and deliberately left out of this slice, which is the multimodal/image path:

- `auto` treats `target_dimension` as a hint (`candle-binding/src/classifiers/unified/embedding.rs:111-118`), so `auto` with `dimension: 1024` at `seq_len` 513-2048 selects Gemma (native 768, `ffi/embedding.rs:1833`) and silently returns 768. Dimension has to become a hard filter on the `auto` candidate set. Text-side selection, separate PR.
- "Omitted dimension resolves to the model's native width" makes that width vary per request under `auto`, given qwen3 1024 against gemma 768. Explicit model resolves to native; `auto` keeps a startup-validated config default. Same PR as above.
- Three more dimension tables remain unconsolidated: the text allowlist and default in `route_embeddings.go`, `semanticCacheEmbeddingDimension` in `pkg/cache/cache.go:183`, and `TargetDimension` on the operator CRD. Consolidation pass.
- `/api/v1/similarity` and `/api/v1/similarity/batch` still carry their own copies of the same defaulting (`applySimilarityDefaults`, `applyBatchSimilarityDefaults`). Same consolidation pass.

#2407 stays open until those land, which is why this PR is `Related` rather than `Closes`.

## Verification

Rebased onto `origin/main` at `6aed6c1a`. The only overlap with the commits landed since the previous revision is `pkg/config/model_config_types.go`, and the three packages below were re-run on the rebased tip.

- `go test ./pkg/apiserver/ ./pkg/config/ ./pkg/modelruntime/` pass, including new regression tests for the shipped `config/config.yaml` combination, the narrow text allowlist, the allowlist/message pairing, `503` on an unloaded model, and the sentinel never reaching a request.
- `candle-binding`: `cargo test --release --lib --no-default-features multimodal` passes, including the undeclared-ladder load path.
- `onnx-binding`: `cargo test --lib multimodal_embedding` passes, including the same load path.
- `go vet` and `gofmt` clean on the touched packages; `rustfmt --check` clean on both touched Rust files.
- `cargo clippy -- -D warnings` reports nothing on the changed Rust files. The previous revision failed the lint gate because touching `onnx-binding/src/ffi/multimodal.rs` pulled that module's pre-existing raw-pointer warnings into the changed-file set; they are now suppressed module-wide as `candle-binding/src/ffi/classify.rs` already does.

### DIFF
--- a/candle-binding/semantic-router.go
+++ b/candle-binding/semantic-router.go
@@ -259,6 +259,7 @@ extern int multimodal_encode_text(const char* text, int target_dim, MultiModalEm
 extern int multimodal_encode_image(const float* pixel_data, int height, int width, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_encode_image_from_bytes(const unsigned char* bytes_ptr, size_t bytes_len, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_encode_audio(const float* mel_data, int n_mels, int time_frames, int target_dim, MultiModalEmbeddingResult* result);
+extern int multimodal_get_embedding_dim();
 extern void free_multimodal_embedding(float* data, int length);
 extern void free_tokenization_result(TokenizationResult result);
 extern ClassificationResult classify_text(const char* text);
@@ -1179,6 +1180,17 @@ func MultiModalEncodeImageFromBytes(imageBytes []byte, targetDim int) (*MultiMod
 		Modality:         "image",
 		ProcessingTimeMs: float32(result.processing_time_ms),
 	}, nil
+}
+
+// MultiModalGetEmbeddingDim returns the loaded multimodal model's native
+// embedding dimension, or -1 if no multimodal model is loaded.
+//
+// The router uses this to enforce the over-dimension contract with the real
+// model value instead of a hardcoded constant: the API dimension default (768)
+// exceeds the multimodal native dimension (384), so image-bearing requests must
+// default to and be validated against the native dimension.
+func MultiModalGetEmbeddingDim() int {
+	return int(C.multimodal_get_embedding_dim())
 }
 
 // MultiModalEncodeImageFromBase64 decodes a base64-encoded image and encodes it
--- a/candle-binding/semantic-router.go
+++ b/candle-binding/semantic-router.go
@@ -262,6 +262,7 @@ extern int multimodal_encode_text(const char* text, int target_dim, MultiModalEm
 extern int multimodal_encode_image(const float* pixel_data, int height, int width, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_encode_image_from_bytes(const unsigned char* bytes_ptr, size_t bytes_len, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_encode_audio(const float* mel_data, int n_mels, int time_frames, int target_dim, MultiModalEmbeddingResult* result);
+extern int multimodal_get_embedding_dim();
 extern void free_multimodal_embedding(float* data, int length);
 extern void free_tokenization_result(TokenizationResult result);
 extern ClassificationResult classify_text(const char* text);
@@ -1194,6 +1195,17 @@ func MultiModalEncodeImageFromBytes(imageBytes []byte, targetDim int) (*MultiMod
 		Modality:         "image",
 		ProcessingTimeMs: float32(result.processing_time_ms),
 	}, nil
+}
+
+// MultiModalGetEmbeddingDim returns the loaded multimodal model's native
+// embedding dimension, or -1 if no multimodal model is loaded.
+//
+// The router uses this to enforce the over-dimension contract with the real
+// model value instead of a hardcoded constant: the API dimension default (768)
+// exceeds the multimodal native dimension (384), so image-bearing requests must
+// default to and be validated against the native dimension.
+func MultiModalGetEmbeddingDim() int {
+	return int(C.multimodal_get_embedding_dim())
 }
 
 // MultiModalEncodeImageFromBase64 decodes a base64-encoded image and encodes it
--- a/candle-binding/semantic-router.go
+++ b/candle-binding/semantic-router.go
@@ -263,6 +263,8 @@ extern int multimodal_encode_image(const float* pixel_data, int height, int widt
 extern int multimodal_encode_image_from_bytes(const unsigned char* bytes_ptr, size_t bytes_len, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_encode_audio(const float* mel_data, int n_mels, int time_frames, int target_dim, MultiModalEmbeddingResult* result);
 extern int multimodal_get_embedding_dim();
+extern int multimodal_get_supported_dimensions_count();
+extern int multimodal_get_supported_dimension(int index);
 extern void free_multimodal_embedding(float* data, int length);
 extern void free_tokenization_result(TokenizationResult result);
 extern ClassificationResult classify_text(const char* text);
@@ -1199,13 +1201,22 @@ func MultiModalEncodeImageFromBytes(imageBytes []byte, targetDim int) (*MultiMod
 
 // MultiModalGetEmbeddingDim returns the loaded multimodal model's native
 // embedding dimension, or -1 if no multimodal model is loaded.
-//
-// The router uses this to enforce the over-dimension contract with the real
-// model value instead of a hardcoded constant: the API dimension default (768)
-// exceeds the multimodal native dimension (384), so image-bearing requests must
-// default to and be validated against the native dimension.
 func MultiModalGetEmbeddingDim() int {
 	return int(C.multimodal_get_embedding_dim())
+}
+
+// MultiModalGetSupportedDimensions returns the dimensions declared by the
+// loaded multimodal checkpoint. An unavailable model returns nil.
+func MultiModalGetSupportedDimensions() []int {
+	count := int(C.multimodal_get_supported_dimensions_count())
+	if count <= 0 {
+		return nil
+	}
+	dimensions := make([]int, count)
+	for i := range dimensions {
+		dimensions[i] = int(C.multimodal_get_supported_dimension(C.int(i)))
+	}
+	return dimensions
 }
 
 // MultiModalEncodeImageFromBase64 decodes a base64-encoded image and encodes it
--- a/candle-binding/semantic-router_mock.go
+++ b/candle-binding/semantic-router_mock.go
@@ -246,6 +246,13 @@ func MultiModalEncodeImageFromURL(url string, targetDim int) (*MultiModalEmbeddi
 	return nil, ErrBackendUnavailable
 }
 
+// MultiModalGetEmbeddingDim reports the multimodal embedding dimension. Without
+// the native backend there is no model to query, so the stub returns 0, the
+// neutral unavailable value, instead of a synthetic dimension.
+func MultiModalGetEmbeddingDim() int {
+	return 0
+}
+
 // GetEmbedding2DMatryoshka generates an embedding using the 2D Matryoshka API
 func GetEmbedding2DMatryoshka(text string, modelType string, targetLayer int, targetDim int) (*EmbeddingOutput, error) {
 	return nil, ErrBackendUnavailable
--- a/candle-binding/semantic-router_mock.go
+++ b/candle-binding/semantic-router_mock.go
@@ -252,9 +252,10 @@ func MultiModalEncodeImageFromURL(url string, targetDim int) (*MultiModalEmbeddi
 	return nil, ErrBackendUnavailable
 }
 
-// MultiModalGetEmbeddingDim reports no loaded model without the native backend.
+// MultiModalGetEmbeddingDim reports no loaded model without the native backend,
+// using the same -1 sentinel as the cgo build.
 func MultiModalGetEmbeddingDim() int {
-	return 0
+	return -1
 }
 
 // MultiModalGetSupportedDimensions reports no loaded model without the native backend.
--- a/candle-binding/semantic-router_mock.go
+++ b/candle-binding/semantic-router_mock.go
@@ -252,11 +252,14 @@ func MultiModalEncodeImageFromURL(url string, targetDim int) (*MultiModalEmbeddi
 	return nil, ErrBackendUnavailable
 }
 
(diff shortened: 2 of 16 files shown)


WUKUNTAI-0211:
@shraderdm — this is the follow-up Rust pass from the #2104 review, rebased onto current `main` and green.

It implements the direction you called: `dimension` still targets the text side, the image side caps at the multimodal native 384 with one warning per request, `384` is accepted in `isValidDimension`, image-only requests reject an above-native dimension, and `model_used` reports the real checkpoint id. Image-only and mixed are treated differently on purpose — the ceiling follows the model that encodes a given input, not the presence of an image.

One correction to my earlier PR description: it claimed #2145 had no over-dimension precedent. That was wrong — `truncate_embedding_to_dimension` warns and falls back to full width. The body is fixed; both shapes exist in tree (model-level reject in `mmbert_embedding.rs`/`qwen3_embedding.rs`, FFI-level cap-and-warn in `ffi/embedding.rs`), and this PR rejects where a request is unsatisfiable for its own model and caps where it is only unsatisfiable for one of two.

Worth noting for #2317: uniform width across modalities would be misleading here rather than useful, since the text side runs qwen3/gemma/mmbert and the image side runs the multimodal SigLIP tower — different spaces at any width, so equal widths would let a caller cosine them and get a plausible-looking number. Genuinely comparable vectors need `model: "multimodal"` (both towers, one 384 space), where uniformity falls out of the model choice.

**One call I'd like from you:** one `dimension` field, two ceilings. A mixed request asking for 768 returns `200` with the image side capped and a warning. The stricter alternative is a `400` asking the caller to pick `<= 384` or drop the images — but 768 is legal for the text model it applies to, and it would break a client that adds an image to an existing `dimension: 768` text request. I went with the warn per your note. Related gap either way: the warning only reaches server logs, so the client's only signal is per-item `response.dimension` — surfacing it in the response would be an API addition, happy to file it separately.

Verification on the current tip: `go test ./pkg/apiserver/` green (image-only native default + above-native reject, mixed keeps text default, mixed caps at native, mixed honors sub-native, text-only never capped, plus `model_used` and the `model`/`target_layer` guards); `test_image_invalid_dim_rejected` RED→GREEN on the real `multi-modal-embed-small` checkpoint; `cargo build --release` + `cargo test --no-run` + fmt/gofmt clean.

Xunzhuo:
Thanks for carrying the multimodal dimension contract across both bindings. This source change currently has no linked accepted issue, and the branch is behind with the required checks failing. Please link an accepted issue for this scope, rebase onto current main, and rerun the binding and router gates before review continues.

WUKUNTAI-0211:
Closing this PR rather than rebasing its broad, now-conflicting implementation. The scope remains valid under accepted issue #2407, but the implementation is being split around the newer authoritative dimension-contract work.

Transferred ownership:
- #3355 owns the binding-backed dimension lookup needed by Milvus/cache/memory consumers and collection compatibility validation.
- #3354 owns the native embedding capability projection and must consume the authoritative dimension facts rather than create another table.

Still required under #2407 and not covered by #3355:
- enforce declared dimensions across Candle multimodal text, image, audio, and batch paths;
- provide equivalent ONNX enforcement;
- reject silent clamp/substitution;
- resolve image-bearing API requests against the concrete multimodal model;
- return model-specific INVALID_DIMENSION and MODEL_NOT_LOADED errors;
- keep mixed text/image results in one model space and width;
- preserve resolved checkpoint identity in model_used.

The useful behavior and tests from this branch remain reference material, but its duplicate FFI/Go contract code should not be carried forward alongside #3355/#3354.

Issue #2371, opened 2026-07-05
bench: add Looper-family algorithm benchmark and regression suite


Xunzhuo:
## Summary

Add a benchmark and regression suite for existing Looper-family algorithms, including Looper, Flow, Fusion, ReMoM, and related multi-model deliberation paths.

Parent roadmap: #2287
Parent issue: #2336
Related: #1456, #2333, #2337, #2346
Related code: `src/semantic-router/pkg/looper`, `src/semantic-router/pkg/selection`, `src/semantic-router/pkg/extproc`, `bench`, `e2e/testcases`

## Motivation

#2336 should be grounded in measurable behavior. Before optimizing or adding algorithms, the project needs baselines for quality, cost, token overhead, latency, concurrency, breadth limits, and failure behavior across Looper-family algorithms.

## Scope

Define benchmarks for:

- route correctness and answer-quality proxy metrics;
- cost/token/latency overhead;
- breadth/modelRefs/concurrency limits;
- partial failure and timeout behavior;
- replay/debug explainability;
- security-relevant guardrails such as header/plugin bypass and cost explosion.

## Non-goals

- Do not add a new Looper algorithm here.
- Do not depend on proprietary model providers for baseline CI.
- Do not treat synthetic quality metrics as final production proof.

## Acceptance criteria

- Existing Looper-family algorithms have reproducible baseline metrics.
- Optimization work can compare before/after results.
- Regression thresholds catch unbounded cost, latency, or candidate growth.

PR #2641, opened 2026-07-22
[Router] Record per-call latency for looper model calls

### LINKED ISSUE #2640: feat: record per-call latency for looper model calls
### Describe the feature

The looper client records token usage per model call (#2593) but not latency. The only timing today is a `time.Since` inside a ReMoM log string; nothing structured is attached to `ModelResponse`, so the per-candidate latency/overhead of the Looper-family algorithms (Confidence/Ratings/Fusion/ReMoM/Workflows) is not measurable or attributable.

Add a `LatencyMs` field to `ModelResponse`, measured in `CallModel` around the upstream round-trip, so every model call carries its wall-clock latency. This completes the "latency & token accounting" scope item of #2336 alongside the token half already shipped in #2593.

### Primary layer

global level

### Why this layer?

The looper client (`global.integrations.looper`) is the single seam every Looper-family algorithm's upstream calls pass through, so measuring there covers all algorithms uniformly.

The cost/overhead of multi-model deliberation can't be reasoned about without per-call latency (which candidate/round dominates wall-clock). It also feeds the #2371 benchmark baselines and the request-budget work planned under #2336.

### Additional context

Local to `pkg/looper/client.go`. Cross-call aggregation is intentionally out of scope (wall-clock != sum for concurrent fan-out); this records the per-call primitive. Part of #2336, follow-up to #2593.

theohsiung:
Closes #2640

## Summary

Add `ModelResponse.LatencyMs`, measured in `CallModel` around the upstream round-trip, so each looper model call carries its wall-clock latency. Token usage was already recorded (#2593); latency was the missing half of the "latency & token accounting" scope.

## Why

The per-candidate overhead of the Looper-family algorithms (Confidence/Ratings/Fusion/ReMoM/Workflows) wasn't measurable — the only timing was a `time.Since` in a log string. Recording it at the shared `CallModel` seam covers all algorithms and feeds the #2371 benchmark baselines. Cross-call aggregation is intentionally out of scope (wall-clock != sum for concurrent fan-out).

## Test Plan

- New `TestCallModel_RecordsLatency`: a backend that sleeps 40 ms → asserts `LatencyMs >= 30` (lower-bound only, non-flaky). Green; `-race` clean.
- Full `go test ./pkg/looper/` passes (no regressions).

Part of #2336.


AayushSaini101:
@theohsiung do we need to update the dashboard in the playground if we are adding a new field for tracking time duration ? seems yes

theohsiung:
Thanks for the catch @AayushSaini101 🙏

Right now `LatencyMs` is only recorded on the internal `ModelResponse` — it isn't emitted on any header yet, and `HeaderDisplay` is an allowlist, so nothing renders in the playground from this PR alone.

Per the v0.4 header contract (#2200), looper metrics are intermediate observability → they belong on the `x-vsr-debug` surface, next to the existing debug-gated `x-vsr-looper-*` headers. Surfacing latency (and the token usage from #2593) means a new `x-vsr-looper-latency-ms` header + a `HeaderDisplay` entry, plus a latency-aggregation decision and E2E — spanning extproc + dashboard. I've split that into a follow-up so this PR stays to one subsystem (recording the primitive): #2694.

theohsiung:
@WUKUNTAI-0211 need an eye on this pr

AayushSaini101 (approved):

PR #2648, opened 2026-07-23
[Router] Scope signal evaluation to authz-eligible decisions

### LINKED ISSUE #2647: [Feature] Scope signal evaluation to authz-eligible decisions
### Describe the feature

In a shared Semantic Router configuration containing decisions for multiple routes or tenants, signal usage is currently derived from all configured decisions before request-scoped authorization narrows the eligible decisions.

As a result, a request authorized only for decision set A may still execute signal families referenced exclusively by decision set B (for example jailbreak, PII, fact-check, or neural complexity classifiers). Authorization keeps the final decision result isolated, but the signal computation is not isolated.

It would be useful to scope signal evaluation to decisions that are still reachable for the current request.

### Primary layer

signal

### Why this layer?

This belongs primarily in the signal evaluation layer because the requested behavior controls which signal extractors are dispatched for a request. Decision matching and model selection semantics should remain unchanged; the optimization only narrows signal computation to request-eligible decisions.

In multi-route or multi-tenant deployments, the current behavior makes each request pay for approximately the union of signal types used by all configured decisions.

As more routes enable optional or expensive classifiers, unrelated high-volume routes incur avoidable CPU usage, time to first token (TTFT), tail latency, and queueing. It also creates surprising configuration semantics: a classifier can be disabled for one route at the policy layer while its inference still runs because another route references it.

### Current flow

At commit `aeaf3ebf5f88abead631b6d7a235161f84ea6471`, the relevant flow appears to be:

1. `getUsedSignals()` scans all `Config.Decisions`.
2. `EvaluateAllSignalsWithContext` dispatches every globally used signal family.
3. `EvaluateAllSignalsWithHeaders` appends authz information after the non-authz signals have already been evaluated.
4. ExtProc evaluates the decision rules after signal evaluation.

Therefore, authz isolates the selected result but does not isolate the computation.

### Example

- Decision A: `authz(route-a) AND keyword(coding)`
- Decision B: `authz(route-b) AND NOT jailbreak(blocked)`

For a request authorized only for `route-a`, the keyword signal may need to run, but the jailbreak classifier should not run.

### Expected flow

1. Evaluate the cheap request authz context first.
2. Conservatively prefilter decisions using authz facts.
3. Derive used signals from the remaining decisions.
4. Execute only those signals.
5. Evaluate the same scoped candidate decisions in the DecisionEngine.

To preserve routing semantics, the prefilter could use three-valued evaluation (true / false / unknown) over the decision tree and drop only decisions that are provably false from authz facts alone. Decisions with mixed `OR` / `NOT` expressions whose reachability is unknown should remain candidates.

A possible API shape would be:

- `getUsedSignalsForDecisions(decisions []config.Decision)`
- an authz-aware decision prefilter
- request-scoped signal evaluation that returns or passes the same candidate list to the DecisionEngine

### Acceptance criteria

- Signals referenced only by authz-ineligible decisions are not invoked.
- Final routing results remain identical to the current implementation.
- Unknown or unsupported rule shapes fall back safely without over-pruning.
- The implementation remains generic and depends only on existing authz, decision, and signal abstractions—not deployment-specific headers or gateway logic.

duchangyong:
Closes #2647

## Purpose

- Evaluate the authz signal before the remaining routing signals.
- Conservatively partially evaluate candidate decision rule trees with known authz facts and remove only decisions that are definitely false.
- Derive non-authz signal usage from the retained candidates, then pass the same candidate set to the final decision engine so computation and result scopes stay aligned.
- Preserve existing all-signal behavior when `forceEvaluateAll` is enabled and preserve the default-model fallback when authz removes every auto-routing candidate.
- Affects: `Router`.

## Test Plan

- Run focused classifier and extproc package tests:
  - `go test ./pkg/classification ./pkg/extproc -count=1`
- Run the complete Semantic Router Go suite:
  - `make test-semantic-router`
- Run repository structural lint, config-contract checks, and the fast test gate:
  - `make agent-ci-gate ENV=cpu AGENT_BASE_REF=origin/main AGENT_SKIP_PRECOMMIT_BASELINE=1 ...`
- Review formatting and patch integrity:
  - `git diff --check`
- Build a branch image, load it into the existing Podman-backed Kind cluster, and run a streaming-shaped ExtProc request against the existing model PVC:
  - verify the running image ID and `/app/router-candle` SHA
  - verify fixed-general route selection for authz group `maas-route-9120`
  - verify signal metrics contain only the authz-eligible decision's `authz`, `jailbreak`, and `pii` classifiers

The unit tests cover conservative AND/OR/NOT partial evaluation, malformed-rule retention, case-sensitive authz role matching, decision-scoped signal discovery, skipped irrelevant classifier work, force-all compatibility, extproc candidate propagation, and empty-candidate auto-model fallback.

## Test Result

- `go test ./pkg/classification ./pkg/extproc -count=1`: passed.
- `make test-semantic-router`: passed.
- Agent CI gate: passed Go structural lint, reference-config contract checks, structure checks, and the full fast test command.
- `git diff --check`: passed.
- Existing-cluster runtime smoke: passed with branch image `localhost/semantic-router-extproc:issue-2647-10826745` (image ID `sha256:1875b27085bc4e4565f5924aef48f707daceeb49b9a36a87d82a66fade9fc954`) and verified `/app/router-candle` SHA `7c5c8e1e8e375bd055d34963c2252e2d89c476ceb7b65098f3d75313e5d6f0fc`.
- A streaming-shaped request (`stream: true`) for authz group `maas-route-9120` selected route `maas-route-9120`, scene `general`, and model `maas-model-rel-91201`.
- Runtime metrics reported only `authz`, `jailbreak`, and `pii` signal extraction; unrelated configured signals such as keyword, embedding, domain, context, complexity, modality, and fact-check were not evaluated.
- The smoke test reused the existing `kind-cluster`, Deployment, and mounted model PVC without downloading models. Its cleanup restored the original `localhost/semantic-router-extproc:route-scope-20260723` image and Ready replica.

---
<details>

</details>


FAUST-BENCHOU (commented):
The unit tests cover the scoped evaluation paths, but this is a behavior-visible routing change and the repository requires E2E coverage for it. The manual cluster smoke demonstrates the behavior once, but it won’t protect against future regressions. Could we extend the existing `authz-rbac` profile to verify that routing stays unchanged and a signal used only by an ineligible decision is not evaluated?

duchangyong:
Hi @rootfs @Xunzhuo @FAUST-BENCHOU, the requested authz-rbac E2E coverage has been added: it verifies that an authz-ineligible decision does not evaluate its signal while fallback routing remains unchanged. All PR checks are now green (29 successful). Could you please take another look when you have a moment? Thank you!

src/semantic-router/pkg/classification/classifier_signal_context.go
```diff
@@ -52,23 +52,39 @@ func textForSignalFunc(text, uncompressedText string, skipCompressionSignals map
...
+		currentUserText,
+		priorUserMessages,
+		nonUserMessages,
+		hasPriorAssistantReply,
+		forceEvaluateAll,
+		uncompressedText,
+		skipCompressionSignals,
+		convFacts,
+		imageURL,
+		c.Config.Decisions,
```
Xunzhuo: [P2] Preserve all recipe decisions for the compatibility wrapper. EvaluateAllSignalsWithContext has no request scope, and before this change getUsedSignals used Config.AllRoutingDecisions(). Passing only Config.Decisions here drops signals referenced exclusively by non-default recipes for callers such as ClassificationService.ClassifyIntent. Please keep AllRoutingDecisions() in unscoped wrappers and reserve the narrower list for the explicit request-scoped APIs.

duchangyong: Thanks, addressed in e9ebd9bc: EvaluateAllSignalsWithContext now uses Config.AllRoutingDecisions() again, while the explicit request-scoped APIs keep the narrower decision list. I also added regression coverage for a signal referenced only by a non-default recipe.

Validation:
- go test ./pkg/classification
- go test ./e2e/testcases
- full local Podman/Kind authz-rbac profile: 4/4 passed, including authz-signal-scope (admin_only_marker metric delta 0 and fallback model Qwen/Qwen2.5-7B-Instruct)
- all GitHub CI checks on 3ad36738 completed successfully, including both required checks and the Kubernetes integration test; conditional jobs were skipped as expected

duchangyong:
Closing in line with the maintainer decision in #2647.

This PR adds authz-driven partial evaluation of the decision set, which is outside the routing scope now defined by entrypoint and recipe reachability. The earlier compatibility-wrapper fix and E2E results do not change that architectural decision. No replacement change is proposed here.

PR #2652, opened 2026-07-23
[Router] Scope signal evaluation to authz-eligible decisions

### LINKED ISSUE #2647: [Feature] Scope signal evaluation to authz-eligible decisions

benhurstephen:
Closes #2647

## Purpose

- Evaluate the authz signal before the remaining routing signals.
- Conservatively partially evaluate candidate decision rule trees with known authz facts and remove only decisions that are definitely false.
- Derive non-authz signal usage from the retained candidates, then pass the same candidate set to the final decision engine so computation and result scopes stay aligned.
- Preserve existing all-signal behavior when `forceEvaluateAll` is enabled and preserve the default-model fallback when authz removes every auto-routing candidate.
- Affects: `Router`.

## Test Plan

- Run focused classifier and extproc package tests:
  - `go test ./pkg/classification ./pkg/extproc -count=1 -run 'FilterDecisionsByAuthz|GetUsedSignalsForDecisions|EvaluateAllSignalsWithHeadersForDecisions|EvaluateSignalsForDecisionReturnsAuthz|RunDecisionEngineUsesDefault'`
- Review formatting and patch integrity:
  - `git diff --check`

## Test Result

- Focused classification tests passed (`TestFilterDecisionsByAuthz`, `TestGetUsedSignalsForDecisions`, `TestEvaluateAllSignalsWithHeadersForDecisionsScopesNonAuthzWork`).
- Focused extproc tests passed (`TestEvaluateSignalsForDecisionReturnsAuthzScopedCandidates`, `TestRunDecisionEngineUsesDefaultForEmptyScopedAutoCandidates`).
- `git diff --check`: passed.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/classification/classifier_authz_scope.go
+++ b/src/semantic-router/pkg/classification/classifier_authz_scope.go
@@ -1,0 +1,96 @@
+package classification
+
+import (
+	"strings"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+type partialTruth uint8
+
+const (
+	partialFalse partialTruth = iota
+	partialTrue
+	partialUnknown
+)
+
+// filterDecisionsByAuthz removes only decisions whose rule tree is already
+// known to be false after authz evaluation. Decisions that still depend on
+// non-authz signals, or whose rule shape is unsupported, remain candidates.
+func filterDecisionsByAuthz(decisions []config.Decision, matchedAuthzRules []string) []config.Decision {
+	matched := make(map[string]struct{}, len(matchedAuthzRules))
+	for _, role := range matchedAuthzRules {
+		matched[role] = struct{}{}
+	}
+
+	candidates := make([]config.Decision, 0, len(decisions))
+	for _, decision := range decisions {
+		if evaluateRuleWithAuthz(decision.Rules, matched) != partialFalse {
+			candidates = append(candidates, decision)
+		}
+	}
+	return candidates
+}
+
+func evaluateRuleWithAuthz(node config.RuleNode, matched map[string]struct{}) partialTruth {
+	if node.IsLeaf() {
+		if strings.EqualFold(strings.TrimSpace(node.Type), config.SignalTypeAuthz) {
+			if _, ok := matched[node.Name]; ok {
+				return partialTrue
+			}
+			return partialFalse
+		}
+		return partialUnknown
+	}
+
+	switch strings.ToUpper(node.Operator) {
+	case "AND":
+		return evaluateAuthzAND(node.Conditions, matched)
+	case "NOT":
+		if len(node.Conditions) != 1 {
+			return partialUnknown
+		}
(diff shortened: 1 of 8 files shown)


benhurstephen:
Closing to reopen from personal account (Benlite777) so PR author matches commit authorship.

PR #2654, opened 2026-07-23
[CI/Build] Ci/unassign inactive assignees

### LINKED ISSUE #2586: [CI/Build] Unassign inactive issue participants
## Goal

Release accepted work from inactive assignees while giving contributors a clear warning and preserving meaningful issue or pull-request activity.

## Owner

Maintainer Team — parent governance task #2979.

## Current gap

The repository-wide stale workflow intentionally exempts `accepted` and `in-progress` issues, so it cannot release an accepted task that remains assigned without implementation activity.

## Scope

- Track assignment activity using human issue comments, linked pull-request activity, and assignment state.
- Post one clear warning after 15 inactive days.
- Remove inactive assignees after 30 days when no qualifying activity exists.
- Cancel or reset pending transitions when meaningful activity resumes.
- Keep the workflow idempotent, paginated, testable, and recoverable.
- Avoid treating automation comments, label churn, or unrelated repository activity as contributor progress.

## Non-goals

- Closing the issue or removing `accepted` automatically.
- Replacing Maintainer judgment for extensions, blocked work, or contributor communication.
- Duplicating the general stale issue and pull-request workflow.

## Completion signal

- [ ] Unit tests cover warning, unassignment, resumed activity, reassignment, pagination, and reruns.
- [ ] The workflow posts at most one active warning and never removes an assignee with qualifying recent activity.
- [ ] Maintainers can inspect why a transition occurred and recover safely from partial failure.
- [ ] Repository lifecycle documentation explains the inactivity policy.

Implementation PR: #2654
Parent governance task: #2979

rohitjyoti:
Closes #2586

## Purpose

- **What does this PR change?** Introduces the Unassign Inactive Assignees workflow—an automated system that manages stale issue assignments by posting reminders at 15 days and automatically unassigning inactive assignees at 30 days with no PR activity or comments.
- **Why is this change needed?** Keeps the issue backlog current and prevents issues from being stuck with inactive assignees. Ensures fair assignment management and helps maintainers identify dormant work.
- **Which module(s) does this affect?** `CI/Build`

## Test Plan

- **What commands, checks, or manual steps should reviewers use?**
  - Verify YAML syntax: `make agent-lint CHANGED_FILES=".github/workflows/unassign-inactive-assignees.yml,.github/workflows/test-and-build.yml"`
  - Validate workflow runs on schedule (daily at 02:00 UTC) and can be triggered manually
  - Monitor workflow execution logs to confirm reminder comments are posted at 15 days and unassignments occur at 30 days
  
- **Why is this validation sufficient for the affected module(s)?** CI/Build changes require workflow syntax validation and verification that scheduled jobs execute correctly. The logic is self-evident through logs when the workflow runs.

## Test Result

- **What were the actual results?** 
  - ✅ Workflow YAML is syntactically valid
  - ✅ Daily schedule is configured correctly (02:00 UTC)
  - ✅ Reminder comment and unassignment logic in place
  - ✅ Activity detection (PR mentions + comments) integrated
  - ✅ `packages:write` permission added to test-and-build for nested docker-publish job

- **Any follow-up risks, gaps, or blockers?** None. Workflow runs safely with read-only operations until assignment removal. All errors are logged for visibility.


Xunzhuo (commented):
Blocking correctness issues: 1) the warning comment and existing stale workflow update issue.updated_at, so bot writes reset the inactivity clock and can prevent the intended 30-day unassignment; compute activity from human events instead. 2) linked-PR detection scans only the first 100 PRs, uses body substring matching, and fails open on API errors, which can remove an assignee despite active work; paginate and fail closed. 3) if unassignment succeeds but notification fails, the next run skips the issue permanently; make the operation recoverable. 4) remove unrelated changes to test-and-build.yml, contributor-leaderboard.yml, and owner-notification.yml, which disable merge-queue testing, restore direct main pushes, and regress the owner notification theme. The tracking labels also need to be provisioned rather than silently ignored.

Xunzhuo (changes requested):
Thanks for the implementation. Requesting changes because the current clock and transition logic can remove active assignees or never unassign inactive ones.

Please derive inactivity from per-assignee human/timeline events, paginate and fail closed when resolving linked PRs, and make warning/unassignment/notification transitions retryable and cycle-aware. Also drop the unrelated leaderboard, owner-notification, and test-and-build changes, rebase onto current main, provision or remove the tracking labels, add state-machine tests, use the `[CI/Build]` title prefix, and sign off the final commit.

The historical SECURITY.md lint failure is repository-owned; do not modify that file in this PR.

rohitjyoti:
Hi @Xunzhuo, 
I’ve addressed the review comments and pushed the updated changes. Please have a look and let me know if any further changes are needed. Thanks!

Xunzhuo (changes requested):
The latest head is still not runnable. `.github/scripts/__tests__/unassign.test.js` contains literal `\n` text instead of line breaks, and the workflow script has the same corruption around `ensureLabels();\n issuesIterator`, leaving invalid JavaScript.

It also re-adds the unrelated 236-line `owner-notification.yml` with `pull_request_target`, which this repository explicitly prohibits. Please remove that file. Finally, make the state machine per-assignee: one active assignee must not keep every inactive assignee, and linked PR activity must be tied to the assignee whose assignment is being evaluated. Add real tests for those transitions after fixing the test file.

FAUST-BENCHOU (commented):
Thanks for adding the inactive assignee management workflow. Automating stale assignment cleanup can help keep issue ownership healthy.

Since this workflow performs write operations (comments, labels, and unassignments), please make sure the permission boundary and trigger conditions are safe. Also, please resolve the current conflicts with main first so we can review against the latest codebase.

.github/workflows/test-and-build.yml
```diff
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+permissions:
```
Xunzhuo, line 9: **[P1] Keep a single workflow permissions mapping**

This adds a second top-level permissions key while the existing one remains at lines 6–7. GitHub Actions rejects the duplicate mapping, so callers of this reusable core test/build workflow cannot run. Merge the permission changes into the existing mapping and verify the workflow can be loaded.

.prowlabels.yaml
```diff
@@ -1,16 +1,40 @@
+area:
```
Xunzhuo, line 1: **[P2] Preserve the current Workgroup and owner taxonomy**

This replaces the canonical wg/* and owner/maintainers definitions with the retired area/track taxonomy and reintroduces lifecycle/pr-state classes. The current maintainer operations contract requires Workgroup ownership and explicitly retires area/track. Preserve the current label inventory and add only the assignment labels this workflow needs.

Xunzhuo (changes requested):
Two regressions remain: the reusable core workflow has duplicate permissions keys, and the label inventory replaces current ownership labels with retired taxonomy. Please fix both before merging. The duplicate YAML key is reproduced; the workflow itself was not executed.

FAUST-BENCHOU:
Should assignment itself count as assignee activity when the assignee did not perform it? A maintainer/bot assignment currently resets the inactivity clock, so the 15/30-day policy effectively starts from assignment time rather than the contributor's last activity. If intentional, it may be worth documenting this explicitly.

rohitjyoti:
Closing the PR as the issue got resolved in this PR: https://github.com/vllm-project/semantic-router/pull/3429

PR #2661, opened 2026-07-24
[Router] close management API Phase 2 redaction gaps (#2463)

### LINKED ISSUE #2463: [Security] Authenticate and isolate the Router management API

AayushSaini101:
Closes #2463

## Purpose

- Completes the **Phase 2 closeout** for management API secret-safe views after Phase 1 auth landed in #2561.
- **What changed**
  - Align `config/config.yaml` `global.services.management_api.auth.roles.operator` with `DefaultManagementAPIRoles()` so the reference config grants operator the full intended permission set (reads + writes), not write-only permissions.
  - Scrub YAML/JSON-style secret assignments (e.g. `api_key: …`, `password: …`) from management API error messages via `scrubSecretsInErrorMessage`, wired through `writeErrorResponse` and auth middleware errors, so parse/deploy failures cannot leak credentials.
  - Enforce reference-config role alignment in tests so YAML roles cannot drift from Go defaults again.
  - Refresh k8s canonical-export golden outputs to include `management_api: {}` now that the service is part of the exported schema.
- **Why**
  - Copying the incomplete YAML operator role into a bearer deployment would silently weaken operator capabilities (e.g. no `config.read` / `classify.invoke`).
  - Config parse/deploy errors can echo snippet text that contains plaintext secrets; those must not appear in API error JSON.
- **Modules:** `Router` (management API / config contract). No CLI, Dashboard, Operator, or E2E behavior changes in this PR.

## Test Plan

Reviewers can validate with:

```bash
cd src/semantic-router
go test ./pkg/apiserver/ -count=1 -run 'Scrub|Redact|ConfigGet|ClassifierInfo|WriteError'
go test ./pkg/config/ -count=1 -run 'ReferenceConfig|ManagementAPI'

### DIFF
--- a/src/semantic-router/pkg/apiserver/config_redact.go
+++ b/src/semantic-router/pkg/apiserver/config_redact.go
@@ -5,10 +5,26 @@ package apiserver
 import (
 	"context"
 	"net/http"
+	"regexp"
 	"strings"
 )
 
 const redactedConfigValue = "[REDACTED]"
+
+// sensitiveAssignmentPattern matches YAML/JSON-style secret assignments that
+// sometimes appear inside parse/validation error strings.
+var sensitiveAssignmentPattern = regexp.MustCompile(
+	`(?i)((?:["']?(?:api[_-]?key|access[_-]?key|password|auth[_-]?password|client[_-]?secret|private[_-]?key|secret)["']?)\s*[:=]\s*)(?:"[^"]*"|'[^']*'|[^\s,}\]]+)`,
+)
+
+// scrubSecretsInErrorMessage removes plaintext credential assignments from
+// management API error messages so parse/deploy failures cannot leak secrets.
+func scrubSecretsInErrorMessage(message string) string {
+	if message == "" {
+		return message
+	}
+	return sensitiveAssignmentPattern.ReplaceAllString(message, "${1}"+redactedConfigValue)
+}
 
 // canViewSecrets reports whether the request principal may see plaintext
 // secrets in config dumps. Requires secret_view (admin via "*" by default).
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1586,8 +1586,15 @@ global:
             - config.read
             - data.read
           operator:
+            - health.read
+            - ready.read
+            - docs.read
+            - metrics.read
+            - classify.invoke
+            - config.read
             - config.write
             - learning.ingest
+            - data.read
             - data.write
           admin:
             - "*"
--- a/src/semantic-router/pkg/apiserver/config_redact_test.go
+++ b/src/semantic-router/pkg/apiserver/config_redact_test.go
@@ -15,6 +15,64 @@ import (
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/services"
 )
 
+func TestScrubSecretsInErrorMessage(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name string
+		in   string
+		want string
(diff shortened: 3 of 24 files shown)

### CODE REVIEW

File: src/semantic-router/pkg/apiserver/config_redact.go [hunk 1]
```diff
@@ -5,11 +5,27 @@ package apiserver
...
+	"regexp"
 	"strings"
 )
 
 const redactedConfigValue = "[REDACTED]"
 
+// sensitiveAssignmentPattern matches YAML/JSON-style secret assignments that
+// sometimes appear inside parse/validation error strings.
+var sensitiveAssignmentPattern = regexp.MustCompile(
+	`(?i)((?:["']?(?:api[_-]?key|access[_-]?key|password|auth[_-]?password|client[_-]?secret|private[_-]?key|secret)["']?)\s*[:=]\s*)(?:"[^"]*"|'[^']*'|[^\s,}\]]+)`,
```
TwilightTechie, line 17: +1, really like this one. Scrubbing secrets out of parse/deploy error messages is exactly the kind of hardening everyone need for enterprise deployments, and the canary-based tests plus the role alignment guard are a nice touch.

Small question: any reason token is not in the pattern? Since the management API auth itself is bearer-token based, an error echoing something like token: sk-abc... would slip through the current list. Happy to see it as a follow-up if you prefer keeping this PR tight.

drivebyer (approved):

src/semantic-router/pkg/apiserver/config_redact.go
AayushSaini101: sounds, valid, feel free to open a MR to update the pattern thanks

PR #2664, opened 2026-07-24
[Fleet-Sim] Add measured B200 context-length sweep

chenxuniu:
## Summary

- Reposts only the measured B200 experiment contribution from #1863 against `main`.
- Archives the 2K–128K context-length sweep, software/GPU environment, and KV-cache maximum-concurrency excerpts.
- Adds a standard-library Python script that validates the CSV calculations and renders the measured table.
- Intentionally excludes all manuscript, LaTeX, and PDF changes so the paper authors can integrate the measurements separately.

## Metric scope

- `decode_tok_s` is aggregate throughput for the TP=8 serving instance; `per_gpu_decode_power_w` is mean one-GPU power.
- The CSV retains both the paper-normalized value (aggregate throughput divided by mean per-GPU power) and an estimated physical 8-GPU system tok/W value.
- The `paper_projected_*` columns are retained only as the analytical reference available when the measurements were collected.

## Reproducibility scope

These artifacts reproduce and validate the archived result table; they do not rerun the GPU experiment from scratch. The original load-generation scripts, raw power traces, and repeat-run samples are not part of this contribution.

## Validation

- `python3 src/fleet-sim/studies/context-length-power-efficiency/scripts/render_b200_table.py`
- Repository Ruff, Black, Markdown, changed-file lint, and supply-chain security checks
- `make vllm-sr-sim-test` — 304 passed

## Review note

This remains a draft so the paper authors can review the metric terminology and integrate the experimental results into the manuscript themselves.

### DIFF
--- a/src/fleet-sim/studies/context-length-power-efficiency/README.md
+++ b/src/fleet-sim/studies/context-length-power-efficiency/README.md
@@ -1,0 +1,80 @@
+# Measured B200 Context-Length Sweep
+
+This directory archives the measured B200 results contributed for the
+context-length power-efficiency study. It contains the experiment summary and
+a small standard-library script that validates and renders the measured table.
+
+It intentionally does not contain or modify the paper's LaTeX or PDF. The
+paper authors can integrate these measurements into the manuscript separately.
+
+## Measurement setup
+
+- Date: April 29, 2026
+- System: one node with 8x NVIDIA B200 GPUs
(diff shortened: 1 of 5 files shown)


Xunzhuo (commented):
Thanks for contributing this measured B200 sweep. I checked the results, provenance, and exact-head CI and don't see a code blocker in the current draft. Once the remaining terminology review is settled, please mark it ready for the final maintainer pass.

PR #2665, opened 2026-07-25
[Bug] Make password resets transactional

### LINKED ISSUE #2482: [Bug] Make credential lifecycle audit events transactionally durable
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

Make dashboard credential mutations, session invalidation, and their security audit event durable as one recoverable operation.

### To Reproduce

```bash
make agent-report ENV=cpu CHANGED_FILES="dashboard/backend/auth/store.go dashboard/backend/auth/session_store.go dashboard/backend/auth/audit_store.go dashboard/backend/auth/admin_meta_handlers.go"
cd dashboard/backend && go test -race ./auth ./router
```

Follow the nearest dashboard `AGENTS.md` and the harness-reported gates.

### Expected behavior

Use one SQLite transaction for the credential mutation, target-user session revocation, and either:

1. the final typed audit row; or
2. a durable outbox row for bounded idempotent delivery.

The handler succeeds only after that durable state commits.

### Impact

##### Audited current-main behavior (2026-07-15)

- Dashboard auth is implemented under `dashboard/backend/auth/` using SQLite.
- Administrator password reset calls `Store.UpdatePassword` in `store.go`, then calls `writeAudit` from the handler.
- `writeAudit` calls `AddAuditLog` after the mutation and discards the returned error.
- The current administrator reset path does not revoke all existing sessions for the target user.
- There is no `dashboard/backend/auth/password_store.go` on current `main`.
- Self-service password change is not a current supported handler, so it must not be described as already transactionally updating password and sessions.

A database or request failure can therefore leave a changed credential without a durable audit event, and previously issued target-user sessions may remain active.

##### Scope

- Define typed credential lifecycle events for administrator reset and any future self-service change/recovery flow.
- Add store-level transaction methods; do not coordinate SQL transactions in HTTP handlers.
- Revoke all target-user sessions when password/credential state changes.
- Preserve actor, target, action, outcome, timestamp, and correlation/idempotency key without credential material.
- Make retries/delivery idempotent and bounded if an outbox is used.
- Add retention/pruning and terminal-failure metrics/content-free logs.
- Ensure cancellation semantics do not report success after an ambiguous commit.

##### Acceptance criteria

- [ ] Successful administrator reset atomically persists the new hash, revokes target sessions, and persists an audit/outbox record.
- [ ] Failure rolls back all three effects; no false success event is written.
- [ ] Existing target-user sessions are rejected after commit.
- [ ] Duplicate retry cannot apply a second semantic mutation or duplicate the event.
- [ ] Audit/outbox payload never contains passwords, hashes, tokens, cookies, authorization headers, or raw bodies.
- [ ] Full-disk, lock/busy, canceled context, insert failure, post-commit response failure, retry, and restart tests exist.
- [ ] A future self-service path must use the same store transaction contract.

##### Likely change surfaces

`dashboard/backend/auth/store.go`, `session_store.go`, `audit_store.go`, `schema.go`, `admin_meta_handlers.go`, tests, metrics/docs, and possibly a narrowly scoped outbox store.

haowu1234:
Fixes #2482

## Summary
- Add a credential lifecycle mutation path for admin password resets that updates the password hash, revokes the target user's active sessions, inserts the audit row, and records idempotency in one SQLite transaction.
- Store keyed retries in `credential_lifecycle_requests` using operation, actor, idempotency key, and an HMAC request fingerprint so same-request retries replay while conflicting bodies return 409.
- Add bounded retention/pruning for credential lifecycle idempotency records plus content-free terminal-failure counters/logs for store failures and post-commit response encoding failures.
- Send `Idempotency-Key` from the dashboard Users page and cover the header/response shape in the auth flow e2e test.

## Compatibility
- Legacy callers without an idempotency key are still accepted and audited with the existing `user.password` action.
- `X-Request-ID` is accepted as a fallback idempotency key for clients that already provide request IDs.
- The store-level credential lifecycle transaction is the reusable boundary for future self-service change/recovery handlers, so HTTP handlers do not coordinate SQL transactions.

## Testing
- `cd dashboard/backend && CGO_ENABLED=1 go test ./auth`
- `cd dashboard/backend && CGO_ENABLED=1 go test -race ./auth ./router`
- `make dashboard-check`
- Local real-server HTTP E2E from `dashboard/backend`: built the dashboard backend, launched it against temporary SQLite stores with setup mode enabled, and verified bootstrap/login/create-user/password-reset, old-session invalidation, old-password rejection, new-password login, idempotent replay, conflict 409, one audit row, one idempotency row, and no password/raw idempotency key in audit `extraJson`.

Additional auth tests cover audit insert rollback, full-disk rollback, SQLite busy/lock rollback, canceled context, keyed retry, conflicting retry, restart/reopen replay, missing fingerprint validation, and post-commit response write failure metrics.


AayushSaini101:
Thanks, @haowu1234 for the implementation, the Pre-commit is failing can you please check and also can add a small demo for the required changes thanks

haowu1234:
Thanks! The pre-commit failure is fixed in `69a39e15`, and the rerun is green now.

I also ran a small real-backend demo against a temporary local dashboard server with fresh SQLite auth/workflow/projection stores:

```text
Dashboard credential lifecycle demo (#2482 / PR #2665)
Server: temporary local dashboard backend on 127.0.0.1:18765
Storage: temporary SQLite auth/workflow/projection DBs

HTTP flow:
- admin login                                      -> 200
- create target user (eeebbb52...)                -> 200
- target login with original password             -> 200
- admin reset with Idempotency-Key                 -> 200 {"ok":true,"replayed":false}
- old target session /api/auth/me after reset      -> 401
- target login with original password after reset  -> 401
- target login with rotated password               -> 200
- same key + same body replay                      -> 200 {"ok":true,"replayed":true}
- same key + different password                    -> 409 credential lifecycle idempotency conflict: key was already used for a different request

SQLite checks:
- user.password audit rows                         -> 1
- credential_lifecycle_requests rows               -> 1
- target sessions revoked / active                 -> 1 / 1
- audit extra_json summary                         -> {"eventType":"credential.admin_password_reset","idempotent":true,"revokedSessions":1,"idempotencyKeyIdChars":16}
- audit extra_json raw password/key leak matches   -> 0
```

This uses the real dashboard HTTP handlers rather than mocked handler calls, so it covers the externally visible reset/retry/conflict behavior plus the durable SQLite audit/idempotency rows.

Xunzhuo (commented):
Thanks for the thorough transactional and idempotency coverage. I did not find a new code blocker in this head, but it no longer merges cleanly with `main`; the conflicts are in the auth schema and Users-page/E2E surfaces. Please rebase and reconcile those current contracts, then request review again.

FAUST-BENCHOU (changes requested):
pls resolve conflitcs

PR #2666, opened 2026-07-25
[Router] vectorstore: bound ingestion memory with streaming batches

Peterren:
Part of #2474 (vector-store ingestion hardening) — implements the "bounded streaming batches" acceptance item. Not a full close of the umbrella issue, whose remaining slices (status retention, saga reconciliation, parser streaming) are tracked separately.

## Purpose

- **What:** The ingestion pipeline embedded *every* chunk of a file and then inserted them all in a single `InsertChunks` call. That holds the whole file's chunk text and embedding vectors in memory simultaneously — an unbounded-by-count per-job memory spike, multiplied across ingestion workers. This PR embeds and stores chunks in fixed-size windows instead (default 64, configurable via `vector_store.ingestion_batch_size`), bounding peak per-job memory to roughly `O(batch_size × embedding_dimension)` regardless of file size.
- **Why:** Large uploads with high-dimensional embeddings could drive large transient allocations; batching makes ingestion memory predictable and bounded. This implements the *bounded streaming batches* acceptance item of the vector-store hardening issue #2474.
- **Module(s):** `Router` (primary), with config-surface updates to `Dashboard` and `E2E`/Helm values.

### Behavior notes

- `ctx` is checked before each batch, so a cancelled lifecycle aborts promptly between batches. Chunk IDs remain globally correct across batch boundaries (indexing is unchanged).
- Batching changes failure semantics: earlier batches are already inserted when a later one fails. To preserve the prior all-or-nothing invariant (**a failed file leaves nothing searchable**), a partial failure now best-effort removes the already-inserted chunks via `DeleteByFileID`, detached from the (possibly cancelled) job context so cleanup still runs during shutdown, but bounded so a wedged backend cannot block the worker.
- **Out of scope (tracked separately under #2474):** streaming the *parser* (text extraction still returns the full document), status TTL/retention, and full transactional/saga reconciliation.

## Test Plan

From `src/semantic-router`:

```
go build ./pkg/...
go build -tags=milvus ./cmd            # real router binary
go test -race ./pkg/vectorstore ./pkg/routerruntime
go test ./pkg/config
golangci-lint run ./pkg/vectorstore/... ./pkg/config/... ./pkg/routerruntime/... \
  --config ../../tools/linter/go/.golangci.yml
```

New unit tests in `pkg/vectorstore/pipeline_batch_test.go`:

- **Batch-boundary correctness** (table): chunk counts of 1/3/5/8/13 against batch sizes 1/4/5 — asserts every chunk is stored, the number of `InsertChunks` calls equals `ceil(N/batch)`, and no batch exceeds the configured size.
- **Default batch size**: unset → 64 (65 chunks → 2 inserts).
- **Partial-failure cleanup**: a failure in a later batch removes the already-inserted chunks; the store ends with 0 searchable chunks and the file counted as failed.
- **First-batch failure**: no insert happened, so no cleanup is attempted.

## Test Result

- All of the above pass. `go test -race` is clean for `pkg/vectorstore`, `pkg/routerruntime`, and `pkg/config`; `golangci-lint` reports **0 issues**; `gofmt`/`go vet` clean; the router binary builds with the `milvus` tag.
- The Kubernetes-backed `rag-hybrid-search` E2E profile is not run locally (requires a cluster); it exercises this path in CI.
- Follow-up: the remaining #2474 items (parser streaming, status retention, saga reconciliation) are separate slices.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/vectorstore/pipeline.go
+++ b/src/semantic-router/pkg/vectorstore/pipeline.go
@@ -46,6 +46,18 @@ type IngestionJob struct {
 // backend or embedder is wedged.
 const defaultStopTimeout = 30 * time.Second
 
+// defaultBatchSize is the number of chunks embedded and inserted per batch when
+// PipelineConfig does not specify one. It bounds peak per-job memory to roughly
+// O(batchSize × embeddingDimension) instead of holding every chunk's text and
+// embedding for a file at once.
+const defaultBatchSize = 64
+
+// batchCleanupTimeout bounds the best-effort removal of already-inserted chunks
+// when a multi-batch ingestion fails partway through. It is detached from the
+// (possibly cancelled) job context so cleanup still runs during shutdown, but
+// is itself bounded so a wedged backend cannot block the worker indefinitely.
+const batchCleanupTimeout = 10 * time.Second
+
 // IngestionPipeline processes file attachment jobs asynchronously.
 // It reads files, extracts text, chunks, embeds, and stores the
 // resulting vectors in the backend.
@@ -56,6 +68,7 @@ type IngestionPipeline struct {
 	embedder     Embedder
 	jobQueue     chan IngestionJob
 	workers      int
+	batchSize    int
 	lifecycleMu  sync.Mutex
 	mu           sync.RWMutex
 	fileStatuses map[string]*VectorStoreFile // vsf_id -> status
@@ -73,6 +86,7 @@ type IngestionPipeline struct {
 type PipelineConfig struct {
 	Workers   int // number of concurrent workers (default 2)
 	QueueSize int // job queue buffer size (default 100)
+	BatchSize int // chunks embedded+inserted per batch (default 64)
 }
 
 // NewIngestionPipeline creates a new ingestion pipeline.
@@ -85,6 +99,10 @@ func NewIngestionPipeline(backend VectorStoreBackend, fileStore *FileStore, mana
 	if queueSize <= 0 {
 		queueSize = 100
 	}
+	batchSize := cfg.BatchSize
+	if batchSize <= 0 {
+		batchSize = defaultBatchSize
+	}
 
 	return &IngestionPipeline{
 		backend:      backend,
@@ -93,6 +111,7 @@ func NewIngestionPipeline(backend VectorStoreBackend, fileStore *FileStore, mana
 		embedder:     embedder,
 		jobQueue:     make(chan IngestionJob, queueSize),
 		workers:      workers,
+		batchSize:    batchSize,
 		fileStatuses: make(map[string]*VectorStoreFile),
 		stopCh:       make(chan struct{}),
 	}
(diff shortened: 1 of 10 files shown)


src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -410,6 +424,75 @@ func (p *IngestionPipeline) processJob(ctx context.Context, job IngestionJob) {
...
+// batchCleanupTimeout so a wedged backend cannot block the worker. Failure to
+// clean up is logged implicitly by leaving the file marked failed; it is not
+// surfaced as a separate error because the job is already failing.
+func (p *IngestionPipeline) cleanupPartialChunks(ctx context.Context, job IngestionJob, inserted bool) {
+	if !inserted {
+		return
+	}
+	cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), batchCleanupTimeout)
+	defer cancel()
+	_ = p.backend.DeleteByFileID(cleanupCtx, job.VectorStoreID, job.FileID)
```
TwilightTechie, line 493: +1 on keeping the all or nothing invariant, nice touch. One thing I've been bitten by before with compensation deletes keyed on file ID: AttachFile doesn't dedupe, so a retry of the same file can run while the old attachment is still completed. If the retry fails here, this delete wipes the previous attachment's chunks too, and it still shows completed with nothing searchable behind it. Tracking the chunk IDs this job inserted and deleting only those would avoid that.

src/semantic-router/pkg/vectorstore/pipeline.go
```diff
@@ -410,6 +424,75 @@ func (p *IngestionPipeline) processJob(ctx context.Context, job IngestionJob) {
...
+func (p *IngestionPipeline) failBatchJob(ctx context.Context, job IngestionJob, inserted bool, code, message string) {
+	p.cleanupPartialChunks(ctx, job, inserted)
+	p.failJob(ctx, job, code, message)
+}
+
+// cleanupPartialChunks best-effort removes chunks already inserted for a job
+// whose ingestion failed partway through. The delete is detached from the
+// job context (which may be cancelled during shutdown) but bounded by
+// batchCleanupTimeout so a wedged backend cannot block the worker. Failure to
+// clean up is logged implicitly by leaving the file marked failed; it is not
```
TwilightTechie, line 485: Just thinking out loud here: If DeleteByFileID itself fails or times out, the stated invariant (failed file leaves nothing searchable) is broken with zero operator signal. Can we log a warning with the store and file IDs (and ideally a metric) when cleanup fails?

TwilightTechie (changes requested):

drivebyer:
Nice work bounding the embed/insert memory! A few things worth a look:

- `MemoryBackend.InsertChunks` rebuilds the full text indexes on every call (`memory_backend.go:120`); batching turns one rebuild per file into ⌈N/64⌉ growing ones. Moving the rebuild out of the per-batch path would avoid this.
- `MilvusBackend` runs a synchronous `Flush` per call (`milvus_backend.go:238`), so a 10k-chunk file flushes ~157 times. Maybe flush once per file, and raise the default batch above 64?
- I'd drop the `ingestion_batch_size` config: users have no good way to pick a value, the right one is backend-dependent, and without an upper bound in `Validate` a huge value silently restores single-insert behavior. A constant plus `PipelineConfig.BatchSize` seems enough (#2535 kept the drain timeout config-only).

Xunzhuo (changes requested):
The ingestion path is still not memory-bounded: it reads the complete upload, extracts the complete text, and materializes all chunks before batching only embedding/storage. Peak memory remains O(file size + all chunks). Please stream the reader/parser/chunker boundary and add a regression that demonstrates bounded memory on a large input.

FAUST-BENCHOU (changes requested):
pls resolve conflicts

PR #2674, opened 2026-07-25
[Router] run shared multimodal input validation in both build modes

### LINKED ISSUE #2619: [Bug] Apply shared presence and NUL validation across Candle native and stub builds
### Proposed Workgroup

### Describe the bug

The Candle native and non-CGO stub builds do not apply the same safe presence and embedded-NUL checks before backend dispatch. A malformed string input can therefore produce different validation behavior depending on whether the native backend is linked.

Parent task: #2675  
Implementation: #2674

### To Reproduce

### Expected behavior

- [ ] The safe presence/NUL subset is implemented by #2674.
- [ ] Shared tests run under both build variants and prove identical pre-dispatch rejection.
- [ ] Remaining dimension, top-k, and tensor-shape work remains explicit in #2675.

### Impact

##### Proposed Workgroup

##### Scope

- Keep the build-tag-free shared validator introduced by #2674.
- Apply identical required-value and embedded-NUL rejection to the bounded multimodal string entrypoints covered by #2674.
- Return the same typed validation outcome before native or stub backend dispatch.
- Run the shared contract tests in both CGO-enabled and non-CGO builds.

##### Non-goals

- Define target-dimension legal sets.
- Define top-k bounds.
- Define image/audio tensor-shape constraints.
- Expand shared validation to every Candle entrypoint in this slice.

Those remaining semantic contracts stay in #2675 and require per-function design and coverage.

TwilightTechie:
## Summary

Closes #2619 (safe subset). Follow-up to #2491.

Request validation for the Candle binding lived **inline inside the CGO implementation**, so the non-CGO stub performed no input checks before failing closed. A malformed request was therefore rejected inconsistently depending on whether the native backend was linked — exactly the gap #2491 (item 3) called out.

This PR introduces a shared, build-tag-free validation file and wires it into both build paths.

## Changes

- **`candle-binding/validation.go`** (new, no build tag) — `validateRequiredText`, which rejects empty strings and strings containing NUL bytes (a NUL cannot cross the cgo boundary intact; `C.CString` would truncate or error).
- **`semantic-router.go`** — the three multimodal string-input entry points now call the shared validator instead of ad-hoc inline checks (empty-value messages preserved verbatim).
- **`semantic-router_mock.go`** — the same three stub functions now validate input **before** returning `ErrBackendUnavailable`, so malformed input is rejected identically in the non-CGO build.
- **`validation_test.go`** (new, no build tag) — runs under both builds; covers the helper directly and drives the public multimodal APIs with empty/NUL input (rejection happens before dispatch, so no linked backend is required).

| Function | Native (CGO) | Stub (non-CGO) |
|---|---|---|
| `MultiModalEncodeText` | shared validator | now validates |
| `MultiModalEncodeImageFromBase64` | shared validator | now validates |
| `MultiModalEncodeImageFromURL` | shared validator | now validates |

## Scope

This is deliberately the **safe subset**: presence + NUL checks only.

It does tighten one previously accepted case, corrected after review: a data URI whose prefix contains a NUL (`data:image/png\x00;base64,...`) used to decode successfully, because everything before `;base64,` is discarded. It is now rejected. It establishes the shared-validation file and the both-modes pattern.

**Deferred to #2675** (new rejection semantics, per-function design, dedicated test coverage): applying the shared validator to the remaining string inputs, and the semantic rules — target-dimension legal sets, top-k bounds, and tensor-shape constraints.

Note on issue linkage: this PR uses `Closes #2619` because the repository's `check-linked-issue` gate requires a closing keyword. #2619 is therefore closed by the safe subset, and the remaining validation work is tracked separately in #2675 rather than left implicit.

## Test Plan

```
CGO_ENABLED=0 go test .        # ok — stub + validation tests
CGO_ENABLED=1 go vet .         # clean
CGO_ENABLED=1 go test .        # validation tests ok (Rust lib built --no-default-features)
gofmt -l                       # clean
```

The build-tag-free `validation_test.go` executes under both build modes, so the shared behavior is exercised, not merely asserted. Existing native multimodal empty-input tests remain compatible (they assert `err != nil`; the empty-value message is unchanged).


drivebyer (commented):
Nice work on the shared validator. One note: `base64Str` and `url` never reach `C.CString` — both are consumed in Go (`DecodeString` / `http.NewRequest`), so the cgo-truncation rationale doesn't apply there (`data:image/png\x00;base64,…` was also accepted before this change.) Might be worth fixing the comment so #2675 doesn't inherit the premise.

TwilightTechie:
Good catch, and you're right on both counts — I verified the call paths:

- `MultiModalEncodeText` → `C.CString(text)` → cgo, so truncation is a real concern there.
- `MultiModalEncodeImageFromBase64` → `base64.StdEncoding.DecodeString(payload)` — pure Go.
- `MultiModalEncodeImageFromURL` → `http.NewRequest(...)` — pure Go.

So the cgo rationale held for one of three sites and I stated it as if it held for all of them.

Your `data:image/png\x00;base64,…` example also exposed something I under-reported in the PR description. Since the parser discards everything before `;base64,`, that input decoded fine before this change and is rejected now. The PR claimed the change touched "no previously accepted input", which was wrong; I've corrected the description rather than leave it standing.

Fixed in 8f2bbd8e. The comment now gives the reason per argument — truncation for the cgo path, malformed-input signal for the other two (NUL in a URL being a request-smuggling smell, and in a data URI able to hide bytes ahead of the separator) — and says explicitly that uniform rejection is a deliberate choice rather than a constraint that exists.

I kept the NUL check on all three rather than narrowing it to the cgo path: a shared validator applying different rules per argument seemed harder to reason about than one refusing a byte no legitimate caller sends. Happy to narrow it if you'd rather the "safe subset" framing hold strictly.

Thanks for flagging it before #2675 built on the premise — that's exactly where it would have done damage.

Xunzhuo (commented):
#3266 is now a strict superset of this shared FFI validation, including target dimensions and tensor shapes. Please consolidate there or rebase this branch to a clearly unique residual delta so the same boundary is not maintained in two competing PRs.

FAUST-BENCHOU:
The shared pre-dispatch validation is a good boundary here. Native and stub builds now reject the same malformed inputs before backend availability matters, while keeping the semantic validation expansion separate. Updating the image build inputs for the new build-tag-free helper also avoids a subtle packaging-only break. LGTM.

PR #2697, opened 2026-07-27
[Router][Dashboard][E2E] surface looper latency and token usage headers

### LINKED ISSUE #2694: feat: surface looper latency and token usage on the x-vsr-debug header + playground
### Describe the feature

The looper client records per-call token usage (#2593) and latency (#2640 / #2641), but neither is exposed on the wire — both live only on the internal `ModelResponse`. Surface them so operators and the dashboard playground can see the cost/overhead of multi-model deliberation.

Per the v0.4 header contract (#2200), looper metrics are intermediate observability, so they belong on the **`x-vsr-debug` surface**, next to the existing debug-gated `x-vsr-looper-*` trace headers (emitted in `appendLooperTraceHeaders`). Scope:

- **Router (`pkg/extproc` + `pkg/looper`)**: add an aggregate latency field to `looper.Response` (decide semantics — wall-clock vs sum across the multi-call execution), then emit `x-vsr-looper-latency-ms` (and looper token usage) inside `appendLooperTraceHeaders`.
- **Dashboard (`dashboard/frontend`)**: add `HeaderDisplay` allowlist entries. The playground already sends `x-vsr-debug: true` (added in #2451), so they render automatically under the debug reveal once added.
- **E2E**: add header assertions.

### Primary layer

global level

### Why this layer?

These are response-header observability metrics governed by the v0.4 header contract; `global.integrations.looper` produces them and the wire contract is global.

The recorded latency/token metrics are currently invisible to operators and the playground, so the overhead of multi-model deliberation can't be observed at the wire. Raised by @AayushSaini101 on #2641.

### Additional context

Follow-up to #2593 (token) and #2641 (latency). Deliberately kept out of #2641 to keep that PR to a single subsystem (recording the primitive); this one spans looper + the extproc header contract + dashboard + E2E.

theohsiung:
Closes #2694

## Purpose

The looper client records per-call token usage (#2593) and latency (#2640/#2641), but neither is exposed on the wire — both live only on the internal `ModelResponse`. This surfaces them as debug-gated response headers so operators and the dashboard playground can observe the cost/overhead of multi-model deliberation.

Modules: `Router` / `Dashboard` / `E2E`

- **Router**: `looper.ExecuteWithLatency` wraps the single `Looper.Execute()` dispatch call site (`req_filter_looper.go`) with wall-clock timing, rather than instrumenting each of the 6 algorithm implementations separately — this captures latency consistently regardless of whether an algorithm dispatches its model calls sequentially or concurrently. `appendLooperTraceHeaders` now emits `x-vsr-looper-latency-ms` and the three `x-vsr-looper-*-tokens` headers (aggregate `Usage` was already populated by #2593) alongside the existing debug-gated looper trace headers.
- **Dashboard**: `HeaderDisplay`/`HeaderReveal` allowlist entries for the four new headers — both already group `x-vsr-looper-*` by prefix, so no other wiring was needed.
- **Router bugfix** (pre-existing, found via full e2e validation — see Test Result): the auto-routing branch that dispatches to the looper never set `ctx.VSRSelectedDecisionName`, so `x-vsr-selected-decision` came back empty for any confidence/ratings-algorithm decision reached via normal decision matching (the direct fusion/remom/workflows dispatchers already set it for their algorithm types). Fixed in `processor_req_body.go`.
- **Router robustness**: `ExecuteWithLatency` now guards against a nil `Looper.Execute` response — no current implementation returns `(nil, nil)`, but the interface contract doesn't forbid it, and this wrapper is the shared instrumentation seam for all 7 implementations.
- **E2E**: new `looper_metrics_probe_decision` (confidence algorithm, 2 model refs reusing the profile's existing LoRA adapters) in the shared `kubernetes`/ai-gateway baseline config, plus a new testcase asserting all four headers as positive integers with `total == prompt + completion`. Also fixes `global.integrations.looper.endpoint`, which this profile never configured (`config.listeners: []` leaves the canonical default `localhost:8899` unreachable) — pointed directly at the profile's vLLM backend instead of standing up new infrastructure.

## Test Plan

- `go test ./pkg/looper/... ./pkg/extproc/... ./pkg/headers/... ./pkg/config/...` from `src/semantic-router`
- `npx vitest run src` from `dashboard/frontend`
- `golangci-lint run --config tools/linter/go/.golangci.agent.yml --new-from-rev origin/main` and the full `make agent-lint` gate
- Full end-to-end validation against a live `kind` cluster (`kubernetes`/ai-gateway profile): `make e2e-setup` then `make e2e-test-only E2E_TESTS="looper-latency-token-headers"`

## Test Result

- All unit tests pass (Go + Vitest), `make agent-lint` passes clean.
- Live e2e run: `looper-latency-token-headers` **passed**, with `x-vsr-looper-latency-ms=74`, `x-vsr-looper-prompt-tokens=16`, `x-vsr-looper-completion-tokens=101`, `x-vsr-looper-total-tokens=117` (117 = 16+101).
- The e2e cluster validation surfaced two real, pre-existing issues along the way, both fixed in this PR: the missing `VSRSelectedDecisionName` assignment above, and the unconfigured looper endpoint in the e2e profile.
- No follow-up risks identified for this PR's scope. Two out-of-scope, low-severity items noted during review and left unfixed: `cmd/fusioneval` bypasses `ExecuteWithLatency` (dormant — it never reads `LatencyMs`), and a pre-existing sibling YAML description (`retention_probe_decision`, unrelated to this PR) has the same unquoted-`#`-truncation issue as one I fixed on my own new decision.


AayushSaini101:
thanks for the dashboard changes @theohsiung can you attached the screeshots as well thanks : )

src/semantic-router/pkg/extproc/processor_req_body.go
```diff
@@ -137,6 +137,11 @@ func (r *OpenAIRouter) handleModelRouting(openAIRequest *openai.ChatCompletionNe
 			"decision":   ctx.VSRSelectedDecision.Name,
 			"algorithm":  ctx.VSRSelectedDecision.Algorithm.Type,
 		})
+		// Track VSR decision information the same way handleAutoModelRouting's
+		// trackVSRDecision does for non-looper decisions, and the direct
+		// fusion/remom/workflows dispatchers already do for their algorithm
+		// types; otherwise x-vsr-selected-decision comes back empty.
+		ctx.VSRSelectedDecisionName = ctx.VSRSelectedDecision.Name
```
TwilightTechie, line 144: Really good catch here. Empty `x-vsr-selected-decision` on the looper path is the kind of thing you only find when something finally asserts on it in a real cluster. Also glad the timing went into one wrapper instead of all 7 algorithms.

src/semantic-router/pkg/extproc/req_filter_looper_response.go
```diff
@@ -65,6 +66,10 @@ func appendLooperTraceHeaders(setHeaders *[]*core.HeaderValueOption, resp *loope
 		newHeaderValueOption(headers.VSRLooperModelsUsed, strings.Join(resp.ModelsUsed, ",")),
 		newHeaderValueOption(headers.VSRLooperIterations, fmt.Sprintf("%d", resp.Iterations)),
 		newHeaderValueOption(headers.VSRLooperAlgorithm, resp.AlgorithmType),
+		newHeaderValueOption(headers.VSRLooperLatencyMs, fmt.Sprintf("%d", resp.LatencyMs)),
```
TwilightTechie: Tried this locally: built a` looper.Response` that skipped `ExecuteWithLatency` and all four headers came back present with value `"0"`, so a missing measurement is indistinguishable from a real zero. Nothing hits that today, but `cmd/fusioneval` already calls `fusion.Execute` directly, so it's one call site away. Skip the append when the value is 0?

src/semantic-router/pkg/headers/headers.go
```diff
@@ -388,4 +388,27 @@ const (
...
+
+	// VSRLooperLatencyMs indicates the wall-clock latency, in milliseconds,
+	// of the full looper execution (all model calls plus algorithm overhead).
+	// Value: "842" (example)
+	VSRLooperLatencyMs = "x-vsr-looper-latency-ms"
+
+	// VSRLooperPromptTokens indicates the aggregate prompt token count
+	// across all model calls made during looper execution.
+	// Value: "512" (example)
+	//nolint:gosec
```
TwilightTechie, line 400: these are G101 false positives on the word "tokens", makes sense to suppress :).

src/semantic-router/pkg/extproc/processor_req_body.go
theohsiung: This was the first e2e test to ever exercise a confidence/ratings-algorithm decision through the normal auto-matching path (fusion/remom/workflows only go through their own direct "pseudo-model" dispatchers, which already set the field) — that's exactly why no unit test had caught it either.

src/semantic-router/pkg/extproc/req_filter_looper_response.go
theohsiung: Fixed in 8faaca5c: appendLooperTraceHeaders now skips each latency/token header when its value is <= 0 instead of emitting "0", so header presence always means "actually measured" (matches the existing appendOptionalHeader convention in this file for empty-string values). Added TestCreateLooperResponseOmitsUnmeasuredZeroValueHeaders. Worth noting: the three token headers come from resp.Usage, which each algorithm's own aggregation populates independently of ExecuteWithLatency, so a bypass like fusioneval only zeroes LatencyMs today — but the fix covers all four uniformly regardless.

src/semantic-router/pkg/headers/headers.go
theohsiung: Right, same pattern as the existing VSRContextTokenCount suppression in this file.

AayushSaini101 (dismissed):

theohsiung:
@AayushSaini101 @TwilightTechie Thx for the review. The screenshots are shown below
2694-playground-header-reveal
2694-playground-header-display

theohsiung:
> @AayushSaini101 @TwilightTechie Thx for the review. The screenshots are shown below 2694-playground-header-reveal 2694-playground-header-display

The playground now surfaces the looper's aggregate latency and token usage — as persistent LOOPER LATENCY (MS) / PROMPT/COMPLETION/TOTAL TOKENS badges in the chat (second image) and in the transient 'Routing this request' reveal shown right after sending (first image).

AayushSaini101 (approved):
Thanks for the testing part and implementation @theohsiung : )

Issue #1843, opened 2026-04-28
bug: router crashes at around 4000 tokens prompt


aleksandaryanakiev:
### Describe the bug

Hello,

**Issue:**
I was testing the Semantic Router using the dashboard "Playground" and "Brain" sections and I found out that when I place a prompt around 4000 tokens the router crashes without any errors in the container.

**Suggestion:**
I dug deeper and found out that there are 4 places where it can crash without providing any logs (Keep in mind that this research was done with AI, I did not validate it):
1. pkg/classification/authz_classifier.go:157
2. pkg/extproc/processor_core.go:49 - no recover() wrapper
3.  Several goroutines lacking panic recovery:
   - pkg/extproc/server.go:437 — config reload goroutine
   - pkg/extproc/processor_res_memory.go:27 — memory store goroutine
   - pkg/extproc/req_filter_modality.go:270,277 — parallel AR text + diffusion calls
   - pkg/extproc/router_selection.go:375 — lookup table populator

### To Reproduce

`"2026-04-28T08:14:12.762Z [DEBUG] Watching for changes in setting files C:\Users\x\.claude...
2026-04-28T08:14:12.767Z [DEBUG] [init] configureGlobalMTLS starting
2026-04-28T08:14:12.768Z [DEBUG] [init] configureGlobalMTLS complete
2026-04-28T08:14:12.768Z [DEBUG] [init] configureGlobalAgents starting
2026-04-28T08:14:12.771Z [DEBUG] [init] configureGlobalAgents complete
2026-04-28T08:14:12.812Z [DEBUG] Using bash path: "C:\Program Files\Git\bin\bash.exe"
2026-04-28T08:14:12.816Z [DEBUG] [LSP MANAGER] initializeLspServerManager() called
2026-04-28T08:14:12.816Z [DEBUG] [LSP MANAGER] Created manager instance, state=pending
2026-04-28T08:14:12.817Z [DEBUG] [LSP MANAGER] Starting async initialization (generation 1)
2026-04-28T08:14:12.820Z [DEBUG] [STARTUP] Loading MCP configs...
2026-04-28T08:14:12.821Z [DEBUG] Found 0 plugins (0 enabled, 0 disabled)
2026-04-28T08:14:12.822Z [DEBUG] Total LSP servers loaded: 0
2026-04-28T08:14:12.822Z [DEBUG] [LSP SERVER MANAGER] getAllLspServers returned 0 server(s)
[...]

PR #2707, opened 2026-07-28
[Router] recover from panics in modelruntime task goroutines

### LINKED ISSUE #2706: bug: [Router] Unrecovered panic in a modelruntime task goroutine crashes a live router on config reload, defeating BestEffort
### Describe the bug

`modelruntime.startTask` runs each task on a bare goroutine with no `recover()`:

```go
// src/semantic-router/pkg/modelruntime/executor.go:222
go func(task Task) {
	r.resultCh <- taskOutcome{
		name: task.Name,
		err:  task.Run(ctx),
	}
}(state.task)
```

In Go an unrecovered panic in *any* goroutine aborts the whole process, so a panicking initializer takes the router down. Three things make this worse than a startup-only failure:

**1. It is reachable on a live, traffic-serving router.** `Execute` is not called only at boot — a config file change re-enters it:

```
fsnotify config change (e.g. ConfigMap update)
  └─ server_config_watch.go:163  goSafely("config_reload_debouncer", …)
      └─ reloadRouterFromFile → reloadRouterFromConfig (server.go:246)
          └─ buildReloadRouter → createRouterClassifier (router_build.go:171)
              └─ Classifier.InitializeRuntime (classifier_lifecycle.go:73)
                  └─ modelruntime.Execute → startTask
                      └─ go func { task.Run(ctx) }      ← unprotected
```

**2. The existing `goSafely` hardening cannot help here.** The reload debouncer is already wrapped (`server_config_watch.go:163`), but `recover()` only catches panics on *its own* goroutine. The panic happens in a grandchild goroutine spawned further down the chain, so the protection added for #1843 never applies.

**3. It defeats the `BestEffort` contract.** Five of the nine classifier tasks are registered `BestEffort: true` (`classifier_lifecycle.go:106-110` — fact_check, hallucination, feedback, preference, language) precisely so their failure stays non-fatal. `BestEffort` only governs *returned errors*; a panic bypasses it entirely and kills the process.

The task bodies are CGO calls into the Rust Candle bindings (`candle_binding.InitModel`, `InitEmbeddingModels…` at `router_runtime.go:185/213/241/376-382`), which is the same surface that produced #1843 ("router crashes at around 4000 tokens prompt", where a panic took the container down with no log line).

There is a second, simpler instance of the same class. `maybePopulateFromReplay` launches `go populateFromReplay(storage, reader)` unprotected at `router_selection.go:567`, while the *identical function* is wrapped 70 lines later at `:640`:

```go
// goSafely so a panic in populateFromReplay (e.g. malformed
// replay-store entry) is logged instead of crashing the whole
// router process — see #1843.
goSafely("lookup_table_populator", func() { … })
```

The one-shot call site was simply missed when the periodic one was hardened. It parses the same replay-store entries and runs on every router build, including reloads.

### To Reproduce

Add a task whose `Run` panics and call `Execute` (this mirrors what a CGO init panic does):

```go
func TestPanicKillsProcess(t *testing.T) {
	_, err := Execute(context.Background(), []Task{{
		Name:       "panicking.best_effort",
		BestEffort: true,                       // explicitly non-fatal
		Run:        func(context.Context) error { panic("simulated CGO model-init crash") },
	}}, Options{})
	t.Logf("Execute returned err=%v", err)      // never reached
}
```

```
$ go test ./pkg/modelruntime/ -run TestPanicKillsProcess
panic: simulated CGO model-init crash

goroutine 22 [running]:
…modelruntime.(*executorRun).startTask.func1(…)
	src/semantic-router/pkg/modelruntime/executor.go:225 +0x4e
created by …modelruntime.(*executorRun).startTask in goroutine 21
	src/semantic-router/pkg/modelruntime/executor.go:222 +0x1dc
FAIL	…/pkg/modelruntime	0.011s
```

The test binary aborts: `Execute` never returns and the `BestEffort` flag makes no difference.

### Expected behavior

A panicking initializer should fail *that task* and be logged, not abort the process — matching the behaviour `goSafely` already provides in `pkg/extproc`. `BestEffort` tasks should stay non-fatal, and required tasks should surface a normal error so dependents are skipped through the existing `failTask`/`markSkipped` path.

One subtlety worth stating, because it constrains the fix: the task goroutine's only job is to send to `r.resultCh`, and `execute()` blocks on `<-r.resultCh` (`executor.go:159`). A `recover()` that swallows the panic without still sending an outcome would leave `r.running` permanently non-zero and convert the crash into a permanent hang. The recovery must deliver a failure outcome on the recovered path too.

### Affected layer

global level

### Additional context

Related prior work: #1843 (the original crash report) and PR #1844, which added `goSafely` and wrapped the four background goroutines in `pkg/extproc`. This issue is the same class in `pkg/modelruntime`, plus the missed sibling call site in `router_selection.go`.

I have a fix and regression tests ready and will open a linked PR.

glitch-ux:
Closes #2706

## Purpose

- **What does this PR change?** `modelruntime.startTask` ran each task on a bare goroutine, so an unrecovered panic in a task aborted the whole router process. This adds a deferred `recover()` in the task goroutine that converts the panic into a normal task error. It also wraps the one-shot `populateFromReplay` call in `maybePopulateFromReplay` with the existing `goSafely` helper.

- **Why is this change needed?** The crash is reachable on a live, traffic-serving router, not just at startup. A config file change re-enters the executor via `reloadRouterFromConfig` → `buildReloadRouter` → `createRouterClassifier` → `InitializeRuntime` → `modelruntime.Execute`, and the task bodies are CGO calls into the Candle bindings — the same surface that produced #1843. The reload debouncer is already wrapped in `goSafely`, but `recover()` only catches panics on its own goroutine, so that protection never reached the task goroutines spawned further down the chain. The panic also bypassed `BestEffort`, which five of the nine classifier tasks set precisely so their failure stays non-fatal.

  For `router_selection.go`, the periodic `populateFromReplay` call was hardened for #1843 at line 640 with the comment *"so a panic in populateFromReplay (e.g. malformed replay-store entry) is logged instead of crashing the whole router process"*. The one-shot call 70 lines earlier was missed; it parses the same replay-store entries and runs on every router build, including reloads.

### Implementation note

The task goroutine's only job is to send to `r.resultCh`, and `execute()` blocks on `<-r.resultCh`. Recovering without still sending an outcome would leave `r.running` permanently non-zero and turn the crash into a permanent hang, so the outcome is sent from the deferred function on both the normal and the recovered path — exactly one send either way. Converting the panic to a `taskOutcome` error (rather than just logging it, as `goSafely` does) lets the existing `failTask`/`markSkipped` logic apply unchanged, which is what preserves `BestEffort` semantics and dependent skipping.

## Test Plan

- `go test -race -count=1 ./pkg/modelruntime/...` — two new regression tests in `executor_test.go`:
  - `TestExecuteRecoversFromBestEffortTaskPanic` — a panicking `BestEffort` task must not abort the process; `Execute` returns nil and the task is marked `TaskFailed` carrying the panic value.
  - `TestExecuteReturnsErrorWhenRequiredTaskPanics` — a panicking required task returns an error and its dependent is `TaskSkipped`.

  Both would abort the test binary if the recovery regressed, and both would hang (and time out) if the outcome send were dropped — so they cover the crash and the deadlock failure modes.
- `go test -count=1 ./pkg/extproc/...` for the `router_selection.go` change.
- `gofmt -l` and `go vet` on both packages.

Go tests in this module link the Rust bindings, so the suites above were run after `cargo build --release` in `candle-binding`, `nlp-binding` and `ml-binding`, with `CGO_LDFLAGS`/`LD_LIBRARY_PATH` pointed at the resulting `target/release` directories.

## Test Result

Reproduction before the fix — the process aborts and `Execute` never returns, despite `BestEffort: true`:

```
=== RUN   TestReproPanicKillsProcess
panic: simulated CGO model-init crash

goroutine 22 [running]:
…modelruntime.(*executorRun).startTask.func1(…)
	src/semantic-router/pkg/modelruntime/executor.go:225 +0x4e
created by …modelruntime.(*executorRun).startTask in goroutine 21
	src/semantic-router/pkg/modelruntime/executor.go:222 +0x1dc
FAIL	…/pkg/modelruntime	0.011s
```

After the fix:

```
$ go test -race -count=1 ./pkg/modelruntime/... -v
--- PASS: TestExecuteRunsDependenciesAndIgnoresBestEffortFailure (0.00s)
--- PASS: TestExecuteSkipsDependentsWhenBestEffortDependencyFails (0.00s)
--- PASS: TestExecuteReturnsErrorForRequiredFailure (0.00s)
--- PASS: TestExecuteRecoversFromBestEffortTaskPanic (0.00s)
--- PASS: TestExecuteReturnsErrorWhenRequiredTaskPanics (0.00s)
--- PASS: TestEmbeddingRuntimeTasksUseOnlyRemoteProviderWhenConfigured (0.00s)
--- PASS: TestPrepareRouterRuntimeProbesRemoteEmbeddingProvider (0.02s)
--- PASS: TestPrepareRouterRuntimeReportsRemoteEmbeddingProbeFailure (0.00s)
PASS
ok  	…/pkg/modelruntime	1.042s

$ go test -count=1 ./pkg/extproc/...
ok  	…/pkg/extproc	1.345s

$ gofmt -l pkg/modelruntime/ pkg/extproc/    # no output
$ go vet ./pkg/modelruntime/... ./pkg/extproc/...   # no output
```

**Follow-up risks / gaps:** this PR deliberately covers only the two call sites on the config-reload crash path. Auditing the module turned up other unprotected goroutines running panic-capable code that are worth hardening separately — most notably the per-request signal-dispatch fan-out in `pkg/classification/classifier_signal_dispatch.go:120`, which runs CGO inference over user-controlled prompt text and is *not* covered by the `Process()`-level recover at `processor_core.go:75` (again because `recover()` is per-goroutine). I'm happy to open a follow-up issue for those if that's useful.

An alternative design would be to lift `goSafely` out of `pkg/extproc` into a shared package and reuse it here. I kept the recovery local to the executor instead, because the executor needs to *convert* the panic into a `taskOutcome` rather than just log it — reusing `goSafely` as-is would swallow the outcome and hang `execute()`. Happy to restructure toward a shared helper if you'd prefer that direction.


AayushSaini101 (approved):
LGTM thanks good enhancements : )

PR #2713, opened 2026-07-28
[Router][E2E][Docs] Isolate multi-factor config by decision

### LINKED ISSUE #2710: bug: isolate multi_factor configuration by matched decision
## Describe the bug

`multi_factor` configuration is currently resolved as process-global state even though algorithm configuration is attached to individual decisions. When multiple decisions define different valid `multi_factor` weights, the runtime can build one selector from the first discovered configuration and use it for every matched decision.

This makes model selection depend on configuration discovery order rather than the policy attached to the matched decision.

## To reproduce

1. Configure two decisions with the same candidate models.
2. Give the decisions opposite `multi_factor` policies, for example quality-only versus cost-only.
3. Make requests that independently match each decision.
4. Observe that both requests can be evaluated with the same selector configuration.

The relevant flow resolves a single multi-factor configuration while building model selection state, and request classification only carries a decision-scoped override for the hybrid selector.

## Expected behavior

The matched decision must always use its own algorithm configuration. Two decisions with opposite valid weights must select independently and consistently in the same process.

An implementation may cache immutable selectors by normalized decision algorithm configuration, or construct a decision-scoped selector safely, but it must not share mutable policy state across decisions.

## Affected layer

Model selection algorithm / routing policy.

## Additional context

A regression test should include at least two decisions that share candidates but use opposite factor weights. A single-decision test cannot detect this isolation failure.

duchangyong:
Closes #2710

## Purpose

- Build `multi_factor` selectors from the algorithm configuration on the matched decision.
- Stop promoting the first decision's `multi_factor` configuration into the global selector registry.
- Add unit and E2E regression coverage for two decisions that use opposite quality-only and cost-only policies.
- Document that `multi_factor` policy is decision-scoped.
- Affected modules: `Router` / `E2E` / `Docs`

## Test Plan

- `make agent-ci-gate ENV=cpu CHANGED_FILES="e2e/profiles/routing-strategies/profile.go e2e/profiles/routing-strategies/values.yaml e2e/testcases/decision_scoped_multi_factor.go src/semantic-router/pkg/extproc/req_filter_classification.go src/semantic-router/pkg/extproc/req_filter_classification_selection_test.go src/semantic-router/pkg/extproc/router_selection.go src/semantic-router/pkg/extproc/router_selection_config_test.go website/docs/tutorials/algorithm/selection/multi-factor.md"`
- The gate runs repository lint and contract checks, the complete Semantic Router test target, and the E2E binary build.
- The new E2E testcase makes real chat requests and verifies both the selected decision and selected model response headers.

## Test Result

- Agent CI gate passed.
- Full `test-semantic-router` target passed.
- E2E test binary built successfully.
- The cluster-backed E2E testcase is included for the upstream E2E workflow; it was not executed by the CPU agent gate.

---
<details>

</details>

### DIFF
--- a/e2e/profiles/routing-strategies/profile.go
+++ b/e2e/profiles/routing-strategies/profile.go
@@ -85,6 +85,7 @@ func (p *Profile) GetTestCases() []string {
 		"keyword-routing",
 		"entropy-routing",
 		"routing-fallback",
+		"decision-scoped-multi-factor",
 	}
 }
 
--- a/e2e/profiles/routing-strategies/values.yaml
+++ b/e2e/profiles/routing-strategies/values.yaml
@@ -25,8 +25,62 @@ config:
           - name: local-vllm
             endpoint: vllm-llama3-8b-instruct.default.svc.cluster.local:8000
             weight: 1
+      - name: premium-model
+        pricing:
+          currency: USD
+          prompt_per_1m: 10
+        backend_refs:
+          - name: local-vllm-premium
+            endpoint: vllm-llama3-8b-instruct.default.svc.cluster.local:8000
+            weight: 1
+      - name: economy-model
+        pricing:
+          currency: USD
+          prompt_per_1m: 1
+        backend_refs:
+          - name: local-vllm-economy
+            endpoint: vllm-llama3-8b-instruct.default.svc.cluster.local:8000
+            weight: 1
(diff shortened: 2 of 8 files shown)


Xunzhuo (approved):
LGTM thanks @duchangyong

PR #2714, opened 2026-07-28
[Feature] Support decision-scoped candidate quality

### LINKED ISSUE #2711: [Feature] Support decision-scoped candidate quality for multi_factor
## Describe the feature

Allow a decision model reference to optionally override the model quality score used by `multi_factor` for that decision.

A generic canonical shape could be:

```yaml
routing:
  decisions:
    - name: coding
      modelRefs:
        - model: model-a
          quality_score: 0.92
        - model: model-b
          quality_score: 0.74
```

The field should be optional and represented so that an explicit zero is distinguishable from an omitted value.

## Primary layer

Algorithm / per-decision model selection.

## Why this layer?

Quality in this case is not a new text signal and does not change boolean decision matching. It is a candidate fact consumed after a decision matches, so the model-selection algorithm is the primary owner. Canonical configuration, CLI schema and config UI are supporting representation surfaces.

The same model can have different measured or curated quality for different decision contexts. A single global model-card `quality_score` cannot represent those differences, so otherwise a correctly matched decision cannot influence final `multi_factor` ranking with context-specific quality.

## Expected behavior

- When `modelRef.quality_score` is present, `multi_factor` uses it for that candidate in the matched decision.
- When it is omitted, `multi_factor` falls back to the model-card global quality score.
- An explicit `quality_score: 0` remains zero and does not trigger fallback.
- The configuration is validated and round-trips through canonical import/export and supported control-plane representations.

## Additional context

Reusing `modelRef.weight` would mix two meanings and would not reliably distinguish an omitted value from an explicit zero. A dedicated optional quality field keeps weighted-distribution semantics separate from quality scoring.

duchangyong:
Closes #2711

## Purpose

- Add an optional `routing.decisions[].modelRefs[].quality_score` override so the same model can carry different quality priors in different decisions.
- Preserve an explicit zero, validate finite values in the `0..1` range, and fall back to the model-card quality score only when the decision override is omitted.
- Apply the override in `multi_factor` selection and expose the same contract through the CLI schema, dashboard editor, reference configuration, and documentation.
- Affected modules: `Router` / `CLI` / `Dashboard` / `Docs` / `E2E`.

## Test Plan

- Run `make agent-ci-gate ENV=cpu`.
- Run the `routing-strategies` E2E profile in an isolated Podman/Kind cluster.
- Run the CLI container integration suite with Podman.
- When the default Prometheus host port is already allocated, rerun runtime lifecycle contracts with the supported `VLLM_SR_PORT_OFFSET` setting and verify the simulator sidecar environment plus `/healthz` endpoint on the offset port.

## Test Result

- `make agent-ci-gate ENV=cpu`: passed, including lint, reference-config checks, the complete Router test suite, CLI tests, Dashboard checks, E2E build, and documentation validation.
- `routing-strategies` E2E: `decision-scoped-multi-factor` passed 1/1 in an isolated Podman/Kind cluster, verifying three real chat-completion routing paths: model-card quality fallback selected `premium-model`, reversed decision-scoped quality overrides selected `economy-model`, and the cost policy continued to select `economy-model` (`verified_policies=3`).
- CLI Podman integration suite: 35 of 39 tests passed on the default ports; four runtime-start tests were blocked by an existing host binding on port 9090.
- The environment propagation, running-container contract, and stop lifecycle tests all passed after applying a port offset. The simulator sidecar contract was also verified through its injected service URL and successful `/healthz` response on the offset port.
- No known code blocker remains for the affected modules.

---
<details>

</details>


src/semantic-router/pkg/config/canonical_config.go
```diff
@@ -246,6 +247,9 @@ func validateCanonicalDecisionModelRefs(decision Decision, modelsByName map[stri
 		if modelRef.Model == "" {
 			continue
 		}
+		if modelRef.QualityScore != nil && (math.IsNaN(*modelRef.QualityScore) || *modelRef.QualityScore < 0 || *modelRef.QualityScore > 1) {
```
Xunzhuo: [P2] Validate quality scores in candidate iterations too. This check only walks Decision.ModelRefs, but ModelRef is also used by candidateIterations[].models. Values such as .nan or 2 therefore pass canonical validation on that path even though the same values are rejected in modelRefs. Please reuse the quality-score validation from validateDecisionCandidateIterationModels and add invalid candidate-iteration cases.

duchangyong: Addressed in `dfd10eec` and extended with behavior-level coverage in `ebcb0c90`.

- The shared quality-score validator is now used for canonical decision model refs, runtime model refs, and `candidateIterations[].models[]`.
- Validation tests cover negative values, values above 1, and non-finite `.nan` values.
- The `routing-strategies` E2E now verifies both model-card fallback and reversed decision-scoped quality overrides through real chat-completion requests; the existing cost policy remains covered.
- `make agent-ci-gate ENV=cpu` and the targeted isolated Podman/Kind E2E pass.

duchangyong:
Closing in line with the maintainer decision in #2711.

This PR introduces a manual quality_score override on each decision's ModelRef. The agreed direction is to derive candidate quality from versioned evaluation and pool qualification under #3041, keeping request-time hard eligibility separate under #3202 and #3116.

The earlier candidate-iteration validation fix addresses the field introduced by this PR; it is not a separate fix for main. Passing checks and the earlier E2E results do not resolve the source-of-truth disagreement, so I am withdrawing this implementation. This does not imply that the replacement work is already complete.

PR #2717, opened 2026-07-28
[Router] make router diagnostics content-free by default

### LINKED ISSUE #2464: security: make router diagnostics content-free by default
## Summary

Make normal router diagnostics metadata-only across request, response, cache, tool, provider, embedding, and memory paths. Owner surfaces: `src/semantic-router/pkg/extproc/response_log_redaction.go`, `src/semantic-router/pkg/extproc/utils.go`, `src/semantic-router/pkg/anthropic/client.go`, `src/semantic-router/pkg/extproc/req_filter_cache.go`, `src/semantic-router/pkg/extproc/req_filter_tools.go`, and `src/semantic-router/pkg/apiserver/route_embeddings.go`.

## Current behavior

Header redaction is covered, but body mutations, immediate responses, provider bodies, queries, tool text, and rewritten content can still reach normal logs. Log levels and truncation do not establish a confidentiality boundary.

## Expected behavior

Normal logs contain correlation metadata, counts, timing, model identity, and approved hashes, never prompt, response, query, tool, or provider content. Exceptional content capture is explicit, short-lived, access-controlled, and separately retained.

## Acceptance

- Define and use one content-logging policy at every listed boundary.
- Remove body mutation and immediate-response bytes before proto logging.
- Replace direct content fields with IDs, byte/token counts, latency, model, and approved hashes.
- Gate exceptional content capture by explicit authorization, sampling, expiry, and a separate secure sink.
- Add a static check or focused lint rule that rejects direct logging of known content fields outside the approved helper.

## Validation

- Add logger-capture canaries for request/response bodies, immediate responses, provider errors, cache queries, tool arguments, embedding queries, and memory rewrites at INFO, DEBUG, and TRACE.
- Preserve credential-header canary coverage.
- Run `cd src/semantic-router && go test -race ./pkg/extproc ./pkg/apiserver ./pkg/anthropic`.
- Run the changed-surface `make agent-lint` and `make agent-ci-gate` commands reported by the harness.

## Related

- Parent audit: #2375
- Security index: #2357
- Credential-header redaction history: #2132

TwilightTechie:
## Summary

Closes #2464.

Router diagnostics interpolated user and provider content directly into log lines — prompts, cache and memory queries, tool-selection text, and raw provider response bodies. Several were at **INFO**, so they were emitted under default production logging, not only when debug was enabled. Where a boundary existed at all it was truncation, which bounds the size of a disclosure without preventing one.

A canary shows the pre-fix behaviour directly. With debug logging on, `utils.go` dumps the ext_proc mutation with `%+v`, and `redactResponseForLog` masked only headers:

```
body_mutation:{body:"{\"messages\":[{\"role\":\"user\",\"content\":\"my social
security number is 123-45-6789 body-canary-001\"}]}"}
```

## Changes

**1. One shared content-logging policy** — `logging.ContentDescriptor` / `ContentDescriptorBytes` return `[content] len=40 sha256=181f8c90469356a0`: byte length plus a truncated digest. That keeps what makes such a line useful for debugging — how large the content was, and whether two occurrences were identical — without disclosing it. It lives in `logging` because the callers span seven packages that all already depend on it. Credentials are deliberately out of scope: those must not be logged at all, not even as a hash.

**2. Body redaction in ext_proc diagnostics** — `redactResponseForLog` now redacts body content on the clone for every phase variant, covering the inline body, the streamed-response mutation, and the immediate-response body. The byte count is preserved; an empty body stays empty so the log still distinguishes "no body" from "body withheld". The phase switch is factored into `responseCommonResponse` so header and body redaction share one enumeration.

**3. Every remaining boundary routed through the descriptor** — 14 sites across `extproc`, `apiserver`, `anthropic`, `cache`, `memory`, `tools`, and `classification`.

Two sites in `classification` (raw vLLM jailbreak and external preference model output) are not in the issue's file list but leak the same way, so they are included.

The truncation helpers those sites relied on (`shortenHybridQuery` and two inline preview blocks) are deleted — leaving them beside the descriptor would imply a boundary that does not exist.

### Deliberately unchanged

- `redis_cache.go:427` and `valkey_cache.go:365` — their `query` variable holds a Redis search expression built from a request id (`@request_id:"<uuid>"`), not user content. Hashing an id would remove debuggability for no privacy gain.
- `user_id` in the two memory paths — an identifier used to correlate a user's retrievals. The issue explicitly permits IDs.

## Test Plan

Red first, then green:

```
# before: 3 canaries fail, prompt visible verbatim in the dump
# after:
go test -race ./pkg/extproc ./pkg/apiserver ./pkg/anthropic ./pkg/observability/logging   # ok
go test -race ./pkg/memory ./pkg/tools ./pkg/classification                               # ok
go build ./pkg/...   # clean
gofmt -l pkg/        # clean
```

New tests: 3 ext_proc canaries (body mutation, immediate response, size metadata retained) and 4 for the descriptor (no content, stable, distinguishing, empty). The canaries format the response exactly as `utils.go` does rather than walking a subset of fields — the existing header-only dump helper cannot observe a body leak, so it would have passed while content still reached the log.

Each of the three commits builds independently.

**Not verified locally:** `pkg/cache` has tests requiring live Valkey and Milvus instances, which are not running on this machine. `TestValkeyCacheIntegration_IPMetricType` was confirmed to fail identically with these changes stashed, and the Milvus smoke test fails during client construction, before the changed log line is reached. CI runs these with the services up.

### Environment

macOS 15 (Darwin 25.5.0, arm64), CPU-only, Go 1.26.3.

## Follow-up

The issue's final acceptance item asks for a static check or lint rule rejecting direct logging of known content fields outside the approved helper. That is tracked in #2718 rather than included here: a naive pattern-based rule flags the two Redis search-expression sites above as false positives, and a gate people learn to ignore is worse than none. It needs a small allowlist or an AST-based check, which is its own change.

### DIFF
--- a/src/semantic-router/pkg/anthropic/client.go
+++ b/src/semantic-router/pkg/anthropic/client.go
@@ -158,7 +158,7 @@ func ToOpenAIResponseBodyWithExt(anthropicResponse []byte, model string, ext *ir
 // stop_reason, cache usage, and server-tool counts survive the
 // flattening that the OpenAI envelope cannot represent.
 func toOpenAIResponseBody(anthropicResponse []byte, model string, ext *ir.IRExtensions) ([]byte, error) {
-	logging.Debugf("Raw Anthropic response (%d bytes): %s", len(anthropicResponse), string(anthropicResponse))
+	logging.Debugf("Raw Anthropic response: %s", logging.ContentDescriptorBytes(anthropicResponse))
 
 	var resp anthropic.Message
 	if err := json.Unmarshal(anthropicResponse, &resp); err != nil {
--- a/src/semantic-router/pkg/apiserver/route_embeddings.go
+++ b/src/semantic-router/pkg/apiserver/route_embeddings.go
@@ -339,8 +339,8 @@ func (s *ClassificationAPIServer) handleBatchSimilarity(w http.ResponseWriter, r
 		ProcessingTimeMs: result.ProcessingTimeMs,
 	}
 
-	logging.Infof("Calculated batch similarity: query='%s', %d candidates, top-%d matches (model: %s, took: %.2fms)",
-		req.Query, len(req.Candidates), len(matches), result.ModelType, result.ProcessingTimeMs)
+	logging.Infof("Calculated batch similarity: query=%s, %d candidates, top-%d matches (model: %s, took: %.2fms)",
+		logging.ContentDescriptor(req.Query), len(req.Candidates), len(matches), result.ModelType, result.ProcessingTimeMs)
 
 	s.writeJSONResponse(w, http.StatusOK, response)
 }
--- a/src/semantic-router/pkg/cache/hybrid_cache_lookup.go
+++ b/src/semantic-router/pkg/cache/hybrid_cache_lookup.go
@@ -41,8 +41,8 @@ func (h *HybridCache) findSimilar(
 		return nil, false, nil
 	}
 
-	logging.Debugf("%s: searching for model='%s', query='%s', threshold=%.3f",
-		logPrefix, model, shortenHybridQuery(query), threshold)
+	logging.Debugf("%s: searching for model='%s', query=%s, threshold=%.3f",
+		logPrefix, model, logging.ContentDescriptor(query), threshold)
 
 	queryEmbedding, err := h.generateEmbedding(query)
 	if err != nil {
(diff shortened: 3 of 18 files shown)


AayushSaini101 (approved):
LGTM thanks : )

PR #2708, opened 2026-07-28
[Docs] harden contributor leaderboard identity resolution against silent duplicate entries

theohsiung:
## Summary

#2699 reported `theohsiung` and `wilsonwu` each showing up as two separate leaderboard entries — one `github:` keyed, one `email:` keyed — because some of their commits (personal Gmail addresses, no GitHub noreply email to fall back on) couldn't be resolved to a GitHub login at generation time. #2702 already fixed the *visible* symptom via a plain data refresh, but that was incidental: enough time had passed for GitHub to link both emails to their accounts, not because anything in the resolution logic changed. The same split can recur for any future contributor in the same situation, with zero visibility until someone notices a duplicate row on the deployed site.

This PR hardens the generator itself, without introducing any fuzzy name-matching or auto-merge (that risks attributing one contributor's commits to a different person):

1. **`scripts/lib/identity-audit.mjs`** (new, unit tested with `node:test`) — pure `findUnresolvedIdentities()` helper that flags any `email:*` identity key with 2+ commits.
2. **`generate-contributor-rank.mjs`**:
   - `readAssociatedPullAuthor()` now logs both failure modes it previously swallowed silently: a thrown error, and a legitimate "no PR found" result. Previously neither left any trace of why a commit fell back to an `email:*` key.
   - After building the "all time" contributor map, run `findUnresolvedIdentities()` over it and `console.warn` anything it flags — visible at generation time (local run or CI logs) instead of requiring someone to eyeball the leaderboard.
   - Added `theohsiung` and `wilsonwu` to `identityOverrides`, since both depend entirely on GitHub's account/email linking (no noreply-email fallback) and could regress again in principle.
3. Data refresh exercising the above — no visible change (already fixed by #2702), confirms the new overrides and audit don't misfire on real data.

Closes #2699 #2704

## Test plan

- [x] `npm test` (new `node:test` suite for `identity-audit.mjs`): 5/5 passing, covering email-vs-github key, threshold boundary, custom threshold, and bot exclusion
- [x] Ran the generator with a valid `GH_TOKEN`: confirmed the new "No associated PR author found" / PR-lookup-failure warnings print correctly, and no "Unresolved GitHub identity" warning fires (both `theohsiung` and `wilsonwu` still resolve to a single `github:` entry each)
- [x] `npx eslint` clean on all touched/new files
- [x] Kept scope to the generator script + its new test module; reverted an incidental `committerActivity.generated.ts` regeneration unrelated to this fix

PR #2720, opened 2026-07-28
[Bug] Merge email-keyed leaderboard entries into matching contributors

### LINKED ISSUE #2699: bug: Contributor leaderboard shows duplicate/split entries for the same contributor
### Describe the bug

The contributor leaderboard (`website/src/data/contributorRank.generated.ts`) shows the same contributor as two separate ranked entries in the same range when some of their commits resolve to a GitHub login and others don't.

Example, `v03ToNow` range:

```json
{ "rank": 2, "name": "Theo Hsiung", "login": "theohsiung", "key": "github:theohsiung", "commits": 34 }
{ "rank": 12, "name": "theohsiung", "key": "email:theobear870924@gmail.com", "commits": 5 }
```

Same split for `theohsiung` in the `all` range (38 vs 5), and for a second contributor, Wilson Wu (`iwilsonwu@gmail.com` split from their GitHub-keyed entry).

### To Reproduce

Search `contributorRank.generated.ts` for `theohsiung` — two objects in the same range array, one keyed `github:theohsiung`, one keyed `email:theobear870924@gmail.com`.

### Expected behavior

One ranked entry per human contributor per range, commits summed across all resolution paths.

### Additional context

Root cause: `resolveIdentity()` in `website/scripts/generate-contributor-rank.mjs` (~1139-1198) resolves each commit through a priority chain (`identityOverrides` table → GitHub Commits API login → PR-author lookup → noreply-email pattern), falling back to `key: email:<address>` when all fail — and nothing merges an `email:*` key back into a matching `github:*` key afterward.

Verified live: the affected commits resolve `commit.author.login` correctly via the GitHub API today, so at generation time GitHub likely hadn't yet linked those Gmail addresses to the accounts as verified emails — a timing issue, not an API-call bug.

Two-tier fix:
1. **Immediate**: add `theobear870924@gmail.com` and `iwilsonwu@gmail.com` to `identityOverrides` (~lines 116-330), same as the 34 contributors already there.
2. **Structural**: add a reconciliation pass after commit collection that re-resolves any leftover `email:*` keys and merges them into a matching `github:*` entry if one is found — so this doesn't require a manual override every time a new contributor's account-linking hasn't caught up by generation time.

Minor related gap: the PR-author fallback's catch block (~1245-1249) silently swallows errors with no `console.warn`, unlike the script's other fallback paths — worth logging so future splits are easier to audit.

Mohith26:
Fixes #2699

## What

The contributor leaderboard splits one human into two ranked entries when some commits resolve to a GitHub login and others only to an email (e.g. theohsiung ranked twice in the same range). Root cause: `resolveIdentity()` has no reconciliation between `github:*` and `email:*` keys.

This adds a deterministic post-collection pass that folds leftover email-keyed entries into a matching github-keyed entry when linked by commit-author evidence the script already collects (shared author email, shared author name, or name equal to the login), summing commits and widening the date range. Ambiguous evidence shared by multiple github-keyed entries is never used, bots are excluded, and no new GitHub API calls are made.

The generated data file is not touched here; it regenerates on the next scheduled run.

## Tests

New `node --test` suite in `scripts/lib/` (the repo's existing convention), including the exact theohsiung split from the issue (34 + 5 must merge into one 39-commit entry), non-merge cases, and immutability. 4 of 8 tests fail against the old pass-through behavior; 13/13 pass with the fix. `npm run lint` clean. Both commits are DCO signed-off.

Credit to @theohsiung for the report in #2699.

### DIFF
--- a/website/scripts/generate-contributor-rank.mjs
+++ b/website/scripts/generate-contributor-rank.mjs
@@ -3,6 +3,7 @@ import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs'
 import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { findUnresolvedIdentities } from './lib/identity-audit.mjs'
+import { mergeEmailKeyedContributors } from './lib/identity-merge.mjs'
 
 const scriptDir = dirname(fileURLToPath(import.meta.url))
 const repoRoot = resolve(scriptDir, '..', '..')
@@ -636,15 +637,29 @@ function collectContributorStats(rows) {
       commits: 0,
       firstCommitDate: row.date,
       latestCommitDate: row.date,
+      emails: new Set(),
+      names: new Set(),
     }
 
     current.commits += 1
     current.firstCommitDate = minDate(current.firstCommitDate, row.date)
     current.latestCommitDate = maxDate(current.latestCommitDate, row.date)
+
+    const normalizedEmail = normalizeEmail(row.email)
+    const normalizedName = normalizeName(row.name)
+
+    if (normalizedEmail) {
+      current.emails.add(normalizedEmail)
+    }
+
+    if (normalizedName) {
+      current.names.add(normalizedName)
+    }
+
     byContributor.set(identity.key, current)
   }
 
-  return byContributor
+  return mergeEmailKeyedContributors(byContributor)
 }
 
 function rankContributorEntries(byContributor, totalCommits, newContributorKeys, reviewStats = new Map()) {
--- a/website/scripts/lib/identity-merge.mjs
+++ b/website/scripts/lib/identity-merge.mjs
@@ -1,0 +1,144 @@
+function normalizeValue(value) {
+  return String(value ?? '').trim().toLowerCase()
+}
+
+function pickEarlierDate(left, right) {
+  if (!left) {
+    return right
+  }
+
+  if (!right) {
+    return left
+  }
+
+  return left < right ? left : right
+}
+
+function pickLaterDate(left, right) {
+  if (!left) {
+    return right
+  }
+
+  if (!right) {
+    return left
+  }
+
+  return left > right ? left : right
+}
+
+function normalizedSet(values) {
+  const normalized = new Set()
+
+  for (const value of values ?? []) {
+    const candidate = normalizeValue(value)
+
+    if (candidate) {
+      normalized.add(candidate)
+    }
+  }
+
+  return normalized
+}
+
+const AMBIGUOUS = Symbol('ambiguous evidence')
+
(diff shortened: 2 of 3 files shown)


drivebyer (commented):
Nice find — the mechanism is real. A few notes from digging in:

- Both cases in #2699 have already self-healed (the API resolves those gmails today). The one live split left is `aby / chenzw0521@gmail.com`, whose account was deleted — its PRs show `user: ghost`, so both the commit API and the PR fallback come back empty. Your pass does fix that one.
- Concern: matching on git author *name* can merge different people — two contributors both committing as `root` collapse into one entry. `AMBIGUOUS` only guards github↔github collisions.
- Suggestion: keep just the name-equals-login path (that's what fixes both cases) and drop the name/email matches. Much smaller, and a lot harder to get wrong.

Mohith26:
Good points, thanks - done in 6df47db: dropped the email/name evidence matching and kept only the name-equals-login path, updated the tests to pin the deliberate non-merges (shared display name or shared email evidence alone no longer merge). All 8 tests green.

drivebyer:
Thanks for the quick turnaround! One correction to my earlier note, after re-checking the data: the pass as-is doesn't catch the one remaining live split — the email entry's author name is `aby`, but the GitHub-keyed entry is `github:aby42` (via the noreply regex), so name-equals-login never fires. Meanwhile #2708 has closed #2699 with the override + audit approach, deliberately avoiding auto-merge.

Suggestion: repurpose this PR as

1. an `identityOverrides` entry: `chenzw0521@gmail.com` → `aby42`;
2. lowering the audit's `minCommits` to 1 so single-commit splits surface.

Same outcome, much less machinery.

Xunzhuo (changes requested):
Thanks for iterating on this. I re-reviewed the latest head against current main. I cannot approve the remaining name-equals-login reconciliation: Git author names are user-controlled, so this can silently attribute an unrelated email-keyed commit to an existing GitHub account.

It also does not fix the current live split: the entries are `aby` / `chenzw0521@gmail.com` and `github:aby42`, so the names do not match. #2708 has already closed #2699 using explicit overrides plus auditing.

I would be comfortable approving a minimal repurpose: remove the merge helper and its tests/integration, add an explicit `chenzw0521@gmail.com` → `aby42` override, lower the unresolved-identity audit default to one commit, add the corresponding regression test, and verify regenerated data contains one two-commit `github:aby42` entry. Please also rebase, update the title/body to current `[Docs]` conventions, and use `Related #2699`.

Xunzhuo:
Closing this as superseded by #2708, which fixed the live split with explicit overrides plus audit coverage and already closed #2699. The remaining name-equals-login heuristic here would still risk misattribution and does not match the affected entry.

PR #2724, opened 2026-07-29
[Router] fix recipe-blind routing reads and pin every read scope

### LINKED ISSUE #2723: [Router] Whole-surface routing reads silently see only the default profile after named recipes
Part of #2331, follow-up to #2612 / #2613.

Those PRs added named recipes next to the flat `IntelligentRouting` decisions field, but ~15 whole-surface read sites kept reading the flat field — silently seeing only the default profile. Concretely: a memory plugin configured on a recipe decision passes validation but never enables the runtime, five contract validators skip recipe decisions entirely, and the replay recorder for a recipe decision is never registered. Every failure is silent, and every new read site has to pick the right scope by hand.

Named recipes are unreleased, so no released configuration is affected — but the reference config already promotes them.

Scope of the fix (#2724):

- rename the flat field `Decisions` → `DefaultDecisions` so the default-profile scope is visible at every read site (config format and API responses unchanged);
- repoint the whole-surface reads at `AllRoutingDecisions()` / `GetDecisionByName()`; sites whose scope really is the default profile stay that way, each with its rationale recorded;
- guard regressions with an allowlist test: a new default-profile read fails CI until it either switches to a whole-surface accessor or states its scope reason.

Out of scope, deferred: removing the flat field so recipes become the single source. The field and the default recipe are one aliased slice (pinned by an invariant test), not two copies — removal buys no additional correctness today and costs an emitter compatibility layer plus a wholesale fixture rewrite. Revisit under #2331 if the recipes-only layout ever forces it.

drivebyer:
Closes #2723 (part of #2331). Follow-up to #2612 / #2613; the flat-field removal sketched alongside that issue is deferred, see Non-goals.

## Purpose

#2612/#2613 added named recipes next to the flat `IntelligentRouting` decisions field, but left ~15 whole-surface read sites reading only the default profile. The failures were silent: a memory plugin configured on a recipe decision passed validation and never enabled the runtime, five contract validators skipped recipe decisions entirely, and the replay recorder for a recipe decision was never registered. Named recipes are unreleased, so no released configuration is affected — but the reference config already promotes them.

This PR is the scoped correction, three layers:

1. **Rename** the flat field `Decisions` → `DefaultDecisions` (mechanical, 106 files). The yaml/json tags keep the config format and the `/info/classifier` wire key byte-identical; the new name makes every default-profile read visible at its call site.
2. **Fix the read sites**: plugin enablement, the five validators, lookups, and the replay recorder now read the whole routing surface via `AllRoutingDecisions()` / `GetDecisionByName()`. Sites whose scope really is the default profile stay that way, each with its rationale recorded.
3. **Guard the future**: an allowlist test pins the complete set of production files that may read the flat field, one scope rationale per entry. A new default-profile read fails CI until it either switches to a whole-surface accessor or states its reason in front of the reviewer.

A review pass over the branch then tightened the scope decisions themselves:

- The process-level selection singleton (`elo`/`multi_factor`/… configs) and the classify-API reasoning/recommendation paths deliberately stay on the default profile — widening them made the global config dependent on YAML recipe order and let recipes leak into an API that previews default-profile routing.
- `candidate_set: tier` learning candidates now follow the request's recipe scope instead of unioning every profile (which could route past a recipe's entrypoint).
- Cross-recipe decision names are now validated case-insensitively; the fold lookup's uniqueness assumption was otherwise false.
- Cleanups: the memory-enablement predicate sank into `config` (was copied in three packages), allocation-free `CountRoutingDecisions()`, and the mirror invariant asserts slice aliasing rather than tautological `DeepEqual`.

**No user-visible behavior change** on any released surface: single-profile configs are bit-for-bit unaffected, and the only behavior changes sit on the unreleased named-recipe surface.

## Non-goals (deferred, tracked)

- **Removing the flat field** ("single source of truth"): the field and the default recipe are one aliased slice, pinned by an invariant test — removal buys no additional correctness today and costs an emitter compatibility layer plus a wholesale fixture rewrite. Revisit in #2331 if the recipes-only layout ever forces it.
- Dashboard recipe awareness: registered as TD045, owned by pl-0038 T12.
- DSL round-trip for recipes/entrypoints: pl-0038 T10.

## Test Plan

- Full main-module `go test` (`pkg/cache` excluded: env-dependent backends), including the cgo-linked `extproc`/`classification`/`services`/`apiserver` suites.
- `make agent-lint` over all changed files, `make check-go-mod-tidy`, markdownlint on the docs.
- New coverage: `recipe_read_sites_test.go` pins every corrected read site; `default_decisions_allowlist_test.go` pins the deliberate default-profile reads (and also matches the `DefaultRecipe()` spelling); `recipes_invariant_test.go` pins the flat-field/default-recipe aliasing; a new loader test pins the case-insensitive cross-recipe name check.

## Test Result

All of the above pass locally; every commit builds and passes lint, so the sequence bisects.

---
<details>

</details>

### DIFF
--- a/dashboard/backend/handlers/topology_response.go
+++ b/dashboard/backend/handlers/topology_response.go
@@ -177,7 +177,7 @@ func appendEvaluatedRulesFromConfig(result *TestQueryResult, configPath string) 
 	}
 
 	matchedSignalNames := buildMatchedSignalNameSet(result.MatchedSignals)
-	for _, decision := range parsedConfig.IntelligentRouting.Decisions {
+	for _, decision := range parsedConfig.IntelligentRouting.DefaultDecisions {
 		if result.MatchedDecision != "" && decision.Name == result.MatchedDecision {
 			continue
 		}
--- a/docs/agent/plans/pl-0038-entrypoints-recipes.md
+++ b/docs/agent/plans/pl-0038-entrypoints-recipes.md
@@ -15,11 +15,12 @@ of it.
   `recipes` adds named additional profiles. A recipe named `default` may only
   appear when the top-level `routing` block carries no profile of its own.
 - Internal representation: normalized `Recipes`/`Entrypoints` on `RouterConfig`;
-  the default recipe's decisions stay bridged into the flat `Decisions` field so
-  current read sites keep working unchanged, while the flat
-  `Signals`/`Projections` fields hold the global registry (the union across
-  recipes, per the issue's "signal registry stays global" non-goal) so one
-  classifier evaluates any recipe's rules.
+  the default recipe's decisions stay bridged into the flat field (renamed
+  `Decisions` → `DefaultDecisions` in #2724 so the default-profile scope is
+  visible at every read site), while the flat `Signals`/`Projections` fields
+  hold the global registry (the union across recipes, per the issue's "signal
+  registry stays global" non-goal) so one classifier evaluates any recipe's
+  rules.
 - Request path: resolve model name → entrypoint → recipe before signal
   evaluation in `pkg/extproc`; decision evaluation reads the per-request recipe.
 - Validation: structural recipe/entrypoint checks first, then cross-surface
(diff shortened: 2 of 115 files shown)


drivebyer:
Superseded by #2741, which completes the rest of #2331 from the issue author and redesigns the recipe model along the way (recipe-local decision names, per-recipe classifier/selector scopes). That branch also fixes the read-scope defects this PR targeted — memory-plugin enablement, replay recorder registration, the contract validators — on the new model, so the whole-surface accessors, the cross-recipe name validation, and the allowlist guard built here no longer apply. The defect catalogue in #2723 should now be tracked against #2741.

Issue #2691, opened 2026-07-27
[Bug] Prevent semantic-cache false hits on negated or antonym queries


WUKUNTAI-0211:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The semantic cache decides a hit by bi-encoder cosine similarity against a
`similarity_threshold` (`pkg/cache/cache_interface.go`,
`pkg/extproc/req_filter_cache.go:118`). Bi-encoder embeddings score
**negated / antonym** query pairs — opposite meaning, near-identical surface
form — *above* the deployed thresholds. The cache then serves a **semantically
opposite** cached answer: a user asking how to **disable** a feature can be
served the cached answer for how to **enable** it.

This is a correctness defect, not just a tuning nuisance, and it is **not**
fixable by threshold calibration (cf. #2341): a genuine paraphrase can score
*lower* than a semantic opposite, so raising the threshold removes legitimate
recall before it removes the negation false-hit.

### To Reproduce

Measured on the router's **own production embedder** — `mmbert` via
candle-binding, the exact path used by the cache
(`vectorstore.CandleEmbedder{modelType:"mmbert"}.Embed` →
`GetEmbeddingWithModelType(text, "mmbert", 384)`), cosine over the 384-dim
output:

| Negated / antonym pair (opposite meaning) | cosine |
|---|---|
| "How to turn **on** dark mode?" vs "How to turn **off** dark mode?" | 0.915 |
| "Is port 6379 **open** by default?" vs "…**closed** by default?" | 0.900 |
| "Should I **commit** this change?" vs "Should I **not** commit this change?" | 0.869 |
| "Does this feature **require** a license?" vs "…does **not require**…?" | 0.853 |
| "Is caching **enabled** in production?" vs "…**disabled**…?" | 0.848 |
| "Is the API rate limit currently **active**?" vs "…**inactive**?" | 0.847 |
| "Can I **increase** my storage quota?" vs "…**decrease**…?" | 0.838 |
| "How do I **enable** two-factor auth?" vs "…**disable**…?" | 0.823 |

False-hit counts against the `similarity_threshold` values shipped in-repo
(`config/plugin/semantic-cache/*.yaml`, `config/config.yaml`
`stores.semantic_cache`):

- `0.80` (memory.yaml) → **8/12** negation pairs false-hit
- `0.86` (config.yaml stores) → **3/12** false-hit
- `0.92` (high-recall) → 0/12, but this cutoff also rejects genuine paraphrases
[...]

PR #2728, opened 2026-07-29
[Router] guard in-memory semantic cache polarity

### LINKED ISSUE #2727: [Bug] Guard the in-memory semantic cache against polarity mismatches
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

### To Reproduce

### Expected behavior

### Impact

### Additional context

WUKUNTAI-0211:
Closes #2727
Refs #2691

## Problem

The in-memory semantic cache can serve an opposite-meaning answer. A bi-encoder
scores `How do I turn on dark mode?` against `How do I turn off dark mode?` at
`0.915`, well above the shipped `0.80` threshold, while a genuine `enable 2FA` /
`turn on 2FA` paraphrase scores only `0.673`. Raising the threshold trades the
false hit for lost recall, so polarity has to be checked separately from
embedding similarity.

## What main already has, and what it is missing

#3075 landed the NLI polarity tier together with the `polarity_guard` config
contract. That contract already names a second tier:

- `config/config.yaml` ships `mode: lexical`, described in its own comment as
  the "model-free default".
- `config.PolarityGuardConfig.NormalizedMode()` defaults to `lexical`.
- `website/docs/.../stores-and-tools.md` documents `lexical` as "always on and
  needs no model".
- `polarity_nli.go` calls the lexical tier "the unconditional floor", and
  `polarityGuardTierNLI` exists so NLI telemetry can "be told apart from the
  lexical tier's events".

The tier itself was never implemented. `PolarityGuardOptions` carries only
`UseNLI`, and `router_components.go` maps config to cache options by reading
`UsesNLI()` alone, so today:

| mode | actual behavior |
|---|---|
| unset / `lexical` (**the shipped default**) | no polarity check at all |
| `nli` | verified, but requires the hallucination explainer model |
| `lexical+nli` | identical to `nli` |

This PR implements that floor.

## Fix

- Add the model-free lexical tier: a rejection when two near-identical queries
  differ only by a negation cue or a known antonym swap.
- Run it from the existing winning-candidate hook. `applyPolarityGuard` checks
  the single best above-threshold candidate, outside the cache lock, and runs
  the cheap tier first — a lexical rejection stands on its own and never pays
  for a model call.
- Tokenize the incoming query once per lookup. `polarityMismatchTokens` takes
  that tokenization, so a caller comparing several candidates normalizes once.
- Label rejections on the shared `cache_negation_reject` event with the tier
  that rejected, so lexical and NLI rejections stay distinguishable.

The multi-candidate guard and the HNSW linear-scan fallback from the earlier
revision of this PR are gone: the guard now sees one candidate per lookup and
adds no unbounded scan under the read lock.

### Ambiguous cues are read in context

The tier cannot be switched off, so a false positive costs every deployment its
recall with no opt-out. Cues that are ambiguous in English are therefore gated
on their neighbours rather than matched bare:

- `no` does not negate inside a hyphenated compound (`no-code`, `no-cache`) or
  before a reference number, so `Show invoice no. 123` and `Show invoice
  no. A123` stay paraphrases of `Show invoice 123` / `Show invoice A123`, while
  `Is there no way to disable it?` still rejects.
- `on`/`off` flip state only next to a toggle verb or at the end of a query, so
  `routing based on embeddings` stays a paraphrase of `routing based off
  embeddings`, while `turn on` / `turn off` and `Is dark mode on?` / `off?`
  still reject.

Auxiliaries are dropped and a conservative plural stem applied before the diff
is taken, so do-support negation reaches its cue instead of falling outside the
two-token gate: `This change works` / `This change doesn't work` and
`It's safe to deploy` / `It isn't safe to deploy` are now rejected.

## Behavior change on a shared path

With the floor in place, a default-configured cache stops serving
opposite-meaning hits it previously served. Three consequences worth calling out
for review:

- The model-backed regression in `negation_regression_test.go` never actually
  ran. Since #3768 the cache consumes a prepared `EmbeddingProvider` instead of
  initializing a model from `EmbeddingModel`, so with the real checkpoint
  present every pair failed at `AddEntry` with `cache embedding provider was
  not prepared` and no assertion was reached. `prepareMmbertProvider` now
  prepares a real mmBERT provider through `native.Runtime.Embedding` and hands
  it to the cache, and a checkpoint that fails to prepare fails the test
  instead of skipping it.
- The NLI unit tests in `polarity_nli_test.go` could no longer use an
  enable/disable swap as their incoming query — the floor rejects it before the
  verifier is reached, so their assertions about verifier call counts,
  fail-open and cancellation would silently stop being exercised. They now
  drive a paraphrase the floor passes, with the fake verifier supplying the
  contradiction, so each tier is tested in isolation.
- A lexical rejection is final; the NLI tier cannot overturn it. That is what
  makes "no model call for the obvious cases" real, and it is why the ambiguous
  cues above are gated rather than matched bare.

## Scope

- In-memory backend only; remote and hybrid backends are unchanged.
- The tier does not cover cue-less semantic opposites, polarity expressed only
  through word order or scope, or non-English polarity. Those remain the NLI
  tier's job, and the docs now say so.
- The model-backed regression is a focused smoke set, not a PAWS/QQP
  calibration corpus.

## Verification

Rebased onto `origin/main` (`26c38b140`) with no conflicts;
`inmemory_cache_search.go` is byte-identical to main apart from the one call
site.

Both model-backed regressions were run against real checkpoints
(`llm-semantic-router/mmbert-embed-32k-2d-matryoshka`,
`tasksource/ModernBERT-base-nli`) fetched locally. Before the fixture change
above, the mmBERT regression exercised nothing:

```
--- FAIL: TestNegationFalseHitRegressionInMemory (3.20s)
    AddEntry("How to turn on dark mode?"): failed to generate embedding:
    cache embedding provider was not prepared
```

With the prepared provider wired in, every above-threshold negation pair is
rejected and every paraphrase control is still served:

```
--- PASS: TestNegationFalseHitRegressionInMemory (4.98s)
    polarity guard exercised on 8/8 above-threshold negation pairs
    paraphrase control exercised on 3/3 above-threshold pairs
```

To confirm the regression is load-bearing rather than vacuously green, the
lexical tier was neutralized in `applyPolarityGuard` (throwaway, not
committed): all eight pairs are then served as false hits at sim 0.94–0.98 —
`Is port 6379 closed by default?` returns the answer cached for `Is port 6379
open by default?` at 0.9796 — while the three paraphrase controls keep
passing.

```
go vet ./pkg/cache/                                          OK
golangci-lint --config tools/linter/go/.golangci.yml \
  --new-from-rev origin/main ./pkg/cache/...                 0 issues
gofmt -l src/semantic-router/pkg/cache/                       clean
go test ./pkg/cache/ -run 'Polarity|Negation' -count=1        ok (16 pass, 0 skip)
go test ./pkg/cache/ -count=1                                 ok (157 pass, 35 skip)
go test ./pkg/extproc/ ./pkg/config/ -count=1                 ok
```

The 35 skips are the Milvus/Valkey/Qdrant/Redis/Hybrid integration tests, which
skip without their containers. The Go unit job does not download model
checkpoints, so both model-backed regressions skip in CI — the evidence above
is local.

Benchmarks through the production lookup finisher, plus a direct one isolating
the tier:

```
BenchmarkPolarityTierAlone/serves-20     2019 ns/op   1056 B/op   10 allocs/op
BenchmarkPolarityTierAlone/rejects-20    2010 ns/op   1056 B/op   10 allocs/op
BenchmarkPolarityGuardServesHit-20       3400 ns/op   1552 B/op   18 allocs/op
BenchmarkPolarityGuardRejects-20         4469 ns/op   2113 B/op   30 allocs/op
BenchmarkPolarityGuardLongQuery-20      11722 ns/op  11792 B/op   30 allocs/op
```

Roughly 2 µs and 10 allocations per lookup, against the ~70 ms per verified hit
the NLI tier costs on CPU. The `no-` compound rewrite is behind a
`strings.Contains` guard, so the regexp stays off the common path — without it
the tier measured 3.4 µs and 16 allocs.


WUKUNTAI-0211:
@Xunzhuo @drivebyer, could you take a look when you have time?

The current in-memory lookup returns the highest embedding-similarity candidate above the configured threshold. It checks partition, user scope, and TTL, but not whether the incoming and cached queries contradict each other.

Bi-encoders can score near-identical negation or antonym pairs highly. For example, `turn on` versus `turn off` scored `0.915`, so the cache can serve instructions for the opposite action.

Raising the threshold is not sufficient because genuine paraphrases can score lower than these hard negatives. This PR adds a conservative, model-free polarity guard before candidate selection.

Thanks for taking a look.

Xunzhuo:
@WUKUNTAI-0211 looking into it, thanks!

Xunzhuo (changes requested):
Thanks for tackling this. A lightweight lexical veto is a reasonable L1 mitigation, and I do not think this PR needs to introduce NLI/cross-encoder verification or solve every form of semantic contradiction. Before merging, though, I think the behavior introduced in this PR needs a few changes:

1. **Constrain ambiguous lexical cues.** The context-free token rules reject legitimate near-paraphrases such as `Show invoice no. 123` / `Show invoice 123` and `routing based on embeddings` / `routing based off embeddings`. Please add controls for these cases and restrict ambiguous cues such as `no` and `on/off` to contexts where they actually express polarity.

2. **Either cover common grammatical negation or narrow the stated contraction support.** For example, `This change works` / `This change doesn't work` and `It's safe to deploy` / `It isn't safe to deploy` exceed the two-token diff gate after only the negative contraction is expanded, so the guard can still allow the opposite cached answer. Please add natural-form regression cases and either normalize the common auxiliary/inflection differences or document this boundary explicitly rather than broadly claiming contraction handling.

3. **Bound and measure the new hot path.** After an HNSW polarity rejection, linearly scanning the full cache while holding `RLock` defeats HNSW for that request; `max_entries=0` is unlimited and an HNSW example allows 5,000 entries. The incoming query is also tokenized and allocated once per candidate. Please pre-tokenize it once, add a benchmark through the production candidate/guard path, and either use bounded over-fetch/fallback or provide evidence that the full-scan latency and lock contention are acceptable at supported sizes.

4. **Split the dashboard E2E image changes from this cache fix.** They are unrelated to the polarity behavior. In addition, the fixed mutable `dashboard:e2e-test` tag ignores `-image-tag`; on a reused Kind cluster, loading a new image under the same tag does not change the Deployment pod template, so the existing pod can continue running the previous checkout.

5. **Align rejection telemetry with the documented semantics.** A higher-ranked candidate can be rejected and a lower-ranked candidate returned as a hit, but that rejection currently emits no `cache_negation_reject` event. Either emit rejection telemetry independently of final hit/miss accounting or clarify that the event represents only polarity-caused final misses.

The broader pairwise semantic verifier, remote cache backends, multilingual/scope-sensitive negation, and embedding hard-negative training are appropriate follow-ups rather than blockers for this bounded L1 change.

drivebyer:
@WUKUNTAI-0211 nice fix, plz fix the conflict : )

drivebyer:
Hi @WUKUNTAI-0211, friendly ping on this PR, could you please resolve the merge conflict

WUKUNTAI-0211:
Hi @drivebyer , I will take a look ASAP. Thanks for the ping!

Xunzhuo (commented):
Thanks for the updates. The hot-path blocker remains: each above-threshold candidate retokenizes the full incoming query, and an HNSW rejection can still trigger a full linear scan under the cache read lock. Main now includes the bounded winning-candidate NLI tier from #3075, and this branch conflicts with that implementation. Please rebase and either drop the overlapping lexical path or replace it with bounded, pretokenized candidate handling plus coverage for ambiguous cues.

FAUST-BENCHOU (changes requested):
pls resolve conflicts

WUKUNTAI-0211:
Rebased onto `main` and rebuilt this on #3075's winning-candidate hook rather than beside it.

On drop versus replace: #3075 shipped the `polarity_guard` contract along with the tier, and that contract already names a lexical tier — `config/config.yaml` ships `mode: lexical`, `NormalizedMode` defaults to it, `PolarityGuardOptions` calls it the unconditional floor, and the docs describe it as always on. The tier behind it was never implemented. `PolarityGuardOptions` carries only `UseNLI`, and `router_components.go` maps config to cache options by reading `UsesNLI()` alone, so a default deployment runs no polarity check at all and `lexical+nli` is indistinguishable from `nli`. That gap is why I replaced the path instead of dropping it.

**Hot path.** `considerSearchCandidate` and the HNSW linear-scan fallback are gone. The tier runs from `applyPolarityGuard` on the single best above-threshold candidate, outside the cache lock, ahead of `applyPolarityNLI` — no full scan under `RLock`, and a lexical rejection never reaches the model. `polarityMismatchTokens` takes the incoming query already tokenized, so it is normalized once per lookup. `BenchmarkPolarityTierAlone` puts the tier at roughly 2 µs and 10 allocations; `BenchmarkPolarityGuardServesHit` and `BenchmarkPolarityGuardLongQuery` run the same thing through `finishFindSimilarSearch`. The `no-` compound rewrite sits behind a `strings.Contains` guard so its regexp stays off the common path — without the guard the tier measured 3.4 µs and 16 allocations.

**Ambiguous cues.** `tokenizeForPolarity` keeps token order now, so `no` and `on`/`off` are read in context. `no` does not negate inside a hyphenated compound (`no-code`, `no-cache`) or before a reference number, so `Show invoice no. 123` and `Show invoice no. A123` are paraphrases again while `Is there no way to disable it?` still rejects. `on`/`off` flip state only next to a toggle verb or at the end of a query, so `routing based on embeddings` stays a paraphrase of `routing based off embeddings`, while `turn on` / `turn off` and a trailing `Is dark mode on?` / `off?` reject. All of them are in `TestPolarityMismatch`.

**Grammatical negation.** Auxiliaries are dropped and a conservative plural stem applied before the diff, so `This change works` / `doesn't work`, `It's safe to deploy` / `It isn't safe to deploy`, `It has worked` / `hasn't worked` and `I can deploy` / `I cannot deploy` reach their cue instead of falling outside the two-token gate — rather than narrowing the contraction claim.

**Telemetry.** With one candidate per lookup a rejection is the final miss, so the two agree by construction; `cache_negation_reject` now carries `tier`, so lexical and NLI rejections stay distinguishable. The scenario behind that point — a higher-ranked candidate rejected while a lower-ranked one is served — cannot arise on the bounded path, since the guard only ever sees the winner. The cost is that a rejected winner is a miss even when a valid lower-ranked entry exists; that is main's existing shape for the NLI tier, not something this PR introduces. The dashboard E2E image changes were reverted out earlier — this PR is the cache plus the docs the tier makes accurate.

Two things to flag. Because the floor runs first, the NLI unit tests in `polarity_nli_test.go` could no longer use an enable/disable swap as their incoming query: the floor rejects it before the verifier is reached, and their assertions about verifier call counts, fail-open and cancellation would quietly stop being exercised. They now drive a paraphrase the floor passes, with the fake verifier supplying the contradiction. And a lexical rejection is final — the NLI tier cannot overturn it, which is what makes "no model call for the obvious cases" real and why the ambiguous cues above are gated rather than matched bare.

If you would still rather drop the tier after that, it closes this PR, but it leaves main shipping a default mode that does nothing, so it needs a follow-up taking `lexical` out of the config contract and the docs. I am happy to open that one instead if you prefer it.

cc @yaojiejia — the `mode: lexical` default from #3075 is the part that never had an implementation behind it.

src/semantic-router/pkg/cache/negation_regression_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func newMmbertCache(t *testing.T) *InMemoryCache {
+	t.Helper()
+	return NewInMemoryCache(InMemoryCacheOptions{
+		SimilarityThreshold: negationRegressionThreshold,
+		MaxEntries:          16,
+		TTLSeconds:          3600, // 0 means "not cached"; use a long TTL so entries stay live
+		Enabled:             true,
+		EvictionPolicy:      FIFOEvictionPolicyType,
+		EmbeddingModel:      "mmbert",
```
Xunzhuo, line 99: **[P2] Pass the prepared mmBERT provider to the regression cache**

`EmbeddingModel` no longer prepares cache inference: `NewInMemoryCache` uses `options.EmbeddingProvider`, and `invokeCacheEmbedding` rejects nil. With the real mmBERT checkpoint present, this test loads the model successfully but fails at its first `AddEntry` with `cache embedding provider was not prepared`, so none of the polarity or paraphrase assertions run. Supply a prepared real mmBERT provider in this fixture and rerun the model-backed test; a provider-only control reaches all eight rejection pairs and three paraphrase controls.

Xunzhuo (changes requested):
The bounded winning-candidate guard resolves the earlier hot-path and ambiguous-cue concerns; focused lexical tests and benchmarks pass. Running the new regression with the real mmBERT checkpoint exposes a fixture failure before any guard assertion. Supplying only the missing prepared provider makes all eight above-threshold polarity rejections and all three paraphrase controls pass. Please wire that provider into the committed regression; the current CI run skips this test because its model fixture is absent.

src/semantic-router/pkg/cache/negation_regression_test.go
WUKUNTAI-0211: Confirmed, and fixed — this was a real hole, thanks for running it against the checkpoint.

`prepareMmbertProvider` now builds a candle mmBERT provider through `native.Runtime.Embedding` and passes it as `EmbeddingProvider`; the legacy `InitMmBertEmbeddingModel` global init is gone, since it no longer reaches the cache after #3768. I reproduced your failure on the old fixture — `AddEntry` returning `cache embedding provider was not prepared`, nothing after it running — and on the new one all eight above-threshold negation pairs reject and all three paraphrase controls still serve. A checkpoint that is present but fails to prepare now fails the test instead of skipping, so the fixture cannot go quiet again.

I also checked the regression is load-bearing rather than vacuously green: neutralizing the lexical tier in `applyPolarityGuard` serves all eight pairs as false hits at 0.94–0.98.

On CI you are right, and it is wider than this test — the Go unit job downloads no checkpoints, so `TestPolarityNLIRegression` skips there too. Would you rather I wire the mmBERT checkpoint into a CI job in this PR, or keep the local proof and track CI coverage separately?

src/semantic-router/pkg/cache/negation_regression_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+// The cache no longer prepares inference itself; it consumes an already
+// prepared provider. A missing checkpoint skips, but a checkpoint that fails to
+// prepare fails: a silently unprepared fixture would leave every assertion
+// below unexercised.
+func prepareMmbertProvider(t *testing.T) *native.EmbeddingProvider {
+	t.Helper()
+	modelPath := findMmbertModel()
+	if modelPath == "" {
+		t.Skip("mmbert embedding model not found (models/mmbert-embed-32k-2d-matryoshka); skipping cgo regression")
```
Xunzhuo, line 68: **[P2] Assign the model-backed test to a provisioned CI profile**

The mandatory Core suite discovers this test, but has no mmBERT checkpoint and no exclusion/profile entry for it. On this exact head, [Core Go, Python and Smoke](https://github.com/vllm-project/semantic-router/actions/runs/35327847513/job/105549576920) aborts with `unexpected mandatory Go result` for `TestNegationFalseHitRegressionInMemory` after this skip. Please register the test with the model-backed validation profile and provide its checkpoint there, so Core does not fail and the native regression actually executes.

Xunzhuo (changes requested):
The prepared-provider fix now passes the real mmBERT regression: all eight above-threshold negation pairs are rejected and all three paraphrase controls still hit. Focused lexical and NLI contract tests also pass. One CI integration blocker remains: Core treats the new model-dependent test as mandatory and fails when it skips. The separate legacy HaluGate corpus was not run because its checkpoint was unavailable.

WUKUNTAI-0211:
Closing as superseded: #3944 landed this guard on `main` (`Fixes #2727`, `Refs #2691`), including the prepared-provider fixture and the model-backed CI profile this branch was still missing, and extends the lexical floor to the remote backends. Thanks @Peterren for carrying it forward.

PR #2736, opened 2026-07-30
fix(config): preserve disabled router replay

### LINKED ISSUE #2735: router_replay.enabled=false is lost during canonical config normalization
### Problem

Canonical v0.3 configuration defaults `global.services.router_replay.enabled` to `true`, while `RouterReplayConfig.Enabled` is tagged with `omitempty` for JSON and YAML.

When `PUT /config/router` normalizes a document that explicitly contains:

```yaml
global:
  services:
    router_replay:
      enabled: false
```

`CanonicalConfigFromRouterConfig` marshals the false value without the `enabled` field. Parsing the normalized document then reapplies the canonical default (`true`). The persisted active configuration therefore enables Router Replay even though the submitted document explicitly disabled it.

This is observable independently from response storage and semantic cache, whose `enabled: false` values survive the same normalization path.

### Expected behavior

An explicit `router_replay.enabled: false` must remain false after canonical export, management API PUT/PATCH normalization, persistence, reload, and readback.

### Proposed fix

Serialize `RouterReplayConfig.Enabled` without `omitempty` and add regression coverage for both canonical export and the management API PUT path.

duchangyong:
Fixes #2735.

## Summary

- serialize `RouterReplayConfig.Enabled` even when it is `false`
- cover canonical export so an explicit disabled value cannot be dropped
- cover `PUT /config/router` normalization and reload

Without this change, canonical normalization omits `enabled: false`; the next parse reapplies the canonical default (`true`) and persists an unexpectedly enabled Router Replay service.

## Tests

- `CGO_ENABLED=0 go test ./pkg/config -count=1`
- repository pre-commit hooks

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_config_deploy_test.go
+++ b/src/semantic-router/pkg/apiserver/route_config_deploy_test.go
@@ -54,6 +54,38 @@ func TestHandleConfigPutReplacesExistingConfig(t *testing.T) {
 	}
 }
 
+func TestHandleConfigPutPreservesDisabledRouterReplay(t *testing.T) {
+	configPath := writeDeployTestBaseConfig(t)
+	payloadCfg := minimalDeployTestConfig("new_route")
+	payloadCfg.RouterReplay = config.RouterReplayConfig{
+		Enabled:      false,
+		StoreBackend: "memory",
+		TTLSeconds:   600,
+	}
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):

Xunzhuo:
Thanks @duchangyong

Issue #2690, opened 2026-07-26
[Epic] Reduce time to the first successful routed request


TwilightTechie:
## Parent Direction

Developer Experience & Ecosystem — #2970

## Goal

Give a new adopter one supported path from installation to a complete, observable routed request before requiring them to understand the full model, configuration, and deployment surface.

## Current-main baseline

The project has a working local-image CLI, a large canonical reference configuration, hosted Playground links, and several local startup paths. It still lacks one maintained first-run profile that combines a minimal config, a successful backend-free request, visible routing evidence, automated smoke coverage, and a reproducible time baseline.

## Current work

- #2572 — resolve the documented hosted Playground credential contract.
- #2744 — ship a minimal validator-valid backend-free config.
- #3184 — repair the dockerless Router plus Envoy development path.
- #3215 — add the one-command evaluation profile, smoke coverage, and time-to-success baseline.

## Scope

- Minimize configuration editing and unnecessary model downloads for the first evaluation.
- Send real OpenAI-compatible requests through a maintained routing path.
- Surface distinct routing outcomes and explain the next model-backed step.
- Keep hosted and local journeys truthful and independently testable.
- Record cold and warm prerequisites when publishing time-to-success results.

## Non-goals

- Treating the evaluation profile as a production configuration.
- Building a second serving architecture.
- Hiding image pulls, downloads, or cached artifacts.
- Replacing reference deployment and security documentation.

## Completion criteria

- One supported command starts the maintained evaluation journey.
- Two requests complete through different routing decisions without an LLM backend.
- The result exposes canonical routing evidence.
- Hosted Playground documentation matches its supported authentication state.
- Local and CI smoke tests protect the journey.
- A reproducible cold and warm time-to-success baseline is published.

TwilightTechie:
Updated the description to reflect the discussion: minimal defaults and a fast containerized evaluation path come first. Docker-free serving and adoption shadow mode are now separated as future work.

PR #2743, opened 2026-08-01
[Router] add minimal validator-valid quickstart config (#2690)

### LINKED ISSUE #2744: [Config] Ship a minimal validator-valid quickstart config for first-run evaluation
## Goal

Ship the minimal maintained configuration required by #2690 so a new user can observe real routing decisions without editing YAML, downloading classifier models, or deploying an LLM backend.

## Proposed Workgroup

## Scope

- Add a small canonical config with model-free keyword routing and at least two distinct logical model outcomes.
- Use fast-response behavior so every maintained probe completes without an upstream model server.
- Keep the file on the strict schema and normal Router validators.
- Add focused config tests proving no classifier artifact or backend is required.
- List the asset in maintained configuration documentation.
- Keep PR #2743 as a candidate implementation after it is rebased onto current main.

## Non-goals

- Adding a new configuration surface.
- Making this the recommended production configuration.
- Implementing the one-command serve profile or timing report; those are tracked by #3215.
- Hiding required runtime or container prerequisites.

## Completion signal

The checked-in config passes canonical validation, needs no Router Model artifact or LLM backend, produces at least two different routing outcomes, and is protected by focused tests.

TwilightTechie:
Fixes #2744. Part of #2690.

A minimal maintained config so a new user can observe real routing decisions without downloading any classifier models or editing YAML.

## Before

The smallest maintained config that passed validation was the exhaustive reference `config/config.yaml` at 2,286 lines. A first run also required 13 classifier model artifacts, and there was no guard preventing a smaller example from regressing into needing models or a backend.

## Now

- `config/quickstart.yaml`: 163 lines (about 60 without comments), passes both the strict canonical schema and runtime validation.
- Routing is model-free: two keyword signals (BM25 and n-gram, computed in-process) drive three decisions that select two distinct models, so different prompts visibly produce different `x-vsr-selected-*` results.
- Zero model artifacts: classifier modules are disabled and embedding paths empty, so the existing `pkg/modeldownload` feature gates skip all downloads. No new machinery was added.
- Every decision carries a `fast_response` plugin, so an OpenAI-compatible request completes with a full response before any LLM backend is deployed.
- Package tests gate the invariants: strict schema, "requires no model artifacts" (asserted with the same `RouterConfig` predicates `pkg/modeldownload` gates downloads on), at least two keyword-conditioned decisions selecting distinct models, `fast_response` on every decision, and every referenced signal declared.
- `config/README.md` and `website/docs/installation/configuration.md` (the enforced parallel asset lists) both list the new asset.

## Intent

Reduce what a new user must read and download before their first successful request, per the discussion on #2690: config cost is active effort and hurts adoption more than the passive model download. This file is documented as an evaluation on-ramp, not a production baseline. The `serve` quickstart profile that consumes this file and the time-to-first-request baseline measurement are follow-ups tracked in #2690.

## Test Plan

- `go test ./pkg/config/`: full package passes, including the three new `TestQuickstart*` tests and the existing `TestDocs*` contract tests.
- Mutation check: an unknown top-level field fails the strict-schema test; enabling a classifier module without routing usage correctly stays green, matching downloader behavior (`routing_signal_usage.go:61`).
- `vllm-sr validate --config config/quickstart.yaml`: clean. 1 listener, 2 keyword signals, 3 decisions, 2 models, `fast_response` on all 3 decisions.
- `make agent-lint` on changed files: all hooks pass (go fmt, go lint, yaml fmt, changed-files lint; the only markdown findings are in gitignored local model downloads under `config/models/`, which CI does not check out).
- `gofmt`, `go vet` on the new test file: clean.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -3,6 +3,7 @@
 `config/` is now the user-facing config surface only.
 
 - `config/config.yaml`: exhaustive canonical reference config
+- `config/quickstart.yaml`: smallest maintained config that passes validation; model-free keyword routing for first-run evaluation (#2690), not a production baseline
 - `config/signal/`: reusable `routing.signals` fragments
 - `config/decision/`: reusable `routing.decisions` rule-shape fragments
 - `config/algorithm/`: reusable `decision.algorithm` snippets
--- a/config/quickstart.yaml
+++ b/config/quickstart.yaml
@@ -1,0 +1,163 @@
+# Quickstart config: the smallest maintained config that passes validation.
+#
+# Purpose: let a new user observe real routing decisions without downloading
+# any classifier models or editing YAML. Routing here uses keyword signals
+# only (BM25 / n-gram matching, computed in-process), and every decision
+# carries a fast_response plugin so a complete OpenAI-compatible response
+# comes back even when no LLM backend is running yet.
+#
+# This file is an evaluation on-ramp, not a production recommendation.
+# The exhaustive reference for every supported field is config/config.yaml.
+# See issue #2690 for background.
+
+version: v0.3
+
+listeners:
+  - name: http-8899
+    address: 0.0.0.0
+    port: 8899
+    timeout: 300s
+
+providers:
+  defaults:
+    # Requests that match no decision below fall through to this model.
+    default_model: quickstart-general
+  models:
+    # Two logical models let the demo show *different* selections for
+    # different prompts. Both point at the same local endpoint; with no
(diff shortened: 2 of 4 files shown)


Xunzhuo (commented):
Current main's quickstart now starts in Dashboard setup mode from an empty workspace, with configuration organized around fragments and recipes. This old standalone `config/quickstart.yaml` conflicts with that maintained flow. Please close it as superseded or rebase to a distinct current-doc gap.

PR #2746, opened 2026-08-02
[Docs] fix blog search only looking at the current page

### LINKED ISSUE #2742: bug: blog search only searches the current page, so most posts can't be found
### Describe the bug

The search box on the blog page only filters the 10 posts shown on that page, not all 21 posts on the site. So most posts can never be found from /blog - the search just says "No posts found" even though the post exists and opens fine by URL.

It also hides the pagination as soon as you type, so you can't move to the next page to keep looking.

### To Reproduce

1. Go to https://vllm-sr.ai/blog
2. Type "Athena" into the search box at the top
3. The page shows "No posts found. Try a different title, topic, or tag."
4. But that post does exist: https://vllm-sr.ai/blog/v0.2-vllm-sr-athena-release
   It's the 12th newest post, so it sits on page 2 and the search can't see it.
5. Also notice the "Older Entries" pagination at the bottom disappears while there is text in the search box.

### Expected behavior

Searching "Athena" should find the v0.2 Athena post.

The search should cover all blog posts, not just the ones rendered on the current page. 

### Affected layer

unsure

### Additional context

The blog has 21 posts and docusaurus.config.ts sets postsPerPage: 10, so there are 3 pages.

In website/src/theme/BlogListPage/index.tsx the filter runs over `items` (~line 153), but Docusaurus only puts the current page's posts into `items` - the full list is never available to that component. Line 239 also hides the paginator whenever a query is active.

Screenshots attached.

page 1 (main screen):

Image

page 2:

Image

Edit: dropping the pagination requirement. It was a workaround for the page-scoped search. Once search covers the whole archive there's nothing to page to, and the paginator's page numbers describe the unfiltered blog, so showing it next to results would be misleading.

abhinav-m22:
Closes #2742

## Purpose

- The search box on /blog only filtered the 10 posts rendered on that page. With 21 posts in the blog, most of the archive could never be found, and the page just said "No posts found" even though the post existed.
- Docusaurus only hands `BlogListPage` the current page's posts, so a client-side filter over props can't see the rest. This adds a small plugin that publishes every published post as global data, and the list page now searches that instead.
- Module: `Docs` (website only)

## Test Plan

- `npx tsc --noEmit` and `npx eslint` on the three changed files
- `npx docusaurus build --locale en` for a real build with SSR
- Manually: open /blog (page 1, not /blog/page/2) and search "athena". That post is 12th by date so it sits on page 2, which is exactly the case that used to fail.

## Test Result

- Build passes with no errors. `.docusaurus/globalData.json` now carries all 21 posts instead of 10.
- Searching "athena" from /blog finds the v0.2 Athena post.
- Nothing changes when you're not searching: still 10 cards, featured card keeps its description, paginator intact.
- If the index is ever missing, the page falls back to the old behaviour rather than breaking.

image

---
<details>

</details>


website/src/theme/BlogListPage/index.tsx
```diff
@@ -236,7 +272,7 @@ export default function BlogListPage(props: Props): React.ReactNode {
                       <p>Try a different title, topic, or tag.</p>
                     </div>
                   )}
-              {!normalizedQuery && <BlogListPaginator metadata={metadata} />}
+              {!isSearching && <BlogListPaginator metadata={metadata} />}
```
Xunzhuo, line 275: Please keep the paginator visible while a search is active, or explicitly update the linked issue to remove that requirement. Issue #2742 states that pagination should not disappear during search, but this condition still removes it whenever isSearching is true. The deploy preview confirms that the paginator disappears after entering “Athena”, so the PR currently addresses only part of the reported behavior.

website/src/data/committerActivity.generated.ts
```diff
@@ -15,66 +15,66 @@ export interface CommitterActivityEntry {
 }
 
 export const committerActivityWindow = {
-  "generatedAt": "2026-07-14",
-  "cutoffDate": "2026-04-14",
+  "generatedAt": "2026-08-01",
```
Xunzhuo: Please revert this file and website/src/data/contributorRank.generated.ts from this PR. They are unrelated to the blog-search fix and change public activity and ranking data as a side effect of running the website generation scripts. Keeping them obscures the functional diff and risks merging unrelated governance-data changes.

Xunzhuo (changes requested):
The archive-wide search works in the deploy preview, and the relevant website build and lint checks pass. I am requesting changes for the two inline issues below: one stated acceptance criterion is not implemented, and unrelated generated contributor data should be removed from this focused fix.

website/src/data/committerActivity.generated.ts
abhinav-m22: Done, both reverted. They were auto-generated, not intentional. PR is back to just the required files for the fix.

website/src/theme/BlogListPage/index.tsx
abhinav-m22: Updated #2742 as suggested. That requirement was a workaround for the page-scoped search, and now that search covers all posts there's nothing left to page to. BlogListPaginator also derives "Page N of M" from the unfiltered blog, so it'd read "Page 1 of 3" above a single result.

Xunzhuo (approved):
Thank you @abhinav-m22

PR #2749, opened 2026-08-02
[Router] fix: enforce replay capture policy so disabled switches drop bodies

### LINKED ISSUE #2748: bug: router_replay stores the full request body even when capture_request_body is false
### Describe the bug

The `router_replay` plugin has a `capture_request_body` switch. Setting it to `false` does not keep the request body out of the replay store. The full body is still persisted, and the replay API still serves it.

The write path ignores the switch end to end:

1. `buildReplayRoutingRecord` copies the body into the record unconditionally: [pkg/extproc/recorder.go#L245-L247](https://github.com/vllm-project/semantic-router/blob/bb72437bbe4e341b2720d96cd28d9d7d1ee62919/src/semantic-router/pkg/extproc/recorder.go#L245-L247)
2. `Recorder.AddRecord` checks `captureRequestBody` only to decide whether to truncate. When the switch is off, it skips truncation and passes the body through untouched: [pkg/routerreplay/recorder.go#L116-L124](https://github.com/vllm-project/semantic-router/blob/bb72437bbe4e341b2720d96cd28d9d7d1ee62919/src/semantic-router/pkg/routerreplay/recorder.go#L116-L124)
3. The store layer persists whatever it receives: [pkg/routerreplay/store/memory.go#L42-L67](https://github.com/vllm-project/semantic-router/blob/bb72437bbe4e341b2720d96cd28d9d7d1ee62919/src/semantic-router/pkg/routerreplay/store/memory.go#L42-L67). With the Postgres, Redis, Milvus, or Qdrant backends the body lands in external durable storage.
4. The one method that does honor the switch, `Recorder.AttachRequest` ([pkg/routerreplay/recorder.go#L212-L220](https://github.com/vllm-project/semantic-router/blob/bb72437bbe4e341b2720d96cd28d9d7d1ee62919/src/semantic-router/pkg/routerreplay/recorder.go#L212-L220)), has no callers in extproc.

So with `capture_request_body: false` the operator gets the opposite of what the flag promises. The body is stored untruncated, because the `max_body_bytes` cap sits inside the same skipped branch. It is then returned by `GET /v1/router_replay/{id}` and by the list endpoint with `showDetails=true`.

The response side is half covered. Response bodies flow through `Recorder.AttachResponse`, which returns early when `capture_response_body` is false. But `AddRecord` has the same unguarded pass-through for `ResponseBody`, so any caller that sets it at add time would leak the same way.

### To Reproduce

1. Enable replay on a decision with the capture switch off:

   ```yaml
   - type: router_replay
     configuration:
       enabled: true
       capture_request_body: false
   ```

2. Send a chat completion request containing sensitive content.
3. Fetch `GET /v1/router_replay?showDetails=true`, or fetch the record by id.
4. `request_body` contains the full request, prompt included.

### Expected behavior

With `capture_request_body: false`, the request body never reaches the store. Same for `capture_response_body` and the response body. Truncation only bounds a leak; the switch must prevent it.

### Affected layer

plugin

drivebyer:
## What

`capture_request_body: false` did not keep the request body out of the replay store. `Recorder.AddRecord` checked the capture switches only to decide whether to truncate, so with capture off the body skipped truncation and was persisted untouched, then served back by the replay detail API. The one method honoring the switch, `AttachRequest`, has no callers in extproc.

This PR enforces the policy in `AddRecord`, the single write path into the store: when a capture switch is off, the corresponding body is dropped and its truncated flag reset. The same guard covers `ResponseBody` as defense in depth (the runtime response path already goes through the guarded `AttachResponse`). A recorder without an explicit capture policy now drops bodies by default, matching `Attach*`.

Fixes #2748

## Test Plan

- New regression tests in `pkg/routerreplay/recorder_test.go`: capture disabled drops bodies and resets truncated flags; a recorder without `SetCapturePolicy` drops by default; capture enabled keeps bodies truncated to `max_body_bytes`.
- `go test ./pkg/routerreplay/...` passes; targeted extproc replay tests (`-run 'RouterReplay|ReplayTool|Trajectory'`) pass after declaring capture policy explicitly in the two fixtures that relied on the pass-through.
- `make agent-lint` on the changed files passes.

### DIFF
--- a/src/semantic-router/pkg/extproc/router_replay_api_test.go
+++ b/src/semantic-router/pkg/extproc/router_replay_api_test.go
@@ -345,6 +345,9 @@ func newReplayAPITestRouterWithRecords(t *testing.T, count int, requestBody stri
 	t.Helper()
 
 	recorder := routerreplay.NewRecorder(store.NewMemoryStore(count+1, 0))
+	// Capture must be enabled for AddRecord to store bodies; the limit must
+	// exceed the 5MiB oversized-body fixtures so they are stored untruncated.
+	recorder.SetCapturePolicy(true, true, 10*1024*1024)
 	for i := 1; i <= count; i++ {
 		recordID := fmt.Sprintf("replay-%03d", i)
 		timestamp := time.Unix(int64(i), 0).UTC()
--- a/src/semantic-router/pkg/extproc/router_replay_trajectory_test.go
+++ b/src/semantic-router/pkg/extproc/router_replay_trajectory_test.go
@@ -157,6 +157,8 @@ func TestHandleRouterReplayTrajectoryCoalescesConsecutiveToolCalls(t *testing.T)
 
 func TestHandleRouterReplayTrajectoryFallsBackToBodyParsing(t *testing.T) {
 	recorder := routerreplay.NewRecorder(store.NewMemoryStore(10, 0))
+	// Body-based trajectory fallback presumes the bodies were captured.
+	recorder.SetCapturePolicy(true, true, 4096)
 	sessionID := "sess-fallback"
 
 	requestBody := `{"messages":[{"role":"user","content":"hello"},{"role":"assistant","tool_calls":[{"id":"call-x","type":"function","function":{"name":"my_tool","arguments":"{}"}}]},{"role":"tool","content":"tool result","tool_call_id":"call-x"}]}`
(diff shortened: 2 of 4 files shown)


AayushSaini101 (approved):
LGTM thanks :)

PR #2753, opened 2026-08-03
[Router] harden learning outcome ingest with ownership and idempotency [ Phase  3 ]

AayushSaini101:
Related #2463

## Purpose

- Harden `POST /v1/router/outcomes` so learning ingest cannot forge provenance or poison experience updates.
- Derive learning `source` from authenticated management principals (not the request body).
- Require `Idempotency-Key`, apply per-principal rate limits, and bind experience updates to owned replay events (replay must exist; model binding must match).
- Update experience only after a successful replay outcome append; duplicate keys return idempotent success without double-counting.

Modules: `Router`

## Test Plan

- [x] `CGO_ENABLED=1 go test -race ./pkg/apiserver ./pkg/extproc ./pkg/routerruntime -run 'RouterOutcome|LearningOutcome|NormalizeRouter|RouterLearningRuntime'`
- [x] Handler tests: missing `Idempotency-Key`, principal-derived `source=operator`, rate-limit enforcement
- [x] Runtime tests: missing replay → `replay_not_found`; model mismatch → `ownership_mismatch`; idempotency dedup; experience only after owned append
- [ ] Optional reviewer check: with bearer auth enabled, call outcomes without `learning.ingest` → 403; with key + owned replay → 200

Validation covers the Phase 3 API policy + runtime ownership seams called out in #2463 without opening Phase 4–6 exposure/audit/matrix scope.

## Test Result

- Targeted race tests for `apiserver`, `extproc`, and `routerruntime` passed locally after rebuilding Rust bindings.
- Pre-existing unrelated `staticcheck` noise in classification OpenVINO paths can still fail full `go-lint` on some local trees; not introduced by this change.
- Follow-up still tracked on #2463: Phase 4 (CLI/Helm exposure isolation), Phase 5 (request policy + audit emit), Phase 6 (full auth matrix + E2E).

---
<details>

</details>

### DIFF
--- a/dashboard/frontend/e2e/feedback-buttons.spec.ts
+++ b/dashboard/frontend/e2e/feedback-buttons.spec.ts
@@ -63,9 +63,11 @@ test.describe('FeedbackButtons', () => {
 
   test('thumbs up sends correct feedback payload', async ({ page }) => {
     let feedbackPayload: Record<string, unknown> | null = null
+    let idempotencyKey: string | null = null
 
     await page.route('**/api/router/v1/router/outcomes', async (route) => {
       feedbackPayload = route.request().postDataJSON()
+      idempotencyKey = route.request().headers()['idempotency-key'] ?? null
       await route.fulfill({
         status: 200,
         headers: { 'Content-Type': 'application/json' },
@@ -84,6 +86,7 @@ test.describe('FeedbackButtons', () => {
     await thumbsUp.click()
 
     expect(feedbackPayload).not.toBeNull()
+    expect(idempotencyKey).toBeTruthy()
     expect(feedbackPayload!.replay_id).toBe(MOCK_REPLAY_ID)
     expect(feedbackPayload!.source).toBe('user')
     expect(feedbackPayload!.target).toBe('model')
@@ -129,9 +132,11 @@ test.describe('FeedbackButtons', () => {
 
   test('thumbs down sends underpowered outcome', async ({ page }) => {
     let feedbackPayload: Record<string, unknown> | null = null
+    let idempotencyKey: string | null = null
 
     await page.route('**/api/router/v1/router/outcomes', async (route) => {
       feedbackPayload = route.request().postDataJSON()
+      idempotencyKey = route.request().headers()['idempotency-key'] ?? null
       await route.fulfill({
         status: 200,
         headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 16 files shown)


AayushSaini101:
Screenshot 2026-08-03 at 3 45 38 PM
This is the architecture till phase 3 : )

drivebyer (commented):
Thanks for tackling this — the trust gap in `/v1/router/outcomes` is real. One concern: a client-supplied Idempotency-Key can't stop repeat voting (callers pick the key), breaks existing clients like the dashboard's feedback buttons, and the check/register isn't atomic under concurrency. Since the replay record already proves the event happened and remembers prior outcomes, could we instead dedupe server-side on `(replay_id, target, target_ref)` via record.Outcomes, and derive decision/tier from the record too? No new header, works across replicas.

AayushSaini101:
> Thanks for tackling this — the trust gap in `/v1/router/outcomes` is real. One concern: a client-supplied Idempotency-Key can't stop repeat voting (callers pick the key), breaks existing clients like the dashboard's feedback buttons, and the check/register isn't atomic under concurrency. Since the replay record already proves the event happened and remembers prior outcomes, could we instead dedupe server-side on `(replay_id, target, target_ref)` via record.Outcomes, and derive decision/tier from the record too? No new header, works across replicas.

@drivebyer thanks for reviewing the MR, i didn't get why the /post/feedback endpoint doesn't work in this case can you please explain a bit

drivebyer:
> > Thanks for tackling this — the trust gap in `/v1/router/outcomes` is real. One concern: a client-supplied Idempotency-Key can't stop repeat voting (callers pick the key), breaks existing clients like the dashboard's feedback buttons, and the check/register isn't atomic under concurrency. Since the replay record already proves the event happened and remembers prior outcomes, could we instead dedupe server-side on `(replay_id, target, target_ref)` via record.Outcomes, and derive decision/tier from the record too? No new header, works across replicas.
> 
> @drivebyer thanks for reviewing the MR, i didn't get why the /post/feedback endpoint doesn't work in this case can you please explain a bit

The dashboard's feedback buttons (FeedbackButtons.tsx) send POST /v1/router/outcomes with no Idempotency-Key, and this PR returns 400 MISSING_IDEMPOTENCY_KEY without it — so after merge, vote fails until the frontend is updated.

AayushSaini101:
> > > Thanks for tackling this — the trust gap in `/v1/router/outcomes` is real. One concern: a client-supplied Idempotency-Key can't stop repeat voting (callers pick the key), breaks existing clients like the dashboard's feedback buttons, and the check/register isn't atomic under concurrency. Since the replay record already proves the event happened and remembers prior outcomes, could we instead dedupe server-side on `(replay_id, target, target_ref)` via record.Outcomes, and derive decision/tier from the record too? No new header, works across replicas.
> > 
> > 
> > @drivebyer thanks for reviewing the MR, i didn't get why the /post/feedback endpoint doesn't work in this case can you please explain a bit
> 
> The dashboard's feedback buttons (FeedbackButtons.tsx) send POST /v1/router/outcomes with no Idempotency-Key, and this PR returns 400 MISSING_IDEMPOTENCY_KEY without it — so after merge, vote fails until the frontend is updated.

thanks @drivebyer added logic for generating key while calling /feedback endpoint from the frontend side thanks

Xunzhuo (approved):

PR #2759, opened 2026-08-04
[Router] add pluggable jailbreak classifier backend (backend enum + http_classify + positive_labels)

theohsiung:
Related #2587

## Purpose

- Replaces `prompt_guard`'s three independent boolean flags (`use_modernbert`, `use_mmbert_32k`, `use_vllm`) with a single `backend` enum (`candle | mmbert32k | http_chat | http_classify`), and adds a new `http_classify` backend so a self-hosted sequence-classifier model (e.g. a Prompt-Guard-2-style model) can be plugged in without disguising it as a chat completion.
- Fixes the root cause behind #2587's original report: `risk_score` could report the argmax-class confidence instead of P(jailbreak), producing "high risk_score + allow" for a confident-but-wrong benign prediction. Every jailbreak backend (Candle LoRA/BERT, ModernBERT, mmBERT-32K, and the generative `http_chat` path) now always returns the full class-probability distribution, never an argmax-only result.
- Adds `positive_labels` config so a custom model whose positive class isn't literally named `jailbreak` (e.g. `INJECTION`, `malicious`) can be configured instead of silently mapping to safe. If configured, router startup now fails fast when none of the configured labels exist in the loaded `jailbreak_mapping` - a misconfigured label can no longer mean "never detected" silently.
- Which module(s) does this affect? `Router` / `Bindings` / `Docs`

This is a **breaking config change**: the three removed boolean flags have no auto-translation. `config/config.yaml`, operator samples, e2e profiles/configs, k8s conversion testdata, and the current (non-versioned) website docs are migrated to `backend:` in this PR; the actual field semantics and migration guidance are documented in `website/docs/tutorials/global/safety-models-and-policy.md` and `config/README.md`.

Category, complexity, and PII classifiers are explicitly out of scope for this PR. `ScoringBackend` and `TokenClassifierBackend` interface names are reserved (not implemented or wired) for the complexity/difficulty-scoring and PII/GLiGuard-style follow-ups discussed on #2587.

## Test Plan

- `cargo fmt --check` / `cargo test --no-default-features --lib` (candle-binding)
- `go build` / `go vet` / `go test -race` / `golangci-lint run --new-from-rev=origin/main` for `pkg/config`, `pkg/classification`, `pkg/apiserver`, `pkg/extproc`, `pkg/k8s`
- `markdownlint` / `yamllint` on the changed docs and config fixtures
- Manual end-to-end validation: served `protectai/deberta-v3-base-prompt-injection-v2` (open, non-gated - the actual target model, `meta-llama/Llama-Prompt-Guard-2-86M`, requires a gated HF license I didn't have access to for this validation) behind a small FastAPI wrapper speaking the `http_classify` contract, then exercised it through the real `HTTPClassifierJailbreakInference` code via both `CheckForJailbreakWithRisk` and the `EvaluateAllSignals` signal-evaluation path, with `positive_labels: [INJECTION]` configured (this model's positive label isn't `jailbreak`).

## Test Result

- All automated checks above pass. The 7 pre-existing `golangci-lint` staticcheck issues in unrelated OpenVINO-stub files, and the pre-existing Rust/Go test failures that need network/GPU access (embedding-model downloads, Milvus/Valkey connectivity), are confirmed unrelated to this change (`--new-from-rev` shows zero new lint issues; none of the failing tests touch any file this PR modifies).
- Manual model validation: the substitute model correctly handled the exact prompt from #2587's report that fooled the bundled `mmbert32k-jailbreak-detector-merged` model (reported safe there; correctly blocked here with `risk_score=1.0`), plus overt, roleplay-framed, and multilingual (Chinese) jailbreaks, with no false positives on benign prompts (including ones containing security-adjacent keywords). It missed one indirect "grandmother-style" social-engineering jailbreak - a known weak spot for this class of classifier, unrelated to this PR's mechanism.
- Follow-up risks/gaps: no real e2e (kind cluster) run yet, only unit/manual validation; the actual gated `Llama-Prompt-Guard-2-86M` model itself was not tested end-to-end.


AayushSaini101:
Thanks for the feature @theohsiung we indeed this: QQ can you please share the testing using the external model via http, and how would we deal in the case of the failure, would we allow retry ? or any kind of fallback, do we plan to extend this for all the classifiers like PII detection, feedback features as well ?

theohsiung:
Thanks for the questions @AayushSaini101!

**1. Testing with the external model via http_classify**

Since the actual target model (`meta-llama/Llama-Prompt-Guard-2-86M`) is gated and I didn't have access, I validated the mechanism with an open substitute: `protectai/deberta-v3-base-prompt-injection-v2`, served behind a small FastAPI wrapper implementing the `http_classify` wire contract (`POST /classify` -> `[{"label", "score"}, ...]`). I exercised it through the real `HTTPClassifierJailbreakInference` code path via both `CheckForJailbreakWithRisk` and `EvaluateAllSignals`, with `positive_labels: [INJECTION]` configured (this model's positive class isn't literally named `jailbreak`).

Results: it correctly caught the exact prompt from #2587's original report that fooled the bundled `mmbert32k-jailbreak-detector-merged` model (reported safe there; blocked here with `risk_score=1.0`), plus overt, roleplay-framed, and multilingual (Chinese) jailbreaks, with no false positives on benign prompts (including ones containing security-adjacent keywords). It missed one indirect "grandmother-style" social-engineering jailbreak - a known weak spot for this class of classifier, unrelated to this PR's mechanism.

**2. Failure handling - retry / fallback**

Honest answer: neither exists yet. `HTTPClassifierJailbreakInference.Classify` is a single HTTP call with a configurable timeout (default 30s, `TimeoutSeconds` in config). On any failure - network error, timeout, non-2xx, unparseable response, or no response label matching the configured `jailbreak_mapping` - it returns an error, no retry attempted.

How that error is handled depends on the caller:
- In the signal-evaluation path used for routing decisions (`evaluateJailbreakSignal`), the error is logged and that content chunk simply contributes nothing to the jailbreak signal - effectively fail-open for that piece, it won't block the request or retry.
- In `CheckForJailbreakWithRisk` (e.g. the security detection API), the error surfaces directly to the caller.

This was a deliberate minimal-first-version choice to keep this PR's scope to the backend abstraction itself. Retry/circuit-breaker/fallback-backend behavior for `http_classify` (and really, for external-model backends in general) feels like a reasonable follow-up - happy to file an issue for it if you agree it shouldn't block this PR.

**3. Extending to other classifiers (PII, etc.)**

Yes - that's exactly what #2760 tracks, deliberately scoped out of this PR to keep the diff narrow: category -> `SequenceClassifierBackend` (same interface this PR already introduces), complexity/difficulty -> `ScoringBackend`, PII -> `TokenClassifierBackend`. Both interface names are already reserved in this PR (unimplemented) so the follow-up plugs into the same shape.

theohsiung:
Re-triggering CI - the `test-and-build` failure was an unrelated network flake (Helm chart repo `jaegertracing.github.io` connection reset, nothing to do with this PR's changes). Closing/reopening to get a fresh run since I don't have rights to re-run failed jobs directly on this repo.

AayushSaini101:
> Thanks for the questions @AayushSaini101!
> 
> **1. Testing with the external model via http_classify**
> 
> Since the actual target model (`meta-llama/Llama-Prompt-Guard-2-86M`) is gated and I didn't have access, I validated the mechanism with an open substitute: `protectai/deberta-v3-base-prompt-injection-v2`, served behind a small FastAPI wrapper implementing the `http_classify` wire contract (`POST /classify` -> `[{"label", "score"}, ...]`). I exercised it through the real `HTTPClassifierJailbreakInference` code path via both `CheckForJailbreakWithRisk` and `EvaluateAllSignals`, with `positive_labels: [INJECTION]` configured (this model's positive class isn't literally named `jailbreak`).
> 
> Results: it correctly caught the exact prompt from #2587's original report that fooled the bundled `mmbert32k-jailbreak-detector-merged` model (reported safe there; blocked here with `risk_score=1.0`), plus overt, roleplay-framed, and multilingual (Chinese) jailbreaks, with no false positives on benign prompts (including ones containing security-adjacent keywords). It missed one indirect "grandmother-style" social-engineering jailbreak - a known weak spot for this class of classifier, unrelated to this PR's mechanism.
> 
> **2. Failure handling - retry / fallback**
> 
> Honest answer: neither exists yet. `HTTPClassifierJailbreakInference.Classify` is a single HTTP call with a configurable timeout (default 30s, `TimeoutSeconds` in config). On any failure - network error, timeout, non-2xx, unparseable response, or no response label matching the configured `jailbreak_mapping` - it returns an error, no retry attempted.
> 
> How that error is handled depends on the caller:
> 
> * In the signal-evaluation path used for routing decisions (`evaluateJailbreakSignal`), the error is logged and that content chunk simply contributes nothing to the jailbreak signal - effectively fail-open for that piece, it won't block the request or retry.
> * In `CheckForJailbreakWithRisk` (e.g. the security detection API), the error surfaces directly to the caller.
> 
> This was a deliberate minimal-first-version choice to keep this PR's scope to the backend abstraction itself. Retry/circuit-breaker/fallback-backend behavior for `http_classify` (and really, for external-model backends in general) feels like a reasonable follow-up - happy to file an issue for it if you agree it shouldn't block this PR.
> 
> **3. Extending to other classifiers (PII, etc.)**
> 
> Yes - that's exactly what #2760 tracks, deliberately scoped out of this PR to keep the diff narrow: category -> `SequenceClassifierBackend` (same interface this PR already introduces), complexity/difficulty -> `ScoringBackend`, PII -> `TokenClassifierBackend`. Both interface names are already reserved in this PR (unimplemented) so the follow-up plugs into the same shape.

Thanks for the explanation, let's focus on small changes, then we can move other requirements : )

AayushSaini101:
@adaamko can we use any detector for writing the blog for this feature, just curious. We have https://github.com/KRLabsOrg/LettuceDetect but it seems this is related to the provided Span-level grounding Span-level grounding verification for RAG, code,  cc: @Xunzhuo

adaamko:
> @adaamko can we use any detector for writing the blog for this feature, just curious. We have https://github.com/KRLabsOrg/LettuceDetect but it seems this is related to the provided Span-level grounding Span-level grounding verification for RAG, code, cc: @Xunzhuo

Hey, yeah you are right that currently Lettucedetect doesnt support guardrails, and that's something we are planning to work on together with @Xunzhuo , maybe based on LFM encoders. Meanwhile maybe Shieldstral would be also a good candidate to test: https://mistral.ai/news/shieldstral/ . Or maybe we could try Opir: https://huggingface.co/knowledgator/opir-multitask-multilang-v1.0 or GliGuard https://huggingface.co/fastino/gliguard-LLMGuardrails-300M . But nothing really supports span based classification, so that's something we should change :)

src/semantic-router/pkg/apiserver/route_model_info_classifiers.go
```diff
@@ -99,7 +103,7 @@ func buildRoutingClassifierModels(
 			Metadata: map[string]string{
 				"enabled":                "true",
 				"jailbreak_mapping_path": promptGuard.JailbreakMappingPath,
-				"model_type":             resolveInlineModelType(promptGuard.UseMmBERT32K, promptGuard.UseModernBERT, false),
+				"model_type":             backend,
```
AayushSaini101: @theohsiung can't we use another name here seems "backend" name is vauge naming here, feel free to share your suggestions thanks : )

src/semantic-router/pkg/config/config.go
```diff
@@ -20,6 +20,24 @@ const (
...
+	// (32K context, YaRN RoPE, multilingual).
+	PromptGuardBackendMmBERT32K = "mmbert32k"
+	// PromptGuardBackendHTTPChat calls an external model with role="guardrail"
+	// through a generative chat-completion prompt (e.g. Qwen3Guard-style).
+	PromptGuardBackendHTTPChat = "http_chat"
+	// PromptGuardBackendHTTPClassify calls an external model with
+	// role="guardrail" through a lightweight sequence-classifier HTTP
+	// contract (text in, full label/score distribution out).
+	PromptGuardBackendHTTPClassify = "http_classify"
+)
```
AayushSaini101, line 53: @theohsiung, we can have two ways only to integrate the model, via the candle ML process, by which it can take connect with any local model, that is configured either mmber32k or any one and otherway using  connection like via http connection only, why do we need http_classify and need to mention mbert32 as hardcore in seperate enum thanks. : )

theohsiung: Good question! These aren't actually interchangeable under one umbrella - let me break down why each needs its own value:

**`candle` vs `mmbert32k`**: these route to genuinely different construction paths (`createJailbreakInferenceCandle()` vs `createMmBERT32KJailbreakInference()`), backed by different FFI calls. mmBERT-32K isn't "any local model" plugged into the same candle loader - it's a structurally different architecture (32K context window, YaRN RoPE scaling, multilingual tokenizer) that needs its own code path, not just a different config value passed into one generic loader.

**`http_chat` vs `http_classify`**: these are different wire protocols, not two names for "just HTTP":
- `http_chat` talks to a generative chat-completion endpoint (e.g. Qwen3Guard-style) - it prompts the model and parses free-form text for a verdict.
- `http_classify` talks to a dedicated sequence-classification endpoint - `POST {"inputs": text}` -> `[{"label", "score"}, ...]`, the standard HuggingFace text-classification pipeline shape.

Collapsing them into one `http` value would mean the router can't tell which contract the endpoint speaks - we'd either need a second field to disambiguate (no simpler than today) or runtime protocol sniffing (more fragile, not less).

Totally hear the "fewer knobs is better" instinct though - if there's a naming scheme that makes the distinction clearer without adding a value, happy to explore it. But I don't think we can drop to two values without losing real information about which backend a given endpoint actually implements.

AayushSaini101: thanks for the explanations @theohsiung Got it thanks for explanation, so the user need to provide the which type of method they are going to use in the configuration for the jailbreak process: like http, http_classify, candel [ for local model], default model mmbert-32 k right ; )

theohsiung: Yep, exactly - one small correction: the default (when `backend` is unset) is `candle`, not `mmbert32k`. So it's `candle` (default, local) / `mmbert32k` (local) / `http_chat` / `http_classify`.

src/semantic-router/pkg/apiserver/route_model_info_classifiers.go
theohsiung: Good catch - you're right that reusing `model_type` here was misleading. This entry doesn't actually have architecture info to report anymore (the old `UseMmBERT32K`/`UseModernBERT` flags that used to drive `model_type` for jailbreak are gone; `candle` now auto-detects bert vs. modernbert at load time, so it's not knowable at config-read time). Renamed the key to `backend` in d7ce394c so it accurately reflects what's being reported, and `model_type` keeps meaning "model architecture" consistently for every other entry in this map.

AayushSaini101:
Overall looks good to me @shraderdm @drivebyer @FAUST-BENCHOU @Xunzhuo can you guys PTAL on this cc: @adaamko you can take a look as well thanks

src/semantic-router/pkg/config/config.go
Xunzhuo, line 53: These values are not peers on one abstraction axis: `candle` is a runtime, `mmbert32k` is a local architecture/loader variant, while `http_chat` and `http_classify` combine transport with two different output semantics. This will make every new architecture or protocol grow this enum and all downstream switches. Please separate at least the semantic kind (`sequence_classifier` vs generative guard verdict), runtime (`candle`/HTTP), and variant (`auto`/`mmbert32k`), or keep this enum only as a compatibility adapter to such an internal descriptor. In particular, `mmbert32k` should be a Candle variant and `http_chat` should not be modeled as a sequence classifier.

Xunzhuo (commented):
Thanks for replacing the mutually exclusive booleans with an explicit backend selection. I do not think the current API boundary is extensible enough yet: the public enum mixes runtime, architecture, transport, and output semantics; the core interface leaks a Candle-specific DTO; and the generative chat path cannot satisfy the same probability contract as a sequence classifier. There are also concrete correctness blockers in the new probability paths, detailed inline.

The config change also needs to be propagated through the full public surface before merge: the Operator CRD still has `use_modernbert` and cannot express `backend`, `positive_labels`, or an external guardrail model; dashboard/builders and several shipped configs still emit removed fields. These gaps make the new backend API unreachable or misleading outside direct YAML configuration.

src/semantic-router/pkg/classification/classifier_jailbreak_init.go
```diff
@@ -101,92 +101,101 @@ func createMmBERT32KJailbreakInitializer() JailbreakInitializer {
...
+	Classify(text string) (candle_binding.ClassResultWithProbs, error)
 }
 
 type JailbreakInferenceImpl struct{}
 
-func (c *JailbreakInferenceImpl) Classify(text string) (candle_binding.ClassResult, error) {
+func (c *JailbreakInferenceImpl) Classify(text string) (candle_binding.ClassResultWithProbs, error) {
 	// Try jailbreak-specific classifier first, fall back to ModernBERT if it fails
-	result, err := candle_binding.ClassifyJailbreakText(text)
+	result, err := candle_binding.ClassifyJailbreakTextWithProbs(text)
```
Xunzhuo, line 158: This backend-neutral seam currently exposes `candle_binding.ClassResultWithProbs`, forcing every HTTP/future backend to depend on a Candle FFI DTO. It also omits `context.Context`, so remote inference cannot inherit request cancellation, deadlines, or tracing. Please define a classification-owned result contract (preferably label-to-score plus explicit score semantics) and use `Classify(ctx context.Context, text string)`. Argmax/confidence should be derived once in the policy layer rather than repeated by each backend.

src/semantic-router/pkg/classification/vllm_classifier.go
```diff
@@ -94,22 +99,19 @@ func (v *VLLMJailbreakInference) Classify(text string) (candle_binding.ClassResu
...
 		isJailbreak, confidence, categories)
 
-	// Map to ClassResult format
+	// Map to ClassResultWithProbs format
 	// Class: 0 = safe, 1 = jailbreak/unsafe
 	class := 0
+	probabilities := []float32{confidence, 1 - confidence}
 	if isJailbreak {
 		class = 1
+		probabilities = []float32{1 - confidence, confidence}
```
Xunzhuo: A generative chat verdict does not provide a calibrated sequence-classifier distribution. Synthesizing one here also hard-codes `0=safe, 1=jailbreak`, while downstream indexes it using the configured mapping. An inverted or multiclass mapping therefore produces the wrong risk (and can map a safe verdict to the positive label). Please give this path a separate generative-guard/verdict interface, or return named evidence with an explicit `uncalibrated` score kind and perform mapping centrally.

src/semantic-router/pkg/classification/http_classifier.go
```diff
@@ -0,0 +1,173 @@
...
+		if !ok || idx < 0 || idx >= numClasses {
+			continue
+		}
+		probabilities[idx] = s.Score
+		if bestIdx == -1 || s.Score > bestScore {
+			bestIdx = idx
+			bestScore = s.Score
+		}
+	}
+	if bestIdx == -1 {
```
Xunzhuo: Requiring only one matching label makes an incomplete response look like a complete distribution: omitted mapping entries remain `0.0`. A common top-k response containing only `benign` will therefore report zero positive risk even though the positive score was simply omitted. Please reject partial/unknown/duplicate labels and non-finite or out-of-range scores, verify the expected label set, and validate normalization before returning success.

src/semantic-router/pkg/classification/classifier_jailbreak_risk.go
```diff
@@ -67,8 +123,8 @@ func (c *Classifier) CheckForJailbreakWithRisk(text string) (bool, string, float
 		return false, "", 0.0, 0.0, fmt.Errorf("unknown jailbreak class index: %d", result.Class)
 	}
 
-	isJailbreak := result.Confidence >= threshold && jailbreakType == jailbreakPositiveLabel
-	riskScore := jailbreakRiskScore(c.JailbreakMapping, result)
+	isJailbreak := result.Confidence >= threshold && isPositiveJailbreakLabel(c.Config.PromptGuard.PositiveLabels, jailbreakType)
```
Xunzhuo: The decision still thresholds the argmax confidence, while `riskScore` below sums all configured positive-label probabilities. With `positive_labels: [jailbreak, INJECTION]`, scores `0.40 benign / 0.35 jailbreak / 0.25 INJECTION`, and threshold `0.5`, this returns risk `0.60` but allows the request. Please have the shared policy layer threshold the summed positive mass (and use the same rule in signal evaluation), otherwise the API can still produce high-risk + allow outcomes.

candle-binding/src/ffi/classify.rs
```diff
@@ -285,6 +285,90 @@ pub extern "C" fn classify_jailbreak_text(text: *const c_char) -> Classification
...
+                    "LoRA jailbreak classifier error: {}, falling back to Traditional BERT",
+                    e
+                );
+                // Don't return - fall through to Traditional BERT classifier
+            }
+        }
+    }
+
+    // Fallback to Traditional BERT classifier
+    if let Some(classifier) = BERT_JAILBREAK_CLASSIFIER.get() {
```
Xunzhuo, line 343: This reads the module-local `classify.rs::BERT_JAILBREAK_CLASSIFIER`, but `init_jailbreak_classifier` writes `ffi/init.rs::BERT_JAILBREAK_CLASSIFIER`. They are distinct `OnceLock`s, so this fallback is always empty for a successfully initialized Traditional BERT model. Please import and use the initializer's static (as is already done for `LORA_JAILBREAK_CLASSIFIER`) and add an FFI regression test for the Traditional BERT with-probabilities path.

src/semantic-router/pkg/config/canonical_defaults.go
```diff
@@ -234,7 +234,7 @@ func defaultPromptGuardModule() CanonicalPromptGuardModule {
 			Enabled:              true,
 			Threshold:            0.7,
 			UseCPU:               true,
-			UseMmBERT32K:         true,
+			Backend:              PromptGuardBackendMmBERT32K,
```
Xunzhuo: This conflicts with the new public contract saying an omitted backend defaults to `candle`. Canonical global resolution starts from these defaults and overlays user YAML, so omitting `backend` retains `mmbert32k`; a user migrating `use_mmbert_32k: false` by removing the old flag can silently select the wrong loader. Either make the canonical default `candle`, or document the canonical default as `mmbert32k` and require `backend: candle` explicitly in the migration guide.

src/semantic-router/pkg/classification/classifier_jailbreak_risk.go
theohsiung: Confirmed and fixed in 0e43c4da: both `CheckForJailbreakWithRisk` and `findBestJailbreakMatch` (signal-evaluation path) now threshold `jailbreakRiskScore`'s summed positive-label mass directly, not argmax confidence. Added a regression test reproducing your exact scenario (`benign=0.40/jailbreak=0.35/INJECTION=0.25`, threshold `0.5`) confirming it now returns `isJailbreak=true`.

src/semantic-router/pkg/classification/vllm_classifier.go
theohsiung: Confirmed and fixed in 7aaaf709: `NewVLLMJailbreakInference` now takes the `jailbreak_mapping`, resolves the positive label's index once at construction (also rejecting mappings with other than 2 classes, since a binary generative verdict can't populate a meaningful distribution over more), and `Classify` writes the distribution at those resolved indices instead of assuming a fixed `{0: safe, 1: jailbreak}` layout. Added a regression test with an inverted mapping (`{jailbreak: 0, safe: 1}`) confirming the verdict is no longer flipped.

src/semantic-router/pkg/classification/http_classifier.go
theohsiung: Confirmed and fixed in f8b1d7f0: `alignScoresToMapping` now requires every configured label to be present exactly once, rejects duplicate labels and non-finite/out-of-[0,1] scores, and rejects a distribution whose scores don't sum to ~1.0 (small floating-point tolerance). A response that omits a configured label - or otherwise isn't a valid complete distribution - now surfaces as an error instead of silently defaulting the missing entries to 0.0.

candle-binding/src/ffi/classify.rs
theohsiung: Good catch, confirmed - and it's pre-existing (predates this PR, verified via git blame). Filed as #2834 rather than fixing inline here: the correct fix (export the static from init.rs, import it in classify.rs, matching LORA_JAILBREAK_CLASSIFIER's existing pattern) is a one-line change, but touching init.rs at all sweeps in ~76 pre-existing clippy::not_unsafe_ptr_arg_deref findings under this repo's file-scoped (not line-scoped) Rust lint gate (see #2761/#2762) - basically every extern "C" fn in that file. Properly clearing that means marking ~39 FFI entry points unsafe fn and auditing Rust-side callers that invoke them without an unsafe block, which felt like its own PR rather than something to fold into this one. Also noted in the issue: BERT_CLASSIFIER and BERT_PII_CLASSIFIER have the identical duplication bug for the same reason, not just the jailbreak one.

src/semantic-router/pkg/config/canonical_defaults.go
theohsiung: Confirmed and addressed in 5cd41ab7 - though as documentation, not a behavior change. Reconsidered both options you offered: changing the canonical default to `candle` would actually break the out-of-the-box canonical config, since the bundled default model_id (`mmbert32k-jailbreak-detector-merged`) is an mmBERT-32K checkpoint that needs the mmbert32k backend's 32K-context/YaRN-RoPE handling - loading it via the plain candle path would be architecturally wrong. So I went with your other option: kept the canonical baseline as mmbert32k, and fixed the code comments (config.go, classifier_jailbreak_init.go) plus the website docs to state this explicitly, with added migration guidance telling users who want plain candle under canonical resolution to set backend: candle explicitly rather than relying on omission.

src/semantic-router/pkg/config/config.go
theohsiung: One correction to what I said above: "default is candle" was incomplete. @Xunzhuo caught that this is only true for the raw fallback in `createJailbreakInference`'s switch statement - canonical config resolution (what the dashboard and canonical export go through) actually defaults `backend` to `mmbert32k` explicitly, matching the bundled model. So: `mmbert32k` (canonical default) / `candle` (raw fallback default, and what you'd get if a config is built without canonical resolution) / `http_chat` / `http_classify`. Clarified in 5cd41ab7.

adaamko:
Took a pass through. The alignScoresToMapping hardening looks solid. The positive_labels fail-fast check is a good call too: a custom model whose positive class isn't literally named jailbreak can't quietly mean "never detected".

Two small things I noticed:
1. The default http_classify timeout is 30s. In the request path it could be a long stall if the classifier hangs, maybe something shorter for a guardrail model would suit better.
2. With a remote endpoint the detector can now fail independently of the router. Today an inference error is logged and the rule is skipped (classifier_signal_jailbreak.go), so the request proceeds as if it had been checked and found safe. That's a reasonable default for availability, but for a security signal it'd be worth making explicit, like an on_error: `allow | block` knob, or at least surfacing "guardrail unavailable" to the caller so downstream systems can react.

Both otherwise this is a LGTM @AayushSaini101 @theohsiung 

On scope: This is only SequenceClassifierBackend interface. Currently it neglects a span level direction, i could take it in a different issue/pr.

AayushSaini101:
> Took a pass through. The alignScoresToMapping hardening looks solid. The positive_labels fail-fast check is a good call too: a custom model whose positive class isn't literally named jailbreak can't quietly mean "never detected".
> 
> Two small things I noticed:
> 
> 1. The default http_classify timeout is 30s. In the request path it could be a long stall if the classifier hangs, maybe something shorter for a guardrail model would suit better.
> 2. With a remote endpoint the detector can now fail independently of the router. Today an inference error is logged and the rule is skipped (classifier_signal_jailbreak.go), so the request proceeds as if it had been checked and found safe. That's a reasonable default for availability, but for a security signal it'd be worth making explicit, like an on_error: `allow | block` knob, or at least surfacing "guardrail unavailable" to the caller so downstream systems can react.
> 
> Both otherwise this is a LGTM @AayushSaini101 @theohsiung
> 
> On scope: This is only SequenceClassifierBackend interface. Currently it neglects a span level direction, i could take it in a different issue/pr.

thanks for the suggestions @adaamko @Xunzhuo , let me know your thoughts on the suggestions @theohsiung : )

theohsiung:
Thanks @AayushSaini101 @Xunzhuo @adaamko for the review — here's what's addressed:

**@Xunzhuo**
- `classifier_jailbreak_init.go`: `SequenceClassifierBackend` no longer leaks `candle_binding.ClassResultWithProbs`; replaced with an owned `SequenceClassificationResult` (probabilities only, argmax derived once in the policy layer) and `Classify` now takes `ctx` (5668802c).
- `config.go` enum: `backend` (candle/mmbert32k/http_chat/http_classify) split into two mutually exclusive fields — `variant` (local Candle model) and `protocol` (remote HTTP wire contract) — so future additions land on the right axis instead of growing one flat list (77d5c195).
- Operator CRD gap: added `variant`/`protocol`/`positive_labels` to the CRD, fixed two samples that had dropped `use_modernbert` without a replacement (54a382bf, fd7286d5).

**@adaamko**
- `http_classify` now defaults to a 5s timeout instead of sharing `http_chat`'s 30s (e55950ab).
- `on_error` handling deferred to a follow-up issue — it's a cross-cutting concern #2760's category/complexity/PII backends will need too, better designed once generically than per-backend here.

Also fixed two CI lint gate issues found while touching this code (unrelated to review comments, but blocking the PR either way).

Xunzhuo (approved):

Xunzhuo:
@theohsiung thanks for the great work on this!

Issue #2469, opened 2026-07-12
[Bug] Enforce supported-version and unknown-field parity across config producers


Xunzhuo:
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

Maintained config producers do not share one supported-version and unknown-field contract. Go Router, Python CLI, Dashboard/DSL, dynamic CRDs, and Operator paths can reject, ignore, or preserve different nested fields, so a document accepted by one producer can change meaning in another.

### To Reproduce

Pass the same config corpus through Router loading, CLI parsing, Dashboard/DSL conversion, both CRD paths, and Operator output. Include missing, malformed, older, and future versions plus nested unknown fields. Compare acceptance, diagnostics, and normalized output.

### Expected behavior

Supported versions are checked before interpretation, migrations are explicit, unknown fields are rejected outside named extension seams, and every maintained producer agrees on acceptance, field-path diagnostics, and normalized output.

### Impact

Producer drift can silently drop or preserve configuration, produce different runtime intent from one input, and make upgrades unsafe.

### Additional context

Completion requires:

- One owned list of supported versions and explicit migrations.
- Stable field-path diagnostics for unknown fields.
- Schemaless objects only at named extension seams with an owner validator.
- One cross-language golden corpus covering versions, migrations, nested unknown fields, and normalized output.
- Operator and CLI constants derived from or tested against the same contract.

Parent Epic: #2326
Related: #2355, #2122

ankit373:
I'd like to take the first acceptance bullet (supported-version gate) as a self-contained slice, if
that split works for you.

What I found while scoping:

- `CanonicalConfig.Version` has **no reader**. `validateCanonicalContract` checks model cards,
  duplicate names and default-model references, but never looks at `Version`.
- The only writers are in `canonical_export.go`, which hardcodes the literal `"v0.3"` twice in
  `CanonicalConfigFromRouterConfig`. There is no supported-version constant in `pkg/config` or in
  `deploy/operator`.
- `RouterConfig` has no `Version` field, so `ParseRoutingYAMLBytes` parses `version:` and discards it.

So `version: v0.1` or `version: v99` is accepted today and interpreted as v0.3, which is the
cross-control-plane drift this issue describes.

Proposed slice, Go only, no CLI/dashboard/operator changes:

1. one `SupportedCanonicalVersion` constant
2. `validateCanonicalContract` rejects unsupported and malformed versions with a `version: ...` field path
3. the two exporter literals derive from the constant
[...]

Issue #2326, opened 2026-07-05
[Epic] Build versioned configuration activation and rollback


Xunzhuo:
## Parent Direction

Enterprise & Environment — #2968

## Goal

Close the remaining gaps in the existing Router configuration APIs so Router, Dashboard, CLI, Helm, and Operator producers share one versioned snapshot, activation, status, and rollback contract.

## Background

Current main already provides `GET`, `PATCH`, and `PUT /config/router`, `/config/router/versions`, `/config/router/rollback`, `/config/hash`, ETag preconditions, timestamped backups, runtime pending/active hashes, and Dashboard deployment projections. This Epic converges and hardens those surfaces rather than introducing a parallel API family or second source of truth.

## Scope

- Define one canonical snapshot identity using stable version, generation, and hash semantics.
- Separate desired snapshot, activation state, effective runtime state, status, and diagnostic dump.
- Validate a complete candidate before activation and return structured field-path errors and warnings.
- Provide structured diff and side-effect-free dry-run behavior.
- Make activation atomic from the caller's perspective and define pending, active, failed, and known-good states.
- Make rollback durable and deterministic across restart; define multi-replica ownership and conflict behavior.
- Expose effective config and status needed to diagnose drift and partial deployment.
- Migrate Dashboard, CLI, Helm, and Operator producers through explicit compatibility shims.

## Interfaces and related work

- Data Plane & Networking consumes validated snapshots through safe reload/drain boundaries such as #2470.
- Operator, CRD, and Helm convergence is tracked by #2355.
- Dashboard projections are tracked by #1610; multi-replica Dashboard state is tracked by #3168.

## Non-goals

- Endpoint scheduling, inference-pool discovery, or provider-specific load balancing.
- Organization administration, virtual keys, tenant quotas, token rate limits, budgets, billing, or settlement.
- Credentials, login sessions, or identity-provider secrets inside generic snapshots.
- Immediate removal of compatible current endpoints.

## Completion criteria

- [ ] One documented snapshot identity and activation-state model covers current management endpoints.
- [ ] Validate, diff, and dry-run return stable structured diagnostics.
- [ ] Concurrent mutation has deterministic precondition and conflict behavior.
- [ ] Activation and known-good rollback survive restart and have defined multi-replica semantics.
- [ ] Effective status exposes active generation/hash and partial-deployment conditions.
- [ ] Dashboard, CLI, Helm, and Operator use or bridge to the same contract.
[...]

PR #2767, opened 2026-08-04
[Config] Gate canonical input on a supported version

ankit373:
Related: #2469

Rebased onto current `main`. The scope changed while this sat: **#3489 landed a minimal version gate**, so this is no longer "add a gate" but "make the gate survive a contract bump and stop duplicating the literal".

## What `main` has now

`validateCanonicalContract` calls `validateCanonicalVersion`, added by #3489:

```go
func validateCanonicalVersion(canonical *CanonicalConfig) error {
	if canonical == nil {
		return fmt.Errorf("config cannot be nil")
	}
	if canonical.Version != "" && canonical.Version != "v0.3" {
		return fmt.Errorf("unsupported config version %q: v0.3 is required", canonical.Version)
	}
	return nil
}
```

That closes the original hole in #2469: `version: v99` is no longer read as v0.3. Two things it does not do.

**1. One accepted version means the first bump breaks every config at once.** The moment that literal becomes `v0.4`, every `v0.3` document on disk stops loading. There is no window to migrate in.

**2. The literal is still spelled in eight non-test places**, so a bump is an eight-file edit, and any site missed is a document the build writes but will not read back:

| site | |
|---|---|
| `pkg/config/canonical_export.go` | x2 |
| `pkg/config/loader.go` (error text) | x2 |
| `pkg/k8s/converter.go` | x2 |
| `pkg/dsl/emitter_yaml.go` | x1 |
| `deploy/operator/controllers/canonical_config_builder.go` | x1 |

The gate is also currently untested: no test references it or its error string.

## Change

Two values, because what a build writes and what it can read are different sets:

```go
// SupportedCanonicalVersion is the canonical config contract this build writes.
const SupportedCanonicalVersion = "v0.3"

// acceptedCanonicalVersions are the contracts this build can read.
var acceptedCanonicalVersions = []string{SupportedCanonicalVersion}
```

A v0.4 release becomes `{"v0.4", "v0.3"}`: new documents are stamped v0.4, existing v0.3 ones keep loading, and dropping v0.3 later is a deliberate one-line edit rather than a side effect of the bump.

The call site and the nil check from #3489 are unchanged; only the body moves, into `canonical_version.go`. The error gains the `version:` field path and names the whole set:

```
version: unsupported "v0.2", this build reads v0.4, v0.3
```

All eight producer and error sites derive from the constant, so `v0.3` appears once in Go:

```
$ grep -rn '"v0\.3"' --include='*.go' src/ deploy/ | grep -v _test
src/semantic-router/pkg/config/canonical_version.go:13:const SupportedCanonicalVersion = "v0.3"
```

An absent `version` still warns rather than failing, unchanged from `main`. Every config in the repo omits it, and making it strict needs the migration path in #2326.

## Addressing the review

> omits the repository-required config documentation and behavior-level E2E

**Docs.** `configuration.md` documented `version` as "Use `v0.3`" and said nothing about what the Router does with it. Added a **Schema version** section (accepted, absent, rejected; the surfaces that share the gate; the bump window) and a troubleshooting entry for the startup error. Both are asserted by `configContractRequiredDocs`, so they fail `TestConfigContractDocsStayAligned` if they drift.

**E2E.** `router-config-version-gate`, registered in `BaselineRouterContract`, posts one canonical document twice to `POST /config/router/validate` on the deployed Router, changing only `version`:

| version | expected |
|---|---|
| `v0.3` | 200, `valid: true` |
| `v0.2` | 422, `CONFIG_VALIDATION_ERROR`, message contains `version:` and `v0.2` |

That endpoint runs the same `ParseYAMLBytes` → `validateCanonicalContract` path as serve and hot-reload but writes nothing, so the case asserts the runtime contract without mutating the deployed config. Asserting the supported version first means a later rejection cannot be blamed on the rest of the document.

## Validation

```
src/semantic-router  go test ./pkg/config ./pkg/dsl ./pkg/k8s   -> ok, ok, ok
deploy/operator      go test ./api/v1alpha1 ./controllers       -> ok, ok
e2e                  go build ./... && go vet ./...             -> clean
dashboard/backend    go build ./...                             -> ok
```

`gofmt` and `go vet` clean on all touched packages.

Unit coverage: supported, absent, whitespace-only, nil, older, future, unprefixed and non-version; the `validateCanonicalContract` boundary; a retained previous contract; and a check that every exporter output passes its own gate. Collapsing the set back to a single compare fails `TestValidateCanonicalVersionAcceptsRetainedContracts`; removing the gate call fails the boundary test.

**The E2E case has not been run against a live cluster**, as I don't have one to hand. It compiles and vets, and the contract it asserts is read off the handler and reproduced at the parse layer:

```
version=v0.3 err=<nil>
version=v0.2 err=version: unsupported "v0.2", this build reads v0.3
```

Happy to move the case to a different profile if `BaselineRouterContract` is the wrong home.

### DIFF
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -57,6 +57,9 @@ var BaselineRouterContract = []string{
 	"language-routing",
 	// Reask signal rule matching and routing (issue #3178)
 	"reask-routing",
+	// The canonical version gate accepts the supported contract and rejects
+	// an unsupported one before interpretation (issue #2469)
+	"router-config-version-gate",
 }
 
 // DashboardContract is the canonical E2E contract for the dashboard API surface.
--- a/src/semantic-router/pkg/config/canonical_version.go
+++ b/src/semantic-router/pkg/config/canonical_version.go
@@ -1,0 +1,39 @@
+package config
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
+)
+
+// acceptedCanonicalVersions are the contracts this build can read.
+// CanonicalConfigVersion is the one it writes; a release that bumps it keeps the
+// outgoing contract listed here for one cycle, so configs written for it keep
+// loading instead of being rejected all at once.
+var acceptedCanonicalVersions = []string{CanonicalConfigVersion}
+
+// validateCanonicalVersion gates canonical input before it is interpreted, so a
+// document written for a different contract cannot be read as this one.
+//
+// An absent version is accepted and warned about: existing documents omit it, and
+// requiring it needs a migration path (see #2326).
+func validateCanonicalVersion(canonical *CanonicalConfig) error {
+	if canonical == nil {
+		return fmt.Errorf("config cannot be nil")
+	}
+
+	v := strings.TrimSpace(canonical.Version)
+	if v == "" {
+		logging.Warnf("version: not set, interpreting as %q", CanonicalConfigVersion)
+		return nil
+	}
+
+	if !slices.Contains(acceptedCanonicalVersions, v) {
+		return fmt.Errorf("version: unsupported %q, this build reads %s",
+			canonical.Version, strings.Join(acceptedCanonicalVersions, ", "))
+	}
+
+	return nil
+}
--- a/website/docs/installation/configuration.md
+++ b/website/docs/installation/configuration.md
@@ -28,7 +28,7 @@ or runtime behavior that differs from the built-in defaults.
 
 | Section | Owns |
 | --- | --- |
-| `version` | Canonical schema version. Use `v0.3`. |
+| `version` | Canonical schema version. Use `v0.3`. See [Schema version](#schema-version). |
 | `listeners` | Public Router listeners and timeouts. |
 | `providers` | Logical provider models, physical backend endpoints, pricing, capabilities, and defaults. |
 | `evaluation` | Optional operator-owned benchmark definitions, versioned index DAGs, and model-linked records. |
@@ -101,6 +101,41 @@ complexity rule's signal errors rather than dropped.
 The [Routing Pipeline](../overview/signal-driven-decisions) explains the design.
 Capability pages under **Capabilities** document each signal, projection,
 decision, algorithm, plugin, and global block.
+
+## Schema version
+
+`version` declares the canonical contract the document is written for. The Router
+reads it before it interprets anything else, so a document written for a
+different contract fails at startup instead of being silently reinterpreted
+under the current one.
+
+| `version` | Behavior |
+| --- | --- |
+| `v0.3` | Accepted. The contract this build reads and writes. |
+| Absent | Accepted, with a startup warning, and interpreted as `v0.3`. |
+| Any other value | Rejected before the document is interpreted. |
+
+A rejected document fails with the offending field path and the set the running
+build accepts:
+
+```
+version: unsupported "v0.2", this build reads v0.3
+```
+
+`vllm-sr config validate`, `vllm-sr serve`, `POST /config/router/validate`, and
+the Dashboard and Operator config paths all apply the same gate, so a document
+that validates on one surface is one the Router will read on the others. Every
+canonical document those surfaces emit is stamped with the version they accept,
+so an exported configuration always loads back.
+
+### Contract bumps
+
+A release that changes the written contract keeps the outgoing one readable for
+one cycle. During that window the Router stamps new documents with the new
+version and still reads the previous one, so existing files keep loading and can
+be migrated incrementally rather than all at once. The error message always
+names the full set the running build reads, which is the authoritative answer
+for a given binary.
 
 ## Capability catalog
 
--- a/deploy/operator/controllers/canonical_config_builder.go
+++ b/deploy/operator/controllers/canonical_config_builder.go
@@ -9,7 +9,7 @@ import (
 
 func (r *SemanticRouterReconciler) buildCanonicalConfig(ctx context.Context, sr *vllmv1alpha1.SemanticRouter) (*routerconfig.CanonicalConfig, error) {
 	canonical := &routerconfig.CanonicalConfig{
-		Version: "v0.3",
+		Version: routerconfig.CanonicalConfigVersion,
 		Listeners: []routerconfig.Listener{
 			{
 				Name:    "grpc-50051",
--- a/e2e/testcases/router_config_version_gate.go
+++ b/e2e/testcases/router_config_version_gate.go
@@ -1,0 +1,209 @@
+package testcases
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/vllm-project/semantic-router/e2e/pkg/fixtures"
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+)
+
+func init() {
+	pkgtestcases.Register("router-config-version-gate", pkgtestcases.TestCase{
+		Description: "Verify the running Router accepts the supported canonical version and rejects an unsupported one before interpreting the document",
+		Tags:        []string{"apiserver", "config", "validation"},
+		Fn:          testRouterConfigVersionGate,
+	})
+}
+
+// versionGateDocument is one canonical document parameterized only by `version`.
+// Both cases submit the same bytes apart from that field, so a difference in the
+// outcome is attributable to the version gate and nothing else.
+const versionGateDocument = `version: %s
+listeners:
+  - name: http-8899
+    address: 0.0.0.0
+    port: 8899
+    timeout: 300s
+providers:
+  defaults:
+    model: local/general
+  models:
+    - name: local/general
+      provider_model_id: my-served-model
+      backend_refs:
+        - name: primary
+          endpoint: host.docker.internal:8000
+          protocol: http
+          provider: vllm
+routing:
+  strategy: priority
+  modelCards:
+    - name: local/general
+      modality: text
+      capabilities: [chat]
+  decisions:
+    - name: default_answer
+      description: Default route.
+      priority: 100
+      rules:
+        operator: AND
+        conditions: []
+      modelRefs:
+        - model: local/general
+`
+
+// unsupportedCanonicalVersion is a contract this build must not read. It is a
+// past contract rather than a future one, so the case stays valid after a bump.
+const unsupportedCanonicalVersion = "v0.2"
+
+type routerConfigValidateResponse struct {
+	Valid          bool   `json:"valid"`
+	NormalizedYAML string `json:"normalized_yaml"`
+}
+
+type routerConfigErrorResponse struct {
+	Error struct {
+		Code    string `json:"code"`
+		Message string `json:"message"`
+	} `json:"error"`
+}
+
+// testRouterConfigVersionGate drives POST /config/router/validate on the running
+// Router. That endpoint runs the same canonical parse as serve and hot-reload but
+// writes nothing, so the case asserts the real runtime contract without mutating
+// the deployed configuration. See issue #2469.
+func testRouterConfigVersionGate(
+	ctx context.Context,
+	client *kubernetes.Clientset,
+	opts pkgtestcases.TestCaseOptions,
+) error {
+	session, err := fixtures.OpenRouterAPISession(ctx, client, opts)
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	httpClient := session.HTTPClient(30 * time.Second)
+	url := session.URL("/config/router/validate")
+
+	supportedVersion, err := validateVersionedDocument(ctx, httpClient, url, "v0.3", opts.Verbose)
+	if err != nil {
+		return err
+	}
+
+	if err := rejectUnsupportedVersion(ctx, httpClient, url, opts.Verbose); err != nil {
+		return err
+	}
+
+	if opts.SetDetails != nil {
+		opts.SetDetails(map[string]interface{}{
+			"supported_version":   supportedVersion,
+			"unsupported_version": unsupportedCanonicalVersion,
+		})
+	}
+
+	return nil
+}
+
+// validateVersionedDocument asserts the supported contract is accepted, so a
+// later rejection cannot be blamed on the rest of the document.
+func validateVersionedDocument(
+	ctx context.Context,
+	httpClient *http.Client,
+	url string,
+	version string,
+	verbose bool,
+) (string, error) {
+	resp, err := postVersionedDocument(ctx, httpClient, url, version)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf(
+			"expected version %q to validate with 200, got %d: %s",
+			version, resp.StatusCode, truncateString(string(resp.Body), 300),
+		)
+	}
+
+	var decoded routerConfigValidateResponse
+	if err := json.Unmarshal(resp.Body, &decoded); err != nil {
+		return "", fmt.Errorf("decode validate response: %w", err)
+	}
+	if !decoded.Valid {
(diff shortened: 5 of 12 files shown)


ankit373:
`test-and-build` went red on an unrelated flake in `pkg/cache`:

```
--- FAIL: TestHybridCachePendingRequest (9.11s)
    cache_test.go:1984: Expected to find cached entry after update
    cache_test.go:1987: Response mismatch: got , want {"answer": "Quantum computing uses qubits..."}
FAIL	.../pkg/cache	132.687s
```

Why it isn't this change:

- `pkg/cache` never references `CanonicalConfig`, `ParseYAMLBytes` or `validateCanonicalContract`, so
  it doesn't go through the gate.
- The gate's error string (`version: unsupported ...`) appears zero times in the run log, so it never
  fired anywhere.
- #2769, branched from the same `main` (2ca8918), passed `test-and-build`.
- The test waits on Milvus visibility after an update. Its own log line says "this may take 30-60
  seconds" and it gave up at 9.11s, while `TestHybridCacheEviction` in the same run took 36.41s.

Later pushes re-triggered CI, so this no longer needs a manual re-run.

src/semantic-router/pkg/config/canonical_version.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
+)
+
+// SupportedCanonicalVersion is the canonical config contract this build implements.
+// Producers of canonical documents must stamp this value; consumers reject anything else.
+const SupportedCanonicalVersion = "v0.3"
```
AayushSaini101, line 13: thanks, @ankit373 cannot we import from this particular env variable, because we need to update this on the regular basis in every release thoughts cc: @FAUST-BENCHOU @drivebyer

ankit373: Pushing back on the env var, for one reason: this value isn't only the version we *accept*, it's the
version we *stamp*. `canonical_export.go`, `pkg/k8s/converter.go`, `pkg/dsl/emitter_yaml.go` and the
operator's `canonical_config_builder.go` all write it onto documents the router emits.

So it cuts both ways. Set it to `v0.4` and this build accepts a v0.4 document and reads it with v0.3
semantics, which is the exact failure the gate exists to turn into a startup error. On the producer
side, a deployment could stamp `v0.4` onto a document this binary wrote to the v0.3 contract, so the
artifact would misreport its own schema. What a build can parse is a property of the build, not of
the environment it runs in.

Your underlying concern is fair, though, and it wasn't fully solved even by this PR. Fixed in
`4ea4a52`: two migration errors in `loader.go` still spelled `v0.3` inline, so a bump needed three
edits. They take the const now, and `v0.3` appears in exactly one place in Go:

```
$ grep -rn 'v0\.3' --include='*.go' src/ deploy/ | grep -v _test | grep -v '// ' | grep -v ServiceVersion
src/semantic-router/pkg/config/canonical_version.go:12:const SupportedCanonicalVersion = "v0.3"
```

Rendered output is unchanged. `canonical_loader_test.go:67` asserts the exact error string and still
passes, and `go vet` is clean on the new format args.

A const is also what the repo already does for contract versions: `JSONRPCVersion`
(`pkg/mcp/types.go`), `SchemaVersion` (`pkg/projectiontrace/trace.go`), `AnthropicAPIVersion`
(`pkg/anthropic/client.go`), `SearchModuleMinVersion` (`pkg/utils/valkey/version.go`).

If what you want is room to accept more than one contract during a migration, that should still be
build-controlled rather than env-controlled. I've implemented it; see the next comment.

ankit373: Pushed as `6adaaff`. Building it out found a real defect in what I'd sent, so thanks for pressing.

With a single accepted version, the moment `SupportedCanonicalVersion` becomes `v0.4` every existing
v0.3 config on disk stops loading. Each contract bump would break all users at once. That is what
your "we update this every release" instinct was pointing at, and my version had it.

Now split into what the build writes and what it reads:

```go
// SupportedCanonicalVersion is the canonical config contract this build writes.
// Producers of canonical documents stamp this value.
const SupportedCanonicalVersion = "v0.3"

// acceptedCanonicalVersions are the contracts this build can read. A release that
// bumps SupportedCanonicalVersion keeps the outgoing contract listed here for one
// cycle, so configs written for it keep loading instead of being rejected at once.
var acceptedCanonicalVersions = []string{SupportedCanonicalVersion}
```

A v0.4 release becomes `{"v0.4", "v0.3"}`. New documents are stamped v0.4, existing v0.3 ones still
load, and dropping v0.3 later is a deliberate one-line edit rather than a side effect of the bump.
The error now names the set: `version: unsupported "v0.2", this build reads v0.4, v0.3`.

Producers are untouched and still stamp the single const, so a document this build writes is always
one it reads back. `TestExportStampsSupportedVersion` asserts that.

Verified by reverting the set back to the single-version compare:
`TestValidateCanonicalVersionAcceptsRetainedContracts` then fails with
`validateCanonicalVersion("v0.4") = version: unsupported "v0.4", this build supports "v0.3", want accepted`.
`pkg/config` is green, `go vet` and gofmt clean.

ankit373:
The red `Run pre-commit hooks check file lint` here is not from this PR. `TestConfigContractDocsStayAligned` fails on current `main` (`a0d75fd`) because #2773 reformatted the URLs the needles match on, and the docs-only path filter skipped the job that would have caught it. Reproduced on a pristine `main` checkout; fix in #2776.

Xunzhuo (changes requested):
This changes the public config-version/startup contract but omits the repository-required config documentation and behavior-level E2E. Please update the maintained configuration docs/proposal/tutorial surfaces as applicable and cover both supported and unsupported versions through the real CLI/runtime path.

ankit373:
@Xunzhuo addressed, and rebased onto current `main`. Description rewritten, because the scope changed while this sat: #3489 landed a minimal gate, so this is now about surviving a contract bump rather than adding the check.

**Docs.** `installation/configuration.md` gains a **Schema version** section covering accepted, absent and rejected values, the surfaces that share the gate, and a **Contract bumps** subsection for the retention window. `troubleshooting/common-errors.md` gains the startup error. Both are asserted by `configContractRequiredDocs`, so they fail `TestConfigContractDocsStayAligned` if they drift rather than going stale quietly.

**E2E.** `router-config-version-gate` in `BaselineRouterContract` posts one canonical document twice to `POST /config/router/validate` on the deployed Router, changing only `version`: `v0.3` must return 200 `valid: true`, `v0.2` must return 422 `CONFIG_VALIDATION_ERROR` naming the field and the value. That endpoint runs the same `ParseYAMLBytes` → `validateCanonicalContract` path as serve and hot-reload but writes nothing, so it asserts the runtime contract without mutating the deployed config. Asserting the supported version first means a rejection cannot be blamed on the rest of the document.

One thing to flag honestly: I have not run the E2E case against a live cluster. It builds and vets, and the request and response shapes are read off `handleConfigValidate`, but the assertion has not been observed passing end to end. Say the word if it belongs in a different profile.

What remains from #3489 is unchanged: same call site, same nil check. Only the body moves, so it accepts a set instead of one value and the eight remaining `"v0.3"` literals derive from the constant.

Xunzhuo (dismissed):
Thanks for carrying this through. The public version gate is complete, the operator manifests cover the supported skew, and the exact head integrates cleanly with current `main`. Approved.

wilsonwu (dismissed):
LGTM

ankit373:
Rebased onto current `main` (24 commits). This dismissed the approval, so flagging exactly what changed to make a re-look cheap.

One conflict, in `e2e/pkg/testmatrix/testcases.go`, and it was purely additive: `main` appended `reask-routing` to `BaselineRouterContract` while this branch appends `router-config-version-gate`. Both entries kept, no other edit.

Nothing else in the three commits changed. Reverified after the rebase:

```
src/semantic-router  go test ./pkg/config ./pkg/dsl ./pkg/k8s   ok, ok, ok
e2e                  go build ./... && go vet                   clean
```

The single-literal invariant the PR exists to create still holds:

```
$ grep -rn "\"v0\.3\"" --include=*.go src/ deploy/ | grep -v _test
src/semantic-router/pkg/config/canonical_version.go:13:const SupportedCanonicalVersion = "v0.3"
```

Context on why this was manual rather than the merge queue: Mergify dequeued the sibling PR #3617 because GitHub refuses to let a GitHub App update a branch when `.github/workflows/` has moved on main, without `workflows` permission. Its own hint says to rebase locally and it will requeue. Same situation here, so I did both by hand.

FAUST-BENCHOU (changes requested):
seems still gets conflicts

ankit373:
@FAUST-BENCHOU thanks, and rather than just resolving the conflict I have rescoped the PR, because the conflict was telling us something.

`main` has since added `CanonicalConfigVersion`, which was the constant half of this PR. That part is done, and my `SupportedCanonicalVersion` had become a second name for the same value, which is worse than what the PR set out to fix. It now builds on `CanonicalConfigVersion` instead.

What is still missing on `main` is the **set**. `validateCanonicalVersion` compares against exactly one version:

```go
if canonical.Version != "" && canonical.Version != CanonicalConfigVersion {
```

So the moment `CanonicalConfigVersion` becomes `v0.4`, every `v0.3` document on disk stops loading at once, with no window to migrate in. `acceptedCanonicalVersions` is what the build *reads*, seeded with the one it *writes*. A v0.4 release becomes `{"v0.4", "v0.3"}`, and dropping v0.3 later is a deliberate edit rather than a side effect of the bump.

Three smaller things come with it: the error gains the `version:` field path and names the accepted set, an absent version warns rather than being silently reinterpreted, and the four literals still on `main` plus the two loader error strings now derive from the constant, so `v0.3` appears once in Go:

```
$ grep -rn '"v0\.3"' --include=*.go src/ deploy/ | grep -v _test
src/semantic-router/pkg/config/canonical_config.go:13:const CanonicalConfigVersion = "v0.3"
```

Verified on the new base:

```
src/semantic-router  go test ./pkg/config ./pkg/dsl ./pkg/k8s   ok, ok, ok
deploy/operator      go test ./controllers                      ok
e2e                  go build ./... && go vet                   clean
```

On the branch itself: @wilsonwu had merged `main` in twice to keep it current, and I appreciate that. Since the rescope meant rebuilding the change on top of today's `main` anyway, the branch is now a single commit on a newer base than those merges. Nothing authored by anyone else was dropped; both merges were integrating `main`, and all of that content is in the new base. I checked the file sets match before force-pushing.

`canonical_export.go` is no longer touched, since `main` already uses the constant there.

Xunzhuo (commented):
Thanks for rebuilding the version gate after the force-push. Four contract blockers remain around gate ordering, malformed versions, cross-surface parity, and whether the new E2E runs in exact-head CI.

src/semantic-router/pkg/config/canonical_version.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
+)
+
+// acceptedCanonicalVersions are the contracts this build can read.
+// CanonicalConfigVersion is the one it writes; a release that bumps it keeps the
+// outgoing contract listed here for one cycle, so configs written for it keep
+// loading instead of being rejected all at once.
+var acceptedCanonicalVersions = []string{CanonicalConfigVersion}
+
+// validateCanonicalVersion gates canonical input before it is interpreted, so a
```
Xunzhuo, line 17: **Gate the raw version before normalization**

Impact: `ParseYAMLBytes` runs raw normalization, environment rewriting, canonical-shape checks, and known-field validation before it reaches this function. A foreign-version document containing a future or unknown field can therefore fail under the current schema before the version gate, after parts of the document have already been rewritten.

Required change: Validate the raw top-level version immediately after the YAML map is parsed, before any normalizer or interpreter. Add a regression combining an unsupported version with an unknown or deprecated field and assert the version error wins.

src/semantic-router/pkg/config/canonical_version.go
```diff
@@ -0,0 +1,39 @@
...
+// document written for a different contract cannot be read as this one.
+//
+// An absent version is accepted and warned about: existing documents omit it, and
+// requiring it needs a migration path (see #2326).
+func validateCanonicalVersion(canonical *CanonicalConfig) error {
+	if canonical == nil {
+		return fmt.Errorf("config cannot be nil")
+	}
+
+	v := strings.TrimSpace(canonical.Version)
```
Xunzhuo, line 27: **Reject malformed version spelling instead of trimming it**

Impact: Trimming turns `"   "` into an absent version and `" v0.3 "` into the supported version. Main's exact comparison rejected both, and the new public contract allows only a genuinely absent value, so this silently broadens accepted input.

Required change: Compare the original value exactly and reserve the compatibility fallback for `""`. Add rejection cases for whitespace-only and padded versions.

website/docs/installation/configuration.md
```diff
@@ -102,6 +102,41 @@ The [Routing Pipeline](../overview/signal-driven-decisions) explains the design.
...
+| Any other value | Rejected before the document is interpreted. |
+
+A rejected document fails with the offending field path and the set the running
+build accepts:
+
+```
+version: unsupported "v0.2", this build reads v0.3
+```
+
+`vllm-sr config validate`, `vllm-sr serve`, `POST /config/router/validate`, and
```
Xunzhuo, line 125: **Make the CLI and Router share the promised contract**

Impact: The CLI first enforces a generated schema whose version is fixed to `v0.3`, then requires `version` in Pydantic. The Router accepts missing or whitespace versions and can retain multiple versions, so the same document is still accepted differently across these surfaces.

Required change: Derive every reader's version gate from one owned contract and add a shared cross-language corpus covering absent, malformed, current, retained, and unsupported versions.

e2e/pkg/testmatrix/testcases.go
```diff
@@ -57,6 +57,9 @@ var BaselineRouterContract = []string{
 	"language-routing",
 	// Reask signal rule matching and routing (issue #3178)
 	"reask-routing",
+	// The canonical version gate accepts the supported contract and rejects
+	// an unsupported one before interpretation (issue #2469)
+	"router-config-version-gate",
```
Xunzhuo, line 62: **Run this contract case in exact-head CI**

Impact: The Envoy AI Gateway workflow replaces this profile's contract with a hard-coded 20-test list that omits `router-config-version-gate`. The green exact-head E2E registered the case but never selected or ran it, so the new runtime behavior remains unverified.

Required change: Include this case in the CI-selected list (or remove the override), then rerun exact-head CI and verify the log shows the test executed.

ankit373:
@Xunzhuo addressed all four, each with a test that fails without the fix.

**Gate ordering.** Version is now read straight off the parsed YAML map before normalization, env rewriting, or known-field validation run. `TestUnsupportedVersionOutranksOtherRejections` combines an unsupported version with a deprecated field and asserts the version error wins.

**Malformed spelling.** Comparison is exact now. `"   "` and `" v0.3 "` are rejected; the compatibility fallback stays reserved for `""` only, since an omitted field decodes to that anyway.

**CLI/Router parity.** Found a real bug here: the schema had no `required` for `version` (absent key already validated), but `UserConfig.version` was a required Pydantic field, so the CLI's full pipeline rejected a document the Router accepts with a warning. Fixed by defaulting it to `""`. The schema's `version` property is now generated from `AcceptedCanonicalVersions()` (enum + the `""` fallback) instead of restated, so it can't drift from the Go gate independently.

Added `router_config_version_gate_conformance.v1.json`: absent/empty/malformed/current/retained cases, with a Go test and a Python test each replaying it against their own surface. Verified it catches the parity bug (reverting the Pydantic default fails the Python side on `version-key-absent`).

**CI coverage.** `router-config-version-gate` added to the Envoy AI Gateway job's hard-coded list.

One flag: main independently reworked `TestRenderTopLevelCollectionAndScalarSections` (the config-schema view renderer moved to a compact `shape`/`fields` format), unrelated to this PR. Didn't rebase onto it, since that test needs updating for the new shape regardless and I wanted this round to land first. Will hit it on the next rebase.

Verified: `go build ./...` (router, e2e, operator), `go vet`, `go test ./pkg/config/... ./pkg/configschema/... ./pkg/dsl/... ./pkg/k8s/...`, and the Python suite including every existing test touching `parse_user_config`/`UserConfig` (two pre-existing `listeners.0.api_keys` failures, confirmed present on the unmodified branch).

PR #2769, opened 2026-08-04
config: honour enabled on the domain and PII classifiers

### LINKED ISSUE #2768: classifier.domain.enabled and classifier.pii.enabled are silently ignored
### What happened

`global.model_catalog.modules.classifier.domain.enabled` and `...classifier.pii.enabled` are silently
ignored. Their sibling modules accept `enabled`, so a config that sets all of them looks consistent and
two of them do nothing.

`CategoryModel` and `PIIModel` (`pkg/config/model_config_types.go`) have no `Enabled` field.
`PromptGuardConfig` does:

| module | embedded type | `enabled` field |
|---|---|---|
| `prompt_guard` | `PromptGuardConfig` | yes |
| `hallucination_mitigation` | n/a | yes |
| `classifier.domain` | `CategoryModel` | **no** |
| `classifier.pii` | `PIIModel` | **no** |

Classification is instead gated implicitly on the model being configured, in
`pkg/classification/classifier_builtin_models.go:183`:

```go
return c.Config.PIIModel.ModelID != "" && c.Config.PIIMappingPath != "" && c.PIIMapping != nil
```

So with `model_ref` set (which `canonical_global.go:278` resolves into `PII.ModelID`), writing
`enabled: false` leaves PII classification **running**.

### Where it already bites

`e2e/config/onnx-binding/config.onnx-binding-test.yaml` sets all five modules off:

```yaml
    modules:
      classifier:
        domain:
          enabled: false      # ignored
        pii:
          enabled: false      # ignored
      hallucination_mitigation:
        enabled: false        # honoured
      prompt_guard:
        enabled: false        # honoured
      feedback_detector:
        enabled: false        # honoured
```

Two of the five don't take effect, so that suite may not be running the configuration it declares.
`config.onnx-classifiers-test.yaml` has the same shape.

`WarnUnknownFields` does flag these at startup, but only as a log line, and the field looks valid
because the neighbouring modules accept it.

### Expected

Either `enabled` is honoured on both classifier modules, or it is rejected so the mistake is visible.
Silently accepting it on a PII control is the worst of the three.

### Notes

Found while scoping the version gate for #2469 (PR #2767). Related: the unknown-field half of #2469, where
7 of 96 canonical configs in this repo currently contain unknown fields, inventory posted there.

Happy to send the fix once you decide which way `enabled` should resolve when `model_ref` is also set.
That's a behaviour question on a security control, so I didn't want to pick unilaterally.

ankit373:
Fixes #2768.

`CategoryModel` and `PIIModel` have no `Enabled` field, so `classifier.domain.enabled` and
`classifier.pii.enabled` are parsed as unknown keys and dropped. `prompt_guard`,
`hallucination_mitigation` and `feedback_detector` all accept `enabled`, so a config that turns all
five off looks consistent while two of them keep running.

`e2e/config/onnx-binding/config.onnx-binding-test.yaml` and `config.onnx-classifiers-test.yaml` both
set all five to `false`, so those suites are not running the configuration they declare.

Classification is currently gated implicitly on the model being configured
(`classifier_builtin_models.go:183`), and `canonical_global.go:278` resolves `model_ref` into
`PII.ModelID`, so `enabled: false` plus a `model_ref` leaves PII classification active.

## Why `*bool` and not `bool`

A plain `bool` makes the zero value `false`, which disables both classifiers for every config built
without the canonical defaults. That replaces "the flag does nothing" with "the flag silently turned
your PII scanner off".

This is not theoretical. I tried the plain-bool version on review feedback (`fe5b968`) and CI failed
four packages: `pkg/classification`, `pkg/apiserver`, `pkg/modeldownload` and `pkg/extproc`. All of
them construct a config in Go rather than loading YAML, so they never reach
`DefaultCanonicalGlobal()`. Reverted in `e3e05a3`.

| `enabled` | behaviour |
|---|---|
| absent (`nil`) | active whenever a model is configured, unchanged from today |
| `false` | disabled, even with `model_ref` set |
| `true` | active, but still requires a configured model |

Nothing changes for any existing config. Only a config that explicitly writes `enabled` sees a
difference, which is the point.

This also matches `PreferenceModelConfig.UseContrastive` in the same struct file, which is a `*bool`
defaulted to true through `WithDefaults()` and read through `ContrastiveEnabled()`.

## Change

- `Enabled *bool` on `CategoryModel` and `PIIModel`, plus an `Active()` method per type over one shared
  `moduleActive` helper.
- All four gates check it, in the same order `IsPromptGuardEnabled` uses, explicit flag first and then
  configuration completeness:
  - `config.IsPIIClassifierEnabled`, `config.IsCategoryClassifierEnabled`
  - `classification.IsPIIEnabled`, `classification.IsCategoryEnabled`
- `config/config.yaml` documents `enabled` on both modules, required by
  `TestReferenceConfigCoversCanonicalPublicSurface`, which caught the omission.

## Testing

`TestClassifierEnabledFlag` covers absent, true and false across both classifiers.
`TestClassifierEnabledDoesNotOverrideMissingConfig` keeps `enabled: true` from implying a model exists.
`TestClassifierEnabledParsesFromYAML` proves the key now reaches the struct instead of being dropped.

Two revert checks: changing `moduleActive` to `enabled != nil && *enabled` (the plain-bool semantics)
fails the absent case, and removing `Active()` from the PII gate fails the explicitly-disabled case.

```
go test ./pkg/config ./pkg/dsl ./pkg/k8s              -> ok, ok, ok
cd deploy/operator && go test ./api/v1alpha1 ./controllers -> ok, ok
```

`gofmt` and `go vet` clean; both modules build.

**Not run locally:** `./pkg/classification` needs the candle native binding and fails to link on this
machine without `cargo`. It fails identically on a pristine `origin/main` checkout, so it is an
environment prerequisite rather than something in this change. `go vet ./pkg/classification` is clean
and CI covers it.

## Open question

I picked `nil` = unchanged because it is the only non-breaking reading. If you would rather `enabled` be
required on these modules, that is a migration and belongs with #2469/#2326, and I'm happy to change it.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -2286,6 +2286,9 @@ global:
         use_vllm: false
       classifier:
         domain:
+          # Omit to keep the historical behaviour of running whenever a model is
+          # configured; false disables the classifier outright.
+          enabled: true
           model_ref: domain_classifier
           model_id: models/mmbert32k-intent-classifier-merged
           threshold: 0.5
@@ -2306,6 +2309,9 @@ global:
           threshold: 0.6
           timeout_seconds: 15
         pii:
+          # Omit to keep the historical behaviour of running whenever a model is
+          # configured; false disables the classifier outright.
+          enabled: true
           model_ref: pii_classifier
           model_id: models/mmbert32k-pii-detector-merged
           threshold: 0.9
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -2286,8 +2286,6 @@ global:
         use_vllm: false
       classifier:
         domain:
-          # Omit to keep the historical behaviour of running whenever a model is
-          # configured; false disables the classifier outright.
           enabled: true
           model_ref: domain_classifier
           model_id: models/mmbert32k-intent-classifier-merged
@@ -2309,8 +2307,6 @@ global:
           threshold: 0.6
           timeout_seconds: 15
         pii:
-          # Omit to keep the historical behaviour of running whenever a model is
-          # configured; false disables the classifier outright.
           enabled: true
           model_ref: pii_classifier
           model_id: models/mmbert32k-pii-detector-merged
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -2286,6 +2286,8 @@ global:
         use_vllm: false
       classifier:
         domain:
+          # Omit to keep the historical behaviour of running whenever a model is
+          # configured; false disables the classifier outright.
           enabled: true
           model_ref: domain_classifier
           model_id: models/mmbert32k-intent-classifier-merged
@@ -2307,6 +2309,8 @@ global:
           threshold: 0.6
           timeout_seconds: 15
         pii:
+          # Omit to keep the historical behaviour of running whenever a model is
+          # configured; false disables the classifier outright.
           enabled: true
           model_ref: pii_classifier
           model_id: models/mmbert32k-pii-detector-merged
--- a/dashboard/frontend/e2e/feedback-buttons.spec.ts
+++ b/dashboard/frontend/e2e/feedback-buttons.spec.ts
@@ -63,9 +63,11 @@ test.describe('FeedbackButtons', () => {
 
   test('thumbs up sends correct feedback payload', async ({ page }) => {
     let feedbackPayload: Record<string, unknown> | null = null
+    let idempotencyKey: string | null = null
 
     await page.route('**/api/router/v1/router/outcomes', async (route) => {
       feedbackPayload = route.request().postDataJSON()
+      idempotencyKey = route.request().headers()['idempotency-key'] ?? null
       await route.fulfill({
         status: 200,
         headers: { 'Content-Type': 'application/json' },
@@ -84,6 +86,7 @@ test.describe('FeedbackButtons', () => {
     await thumbsUp.click()
 
     expect(feedbackPayload).not.toBeNull()
+    expect(idempotencyKey).toBeTruthy()
     expect(feedbackPayload!.replay_id).toBe(MOCK_REPLAY_ID)
     expect(feedbackPayload!.source).toBe('user')
     expect(feedbackPayload!.target).toBe('model')
@@ -129,9 +132,11 @@ test.describe('FeedbackButtons', () => {
 
   test('thumbs down sends underpowered outcome', async ({ page }) => {
     let feedbackPayload: Record<string, unknown> | null = null
+    let idempotencyKey: string | null = null
 
     await page.route('**/api/router/v1/router/outcomes', async (route) => {
       feedbackPayload = route.request().postDataJSON()
+      idempotencyKey = route.request().headers()['idempotency-key'] ?? null
       await route.fulfill({
         status: 200,
         headers: { 'Content-Type': 'application/json' },
@@ -150,6 +155,7 @@ test.describe('FeedbackButtons', () => {
 
     await expect(thumbsDown).toHaveAttribute('aria-pressed', 'true')
     expect(feedbackPayload).not.toBeNull()
+    expect(idempotencyKey).toBeTruthy()
     expect(feedbackPayload!.replay_id).toBe(MOCK_REPLAY_ID)
     expect(feedbackPayload!.target_ref).toBe(MOCK_MODEL)
     expect(feedbackPayload!.verdict).toBe('underpowered')
--- a/dashboard/frontend/src/components/FeedbackButtons.tsx
+++ b/dashboard/frontend/src/components/FeedbackButtons.tsx
@@ -99,7 +99,10 @@ const FeedbackButtons: React.FC<FeedbackButtonsProps> = ({
 
       const res = await fetch(OUTCOMES_API, {
         method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
+        headers: {
+          'Content-Type': 'application/json',
+          'Idempotency-Key': crypto.randomUUID(),
+        },
         body: JSON.stringify(body),
       })
 
--- a/e2e/testcases/security_policy_apply.go
+++ b/e2e/testcases/security_policy_apply.go
@@ -74,17 +74,18 @@ func newTestPolicy() securityPolicyPayload {
 	return securityPolicyPayload{
 		RoleMappings: []roleMappingPayload{
 			{
-				Name:      "premium",
-				Subjects:  []subjectPayload{{Kind: "Group", Name: "paying-customers"}},
-				Role:      "premium_tier",
-				ModelRefs: []string{"gpt-4"},
+				Name:     "premium",
+				Subjects: []subjectPayload{{Kind: "Group", Name: "paying-customers"}},
+				Role:     "premium_tier",
+				// Match Helm chart default routing.modelCards name after values merge.
+				ModelRefs: []string{"replace-with-your-model"},
 				Priority:  10,
 			},
 			{
 				Name:      "free",
 				Subjects:  []subjectPayload{{Kind: "Group", Name: "free-users"}},
 				Role:      "free_tier",
-				ModelRefs: []string{"gpt-3.5-turbo"},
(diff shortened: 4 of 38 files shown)


drivebyer (commented):
Thanks for the fix, verified it works end to end. One suggestion: since `resolveCanonicalGlobal` merges over `DefaultCanonicalGlobal()`, a plain `bool` defaulted to `true` in `defaultClassifierModule()` would keep unmentioned configs unchanged too, matching how `prompt_guard` does it, and drops the `*bool` / `Active()` machinery.

ankit373:
Done in `9b9e1d5`. Plain `bool` defaulted in `defaultClassifierModule()`, `*bool` and `Active()` gone,
so it reads like `prompt_guard`.

I checked the merge holds on every path first. `DefaultGlobalConfig()` is itself
`DefaultCanonicalGlobal()` then `applyCanonicalGlobal()`, so all three producers of a `RouterConfig`
inherit the default. The one branch where a zero `bool` would clobber it is the typed-override path in
`resolveCanonicalGlobal`, and that is unreachable today: `normalizeCanonicalConfig` is the only
caller, and `attachCanonicalGlobalOverride` always sets `globalOverrideRaw` when `global:` is present,
so the raw payload wins and only carries keys the user actually wrote.

It cost six existing specs, which build a `RouterConfig` without the canonical defaults:

- `routing_signal_usage_test.go`, `newRoutingSignalUsageTestConfig()` (4 subtests)
- `config_test.go`, the two "properly configured" cases, via `parseLegacyRuntimeConfigForTest` which
  does `cfg := &RouterConfig{}` plus `yaml.Unmarshal` with no defaults

Both now set `enabled` explicitly, same as the `PromptGuard` literal three lines below in that
fixture. Worth flagging: a hand-built `RouterConfig` has to opt in now.

Verified by reverting each half separately with `-count=1`:

- drop `Enabled: true` from `defaultClassifierModule()`, and `TestClassifierEnabledDefaultsToOn` fails
- drop the `Enabled &&` gate, and `TestClassifierEnabledGatesRuntimePredicates/disabled_but_configured`
  fails with `IsPIIClassifierEnabled() = true, want false`, which is #2768 again

The tests exercise `resolveCanonicalGlobal` rather than struct literals, since the defaults merge is
where the safety now lives. `pkg/config` green (357 specs), `go build ./pkg/...` clean, gofmt clean.

ankit373:
One more check, since the whole question was whether omitting the key can silently disable a
classifier. I ran the real `config/config.yaml` through `ParseYAMLBytes` in three states:

| `config/config.yaml` | `IsCategoryClassifierEnabled` | `IsPIIClassifierEnabled` |
|---|---|---|
| as shipped (`enabled: true`) | true | true |
| flipped to `enabled: false` | false | false |
| both `enabled:` keys deleted | true | true |

Row 2 is #2768 fixed. Row 3 is the case the `*bool` was guarding, now covered by the default in
`defaultClassifierModule()`.

Nothing outside the defaults can miss it either. The only non-test constructions of these structs in
the tree are the defaults themselves:

```
$ grep -rn 'CategoryModel{\|PIIModel{' --include='*.go' src/ deploy/ | grep -v _test
src/semantic-router/pkg/config/canonical_defaults.go:247:  CategoryModel: CategoryModel{
src/semantic-router/pkg/config/canonical_defaults.go:257:  PIIModel: PIIModel{
```

The operator and k8s converter reach the router as YAML, so they go through the same merge.

No code change from this, `fe5b968` is still the tip. I dropped the harness rather than ship it: it
works by string-replacing the reference config and would rot the first time that file is reformatted.
The committed tests cover both behaviours through `resolveCanonicalGlobal`.

ankit373:
Correcting something I said above, because I argued your side better than my own and you should have
the full picture before this merges.

I said the `*bool` was the outlier and plain `bool` was the house pattern. That holds for the flat
toggles in `model_config_types.go` (`prompt_guard`, `mcp`, `feedback_detector`, `prompt_compression`),
but not for the package: `pkg/config` has 11 non-test `Enabled *bool` fields. Two of them sit closer to
this change than `prompt_guard` does.

`defaultClassifierModule()`, the function you pointed me at, already defaults a tri-state pointer on
the field immediately after the one I changed:

```go
Preference: PreferenceModelConfig{
    UseContrastive: canonicalBoolPtr(true),
},
```

And `model_config_types.go`, the file I deleted `Active()` from, already has that accessor shape three
structs above `CategoryModel`:

```go
func (c PreferenceModelConfig) ContrastiveEnabled() bool {
    return *c.WithDefaults().UseContrastive
}
```

So both patterns are established here, and "match `prompt_guard`" and "match `Preference`" point
opposite ways inside the same function.

The real difference:

- plain `bool` is correct because the typed-override branch of `resolveCanonicalGlobal` is unreachable.
  Nothing enforces that, it holds by reachability. A second caller, or an operator path that builds a
  `CanonicalConfig` in Go instead of YAML, would marshal `enabled: false` from a zero value and
  silently disable both classifiers.
- `*bool` cannot express that bug regardless of caller, at the cost of an accessor.

The six spec changes are the evidence that the switch isn't behaviour-neutral: any Go caller building a
`RouterConfig` has to opt in now.

Not asking you to revert. You own the convention, `prompt_guard` consistency is a real argument, and
`fe5b968` stays as the tip. I just don't want it merged on a claim of mine that was wrong. If you'd
rather have the pointer version back, say so and it's one commit.

ankit373:
CI settled the question, and not the way I reported it. I've reverted to the pointer version in
`e3e05a3`, which puts the tree back to `23dd579`.

The plain `bool` broke four packages. My local run only covered `pkg/config`, which is why I missed it:

| package | failures |
|---|---|
| `pkg/classification` | `TestClassifier` (16 specs: PII init, PII classification, domain signal), `TestNewLegacyClassifierFromConfigRequiresUsedCoreSignalMappings`, `TestLoadClassifierMappingsRequiresUsedCoreSignalMappings`, `TestSignalGroupSoftmaxExclusiveChoosesSingleDomainWinner`, `TestInitializeRuntimeInitializesCoreSignalClassifiersWhenUsed` |
| `pkg/apiserver` | the four `TestBuildModelsInfoResponse*` tests |
| `pkg/modeldownload` | `TestBuildModelSpecsIncludesUsedCoreClassifierModels`, `TestBuildModelSpecsIncludesConfigDerivedRequiredFiles` |
| `pkg/extproc` | fails in the same run |

Representative failure: `runtime_state_test.go:271: expected model info entries from live config`.

The cause is the one I raised as theoretical and then talked myself out of. All of these build a config
in Go rather than loading YAML, so they never touch `DefaultCanonicalGlobal()` and a plain bool reads
`false`. It isn't the six specs I found in `pkg/config`, it's every hand-built config in the tree.

For contrast, `23dd579` (the pointer version) is green: 42 success, 9 skipped, 4 neutral, zero
failures.

So the reachability argument was too narrow. It's true that the typed-override branch of
`resolveCanonicalGlobal` is unreachable, but that was never the whole exposure. Any Go construction of
`Classifier`/`RouterConfig` is, and there are a lot of them.

If you still want plain `bool` for consistency with `prompt_guard`, it's doable, but the price is
setting `Enabled: true` in roughly 25 fixtures across four packages, and every future test that builds
one of these structs has to remember. Happy to do that if you'd prefer it. Otherwise the pointer
version matches `PreferenceModelConfig.UseContrastive` in the same struct file and needs no fixture
changes.

One unrelated failure in the same run: `TestConfigContractDocsStayAligned` says
`website/docs/api/apiserver.md is missing required text "http://localhost:8080"`. This branch touches
no docs, and that string is present 23 times on current `main`, so I don't think it's from this PR.

ankit373:
The red `Run pre-commit hooks check file lint` here is not from this PR. `TestConfigContractDocsStayAligned` fails on current `main` (`a0d75fd`) because #2773 reformatted the URLs the needles match on, and the docs-only path filter skipped the job that would have caught it. Reproduced on a pristine `main` checkout; fix in #2776.

drivebyer (approved):
LGTM

PR #2773, opened 2026-08-04
[Docs] refresh API reference with curl examples.

AayushSaini101:
## Summary

The public API reference was outdated relative to the live router surface.
This updates the docs so a first-time user can understand **which port to call**,
**what each endpoint does**, and **how a request/response looks**.

### Changes

- **Apiserver (`:8080`)** — `website/docs/api/apiserver.md`
  - Quick start: `/health` → `/api/v1/classify/intent`
  - Auth + common error envelope
  - Full endpoint index aligned with `routes_catalog.go` (eval, NLI, embeddings,
    similarity, models, config/recipes/KBs, memory/vector/files, outcomes,
    startup-status, …)
  - Worked `curl` examples with sample JSON responses

- **Router data plane (`:8801`)** — `website/docs/api/router.md`
  - Clear port split vs apiserver
  - `curl` examples for Chat Completions and Responses API
  - Router Replay section: list / detail / aggregate / trajectory + linking
    outcomes back to `:8080`

Fixes #2772

## Test plan

- [x] Targeted `markdownlint` on the two updated API docs
- [x] Spot-check paths against apiserver route catalog and ExtProc replay handlers
- [ ] Preview docs locally (`cd website && npm start`) and review API sidebar pages
- [ ] Confirm `/docs` and `/openapi.json` tip still matches a running apiserver

PR #2776, opened 2026-08-04
test(config): unfence the apiserver doc needles so a table is not a failure

### LINKED ISSUE #2777: TestConfigContractDocsStayAligned fails on main: #2773 reformatted the URLs the needles match on
`TestConfigContractDocsStayAligned` fails on current `main` (`a0d75fd`), so every PR whose changed files
route through the Go tests for `pkg/config` is red for a reason unrelated to the PR.

## Reproduce

```
$ git checkout a0d75fd && cd src/semantic-router
$ go test ./pkg/config/ -count=1 -run TestConfigContractDocsStayAligned
    docs_contract_test.go:609: website/docs/api/apiserver.md is missing required text "`http://localhost:8080`"
FAIL
```

## Cause

`configContractRequiredDocs` asserts on exact backtick placement for
`website/docs/api/apiserver.md`:

```go
"`http://localhost:8080`",
"`GET /openapi.json`",
"`GET /config/router`",
```

#2773 reorganised that page. Nothing documented was lost, but both spans changed shape:

- the base URL now sits inside a longer span, `` `GET http://localhost:8080/api/v1` ``
- `/config/router` moved into an endpoint table, so it is two spans in two cells:
  `| `GET` | `/config/router` | Current router config as JSON |`

Counts on `main`: bare `http://localhost:8080` appears 23 times, the fenced `` `http://localhost:8080` ``
zero times. `/config/router` appears 16 times. So the page documents more than it did before and the
assertion still fails.

The zh-Hans translation at
`website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/api/apiserver.md` was not touched by #2773
and still matches all three needles, so the two pages have drifted.

## Why it merged green

On #2773, both `test-and-build` and `Run pre-commit hooks check file lint` were **skipped**, because the
changed-files gate saw a docs-only change. The assertion that guards those docs lives in
`src/semantic-router/pkg/config/docs_contract_test.go`, so nothing ran it.

That part seems worth a separate look: a Go test that asserts on docs content will keep getting bypassed
by a docs-only path filter, so this can recur with any docs edit.

## Fix

PR #2776 drops the fences and shares one needle slice between the English page and the zh-Hans
translation, so the assertion survives an endpoint moving between prose and a table. It still guards
content: replacing `/config/router` in the page fails the test.

ankit373:
`TestConfigContractDocsStayAligned` fails on current `main` (`a0d75fd`), so every PR that runs the
Go tests over `pkg/config` is red through no fault of its own.

```
$ git checkout a0d75fd && cd src/semantic-router
$ go test ./pkg/config/ -count=1 -run TestConfigContractDocsStayAligned
    docs_contract_test.go:609: website/docs/api/apiserver.md is missing required text "`http://localhost:8080`"
FAIL
```

## Cause

The needles asserted on exact backtick placement:

```go
"`http://localhost:8080`",
"`GET /openapi.json`",
"`GET /config/router`",
```

#2773 reorganised the page, and both of those spans changed shape without losing anything:

- the base URL is now inside a longer span, `` `GET http://localhost:8080/api/v1` ``
- `/config/router` moved into an endpoint table, so it is two spans in two cells:
  `| `GET` | `/config/router` | Current router config as JSON |`

Counts on `main` for `website/docs/api/apiserver.md`: bare `http://localhost:8080` 23 times, the
fenced `` `http://localhost:8080` `` 0 times. So the page documents more than it did before and the
test still fails.

## Why it merged green

On #2773 both `test-and-build` and `Run pre-commit hooks check file lint` were **skipped**, because the
changed-files gate saw a docs-only change. The assertion that guards those docs lives in
`src/semantic-router/pkg/config/docs_contract_test.go`, so nothing ran it. Worth a thought separately:
a Go test that asserts on docs content will keep getting bypassed by a docs-only path filter.

## Change

Test-only. The three needles lose the fences and move to one `apiserverDocNeedles` slice shared by the
English page and the zh-Hans translation, which #2773 did not touch and which still matched the old
form. Sharing them keeps the two from drifting the next time either page is rewritten.

The assertion still guards content rather than formatting. Replacing `/config/router` in the page (16
occurrences) still fails:

```
docs_contract_test.go:610: website/docs/api/apiserver.md is missing required text "/config/router"
```

`go test ./pkg/config -count=1` passes, `go vet` and gofmt clean.

I hit this on #2767 and #2769 and initially assumed it was mine, so opening it separately rather than
folding it into either.

Closes #2777

### DIFF
--- a/src/semantic-router/pkg/config/docs_contract_test.go
+++ b/src/semantic-router/pkg/config/docs_contract_test.go
@@ -16,6 +16,15 @@ func repoRel(parts ...string) string {
 	return filepath.Join(parts...)
 }
 
+// apiserverDocNeedles is shared by the English page and its zh-Hans translation.
+// These stay unfenced so the assertion survives the page moving an endpoint between
+// prose and a table, which is what broke it in #2773.
+var apiserverDocNeedles = []string{
+	"http://localhost:8080",
+	"/openapi.json",
+	"/config/router",
+}
+
 var configContractRequiredDocs = []docNeedles{
 	{
 		path: "config/README.md",
@@ -132,12 +141,8 @@ var configContractRequiredDocs = []docNeedles{
 		},
 	},
 	{
-		path: repoRel("website", "docs", "api", "apiserver.md"),
-		needles: []string{
-			"`http://localhost:8080`",
-			"`GET /openapi.json`",
-			"`GET /config/router`",
-		},
+		path:    repoRel("website", "docs", "api", "apiserver.md"),
+		needles: apiserverDocNeedles,
 	},
 	{
 		path: repoRel("website", "docs", "troubleshooting", "common-errors.md"),
@@ -209,12 +214,8 @@ var configContractRequiredDocs = []docNeedles{
 		},
 	},
 	{
-		path: repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "api", "apiserver.md"),
-		needles: []string{
-			"`http://localhost:8080`",
-			"`GET /openapi.json`",
-			"`GET /config/router`",
-		},
+		path:    repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "api", "apiserver.md"),
+		needles: apiserverDocNeedles,
 	},
 	{
 		path: repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "proposals", "nvidia-dynamo-integration.md"),


theohsiung (approved):
Verified locally: checked out this branch, `go test ./pkg/config/... -run "TestConfigContractDocsStayAligned|TestCurrentConfigDocsAvoidRetiredCanonicalExamples" -v` passes (was failing on main before this), full `go test ./pkg/config/...`, `go vet`, and `gofmt` are all clean. Diff is minimal and test-only (single file, 13/-12), and the fix correctly de-couples the assertion from markdown formatting (prose vs. table) while still guarding actual content - confirmed the zh-Hans translation (untouched by #2773, still in the old backtick-wrapped form) still matches the now-unfenced needles, since a wrapped string still contains its own unwrapped substring.

The two red checks here (`check-linked-issue`, `component-benchmarks`) aren't about this change: `check-linked-issue` ran on the PR-opened payload before `Closes #2777` was added to the body (stale, needs a re-run), and `component-benchmarks` failed on `download-models-perf` hitting a CDN "Local entry not found" error - unrelated infra flake, not this diff.

@AayushSaini101 flagging you in since #2773 is what this fixes - might be useful for you to glance at, and if you have write access it'd unblock re-running that stale `check-linked-issue` job.

drivebyer (approved):

PR #2778, opened 2026-08-05
fix: prevent substring memory leaks on payload truncations in replay …

### LINKED ISSUE #2780: [Router][Memory] resolve OOM risk from Go string-slice retention in router replay payload truncation
## Summary

Router Replay truncates large payload strings (`RequestBody`, `ResponseBody`, `Prompt`, `ToolDefinitions`, tool-step `Arguments` / `Output`) with plain Go slicing (`s[:max]`). Because Go string slices share the original backing array, truncating a multi-MB payload to a few KB can still pin the **full** original allocation until the truncated record is released. Under heavy traffic that retained memory can grow into an OOM.

**Fix:** wrap truncated substrings with `strings.Clone()` so each stored field owns a small independent backing array.

## Problem

In `src/semantic-router/pkg/routerreplay/recorder.go`, truncation currently looks like:

```go
rec.RequestBody = rec.RequestBody[:policy.maxBodyBytes]
rec.ResponseBody = rec.ResponseBody[:policy.maxBodyBytes]
rec.Prompt = rec.Prompt[:max]
rec.ToolDefinitions = rec.ToolDefinitions[:max]
// similarly for ToolTraceStep.Arguments / Output
```

Go string slicing does **not** copy bytes. If `RequestBody` was a 50MB string and policy caps at ~4KB, the truncated value still references the 50MB array. That retention lasts as long as the truncated record lives in:

- the recorder / async write path
- the in-memory replay store
- any other long-lived holder of `RoutingRecord`

Related prior work already bounds *length* and *drop* policy (e.g. capture switches, `#2748`, `#1835` step caps), but length limits alone do not release the backing store after a slice truncate.

## Impact

- Memory leak / high RSS when `router_replay` captures large bodies or tool traces
- Potential OOM under sustained large-payload traffic even when `max_body_bytes` / `max_tool_trace_bytes` look “safe”
- Affects deployments that enable request/response body capture or large tool traces

## Proposed solution

After every truncation that keeps a prefix of a large string, allocate a fresh copy:

```go
rec.RequestBody = strings.Clone(rec.RequestBody[:policy.maxBodyBytes])
```

Apply the same pattern to:

- `rec.RequestBody`
- `rec.ResponseBody`
- `rec.Prompt`
- `rec.ToolDefinitions`
- `step.Arguments`
- `step.Output`

(and any shared helper such as `truncateBody` / `truncateToolTraceStep` so all paths stay consistent, including `UpdateToolTrace`).

`strings.Clone` forces a new backing array sized to the truncated content and severs the reference to the original large payload so GC can reclaim it.

## Acceptance criteria

- [ ] All router-replay truncation sites that keep a sliced prefix use `strings.Clone` (or equivalent copy)
- [ ] Unit tests cover truncated fields after large-input truncation (length, truncation flags, helper paths including `UpdateToolTrace`)
- [ ] Existing truncation flag behavior unchanged (`RequestBodyTruncated`, `PromptTruncated`, step `Truncated`, etc.)
- [ ] No behavior change for payloads already under the limit
- [ ] Focused `routerreplay` package tests / `make test-semantic-router` pass

## Suggested test sketch

1. Build a large string for `RequestBody` / `Prompt` / step fields.
2. Persist via `AddRecord` / `UpdateToolTrace` with a small max.
3. Confirm stored length equals max and truncation flags are set.
4. Ensure truncate helpers copy rather than share backing storage (regression against plain `s[:max]`).

## Related code

- `src/semantic-router/pkg/routerreplay/recorder.go` (`AddRecord`, `applyMaxToolTraceBytes`, `truncateToolTraceStep`, `truncateBody`)
- Capture policy / max body config via `global.services.router_replay` and route-local `router_replay` plugin

## Labels (suggested)

`bug`, `router`, `memory`, `performance`

gagandhakrey:
closes: #2780

## Overview
This PR addresses a severe memory leak and potential OOM (Out of Memory) vulnerability caused by Go's underlying string slicing behavior when processing large request and response payloads. 

## The Problem
Previously, large payload strings (such as `RequestBody`, `ResponseBody`, `Prompt`, and `ToolTrace` fields) were being truncated using standard Go string slicing (e.g., `rec.RequestBody[:policy.maxBodyBytes]`). 

Because Go strings share their underlying backing array when sliced, truncating a large 50MB payload down to a 4KB snippet inadvertently retained the entire 50MB array in memory. This memory was held for as long as the truncated record sat in the Recorder's queue or memory store. Under heavy traffic, this prevented the garbage collector from reclaiming memory, causing a massive memory leak and eventual OOM crashes.

## The Solution
Wrapped all relevant string truncations with `strings.Clone()`. 

This forces Go to allocate a fresh, minimal byte array that is exactly the size of the truncated string, safely severing the reference to the original massive payload. 

**Affected fields updated:**
* `rec.RequestBody`
* `rec.ResponseBody`
* `rec.Prompt`
* `rec.ToolDefinitions`
* `step.Arguments`
* `step.Output`

## Impact
* **Immediate GC Reclamation:** The garbage collector can now immediately reclaim the memory used by the original large payloads as soon as the caller is done with them.
* **Router Stability:** Significantly reduces and stabilizes the memory footprint of the router, entirely eliminating the OOM vulnerability during payload capture and truncation.

### DIFF
--- a/src/semantic-router/pkg/routerreplay/recorder.go
+++ b/src/semantic-router/pkg/routerreplay/recorder.go
@@ -2,6 +2,7 @@ package routerreplay
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
@@ -147,7 +148,7 @@ func (r *Recorder) AddRecord(rec RoutingRecord) (string, error) {
 	// dropped here — truncation alone only bounds a leak (see #2748).
 	if policy.captureRequestBody {
 		if len(rec.RequestBody) > policy.maxBodyBytes {
-			rec.RequestBody = rec.RequestBody[:policy.maxBodyBytes]
+			rec.RequestBody = strings.Clone(rec.RequestBody[:policy.maxBodyBytes])
 			rec.RequestBodyTruncated = true
 		}
 	} else {
@@ -157,7 +158,7 @@ func (r *Recorder) AddRecord(rec RoutingRecord) (string, error) {
 
 	if policy.captureResponseBody {
 		if len(rec.ResponseBody) > policy.maxBodyBytes {
-			rec.ResponseBody = rec.ResponseBody[:policy.maxBodyBytes]
+			rec.ResponseBody = strings.Clone(rec.ResponseBody[:policy.maxBodyBytes])
 			rec.ResponseBodyTruncated = true
 		}
 	} else {
@@ -185,11 +186,11 @@ func applyMaxToolTraceBytes(rec *RoutingRecord, max int) {
 		return
 	}
 	if len(rec.Prompt) > max {
-		rec.Prompt = rec.Prompt[:max]
+		rec.Prompt = strings.Clone(rec.Prompt[:max])
 		rec.PromptTruncated = true
 	}
 	if len(rec.ToolDefinitions) > max {
-		rec.ToolDefinitions = rec.ToolDefinitions[:max]
+		rec.ToolDefinitions = strings.Clone(rec.ToolDefinitions[:max])
 		rec.ToolDefinitionsTruncated = true
 	}
 	truncateToolTraceSteps(rec.ToolTrace, max)
@@ -237,11 +238,11 @@ func capToolTraceStepCount(trace *ToolTrace, max int) {
 // and Truncated is set to signal that truncation happened.
 func truncateToolTraceStep(step *ToolTraceStep, max int) {
 	if len(step.Arguments) > max {
-		step.Arguments = step.Arguments[:max]
+		step.Arguments = strings.Clone(step.Arguments[:max])
 		step.Truncated = true
 	}
 	if len(step.Output) > max {
-		step.Output = step.Output[:max]
+		step.Output = strings.Clone(step.Output[:max])
 		step.Truncated = true
 	}
 }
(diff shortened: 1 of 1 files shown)


src/semantic-router/pkg/routerreplay/recorder.go
```diff
@@ -157,7 +158,7 @@ func (r *Recorder) AddRecord(rec RoutingRecord) (string, error) {
 
 	if policy.captureResponseBody {
 		if len(rec.ResponseBody) > policy.maxBodyBytes {
-			rec.ResponseBody = rec.ResponseBody[:policy.maxBodyBytes]
+			rec.ResponseBody = strings.Clone(rec.ResponseBody[:policy.maxBodyBytes])
 			rec.ResponseBodyTruncated = true
 		}
 	} else {
```
AayushSaini101: thanks @gagandhakrey can you optimised these statements as well there are duplicates lines in the flow thanks : )

gagandhakrey: thanks Aayush for pointing out , done(refactored the code)

AayushSaini101 (dismissed):
LGTM thanks : )

AayushSaini101:
@gagandhakrey seems there are some conflicts are present can you fix them thanks

AayushSaini101 (dismissed):

AayushSaini101:
thanks @gagandhakrey pre-commit check is failing seems due to some formatting issue can you check thanks : )

gagandhakrey:
got it , fixed it will ping you once it ci is green

AayushSaini101 (approved):

PR #2779, opened 2026-08-05
[Docs] propose routing inference runtime and model ecosystem

Xunzhuo:
## Summary
- propose a router-owned inference runtime with typed task/codec contracts, immutable deployments, driver negotiation, session lifecycle, and receipt-backed support levels
- define a Day-0 vendor onboarding pipeline and an initial 36-deployment L2 verification target across local and remote execution drivers
- add the proposal to the documentation sidebar and quote Mermaid node labels for consistent rendering

## Test plan
- [x] Parse the routing-runtime diagram with Mermaid 11.14.0
- [x] `cd website && npx docusaurus build --locale en`

Part of #2782

### DIFF
--- a/website/docs/proposals/routing-inference-runtime-and-model-ecosystem.md
+++ b/website/docs/proposals/routing-inference-runtime-and-model-ecosystem.md
@@ -1,0 +1,1081 @@
+# Routing Inference Runtime and Model Portfolio
+
+**Status:** Proposal<br />
+**Target:** vLLM Semantic Router model-runtime roadmap
+
+## Summary and decisions
+
+Routing-time inference already includes sequence and token classifiers,
+embeddings, rerankers, and small decoders used by signals, algorithms, and
+plugins. This RFC defines common model identity, task contracts, deployment,
+placement, and support evidence beneath the existing signal → projection →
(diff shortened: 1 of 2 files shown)


adaamko:
Read through the proposal, I like the task/codec split plus receipts direction.

There is one gap I think, mainly on the safety side. The task table assigns jailbreak to sequence_classify (line 373), while token_classify right below it lists "policy matching" as a consumer (line 376). For jailbreak and harm the more useful output is which span is the attack, rather than one label for the whole prompt, the same shape PII already consumes. Right now the guardrail models in the list are either document-level or free-form-rule (GliGuard at L1 through GLiNER2, Qwen3Guard-Stream via stream_guard), so a trained span-level safety model has no defined home. Could safety signals be explicitly allowed to consume token_classify / token_spans.v1? @Xunzhuo

FAUST-BENCHOU (commented):
conflicts

Xunzhuo:
Closing this proposal as superseded by the implemented runtime and current documentation, including #3769. Remaining qualification, compatibility receipts and model-family work stays tracked in #2782, #3196 and #2974.

PR #2781, opened 2026-08-05
[CI/Build] isolate management API exposure by default (#2463 Phase 4)

AayushSaini101:
Related #2463 (Phase 4: deployment exposure isolation).

Stop advertising the router management API (`:8080` / `classify-api`) by default on local host publish and primary Kubernetes Services, while keeping in-cluster / docker-network access for dashboard and tests.

### What changed

**Local CLI (`vllm-sr serve`)**
- Host port map for `:8080` is **off by default**
- Opt-in: `--publish-management-api` or `VLLM_SR_PUBLISH_MANAGEMENT_API=true`
- Dashboard/envoy still reach `router:8080` on the docker network

**Helm**
- Primary Service: gRPC only by default
- Always-on `*-management` ClusterIP Service for in-cluster access
- Dashboard `TARGET_ROUTER_API_URL` points at `*-management`
- `managementApi.expose: true` also advertises `classify-api` on the primary Service
- NOTES updated for port-forward to the management Service

**Static deploy (KServe / Istio)**
- Remove `classify-api` from the main Service by default
- Opt-in manifests: `service-management.yaml` (+ KServe `route-management.yaml`)

**E2E / scripts**
- Memory integration readiness uses `docker exec … /ready` (no host `:8080`)
- Router API sessions port-forward `semantic-router-management`

### Operator usage

```bash
# default — management API not on localhost
vllm-sr serve

# host tools (curl / eval / rag against :8080)
vllm-sr serve --publish-management-api

# Helm: also put classify-api on the primary Service
helm upgrade ... --set managementApi.expose=true
```

## Test plan

- [x] `make vllm-sr-test`
- [x] `make helm-lint`
- [x] CLI unit coverage: default omits `*:8080` publish; opt-in maps host api port
- [ ] Manual: `vllm-sr serve` then confirm `localhost:8080` is unreachable; with `--publish-management-api` it works
- [ ] Manual/Helm: dashboard still reaches management via `*-management` Service
- [ ] E2E profiles that use `OpenRouterAPISession` still pass against `semantic-router-management`


AayushSaini101:
@Xunzhuo @drivebyer @FAUST-BENCHOU Need your review thanks : )

AayushSaini101:
@drivebyer @drivebyer @FAUST-BENCHOU @Xunzhuo can you please review it thanks

drivebyer (dismissed):
---

drivebyer (commented):
Nice catch on the LAN exposure. Would a loopback publish work here? -p 127.0.0.1:8080:8080 in [append_port_mappings](https://github.com/vllm-project/semantic-router/blob/main/src/vllm-sr/cli/container_run_command.py#L93-L95) blocks LAN access the same way, without a new flag. The dashboard isn't affected either way since it reaches router:8080 over the docker network, and it keeps vllm-sr eval / rag and the curl localhost:8080 examples in the docs working.

WUKUNTAI-0211 (changes requested):
Thanks — the Helm/KServe split is solid. I checked PR #2781 (`18890fa2`) merged onto current main (`e7cbab82`): rendered Helm default / `managementApi.expose=true`, rendered the Istio kustomization, ran `pytest tests/test_split_runtime_stack.py -q` (**13 passed**), and exercised the falsey publish-env path.

## What's good

- Default Helm render: primary Service is gRPC-only; unconditional `semantic-router-management` carries `classify-api`. With `managementApi.expose=true`, the primary also carries `classify-api`.
- Primary and management Service selectors are identical; the E2E fixture port-forward rename is correct for the standard Helm release.
- The memory test readiness change to `docker exec ... /ready` correctly removes its host-`:8080` dependency.
- API exposure is fail-closed: falsey/unknown env values do not publish `:8080`.

## Findings

### Important — blocking

1. **Default `vllm-sr serve` now breaks documented `eval`/`rag` usage.**

   The router no longer publishes host `:8080` by default, but `eval.py` and `rag.py` still default to `localhost:8080`; both are direct HTTP clients. `src/vllm-sr/README.md` still documents `vllm-sr serve` followed by eval/rag, and their connection errors tell users to start serve again.

   `serve --help` mentions `--publish-management-api`, but the README/errors must match. Please either preserve loopback-only `127.0.0.1:8080` publishing, or update the workflows and errors to require `vllm-sr serve --publish-management-api`.

2. **The observability dashboard targets a removed Service port.**

   `deploy/kubernetes/observability/dashboard/configmap.yaml:12` still points to:

   ```yaml
   http://semantic-router.vllm-semantic-router-system.svc.cluster.local:8080
   ```

   The default Helm primary Service no longer exposes `:8080`. This exact ConfigMap is consumed by the dashboard E2E deployment, so dashboard API/config calls fail. Point it at `semantic-router-management...:8080`, matching the chart dashboard change.

### Important

3. **Istio does not install the new management Service.**

   `service-management.yaml` is new but absent from `deploy/kubernetes/istio/kustomization.yaml`. `kubectl kustomize` produces no `:8080` Service. Add it to `resources` or document an explicit apply step.

4. **`VLLM_SR_PUBLISH_MANAGEMENT_API=false` logs “ENABLED”.**

   Runtime behavior is safe—`:8080` stays unpublished—but `runtime_support.py` treats any non-empty value as enabled while `container_start.py` parses actual truthy values. Use one low-level boolean parser for both decisions and logging.

## Follow-up

`deploy/openshift/` legacy manifests still expose `:8080`; they are self-consistent, but the Phase 4 policy is not yet uniform. Please track that explicitly.

## Verdict

**Request changes** — findings 1 and 2 break documented/deployed behavior.

Xunzhuo (changes requested):
Thanks for tightening the default management-API exposure. In addition to the existing review findings, I found these unresolved cross-file and upgrade regressions:

1. **Helm ingress breaks with the new default.** `deploy/helm/semantic-router/templates/ingress.yaml:37` still sends traffic to the primary Service, and the default ingress backend port remains `8080`. That port is now absent unless `managementApi.expose=true`, so existing ingress-enabled releases will start returning backend errors after upgrade. Please route the ingress to the dedicated management Service, or validate that ingress requires explicit management exposure.

2. **The documented Helm port-forward helper now fails.** `tools/make/helm.mk:371` still runs `kubectl port-forward ... svc/$(HELM_RELEASE_NAME) 8080:8080`, but the primary Service no longer has port 8080. `make helm-port-forward-api` is referenced by the Helm quick-start docs. It should forward `svc/$(HELM_RELEASE_NAME)-management`.

3. **KServe upgrades leave the old API Route behind.** Removing the second object from `deploy/kserve/route.yaml` does not delete a Route previously created by `oc apply`. On upgrade, `semantic-router-kserve-api` remains, targets the primary Service whose `classify-api` port was removed, and `deploy.sh` later reports that stale route as available. Please add migration cleanup for the old default Route while preserving an explicitly opted-in management Route.

4. **The new exposure toggle is not type-checked.** `managementApi` is missing from `deploy/helm/semantic-router/values.schema.json`. Helm treats a non-empty string such as `managementApi.expose: \"false\"` or `--set-string managementApi.expose=false` as truthy, which unintentionally exposes `classify-api`. Please define `managementApi.expose` as a boolean in the schema and use a strict boolean check.

5. **`--publish-management-api` silently does nothing for Kubernetes targets.** The option is accepted and logged before target selection, but `--target k8s` exposure is controlled by the Helm value rather than the pod environment variable. Please either map the option to `managementApi.expose` for Kubernetes or reject/document it as local-container-only.

Requesting changes because findings 1–3 break supported deployment and upgrade workflows.

FAUST-BENCHOU (changes requested):
pls resolve conflicts

PR #2784, opened 2026-08-05
fix(cli): forward the api keys a config actually names

### LINKED ISSUE #2775: bug: missing api keys in launcher's allowlist
### Describe the bug

The launcher's allowlist contains only:
```
("HF_TOKEN", True),
("ANTHROPIC_API_KEY", True),
("OPENAI_API_KEY", True),
```

### To Reproduce

runtime_support.py:
```
PASSTHROUGH_ENV_RULES = (
    ("HF_ENDPOINT", False),
    ("HF_TOKEN", True),
    ("HF_HOME", False),
    ("HF_HUB_CACHE", False),
    ("VLLM_SR_DETERMINISTIC_EMBEDDINGS", False),
    ("ANTHROPIC_API_KEY", True),
    ("OPENAI_API_KEY", True),
    ("OPENCLAW_BASE_IMAGE", False),
    ("SR_LOG_LEVEL", False),
    ("SR_LOG_ENCODING", False),
    ("SR_LOG_DEVELOPMENT", False),
    ("SR_LOG_ADD_CALLER", False),
)
```

### Expected behavior

Parsing all api keys, or way to add them into container at runtime

### Affected layer

global level

### Additional context

_No response_

ankit373:
Closes #2775

`PASSTHROUGH_ENV_RULES` is a static allowlist, but `api_key_env` is free-form
(`canonical_providers.go:44`, read via `os.Getenv(ref.APIKeyEnv)` in
`canonical_global.go:354`). Any provider credential whose name is not one of those
literals never reaches the container, and the failure is silent: the key resolves to an
empty string inside the router.

The shipped reference config hits this today:

```
$ python -c "from cli.commands.runtime_support import PASSTHROUGH_ENV_RULES, config_env_references; \
  print(sorted(config_env_references('config/config.yaml') - {n for n,_ in PASSTHROUGH_ENV_RULES}))"
['EMBEDDING_API_KEY', 'VLLM_SR_PRIMARY_API_KEY']
```

`config/config.yaml:40` names `VLLM_SR_PRIMARY_API_KEY` for the primary provider and
`:2124` names `EMBEDDING_API_KEY`. Neither is forwarded.

## Change

Read the env var names back off the config rather than guessing them ahead of time:

- `config_env_references()` collects `api_key_env` values and `${VAR}` interpolations,
  including the `${VAR:-default}` and `${VAR-default}` forms `resolveBracedEnvReference`
  supports.
- `append_passthrough_env_vars(env_vars, config_path)` forwards those in addition to the
  static rules, masked in logs, and skips names a static rule already handled.

The interpolation scan is restricted to upper case on purpose. `config/config.yaml:1163`
uses `${user_content}` and `${top_k}` as MCP tool-argument placeholders, and those are not
environment variables.

## Second half: the same list decided what was secret

`_sync_env_secret` put only sensitive names into the K8s Secret and `_translate_env_vars`
wrote everything else into the Helm `env:` list. Both derived "sensitive" from that same
allowlist, so an unrecognized provider key would not merely have been missing, it would
have been **written into values.yaml in plain text** once it reached `env_vars` by any
route.

`sensitive_env_names(config_path)` now treats anything the config names as a credential,
and both call sites use it. Both already had the config path in scope, so this is one
argument each.

## Testing

`src/vllm-sr` suite: 550 passed, up from 546. `black` clean across all 112 files.

Four new tests, each verified to fail without the change by disabling the discovery and
re-running:

| test | fails without the fix on |
|---|---|
| `test_append_passthrough_env_vars_forwards_keys_named_by_the_config` | `GEMINI_API_KEY` and `VALKEY_PASSWORD` absent from `env_vars` |
| `test_config_env_references_reads_api_key_env_and_interpolations` | empty set returned |
| `test_sensitive_env_names_covers_config_named_credentials` | `GEMINI_API_KEY` not treated as sensitive |
| `test_config_named_api_key_excluded_from_plain_env` | key present in the plain Helm `env:` list |

The last one is the plain-text case, so it fails loudly rather than leaking quietly.

Unreferenced names are not forwarded, a missing or unparseable config yields an empty set
rather than an error, and `${VAR}` inside a YAML comment is ignored because the scan runs
over the parsed document.

### DIFF
--- a/src/vllm-sr/cli/commands/runtime.py
+++ b/src/vllm-sr/cli/commands/runtime.py
@@ -103,7 +103,7 @@ def _execute_serve(
     validate_setup_mode_flags(setup_mode, minimal, readonly)
 
     env_vars: dict[str, str] = {}
-    append_passthrough_env_vars(env_vars)
+    append_passthrough_env_vars(env_vars, config_path)
     apply_runtime_mode_env_vars(
         env_vars,
         minimal,
--- a/src/vllm-sr/cli/commands/runtime_support.py
+++ b/src/vllm-sr/cli/commands/runtime_support.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import os
+import re
+from itertools import chain
 from pathlib import Path
 
 import yaml
@@ -42,6 +44,12 @@ RUNTIME_CONFIG_PATH_ENV = "VLLM_SR_RUNTIME_CONFIG_PATH"
 SOURCE_CONFIG_PATH_ENV = "VLLM_SR_SOURCE_CONFIG_PATH"
 RUNTIME_ALGORITHM_OVERRIDE_ENV = "VLLM_SR_ALGORITHM_OVERRIDE"
 
+# Mirrors resolveBracedEnvReference, including ${NAME:-default}. Upper case only, so
+# MCP tool-argument placeholders like ${user_content} are not read as env vars.
+_ENV_REFERENCE = re.compile(r"\$\{([A-Z_][A-Z0-9_]*)(?::?-[^}]*)?\}")
+
+API_KEY_ENV_FIELD = "api_key_env"
+
 PASSTHROUGH_ENV_RULES = (
     ("HF_ENDPOINT", False),
     ("HF_TOKEN", True),
@@ -57,6 +65,8 @@ PASSTHROUGH_ENV_RULES = (
     ("SR_LOG_DEVELOPMENT", False),
     ("SR_LOG_ADD_CALLER", False),
 )
+
+_STATIC_SENSITIVE = frozenset(name for name, masked in PASSTHROUGH_ENV_RULES if masked)
 
 
 def _finalize_runtime_config_write(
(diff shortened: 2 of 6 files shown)


drivebyer (commented):
Thanks for the fix — the discovery approach makes sense. Two things I think need addressing:

1. Env names are discovered from the **source** config, but `_sync_env_secret` / `_translate_env_vars` re-derive sensitivity from the **effective** config (which `resolve_effective_config_path` may rewrite). If the two diverge, a forwarded key could be treated as non-sensitive and end up in plaintext Helm values — the exact leak this PR closes. Deriving both from the same file would fix it.

2. Discovered names have no exclusion list, so a config referencing e.g. `${DISABLE_DASHBOARD}` or `${HOME}` would forward host values and override CLI-managed vars. A small denylist (CLI control vars + `PATH`/`HOME`/`USER`) seems worth adding.

ankit373:
Both real, fixed in bfaf479.

1. `deploy()` was checking sensitivity against `config_file` (the effective config), not the config `append_passthrough_env_vars` actually discovered names from. Threaded `source_config_file` through `K8sBackend.deploy` and `translate_config_to_helm_values` so both `_sync_env_secret` and the plain-env filter check the same file discovery used. Added a test that reproduces the leak: a name present only in the source config now stays masked even when the effective config no longer has it.

2. Added `_DISCOVERY_DENYLIST` in `runtime_support.py`: `PATH`, `HOME`, `USER`, `SHELL`, `PWD`, `LOGNAME`, plus the CLI-owned override vars (`DISABLE_DASHBOARD`, `DASHBOARD_READONLY`, `DASHBOARD_PLATFORM`, `VLLM_SR_PLATFORM`, setup-mode/algorithm-override env names). `config_env_references` subtracts it before returning, so it protects discovery and the sensitivity check in one place.

56 tests in the two affected files, 555 in the full suite, all green. Reverted each fix independently and confirmed the new tests fail on the specific assertion before restoring.

drivebyer (approved):
LGTM

Issue #2715, opened 2026-07-28
[Bug] Inverted cache-hit criterion: IP vector distance is treated as a similarity


drivebyer:
### Describe the bug

With `metric_type: IP`, every Valkey/Redis-backed lookup applies an inverted hit criterion: identical queries never match, unrelated ones do.

RediSearch and valkey-search return `vector_distance` as a **distance** for all three metrics — `IP: d = 1 - u·v`, smaller means closer ([Redis](https://redis.io/docs/latest/develop/ai/search-and-query/vectors/), [Valkey](https://valkey.io/commands/ft.create/)). The conversion helper passes that value straight through for `IP`:

```go
case "IP":
    return distance   // should be 1 - distance
```

Callers then compare it against a larger-is-more-similar threshold, so the comparison runs backwards.

The same code exists in four places, three of them affected:

| Location | Consumer | Comparison |
|---|---|---|
| `pkg/cache/redis_cache.go:598` | semantic cache (Redis) | `redis_cache.go:700` |
| `pkg/cache/valkey_cache_helpers.go:177` | semantic cache (Valkey) | `valkey_cache.go:550` |
| `pkg/memory/valkey_store_helpers.go:391` | agentic memory retrieval | `valkey_store.go:292` |
| `pkg/vectorstore/valkey_backend_parse.go:238` | vector store / RAG retrieval | `hybrid.go:523`, `memory_backend.go:300` |

### To Reproduce

1. Configure any Valkey- or Redis-backed component with `metric_type: IP` (the shipped Milvus example already uses `IP`, and the operator CRD defaults to it).
2. Store an entry, then query with the byte-identical text.
3. The lookup misses. Querying with unrelated text hits instead.

The arithmetic: identical normalized vectors give `u·v = 1`, so `d = 0`, so the passthrough reports similarity `0.0` — below every threshold. Orthogonal vectors give `d = 1` and report `1.0`.

### Expected behavior

Repeating a stored query verbatim returns the cached entry. `1 - distance` recovers the inner product, which for the L2-normalized embeddings this repo produces equals the cosine similarity.

### Additional context

Three existing tests assert the wrong value and would keep the bug alive:

- `pkg/cache/valkey_cache_test.go:126` — `{"IP passthrough", "IP", 0.75, 0.75}`
- `pkg/memory/valkey_store_test.go:196` — `{"IP identity", "IP", 0.95, 0.95, 0.001}`
[...]

PR #2786, opened 2026-08-05
[Router] fix: convert Milvus L2 scores to similarities before threshold comparison

drivebyer:
## What

Under `metric_type: L2`, Milvus returns the squared Euclidean distance (smaller = closer), but semantic-cache lookup (`FindSimilarWithThreshold`) and RAG retrieval (`SearchDocuments`) compared the raw score directly against a similarity threshold. The comparison is inverted: an exact match scores 0.0 and is reported as a MISS, while an unrelated entry with a large distance clears the threshold and is served as a bogus HIT.

Same defect shape as the Valkey/Redis fix in #2716 — this is the Milvus sibling.

## Changes

- Convert scores per metric before comparing: L2 → `1/(1+d)` (same compression as #2716), IP/COSINE pass through.
- Normalize `metric_type` at construction (empty → `IP`, uppercased) so the inline-config path can't reintroduce the bug via a lowercase `l2`; warn on unrecognized values.
- Surface SDK per-result errors (`SearchResult.Err`) instead of misreporting them as threshold misses.
- Log the metric type alongside similarities in cache hit/miss events.

Note: L2 thresholds now compare against `1/(1+d)`, so existing L2 deployments keep a correct orientation with a slightly stricter effective cutoff (0.8 ≈ cos ≥ 0.875 for normalized vectors).

## Testing

Unit tests cover the conversion table, the #2716 regression shape (exact match hits, distant entry misses), metric normalization, and fake-client wiring for `SearchDocuments` (conversion before threshold, caller metric canonicalization, error propagation). `go build` / `go vet` / `gofmt` clean; `go test ./pkg/cache/` matches the pre-change baseline (remaining failures need a local Milvus/Valkey).

Related #2715 (the Milvus sibling of the Valkey/Redis fix in #2716)


src/semantic-router/pkg/cache/milvus_cache_admin.go
```diff
@@ -105,24 +108,28 @@ func (c *MilvusCache) SearchDocuments(ctx context.Context, collectionName string
 	if len(searchResult) == 0 || searchResult[0].ResultCount == 0 {
 		return nil, nil, nil // No results, but not an error
 	}
```
Xunzhuo, line 114: Could we move this error check before the `ResultCount == 0` return? `SearchResult.Err` can coexist with a zero result count, so the current ordering still reports that failure as a normal no-result response instead of surfacing it. A regression test with `ResultCount: 0` and a non-nil `Err` would cover the case.

drivebyer: Nice catch, fixed in next commit.

AayushSaini101 (approved):

PR #2815, opened 2026-08-07
[Performance] Replace O(N) Redis scan with conversation index lookup

gagandhakrey:
# perf(responsestore): replace O(N) Redis scan with secondary index in `ListResponsesByConversation`

## Issue

`ListResponsesByConversation` performed a full keyspace `SCAN` across **all** response keys in Redis, followed by an individual `GET` for every key before filtering by `conversationID` in memory.

This resulted in an **O(N)** lookup where every request scanned the entire response dataset regardless of how many responses belonged to the requested conversation.

As the dataset grew, this approach led to:

- **Excessive Redis operations** — one `SCAN` and one `GET` per stored response.
- **High latency** — response time increased linearly with the total number of responses.
- **Poor scalability** — lookup cost depended on the total dataset size instead of the requested conversation.
- **Unnecessary CPU and memory overhead** — every response was fetched and unmarshalled before most were discarded.

Additionally, `DeleteConversation(deleteResponses=true)` internally called `ListResponsesByConversation`, which respected `DefaultListLimit=20`. As a result, conversations containing more than 20 responses could leave orphaned response keys in Redis.

---

## Fix

Introduced a Redis **SET-based secondary index**:

```
sr:index:conversation_responses:<conversationID>
```

Each conversation now maintains a set of response IDs, allowing lookups to operate on only the relevant responses.

| Method | Change |
|--------|--------|
| **`StoreResponse`** | Adds the response ID to the conversation index (`SADD`) and refreshes the index TTL. |
| **`UpdateResponse`** | If `ConversationID` changes, removes the response ID from the old index (`SREM`) and adds it to the new one (`SADD`). When unchanged, performs `SADD` to keep the index self-healing and refreshes the TTL. |
| **`DeleteResponse`** | Retrieves the response to determine its conversation, deletes the response, and removes its ID from the conversation index. |
| **`ListResponsesByConversation`** | Replaces full keyspace `SCAN` with `SMEMBERS`, followed by a pipelined fetch of only the matching response keys. |
| **`DeleteConversation`** | Uses the conversation index directly instead of `ListResponsesByConversation`, ensuring every response is deleted regardless of list limits, then removes the index key. |
| **`AddResponseToConversation`** | Previously a no-op for Redis. Now updates the conversation index to match the behavior of the in-memory store. |

---

## Post-Fix Impact

| Metric | Before | After |
|--------|--------|-------|
| **Time Complexity** | O(N) where N = total responses | O(K) where K = responses in the target conversation |
| **Redis Operations** | 1 `SCAN` + N `GET`s | 1 `SMEMBERS` + pipelined fetch |
| **Network Round Trips** | Linear with total responses | Constant index lookup + batched fetch |
| **CPU & Memory Usage** | Deserialize every stored response | Deserialize only matching responses |
| **Scalability** | Depends on total dataset size | Depends only on conversation size |
| **DeleteConversation** | Could leave orphaned responses beyond the default limit | Deletes all responses and removes the conversation index |
| **Index Consistency** | N/A | Automatically maintained during create, update, delete, and conversation migration |

---

## Test Cases

All integration tests reside in:

```
pkg/responsestore/redis_store_integration_test.go
```

and run against a live Redis instance (tests are skipped when Redis is unavailable).

### Existing Tests

| Test Case | Validation |
|-----------|------------|
| **List responses by conversation** | Stores multiple responses in the same conversation and verifies that `ListResponsesByConversation` returns the expected results using the new index path. |
| **List responses with limit** | Ensures `ListOptions.Limit` continues to work correctly on top of the index-based lookup. |

### New Tests

| Test Case | Validation |
|-----------|------------|
| **Index updated on delete** | Deletes a response and verifies it no longer appears in conversation listings, confirming the response ID is removed from the conversation index. |
| **Index updated when conversation changes** | Updates a response's `ConversationID` and verifies it disappears from the old conversation and appears in the new one. |
| **AddResponseToConversation updates index** | Verifies the Redis implementation now correctly maintains the conversation index instead of acting as a no-op. |
| **DeleteConversation deletes all responses** | Verifies conversations containing more than the default list limit have all associated responses deleted and the conversation index removed. |

### Test Execution

```bash
cd src/semantic-router
go test ./pkg/responsestore -v -count=1
```

All integration tests pass successfully.

---

## Files Changed

- `pkg/responsestore/redis_store.go`
- `pkg/responsestore/redis_store_integration_test.go`
```

### DIFF
--- a/src/semantic-router/pkg/responsestore/redis_store.go
+++ b/src/semantic-router/pkg/responsestore/redis_store.go
@@ -429,10 +429,12 @@ func (s *RedisStore) UpdateResponse(ctx context.Context, response *responseapi.S
 			}
 		}
 	} else if response.ConversationID != "" {
-		// ConversationID did not change, but we should refresh the TTL of the index
-		// since the response itself just had its TTL refreshed.
+		// ConversationID did not change, but we should ensure the response is in the index
+		// and refresh the TTL since the response itself just had its TTL refreshed.
 		idxKey := s.buildKey(ConversationIndexPrefix + response.ConversationID)
-		if s.ttl > 0 {
+		if err := s.client.SAdd(ctx, idxKey, response.ID).Err(); err != nil {
+			logging.Warnf("RedisStore: failed to refresh response %s in conversation index %s: %v", response.ID, response.ConversationID, err)
+		} else if s.ttl > 0 {
 			s.client.Expire(ctx, idxKey, s.ttl)
 		}
 	}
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Closing this as a duplicate of #2817, the later and broader implementation of the same Redis conversation index path. Please carry any unique deletion and integration coverage from this draft into #2817 before its review.

PR #2817, opened 2026-08-08
[Performance] Replace O(N) Redis scan with a secondary index

gagandhakrey:
Description  Polished with AI

<html><head></head><body><h1>perf: optimize ListResponsesByConversation with a secondary Redis index</h1><h2>Summary</h2><p>Replaced the full Redis keyspace scan in <code inline="">ListResponsesByConversation</code> with a secondary sorted-set index keyed by conversation ID.</p><p>The read path now scales with the number of responses in the requested conversation instead of the total number of responses stored in Redis.</p><h2>Issue</h2><p><code inline="">RedisStore.ListResponsesByConversation</code> previously:</p><ol><li><p>Scanned the entire response keyspace.</p></li><li><p>Issued an individual <code inline="">GET</code> for every response key.</p></li><li><p>Unmarshalled every response.</p></li><li><p>Filtered by <code inline="">conversationID</code> in memory.</p></li></ol><p>This made the operation <strong>O(N)</strong> in the total number of stored responses rather than the number of responses belonging to the requested conversation.</p><p>As the Redis dataset grew, latency increased linearly and most of the fetched payloads were immediately discarded.</p><h2>Fix</h2><p>Added a secondary Redis index:</p><pre><code class="language-text">sr:conversation-index:{conversation_id}
</code></pre><p>The index is a sorted set containing response IDs scored by <code inline="">created_at</code>.</p><h3>Read path</h3><p>The new flow is:</p><pre><code class="language-text">ZRANGE conversation-index
        ↓
response IDs for requested conversation
        ↓
pipelined GET
        ↓
unmarshal only requested responses
</code></pre><p>This reduces the read path to:</p><ul><li><p>1 <code inline="">ZRANGE</code></p></li><li><p>1 pipelined batch of <code inline="">GET</code>s</p></li></ul><p>instead of scanning the entire response keyspace.</p><h3>Write path changes</h3><ul><li><p><strong><code inline="">StoreResponse</code></strong></p><ul><li><p>Replaced <code inline="">EXISTS</code> + <code inline="">SET</code> with atomic <code inline="">SET NX EX</code>.</p></li><li><p>Preserves the existing <code inline="">ErrAlreadyExists</code> contract.</p></li><li><p>Eliminates the pre-existing TOCTOU race.</p></li><li><p>Writes the payload before adding the index entry, making prune-on-missing safe even when Redis Cluster places the keys on different nodes.</p></li></ul></li><li><p><strong><code inline="">UpdateResponse</code> / <code inline="">DeleteResponse</code></strong></p><ul><li><p>Read the stored conversation ID to move or remove the corresponding index entry.</p></li><li><p><code inline="">UpdateResponse</code> no longer requires a separate <code inline="">EXISTS</code> probe.</p></li></ul></li><li><p><strong>Expiry</strong></p><ul><li><p>Response payloads already expire through their TTL.</p></li><li><p>Sorted-set members do not expire individually, so the index key TTL is refreshed whenever an entry is added.</p></li><li><p>TTL refresh is guarded by <code inline="">s.ttl &gt; 0</code> because <code inline="">EXPIRE key 0</code> deletes the key.</p></li><li><p>Listings prune index entries whose response payload no longer exists.</p></li></ul></li><li><p><strong>Index write failures</strong></p><ul><li><p>Logged but not returned to the caller.</p></li><li><p>The response payload is already durable, and returning the index error would cause callers to retry and potentially receive <code inline="">ErrAlreadyExists</code>.</p></li></ul></li><li><p><strong>Conversation ID validation</strong></p><ul><li><p>Retained the existing <code inline="">ConversationID</code> equality check as a guard against index skew.</p></li></ul></li><li><p><strong>Redis Cluster</strong></p><ul><li><p>No multi-key commands span Redis Cluster slots.</p></li><li><p>The only same-key pipeline pair, <code inline="">ZADD</code> + <code inline="">EXPIRE</code>, operates on the index key.</p></li></ul></li></ul><p>The index prefix intentionally does not start with <code inline="">conversation:</code> so these keys are excluded from the <code inline="">sr:conversation:*</code> scan used by <code inline="">ListConversations</code>.</p><h2>Related Bugs Fixed</h2><h3>1. <code inline="">DeleteConversation</code> cascade limit</h3><p><code inline="">DeleteConversation(id, deleteResponses=true)</code> previously called <code inline="">ListResponsesByConversation</code> with the default <code inline="">DefaultListLimit</code> of <strong>20</strong>.</p><p>As a result, conversations with more than 20 responses silently deleted only the first 20 responses.</p><p>The cascade now reads directly from the conversation index and deletes all indexed responses along with the index key.</p><h3>2. Response ordering</h3><p><code inline="">SCAN</code> does not guarantee ordering, despite <code inline="">ResponseStore</code> documenting chronological ordering.</p><p>The new index uses <code inline="">created_at</code> as the sorted-set score, so responses are now returned chronologically, matching <code inline="">MemoryStore</code>.</p><h2>Performance Impact</h2><p>Measured against local Redis with:</p><ul><li><p>5,000 responses</p></li><li><p>500 conversations</p></li><li><p>5,500 total Redis keys</p></li><li><p>10 responses in the requested conversation</p></li></ul>
Metric | Before | After
-- | -- | --
Cost scales with | Total responses in Redis | Responses in requested conversation
Redis operations | ~550 SCAN + 5,000 sequential GET | 1 ZRANGE + 10 pipelined GET
Commands | ~5,550 | 1 ZRANGE + 10 pipelined GET
Payloads unmarshalled | 5,000 | 10
Discarded payloads | 4,990 | 0
Latency | Grows linearly with dataset size | ~263µs

<h2>Behavioral Changes</h2><h3>Legacy data</h3><p>Responses written before the secondary index was introduced do not have index entries.</p><p>Therefore:</p><ul><li><p>They are not returned by <code inline="">ListResponsesByConversation</code>.</p></li><li><p>They are not removed by the new <code inline="">DeleteConversation</code> cascade.</p></li><li><p>They remain retrievable by ID.</p></li><li><p>They remain available through <code inline="">GetConversationChain</code>.</p></li><li><p>They expire naturally through their existing TTL, 30 days by default.</p></li></ul><p>A backfill was intentionally not included in this change.</p><p><code inline="">ListResponsesByConversation</code> currently has no production callers; the extproc filter uses <code inline="">GetConversationChain</code>.</p><h3>Pagination</h3><p>Listings now return the <strong>oldest N responses</strong> because the index is ordered by <code inline="">created_at</code>.</p><p><code inline="">ListOptions.Order</code> remains ignored by both Redis and memory backends.</p><p>Previously, Redis returned an arbitrary N responses due to <code inline="">SCAN</code> ordering. The new behavior matches <code inline="">MemoryStore</code>.</p><h3>Concurrent duplicate stores</h3><p>Concurrent <code inline="">StoreResponse</code> calls using the same response ID previously could both succeed, with the last write winning.</p><p>With atomic <code inline="">SET NX EX</code>, one caller now receives <code inline="">ErrAlreadyExists</code>, matching the documented contract.</p><h3>Cascade deletion</h3><p><code inline="">DeleteConversation(..., deleteResponses=true)</code> now removes <strong>all indexed responses</strong>, rather than being limited to the first 20.</p><h2>Testing</h2><p>New coverage was added to <code inline="">redis_store_test.go</code>.</p><p>This was intentionally kept outside the integration test file because <code inline="">make test</code> runs <code inline="">go test</code> without <code inline="">-tags=integration</code>. The existing unit test file therefore executes in CI, while the integration file does not.</p><h3>Added tests</h3><ul><li><p><strong><code inline="">TestRedisConversationIndexKeyIsolation</code></strong></p><ul><li><p>Requires no Redis.</p></li><li><p>Verifies index keys match neither the <code inline="">sr:conversation:</code> nor <code inline="">sr:response:</code> scan prefixes.</p></li><li><p>Prevents <code inline="">ListConversations</code> from accidentally interpreting sorted-set index keys as conversation JSON.</p></li><li><p>Runs on every CI job.</p></li></ul></li><li><p><strong><code inline="">TestRedisConversationIndexListing</code></strong></p><ul><li><p>Conversation isolation.</p></li><li><p>Chronological ordering.</p></li><li><p>Index contents.</p></li><li><p>Unknown and empty conversation IDs.</p></li><li><p>Index removal on delete.</p></li><li><p>Moving responses between conversations.</p></li><li><p>Pruning index entries whose payload has expired/deleted.</p></li></ul></li><li><p><strong><code inline="">TestRedisDeleteConversationCascade</code></strong></p><ul><li><p>Creates <code inline="">DefaultListLimit + 5</code> responses.</p></li><li><p>Verifies the cascade deletes all responses rather than stopping at the first page.</p></li><li><p>Verifies responses belonging to other conversations are untouched.</p></li></ul></li><li><p><strong><code inline="">TestRedisStoreResponseRejectsDuplicate</code></strong></p><ul><li><p>Verifies duplicate writes are rejected.</p></li><li><p>Confirms the original payload remains unchanged.</p></li><li><p>Confirms only one index entry exists.</p></li></ul></li></ul><h3>Test isolation</h3><p>Redis-backed tests use a unique per-run key prefix and <code inline="">t.Cleanup</code> instead of flushing Redis DB 0.</p><p>This is necessary because CI shares DB 0 with other test suites.</p><p>Verified that test runs leave zero keys behind.</p><h2>Validation</h2><p>Verified:</p><ul><li><p><code inline="">go test</code> without build tags</p></li><li><p><code inline="">go test</code> with integration build tags</p></li><li><p>Redis-backed tests with Redis available</p></li><li><p>Redis-free tests with Redis unavailable</p></li><li><p><code inline="">gofmt</code></p></li><li><p><code inline="">go vet</code> with both tag configurations</p></li><li><p><code inline="">golangci-lint v2.5.0</code> using repository configuration</p></li></ul><p>All checks pass with <strong>0 lint/vet issues</strong>.</p><p><code inline="">redis_store_integration_test.go</code> is unchanged.</p><h2>Files Changed</h2><ul><li><p><code inline="">src/semantic-router/pkg/responsestore/redis_store.go</code></p><ul><li><p>Secondary conversation index</p></li><li><p>Optimized listing</p></li><li><p>Atomic response creation</p></li><li><p>Index maintenance</p></li><li><p>Cascade deletion changes</p></li></ul></li><li><p><code inline="">src/semantic-router/pkg/responsestore/redis_store_test.go</code></p><ul><li><p>Secondary index and behavioral regression tests</p></li></ul></li><li><p><code inline="">config/runtime/response-api/redis-simple.yaml</code></p><ul><li><p>Updated Redis key-schema documentation</p></li></ul></li></ul><h2>Reviewer Focus</h2><p>The main areas worth reviewing are:</p><ol><li><p><strong>Index consistency and write ordering</strong> — payload is persisted before the index entry.</p></li><li><p><strong>TTL behavior</strong> — index TTL is refreshed independently from response payload TTL.</p></li><li><p><strong>Failure semantics</strong> — index write failures are logged rather than returned.</p></li><li><p><strong>Legacy data behavior</strong> — pre-index responses are intentionally not backfilled.</p></li><li><p><strong>Redis Cluster compatibility</strong> — no cross-slot multi-key operations are introduced.</p></li><li><p><strong>Pagination semantics</strong> — results are now deterministic and chronological.</p></li></ol></body></html>

### DIFF
--- a/config/runtime/response-api/redis-simple.yaml
+++ b/config/runtime/response-api/redis-simple.yaml
@@ -49,8 +49,13 @@ tls_enabled: false
 #     pool_size: 10
 
 # Redis Key Schema:
-# sr:response:{response_id}      → JSON blob of StoredResponse
-# sr:conversation:{conv_id}      → JSON blob of StoredConversation
+# sr:response:{response_id}         → JSON blob of StoredResponse
+# sr:conversation:{conv_id}         → JSON blob of StoredConversation
+# sr:conversation-index:{conv_id}   → sorted set of response IDs in that conversation,
+#                                     scored by created_at. Maintained on write so
+#                                     listing a conversation is a single ZRANGE plus
+#                                     one pipelined GET per response, instead of a
+#                                     scan over every response key.
 #
 # Both Response API and Semantic Cache share DB 0 but use different base prefixes:
 # - Response API: sr:*
--- a/config/runtime/response-api/redis-simple.yaml
+++ b/config/runtime/response-api/redis-simple.yaml
@@ -49,13 +49,9 @@ tls_enabled: false
 #     pool_size: 10
 
 # Redis Key Schema:
-# sr:response:{response_id}         → JSON blob of StoredResponse
-# sr:conversation:{conv_id}         → JSON blob of StoredConversation
-# sr:conversation-index:{conv_id}   → sorted set of response IDs in that conversation,
-#                                     scored by created_at. Maintained on write so
-#                                     listing a conversation is a single ZRANGE plus
-#                                     one pipelined GET per response, instead of a
-#                                     scan over every response key.
+# sr:response:{response_id}      → JSON blob of StoredResponse
+# sr:conversation:{conv_id}      → JSON blob of StoredConversation
+# sr:conversation-index:{conv_id} → sorted set of response IDs, scored by created_at
 #
 # Both Response API and Semantic Cache share DB 0 but use different base prefixes:
 # - Response API: sr:*
--- a/src/semantic-router/pkg/responsestore/redis_store.go
+++ b/src/semantic-router/pkg/responsestore/redis_store.go
@@ -36,6 +36,16 @@ const (
 	// ConversationKeyPrefix for conversation keys
 	// Combined with key_prefix (default "sr:"): sr:conversation:conv_xxxxx
 	ConversationKeyPrefix = "conversation:"
+
+	// ConversationIndexKeyPrefix for the secondary index that maps a conversation
+	// to the responses it contains. The index is a sorted set scored by the
+	// response created_at, so listing a conversation costs one ZRANGE plus one
+	// pipelined GET per member instead of a scan over the whole keyspace.
+	// Combined with key_prefix (default "sr:"): sr:conversation-index:conv_xxxxx
+	//
+	// The prefix intentionally does not start with ConversationKeyPrefix so index
+	// keys are not matched by the sr:conversation:* scan in ListConversations.
+	ConversationIndexKeyPrefix = "conversation-index:"
 )
 
 // The function validates configuration, establishes connection, and tests connectivity.
@@ -281,6 +291,12 @@ func (s *RedisStore) buildKey(suffix string) string {
 	return s.keyPrefix + suffix
 }
 
+// conversationIndexKey returns the key of the sorted set indexing the responses
+// that belong to a conversation.
+func (s *RedisStore) conversationIndexKey(conversationID string) string {
+	return s.buildKey(ConversationIndexKeyPrefix + conversationID)
+}
+
 func (s *RedisStore) CheckConnection(ctx context.Context) error {
 	if !s.enabled {
 		return fmt.Errorf("redis store is disabled")
@@ -319,22 +335,23 @@ func (s *RedisStore) StoreResponse(ctx context.Context, response *responseapi.St
 
 	key := s.buildKey(ResponseKeyPrefix + response.ID)
 
-	exists, err := s.client.Exists(ctx, key).Result()
-	if err != nil {
-		return fmt.Errorf("failed to check response existence: %w", err)
-	}
-	if exists > 0 {
-		return ErrAlreadyExists
-	}
-
 	data, err := json.Marshal(response)
 	if err != nil {
 		return fmt.Errorf("failed to serialize response: %w", err)
 	}
 
-	if err := s.client.Set(ctx, key, data, s.ttl).Err(); err != nil {
+	// SET NX enforces the "must not already exist" contract in a single atomic
+	// command, and guarantees the payload is readable before the ID is published
+	// to the conversation index below.
+	stored, err := s.client.SetNX(ctx, key, data, s.ttl).Result()
+	if err != nil {
 		return fmt.Errorf("failed to store response in Redis: %w", err)
 	}
+	if !stored {
+		return ErrAlreadyExists
+	}
+
+	s.indexResponse(ctx, response.ConversationID, response.ID, response.CreatedAt)
 
 	return nil
 }
@@ -375,12 +392,11 @@ func (s *RedisStore) UpdateResponse(ctx context.Context, response *responseapi.S
 
 	key := s.buildKey(ResponseKeyPrefix + response.ID)
 
-	exists, err := s.client.Exists(ctx, key).Result()
+	// Reading the stored conversation doubles as the existence check and tells us
+	// whether the response is moving between conversations.
+	previousConversationID, err := s.storedConversationID(ctx, response.ID)
 	if err != nil {
-		return fmt.Errorf("failed to check response existence: %w", err)
-	}
-	if exists == 0 {
-		return ErrNotFound
+		return err
 	}
 
 	data, err := json.Marshal(response)
@@ -391,6 +407,11 @@ func (s *RedisStore) UpdateResponse(ctx context.Context, response *responseapi.S
 	if err := s.client.Set(ctx, key, data, s.ttl).Err(); err != nil {
 		return fmt.Errorf("failed to update response in Redis: %w", err)
 	}
+
+	if previousConversationID != "" && previousConversationID != response.ConversationID {
+		s.unindexResponse(ctx, previousConversationID, response.ID)
+	}
+	s.indexResponse(ctx, response.ConversationID, response.ID, response.CreatedAt)
 
 	return nil
 }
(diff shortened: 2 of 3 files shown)


Zheng-Lu:
This is a solid fix for the core problem. I went through the diff and the new tests in detail rather than just the summary; a few things I'd like confirmed before merging, none of which require a design change:

1. **Do the new Redis-backed tests actually run in CI?**
   `newConversationIndexStore` skips (`t.Skipf`) when `localhost:6379` isn't reachable. Since these live in `redis_store_test.go` rather than the `integration`-tagged file, they run in the same job as `go test` — but only if that job actually has Redis listening. Can you confirm (e.g. a CI log link) that they're executing and not silently skipping?

2. **Index-write failures in `StoreResponse` are logged only, with no counter/metric.** 
   A failed `indexResponse` leaves the response payload durable but permanently absent from `ListResponsesByConversation` for its conversation, with nothing but a log line marking it. Given the goal here is eliminating silent conversation-history gaps, could we at least get a metric on this path so it's observable in prod, even keeping the fail-open behavior?

3. **`deleteConversationResponses`'s pipeline error is swallowed.** 
   On `pipe.Exec` failure, `DeleteConversation(..., deleteResponses=true)` still returns `nil`. The "conversation key is already gone, nothing to retry" rationale makes sense for the conversation key itself, but I'm less sure it extends to silently reporting success when the response deletions failed. Worth at least a metric/log distinguishing "conversation deleted, cascade fully succeeded" from "conversation deleted, cascade partially failed"?

4. **Minor doc nit:** 
   "No cross-slot multi-key operations are introduced" isn't quite accurate — `deleteConversationResponses` pipelines `Del` across multiple response keys plus the index key, which can span cluster slots. That's fine in practice (`Pipeline()` isn't atomic and go-redis already shards cluster pipelines per-node — `fetchResponsesPipelined` does the same today), just worth tightening the description so it doesn't read as a stronger guarantee than what's actually true.

*Not asking for changes here, but flagging for the record:* 
`ListResponsesByConversation` still does `ZRANGE 0 -1` (full index) before truncating to the pagination limit in memory, so a single very long conversation still costs `O(conversation size)` per call regardless of `opts.Limit`. Fine today since this method has no production callers yet, but worth tracking before #2999 wires `conversation_id`-based history through it — same for the legacy/pre-index backfill gap, which is explicitly and reasonably deferred here but shouldn't get lost once this path goes live.

Nice fix overall

Xunzhuo (changes requested):
The new conversation index hides every response stored before upgrade. Writes also persist the response first and swallow index failure; a retry then sees `AlreadyExists` and cannot repair discovery. Please provide migration or lazy fallback for existing data plus atomic or repairable index writes, with upgrade and partial-failure regressions.

Zheng-Lu:
> The new conversation index hides every response stored before upgrade. Writes also persist the response first and swallow index failure; a retry then sees `AlreadyExists` and cannot repair discovery. Please provide migration or lazy fallback for existing data plus atomic or repairable index writes, with upgrade and partial-failure regressions.

Hi @Xunzhuo, since #2999 is currently blocked by this bug and this draft PR is stale for  a while, so I'm taking over this bug fix and created a new PR #3502 that took your suggestion into account.

gagandhakrey:
thanks @xunzhuo for the review and @zheng-lu for picking this up. i was aware of the issues with my current pr and that it required much more work thorough testing, but unfortunately i was busy with something else, so i left the pr as a draft. really appreciate you taking it over. i will close my pr now

PR #2818, opened 2026-08-08
[Router] Persist router replay recipe identity in PostgreSQL

### LINKED ISSUE #2792: bug: persist recipe identity in PostgreSQL Router Replay records
## Summary

Persist the routing recipe on the PostgreSQL Router Replay path so recipe filters and aggregates have the same data as the in-memory and JSON-backed implementations.

Parent milestone: #2791

## Background and current state

`routerreplay/store.Record` contains a `Recipe` field and the ExtProc recorder populates it from the resolved routing context. The PostgreSQL create-table and insert column lists in `src/semantic-router/pkg/routerreplay/store/postgres.go` do not include a `recipe` column. As a result, a record written through this backend cannot round-trip recipe identity through the dedicated field.

This issue is deliberately narrow. It does not redesign Router Replay or create a billing ledger.

## Scope

- Add a nullable `recipe` column to the PostgreSQL schema and migration path.
- Include `recipe` in insert, select, scan, update/codec, and test fixtures where applicable.
- Preserve existing rows and treat a missing historical recipe as unknown rather than inventing a value.
- Verify list, detail, filter, and aggregate behavior on PostgreSQL.

## Acceptance criteria

- [ ] A record with `Record.Recipe` survives a PostgreSQL write/read round trip.
- [ ] Recipe filters and aggregate grouping return the expected PostgreSQL records.
- [ ] Existing tables are upgraded without destructive migration.
- [ ] Insert columns, placeholders, and argument alignment tests remain exact.
- [ ] Redis, memory, and other Replay stores retain current behavior.

## Likely change surfaces

- `src/semantic-router/pkg/routerreplay/store/postgres.go`
- `src/semantic-router/pkg/routerreplay/store/postgres_record_row.go`
- `src/semantic-router/pkg/routerreplay/store/postgres_record_codec.go`
- Router Replay PostgreSQL tests

## Validation

Run the repository harness for the touched files, then the focused Router Replay store and aggregate tests. Include a real PostgreSQL integration case rather than relying only on query-string assertions.

olifarhaan:
Closes #2792

## Purpose

`Record.Recipe` is populated by the ExtProc recorder on every request, and the Router Replay API filters on it, free-text searches it, and builds its `available_recipes` dropdown from it. PostgreSQL was the only backend that maps records column by column, since memory, Redis, Qdrant, and Milvus marshal the whole `Record`, so it was the only one that dropped the field. Writes discarded the recipe and reads always returned `""`, which left `?recipe=` matching nothing and the dropdown empty on Postgres-backed deployments

This carries recipe through the insert, select, scan, and decode paths, and adds the column through the existing `ADD COLUMN IF NOT EXISTS` block so already-created tables upgrade without a destructive migration. Rows written before the column existed scan as NULL and decode to an empty recipe rather than failing the read

The column is appended to the end of the lists rather than slotted next to `decision`. Inserting it in the middle would renumber every `$N` placeholder after it for no behavioral gain

Two commits: the fix, then the tests. Module affected: `Router`

While adding the tests I also filled in a gap in the existing coverage. There was an alignment test for the INSERT columns against the `$N` placeholders and against `postgresInsertRecord.args()`, but nothing checking the INSERT list against `postgresRecordSelectColumns`. `scanDestinations()` is positional, so a column added to one list but not the other round trips as a zero value instead of failing loudly, which is exactly how recipe stayed invisible

## Test Plan

The changed files are pure Go with no new dependencies, so the repo-native Go gates cover them:

```bash
make check-go-mod-tidy
make go-lint
make test-semantic-router
pre-commit run --files src/semantic-router/pkg/routerreplay/store/postgres.go \
  src/semantic-router/pkg/routerreplay/store/postgres_record_row.go \
  src/semantic-router/pkg/routerreplay/store/postgres_insert_query_test.go \
  src/semantic-router/pkg/routerreplay/store/postgres_recipe_column_test.go
make agent-ci-lint CHANGED_FILES="src/semantic-router/pkg/routerreplay/store/postgres.go,src/semantic-router/pkg/routerreplay/store/postgres_record_row.go,src/semantic-router/pkg/routerreplay/store/postgres_insert_query_test.go,src/semantic-router/pkg/routerreplay/store/postgres_recipe_column_test.go"
```

To confirm the new tests actually fail when the fix is broken rather than passing vacuously, each part of the change can be reverted in isolation and the suite re-run:

```bash
go test ./pkg/routerreplay/store/ -count=1 -run Postgres -v
```

## Test Result

All gates pass locally on darwin/arm64 with `golangci-lint` v2.5.0, matching the version pinned in `.github/workflows/pre-commit.yml`. `make test-semantic-router` reports one failure, `TestHybridCachePendingRequest` in `pkg/cache`, which is a known flake unrelated to this change. It fails on other contributors' PRs that do not touch `pkg/cache` either (#2530, #2507, #2810), and it passed on my own PR #2819. The test writes to the Milvus-backed hybrid cache, sleeps a fixed 100ms "for indexing", then asserts the entry is findable, so a loaded runner fails it. Locally it cannot run at all without a Milvus instance, which is why `make test-semantic-router` defaults `SKIP_MILVUS_TESTS=true`. The `test-and-build` job is flaky on `main` for the same reason, failing 4 of the last 8 runs there on a rotating set of tests. Everything else is green: `make go-lint` reports 0 issues, `make check-go-mod-tidy` passes, `make agent-ci-lint` exits 0, and every pre-commit hook that applies to these files passes including the supply chain AST scan and the agent changed-files lint

The six `Postgres` tests in the package pass, four of them new. Reverting each part of the fix in turn fails at least one test in every case:

| Reverted | Fails |
| --- | --- |
| `recipe` removed from `postgresRecordSelectColumns` | `TestPostgresInsertSelectColumnAlignment`, `TestPostgresBackendWiresDemotedHeaderFields` |
| `recipe` removed from `args()` | `TestPostgresInsertQueryColumnArgsAlignment`, `TestPostgresRecordCarriesRecipe` |
| `&row.recipe` removed from `scanDestinations()` | `TestPostgresInsertSelectColumnAlignment`, `TestPostgresRecordCarriesRecipe` |
| decode assignment removed | `TestPostgresRecordRowDecodesRecipe` |
| `ALTER TABLE` migration removed | `TestPostgresCreateTableAddsRecipeColumn` |

One gap worth flagging. There is no live PostgreSQL round trip anywhere in this package, so the write and read paths are pinned through the production mapping functions rather than through a real driver. The column ordering that makes that equivalent is itself now pinned by `TestPostgresInsertSelectColumnAlignment`. I also could not run `pkg/extproc` locally, since it fails to link without the Rust bindings on this machine, but this change touches no extproc code and the API-level filter reads `Record.Recipe` unchanged

### DIFF
--- a/src/semantic-router/pkg/routerreplay/store/postgres.go
+++ b/src/semantic-router/pkg/routerreplay/store/postgres.go
@@ -36,8 +36,8 @@ const postgresInsertQueryTemplate = `
 			prompt_tokens, cached_prompt_tokens, cache_write_tokens, completion_tokens, total_tokens,
 			actual_cost, baseline_cost, cost_savings, currency, baseline_model,
 			session_id, turn_index, previous_response_id, conversation_id,
-			cache_similarity, context_token_count, hallucination_span_details
-		) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26, $27, $28, $29, $30, $31, $32, $33, $34, $35, $36, $37, $38, $39, $40, $41, $42, $43, $44, $45, $46, $47, $48, $49, $50, $51, $52, $53, $54, $55, $56, $57, $58, $59, $60)
+			cache_similarity, context_token_count, hallucination_span_details, recipe
+		) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26, $27, $28, $29, $30, $31, $32, $33, $34, $35, $36, $37, $38, $39, $40, $41, $42, $43, $44, $45, $46, $47, $48, $49, $50, $51, $52, $53, $54, $55, $56, $57, $58, $59, $60, $61)
 	`
 
 const postgresCreateTableQueryTemplate = `
@@ -130,6 +130,7 @@ const postgresCreateTableQueryTemplate = `
 		ALTER TABLE {{table}} ADD COLUMN IF NOT EXISTS cache_similarity REAL DEFAULT 0;
 		ALTER TABLE {{table}} ADD COLUMN IF NOT EXISTS context_token_count INTEGER DEFAULT 0;
 		ALTER TABLE {{table}} ADD COLUMN IF NOT EXISTS hallucination_span_details JSONB;
+		ALTER TABLE {{table}} ADD COLUMN IF NOT EXISTS recipe VARCHAR(255);
 		CREATE INDEX IF NOT EXISTS idx_{{table}}_timestamp ON {{table}} (timestamp DESC);
 		CREATE INDEX IF NOT EXISTS idx_{{table}}_created_at ON {{table}} (created_at);
 		CREATE INDEX IF NOT EXISTS idx_{{table}}_request_id ON {{table}} (request_id);
--- a/src/semantic-router/pkg/routerreplay/store/postgres_insert_query_test.go
+++ b/src/semantic-router/pkg/routerreplay/store/postgres_insert_query_test.go
@@ -49,6 +49,31 @@ func TestPostgresInsertQueryColumnArgsAlignment(t *testing.T) {
 	}
 }
 
+// TestPostgresInsertSelectColumnAlignment pins the write path against the read
+// path. postgresRecordRow.scanDestinations() is positional, so it only decodes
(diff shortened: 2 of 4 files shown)


olifarhaan:
test-and-build failure is TestHybridCachePendingRequest in pkg/cache, which this PR does not touch. It reproduces on unmodified main.

olifarhaan:
Root-caused the red check: it is a Milvus read-consistency race, now filed as #2822, with the underlying config bug as #2821.

Xunzhuo (approved):
LGTM thanks

PR #2820, opened 2026-08-08
[Router] Require classifier runtime weights so partial downloads heal

### LINKED ISSUE #2669: [Bug] Restore incomplete ModernBERT runtime weights during model healing
### Proposed Workgroup

### Describe the bug

Model completeness checks omit required ModernBERT runtime weights, so a partial download can be treated as complete and the healing path never fetches the missing files.

### To Reproduce

### Expected behavior

Completeness should be derived from the resolved model and runtime contract, and healing should restore every required artifact before activation.

### Impact

##### Completion signal

Tests cover partial and corrupt ModernBERT artifacts and prove that the deployment either heals to a complete immutable identity or remains not ready with an actionable error.

Parent Epic: #2782

olifarhaan:
Closes #2669

## Purpose

On the mmBERT-32K backend, which is the default for all three, the category, PII, and jailbreak classifiers load through `TraditionalModernBertTokenClassifier::new_with_variant`. It hard-reads `config.json`, `tokenizer.json`, and `model.safetensors` from the model root and fails outright if any is missing

Completeness only demanded `config.json` plus any weight-shaped file, and `hasModelWeights` matches `*.safetensors`, `*.bin`, and `*.onnx` recursively. So an interrupted download that left the companion mapping and a nested adapter blob behind read as complete, was never re-fetched, and since these classifiers register with `bestEffort: false`, every restart failed the same way

`addClassifierModelRequiredFiles` mirrors the existing `addEmbeddingModelRequiredFiles` from #2172 and records the root weights and tokenizer for the three models. Module affected: `Router`

On the LoRA caveat raised in the issue: this is gated on `use_mmbert_32k`, which is exactly the flag selecting the initialiser that has no LoRA fallback. With it off, PII and jailbreak init auto-detects LoRA directories that legitimately carry adapter weights instead of root weights, so those keep the looser heuristic. A test pins that carve-out

## Test Plan

```bash
go test ./pkg/modeldownload/ -count=1
make test-semantic-router
make go-lint
make check-go-mod-tidy
make agent-ci-lint CHANGED_FILES="src/semantic-router/pkg/modeldownload/config_parser.go,src/semantic-router/pkg/modeldownload/classifier_completeness_test.go"
```

To confirm the tests are not vacuous, revert each part of the fix and re-run

## Test Result

All four new tests fail on the base for the right reason, for example:

```
--- FAIL: TestBuildModelSpecsRequiresClassifierRuntimeWeights
    models/mmbert32k-pii-detector-merged RequiredFiles = []string{"config.json", "pii_type_mapping.json"}, missing "model.safetensors"
--- FAIL: TestPartialClassifierDirReportedIncomplete
    partial classifier dir reported complete; the runtime hard-loads model.safetensors and would fail at init
```

Reverting each part of the change in turn fails at least one test every time:

| Reverted | Fails |
| --- | --- |
| `addClassifierModelRequiredFiles` call dropped | `TestBuildModelSpecsRequiresClassifierRuntimeWeights`, `TestPartialClassifierDirReportedIncomplete` |
| `use_mmbert_32k` gate dropped, so LoRA also demands root weights | `TestLoRAClassifierKeepsHeuristicCompleteness` |

`make go-lint` reports 0 issues, `make check-go-mod-tidy` and `make agent-ci-lint` pass, and every applicable pre-commit hook passes. `make test-semantic-router` reports one failure, `TestHybridCachePendingRequest` in `pkg/cache`, which is a known flake unrelated to this change. It fails on other contributors' PRs that do not touch `pkg/cache` either (#2530, #2507, #2810), and it passed on my own PR #2819. The test writes to the Milvus-backed hybrid cache, sleeps a fixed 100ms "for indexing", then asserts the entry is findable, so a loaded runner fails it. Locally it cannot run at all without a Milvus instance, which is why `make test-semantic-router` defaults `SKIP_MILVUS_TESTS=true`. It is currently failing on several unrelated PRs and on `main` itself

One thing I did not change: `DefaultRequiredFiles` still allows the loose nested-weight heuristic for every other model. Tightening that globally would be a much wider change than this issue

The `models/` prefix is now a named constant, since the linter flags the third repetition this change would have introduced

### DIFF
--- a/src/semantic-router/pkg/modeldownload/classifier_completeness_test.go
+++ b/src/semantic-router/pkg/modeldownload/classifier_completeness_test.go
@@ -1,0 +1,157 @@
+package modeldownload
+
+import (
+	"path/filepath"
+	"slices"
+	"testing"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+const (
+	testCategoryModelPath  = "models/mmbert32k-intent-classifier-merged"
+	testPIIModelPath       = "models/mmbert32k-pii-detector-merged"
+	testJailbreakModelPath = "models/mmbert32k-jailbreak-detector-merged"
+)
+
+// newMmBERT32KClassifierConfig mirrors the shipped defaults: all three classifiers on the
+// mmBERT-32K backend, pointed at merged (non-LoRA) model directories, with routing that
+// actually consumes each signal so the models survive the optional-model feature gates.
+func newMmBERT32KClassifierConfig() *config.RouterConfig {
+	return &config.RouterConfig{
+		MoMRegistry: map[string]string{
+			testCategoryModelPath:  "llm-semantic-router/mmbert32k-intent-classifier-merged",
+			testPIIModelPath:       "llm-semantic-router/mmbert32k-pii-detector-merged",
+			testJailbreakModelPath: "llm-semantic-router/mmbert32k-jailbreak-detector-merged",
+		},
+		InlineModels: config.InlineModels{
+			Classifier: config.Classifier{
+				CategoryModel: config.CategoryModel{
+					ModelID:             testCategoryModelPath,
+					UseMmBERT32K:        true,
+					CategoryMappingPath: testCategoryModelPath + "/category_mapping.json",
+				},
+				PIIModel: config.PIIModel{
+					ModelID:        testPIIModelPath,
+					UseMmBERT32K:   true,
+					PIIMappingPath: testPIIModelPath + "/pii_type_mapping.json",
+				},
+			},
+			PromptGuard: config.PromptGuardConfig{
+				Enabled:              true,
+				ModelID:              testJailbreakModelPath,
+				Variant:              config.PromptGuardVariantMmBERT32K,
+				JailbreakMappingPath: testJailbreakModelPath + "/jailbreak_type_mapping.json",
+			},
+		},
+		IntelligentRouting: config.IntelligentRouting{
+			Decisions: []config.Decision{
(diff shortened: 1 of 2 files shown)


Xunzhuo:
Thanks for adding the classifier completeness regressions. This branch currently conflicts with main, so please rebase it onto current main and rerun the model-download tests; I will review the refreshed head.

Xunzhuo (changes requested):
Provisioning discovers request-reachable named recipes, but required classifier files are added only from flat/default fields. A named reachable mmBERT recipe can therefore download without mandatory root weights/tokenizer and be accepted as complete. Please apply completeness requirements to the same reachable profiles and cover a named entrypoint.

adaamko (commented):
agree with xunzhuo on named recipes, and the fix is probably structural rather than another loop: ExtractRequiredFilesByModel already walks the config by reflection and picks up mapping files from named recipes for free, while addClassifierModelRequiredFiles reads three flat fields. if the root-weight rule hangs off the same walk (any CategoryModel / PIIModel / PromptGuardConfig it visits, with the same mmbert32k + local checks), reachable named recipes are covered without a second list to maintain.

on the carve-out: the comment and test say "LoRA", but what the code skips is the whole candle backend, and a merged non-LoRA directory on candle is exactly as exposed to the half-download case as the mmbert32k one. fine to leave that for later, but name it as the candle backend so nobody reads the LoRA test as covering it.

conflicts with #3533 and #2828 in the same function, see my note on #2828 for an order.

FAUST-BENCHOU (changes requested):
pls resolve conflicts

PR #2828, opened 2026-08-09
[Router] Download aliased model paths to the directory the runtime loads

### LINKED ISSUE #2676: [Bug] Validate completeness for alias-form embedding model paths
### Proposed Workgroup

### Describe the bug

Embedding models referenced through an alias-form path can bypass artifact completeness requirements that are enforced for canonical paths.

### To Reproduce

### Expected behavior

Alias resolution should happen before completeness validation so equivalent model references enforce the same required artifact set.

### Impact

##### Completion signal

Canonical and alias-form references share completeness, healing, identity, and failure tests.

Parent Epic: #2782

olifarhaan:
Closes #2676

## Purpose

`ResolveModelPath` lets a config name a model by a registry alias, and every consumer of a `*ModelPath` field uses it before handing the path to candle: the embedding runtime, local classifier signal rules, and the modality detector. The download collector did not. It recorded the literal configured string and skipped anything without a `models/` prefix, so the two halves disagreed about which directory holds the model.

A bare alias (`mmbert_model_path: mmbert`) was therefore never downloaded at all, and a `models/<alias>` form downloaded a full snapshot into a directory the runtime never opens. Either way the router starts, finds the canonical directory missing, and the embedding runtime fails against a model the config legitimately declared.

`recordModelPath` now canonicalizes the fields the runtime canonicalizes, `addEmbeddingModelRequiredFiles` keys the #2172 weight requirements off the same resolved path, and the embedding and modality feature gates compare in that same path space so a disabled model is still skipped. `ModelID` deliberately stays literal: the built-in classifiers pass it to the bindings verbatim, so resolving it here would move the download away from the directory they load. Module affected: `Router`

The issue frames this as completeness requirements being bypassed. That is the smaller half. The requirements are recorded against a directory that is also the wrong download target, so the snapshot itself lands in the wrong place.

## Test Plan

```bash
cd src/semantic-router && go test ./pkg/modeldownload/ -count=1
make test-semantic-router
make go-lint
make check-go-mod-tidy
make agent-ci-lint CHANGED_FILES="src/semantic-router/pkg/modeldownload/config_parser.go,src/semantic-router/pkg/modeldownload/feature_gates.go,src/semantic-router/pkg/modeldownload/model_path_alias_test.go"
```

To confirm the tests are not vacuous, revert each part of the fix and re-run.

For the behavior itself, take `config/config.yaml`, point `mmbert_model_path` at an alias of the same model, load it through `config.Load`, and compare `BuildModelSpecs` output against `config.ResolveModelPath` on the configured value.

## Test Result

Loading the shipped `config/config.yaml` with only `mmbert_model_path` changed, on `main`:

```
configured mmbert_model_path: "mmbert"
runtime loads:                "models/mmbert-embed-32k-2d-matryoshka"
downloader fetches:           nothing

configured mmbert_model_path: "models/mom-embedding-ultra"
runtime loads:                "models/mmbert-embed-32k-2d-matryoshka"
downloader fetches:           "models/mom-embedding-ultra"
```

On this branch both forms agree with the runtime:

```
configured mmbert_model_path: "mmbert"
runtime loads:                "models/mmbert-embed-32k-2d-matryoshka"
downloader fetches:           "models/mmbert-embed-32k-2d-matryoshka"

configured mmbert_model_path: "models/mom-embedding-ultra"
runtime loads:                "models/mmbert-embed-32k-2d-matryoshka"
downloader fetches:           "models/mmbert-embed-32k-2d-matryoshka"
```

The four new tests fail on the base, for example:

```
--- FAIL: TestBuildModelSpecsTargetsTheDirectoryTheRuntimeLoads/bare_alias
    BuildModelSpecs() returned 0 specs for "mmbert", want 1 for "models/mmbert-embed-32k-2d-matryoshka"
--- FAIL: TestBuildModelSpecsTargetsTheDirectoryTheRuntimeLoads/models-prefixed_alias
    BuildModelSpecs() downloads "models/mom-embedding-ultra", but the runtime loads "models/mmbert-embed-32k-2d-matryoshka"
```

Reverting each part of the change in turn fails at least one test every time:

| Reverted | Fails |
| --- | --- |
| collector resolution | `TargetsTheDirectoryTheRuntimeLoads/bare_alias`, `/models-prefixed_alias`, `AliasedEmbeddingModelKeepsWeightRequirements` |
| required-files resolution | `AliasedEmbeddingModelKeepsWeightRequirements` |
| feature-gate resolution | `AliasedEmbeddingModelStillSkippedForRemoteBackend` |
| `ModelID` carve-out | `KeepsClassifierModelIDLiteral` |

`make test-semantic-router`, `make go-lint` (0 issues), `make check-go-mod-tidy`, and `make agent-ci-lint` all pass.

One thing I did not change: `ModelID` and `*ModelPath` still resolve aliases differently across the repo, and this PR matches that split rather than settling it. Unifying it means touching the classifier mapping paths too, which is a much wider change than this issue. Worth noting for merge order that #2820 also edits `config_parser.go`; whichever lands second needs a small rebase.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -64,7 +64,7 @@ func recordModelPath(fieldName string, field reflect.Value, paths *[]string, see
 		return
 	}
 
-	path := field.String()
+	path := resolveAliasForField(fieldName, field.String())
 	if path == "" || !strings.HasPrefix(path, "models/") || seen[path] {
 		return
 	}
@@ -82,6 +82,20 @@ func isModelPathField(fieldName string) bool {
 		fieldName == "Qwen3ModelPath" ||
 		fieldName == "GemmaModelPath" ||
 		strings.HasSuffix(fieldName, "ModelPath")
+}
+
+// resolveAliasForField canonicalizes the registry aliases the runtime itself resolves,
+// so the snapshot lands in the directory that will actually be loaded.
+//
+// Every consumer of a *ModelPath field passes it through config.ResolveModelPath first:
+// the embedding runtime, local classifier signal rules, and the modality detector. A
+// ModelID reaches the candle bindings verbatim, so canonicalizing one here would move
+// its download away from the directory the classifier opens.
+func resolveAliasForField(fieldName, path string) string {
+	if fieldName == "ModelID" {
+		return path
+	}
+	return config.ResolveModelPath(path)
 }
 
 // isModelDirectory checks if a path looks like a model directory (not a file)
@@ -185,7 +199,7 @@ func addEmbeddingModelRequiredFiles(cfg *config.RouterConfig, requiredFilesByMod
 	}
 
 	// MmBertModelPath holds the configured semantic embedding model directory.
-	path := cfg.MmBertModelPath
+	path := config.ResolveModelPath(cfg.MmBertModelPath)
 	if path == "" || !strings.HasPrefix(path, "models/") {
 		return
 	}
--- a/src/semantic-router/pkg/modeldownload/feature_gates.go
+++ b/src/semantic-router/pkg/modeldownload/feature_gates.go
@@ -13,12 +13,14 @@ var optionalModelFeatureGates = []modelFeatureGate{
 			return !cfg.EmbeddingModels.UsesRemoteEmbeddingBackend()
 		},
 		paths: func(cfg *config.RouterConfig) []string {
+			// Canonical form, so these match the paths resolveAliasForField collected
+			// for the same fields. Gates over ModelID stay literal for the same reason.
 			return []string{
-				cfg.Qwen3ModelPath,
-				cfg.GemmaModelPath,
-				cfg.MmBertModelPath,
-				cfg.MultiModalModelPath,
-				cfg.BertModelPath,
+				config.ResolveModelPath(cfg.Qwen3ModelPath),
+				config.ResolveModelPath(cfg.GemmaModelPath),
+				config.ResolveModelPath(cfg.MmBertModelPath),
+				config.ResolveModelPath(cfg.MultiModalModelPath),
+				config.ResolveModelPath(cfg.BertModelPath),
 			}
 		},
 	},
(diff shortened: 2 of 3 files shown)


theohsiung:
Hi @olifarhaan, sorry for the late review, can you solve the conflicts. Thanks a lot

Xunzhuo (changes requested):
Alias resolution is applied to every reflected `*ModelPath`, but not every consumer resolves aliases: GMTRouter copies its `ModelPath` literally into runtime config. A literal directory that collides with a registry alias is then downloaded elsewhere. Please whitelist the actual alias-resolving owners/fields and add a GMTRouter collision regression.

adaamko (commented):
xunzhuo's point stands, and i'd flip the shape: resolving every reflected *ModelPath and carving out ModelID is the inverse of the contract. GMTRouterSelectionConfig.ModelPath is copied literally into the runtime, so a models/<alias> there is downloaded to the canonical dir and loaded from the literal one. the fields whose consumers actually call ResolveModelPath are the embedding runtime paths, the generic local classifier and the modality detector, so the whitelist is short and everything else stays literal.

the gate side is right but its now the same canonical-vs-literal decision encoded twice, once in recordModelPath and once in feature_gates.go. one helper that answers "resolved or literal" per field, used by both, keeps them from drifting the next time a path field is added.

this conflicts with #3533 and #2820 in the same stretch of BuildModelSpecs. suggested order: #3533 first (smallest, a bug, exclude map keyed by path), then #2820, then this one rebased on both, because alias resolution has to key the exclude map and the classifier required-files map by the resolved path. i'll turn the rebase review around quickly.

FAUST-BENCHOU (changes requested):
conflicts

Xunzhuo:
Thanks @olifarhaan for identifying and testing the alias/download mismatch. Closing as superseded by merged #3944: main now resolves aliases consistently across provisioning, runtime paths, required files and feature gates, with regressions for bare and models-prefixed aliases. Classifier ModelID aliases are also resolved by the runtime, so the older literal-ModelID carve-out is no longer needed.

PR #2837, opened 2026-08-10
[Dashboard] report partial user edits honestly and refresh the table

### LINKED ISSUE #2836: bug: user edit dialog saves the role change, then reports it as a total failure
### Describe the bug

The user edit dialog sends two requests: a `PATCH` for role and status, then a `POST` to rotate the password. If the password request fails, the role and status change has already committed, but the dialog shows a bare error with no sign that anything saved. `fetchUsers()` only runs on the success path, so the table keeps showing the old role until you reload.

The password request is easy to fail by accident. bcrypt rejects anything over 72 bytes, and neither the input nor the handler checks the length, so the server returns 500 with the raw bcrypt error text.

Net result: you are told the edit failed, half of it persisted, and the UI shows a value the server no longer has.

### To Reproduce

1. Log in as admin, go to `/users`.
2. Click **Edit** on a non-admin user. Note their role, for example `read`.
3. Change **Role** to `write`.
4. In **New password**, paste ~73 characters
5. Save. The dialog shows `bcrypt: password length exceeds 72 bytes`.
6. Cancel. The table still shows `read`.
7. Reload. The role is now `write`.

Network tab shows `PATCH /api/admin/users/{id}` 200, then `POST /api/admin/users/password` 500.

### Expected behavior

1. A password over 72 bytes should be rejected before anything is sent, so the role change cannot commit on its own. Nothing should be saved and the table should be unchanged.
2. If the password step fails after the role change has committed for any other reason (expired session, dropped connection, target user deleted), the message should say the role and status were saved and only the password failed.
3. The table should refresh on the failure path either way, so it never shows a value the server no longer has.
4. The server should return 400 with a clear message for an oversized password, not 500 leaking the bcrypt error.

### Affected layer

None

### Additional context

Code: `UsersPage.tsx#L291-L323`, `admin_meta_handlers.go#L106-L110`, `service.go#L84-L90`, `UsersPageUserDialog.tsx#L210-L229`.

The 500 is folded into the same fix rather than filed separately: the frontend change covers every partial-failure cause (expired session, dropped connection, deleted user), while the 400 mapping removes the one an admin hits by accident and stops a client input error from reading as a server fault to every API caller. Both encode the same 72 byte rule, so splitting them would leave it defined in only one place.

Separately, there is no minimum password length either, so a single character password is accepted. Happy to add one if you want to pick a number.

https://github.com/user-attachments/assets/c3ce7e08-ba38-4d80-a478-52c2c3c66d90

abhinav-m22:
Closes #2836

## Purpose

The user edit dialog writes role and status with a `PATCH`, then rotates the password with a `POST`. If the second call failed, the first had already committed but the dialog reported a flat failure and never called `fetchUsers()`, so the table kept showing the old role until a manual reload.

- Validate the password before any request is sent, in bytes rather than characters
- Track whether the `PATCH` committed, and say so when a later step fails
- Refresh the table on every failure path, since a failed request does not prove nothing changed
- Map an oversized password to 400 instead of 500, enforced inside `Service.HashPassword` so all call sites are covered, and stop echoing hashing errors to the client

Module: `Dashboard`

No `maxLength` on the input on purpose: it counts UTF-16 units and would silently truncate a pasted password into one the user cannot sign in with.

https://github.com/user-attachments/assets/60b81ac7-8a69-4bbf-8e39-dcb813f49279

Screenshot From 2026-08-10 18-52-09

---
<details>

</details>


Xunzhuo:
@abhinav-m22 thanks, this is cool ! cc @AayushSaini101

AayushSaini101 (approved):
LGTM thank you for the fix : )

PR #2842, opened 2026-08-11
[Router] Honor replay Milvus consistency level

### LINKED ISSUE #2840: bug: Router replay Milvus store ignores the configured consistency_level
### Describe the bug

### To Reproduce

### Expected behavior

### Affected layer

global level

### Additional context

iroiro147:
Fixes #2840

## Purpose

Honor router replay Milvus consistency_level configuration when creating the collection and when reading records. The documented levels are mapped case-insensitively, with Session as the read-your-writes default and a warning for unknown values.

## Test Plan

- go test ./pkg/routerreplay/store -run TestMilvusConsistencyLevel -count=1
- go vet ./pkg/routerreplay/store
- gofmt and git apply --check --recount

The commit is DCO signed-off.


src/semantic-router/pkg/routerreplay/store/milvus.go
```diff
@@ -23,15 +24,16 @@ const (
...
+	collectionName   string
+	consistencyLevel entity.ConsistencyLevel
+	ttl              time.Duration
+	asyncWrites      bool
+	asyncChan        chan asyncOp
+	done             chan struct{}
+	writerDone       chan struct{}
+	updateMu         sync.Mutex
+	lifecycle        *storageLifecycle
 }
```
drivebyer: Track whether this process has written yet, and use it to floor Session at Bounded on reads — see the review comment for why the SDK degrades Session to Eventually without a session timestamp.

```suggestion
type MilvusStore struct {
	client           client.Client
	collectionName   string
	consistencyLevel entity.ConsistencyLevel
	wrote            atomic.Bool
	ttl              time.Duration
	asyncWrites      bool
	asyncChan        chan asyncOp
	done             chan struct{}
	writerDone       chan struct{}
	updateMu         sync.Mutex
	lifecycle        *storageLifecycle
}

// readLevel returns the consistency level to use for reads. The SDK degrades
// Session to Eventually until this client has written to the collection (no
// session timestamp exists yet), which is weaker than the Bounded these reads
// effectively ran at before the level became configurable. Flooring Session at
// Bounded until this process has written keeps the fix from ever being a
// downgrade, while still giving read-your-writes once a session timestamp exists.
func (m *MilvusStore) readLevel() entity.ConsistencyLevel {
	if m.consistencyLevel == entity.ClSession && !m.wrote.Load() {
		return entity.ClBounded
	}
	return m.consistencyLevel
}
```

Three companion edits fall outside the diff hunks, so I can't attach suggestions to them: add `"sync/atomic"` to the import block, and `m.wrote.Store(true)` after the successful `m.client.Insert` in `Add` and after the successful `m.client.Upsert` in `replaceRecord`.

drivebyer (commented):
Blocking on the unset default: `milvusConsistencyLevel("")` returns Session, and the SDK degrades Session to Eventually until this client's first write (`options.go:240-245`) — weaker than the Bounded these reads run at today. Floor it; suggestions inline. Closing my duplicate #2949.

src/semantic-router/pkg/routerreplay/store/milvus.go
```diff
@@ -295,7 +302,7 @@ func (m *MilvusStore) getRecord(ctx context.Context, id string) (Record, bool, e
 
 	// Query by ID
 	expr := fmt.Sprintf("id == '%s'", id)
-	result, err := m.client.Query(ctx, m.collectionName, nil, expr, []string{"id", "timestamp", "data"})
+	result, err := m.client.Query(ctx, m.collectionName, nil, expr, []string{"id", "timestamp", "data"}, client.WithSearchQueryConsistencyLevel(m.consistencyLevel))
```
drivebyer:
```suggestion
	result, err := m.client.Query(ctx, m.collectionName, nil, expr, []string{"id", "timestamp", "data"}, client.WithSearchQueryConsistencyLevel(m.readLevel()))
```

src/semantic-router/pkg/routerreplay/store/milvus.go
```diff
@@ -350,6 +357,7 @@ func (m *MilvusStore) listRecords(ctx context.Context) ([]Record, error) {
 		"",
 		[]string{"id", "timestamp", "data"},
 		client.WithLimit(10000),
+		client.WithSearchQueryConsistencyLevel(m.consistencyLevel),
```
drivebyer:
```suggestion
		client.WithSearchQueryConsistencyLevel(m.readLevel()),
```

src/semantic-router/pkg/routerreplay/store/milvus.go
```diff
@@ -23,15 +24,16 @@ const (
...
-	ttl            time.Duration
-	asyncWrites    bool
-	asyncChan      chan asyncOp
-	done           chan struct{}
-	writerDone     chan struct{}
-	updateMu       sync.Mutex
-	lifecycle      *storageLifecycle
+	client           client.Client
+	collectionName   string
+	consistencyLevel entity.ConsistencyLevel
```
drivebyer, line 30:
```suggestion
	consistencyLevel entity.ConsistencyLevel
	wrote            atomic.Bool
```

src/semantic-router/pkg/routerreplay/store/milvus.go
drivebyer, line 39:
```suggestion
}

// readLevel floors Session at Bounded until this process has written, since the
// SDK degrades Session to Eventually while no session timestamp exists.
func (m *MilvusStore) readLevel() entity.ConsistencyLevel {
	if m.consistencyLevel == entity.ClSession && !m.wrote.Load() {
		return entity.ClBounded
	}
	return m.consistencyLevel
}
```

drivebyer (commented):
Re-split the struct suggestion — the earlier range spanned deleted lines, which GitHub can't apply. Two single-line suggestions below. Still needs `"sync/atomic"` in the imports, plus `m.wrote.Store(true)` after the `Insert` in `Add` and the `Upsert` in `replaceRecord`.

drivebyer (approved):

Issue #2215, opened 2026-06-17
bug: Streaming response terminated prematurely due to preview usage chunk


seeger12138:
### Describe the bug

# Bug Report: Streaming response terminated prematurely due to preview usage chunk

## Summary

When an upstream provider (e.g.,  gateway) sends a "preview" chunk with `usage` data but no content (`choices: []`) as the **first** chunk in a streaming response, the Router's Anthropic SSE emitter misidentifies it as a stream termination signal, causing:

1. **Premature stream termination**: `message_stop` is emitted after the first chunk, making the Anthropic client (Claude Code) believe the response is complete
2. **Wrong `completion_tokens` recorded**: The final usage captured comes from the preview chunk (`completion_tokens: 4`) instead of the real final chunk (`completion_tokens: 252`)
3. **Content lost**: All subsequent content deltas and tool call deltas are ignored by the caching/metrics layer

## Affected Files

- **Bug location**: `src/semantic-router/pkg/anthropic/sse_out.go`
  - Lines 153–161: `emitChunkEvents` handling of empty-`Choices` chunks
  - Lines 599–605: `chunkHasUsage` helper
  - Lines 215–221: `isTerminalChunk` helper
- **Trigger point**: `src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic_client.go`
  - Line 82: `streamDone || strings.Contains(chunkStr, "data: [DONE]")` calls `finalizeStreamingResponse`
- **Idempotency guard**: `src/semantic-router/pkg/extproc/processor_res_body_streaming.go`
  - Line 73: `ctx.StreamingComplete` check prevents subsequent chunks from correcting the data

## Trigger Conditions

1. The upstream (e.g., gateway) sends a usage-only preview chunk at the start of the stream, in the shape of:
   ```json
   {"id":"...","object":"chat.completion.chunk","choices":[],"usage":{"prompt_tokens":37494,"completion_tokens":4,"total_tokens":37498}}
   ```
   Key characteristics: `choices` is an empty array, `usage.total_tokens > 0`, no `finish_reason`

2. The Router has an Anthropic client connected (`ClientProtocol = "anthropic"`, i.e., accessed via `/v1/messages` path)

3. The Router is configured with `api_format: openai` for upstream communication

## Upstream Response Example

First chunk (preview, usage-only, `choices: []`):

```json
[...]

PR #2846, opened 2026-08-11
[Router] fix streaming truncation when backend sends usage on every chunk

### LINKED ISSUE #2845: bug: streaming truncated to first token when backend sends usage on every chunk
### Describe the bug

With an OpenAI-compatible backend that attaches cumulative `usage` to **every** streaming chunk (observed with Baseten Model APIs), the Anthropic-facing SSE translation terminates the stream after the **first** content delta. All subsequent content is silently discarded.

Every streamed response is truncated to a single token. Non-streaming responses are unaffected, which makes this confusing to diagnose.

Because Claude Code always streams, every turn returns one token, making the router unusable with such a backend.

### To Reproduce

Backend: Baseten Model APIs, `api_format: openai`.

Upstream sends this — note `usage` on every chunk, `finish_reason` only on the last:

```
[00] delta.content='1'         finish=None   usage={prompt:79, completion:1, ...}
[01] delta.content=', 2, 3'    finish=None   usage={...}
[02] delta.content=', 4, 5'    finish=None   usage={...}
[03] delta.content=', 6, 7'    finish=None   usage={...}
[04] delta.content=', 8, 9'    finish=None   usage={...}
[05] delta.content=', 10'      finish=stop   usage={...}
[06] delta.content=None        finish=None   usage={...}
[07] data: [DONE]
```

The router emits a complete-but-empty Anthropic sequence after chunk `[00]`:

```
event: message_start
event: content_block_start
event: content_block_delta      delta.text='1'
event: content_block_stop
event: message_delta            stop_reason=end_turn  output_tokens=1
event: message_stop
```

The same request with `"stream": false` returns the full answer, confirming both the backend and the non-streaming path are fine:

| `stream` | text delivered | output_tokens |
|---|---|---|
| `false` | `1, 2, 3, 4, 5, 6, 7, 8, 9, 10` | 28 |
| `true`  | `1`                            | 1 |

### Expected behavior

The stream should continue until `finish_reason` arrives (or a usage-only trailer), delivering all content deltas. Expected `output_tokens=28` and the full text in the streaming case.

### Root cause

`src/semantic-router/pkg/anthropic/sse_out.go`

```go
func isTerminalChunk(finishReason string, chunk openai.ChatCompletionChunk) bool {
	return finishReason != "" || chunkHasUsage(chunk)
}
```

`chunkHasUsage()` is true for chunk `[00]`, so the first content delta is treated as terminal. `emitChunkEvents` then emits `content_block_stop` + `message_delta` + `message_stop` and returns `done`, so the caller stops consuming the stream.

This is a **distinct case from #2215**. That issue covered a usage-only *preview* chunk (`"choices": []`) arriving before any content, fixed with the `state.contentStarted()` guard in the `len(chunk.Choices) == 0` branch. The present bug is a **content** chunk (`choices` non-empty) that also carries usage, which never reaches that branch.

### Suggested fix

Only `finish_reason` should terminate a content chunk. Usage-only trailers still terminate via the existing no-choices branch.

```go
func isTerminalChunk(finishReason string, chunk openai.ChatCompletionChunk) bool {
	if finishReason != "" {
		return true
	}
	if len(chunk.Choices) > 0 {
		return false
	}
	return chunkHasUsage(chunk)
}
```

Behaviour matrix — only the last row changes:

| Chunk shape | Before | After |
|---|---|---|
| `finish_reason` set (± usage) | terminal | terminal |
| content only, no usage | not terminal | not terminal |
| usage-only, no choices, after content | terminal | terminal |
| usage-only, no choices, before content (#2215) | skipped | skipped |
| **content + usage, no finish_reason** | **terminal** | **not terminal** |

### Verification

- All 183 existing `pkg/anthropic` tests pass with the change.
- Live traffic against Baseten (GLM-5.2-Fast and Kimi-K2.7-Code): streaming, streaming tool calls, multi-turn, `cache_control`, and non-streaming all correct.
- 97 real Claude Code requests through the router: completion tokens went from uniformly `1` to median 83 / max 1330.

### Separate observation (not part of the fix)

`data: [DONE]` is consumed by `extractSSEDataLines` and emits nothing client-facing (`client_stream.go`). It does trigger `finalizeStreamingResponse` for internal accounting, but no `message_stop` reaches the client. So a backend that never sends `finish_reason` and never sends a usage-only trailer would hang an Anthropic client — both before and after this change. A `[DONE]` backstop may be worth considering, but that alters a deliberate design decision documented in the `sse_out.go` header comment, so it is not bundled here.

### Affected layer

global level

nandi19k:
Closes #2845

## Purpose

- **What does this PR change?** `isTerminalChunk` in the Anthropic outbound SSE emitter no longer treats a *content* chunk as terminal merely because it carries `usage`. Only `finish_reason` terminates a content chunk; usage-only trailers (`len(chunk.Choices) == 0`) still terminate via the existing no-choices branch.

- **Why is this change needed?** Backends that attach cumulative `usage` to every streaming chunk (observed: Baseten Model APIs) had every streamed response truncated to its first token. The first content delta satisfied `chunkHasUsage()`, so the emitter fired `content_block_stop` + `message_delta` + `message_stop` and stopped consuming the stream. Non-streaming was unaffected, which is what made it confusing to diagnose.

  This is distinct from #2215: that covered a usage-only *preview* chunk (`"choices": []`) before any content and was fixed with the `state.contentStarted()` guard. This bug is a **content** chunk that also carries usage, which never reaches that branch.

Behaviour matrix — only the last row changes:

| Chunk shape | Before | After |
|---|---|---|
| `finish_reason` set (± usage) | terminal | terminal |
| content only, no usage | not terminal | not terminal |
| usage-only, no choices, after content | terminal | terminal |
| usage-only, no choices, before content (#2215) | skipped | skipped |
| **content + usage, no finish_reason** | **terminal** | **not terminal** |

## Test Plan

Unit:

```bash
cd src/semantic-router
go test ./pkg/anthropic/...
go test ./pkg/anthropic/... -run TestEmitAnthropicSSEChunk -v
```

The new regression test `TestEmitAnthropicSSEChunk_UsageOnEveryContentChunk` drives three chunks — two content+usage chunks with `finish_reason: null`, then a terminal `finish_reason: "stop"` chunk — and asserts that the first two do **not** end the stream, that both text deltas reach the client, and that exactly one `message_stop` is emitted.

Pre-existing tests covering the adjacent cases, which must stay green:

- `TestEmitAnthropicSSEChunk_SplitFinishReasonAndUsage`
- `TestEmitAnthropicSSEChunk_UsageOnlyTerminal`
- `TestEmitAnthropicSSEChunk_UsageOnlyTerminalClosesThinkingAndToolBlocks`
- `TestEmitAnthropicSSEChunk_PreviewUsageChunkNotTerminal` (#2215)

Manual, against a live backend that sends per-chunk usage (Baseten Model APIs, `api_format: openai`), through Envoy + ExtProc:

1. streaming request — expect the full answer, not one token
2. streaming request with a long answer — expect many deltas
3. streaming request with `tools` — expect `tool_use` and `stop_reason: tool_use`
4. multi-turn conversation
5. `system` block with `cache_control`
6. non-streaming request — regression check
7. a second model — confirm routing is unaffected

## Test Result

- **All 183 existing `pkg/anthropic` tests pass**, 0 failures.
- New regression test passes with the fix. Without the fix it fails at the first assertion: chunk 1 (`content` + `usage`, `finish_reason: null`) returns `done == true` where the test expects `false`.
- Manual, all 7 cases correct:

  | case | deltas | output_tokens | stop_reason |
  |---|---|---|---|
  | short stream | 6 | 28 | `end_turn` |
  | long stream | 21 | 118 | `end_turn` |
  | stream + tool call | 2 | 10 | `tool_use` |
  | second model | 8 | 34 | `end_turn` |
  | multi-turn | 6 | 28 | `end_turn` |
  | system + `cache_control` | 8 | 43 | `end_turn` |
  | non-streaming | n/a | 28 | `end_turn` |

- Real client traffic: 97 requests from an Anthropic-format client (Claude Code) through the router. Completion tokens went from uniformly `1` before the fix to median 83 / max 1330 after. Routing behaviour unchanged (97 decisions across two backend models, 0 errors).

- **Follow-up risk, not addressed here:** `data: [DONE]` is consumed by `extractSSEDataLines` and emits nothing client-facing; it only triggers `finalizeStreamingResponse` for internal accounting. A backend that never sends `finish_reason` and never sends a usage-only trailer would therefore hang an Anthropic client — true both before and after this change. A `[DONE]` backstop may be worth considering, but it alters a deliberate design decision documented in the `sse_out.go` header comment, so it is deliberately out of scope.

<details>

</details>


AayushSaini101 (approved):
Good catch thanks for the fix : )

PR #2850, opened 2026-08-11
[Dashboard] key the web search rate limit on the session user

### LINKED ISSUE #2849: bug: one user can bypass the web search rate limit and block everyone
### Describe the bug

The web search endpoint has two limits. One allows 5 requests per minute per client, and it exists to stop any single person hogging the service. The other allows 30 requests per minute in total across all users, and it exists to stop DuckDuckGo banning us.

The first limit is what protects the second. Without it, one person can eat the whole shared budget.

The problem is how the code decides who a client is. It reads the `X-Forwarded-For` header and trusts it. That header is just text the caller puts in the request, so anyone can write anything in it.

So the attack is to send a different fake value every time. Every request looks like it came from a new person, so the limit of 5 never triggers. One user can then take all 30 shared slots in a few seconds, and everyone else gets rejected for the rest of the minute, including admins who have made no requests at all. Run it on a loop and the lockout never lifts.

Any logged in account can do this. The route needs the `tools.use` permission, which is granted to the lowest role, `read`. So a read only account can deny web search to admins.

### To Reproduce

Run the dashboard backend on :8700 and get a token:

```bash
TOKEN=$(curl -s -X POST http://localhost:8700/api/auth/login \
  -d '{"email":"admin@example.com","password":"<your password>"}' | jq -r .token)
```

Send `{}` as the body throughout. An empty query is rejected after the rate limit check but before any call to DuckDuckGo, so the test runs instantly and does not depend on the network. Read the codes as: `400` means the limiter let the request through, `429` means it refused.

**1. Confirm the limit works normally.** Six requests, no spoofed header:

```bash
for i in $(seq 1 6); do
  curl -s -o /dev/null -w "request $i -> %{http_code}\n" \
    -X POST http://localhost:8700/api/tools/web-search \
    -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' -d '{}'
done
```

Gives `400 400 400 400 400 429`. The sixth is refused, which is the limit doing its job.

**2. Wait 60 seconds.** The window is one minute and the shared budget carries across steps, so skip this and the later numbers come out wrong.

**3. Bypass the limit.** The same six requests, one spoofed header each:

```bash
for i in $(seq 1 6); do
  curl -s -o /dev/null -w "request $i -> %{http_code}\n" \
    -X POST http://localhost:8700/api/tools/web-search \
    -H "Authorization: Bearer $TOKEN" \
    -H "X-Forwarded-For: 10.0.0.$i" \
    -H 'Content-Type: application/json' -d '{}'
done
```

Gives six `400`s and no `429`. Same user, same session, same machine, but the limit is gone.

**4. Wait 60 seconds, then take the whole shared budget:**

```bash
sleep 60
for i in $(seq 1 30); do
  curl -s -o /dev/null -w "%{http_code} " \
    -X POST http://localhost:8700/api/tools/web-search \
    -H "Authorization: Bearer $TOKEN" \
    -H "X-Forwarded-For: 10.1.$((i/256)).$((i%256))" \
    -H 'Content-Type: application/json' -d '{}'
done; echo
```

Gives 30 `400`s. One user has now used the entire minute's budget for the whole deployment.

**5. Confirm everyone else is blocked.** Straight away, log in as a different user in a private window and run a search from the dashboard UI. It fails with a rate limit error even though that user has made no requests at all. The same happens from curl with no spoofed header:

```bash
curl -s -o /dev/null -w "innocent user -> %{http_code}\n" \
  -X POST http://localhost:8700/api/tools/web-search \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' -d '{}'
```

Gives `429`.

**6. Confirm it clears, and that it can be held open.** Wait 60 seconds and repeat step 5. It returns `400` again. Now repeat step 4 on a loop, and step 5 keeps returning `429` for as long as you keep going.

### Expected behavior

1. Changing `X-Forwarded-For` should not put a request in a fresh bucket. The sixth request in step 3 should still be refused, exactly like step 1.
2. One user should not be able to consume the entire shared budget.
3. A user who has made no requests should never be refused because of somebody else's traffic.

### Affected layer

None

### Additional context

The relevant code is `getClientIP` and `isAllowed` in `dashboard/backend/handlers/websearch.go`, plus the limit constants near the top of that file. The `tools.use` permission on the `read` role is in `dashboard/backend/auth/schema.go`.

One thing worth saying about the fix, because it may not be obvious. This route is always authenticated. Every path under `/api/` requires a session, and `/api/tools/` requires `tools.use`, so the handler always has a user ID that came from a signed session. Keying the limit on the user instead of on an address removes the problem completely, and there is no trusted proxy list to set up or get wrong. It also fixes two smaller things that are wrong today: everyone behind one office IP currently shares a single bucket of 5, and an IPv6 user has a whole range of addresses to rotate through for free.

Two related problems in the same code, small enough to fix in the same PR:

The shared budget is charged before the request body is read, so requests that never reach DuckDuckGo still spend DuckDuckGo's quota. That is also what makes the repro above work with an empty query.

The guard that caps tracked clients at 10000 refuses any client it has not seen once the table is full. A guard meant to protect memory should never deny service to someone new. It should drop an old entry instead. This cannot actually trigger today because the shared limit keeps the table at around 30 entries, but the failure mode is wrong either way.

abhinav-m22:
Closes #2849

## Purpose

The web search per-client rate limit was keyed on the `X-Forwarded-For` header, which any caller can set. Sending a different value on each request put every request in a fresh bucket, so one user could take the entire 30 per minute shared budget and lock every other user out, including admins. Any `read` role account could do it.

- Key the limit on the session user ID, which cannot be forged. The route is always authenticated, so no trusted proxy configuration is needed
- Charge the shared upstream budget only for requests that actually reach DuckDuckGo
- Make the tracked-client cap evict an old entry instead of refusing a client it has not seen

Module: `Dashboard`

## Test Result

Eight new tests in `websearch_ratelimit_test.go` pass. Verified they catch the bug rather than just passing: with the old header-trusting key temporarily restored, four of them fail, including the main bypass test.

Manual run against the fix, screenshot attached:

1. **Six requests, no spoofed header.** Baseline. Five allowed, sixth refused. Unchanged from before the fix, which confirms the limit is still enforced and was not simply switched off.
2. **Six requests, a different `X-Forwarded-For` on each.** Five allowed, sixth refused. This is the fix. Before, all six were allowed because each fake address opened a fresh bucket.
3. **Thirty requests, a different `X-Forwarded-For` on each.** Five allowed, twenty five refused. Before, all thirty were allowed and consumed the entire shared budget for the deployment.
4. **A second, real account immediately afterwards.** Allowed. Before, this returned a rate limit error even though that user had made no requests at all.

Screenshot From 2026-08-11 15-34-09

| Step | Before | After |
|---|---|---|
| 1. No spoofed header | `400 400 400 400 400 429` | `400 400 400 400 400 429` |
| 2. Spoofed header | `400` x6 | `400 400 400 400 400 429` |
| 3. Thirty spoofed | 30 allowed | 5 allowed, 25 refused |
| 4. Second user | `429` | `400` |

Step 3 also shows the upstream budget fix: those requests carry an empty query, so none of them reaches DuckDuckGo and none of them spends DuckDuckGo's quota. Previously all thirty were charged against it.

---
<details>

</details>


abhinav-m22:
@Xunzhuo, @AayushSaini101 Could you please review the PR?

AayushSaini101 (approved):
LGTM thanks : )

Issue #2761, opened 2026-08-04
bug: rust-lint CI gate scopes by changed file, not changed line, sweeping in unrelated pre-existing clippy debt


theohsiung:
### Describe the bug

The Rust lint step in the agent harness (`tools/agent/scripts/agent_support.py`'s `run_rust_lint` / `collect_changed_rust_messages`, invoked by `make agent-lint` and the "Run pre-commit hooks check file lint" CI job) already tries to scope `cargo clippy` findings to only the current change, but it does so at the **file** level, not the **line/hunk** level.

Concretely: if a PR edits any line in an existing Rust file that already has unrelated pre-existing clippy findings elsewhere in that same file, all of that file's pre-existing findings get treated as "relevant to this change" and block the PR - even though none of them are anywhere near the actual diff.

This is the Rust-side version of a gap that the Go lint step doesn't have: `golangci-lint run --new-from-rev <base>` does real hunk-aware scoping natively, but the custom Rust filtering only compares file paths, not line ranges.

Confirmed with a controlled comparison using the exact CI-pinned toolchain (`rustc`/`clippy` 1.90): running `cargo clippy --all-targets -- -D warnings ...` (the exact command the CI step runs) produces the same 266 errors on a clean, unmodified `origin/main` checkout as on a branch with real, unrelated feature changes - i.e. this is 100% pre-existing crate-wide debt, not something introduced by editing those files.

### To Reproduce

1. Edit any single line in `candle-binding/src/model_architectures/lora/bert_lora.rs` (or `bert.rs`, or `src/classifiers/lora/security_lora.rs` - all three currently carry pre-existing clippy debt at unrelated lines).
2. Run `make agent-lint CHANGED_FILES="<that file>"`, or open a PR touching that file.
3. Observe that clippy findings from lines you never touched (e.g. a pre-existing `enum` size-difference warning, or a `clone()` that could be `slice::from_ref()`) are reported as blocking, even though they're outside your actual diff.

### Expected behavior

Only clippy findings whose span overlaps the PR's actual changed line ranges should block the check - mirroring how the Go lint step already behaves via `golangci-lint`'s native `--new-from-rev` hunk-aware scoping. Pre-existing findings elsewhere in a touched file should be ignored, the same way they would be if they were in an untouched file.

### Affected layer

unsure

### Additional context

[...]

PR #2875, opened 2026-08-12
[Bindings] unify duplicated BERT_CLASSIFIER/BERT_PII_CLASSIFIER/BERT_JAILBREAK_CLASSIFIER OnceLocks

### LINKED ISSUE #2834: bug: BERT_JAILBREAK_CLASSIFIER OnceLock duplicated across ffi/init.rs and ffi/classify.rs, Traditional BERT jailbreak fallback always empty
### Describe the bug

`candle-binding/src/ffi/init.rs` and `candle-binding/src/ffi/classify.rs` each declare their own module-private `static BERT_JAILBREAK_CLASSIFIER: OnceLock<Arc<BertClassifier>>`. These are two distinct statics that happen to share a name - Rust does not merge them.

`init_jailbreak_classifier` (in `init.rs`) writes to `init.rs`'s copy when it detects a Traditional (non-LoRA) BERT model:

```rust
match BertClassifier::new(model_path, num_classes as usize, use_cpu) {
    Ok(classifier) => BERT_JAILBREAK_CLASSIFIER.set(Arc::new(classifier)).is_ok(),
    ...
}
```

But `classify_jailbreak_text` and `classify_jailbreak_text_with_probabilities` (in `classify.rs`) read from `classify.rs`'s own copy as a fallback after the LoRA path:

```rust
// Fallback to Traditional BERT classifier
if let Some(classifier) = BERT_JAILBREAK_CLASSIFIER.get() {
```

Since `classify.rs`'s static is never written to, this `.get()` always returns `None`. The Traditional BERT jailbreak fallback path is dead code today - initializing a Traditional BERT jailbreak model succeeds (`init_jailbreak_classifier` returns `true`), but every subsequent classification call silently falls through past the fallback (returning whatever `default_result`/empty behavior follows) instead of using the loaded model.

Note `LORA_JAILBREAK_CLASSIFIER` does not have this bug - it's declared `pub` once in `init.rs` and correctly imported into `classify.rs` via `use crate::ffi::init::{..., LORA_JAILBREAK_CLASSIFIER, ...};`. `BERT_CLASSIFIER` and `BERT_PII_CLASSIFIER` (intent and PII classifiers) have the identical duplication bug for the same reason - this report only covers the jailbreak case since that's what surfaced it, but the same fix pattern likely applies to those two as well.

### To Reproduce

1. Prepare a Traditional (non-LoRA) BERT jailbreak classifier model directory (no `lora_config.json`, no LoRA weights in `model.safetensors`) with at least 2 labels in `config.json`.
2. Call `init_jailbreak_classifier(model_path, num_classes, use_cpu)` - it returns `true` and detects `ModelType::Traditional`.
3. Call `classify_jailbreak_text(text)` or `classify_jailbreak_text_with_probabilities(text)`.
4. Observe the Traditional BERT fallback branch's `BERT_JAILBREAK_CLASSIFIER.get()` is always `None`, even though step 2 reported successful initialization.

### Expected behavior

`classify_jailbreak_text`/`classify_jailbreak_text_with_probabilities` should use the classifier that `init_jailbreak_classifier` actually initialized. Fix: declare `BERT_JAILBREAK_CLASSIFIER` `pub` once in `init.rs` (matching `LORA_JAILBREAK_CLASSIFIER`'s existing pattern) and import it into `classify.rs` instead of redeclaring a second static there.

### Affected layer

plugin

### Additional context

**Why this needs to be its own PR, not a quick fix folded into another change:** `init.rs` currently has ~76 pre-existing `clippy::not_unsafe_ptr_arg_deref` findings (nearly every `#[no_mangle] pub extern "C" fn` in the file dereferences a raw pointer without being marked `unsafe`) - out of 110 crate-wide. The CI Rust-lint gate scopes by changed *file*, not changed *line* (see #2761/#2762), so making even this one-line visibility change to `init.rs` sweeps in all 76 as if they were newly introduced, blocking CI.

Fixing that properly means marking ~39 FFI entry points `unsafe fn` and auditing/updating any Rust-side callers (tests, `oncelock_concurrent_test.rs`, etc.) that call them without an `unsafe` block - real work, and a much bigger, riskier change than this bug fix itself. Filing this separately so the OnceLock fix doesn't get entangled with (or blocked by) that unsafe-marking cleanup.

Discovered during review of #2759 (`@Xunzhuo`, thanks for catching it).

theohsiung:
Closes #2834

## Purpose

- `candle-binding/src/ffi/init.rs` and `candle-binding/src/ffi/classify.rs` each declared their own module-private `OnceLock<Arc<BertClassifier>>` under the names `BERT_CLASSIFIER`, `BERT_PII_CLASSIFIER`, and `BERT_JAILBREAK_CLASSIFIER`. Since these are distinct statics that happen to share a name, every read of them in `classify.rs` (`classify_text`, and the Traditional BERT fallback paths for PII/jailbreak) keyed off a copy that `init.rs`'s `init_classifier`/`init_pii_classifier`/`init_jailbreak_classifier` never wrote to - those Traditional BERT fallback paths always read an empty `OnceLock`, even after the caller reports successful initialization.
- `BERT_CLASSIFIER` was a partial exception: `classify.rs`'s own `init_generic_classifier` wrote to (and `classify_text` read from) `classify.rs`'s copy, so that specific pair was internally self-consistent - but still disconnected from `init.rs`'s `init_classifier`/`InitClassifier` (Go exposes both as distinct, separately-callable entry points for the same conceptual classifier slot).
- Fix: declare all three `pub` once in `init.rs` (matching the existing `LORA_JAILBREAK_CLASSIFIER` pattern) and import them into `classify.rs` instead of redeclaring.
- Which module(s) does this affect? `Bindings`

Discovered and deliberately deferred during review of #2759 (`@Xunzhuo`) - see #2834 for why it was filed separately rather than fixed inline there (touching `init.rs` sweeps in unrelated pre-existing clippy debt under this repo's file-scoped, not line-scoped, Rust lint gate - see #2761/#2762).

## Test Plan

- `cargo build --release --no-default-features` (candle-binding)
- `cargo test --no-default-features --lib` (candle-binding)
- `cargo fmt -- --check` (candle-binding)
- Traced every Go-side caller (`InitClassifier`/`ClassifyText`/`InitGenericClassifier` in `candle-binding/semantic-router.go`, and their production callers under `src/semantic-router/pkg/`) to characterize the actual behavior change.

## Test Result

- Build, fmt, and the FFI test suite are all clean (`ffi::init_test`, `ffi::classify_test`, `ffi::oncelock_concurrent_test` all pass; the only failures in the full `cargo test --lib` run are the same pre-existing network/GPU-gated embedding-model tests unrelated to this change).
- **No behavior change for any currently-exercised Go code path**: `InitClassifier` (`init_classifier`) and `ClassifyText` (`classify_text`) have zero production callers in `src/semantic-router` today - only `InitGenericClassifier` is called, and Traditional BERT is a rarely-hit fallback behind LoRA/ModernBERT auto-detection for jailbreak/PII (per `detect_model_type`). This closes a latent bug that would otherwise silently misbehave the moment any of these five entry points sees real production traffic, without changing today's behavior.
- Follow-up risk: `init.rs` carries ~78 pre-existing Rust-lint findings (76 `clippy::not_unsafe_ptr_arg_deref` - marking ~39 FFI entry points `unsafe fn` is real, separate work; 2 unrelated `clippy::unused_io_amount`) that this PR's changed-file list may sweep into the CI Rust-lint gate per the file-scoped (not line-scoped) issue tracked in #2761/#2762. Pushing as-is to see the actual CI outcome before deciding whether that needs its own follow-up.


WUKUNTAI-0211 (approved):
LGTM, thanks.

This correctly removes the duplicated module-local `OnceLock`s and makes the initializer and classification paths share the same `ffi::init` state for the generic, PII, and Traditional-BERT jailbreak classifiers. Marking the raw-pointer C-ABI initializers `unsafe` is also appropriate and does not change the C/cgo ABI.

Verified locally:
- `cargo test --no-default-features --lib ffi::init_test` — 29 passed
- `make test-binding-minimal`
- `make agent-validate`

Non-blocking follow-up: a regression test with a successful Traditional BERT fixture initialization followed by `classify_jailbreak_text()` would directly lock in the original #2834 failure mode.

theohsiung:
@mergifyio queue

Issue #2868, opened 2026-08-11
[Feature] Add tenant-aware entrypoint rules and caller-filtered model discovery


Xunzhuo:
## Describe the feature

Extend `entrypoints` from a static virtual-model-to-recipe mapping into a named, entrypoint-local rule table that can select a recipe from trusted request headers and the inference request path.

The same rules must be the source of truth for both caller-specific model discovery and invocation:

- `GET /v1/models` lists only entrypoint aliases visible to the current caller.
- Model-bearing inference requests resolve the requested alias, evaluate the entrypoint's local rules, and select exactly one recipe before any recipe-scoped signals or decisions run.
- A caller that may use multiple entrypoints is represented by matching rules under each entrypoint; this proposal does not add a global entrypoint grant table.
- Existing `{model_names, recipe}` entrypoints remain unconditional/public for compatibility.

The initial matcher surface should be intentionally small:

- header `Exact` matching;
- path `Exact` and segment-aware `PathPrefix` matching;
- no method, query-parameter, regex, arbitrary body, or CEL matcher in v1.

## Primary layer

global level

## Why this layer?

Entrypoint resolution is a request boundary that runs before recipe-local signals, decisions, algorithms, and plugins. It affects the canonical config contract, OpenAI-compatible discovery, ExtProc request processing, control-plane APIs, and dashboard behavior, so it should be implemented as a dedicated pre-recipe routing capability rather than hidden inside a signal or plugin.

## Why do you need this feature?

The current contract maps every entrypoint alias directly to one recipe and `/v1/models` publishes a static, unfiltered catalog. It cannot express cases such as:

- tenant A using `Auto` through recipe A;
- a specific user in tenant A using the same `Auto` alias through recipe B;
- that user also being permitted to discover and invoke a separate `Code` entrypoint.

Using one local rule table for discovery and invocation prevents the model catalog and the actual routing boundary from drifting apart. It also prevents a conditional entrypoint miss from falling through to concrete-model passthrough or the default recipe.

## Scope

- Add stable entrypoint and rule names, `rules[].matches[]`, header/path matchers, and recipe actions to the canonical config.
- Add an immutable compiled resolver with explicit `matched`, `claimed_no_match`, `unclaimed`, and `ambiguous` outcomes.
- Filter both ExtProc and router-API implementations of `GET /v1/models` through the same resolver.
- Define caller visibility projection for collection discovery and an optional `for_path` query for endpoint-specific model discovery.
[...]

PR #2884, opened 2026-08-13
[Router] Add conditional entrypoint rules for tenant/user/path-aware routing

shaunakgalvankar:
Related #2868

## Purpose

- Adds a conditional entrypoint rule system so a single virtual model name (e.g. `Auto`) can route
  to different recipes based on caller headers (tenant/user) and/or request path, instead of only
  the existing flat `{model_names, recipe}` mapping.
- Fixes an unsafe fallback: a claimed entrypoint alias whose rules don't match the caller now
  returns a hard 404, and can never silently fall through to passthrough or a default recipe.
- Config validation rejects ambiguous rule tables at load time (e.g. two rules that could both
  match the same request with neither more specific than the other); the resolver defensively
  re-checks the same specificity logic at request time and fails closed (500) if it's ever reached
  anyway.
- Also fixes two pre-existing bugs surfaced by this change:
  - `IsEntrypointModelName`/`RecipeForRequestModel` read `entrypoint.Recipe` directly, so a
    rules-based entrypoint's virtual name would have been misclassified as a concrete backend
    model and routed to passthrough — reintroducing the exact issue #2868 exists to fix.
  - `ReachableRoutingRecipes` (which drives startup resource discovery — model downloads,
    classifier provisioning) only walked `entrypoint.Recipe`, so a recipe reachable only via
    `rules[].recipe` would pass config validation but have nothing provisioned behind it at
    request time.
- Module(s): `Router`

## Scope

This is a phased Phase 1 PR against a large, multi-issue RFC (issue #2868 links 5 separate
tracking issues in its own "Related work" section). In scope:
- Config schema: `entrypoints[].rules[]` (header + path matchers, AND within a match, OR across
  matches) alongside the existing `recipe` field (mutually exclusive, validated).
- A compiled resolver (`ResolveEntrypoint`) with explicit `Matched | ClaimedNoMatch | Unclaimed |
  Ambiguous` outcomes.
- ExtProc wiring for all three inference paths (`/v1/chat/completions`, `/v1/responses`,
  `/v1/messages`) via the shared pre-routing stage, so a denied conditional entrypoint returns 404
  before any backend call.
- A new `config/config.yaml` reference example using the new schema, with reference-config
  coverage-test assertions extended to match.

Explicitly deferred to follow-up PRs (tracked by #2868's own linked issues):
- Caller-filtered `/v1/models` discovery (`for_path`/rule-aware listing).
- Management CRUD APIs for rules (`/config/router/entrypoints/*`, dry-run resolve).
- Eval API `request_context` support — the classification API has no HTTP path/headers, so a
  virtual model name behind conditional rules now correctly returns "unknown routing model" there
  rather than guessing (covered by a new regression test); extending eval with a request context
  is the natural way to lift this later.
- Dashboard/Topology/Playground, CLI/DSL/Helm/Operator translation of `rules`.
- Recipe-deletion guard's friendly `409 RECIPE_IN_USE` for a recipe used only via `rules[].recipe`
  (falls back to a still-safe `400 CONFIG_VALIDATION_ERROR`).

## Test Plan

- `go build ./...` (all packages except the pre-existing, unrelated `cmd/wasm` target, which
  requires `GOOS=js GOARCH=wasm` and fails identically on `main`).
- `go test ./pkg/config/... ./pkg/extproc/... ./pkg/publicmodels/... ./pkg/services/...
  ./pkg/apiserver/... ./pkg/modeldownload/... ./pkg/dsl/...`
- `go vet ./...`
- `golangci-lint run --new-from-rev=<base> ./...` (scoped to this PR's changed lines)
- New tests added: config-schema validation/normalization, specificity/dominance/ambiguity truth
  table, resolver (all four statuses, legacy entrypoints unaffected), canonical round-trip, ExtProc
  wiring (matched/denied/ambiguous/legacy across all three protocols, never-passthrough
  regression), and a classification-API regression proving the documented request-context
  limitation.
- Why sufficient: every existing test in the affected packages passes unmodified (legacy
  `{model_names, recipe}` entrypoints are byte-identical in behavior and error strings), so this is
  additive with a full regression net around the two bug fixes and the new denial path.

## Test Result

- All of the above pass locally: `pkg/config`, `pkg/extproc`, `pkg/publicmodels`, `pkg/services`,
  `pkg/apiserver`, `pkg/modeldownload`, `pkg/dsl` all green; `go vet` and `golangci-lint` report no
  new issues.
- Known gaps/follow-ups (not blockers): `recipes.go` crossed the 400-line structure-lint *warn*
  threshold (not the 800-line error threshold); `publicmodels`'s entrypoint description for
  conditional entrypoints is a single generic string until caller-filtered discovery lands;
  `dsl/decompiler_scopes.go` does not yet special-case rules-based entrypoints (should reject
  cleanly rather than emit a lossy `recipe: ""` — left as a follow-up).


Xunzhuo (commented):
@shaunakgalvankar this is blocked on issue intake: #2868 is still needs-acceptance. Please wait for the intake decision and, if accepted, make #2868 the implementation contract in the PR description before requesting another review.

Xunzhuo (changes requested):
Formalizing the blocker from my prior review: #2868 is still needs-acceptance and is only referenced as related. Please wait for intake and make the accepted implementation issue explicit before another review.

PR #2886, opened 2026-08-13
[Bug] Use configured embedding models for ML selectors

### LINKED ISSUE #2883: [Bug] ML-model-selection E2E aborts on embedding dimension mismatch
### Describe the bug

The `ml-model-selection` E2E profile crashes the semantic-router process when an SVM selector receives a feature vector whose dimension does not match the pretrained model. The E2E configuration declares a 1024-dimensional model-selection embedding, while the runtime selection embedding provider uses the canonical mmBERT configuration with a target dimension of 768. After appending the 14-category one-hot vector, the runtime supplies 782 features, while the Qwen3-trained selector models expect 1038 features. KMeans reports a selection error and falls back, but the SVM Rust implementation panics during `decision_function`. Because the panic crosses the C FFI boundary, the process terminates with `SIGABRT`. Subsequent E2E requests return empty HTTP 500 responses.

### To Reproduce

 ```bash
 make e2e-test \
   E2E_PROFILE=ml-model-selection \
   E2E_VERBOSE=true \
   E2E_KEEP_CLUSTER=false
 ```
Then we can get the err in semantic-router-logs.txt:

 ```
thread '<unnamed>' panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ndarray-0.15.6/src/impl_ops.rs:307:5:
called `Result::unwrap()` on an `Err` value: ShapeError/IncompatibleShape: incompatible shapes
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread '<unnamed>' panicked at library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x7f1d0d7ca2b2 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hdcfcb6d4c8489523
   1:     0x7f1d0d7e8333 - core::fmt::write::h8a494366950f23bb
   2:     0x7f1d0d7c8153 - std::io::Write::write_fmt::h6556609fca33d0b1
   3:     0x7f1d0d7ca102 - std::sys::backtrace::BacktraceLock::print::hb2a626a81e06b2dc
   4:     0x7f1d0d7cb0d3 - std::panicking::default_hook::{{closure}}::h4f78485264f12d10
   5:     0x7f1d0d7caeb5 - std::panicking::default_hook::h2c66fc99e962531d
   6:     0x7f1d0d7cbaa5 - std::panicking::rust_panic_with_hook::h33ac55f64bbd807d
   7:     0x7f1d0d7cb806 - std::panicking::begin_panic_handler::{{closure}}::h30e7cb89678a57fe
   8:     0x7f1d0d7ca7b9 - std::sys::backtrace::__rust_end_short_backtrace::hed60f27456c16ced
   9:     0x7f1d0d7cb4cd - __rustc[de2ca18b4c54d5b8]::rust_begin_unwind
  10:     0x7f1d0d7e6edd - core::panicking::panic_nounwind_fmt::h6c46f1098922b4b6
  11:     0x7f1d0d7e6f72 - core::panicking::panic_nounwind::he501508d405a4565
  12:     0x7f1d0d7e7096 - core::panicking::panic_cannot_unwind::h864cccdfd8b0af98
  13:     0x7f1d0d7a65cd - ml_svm_select
  14:          0x26897c3 - <unknown>
  15:           0x4cefc4 - <unknown>
thread caused non-unwinding panic. aborting.
SIGABRT: abort
PC=0x7f1d0d3b171c m=4 sigcode=18446744073709551610
signal arrived during cgo execution
 ```

Then the router process exits, and subsequent requests return HTTP 500 with an empty response body.

### Expected behavior

The embedding configuration used during inference should match the configuration used to train the pretrained selector models.

### Affected layer

unsure

### Additional context

This was first observed in #2738. But reproduced on a clean checkout of `upstream/main` (`fa39d0ac`)

- `global.router.model_selection.ml.embedding_dim` is set to `1024`, but the selection embedding provider actually reads the canonical global embedding configuration, which resolves to mmBERT with `target_dimension: 768`.
- Feature vectors are the embedding plus a 14-category one-hot vector: selector training used 1024 + 14 = 1038 features (Qwen3), while runtime inference currently produces 768 + 14 = 782 (mmBERT).
- This profile only runs when a PR's diff touches `e2e/profiles/ml-model-selection/**`, `src/semantic-router/pkg/modelselection/**`, `src/training/model_selection/ml_model_selection/**`, or `ml-binding/**` (`tools/agent/test-domain-registry.yaml`, `selection: pr`), and it is not part of the nightly E2E matrix. Neither of two recent PRs that touched adjacent paths actually scheduled it in CI, which likely explains why a five-month-old regression only surfaced now.

altale:
Closes #2883

## Purpose

- Allow each model-selection family to request the embedding space it requires.
- Keep general-purpose selectors such as RouterDC and Hybrid on the default semantic embedding model.
- Add `model_type` to the ML selector configuration so KNN, KMeans, SVM, and MLP can use the embedding model
associated with their pretrained artifacts.
- Configure the ML model-selection profile to use Qwen3 with 1024-dimensional embeddings, while the default semantic
embedding model remains mmBERT with 768-dimensional embeddings.
- Pass the requested model type and target dimension through the existing embedding backend instead of maintaining
separate embedding callbacks for general and ML selectors.
- Increase the `ml-model-selection` router memory limit from 8 GiB to 12 GiB because the profile now intentionally loads both embedding models: mmBERT for general-purpose selectors and Qwen3 for pretrained ML selectors. The previous 8 GiB limit caused the router container to be OOM-killed during startup/runtime.
- Update the reference configuration and ML model-selection documentation.

This fixes a dimension mismatch where pretrained ML selector artifacts expected Qwen3 1024-dimensional embeddings but
received embeddings from the default mmBERT 768-dimensional space.

## Test Plan

```bash
make e2e-test \
  E2E_PROFILE=ml-model-selection \
  E2E_VERBOSE=true \
  E2E_KEEP_CLUSTER=false
```

## Test Result

Before the fix:
```
[Test] ✗ HTTP 500 Error for question: If Real GDP = $200 billion and the price index = 200 Nominal GDP is
  Expected category: economics
  Response Headers:
  Date: Tue, 11 Aug 2026 17:10:00 GMT
  Content-Length: 0
  Response Body: 
```
After the fix:
```
================================================================================
TEST RESULTS
================================================================================
✅ PASSED - model-selection (35.257609581s)
✅ PASSED - domain-classify (12m28.042388446s)
✅ PASSED - chat-completions-request (6.946008831s)
================================================================================
Total: 3 | Passed: 3 | Failed: 0
================================================================================
[Runner] ✅ All tests passed!
```

---
<details>

</details>


altale:
Hi @AayushSaini101, could one of you take a look when you have time? The merge conflict has been resolved, and all CI checks are now green, including the `ml-model-selection` E2E profile. I’d really appreciate your time and review.  Thank you very much!

src/semantic-router/pkg/extproc/router_selection.go
```diff
@@ -334,17 +335,23 @@ func buildMLSelectionConfig(cfg *config.RouterConfig) *selection.MLSelectorConfi
 		return nil
 	}
 
+	embeddingDim := mlCfg.EmbeddingDim
+	if embeddingDim <= 0 && mlCfg.ModelType == config.EmbeddingModelTypeQwen3 {
+		embeddingDim = 1024
```
ramkrishs: This Qwen3→1024 default is now hardcoded here too, separate from selectionEmbeddingDimension above — could this call that helper instead so the two defaults can't drift apart?

ramkrishs (changes requested):
Left an inline comment above — requesting changes to make sure it doesn't get lost before merge.

altale:
@ramkrishs. Thank you so much for the review. Good catch. On closer inspection, passing 0 is preferable because it lets Candle use the model’s native dimension instead of hardcoding all Qwen3 variants to 1024.  The E2E profile still explicitly uses 1024 to match its pretrained artifacts, and all tests pass. I’ve removed the fallback accordingly. WDYT? Thanks.

ramkrishs (approved):
LGTM

AayushSaini101:
> Hi @AayushSaini101, could one of you take a look when you have time? The merge conflict has been resolved, and all CI checks are now green, including the `ml-model-selection` E2E profile. I’d really appreciate your time and review. Thank you very much!

thanks resolved : )

PR #2887, opened 2026-08-13
[Router] Carry compute budget and routing evidence into Looper execution

### LINKED ISSUE #2861: [Feature] Carry compute policy and routing evidence into Looper execution
## Goal

Carry request-scoped compute policy and immutable routing evidence into Looper execution so algorithms can make bounded, explainable decisions using existing signals, decision confidence, model pricing, and remaining budget.

## Proposed Workgroup

MoM & Routing

## Background

Current Looper requests receive candidates and static algorithm configuration but not the evidence that selected the decision or one atomic cross-stage budget ledger. Per-call token limits are insufficient because concurrent children can independently approve work against the same remaining balance.

Parent delivery area: #2336. The hard non-overridable call-amplification floor is #1456; attempt accounting is #2855; matched-budget evaluation is #2858. Active implementation is tracked in #2887.

## Work item shape

This is one shared execution-contract leaf. Routing evidence, compute policy, atomic reservation, terminal reconciliation, exhaustion behavior, Replay, and tests must land together; adaptive-N, Best-of-N, and other consumers remain separate issues.

## Scope

- Define bounded, content-minimized `RoutingEvidence` containing the selected decision confidence and qualified matched-signal facts.
- Define request-scoped `ComputeBudget` for prompt, completion, and total tokens, estimated cost, wall time, concurrency, and the non-overridable call cap from #1456.
- Define an atomic `BudgetLedger` for reserved, estimated, committed, released, and consumed resources with typed exhaustion reasons.
- Reserve worst-case aggregate work before a concurrent stage dispatches; never let multiple children approve against the same balance.
- Reconcile successful, failed, timed-out, cancelled, and partially completed attempts exactly once.
- Let algorithms query remaining budget before another stage or batch and return the best already available policy-valid result or a typed failure on exhaustion.
- Preserve existing algorithm schedules when no additional policy is configured.

Add bounded request/recipe compute-policy fields with explicit validation and precedence. Missing usage or unknown pricing follows a documented conservative policy. Request input cannot raise operator-owned limits or the hard call cap.

## Non-goals

- Duplicate or weaken #1456's total-call limit.
- Add adaptive-N, Best-of-N, or cross-algorithm chaining.
- Expose raw prompts, responses, or sensitive signal values.
- Turn per-request budgeting into global capacity scheduling.

## Completion signal

- [ ] Looper execution receives typed immutable routing evidence and compute policy.
- [ ] Aggregate work is reserved atomically before dispatch and reconciled exactly once afterward.
- [ ] Missing usage and unknown pricing use conservative documented behavior.
- [ ] Exhaustion has a stable reason and deterministic result or failure semantics.
- [ ] The hard call cap cannot be raised or bypassed.
- [ ] Replay records budget decisions and final consumption without sensitive content.

shaunakgalvankar:
## Summary

Adds three request-scoped contracts to `looper.Request` so Looper algorithms
(Confidence, Ratings, ReMoM, Fusion, Router Flow) can bound cost/latency and
carry the routing context that selected them, per #2861:

- **RoutingEvidence** — decision confidence and matched signal identifiers
  (config-defined, never prompt content), threaded through from data that
  already existed but was previously dropped before reaching Looper.
- **ComputeBudget** — optional per-decision `algorithm.budget` (token/cost/
  wall-time ceiling). Every field defaults to unlimited, so decisions that
  don't configure it are unaffected.
- **BudgetLedger** — the concurrency-safe live tracker. Every algorithm
  consults it before starting more work and records real usage after each
  call, through two shared helpers (`CheckBudget`/`RecordBudgetUsage*`)
  rather than duplicating ledger logic five times. Escalation-shaped
  algorithms (Confidence, ReMoM, Router Flow, including its tool-call resume
  path) stop early and return the best result already obtained; single-shot
  algorithms (Ratings, Fusion) check once before starting, since their
  stages aren't independently truncatable. Wall-time is enforced once, in
  `ExecuteWithLatency`.

Deliberately does not duplicate the separate, unmerged hard call-amplification
cap (#1456) — `BudgetLedger` tracks only tokens/cost/time, and checks it at
the same insertion points a future call-count check would use. Config
validation rejects `algorithm.budget` on algorithm types that don't execute
through Looper, so a budget can't be silently unenforced.

Rebased onto current `main`: `algorithm.budget` validation now delegates to
the canonical `config.IsLooperAlgorithmType`/`SupportedLooperAlgorithmTypes`
catalog rather than a local duplicate, and the Looper-construction path
matches the current `createLooper`/`looper.Factory` error-handling shape.

## Testing

- `go vet`, `gofmt`, `golangci-lint` clean (zero new findings)
- `go test -race ./pkg/looper/...` — new unit + integration tests, including
  a concurrent-ledger race test and two end-to-end tests proving a real
  HTTP round-trip is deterministically truncated once a budget is exhausted
- `go test ./pkg/config/...`, `go test ./pkg/extproc/...` — full suites pass
- Repo structure-check gate passes for every touched file
- `config/config.yaml` reference updated; doc-consistency tests pass

Closes #2861


AayushSaini101:
Thanks for updating the config @shaunakgalvankar can you add a recording for end-to-end testing according the acceptance criteria thanks : )

Xunzhuo (commented):
@shaunakgalvankar this is blocked on issue intake: #2861 is still needs-acceptance, so I cannot approve its implementation yet. The requested end-to-end recording is also still missing. Please add that evidence and wait for the intake decision before another review round.

Xunzhuo (changes requested):
Formalizing the blocker from my prior review: #2861 is still needs-acceptance, and the requested end-to-end recording is still missing. Please provide that evidence and wait for issue intake before another review.

PR #2888, opened 2026-08-14
[Test] Restore ONNX binding test compilation against the current API

iroiro147:
Update stale callsites so `go test ./...` regains compilation against current API surface.

- `IsModelInitialized()` returns `(rust, go)` states; assert both, and drop stale resets of removed globals.
- `GetEmbedding2DMatryoshka` calls include the modelType argument across single, truncated, layer, combination, consistency, normalization, and benchmark paths.
- Replace removed batch API with sequential `TestGetEmbeddingsBatch` loop and an empty-input error case; adjust `BenchmarkBatchEmbeddings` to iterate sequentially.
- `TestGetMatryoshkaConfig` now asserts via `GetEmbeddingModelsInfo` instead of a retired helper.
- `GetEmbedding` passes a max-length bound and treats the returned slice directly.

Refs #2477 — this PR is the compilation half only. The provider-independent CI lane #2477 also asks for is NOT included, so this should not auto-close the issue.


Xunzhuo (changes requested):
@iroiro147 Two blockers remain: this closes #2477 but only changes the test file—the required provider-independent CI lane is still missing. Also `_ = true` does not reset initialization state, so `InitWithEmptyPath` no longer tests a fresh invalid initialization. Please add the required CI coverage and make that lifecycle test real.

Xunzhuo (changes requested):
The updated test fixes still cannot be approved before issue intake: #2477 remains needs-acceptance. Please wait for that decision; no further code-specific blocker is being added in this review.

iroiro147:
@Xunzhuo Both points taken. Addressing the code one and correcting the scope claim; deferring on the third as you asked.

**The lifecycle test — you are right, and `_ = true` was worse than useless.** Pushed `7d639f4f`.

The original reset `initOnce` and `modelInitialized` before this subtest so it exercised a genuinely fresh invalid initialization. Both package-level vars are gone: init state now lives in Rust behind `is_mmbert_model_initialized()`, `InitMmBertEmbeddingModel` guards only with `initMu`, and the binding exposes no reset. So there was nothing left for a subtest to zero — and my placeholder quietly converted a fresh-init test into a re-init test while looking like it had handled it. That is the worse failure of the two, because it reads as covered.

Since state cannot be reset, ordering is the only honest substitute: `InitWithEmptyPath` now runs **first**, the one point in the function where the model really is uninitialized, and asserts that precondition explicitly:

```go
if r, g := IsModelInitialized(); r || g {
    t.Fatalf("precondition: model already initialized (r=%v g=%v); "+
        "this subtest must run before any successful init", r, g)
}
```

So a future reordering fails loudly instead of silently weakening the test the way my placeholder did.

If you would rather the binding grew a real teardown hook, that is a better long-term answer than relying on subtest order, and I am happy to add one — but it changes the binding's public surface, so I did not want to do it inside a test-restoration PR without asking.

**On scope — you are right that the PR overclaimed, and I have corrected it.** The body said `Fixes #2477` while changing only the test file. It now reads `Refs #2477` and states explicitly that this is the compilation half and that the provider-independent CI lane is *not* included, so it cannot auto-close the issue. That was my error, not an ambiguity in your request.

On whether the CI lane belongs in this PR or its own: my preference is separate, because restoring compilation is a prerequisite that can be reviewed on its own terms, and a CI lane I cannot exercise locally deserves its own review surface. But it is your call, and if you want them landed together I will add it here.

**On intake — understood, and I will wait.** I am not asking for review while #2477 is `needs-acceptance`. The two changes above are so the PR is correct whenever that decision lands, not to reopen the approval question.

Xunzhuo (commented):
The fresh-invalid-init test is now honest and #2477 has since been accepted. This remains only the test half of that scope, though, and lacks the provider-independent CI lane; #2515 is the current superset with the implementation, mandatory CI, and ABI receipts. Please consolidate there rather than keep a second behind implementation.

iroiro147:
@Xunzhuo Agreed — closing in favour of #2515. I checked it at code level rather than on the description.

#2515 rewrites the same `onnx-binding/semantic-router_test.go`, and adds everything this PR was missing: the provider-independent CI lane (`onnx-binding-ci.yml`, `test-and-build.yml`), the Go-side implementation, and the ABI layout receipts. It is a strict superset.

The one piece I wanted to confirm before closing was the fresh-init test, since that is the only thing this PR contributed that #2515 might have lacked. It does not lack it — **its version is better than mine.** Mine made `InitWithEmptyPath` honest by forcing it to run first and asserting the model was uninitialized, which still depends on subtest order. @TwilightTechie's is a standalone `TestInitWithEmptyPath` that relies on the FFI rejecting an empty path *before* consulting global state, and additionally asserts that a failed init does not change the initialization state. That is independent of order and checks a stronger invariant, so there is nothing here worth porting across.

Thanks for pointing me at it instead of leaving two PRs to diverge.

Issue #2793, opened 2026-08-06
ci: run Dashboard backend Go tests in the required Dashboard gate


Xunzhuo:
## Summary

Make the required Dashboard CI gate execute the existing Dashboard backend Go tests instead of checking only lint, TypeScript, frontend unit tests, and `go mod tidy`.

Parent milestone: #2791

## Background and current state

`tools/make/dashboard.mk` defines `dashboard-test-backend` as `go test ./...`, but `dashboard-check` does not depend on it. `.github/workflows/dashboard-test.yml` runs `make dashboard-check` and then builds the Dashboard. A backend regression can therefore pass the named Dashboard check without executing the backend test suite.

## Scope

- Add the backend test target to the required Dashboard check or an equivalent required CI step.
- Preserve useful failure output and dependency caching.
- Keep local and CI entrypoints aligned.
- Document when race or integration variants run separately from the fast gate.

## Acceptance criteria

- [ ] A failing Dashboard backend Go test fails the required Dashboard workflow.
- [ ] `make dashboard-check` locally exercises the same fast backend tests as CI.
- [ ] The target does not silently skip packages because of working-directory or build-tag differences.
- [ ] CI duration and logs remain visible enough to diagnose the failing package.
- [ ] Harness documentation and help text describe the updated gate.

## Likely change surfaces

- `tools/make/dashboard.mk`
- `.github/workflows/dashboard-test.yml`
- Dashboard CI documentation or harness task metadata

## Validation

Prove the gate with a controlled failing-test check in CI configuration tests or an equivalent workflow fixture, then run `make dashboard-check` and `make dashboard-build`.

PR #2889, opened 2026-08-14
[Dashboard] resolve setup mode from a single canonical source

### LINKED ISSUE #2795: bug: make Dashboard setup mode a single-source contract
## Summary

Make Dashboard setup mode resolve from one canonical source so login/bootstrap, routing, settings, and activation cannot disagree about whether first-run setup is active.

Parent milestone: #2791

## Background and current state

`GET /api/setup/state` derives `setupMode` from `setup.mode` in the loaded config through `hasSetupMode` in `dashboard/backend/handlers/setup.go`. `GET /api/settings` returns `cfg.SetupMode`, which is populated from Dashboard flags/environment in `dashboard/backend/config/config.go`. The auth service also receives `cfg.SetupMode` during startup.

Those values can diverge. A deployment can therefore show setup mode in one surface while login/bootstrap or route redirects use another value.

## Scope

- Define one canonical setup-mode owner and document precedence for legacy flags.
- Make `/api/setup/state`, `/api/settings`, bootstrap gating, authenticated route fallback, and activation use the same resolved value.
- Keep existing local first-run behavior working while deprecating the duplicate input.
- Expose a diagnostic reason when setup state cannot be read safely.

## Acceptance criteria

- [ ] All backend and frontend setup decisions use one resolved setup state.
- [ ] Conflicting legacy inputs produce a deterministic result and a visible warning.
- [ ] Activation transitions every surface out of setup mode without restart-dependent disagreement.
- [ ] Login/bootstrap cannot open solely because a stale environment value disagrees with the active config.
- [ ] Unit and E2E tests cover config-only, legacy-flag-only, conflicting, activation, and restart cases.

## Likely change surfaces

- `dashboard/backend/config/config.go`
- `dashboard/backend/handlers/setup.go`
- `dashboard/backend/handlers/settings.go`
- `dashboard/backend/router/auth_routes.go`
- `dashboard/frontend/src/contexts/SetupContext.tsx`
- setup and auth E2E tests

## Validation

Use the Dashboard backend tests, frontend setup contract tests, and the repository setup-wizard E2E profile selected by the harness.

abhinav-m22:
Closes #2795

## Purpose

Setup mode had two sources that could disagree:

- `setup.mode` in the router config, re-read from disk on every request
- the `--setup-mode` / `DASHBOARD_SETUP_MODE` flag, frozen at process start

The CLI sets the flag from the config at launch, so they match at startup and only at startup.

The flag gated `OpenBootstrapEnabled()`, which controls the unauthenticated `/api/auth/bootstrap/register` endpoint. Against an activated config, a stale env value opened that endpoint with nothing visible in the UI:

```
$ curl -s localhost:8700/api/setup/state | jq .setupMode
false                                   # UI looks normal

$ curl -s localhost:8700/api/auth/bootstrap/can-register
{"canRegister":true}                    # first-admin creation is open
```

The same frozen bool also meant the window did not close after activation, because activation restarts router and envoy but not the dashboard.

This PR resolves setup mode from one source, the config file, read live behind a small cache, and makes the auth gate check it at call time.

- Module: `Dashboard`

## What changed

- New `dashboard/backend/setupmode` package. The config file decides. The legacy flag is read only so a disagreement can be reported.
- The bootstrap gate reads the resolver at call time instead of a bool captured at startup.
- `/api/setup/state`, `/api/settings` and the three setup write endpoints now use the same resolved value.
- `/api/setup/state` gained a `reason` field for a config that cannot be read, or a stale flag.
- Startup logs the resolved state. A conflict warns once per change, not per request.
- `--setup-mode` is deprecated. It still works, it just no longer decides anything.

## Severity

Row 1 is live whenever no admin exists yet, and nothing in the UI shows it.

Row 2 (activation drift) is a **latent** open door, not a live one. By activation time an admin usually exists, so `CanBootstrap()` returns false and nobody can claim the instance. It goes live only if the user table is empty again, for example an ephemeral auth volume.

## Before / after

| # | Scenario | Before (`main`) | After (this branch) |
|---|---|---|---|
| 1 | Stale `DASHBOARD_SETUP_MODE=true`, activated config | `canRegister: true`, register returns 200 and claims the instance <br>before-1 | `canRegister: false`, register returns 403 <br>after-1 |
| 2 | Activation drift, no restart | `/api/setup/state` flips to false, `canRegister` stays true <br>before-2 | both false, no restart needed <br>after-2 |
| 3 | Config on, flag off | UI pins you to `/setup` but `canRegister` is false, dead end <br>before-3 | `canRegister: true`, first run works <br>after-3 |
| 4 | Broken config, flag true | n/a | `canRegister: false`, fails closed <br>after-4 |

## Test Plan and Result

- `go test -race` per package, all green: `setupmode` (100% statement coverage), `auth`, `router`, `handlers`. `golangci-lint` reports 0 issues. New tests cover the resolver (missing, malformed and unreadable config, cache hit, invalidation, errors never cached, conflict logging, concurrency), the bootstrap gate (closes on activation with no restart, stale flag and malformed config cannot open it, unwired service fails closed) and the handlers (`reason` omitted when clean, present on conflict, 200 on unreadable config, write endpoints gated).
- Manual: the four scenarios in the table above, run on `main` and on this branch. The 11 existing tests in `bootstrap_gate_test.go` pass unmodified; that file's diff is additions only.

<details>
<summary>go test and lint output</summary>

```
$ go test ./setupmode/... ./auth/... ./router/... -race
ok   .../dashboard/backend/setupmode    1.09s
ok   .../dashboard/backend/auth        67.94s
ok   .../dashboard/backend/router       8.40s

$ go test ./handlers/... -race -run 'Setup|Settings'
ok   .../dashboard/backend/handlers     4.15s

$ golangci-lint run ./... --config ../../tools/linter/go/.golangci.yml
0 issues.
```

</details>

## Notes for review

- **No frontend changes.** `SetupContext` already read only `/api/setup/state`, which was already the live source. Fixing the backend behind it makes the frontend correct with no edit. Zero lines under `dashboard/frontend/src`.
- **`SettingsResponse.SetupMode` is read by nothing today.** `ReadonlyContext` is the only consumer of `/api/settings` and never touches it. Kept and repointed rather than removed, since dropping a JSON field breaks callers for no gain.
- **`/api/setup/state` now returns 200 with a `reason` instead of 500** when the config cannot be read. Effective behaviour is the same, since the frontend already treated a failed fetch as "not in setup mode". The reason now travels with it. Easy to switch back if you prefer.
- **`reason` is served unauthenticated**, so it names the config file but never its contents and never the absolute path. YAML errors quote values and `*fs.PathError` embeds the path, so both are dropped or unwrapped. The full path is in the startup log.
- **Why the config file wins:** the flag is a copy of the config taken at launch that never refreshes. Requiring both was considered, but that leaves a setup-mode config with no env var unable to create its first admin.

## Out of scope

- CI gating of the backend Go tests. That is #2793, to be addressed in a separate PR.
- Three other copies of this check outside the dashboard: `start-router.sh`, `cli/bootstrap.py`, `cli/service_defaults.py`. Unifying them crosses three languages.
- Removing `--setup-mode`. The conflict warning only works while the value is still read. Removable once the CLI stops setting `DASHBOARD_SETUP_MODE`.

---

<details>

</details>


abhinav-m22:
@Xunzhuo @drivebyer @WUKUNTAI-0211 can you review this one please?

abhinav-m22:
@Xunzhuo @drivebyer @WUKUNTAI-0211 can you review this PR please?

AayushSaini101:
@wilsonwu can you take a look on this if possible thanks : )

wilsonwu:
> @wilsonwu can you take a look on this if possible thanks : )

Reviewed this against the scope and acceptance criteria of #2795. The implementation looks solid overall: setup.mode is now the canonical source, and the resolved value is consistently used by setup state, settings, bootstrap gating, and activation flows. Legacy flag conflicts are handled deterministically with useful diagnostics, and the activation flow updates all relevant surfaces without requiring a dashboard restart. The unit and frontend E2E coverage also exercise the key scenarios well.

I don’t see any blocking issues with this implementation. LGTM.

AayushSaini101 (approved):

PR #2891, opened 2026-08-14
[Fleet-Sim] fix: correct preemption bookkeeping in Instance._start_next

### LINKED ISSUE #2890: bug: fleet-sim Instance preemption leaves the admitted request queued and can double-fire a stale completion event
### Describe the bug

`Instance._start_next` (`src/fleet-sim/fleet_sim/core/instance.py`) has two related bookkeeping bugs in how it handles preemption when the KV block budget is tight:

1. It peeks `req = self._queue[0]` before running the preemption loop. That loop calls `self._queue.appendleft(victim)` for each evicted request, which shifts the queue, so by the time the function reaches the unconditional `self._queue.popleft()`, it removes whichever victim was just re-queued instead of `req`. The request actually being admitted is never dequeued: it becomes active (slot, KV blocks, a scheduled completion event) while also staying in `self._queue`. Depending on timing, this either causes a later retry to admit the same request a second time, or, if nothing else frees a slot, an infinite loop where `advance_to` keeps retrying the same still-queued request without making time progress.

2. A preempted request's original completion event is left in `self._events` (the min-heap). The only guard against it firing later is `req.preempted`, which is reset to `False` on re-admission (`_start_next`, success path). If the stale event's scheduled time is reached *after* that reset, `advance_to`'s `if req.preempted: continue` no longer skips it, and it is processed as a second, spurious completion for the same request: `active_slots`/`used_kv_blocks` get double-decremented and `total_requests`/`completed` double-count it.

### To Reproduce

Directly driving `Instance` with a `ManualProfile` (`W=1, H=0` so completion times are deterministic) and three requests where the third forces evicting the first shows the queue desync: after the admission that triggers eviction, the just-admitted request is still present in `self._queue` while also appearing in `self._active_reqs`.

```python
from fleet_sim.core.instance import Instance
from fleet_sim.core.request import Request
from fleet_sim.gpu_profiles.manual import ManualProfile

profile = ManualProfile(
    name="test", W=1.0, H=0.0, chunk=1000, blk_size=1,
    total_kv_blks=12, max_slots=100, cost_per_hr=1.0, calibration_ctx=6,
)
inst = Instance(instance_id=0, pool_id="p", gpu=profile, max_ctx=6)

a = Request(req_id=1, arrival_time=0, l_in=1, l_out=9)  # 10 blocks
b = Request(req_id=2, arrival_time=0, l_in=1, l_out=1)  # 2 blocks
c = Request(req_id=3, arrival_time=0, l_in=1, l_out=9)  # 10 blocks

for r in (a, b, c):
    inst.accept(r)

inst._start_next(0.0)  # admits a
inst._start_next(0.0)  # admits b (budget exactly full: 10 + 2 == 12)
inst._start_next(1.0)  # c's own admission evicts a

print(list(inst._queue))       # [c] on main: c never left the queue
print(inst._active_reqs)       # [b, c]: c is active *and* still queued
```

On `main`, the last `_start_next` call leaves `c` in `self._queue` while it is also in `self._active_reqs`. The second (stale-event) bug can be shown by manually re-admitting a preempted request before its original event's scheduled time and then running `advance_to`: the same request shows up twice in the returned `completed` list.

### Expected behavior

A request that has been admitted (whether or not admitting it required evicting another) should be removed from `self._queue` exactly once and never appear in both `self._queue` and `self._active_reqs` simultaneously. A request should produce exactly one completion per admission: an old event from a preemption that has since been re-admitted (once, or through several preempt/re-admit cycles) should never fire again.

### Affected layer

plugin

### Additional context

Found while reviewing the KV-budget preemption path in the fleet capacity-planning simulator (`src/fleet-sim`); no external report yet. I have a fix and a regression test ready and will open a PR referencing this issue.

AmirF194:
Closes #2890

## Purpose

`Instance._start_next` has two bookkeeping bugs when the KV block budget forces a preemption (see #2890 for the full description and a standalone repro):

1. It peeks the queue head into `req` before the preemption loop runs, but the loop's `appendleft(victim)` shifts the queue, so the later unconditional `self._queue.popleft()` removes the wrong item. The admitted request is never dequeued: it becomes active while also staying in `self._queue`.
2. A preempted request's old completion event stays in the heap. The only guard is `req.preempted`, which the re-admission path resets to `False`; if that happens before the stale event's scheduled time, it fires as a second, spurious completion, double-counting `total_requests`/`completed` and double-decrementing `active_slots`/`used_kv_blocks`.

Module: `Fleet-Sim`

## Fix

- Pop `req` from the queue before running the preemption loop, and put it back at the correct position (after the just-evicted victims) if admission still fails, instead of blindly popping whatever is now at the front.
- Add `Request.admission_epoch`, bumped on every preemption and every (re-)admission, stamped onto each `_Event` at push time. `advance_to` now drops an event whose stamped epoch no longer matches the request's current epoch, which correctly invalidates a stale event regardless of how many preempt/re-admit cycles a request has been through (the old `preempted`-flag check only handled one cycle).

## Test Plan

`tests/test_instance.py` (new) drives `_start_next` directly so each admission/eviction step can be asserted without depending on how a full simulation happens to schedule events:

- `test_admission_that_evicts_removes_the_admitted_request_from_queue`: after an admission that forces an eviction, the admitted request is not left in `self._queue`, and the evicted request is queued instead of also being active.
- `test_stale_event_from_a_preempted_admission_is_ignored_after_readmission`: a manually-injected stale event (mismatched epoch) is skipped by `advance_to`, and the fresh event for the same request still fires exactly once.

Both fail on `main` (the first with an `AssertionError`, the second with a `TypeError` since `_Event` has no `epoch` field yet) and pass on this branch.

## Test Result

- New tests pass on this branch, fail on `main`.
- Full existing suite still passes: `python -m pytest tests/` (306 passed), verified in a clean `python:3.11-slim` container, and the new/existing `tests/test_instance.py` and core suites also verified on `python:3.10-slim`.
- `ruff check` and `black --check` are clean on the changed files (the repo's `fleet_sim/__init__.py` etc. already fail `ruff check` on `main`, unrelated to this change).
- Not verified: behavior of this fix under the router/optimizer's full CDF-driven simulation loop; the fix is exercised directly against `Instance`.

### DIFF
--- a/src/fleet-sim/fleet_sim/core/instance.py
+++ b/src/fleet-sim/fleet_sim/core/instance.py
@@ -52,6 +52,7 @@ class _Event:
 
     time: float
     req: Request
+    epoch: int  # req.admission_epoch at push time; stale if it no longer matches
 
     def __lt__(self, other: _Event) -> bool:
         return self.time < other.time
@@ -189,9 +190,10 @@ class Instance:
                 while self._events and self._events[0].time <= self._now:
                     ev = heapq.heappop(self._events)
                     req = ev.req
-                    if req.preempted:
-                        # This completion event belongs to a preempted request;
-                        # the slot/blocks were already released at preemption time.
+                    if ev.epoch != req.admission_epoch:
+                        # Stale: req was preempted and/or re-admitted since this
+                        # event was scheduled, so its slot/blocks belong to a
+                        # later admission (or none, if still preempted).
                         continue
                     req.end_time = self._now
                     req.state = RequestState.DONE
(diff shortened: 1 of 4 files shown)


Xunzhuo (commented):
Thanks

Issue #2856, opened 2026-08-11
[Refactor] Unify Looper candidate-panel execution semantics


Xunzhuo:
## Goal

Extract a shared candidate-panel execution primitive for Looper-family algorithms so concurrency, ordering, timeout, quorum, cancellation, draining, accounting, and partial failure follow one explicit contract.

## Proposed Workgroup

MoM & Routing

## Background

Fusion, ReMoM, Workflows, and Ratings currently use similar but different collectors. Their cancellation ownership, ordering, quorum, late-result, and accounting behavior has drifted, making correctness fixes and future bounded algorithms harder to implement safely.

Parent: #2336. Related: amplification budget #1456 and attempt trace #2855.

## Scope and approach

- Introduce one internal panel executor with explicit concurrency, timeout/deadline, usable quorum, ordering, and per-attempt error-policy inputs.
- Accept a caller-supplied stage-specific usability predicate that distinguishes transport success from a protocol-valid result for that algorithm and stage.
- Preserve deterministic configured ordering independently of completion order.
- Define cancellation ownership, stop new dispatch promptly, and drain already-started work without goroutine or semaphore leaks.
- Return structured attempted, usable, failed, timed-out, and cancelled results with provenance and usage.
- Expose usable count, failure classes, cancellation state, and whether quorum was met; the calling algorithm applies its own explicit panel-level fallback policy.
- Migrate Fusion and ReMoM first, then Workflows and Ratings with behavior-preserving tests.

## Non-goals

- Add Best-of-N, adaptive-N, semantic deduplication, or another public algorithm.
- Change algorithm output bodies during the initial migration.
- Merge tool-trajectory state into the generic panel executor.

## Completion criteria

- [ ] Migrated algorithms share one tested concurrency, usable-quorum, and cancellation primitive.
- [ ] Ordering and documented partial-failure behavior remain compatible unless a separate bug explicitly changes them.
- [ ] Exact/over/below quorum, timeout, fail-fast, skip, cancellation, late completion, and all-failed cases are tested.
- [ ] Attempt and usage accounting covers cancelled and in-flight calls.
- [ ] No goroutine or semaphore-slot leaks occur under cancellation.
- [ ] Existing Looper microbenchmarks remain within repository thresholds.

## Validation

Run targeted Looper tests under `-race`, the existing algorithm suites, and the Looper performance benchmark gate selected by the repository harness.

## Shared-seam boundary

[...]

PR #2892, opened 2026-08-14
[Router] Unify Looper candidate-panel execution semantics for Fusion and ReMoM

shaunakgalvankar:
## Summary

Fusion, ReMoM, Workflows, and Ratings each own a separate
goroutine/semaphore/collector implementation for dispatching a panel of
concurrent model calls. Per #2856, this introduces a shared executor and
migrates Fusion and ReMoM onto it (the issue's own stated priority);
Workflows and Ratings are deferred and tracked (not dropped).

- **`RunPanel`** (`panel_executor.go`) owns concurrency limiting, quorum,
  timeout, cancellation, and deterministic index-ordered results, returning
  structured per-attempt outcomes (`success`/`failed`/`cancelled`) instead
  of the ad hoc tuples each algorithm returned before.
- **Fusion and ReMoM migrated**, each preserving its exact external
  contract (return shapes, partial-response-on-error/timeout semantics,
  quorum/usable-response counting) verified against the pre-migration code
  path by path.
- **Two deliberate behavior fixes**, both explicitly what the issue asks
  for rather than incidental changes:
  - ReMoM's result ordering is now deterministic (index-based) — it
    previously appended in completion order.
  - ReMoM's in-flight stragglers are now cancelled promptly whenever
    quorum/fail-fast triggers, in every case — previously this only
    happened as a side effect of `RoundTimeoutSeconds` being configured,
    and was a no-op otherwise.
- **One narrow, explicitly-documented difference**: in ReMoM's fail mode
  specifically, a response that aborts the round for being unusable
  content no longer contributes its own token usage to that round's
  partial accounting (fail mode is opt-in, not ReMoM's default, and the
  whole request fails regardless).
- Workflows and Ratings migration tracked as
  [TD047](tools/agent/docs/tech-debt/td-047-looper-panel-executor-workflows-ratings-migration.md)
  under the existing architecture-debt plan.

## Testing

- `go vet`, `gofmt`, `golangci-lint` clean (zero new findings)
- `go test -race ./pkg/looper/...` — full package, 3 repeated runs, all
  green, including a new `goleak`-covered test exercising every early-return
  path (quorum with/without cancellation, fail-fast, timeout, full drain)
- Existing `fusion_test.go`/`remom_test.go` suites pass unmodified; only the
  3 ReMoM tests that called the now-deleted internal collector function
  directly were ported to integration-level tests against
  `executeParallelCalls`
- `go test ./pkg/extproc/...` — full downstream regression suite passes
- Benchmarks (`BenchmarkPanel_RunPanel`, plus existing
  `BenchmarkFusion_Execute`/`BenchmarkReMoM_Execute`) confirm the executor's
  own overhead (~10-100µs) is negligible against the real HTTP cost that
  dominates end-to-end execution (milliseconds)
- Repo structure-check gate passes for every touched file

Related #2856


src/semantic-router/pkg/looper/panel_executor.go
```diff
@@ -0,0 +1,319 @@
...
+	n := len(calls)
+	if n == 0 {
+		return &PanelResult{QuorumMet: policy.MinSuccessful <= 0}
+	}
+
+	execCtx, cancel := context.WithCancel(ctx)
+	if policy.Timeout > 0 {
+		execCtx, cancel = context.WithTimeout(ctx, policy.Timeout)
+	}
+	defer cancel()
```
Xunzhuo, line 182: **Honor CancelOnReturn=false**

Impact: Context-aware stragglers are canceled after quorum or fail-fast even when callers explicitly request background completion, violating the executor contract and potentially dropping unfinished work.

Required change: Do not cancel the execution context on an early return when CancelOnReturn is false, while preserving timeout cleanup; add a context-aware regression test proving stragglers finish normally.

Xunzhuo (commented):

FAUST-BENCHOU:
Nice consolidation. The shared executor keeps the important Fusion/ReMoM semantics explicit—quorum, fail-fast, timeout, cancellation, and deterministic result ordering—without hiding the algorithm-specific error handling. The leak-safety coverage is also a good addition. LGTM from my pass.

PR #2893, opened 2026-08-14
[Router] Honor consistency_level in the router replay Milvus store

### LINKED ISSUE #2840: bug: Router replay Milvus store ignores the configured consistency_level

uuzzrm:
Closes #2840

## Purpose

- The router replay Milvus store parsed `consistency_level` into its config but never applied it: collections were created at the SDK default and every `Query` ran at the collection's level, so the field was effectively dead config.
- Every mutation on this store (`UpdateStatus`, `AttachRequest`, `AttachResponse`, `AppendOutcome`, `UpdateHallucinationStatus`, `UpdateUsageCost`, `UpdateToolTrace`) is a Get -> delete -> re-insert. With Bounded reads (the SDK default), a stale Get inside the graceful-time window can silently drop fields an earlier update already persisted.
- This is the same bug class fixed for the semantic cache in #2839, applied to the router replay consumer:
  - resolve the configured level once in `NewMilvusStore`; unset falls back to the documented `Session` default, which gives the update chain read-your-writes within the client session
  - pass it to `CreateCollection` via `client.WithConsistencyLevel`
  - pass it to every `Query` via `client.WithSearchQueryConsistencyLevel`, routed through a shared `queryOptions` helper so future call sites cannot silently drop it
  - unrecognized values log a warning instead of being silently ignored
- Affects the `Router` module.

## Test Plan

- `go build ./pkg/routerreplay/...` and `go vet ./pkg/routerreplay/...`
- `gofmt -l pkg/routerreplay/store/`
- `go test ./pkg/routerreplay/... -count=1`
- New unit tests cover the name -> `entity.ConsistencyLevel` mapping (including unset and unrecognized) and assert the level the query options actually apply, using a `ClCustomized` sentinel so "option did nothing" cannot masquerade as an explicit level.

## Test Result

- Build, vet, and gofmt clean; all router replay package tests pass.
- The sample config value `Bounded` maps to `ClBounded`; unset configs now get the documented `Session` default instead of the SDK's `Bounded`.

---
<details>

- [x] PR title uses a module-aligned prefix: `[Router]`

</details>

### DIFF
--- a/src/semantic-router/pkg/routerreplay/store/milvus.go
+++ b/src/semantic-router/pkg/routerreplay/store/milvus.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -12,6 +13,7 @@ import (
 	"github.com/milvus-io/milvus-sdk-go/v2/entity"
 
 	milvuslifecycle "github.com/vllm-project/semantic-router/src/semantic-router/pkg/milvus"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


uuzzrm:
Duplicate of #2896 (same consistency_level fix for the Milvus stores). Closing this one.

Issue #2822, opened 2026-08-08
bug: TestHybridCachePendingRequest is flaky, sleeps 100ms against a Bounded-consistency read


olifarhaan:
### Describe the bug

`TestHybridCachePendingRequest` in `pkg/cache` fails intermittently in the `test-and-build` job, on PRs that do not touch `pkg/cache` at all. Across the last 26 workflow runs the job was skipped by the path filter 10 times, and of the 15 that actually executed it, 3 failed, so roughly a 1 in 5 failure rate

Recent examples, none of which change anything under `pkg/cache`: #2530, #2507, #2810, #2818. It also passed on #2819, which is a sibling PR from the same branch point, so the outcome does not track the change under test

The failure is always the same:

```
=== RUN   TestHybridCachePendingRequest
    cache_test.go:1941: Starting TestHybridCachePendingRequest - this may take 30-60 seconds...
    cache_test.go:1984: Expected to find cached entry after update
    cache_test.go:1987: Response mismatch: got , want {"answer": "Quantum computing uses qubits..."}
--- FAIL: TestHybridCachePendingRequest (9.24s)
```

The cause is a fixed sleep against a read that is not guaranteed to be current:

```go
err = cache.UpdateWithResponse("req1", testResponse, -1)
// Wait for indexing
time.Sleep(100 * time.Millisecond)
response, found, err := cache.FindSimilar("gpt-4", testQuery)
```

`HybridCache.UpdateWithResponse` writes only to Milvus ([`hybrid_cache.go:425`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/cache/hybrid_cache.go#L425)). The in-memory HNSW index supplies the candidate, but the response body is fetched from Milvus by `fetchResponseFromCandidates` ([`hybrid_cache_lookup.go:126`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/cache/hybrid_cache_lookup.go#L126)), and the fallback path queries Milvus as well. Those reads run at Bounded consistency, which serves a guarantee timestamp behind now by the graceful time, 5s by default. On an idle runner the write has usually propagated inside 100ms and the test passes; under load it has not, and the test fails

### To Reproduce

1. Start a Milvus instance and run the job the way CI does, with `SKIP_MILVUS_TESTS=false`, as set in `.github/workflows/test-and-build.yml:232`
[...]

PR #2896, opened 2026-08-14
[Router] Honor the configured consistency_level in the Milvus semantic cache

### LINKED ISSUE #2821: bug: Milvus cache ignores the configured consistency_level and always reads at Bounded
### Describe the bug

The Milvus semantic cache never applies the configured `consistency_level`. The field is parsed into config in [`runtime_config.go:177`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/config/runtime_config.go#L177) and [`:420`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/config/runtime_config.go#L420), and the repo's own fixtures set `consistency_level: "Session"`, but no cache code path reads it. The only non-test consumer of `.ConsistencyLevel` in the tree is [`router_replay_setup.go:332`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/extproc/router_replay_setup.go#L332), which is a different subsystem

`MilvusCache` passes no consistency option when it creates the collection and none on any read, so milvus-sdk-go falls back to its default:

- `entity.DefaultConsistencyLevel = ClBounded` (`entity/collection.go:19`)
- `CreateCollection` applies that default unless `WithConsistencyLevel` is given (`client/collection.go:97`)
- `makeSearchQueryOption` starts at `ClBounded` and otherwise adopts the collection's level (`client/options.go:222-228`)

Bounded consistency serves reads at a guarantee timestamp behind now by the graceful time, 5s by default in Milvus. So an operator who sets `Strong` or `Session` silently gets Bounded, and a cache read shortly after a write can miss an entry that exists

### To Reproduce

1. Configure the semantic cache with the Milvus backend and set `search.consistency_level: "Strong"`
2. Write an entry, for example through `AddPendingRequest` followed by `UpdateWithResponse`
3. Immediately call `FindSimilar` for the same query and model
4. Observe a miss, and note that no `WithSearchQueryConsistencyLevel` reaches the client regardless of the configured value

`MilvusCache.GetByID` at [`milvus_cache_fetch.go:117`](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/cache/milvus_cache_fetch.go#L117) shows the shape: `c.client.Query(ctx, collection, partitions, expr, outputFields)` with no `SearchQueryOptionFunc` argument

### Expected behavior

The configured `consistency_level` should reach Milvus, both on collection creation and on every search and query the cache issues. An unset value should keep today's default so nothing changes for existing deployments

### Affected layer

global level

### Additional context

The user-visible symptom in production is a cache miss on an entry that was just written, so the request goes to the model again and costs a call. Nothing crashes, which is probably why this has gone unnoticed

The same staleness is also the reason `TestHybridCachePendingRequest` fails intermittently in CI, tracked separately in #2822. That test writes, sleeps a fixed 100ms, then asserts the entry is findable, and every read path for the response body goes to Milvus. Fixing the plumbing here will not by itself fix that test, since `createTestMilvusConfig` does not set `consistency_level` at all

Found while looking at CI failures on unrelated PRs, not through a production incident

uuzzrm:
Closes #2821

## Purpose

- The Milvus semantic cache parses `search.consistency_level` but never passes it to the SDK. `CreateCollection` and every search/query the cache issues (`milvusSearchSimilarVectors`, `SearchDocuments`, `GetByID`, `GetAllEntries`, `UpdateWithResponse`, `FindExact`) fall back to the SDK default (`Bounded`), so a read issued shortly after a write can miss an entry that already exists and the request is routed to the model again. The repo's own config already sets the field (canonical `config.yaml` uses `Bounded`, the stricter profile `Strong`, and the standalone `milvus.yaml` example `Session`), so the configured value is silently ignored today.
- This PR plumbs the configured level into collection creation, every search, and every query the Milvus cache issues. An empty or unrecognized level passes no option, so the SDK default stays in effect and existing deployments see no change.
- Module: Router (semantic cache backend).

## Test Plan

- `go test ./pkg/cache/ -run 'ConsistencyLevel|HonorsConsistency|KeepsSDKDefault' -count=1`: the resolver table covers case/whitespace normalization and the SDK-default fallback; a recording fake client asserts the option actually reaches `Search`, `Query`, and `CreateCollection`, using `ClCustomized` as a sentinel so an option that does nothing cannot masquerade as an explicit level.
- `go test ./pkg/cache/ -count=1` (full package suite) and `go vet ./pkg/cache/`, `gofmt` — all clean.
- Local note: this dev box has no C toolchain, so the valkey-glide dependency (cgo-only) cannot compile here; verification ran with an overlay stubbing `valkey_cache.go`/exact-`valkey` file. The changed files and the Linux CI path are unaffected.

## Test Result

- All new tests pass; the full `pkg/cache` suite still passes; `go vet` and `gofmt` report no issues.
- No follow-up risks identified. `#2822` (the intermittent `TestHybridCachePendingRequest` failure) is tracked separately and also needs `createTestMilvusConfig` to set a level, which this PR intentionally leaves alone.

### DIFF
--- a/src/semantic-router/pkg/cache/milvus_cache_admin.go
+++ b/src/semantic-router/pkg/cache/milvus_cache_admin.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"github.com/milvus-io/milvus-sdk-go/v2/client"
 	"github.com/milvus-io/milvus-sdk-go/v2/entity"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
@@ -88,6 +89,16 @@ func (c *MilvusCache) SearchDocuments(ctx context.Context, collectionName string
 		filterExpr = fmt.Sprintf("%s != \"\"", contentField)
 	}
 
+	// The cache's configured consistency level applies only to its own
+	// collection: a foreign collection (e.g. a RAG knowledge base) keeps its
+	// server-side level, which the SDK adopts when no option is passed.
+	// Overriding it here could weaken reads — a Session level degrades to
+	// Eventually on a collection this process never writes to.
+	var searchOpts []client.SearchQueryOptionFunc
+	if collectionName == c.collectionName {
+		searchOpts = c.searchQueryOptions()
+	}
+
 	// Use Milvus Search with collection-specific or default parameters
 	searchResult, err := c.client.Search(
 		ctx,
@@ -100,7 +111,7 @@ func (c *MilvusCache) SearchDocuments(ctx context.Context, collectionName string
 		entity.MetricType(actualMetricType),
 		topK,
 		searchParam,
-		c.searchQueryOptions()...,
+		searchOpts...,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("milvus search failed: %w", err)
--- a/src/semantic-router/pkg/cache/milvus_cache_admin.go
+++ b/src/semantic-router/pkg/cache/milvus_cache_admin.go
@@ -100,6 +100,7 @@ func (c *MilvusCache) SearchDocuments(ctx context.Context, collectionName string
 		entity.MetricType(actualMetricType),
 		topK,
 		searchParam,
+		c.searchQueryOptions()...,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("milvus search failed: %w", err)
--- a/src/semantic-router/pkg/cache/exact_cache_milvus.go
+++ b/src/semantic-router/pkg/cache/exact_cache_milvus.go
@@ -32,6 +32,7 @@ func (c *MilvusCache) FindExact(
 		nil,
 		expr,
 		[]string{"response_body"},
+		c.searchQueryOptions()...,
 	)
 	if err != nil {
 		return LookupResult{}, fmt.Errorf("milvus exact lookup failed: %w", err)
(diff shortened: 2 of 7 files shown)


src/semantic-router/pkg/cache/milvus_cache_admin.go
```diff
@@ -100,6 +100,7 @@ func (c *MilvusCache) SearchDocuments(ctx context.Context, collectionName string
 		entity.MetricType(actualMetricType),
 		topK,
 		searchParam,
+		c.searchQueryOptions()...,
```
drivebyer: `SearchDocuments` is the RAG retrieval entry point (`req_filter_rag_milvus.go:89`) and searches an **external knowledge-base collection**, not the cache's own collection. Passing the cache's `search.consistency_level` here changes behavior for that foreign collection:

- **Before this PR**: with no option passed, the SDK adopts the target collection's server-side level (`client/options.go:225-227` reads it from the meta cache populated by `DescribeCollection`). A KB collection created with `Strong` by its ingestion job was read at `Strong`.
- **After this PR**: the cache's level overrides it. With the `Session` value shipped in the repo's own samples (`config/runtime/response-cache/milvus.yaml:75`), the session-timestamp lookup **misses** — this process never writes to the KB collection — and `makeSearchQueryOption` falls back to `EventuallyTimestamp` (`client/options.go:241-244`). The RAG search silently degrades from Strong to Eventually, so freshly ingested documents become invisible.

Suggested fix — only pass the options when searching the cache's own collection:

```go
	var searchOpts []client.SearchQueryOptionFunc
	if collectionName == c.collectionName {
		searchOpts = c.searchQueryOptions()
	}
	searchResult, err := c.client.Search(
		...
		searchParam,
		searchOpts...,
	)
```

(needs `"github.com/milvus-io/milvus-sdk-go/v2/client"` added to this file's imports)

A ready-made commit with this change plus a `TestMilvusCacheSearchDocumentsConsistencyScope` test (asserting own-collection reads carry the option and foreign-collection reads don't) is available to cherry-pick:

```
git fetch https://github.com/vllm-project/semantic-router claude/pr-2896-review-wj8ogz
git cherry-pick a1202bc
```

---
_Generated by [Claude Code](https://claude.ai/code)_

drivebyer (commented):
The core consistency fix is solid, but the cache-level setting leaks into external RAG collections and can downgrade reads; this must be fixed before merging.

drivebyer (approved):

PR #2898, opened 2026-08-15
[Router] fix: match keyword signals against prior user messages

### LINKED ISSUE #2880: [Bug] Keyword signals only evaluate the last user message, ignoring conversation history

uuzzrm:
Closes #2880

## Purpose

- Keyword signals were only ever evaluated against the **current** user message. In a multi-turn conversation, a privacy or policy marker (e.g. "do not upload to cloud", "local model only") that appeared in an earlier user turn would silently stop matching, so the router could route a request to a cloud/frontier model even though the user had explicitly asked to keep it local.
- This change feeds the prior user messages into keyword evaluation (in conversation order, skipping empty entries) so exact-match markers keep working across turns. Single-turn requests are unaffected — no prior messages means the behavior is identical to before.
- Affected module: `Router` (classification signal dispatch).

## Test Plan

- `go test ./pkg/classification/ -run 'TestKeywordSignal' -count=1`
- The regression test `TestKeywordSignalMatchesPriorUserMessages` builds a privacy recipe, sends a conversation where only the **prior** user message contains the marker, and asserts the keyword rule matches. I also verified it fails on the unfixed code (matched rules: `[]`), so it actually guards the bug.
- `TestKeywordSignalSingleTurnUnchanged` keeps the existing single-turn path honest.
- `go test ./pkg/services/ -count=1` passes (the services layer feeds `priorUserMessages` into the same dispatcher).

## Test Result

- Focused tests pass; `go vet ./pkg/classification/` is clean; the file is gofmt-clean.
- Full-suite BM25/N-gram tests can't run on this machine (the Windows build has no CGo/Rust binding, so those classifiers fail closed at construction), which is unrelated to this change. The repo agent CI gate will run the complete matrix.

### DIFF
--- a/src/semantic-router/pkg/classification/classifier_signal_dispatch.go
+++ b/src/semantic-router/pkg/classification/classifier_signal_dispatch.go
@@ -1,6 +1,7 @@
 package classification
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
@@ -31,7 +32,13 @@ func (c *Classifier) buildSignalDispatchers(
 	dispatchers := []signalDispatch{
 		{
 			config.SignalTypeKeyword, "Keyword",
-			func() { c.evaluateKeywordSignal(results, mu, textForSignal(config.SignalTypeKeyword)) },
+			func() {
+				c.evaluateKeywordSignal(
+					results,
+					mu,
+					keywordSignalText(textForSignal(config.SignalTypeKeyword), priorUserMessages),
+				)
+			},
 		},
 		{
 			config.SignalTypeEmbedding, "Embedding",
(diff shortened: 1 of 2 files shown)


uuzzrm:
Duplicate of #2897, closing this one.

PR #2899, opened 2026-08-15
[Dashboard] Record privileged config and MCP actions in the audit log

### LINKED ISSUE #2825: [Bug] Privileged config changes are never written to the audit log
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

The dashboard has a full audit log: a table, an admin API at `/api/admin/audit-logs` gated behind `users:manage`, and a UI panel under **Users --> Audit logs**.

It only ever records user management. Config deploys, rollbacks, knowledge base edits, security policy changes and MCP server changes are not recorded at all.

There is a purpose-built `AuditMiddleware` in [middleware.go#L327](https://github.com/vllm-project/semantic-router/blob/main/dashboard/backend/auth/middleware.go#L327) that wraps a handler and writes an audit row. It is never called.

The only writers are the five `writeAudit(...)` calls in the auth package, covering user create/update/delete, password change, and bootstrap registration.

So actions that each have their own permission (`config:deploy`, `config:write`, `security:manage`) leave no trace of who performed them:

- `POST /api/router/config/deploy`
- `POST /api/router/config/rollback`
- `POST /api/router/config/update`, `/global/update`, `/global/raw/update`
- knowledge base create / update / delete
- `/api/security/` policy changes
- `/api/mcp/servers` changes

The router's own API server declares the same contract with nothing behind it. [route_policy.go#L33-L47](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/apiserver/route_policy.go#L33-L47) defines ten audit actions, and `routes_catalog.go` attaches them to 19 mutation routes - knowledge base create/update/delete, config PATCH/PUT/rollback, recipe save/delete, memory delete, vector store and file writes. `AuditAction` is only ever written onto the route struct at [route_policy.go#L67](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/apiserver/route_policy.go#L67) and never read back, and the router has no audit table, writer or log sink of any kind - the audit table exists only in the dashboard's auth database. So there is nothing for those 19 declarations to feed.

Authorisation and audit answer different questions. Authorisation asks "may this user deploy?"; audit asks "who deployed the config that broke production?" The second cannot be reconstructed after the fact if nothing wrote the row at the time.

### To Reproduce

1. Start the stack with `vllm-sr serve` and log in to the dashboard at `http://localhost:8700` as an admin.
2. Note the current audit logs
3. Do something privileged: edit a knowledge base threshold and deploy the config.
4. Re-read the audit log. The count is unchanged.
5. As a control, create a user, then re-read. The count goes up by one, confirming the log itself works.

### Expected behavior

Every request requiring a write permission should produce an audit row. Config deploy and rollback at minimum.

### Impact

### Additional context

On the dashboard side, `AuditMiddleware` already does the right thing and just needs wiring in `core_routes.go`. On the router side the hook point also exists: `wrapRouteHandler` in [middleware.go#L23](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/apiserver/middleware.go#L23) already receives the route (with its `AuditAction`), the resolved principal and the request ID, so emitting there would cover all 19 routes at once rather than wiring handlers one by one. A sink for the router still has to be decided.

One caveat: `auditResponseWriter` only implements `WriteHeader`, so wrapping a streaming handler such as `/api/router/config/nl/generate/stream` would hide the underlying `http.Flusher` and break SSE. It should forward `Flush`/`Hijack`, or streaming routes should be skipped.

Unrelated to `global.services.router_replay`, which records routing decisions per request rather than administrative actions, despite its docs describing `postgres` as "audit storage".

Happy to send a PR once the approach is settled.

##### Affected layer

None

uuzzrm:
Closes #2825

## Purpose

- The dashboard has a working audit log, but it only ever records user management. Config deploys, rollbacks, config updates, knowledge base writes, security policy changes, and MCP server changes were never recorded, so there was no way to answer "who changed the config that broke production?"
- This wires the existing (previously unused) `AuditMiddleware` around the dashboard's privileged mutation routes so every write-permission request produces an audit row.
- Module affected: `Dashboard`

## What changed

- `auth.Service.Audit(...)` — small wrapper around `AuditMiddleware` so route wiring doesn't need direct access to the store. Passes through untouched when auth is unavailable.
- `router/audit.go` — `auditMutation` wrapper that records non-read methods under a per-route action, and `kbAuditAction` which maps KB proxy methods to the router-side `knowledge_base.save` / `knowledge_base.delete` vocabulary.
- Wired routes: `/api/router/config/deploy`, `/api/router/config/rollback`, `/api/router/config/update`, `/api/router/config/global/update`, `/api/router/config/global/raw/update`, `/api/router/config/defaults/update`, `/api/router/config/kbs` (+ `/kbs/`), `/api/security/policy` (PUT), and `/api/mcp/servers` create/update/delete.
- `auditResponseWriter` now forwards `Flush` and exposes `Unwrap`, so the audit wrapper is safe to sit in front of streaming handlers in the future (the issue called this out as a caveat).
- `Server.Close` now also closes the auth and workflow stores. The previous implementation only closed the config-projection store, which leaked sqlite file handles on shutdown (and made Windows test cleanup flaky).

## Test Plan

- `go build ./...` in `dashboard/backend`
- `go vet ./auth/... ./router/...`
- `go test ./auth/... ./router/...`
- New regression tests:
  - `TestAuditMutationRecordsConfigDeploy` — authenticated POST to a wrapped route writes one audit row with the actor, action, resource, method, and status code.
  - `TestAuditMutationSkipsReads` — GET on a wrapped route writes no row, keeping the log a record of writes only.
  - `TestAuditMutationUnavailableStorePassesThrough` — nil auth service leaves the handler unwrapped instead of panicking.

## Test Result

- Build, vet, and the full `auth` + `router` test suites pass (including the previously flaky `TestBuiltInOpenClawMCPConnectsThroughInternalLoopbackRoute`, which is now stable on Windows because the stores are closed properly).
- The `handlers` and `mcp` package failures seen in a full `go test ./...` run are pre-existing on this machine (temp-dir file locks and docker-status probing) and reproduce on a clean checkout without this change.
- Follow-up: the router-side apiserver still declares audit actions on its 19 mutation routes with no sink behind them. That side needs a sink decision (as noted in the issue) and is intentionally left out of this PR.

### DIFF
--- a/dashboard/backend/auth/middleware.go
+++ b/dashboard/backend/auth/middleware.go
@@ -423,6 +423,21 @@ func (w *auditResponseWriter) WriteHeader(status int) {
 	w.ResponseWriter.WriteHeader(status)
 }
 
+// Unwrap lets http.ResponseController reach the underlying writer, so wrapped
+// handlers keep access to Flush/Hijack without losing the audit capture.
+func (w *auditResponseWriter) Unwrap() http.ResponseWriter {
+	return w.ResponseWriter
+}
+
+// Flush forwards to the underlying writer when it supports streaming, so the
+// audit wrapper can sit in front of SSE or chunked handlers without breaking
+// their flush behavior.
+func (w *auditResponseWriter) Flush() {
+	if flusher, ok := w.ResponseWriter.(http.Flusher); ok {
+		flusher.Flush()
+	}
+}
+
 func (w *auditResponseWriter) statusCodeOr200() int {
 	if w.status == 0 {
 		return http.StatusOK
--- a/dashboard/backend/auth/service.go
+++ b/dashboard/backend/auth/service.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 	"sync"
 	"time"
@@ -250,6 +251,27 @@ func (s *Service) CanBootstrap(ctx context.Context) (bool, error) {
 		return false, err
 	}
 	return count == 0, nil
+}
+
+// Audit wraps a handler so that every request it serves is recorded in the
+// audit log after the handler completes. The action and resource are recorded
+// verbatim; the authenticated user, method, path, IP, user agent and status
+// code are captured from the request/response. When the service has no store
+// (auth unavailable), the handler passes through unwrapped.
+func (s *Service) Audit(action, resource string, next http.HandlerFunc) http.HandlerFunc {
+	if s == nil || s.store == nil {
+		return next
+	}
+	return AuditMiddleware(s.store, action, resource, next)
+}
+
+// Close releases the underlying store. Safe to call when the service was never
+// backed by a store (auth unavailable).
+func (s *Service) Close() error {
+	if s == nil || s.store == nil {
+		return nil
+	}
+	return s.store.Close()
 }
 
 func defaultAdminName(name string) string {
--- a/dashboard/backend/router/audit.go
+++ b/dashboard/backend/router/audit.go
@@ -1,0 +1,52 @@
+package router
+
+import (
+	"net/http"
+
+	auth "github.com/vllm-project/semantic-router/dashboard/backend/auth"
+)
+
+// auditMutation wraps a handler so that state-changing requests are recorded
+// in the audit log. Read-only and preflight methods (GET, HEAD, OPTIONS) pass
+// through unrecorded so the log stays a record of writes; actionFor maps the
+// remaining methods to the audit action, and an empty action passes through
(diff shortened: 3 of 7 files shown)


Xunzhuo:
Thanks for taking this on. The audit wiring is directionally right, and the focused test exercises an authenticated config write. The branch now conflicts with current `main` across auth middleware and route registration after the CSRF, session, and Evaluation Plane changes. Please rebase and preserve the current write protections while resolving those conflicts; I’ll review the resolved head.

PR #2900, opened 2026-08-15
[Fleet-Sim] fix: correct preemption queue bookkeeping in Instance

### LINKED ISSUE #2890: bug: fleet-sim Instance preemption leaves the admitted request queued and can double-fire a stale completion event

uuzzrm:
Closes #2890

## Purpose

- Fixes two preemption bookkeeping bugs in `Instance._start_next` (`src/fleet-sim`) that corrupt the simulation whenever admitting a request requires evicting an active one:
  1. Queue desync: the head was peeked, then popped *after* victims were re-queued with `appendleft`, so the `popleft` removed the last victim instead of the candidate.  The admitted request stayed in the queue while also becoming active, which could double-admit it later or stall `advance_to` on the same still-queued request.
  2. Stale completion events: a preempted request's original completion event stayed in the event heap.  Once the request was re-admitted (the `preempted` flag is reset), the stale event could fire and double-decrement slots/blocks and double-count the completion.
- Also guards `advance_to` / `next_event_time` against spinning when the KV budget cannot fit even the head request (or when admitting it would just swap victims), so the clock advances to the next event instead of looping forever.
- Module: `Fleet-Sim`.

## Test Plan

- Added `src/fleet-sim/tests/test_instance.py` covering:
  - admitted-via-preemption request is removed from the queue exactly once
  - stale completion event does not complete a re-admitted request twice
  - `advance_to` returns instead of hanging when the budget cannot fit anything
  - `next_event_time` does not claim immediate service when the head cannot be admitted
  - a full preemption run drives every request to `DONE` exactly once
- Ran `pytest tests/test_instance.py` (5 passed) and the rest of the fleet-sim suite (199 passed; the 5 pre-existing failures in `test_optimizer.py` are a Windows GBK subprocess-decode issue unrelated to this change).
- Ran `ruff check` and `ruff format --check` on the changed files (clean).

## Test Result

- `pytest tests/test_instance.py -q` → `5 passed`
- `ruff check` → all checks passed; `ruff format --check` → 3 files already formatted

---
<details>

</details>

### DIFF
--- a/src/fleet-sim/fleet_sim/core/instance.py
+++ b/src/fleet-sim/fleet_sim/core/instance.py
@@ -258,23 +258,7 @@ class Instance:
         req = self._queue.popleft()
         req_blocks = math.ceil((req.l_in + req.l_out) / self.gpu.blk_size)
 
-        # ── Preempt longest active request if KV budget is tight ──────────
-        n_victims = 0
-        while (
-            self._used_kv_blocks + req_blocks > self._total_kv_blocks
-            and self._active_reqs
-        ):
-            victim = max(self._active_reqs, key=lambda r: r.l_in + r.l_out)
-            victim_blocks = math.ceil((victim.l_in + victim.l_out) / self.gpu.blk_size)
-            # Invalidate the victim's pending completion event via the preempted flag
-            victim.preempted = True
-            self._active_reqs.remove(victim)
-            self._active_slots -= 1
-            self._used_kv_blocks -= victim_blocks
-            # Re-queue the victim at head so it is retried first
-            self._queue.appendleft(victim)
-            self.total_preempted += 1
-            n_victims += 1
+        n_victims = self._preempt_for(req_blocks)
 
         # If still no room (budget too small for even 1 req), put the
         # candidate back right after the victims that were just re-queued,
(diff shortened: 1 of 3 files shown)


uuzzrm:
Looks like the same fix already landed upstream via #2891 (same `_start_next` preemption bookkeeping, same test coverage for #2890). Closing mine to avoid the duplicate — thanks!

PR #2903, opened 2026-08-15
[CLI] fix: anchor Envoy path rewrite so a v1 base_url prefix is not applied twice

### LINKED ISSUE #2885: bug: <title>Envoy path rewrite applied twice for backends whose base_url path starts with /v1 (e.g. Gemini OpenAI-compatible endpoint)
### Describe the bug

When a backend's `base_url` path starts with `/v1`, the generated Envoy `regex_rewrite` rule is applied twice, producing a corrupted upstream path and a 404 from the provider.

Google's Gemini OpenAI-compatible endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`) hits this consistently. Azure OpenAI (`/openai/v1`) is unaffected, because its path does not start with `/v1` and therefore cannot match the rewrite pattern a second time.

The generated rule looks like this:

```yaml
regex_rewrite:
  pattern:
    google_re2: {}
    regex: "^/v1(.*)$"
  substitution: "/v1beta/openai\\1"
```

The rewrite appears to run a second time after `ext_proc` injects the `x-selected-model` header and the route is recomputed. On the second pass the already-rewritten path `/v1beta/openai/chat/completions` still matches `^/v1(.*)$`, capturing `beta/openai/chat/completions`, which yields:

```
/v1beta/openai + beta/openai/chat/completions
= /v1beta/openaibeta/openai/chat/completions
```

Routing itself is correct — the router selects the right model and injects the credential properly:

```
"routing_decision" ... "original_model":"gemini-3.5-flash", "selected_model":"gemini-3.5-flash"
```

```
'authorization', 'Bearer <REDACTED>'
'x-selected-model', 'gemini-3.5-flash'
```

So this is purely a path-rewrite problem, not an auth or model-selection problem.

### Note on debugging

The Envoy access log is misleading here. Its format uses `%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%`, so it prints the correct-looking original path while the actual upstream request uses the doubled path:

```
[2026-08-13T08:08:04.304Z] "POST /v1beta/openai/chat/completions HTTP/1.1" 404 - 80 0 111 103 ...
```

The corruption is only visible in Envoy debug output (`':path'` lines).

### Environment

- vllm-sr version: 0.3.0 (installed via pip)
- Deployment target: docker (local)
- Envoy image: `envoyproxy/envoy:v1.34-latest`
- Router image: `ghcr.io/vllm-project/semantic-router/vllm-sr:latest`
- Dashboard image: `ghcr.io/vllm-project/semantic-router/dashboard:latest`
- Host OS: Ubuntu 24.04

### Workaround

Making the pattern unable to match its own output prevents the second rewrite. Requiring a path-segment boundary plus a known endpoint name works:

```yaml
regex: "^/v1/(chat|models|embeddings|completions)(.*)$"
substitution: "/v1beta/openai/\\1\\2"
```

First pass: `/v1/chat/completions` → `/v1beta/openai/chat/completions`
Second pass: `/v1beta/...` does not match `^/v1/(chat|models|...)`, so the path is left alone.

This has to be re-applied manually to `~/.vllm-sr/envoy.yaml` after every `vllm-sr serve`, since the file is regenerated on each start. When `default_model` also points at the affected backend, a second identical rewrite block is generated for the default route and both need patching.

### 1. Configure a backend whose `base_url` path starts with `/v1`

`config.yaml`:

```yaml
version: v0.3
listeners:
- name: http-8899
  address: 0.0.0.0
  port: 12002
  timeout: 300s
providers:
  defaults:
    default_model: gemini-3.5-flash
  models:
  - name: gemini-3.5-flash
    provider_model_id: gemini-3.5-flash
    backend_refs:
    - name: gemini-3.5-flash
      protocol: https
      weight: 30
      base_url: https://generativelanguage.googleapis.com/v1beta/openai
      provider: openai
      api_key: <YOUR_GEMINI_API_KEY>
  - name: gpt-5.5
    provider_model_id: gpt-5.5
    backend_refs:
    - name: gpt-5.5
      protocol: https
      weight: 30
      base_url: https://<your-resource>.openai.azure.com/openai/v1
      provider: openai
      api_key: <YOUR_AZURE_API_KEY>
routing:
  modelCards:
  - name: gemini-3.5-flash
    description: Fast general-purpose model
    capabilities:
    - general_qa
    - coding
    - quick_responses
  - name: gpt-5.5
    description: Advanced reasoning model
    capabilities:
    - reasoning
    - complex_tasks
    - analysis
  decisions:
  - name: default-route
    description: Router DC based intelligent routing
    priority: 100
    rules:
      operator: AND
    modelRefs:
    - model: gemini-3.5-flash
    - model: gpt-5.5
    algorithm:
      type: router_dc
```

### 2. Validate and start

```bash
vllm-sr validate --config config.yaml
```

Validation passes with no warnings.

### 3. Confirm the generated rewrite rules

```bash
docker exec vllm-sr-envoy-container grep -n "regex:\|substitution:" /etc/envoy/envoy.yaml
```

Output — note that the Gemini rule appears twice (once for the model route, once for the default route, because `default_model` points at it):

```
57:                      regex: "^/v1(.*)$"
58:                    substitution: "/v1beta/openai\\1"
79:                      regex: "^/v1(.*)$"
80:                    substitution: "/openai/v1\\1"
102:                      regex: "^/v1(.*)$"
103:                    substitution: "/v1beta/openai\\1"
```

### 4. Send a request to the Gemini-backed model

```bash
curl -v http://localhost:12002/v1/chat/completions \
  -d '{"model": "gemini-3.5-flash", "messages": [{"role":"user","content":"1+1?"}]}'
```

Response:

```
< HTTP/1.1 404 Not Found
< content-type: text/html
< server: envoy
< content-length: 0
< server-timing: gfet4t7; dur=36
< x-vsr-response-path: upstream
```

### 5. Inspect the actual upstream path in Envoy debug output

```bash
docker logs vllm-sr-envoy-container 2>&1 | grep "':path'" | tail -3
```

```
':path', '/v1/chat/completions'
':path', '/envoy.service.ext_proc.v3.ExternalProcessor/Process'
':path', '/v1beta/openaibeta/openai/chat/completions'
```

The last line is the corrupted upstream path.

### 6. Confirm the endpoint and credentials are fine

Calling the same endpoint directly from the host with the same API key returns 200:

```bash
curl "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions" \
  -H "Authorization: Bearer $GEMINI_API_KEY" \
  -d '{"model": "gemini-3.5-flash", "messages": [{"role":"user","content":"1+1?"}]}'
```

```
< HTTP/2 200
{"choices":[{"finish_reason":"stop", ... "model":"gemini-3.5-flash", ...}]}
```

For reference, this endpoint returns **400** (not 404) for a bad API key, a missing `Authorization` header, or a nonexistent model name — which is how the 404 was traced to the path rather than to auth or model naming:

```bash
# bad key           -> 400 {"error": {"code": 400, "message": "Please pass a valid API key", ...}}
# no auth header    -> 400
# bad model name    -> 400
```

### Note

Routing to the Azure OpenAI model in the same config works correctly throughout, since `/openai/v1` cannot match `^/v1(.*)$` on a second pass.

### Expected behavior

The upstream request should use the path built from `base_url` exactly once:

```
/v1beta/openai/chat/completions
```

and the request should succeed with a normal OpenAI-compatible response, the same as calling the provider endpoint directly.

Concretely:

- A backend whose `base_url` path begins with `/v1` should be routed correctly, without the prefix being applied twice.
- The generated rewrite should be safe to evaluate more than once — applying it to its own output should be a no-op rather than corrupting the path.
- When `default_model` points at such a backend, the extra rewrite block generated for the default route should not compound the problem.

### Suggested fix

Two options, either alone would resolve it:

1. **Generate a rewrite that cannot match its own output.** Anchoring on a path-segment boundary and the known endpoint names is enough, e.g. `^/v1/(chat|models|embeddings|completions)(.*)$` → `<base_path>/\1\2`. A more general approach would be to derive the pattern from the incoming path prefix in a way that is guaranteed idempotent for any `base_url`.

2. **Prevent the rewrite from running a second time** when the route is recomputed after `ext_proc` mutates the request headers — for example by rewriting based on `x-envoy-original-path` when it is present, rather than on the current `:path`.

### Additional suggestion (diagnostics)

The default Envoy access-log format uses `%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%`, which hides this class of bug: the log shows the correct-looking pre-rewrite path while the upstream receives something different. Logging the post-rewrite upstream path (or both) would make path-rewrite issues visible without turning on Envoy debug logging.

### Related note

Since `~/.vllm-sr/envoy.yaml` is regenerated on every `vllm-sr serve`, there is currently no supported way to persist a local fix to the generated Envoy config. An override mechanism, or the ability to supply a custom Envoy template, would help as a general escape hatch for provider-specific path quirks.

### Affected layer

None

### Additional context

_No response_

uuzzrm:
Closes #2885

## Purpose

- Backends whose base_url path starts with /v1 (Gemini's generativelanguage.googleapis.com/v1beta/openai is the one I hit) generate an Envoy regex_rewrite of ^/v1(.*) with substitution <prefix>backslash1. After ext_proc injects x-selected-model and Envoy re-runs route selection, the already-rewritten path still matches the pattern, so the rewrite fires again and the prefix ends up doubled: /v1/chat/completions -> /v1beta/openai/chat/completions -> /v1beta/openaibeta/openai/chat/completions (404 from upstream).
- Azure (/openai/v1) and other prefixes that don't start with /v1 never match a second time, which is why this only showed up with Gemini-style base URLs.
- Fix: anchor the pattern at the path-segment boundary (^/v1/(.*), substitution <prefix>/backslash1) so applying it to its own output is a no-op. This affects the model route and the default route (both emit the same block).
- Module: CLI (envoy template + config generator tests)

## Test Plan

- python -m pytest src/vllm-sr/tests/test_config_generator.py - full file, includes existing rewrite assertions plus a new regression test for a /v1-prefixed base_url
- New test also simulates applying the generated rewrite twice to confirm the second pass is a no-op
- Ran the whole src/vllm-sr/tests suite as a sanity check (547 passed; the 32 failures are pre-existing os.fchmod issues on Windows, unrelated to this change)

## Test Result

8/8 tests in test_config_generator.py pass, including the new test_backend_ref_v1_prefixed_base_url_rewrite_is_idempotent.

Verified against several real base URLs that the rewritten path is stable after a second application:
- /v1beta/openai (Gemini) - now rewritten exactly once
- /openai/v1 (Azure), /api/v1, /compatible-mode/v1 - unchanged behavior

### DIFF
--- a/src/vllm-sr/cli/templates/envoy.template.yaml
+++ b/src/vllm-sr/cli/templates/envoy.template.yaml
@@ -65,12 +65,15 @@ static_resources:
                   # Rewrite Host header to match upstream server
                   host_rewrite_literal: "{{ model.endpoints[0].host_authority }}"
                   {% if model.path_prefix %}
-                  # Prepend path prefix to all requests (e.g., /compatible-mode + /v1/chat/completions = /compatible-mode/v1/chat/completions)
+                  # Prepend path prefix to all requests (e.g., /v1/chat/completions -> /compatible-mode/v1/chat/completions).
+                  # The pattern is anchored at the path-segment boundary so re-evaluating the
+                  # route after ext_proc header injection is a no-op: the rewritten path no
+                  # longer starts with /v1/ and cannot match the pattern a second time.
                   regex_rewrite:
                     pattern:
                       google_re2: {}
-                      regex: "^/v1(.*)$"
-                    substitution: "{{ model.path_prefix }}\\1"
+                      regex: "^/v1/(.*)$"
+                    substitution: "{{ model.path_prefix }}/\\1"
                   {% endif %}
               {% endfor %}
               {% for model in anthropic_models %}
@@ -117,12 +120,15 @@ static_resources:
                   host_rewrite_literal: "{{ models[0].endpoints[0].host_authority }}"
                   {% endif %}
                   {% if models and models[0].path_prefix %}
-                  # Prepend path prefix to all requests (e.g., /compatible-mode + /v1/chat/completions = /compatible-mode/v1/chat/completions)
+                  # Prepend path prefix to all requests (e.g., /v1/chat/completions -> /compatible-mode/v1/chat/completions).
+                  # The pattern is anchored at the path-segment boundary so re-evaluating the
+                  # route after ext_proc header injection is a no-op: the rewritten path no
+                  # longer starts with /v1/ and cannot match the pattern a second time.
                   regex_rewrite:
                     pattern:
                       google_re2: {}
-                      regex: "^/v1(.*)$"
-                    substitution: "{{ models[0].path_prefix }}\\1"
+                      regex: "^/v1/(.*)$"
+                    substitution: "{{ models[0].path_prefix }}/\\1"
                   {% endif %}
           http_filters:
           {% if listener.api_keys %}
--- a/src/vllm-sr/tests/test_config_generator.py
+++ b/src/vllm-sr/tests/test_config_generator.py
@@ -1,3 +1,4 @@
+import re
 import sys
 from pathlib import Path
 
@@ -106,7 +107,8 @@ def test_backend_ref_ip_port_path_produces_correct_envoy_cluster_and_route(
 ):
     """Backend ref http://10.0.0.1:8000/v1 should split into address=10.0.0.1,
     port=8000, host_authority=10.0.0.1:8000, path_prefix=/v1, and the route
-    should use regex ^/v1(.*)$ to avoid duplicating /v1."""
+    should use the segment-anchored regex ^/v1/(.*)$ so the rewrite is
+    idempotent and /v1 is never duplicated."""
     rendered = _render_envoy_config(
         tmp_path,
         monkeypatch,
@@ -155,8 +157,8 @@ routing:
     route = _model_route(rendered, "test-model")
     route_action = route["route"]
     assert route_action["host_rewrite_literal"] == "10.0.0.1:8000"
-    assert route_action["regex_rewrite"]["pattern"]["regex"] == "^/v1(.*)$"
-    assert route_action["regex_rewrite"]["substitution"] == "/v1\\1"
+    assert route_action["regex_rewrite"]["pattern"]["regex"] == "^/v1/(.*)$"
+    assert route_action["regex_rewrite"]["substitution"] == "/v1/\\1"
 
 
 def test_provider_reliability_renders_retry_outlier_and_least_request(
@@ -279,8 +281,8 @@ routing:
     route_action = route["route"]
     # standard port 443 → host_authority should omit port
     assert route_action["host_rewrite_literal"] == "api.example.com"
-    assert route_action["regex_rewrite"]["pattern"]["regex"] == "^/v1(.*)$"
-    assert route_action["regex_rewrite"]["substitution"] == "/compatible-mode/v1\\1"
+    assert route_action["regex_rewrite"]["pattern"]["regex"] == "^/v1/(.*)$"
+    assert route_action["regex_rewrite"]["substitution"] == "/compatible-mode/v1/\\1"
 
 
 def test_backend_ref_https_base_url_uses_tls_and_explicit_extra_headers(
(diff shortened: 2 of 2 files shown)


uuzzrm:
Duplicate of #2902 - same fix for #2885, and that PR already has all checks green. Closing this one to keep the issue's PR list clean.

PR #2906, opened 2026-08-15
[CLI][E2E] Fix idempotent base URL path rewrites

### LINKED ISSUE #2885: bug: <title>Envoy path rewrite applied twice for backends whose base_url path starts with /v1 (e.g. Gemini OpenAI-compatible endpoint)

muchengl:
Part of #2885

This PR addresses the known failing pattern from #2885 where `clear_route_cache` route recomputation can apply a backend base URL path rewrite more than once, for example `/v1/chat/completions` being rewritten through a `/v1beta/openai` base path repeatedly. It does not close #2885; broader root-cause cleanup remains tracked there.
## Purpose

This PR fixes a backend base URL path rewrite regression in the `vllm-sr` generated Envoy config.

The issue appears when an OpenAI-compatible backend is configured with a `base_url` that already contains a path beginning with text that looks like `/v1`, for example:

```yaml
base_url: https://generativelanguage.googleapis.com/v1beta/openai
provider: openai
```

Before this change, the generated Envoy route used this broad rewrite matcher:

```text
^/v1(.*)$
```

That matcher also matches paths such as `/v1beta/openai/...`. When the route cache is recomputed, the already rewritten upstream path can be rewritten again, producing an invalid path like:

```text
/v1beta/openaibeta/openai/chat/completions
```

This PR changes the generated rewrite matcher to only match a complete `/v1` path segment:

```text
^/v1([/?].*)?$
```

With that matcher, `/v1/chat/completions` still rewrites to the configured backend base path, but `/v1beta/...` no longer matches as if it were the request-facing `/v1` segment.

Affected modules: `CLI` / `E2E`

### Implementation Notes

- Updates `src/vllm-sr/cli/templates/envoy.template.yaml` for both selected-model routes and the default fallback route.
- Adds config generator tests that assert the rendered Envoy regex/substitution and exercise the rewrite idempotency behavior with `re.sub`.
- Extends the CLI integration test helper so API-only smoke configs can point at a `base_url` backend instead of only an `endpoint` backend.
- Adds a minimal path-recording OpenAI-compatible mock upstream for CLI integration tests. The integration test sends a real request through `vllm-sr serve` / Envoy and verifies the path received by the mock upstream.
- Adds skipped TODO coverage for the remaining root-cause gap where a future rewrite such as `/v1/chat -> /v1/provider/chat` can still be vulnerable because the rewritten path continues to begin with the `/v1` segment.

The E2E test changes are somewhat larger than the template fix itself because they cover the real serve -> Envoy -> upstream path behavior. If maintainers prefer a lighter or more reusable test structure, I can revise this part.

### Known Follow-Up

This PR fixes the current `/v1beta/openai` double-rewrite failure, but it does not fully remove the underlying rewrite idempotency risk for every possible backend base path. The skipped TODO tests document the remaining case where the rewritten backend path itself still starts with `/v1`, such as:

```text
/v1/chat/completions -> /v1/provider/chat/completions
```

## Test Plan

- Run focused config generator tests to verify rendered Envoy route behavior.
- Run the vllm-sr CLI test suite to verify the new integration tests are discovered and skipped by default unless integration mode is enabled.
- Use the mock OpenAI-compatible upstream to keep the integration path deterministic and avoid depending on a live provider.
- Keep the known `/v1/provider/...` root-cause case as skipped TODO coverage so a future fix has an executable target.

Full live CLI integration coverage remains available through:

```bash
make vllm-sr-test-integration
```

## Test Result

```text
.venv-agent/bin/python -m pytest src/vllm-sr/tests/test_config_generator.py
8 passed, 1 skipped
```

```text
make vllm-sr-test
Ran 44 tests in 63.489s
OK (skipped=8)
Tests executed: 36
Tests skipped: 8
Failures: 0
Errors: 0
```

---
<details>

</details>

### DIFF
--- a/e2e/testing/vllm-sr-cli/README.md
+++ b/e2e/testing/vllm-sr-cli/README.md
@@ -25,7 +25,7 @@ Docker or Podman access.
 | `test_unit_serve.py` | Config bootstrap, mounts, ports, image pull policy, tokens, and read-only mode. |
 | `test_unit_lifecycle.py` | `status`, `logs`, `stop`, `dashboard`, and `config` command construction. |
 | `test_unit_runtime_topology.py` | Split-runtime discovery, cleanup, timeouts, and Docker/Podman selection. |
-| `test_integration.py` | Live health, management APIs, model visibility, sidecars, lifecycle, and pull policies. |
+| `test_integration.py` | Live health, management APIs, model visibility, path rewrites, sidecars, lifecycle, and pull policies. |
 | `cli_test_base.py` | Shared command and container helpers. |
 | `run_cli_tests.py` | Prerequisite checks, discovery, filtering, and reporting. |
 
--- a/e2e/testing/vllm-sr-cli/cli_test_base.py
+++ b/e2e/testing/vllm-sr-cli/cli_test_base.py
@@ -24,6 +24,9 @@ import yaml
 from cli.runtime_stack import DEFAULT_STACK_NAME, resolve_runtime_stack
 
 HTTP_STATUS_OK = 200
+AGENT_SMOKE_CONFIG_PATH = (
+    Path(__file__).resolve().parents[2] / "config" / "config.agent-smoke.cpu.yaml"
+)
 
 
 def _coerce_timeout_stream(value: str | bytes | None) -> str:
@@ -33,6 +36,15 @@ def _coerce_timeout_stream(value: str | bytes | None) -> str:
     if isinstance(value, bytes):
         return value.decode(errors="replace")
     return value
+
+
+def _api_only_global_config() -> dict[str, object]:
+    """Reuse the canonical smoke global block for API-only integration tests."""
+    smoke_config = yaml.safe_load(AGENT_SMOKE_CONFIG_PATH.read_text(encoding="utf-8"))
+    global_config = smoke_config.get("global")
+    if not isinstance(global_config, dict):
+        raise AssertionError(f"{AGENT_SMOKE_CONFIG_PATH} must define a global mapping")
+    return global_config
 
 
 class CLITestBase(unittest.TestCase):
@@ -298,9 +310,24 @@ class CLITestBase(unittest.TestCase):
         port: int = 8888,
         model_name: str = "test-model",
         endpoint: str = "host.docker.internal:8000",
+        base_url: str | None = None,
+        provider: str | None = None,
+        api_only: bool = False,
     ) -> str:
         """Write a minimal runnable canonical v0.3 config into the temp workspace."""
         config_path = Path(self.test_dir) / "config.yaml"
+        backend_ref: dict[str, object] = {
+            "name": "primary",
+            "weight": 100,
+        }
+        if base_url is not None:
+            backend_ref["base_url"] = base_url
+        else:
+            backend_ref["endpoint"] = endpoint
+            backend_ref["protocol"] = "http"
+        if provider is not None:
+            backend_ref["provider"] = provider
+
         config = {
             "version": "v0.3",
             "listeners": [
@@ -320,14 +347,7 @@ class CLITestBase(unittest.TestCase):
                     {
                         "name": model_name,
                         "provider_model_id": model_name,
-                        "backend_refs": [
-                            {
-                                "name": "primary",
-                                "weight": 100,
-                                "endpoint": endpoint,
-                                "protocol": "http",
-                            }
-                        ],
+                        "backend_refs": [backend_ref],
                     }
                 ],
             },
@@ -344,6 +364,8 @@ class CLITestBase(unittest.TestCase):
                 ],
             },
         }
+        if api_only:
+            config["global"] = _api_only_global_config()
         config_path.write_text(
             yaml.safe_dump(config, sort_keys=False),
             encoding="utf-8",
--- a/e2e/testing/vllm-sr-cli/mock_openai_upstream.py
+++ b/e2e/testing/vllm-sr-cli/mock_openai_upstream.py
@@ -1,0 +1,27 @@
+"""Minimal OpenAI-compatible upstream used by vllm-sr CLI integration tests."""
+
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+
+
+class Handler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        length = int(self.headers.get("Content-Length", "0"))
+        self.rfile.read(length)
+        print(self.path, flush=True)
+        body = (
+            b'{"id":"mock","object":"chat.completion",'
+            b'"model":"test-model","choices":[{"index":0,'
+            b'"message":{"role":"assistant","content":"ok"},'
+            b'"finish_reason":"stop"}]}'
+        )
+        self.send_response(200)
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+    def log_message(self, format, *args):
+        return
+
+
+ThreadingHTTPServer(("0.0.0.0", 18080), Handler).serve_forever()
--- a/e2e/testing/vllm-sr-cli/test_integration.py
+++ b/e2e/testing/vllm-sr-cli/test_integration.py
@@ -14,11 +14,17 @@ import subprocess
 import time
 import unittest
 from contextlib import contextmanager
+from pathlib import Path
 from unittest import mock
 from urllib import error as urllib_error
 from urllib import request as urllib_request
 
 from cli_test_base import HTTP_STATUS_OK, CLITestBase
+
+DEFAULT_MOCK_OPENAI_IMAGE = "ghcr.io/vllm-project/semantic-router/vllm-sr-sim:latest"
+MOCK_OPENAI_IMAGE_ENV = "VLLM_SR_SIM_IMAGE"
+MOCK_OPENAI_SERVER_PORT = 18080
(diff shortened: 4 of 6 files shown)


Xunzhuo (changes requested):
The proposed matcher fixes the `/v1beta/...` case, but it is not generally idempotent for valid backend base paths that begin with the complete `/v1/` segment.

For example, with `base_url: https://api.example.com/v1/provider`:

```text
/v1/chat/completions
→ /v1/provider/chat/completions
→ /v1/provider/provider/chat/completions
```

The skipped test `test_base_url_path_rewrite_idempotency_todo_for_v1_segment_prefix` already demonstrates this failure. Because this PR says it closes #2885 and describes the rewrite as idempotent, please either:

1. implement rewriting from an immutable original path and enable the skipped test; or
2. explicitly scope this PR as a narrow `/v1beta` mitigation, keep #2885 open, and link a follow-up issue for the general fix.

The branch also currently conflicts with `main` in `e2e/testing/vllm-sr-cli/README.md` and `e2e/testing/vllm-sr-cli/test_integration.py`, so it needs a rebase and fresh CI.

Xunzhuo (approved):
thanks @muchengl

PR #2915, opened 2026-08-17
[Router] Honor x-vsr-debug on exact response-cache hits

### LINKED ISSUE #2911: [Router] Exact response-cache hit omits x-vsr-cache-similarity under x-vsr-debug
### Describe the bug

An exact response-cache hit never emits `x-vsr-cache-similarity`, even when the
request opted into the debug surface with `x-vsr-debug: true`. A semantic hit on
the same request would emit it.

The exact-hit path sets the request-owned score correctly:

```go
// src/semantic-router/pkg/extproc/req_filter_cache_exact.go:92
ctx.VSRCacheSimilarity = 1
```

so the Replay record and telemetry are right. But it then hardcodes the demoted
values when building the immediate response, without consulting the debug flag:

```go
// src/semantic-router/pkg/extproc/req_filter_cache_exact.go:107-114
response := r.createCacheHitResponse(
    ctx,
    result.ResponseBody,
    "",                            // category
    ctx.VSRSelectedDecisionName,
    nil,                           // matchedKeywords
    0,                             // similarity
)
```

and the header is gated on a positive value:

```go
// src/semantic-router/pkg/utils/http/response.go:547
if len(similarity) > 0 && similarity[0] > 0 {
    setHeaders = append(setHeaders, &core.HeaderValueOption{
        Header: &core.HeaderValue{
            Key:      "x-vsr-cache-similarity",
            RawValue: []byte(fmt.Sprintf("%.4f", similarity[0])),
        },
    })
}
```

The semantic path instead routes the same three fields through
`cacheDetailForSurface(ctx, categoryName)` (`req_filter_cache.go:307`), which is
what actually honors `x-vsr-debug`. `CreateCacheHitResponse`'s own doc comment
states the intended contract:

> each is emitted only when non-empty, so the caller demotes them by passing
> empty values unless the request opted into `x-vsr-debug`

The exact caller demotes unconditionally, so it never satisfies the second half
of that sentence. The same hardcoding also drops the category and matched
keywords on an exact hit.

#### The Response API path has the same gap

`createCacheHitResponse` forwards the same hardcoded `0` into the Response API
branch, which applies an identical positive-value gate:

```go
// src/semantic-router/pkg/extproc/req_filter_response_api_immediate.go:331-333
if similarity > 0 {
    setHeaders = append(setHeaders, rawHeaderOption("x-vsr-cache-similarity", fmt.Sprintf("%.4f", similarity)))
}
```

So a Response API request that takes the exact path loses the header for the
same reason. A fix should cover both surfaces, not just the chat-completions one.

#### There is no later stage that recovers the header

`addDecisionDetailHeaders` does emit the header from request state:

```go
// src/semantic-router/pkg/extproc/processor_res_header_mutation.go:365-367
if ctx.VSRCacheSimilarity > 0 {
    builder.addFloat("x-vsr-cache-similarity", float64(ctx.VSRCacheSimilarity))
}
```

but that builder only runs on the upstream response-header phase. A cache hit
returns an `ImmediateResponse`, so the request never reaches it. The header is
lost for good, not merely reordered.

### To Reproduce

1. Enable the response cache with a backend that supports the exact fast path
   (in-memory is enough).
2. Set the decision's `response_cache.mode` to `exact` or `exact_then_semantic`.
   This step is required: `exactCacheEnabledForRequest`
   (`semantic_cache_scope.go:40-44`) defaults to `semantic`, and under the
   default mode the exact fast path is never entered, so the bug does not
   reproduce.
3. Send a request, let it populate the cache.
4. Replay the identical request with `x-vsr-debug: true` so it takes the exact
   path.
5. Inspect the response headers.

Observed: `x-vsr-cache-hit: true` is present, `x-vsr-cache-similarity` is
absent.

Expected: `x-vsr-cache-similarity: 1.0000`, matching what a semantic hit emits
for its own score.

The header-layer half reproduces without any router wiring, by calling
`CreateCacheHitResponse` with the exact path's literal argument list and then
with the arguments it should pass under `x-vsr-debug`:

```
exact-path args ("", nil, 0):  similarity header present=false value=""
intended args ("math", kw, 1): similarity header present=true  value="1.0000"
```

### Expected behavior

An exact hit under `x-vsr-debug: true` emits `x-vsr-cache-similarity: 1.0000`,
and the category / matched-keyword headers follow the same debug-surface rule
the semantic path uses. Without the debug opt-in the headers stay demoted, as
today. The same holds for the Response API immediate response.

### Affected layer

global level

### Additional context

Introduced with the exact-first cache path in #2813; the semantic path has
routed through `cacheDetailForSurface` since #2205.

Noticed while reviewing the #2473 follow-up. It is deliberately out of scope
there: that change is about context propagation and miss similarity, and it does
not touch `req_filter_cache_exact.go`. Filing separately so the fix stays
reviewable on its own.

A fix should also cover the missing tests: there is currently no assertion that
an exact hit plus `x-vsr-debug: true` produces the similarity header, on either
the chat-completions or the Response API surface.

WUKUNTAI-0211:
Closes #2911

## Purpose

- An exact response-cache hit sets `ctx.VSRCacheSimilarity = 1` (`req_filter_cache_exact.go:92`), so the replay record and telemetry carry the right score, but it then built the immediate response with a hardcoded empty category, `nil` matched keywords and similarity `0`. Both immediate-response surfaces gate the header on a positive value (`utils/http/response.go:547`, `req_filter_response_api_immediate.go:331`), so an exact hit never emitted `x-vsr-cache-similarity` — even when the request opted in with `x-vsr-debug: true`, while a semantic hit on the same request did. The category and matched-keyword headers were dropped the same way.
- The semantic path routes the same three fields through `cacheDetailForSurface(ctx, categoryName)` (`req_filter_cache.go:307`), which is what actually honors the opt-in. This routes the exact path through the same helper (`req_filter_cache_exact.go:108`). Without the opt-in the detail stays demoted exactly as before — no change to the lean surface.
- No later stage recovers the header: `addDecisionDetailHeaders` only runs on the upstream response-header phase, and a cache hit returns an `ImmediateResponse`, so the request never reaches it.
- Module: `Router`. Behavior is documented already — `website/docs/troubleshooting/vsr-headers.md:115` lists `x-vsr-cache-similarity` on the `debug` surface; this makes the exact path match what the docs promise, so no docs change is needed.
- Scope note: `performExactCacheLookup` is also reached from the personalized-cache path (`req_filter_cache_personalized.go:37`), so that surface now honors the opt-in too. Same contract, intentional.

## Test Plan

- `make test-semantic-router` — unit tests for the whole router, to catch regressions on the lean surface.
- Targeted: `go test -count=1 -run 'TestExactCacheHit' -v ./pkg/extproc/` with `LD_LIBRARY_PATH` set to the `candle-binding` / `ml-binding` / `nlp-binding` release dirs (as `test-semantic-router` does). `-count=1` matters — a cached result will report `ok` without running anything.
- Red/green check: revert only the `req_filter_cache_exact.go` hunk, keep the tests, and confirm the two opt-in tests fail while the no-opt-in test still passes. That is what proves the tests falsify the bug rather than just describing the new code.
- Three new tests cover the three surfaces this touches: an exact hit under `x-vsr-debug` emits the similarity, category and matched-keyword headers; an exact hit without the opt-in emits none of them; and the Response API immediate response follows the same rule.

## Test Result

- Green with the fix:

```
=== RUN   TestExactCacheHitEmitsDetailUnderDebug
--- PASS: TestExactCacheHitEmitsDetailUnderDebug (0.00s)
=== RUN   TestExactCacheHitDemotesDetailWithoutDebug
--- PASS: TestExactCacheHitDemotesDetailWithoutDebug (0.00s)
=== RUN   TestExactCacheHitEmitsDetailUnderDebugForResponseAPI
--- PASS: TestExactCacheHitEmitsDetailUnderDebugForResponseAPI (0.00s)
ok      .../pkg/extproc  0.082s
```

- Red/green verified: reverting only the `req_filter_cache_exact.go` hunk while keeping the tests turns `TestExactCacheHitEmitsDetailUnderDebug` and `TestExactCacheHitEmitsDetailUnderDebugForResponseAPI` red — both report the similarity, category and matched-keyword headers as empty, which is exactly the #2911 symptom. `TestExactCacheHitDemotesDetailWithoutDebug` passes in both directions, so it stands as the guard that the lean surface is unchanged.

- Full package: `go test -count=1 ./pkg/extproc/` → `ok 2.020s`. `gofmt -l` on both files is empty, `go vet ./pkg/extproc/` is clean.
- Gap, called out deliberately: no E2E. `e2e/testcases/cache.go` exercises the cache under the default `semantic` mode, and `exactCacheEnabledForRequest` (`semantic_cache_scope.go:40`) defaults to `semantic`, so the exact fast path is only reachable after setting a decision's `response_cache.mode` to `exact` / `exact_then_semantic`. The defect is entirely in immediate-response header construction and the unit tests pin all three surfaces at that layer, so I kept the change to unit coverage. Happy to add an exact-mode E2E config if a reviewer wants the wire-level assertion.


AayushSaini101 (approved):

WUKUNTAI-0211:
@mergifyio requeue

PR #2813, opened 2026-08-06
[Optimized]: Harden cache and context handling

Xunzhuo:
Related #2375

## Summary
- add hard-partitioned exact-first response caching across supported local and shared backends, with protocol-correct replay and zero-cost hit telemetry
- add structure-preserving route-local context compression for OpenAI and Anthropic tool results, including JSON, multimodal, RAG policy, Response API, streaming, and Looper paths
- add learning continuity and conservative upstream health, retry, load-balancing, and outlier policies

## Test plan
- [x] Full semantic-router test gate
- [x] CLI unit and integration gates
- [x] Dashboard lint, type-check, and unit tests
- [x] Changed-file pre-commit and repository fast CI gates
- [x] CPU feature gate and local stack smoke
- [x] Context-compression race and fake-upstream body-capture tests
- [x] ROCm image build, execution-provider check, targeted package tests, and live Envoy compression smoke

PR #2919, opened 2026-08-17
[Router] cache: propagate request context through cache backends

### LINKED ISSUE #2473: cache: propagate cancellation and return similarity per lookup

WUKUNTAI-0211:
Closes #2473

## The problem

Every cache backend calls its driver with `context.Background()`. On `main` today:

```
$ grep -c 'context.Background()' src/semantic-router/pkg/cache/*.go   # non-test
8 valkey_cache.go   7 redis_cache.go   7 milvus_cache.go   6 qdrant_cache.go
3 exact_cache_milvus.go   2 exact_cache_{valkey,redis,qdrant}.go   ...
```

Two consequences:

**A storage call outlives the request that issued it.** When Milvus, Redis, Qdrant or Valkey stalls, the query has no deadline and no cancellation tied to anything: the goroutine and its connection stay pinned until the driver's own timeout — if it has one — fires. One per abandoned request, with nothing reclaiming them. This is the failure mode #2222 tracks.

**A constructor that fails half-way leaks the client it dialed.** The two failure paths in `NewRedisCache` do not agree with each other:

```go
if err := cache.CheckConnection(); err != nil {
    logging.Debugf("RedisCache: failed to connect: %v", err)
    return nil, err                          // client never closed
}
if err := cache.initializeIndex(); err != nil {
    _ = redisClient.Close()                  // this path does close it
    return nil, fmt.Errorf(...)
}
```

`NewValkeyCache` has the same asymmetry. A retrying config reload against an unreachable backend accumulates one connection pool per attempt.

## What this changes

`CheckConnection`, `AddEntry`, `LookupSimilarWithThreshold`, `FindExact` and `AddExact` take a `context.Context` and hand it to the driver, across all six backends (in-memory, hybrid, Milvus, Qdrant, Redis, Valkey) plus the `windows || !cgo` stubs. The root is the ext_proc stream context (`processor_core.go`: `TraceContext: stream.Context()`), which #2813 already wired into the lookup and write paths — so the storage call is now bounded by the request that owns it rather than by nothing at all.

Both constructors release a partially built client through a `releaseOnFailure` seam, so the two failure paths agree.

## Cancellation is deliberately not symmetric

Threading a context is the mechanism; where it is honored is a decision:

- **Read path honors it.** A lookup is work still ahead of the request. An abandoned request stops the storage call instead of running to completion on a connection nobody is waiting for, and a canceled or expired context is reported as an error rather than as a fail-open miss.
- **Write path detaches from it.** The response being stored has already been computed, so honoring a client disconnect there would discard a cache fill without saving any work. Both response-cache write paths use `context.WithoutCancel`, which keeps the trace values so the write stays on the request's span.

Embedding is a synchronous CGO call and cannot be interrupted mid-flight, so cancellation there is best-effort: checked once before the embed starts and again before any state is published. The two-phase legacy methods (`AddPendingRequest` / `UpdateWithResponse`) stay context-free and forward `context.Background()` to the context-aware core — they are not on a request path, so there is no caller context to honor.

## Also in scope (remaining #2488 review items)

- **Stub alignment.** The `windows || !cgo` stubs stay assignable to `CacheBackend`, with compile-time interface assertions so the next contract change fails at the definition site instead of in a downstream build.
- **Hybrid fetch failures.** A cross-model HNSW candidate is a miss; a real Milvus fetch failure is now an error instead of hiding behind the cache's fail-open path, which both masked the outage and let an error path publish a hit-level score.
- **Miss scores aligned across all six backends.** A below-threshold miss keeps reporting its rejected candidate — that score is this request's own measurement now that it rides `LookupResult` instead of shared backend state, and it is what makes a miss diagnosable against the threshold on the debug and Replay surfaces. An error still carries no score.

## Tests

Unit (`cd src/semantic-router && go test -race ./pkg/cache/... ./pkg/extproc`):

- Barrier-controlled concurrent lookups on one backend, asserting each result carries its own score.
- Cancellation before the embed and during the embed, on both the lookup and the write path, including the assertion that a canceled write publishes no entry.
- A table over all four remote backends proving each search path surfaces the request cancellation rather than a fail-open miss.
- The detached write context surviving a canceled request while keeping its trace values, and falling back to `Background` when a request has none.
- Constructor-failure cleanup for Redis.
- Hybrid regression coverage separating a cross-model miss from a fetch failure.

E2E (`semantic-cache` testcase): the profile previously printed a hit-rate summary and returned `nil` no matter what happened underneath. It now has a verdict — priming failures are collected instead of skipped, a run with zero executed requests fails as unasserted, and each request's `x-vsr-cache-similarity` is validated (a hit must carry a score in `(0,1]`; a miss may omit the header or report its best rejected candidate in `[0,1)`). Hit rate stays a measurement rather than a gate, since `acceptance_contracts.go` already enforces a floor for it.

## Limits worth stating

- The **E2E cancellation probe pins the end-to-end contract the issue asks for, not the backend change that motivated it**: a response that never arrives writes nothing either way, so the probe would also pass before the backends took a context. The falsifiable cancellation coverage lives at unit level in `pkg/cache`. The probe runs last so it cannot disturb the hit-rate measurement, and a request that beats the cancel window is reported and skipped, so it degrades to a no-op instead of flaking when upstream answers quickly.
- **Constructor cleanup is proven end to end only on the Redis path.** Valkey's client builder rejects a refused dial before any client exists, so its test pins that instead of the release.
- The unexported `searchFn` / `queryByIDFn` / `closeClient` fields are **test seams**. No production path sets them; they exist because constructor cleanup and backend-search failure are not observable from outside a constructor that returns `(nil, err)`.
- `runBenchmarkScenario` is split into two helpers — the `USE_HNSW` override and the bounded-concurrency query loop — so the touched file passes the repository structure gate and the changed-file `funlen` budget. The extracted code is unchanged logic.

## Relationship to #2813

#2813 made the lookup result request-owned (`LookupResult`, no shared similarity state) and wired the stream context into the extproc lookup and write paths. It stopped at the extproc boundary. This PR carries that context the rest of the way into the backends, which is the half of #2473 that remains.

### DIFF
--- a/e2e/testcases/cache.go
+++ b/e2e/testcases/cache.go
@@ -4,16 +4,25 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
 
 	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
 )
+
+// cacheCancelWindow is how long the cancellation probe lets a request run before
+// abandoning it. Long enough that the router has started the ext_proc exchange,
+// short enough that a real upstream answer is unlikely.
+const cacheCancelWindow = 250 * time.Millisecond
 
 func init() {
 	pkgtestcases.Register("semantic-cache", pkgtestcases.TestCase{
@@ -38,9 +47,11 @@ type CacheResult struct {
 	OriginalQuestion string
 	SimilarQuestion  string
 	CacheHit         bool
+	Similarity       float64 // this request's score: the match on a hit, the best rejected candidate on a miss (0 when absent)
 	Error            string
 }
 
+//nolint:cyclop,gocognit // Existing E2E orchestration branches by request outcome.
 func testCache(ctx context.Context, client *kubernetes.Clientset, opts pkgtestcases.TestCaseOptions) error {
 	if opts.Verbose {
 		fmt.Println("[Test] Testing semantic cache functionality")
@@ -61,6 +72,7 @@ func testCache(ctx context.Context, client *kubernetes.Clientset, opts pkgtestca
 
 	// Run cache tests
 	var results []CacheResult
+	var setupFailures []string
 	totalRequests := 0
 	cacheHits := 0
 
@@ -69,13 +81,18 @@ func testCache(ctx context.Context, client *kubernetes.Clientset, opts pkgtestca
 		if opts.Verbose {
 			fmt.Printf("[Test] Sending original question: %s\n", testCase.OriginalQuestion)
 		}
-		_, err := sendChatRequest(ctx, testCase.OriginalQuestion, localPort, opts.Verbose)
+		origResp, err := sendChatRequest(ctx, testCase.OriginalQuestion, localPort, opts.Verbose)
 		if err != nil {
 			if opts.Verbose {
 				fmt.Printf("[Test] Error sending original question: %v\n", err)
 			}
+			// Without a priming request the similar questions cannot hit, so
+			// record it instead of silently skipping the case.
+			setupFailures = append(setupFailures,
+				fmt.Sprintf("original question %q: %v", testCase.OriginalQuestion, err))
 			continue
 		}
+		origResp.Body.Close()
 
 		// Wait a bit to ensure cache is populated
 		time.Sleep(1 * time.Second)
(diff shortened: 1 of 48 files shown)


Xunzhuo (approved):

Xunzhuo (commented):
Thanks for carrying the request context through the cache backends and tightening the E2E assertions. I found two issues worth addressing:

1. **[P1] Keep detached cache writes bounded.** `cacheWriteContext` uses `context.WithoutCancel` (and falls back to `context.Background`), which removes both cancellation and the request deadline. Cache writes are still synchronous in the response path, and Milvus/Qdrant do not add a per-operation write timeout here, so a stalled backend can still pin the ExtProc goroutine/connection indefinitely—the failure mode this PR is intended to eliminate. Please derive a bounded server-owned context, for example `context.WithTimeout(context.WithoutCancel(parent), cacheWriteTimeout)`; if the intent is also not to delay the client response, use a bounded asynchronous write path.

2. **[P2] Preserve the benchmark threshold.** `runCacheFindSimilarBench` changed from `FindSimilar` (which used the configured threshold for each backend: 0.8 for memory and normalized-equivalent 0.9 for Redis/Valkey) to `LookupSimilarWithThreshold(..., 0)`. With threshold 0, almost every KNN candidate becomes a hit, so the intended 70% hit / 30% novel-miss mix, reported hit rate, and latency comparison are no longer representative. Please keep `FindSimilar` or pass the equivalent configured threshold.

The rest of the context plumbing, constructor cleanup, Hybrid error handling, and E2E verdict changes look directionally sound.

Issue #2924, opened 2026-08-17
[Feature] Enforce bounded request byte-size limits on the http_classify connector


theohsiung:
### Describe the feature

#2759's `http_classify` connector (`http_classifier.go`) reads the full response body with `io.ReadAll(resp.Body)` with no size cap. RFC #2779's Phase 1 scope explicitly lists "generalize the current direct `/classify` connector with bounded I/O" and "HTTP classifier boundary tests reject request/response byte and batch limits at configured N+1 and assert deadline cancellation" as exit criteria.

Scope for this issue: enforce a configurable request/response byte-size limit (reject with a clear error above it) and add a boundary test asserting rejection at the configured limit + 1 byte, plus a test asserting the existing deadline/cancellation behavior actually cancels in-flight requests.

Explicitly out of scope: batch-size limits. The current wire contract (`httpClassifyRequest{Inputs: text}`) is single-string only — there is no batch dimension to bound. Adding batch limits first requires adding batching, which is separate, larger work; this issue only leaves a comment noting where that limit would apply once batching exists.

### Primary layer

global level

### Why this layer?

Same rationale as #2759/#2760: this is router-owned connector config/behavior, not a routing-rule concern.

### Why do you need this feature?

Closes a concrete gap between the merged #2759 connector and RFC #2779 Phase 1's explicit exit criteria for that same connector, without taking on unrelated scope (batching) to do it.

### Additional context

Loosely related to #2760 (touches the same file as the prerequisite label-mapping refactor) but not itself part of #2760's category/complexity/PII scope — filed as a standalone, cross-referenced issue.

guan404ming:
Hi @theohsiung @Xunzhuo I just open [pr](https://github.com/vllm-project/semantic-router/pull/3093) to resolve this issue please take a look, thanks!

theohsiung:
Cool thx for your awesome work!

Issue #2920, opened 2026-08-17
[Feature] Add a shared remote-backend block for built-in classifier signals


theohsiung:
### Goal

Let built-in classifier signals declare a common local-or-remote backend block, beginning with category, without duplicating transport and lifecycle fields.

### Scope

- Define the shared backend reference and validation.
- Migrate category first while preserving current local behavior.
- Reuse typed deployment and connector contracts from #2782.
- Specify unavailable and incompatible-backend behavior.

### Non-goals

- A universal observation registry.
- Routing decisions or fallback-model selection.
- Signal-specific result contracts covered by sibling issues.

Parent Epic: #2760

theohsiung:
Note (superseded below): #2918 originally left the HTTP transport type (`HTTPClassifierJailbreakInference`) jailbreak-specific, only generalizing the label-validation logic — meaning this issue would likely have needed to copy-paste that type under a new name. See the next comment: this got resolved before #2918 merged, so there's nothing left to decide here on that front.

theohsiung:
Update: this is now done in #2918/#2930 — `HTTPClassifierJailbreakInference` is generalized into `HTTPClassifierInference` (mapping field is `sequenceLabelMapping`), validated against a real external model (`LLM-Semantic-Router/category_classifier_modernbert-base_model`) before merging. This issue can construct `HTTPClassifierInference` directly — no transport-type decision left to make here. What's still this issue's job: `CategoryModel`'s own `Variant`/`Protocol` config fields, its backend-selection validator, and the dependency-construction wiring (mirroring `buildJailbreakDependencies`).

theohsiung:
**Superseded — error policy is now owned by #3106.**

When I wrote this, `CategoryModel.OnError` looked like a per-signal field this issue should add, mirroring `PromptGuardConfig.OnError`. Since then #3106 was opened to define backend error policy for classifier signals as a whole; it is `accepted` and being implemented in #3145 with a different mechanism — a three-valued `Unknown` propagated through the rule engine plus a configurable `on_unknown`, rather than a per-module `on_error` bolted onto each consumer.

The distinction matters: anything that resolves the failure at the leaf collapses to a bool before `NOT`/`AND`/`OR` see it, which is the fail-closed-becomes-fail-open case discussed on #3106. So error policy for this signal should follow #3106 rather than be settled here, and this issue should stay scoped to the backend block itself.

Original text below for context.

---

[...]

Issue #2760, opened 2026-08-04
[Epic] Unify built-in classifier backend and result contracts


theohsiung:
### Goal

Move built-in classifier signals onto shared, versioned backend and result contracts while preserving signal-specific semantics.

### Scope

- Add a shared remote-backend block for built-in classifier signals.
- Define versioned score and token-span result contracts.
- Migrate category, complexity, PII, hallucination, and prompt-guard integrations incrementally.
- Keep configuration validation and unavailable behavior consistent.

### Child issues

- #2920
- #2921
- #2922
- #2928
- #3126

### Non-goals

- A universal routing observation registry.
- Recipe decisions, decision trees, or safe-model selection.
- A shared HTTP connector implementation; that belongs to #2782.

Parent Workgroup: #2966

adaamko:
Hey, reviewed the scope, i have a few notes on contracts and a bigger point.

1. Spans: worth stating the offset unit explicitly.  #2779 defines `token_classify` / `token_spans.v1` for this shape — one contract is better than two.

`TokenEntity` already means character positions in the original text. HTTP contract must say so, or providers guess. Needs stating:

- offsets are character offsets into the exact request string — not token indices, not offsets into a normalized or truncated copy;
- counting unit: code points or UTF-8 bytes;
- overlapping and nested spans allowed or not (`PERSON` inside `ADDRESS`);
- truncation past model context: dropped spans, or error;
- confidence semantics: per span, calibrated or not.

2. Category:  reuse label alignment from #2759 

`alignScoresToMapping` rejects unknown labels, duplicates, missing mapping labels, score sets not summing to ≈1.

3.  `ScoringBackend` needs declared range and direction

Raw score plugs into existing threshold unchanged only if providers agree on scale. One difficulty scorer emits `[0,1]`, high = hard. Another emits logits, high = easy. Thresholds invert, no error raised. Require `[0,1]`, higher = harder, validate on first call.

4. Runtime label sets

GLiNER-style entity model takes label set per request, not at load time. Fixed-label contract pins it to static set, loses that property. Decide now whether token contract carries optional per-request label set — close to `rule_token_score` in #2779  — even if first implementation ignores it.

5. Failure policy

Unreachable endpoint logs error and skips rule, so request looks checked and clean. Jailbreak: security gap. PII: compliance gap. Same code path. One shared `on_error: allow | block` knob for guardrail plus new classifiers beats per-signal defaults.


[...]

PR #2930, opened 2026-08-18
[Router][Operator][E2E][Docs] generalize label-mapping validation and add on_error fail-closed contract for classifier backends

### LINKED ISSUE #2918: feature: generalize http_classify label-mapping validation + shared on_error contract for classifier backends
### Describe the feature

#2759 added `alignScoresToMapping`/`assignScoreToMapping` in `http_classifier.go` to validate an `http_classify` response against a label mapping (reject unknown/duplicate/missing labels, enforce the scores sum to ~1.0). It's hardcoded to `*JailbreakMapping`, so #2760's category classifier can't reuse it without a copy-paste.

This issue generalizes that validation behind a small interface (e.g. exposing `GetIndexForType`/`GetTypeFromIndex`/label count) that both `JailbreakMapping` and `CategoryMapping` can satisfy, so `alignScoresToMapping`/`assignScoreToMapping` become one shared implementation instead of two near-identical copies.

It also adds a shared `on_error: allow|block` config knob usable across `prompt_guard`, category, and PII backends. Today, an unreachable classifier endpoint logs an error and silently skips the rule — the request looks checked and clean when it wasn't. @adaamko flagged this on #2760 as a security gap for jailbreak and a compliance gap for PII, hitting the same code path either way; a single shared knob replaces per-signal implicit defaults.

Note: `on_error` already exists with different semantics (`skip|fail`) in `fusion_config.go`/`remom_config.go` — name this field distinctly enough to avoid confusion with that existing, unrelated pattern.

This is a prerequisite refactor for the category/complexity/PII sub-issues under #2760.

### Primary layer

global level

### Why this layer?

Model backend selection and error-handling policy for a built-in classifier is router-owned config, not a routing-rule concern — mirrors where `prompt_guard.backend` already lives, under `global.model_catalog.modules`.

Without this, category/complexity/PII would each need their own copy of label-mapping validation and their own ad hoc error-handling default, instead of reusing one contract — increasing the amount of code #2760 has to add and review.

### Additional context

Sub-issue of #2760. Follow-up from @adaamko's review comment on #2760 (2026-08-14).

theohsiung:
Closes #2918

## Purpose

Two parts, deliberately: the prerequisite refactor #2918 asked for, and the fail-open paths that surfaced while proving it actually works. Modules affected: `src/semantic-router`, `deploy/operator`, `e2e`, `website/docs`.

**Part 1 — the prerequisite refactor (#2918)**

- Generalizes #2759's `alignScoresToMapping`/`assignScoreToMapping` behind a `sequenceLabelMapping` interface; `CategoryMapping` satisfies it too.
- Generalizes `HTTPClassifierJailbreakInference` → `HTTPClassifierInference` so #2920 reuses it directly. Validated against a real external model (a 14-label category classifier) over the live `http_classify` contract, not a mock.
- Adds a shared `on_error: allow|block` contract as an embeddable `ClassifierOnErrorConfig`, per @adaamko's [point 5](https://github.com/vllm-project/semantic-router/issues/2760#issuecomment-5294683410) ("one shared knob … beats per-signal defaults") and the issue's explicit `allow|block` naming.

**Part 2 — making `block` actually close**

Wiring the knob to one real unreachable classifier showed the fail-closed path held on only one of four paths. Each of these was a fail-open that made `on_error: block` a partial or silent no-op:

- **Request path:** a genuine detection was being replaced by the error sentinel (erasing the real label/score that feed `jailbreak_type` and the `jailbreak.type` span attribute, so a replayed real attack looked like a guardrail outage); a successful-but-uninterpretable result (argmax outside the mapping) was treated as verified-safe.
- **Response path:** `res_filter_jailbreak.go` never consulted `IsBlock()` — it had exactly one call site repo-wide. A guardrail outage still served unverified model output with 200.
- **`method: contrastive` rules:** entirely exempt; a request-time embedding failure left `MaxScore` at `-1`, i.e. "no match".
- **Config, two ways:** the Operator CRD had no `on_error` field, so the API server pruned it and operator-managed deployments silently failed open (same gap class as #2759's `variant`/`protocol` miss — extended the regression test that miss left behind). And nothing validated that an enabled `prompt_guard` was wired up: a mistyped `model_role` left `config.Parse()` clean and `IsBlock()` true while the signal never dispatched.
- **Mapping arity/consistency:** the four accepted label-map shapes now collapse to one agreed set at load. Previously `LabelCount()` and `GetJailbreakTypeFromIndex` read different maps, so a resolvable class could have no distribution slot — a truncated top-1 response passed the completeness guard and read as `0.0`. `LabelCount() == 0` was also constructible, which fails every request while blaming the server.

**Scope boundaries.** `CategoryModel`/`PIIModel` embedding `ClassifierOnErrorConfig` stays with #2920/#2922 — the type is embeddable now and notes are on #2920/#2921/#2922/#2923, but those backends don't exist yet. Bounded I/O on the connector stays #2924; this touches `http_classifier.go` but deliberately adds no byte limits.

**On #2779.** The label side aligns: `label_distribution.v1`'s `preserve_all_declared_labels: true` is exactly the invariant `alignScoresToMapping` enforces. The failure side does not — #2779 places `on_error` per classifier instance with `unknown`, this is per-module with `allow|block`, which is what #2760 point 5 explicitly asked for over per-signal defaults. Reconciling later is a semantic remap, not a shim (`unknown` = "no verdict"; `block` = "record a detection at confidence 1.0"). Discussion in the thread below.

## Test Plan

```sh
# unit + race, from src/semantic-router
go test ./pkg/classification/ ./pkg/config/ ./pkg/extproc/ -count=1
go test ./pkg/config/ -count=1 -race

# operator module
cd deploy/operator && go test ./controllers/ ./api/... -count=1
make manifests generate bundle          # CRD bases, OLM bundle
make docs-crd                           # generated CRD reference

# lint gates
make agent-lint CHANGED_FILES="..."     # per changed file
make go-lint markdown-lint yaml-lint

# e2e (manual-only profile, real kind cluster)
make e2e-test E2E_PROFILE=jailbreak-onerror
```

Plus two targeted regression checks written for this PR: a table over all four accepted mapping shapes through the new canonicalization, and parsing every router config in-repo through the new prompt_guard validator.

## Test Result

- `pkg/classification`, `pkg/config`, `pkg/extproc` pass, including `-race`. Operator controller + api tests pass; `Verify Manifests` and `Validate OLM Bundle` confirm the regenerated CRD and bundle.
- **e2e `jailbreak-onerror` on a local 3-node kind cluster** (real Envoy Gateway + AI Gateway + semantic-router; upstream is `mock-vllm`, guardrail deliberately unreachable at `127.0.0.1:1`):

  ```
  [Test] blocked=true selected_decision="block_on_classifier_error" status=200
  ✅ Test jailbreak-onerror-block passed
  ```

  `status=200` because `fast_response` returns the decline as a normal completion — the block surfaces as `x-vsr-fast-response: true` plus the decision name, which is what the test asserts.
- Canonicalization is covered by a table over all four accepted mapping shapes, in both directions. No `jailbreak_type_mapping.json` ships in-repo (the models are fetched at runtime), so the shapes that matter in practice are the e2e ConfigMap's inline mapping and what the training pipeline writes (`label_to_id` + `id_to_label` from the same dicts, so consistent and contiguous by construction). All 10 router configs in-repo still parse, 9 of them with `prompt_guard` enabled — the new validator rejects nothing that previously worked.
- Full CI green (29 checks), DCO green, Codecov reports all modified and coverable lines covered.

**Remaining blockers / risk.** Needs ≥1 approval; no technical blockers. One found-on-the-way defect is deliberately *not* fixed here: `prompt_guard` is silently non-functional under the Operator, because the CRD's `Enabled bool` carries `omitempty` (so `enabled: false` is unrepresentable and falls back to default `true`) while `jailbreak_mapping_path` lacks `omitempty` (so `""` blanks the default path). Fixing it changes how those fields serialize — asked in the thread below whether to file separately or fold in. Also pre-existing and out of scope: the response cache is written before the response guardrail runs, and jailbreak inference fans out unbounded serial calls.

Note `7407dc5` does not compile on its own — an unrelated test file was swept into it and its implementation lands in the next commit. That only affects per-commit review, not `main`: `.mergify.yml` sets `merge_method: squash`, so the intermediate state never lands. Say the word if you'd still prefer the history rewritten.

---
<details>

</details>


AayushSaini101:
Thanks for raising the error handling request, pre-commit hook is failing can you check, and also can you make the comments shorter or remove the comments where it is not required for easy review thanks @theohsiung

theohsiung:
> Thanks for raising the error handling request, pre-commit hook is failing can you check, and also can you make the comments shorter or remove the comments where it is not required for easy review thanks @theohsiung

Sure, will take a look! Thx for the advise @AayushSaini101

AayushSaini101:
Thanks for working on this, @theohsiung can you add a small demo of this working case during the on_error as well it would be great to review end to end thanks

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
```diff
@@ -170,6 +180,59 @@ func (c *Classifier) evaluateBERTJailbreakRule(rule config.JailbreakRule, conten
...
 
+// jailbreakCandidateOutcome is the disposition of one cached result within
+// findBestJailbreakMatch's scan. A dedicated type (rather than two
+// independent bools) makes the invalid combination - both "matched" and
+// "fail-closed" true at once - unrepresentable.
+type jailbreakCandidateOutcome int
+
+const (
+	// jailbreakCandidateNone means this cached result contributed nothing:
+	// a classify error tolerated by on_error: skip, an unknown class index,
```
adaamko: `on_error: skip` - validator only accepts `allow|block`.

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
```diff
@@ -170,6 +180,59 @@ func (c *Classifier) evaluateBERTJailbreakRule(rule config.JailbreakRule, conten
...
+	// or a risk score below the rule's threshold. The zero value, so a
+	// zero-value jailbreakCandidate is safely "no contribution".
+	jailbreakCandidateNone jailbreakCandidateOutcome = iota
+	// jailbreakCandidateMatched means jailbreakType/riskScore are a genuine
+	// content-based match.
+	jailbreakCandidateMatched
+	// jailbreakCandidateFailClosed means on_error: fail turned a classify
```
adaamko: same here, `on_error: fail` should read `block`.

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
```diff
@@ -15,6 +15,16 @@ type cachedJailbreakResult struct {
...
 
+// jailbreakClassificationErrorType is the sentinel jailbreak type reported
+// when config.PromptGuardOnErrorBlock forces a rule to match because
+// inference itself failed (e.g. an unreachable http_chat/http_classify
+// endpoint), rather than because the content was actually classified as a
+// jailbreak. It is distinguishable from a real detected type in
+// results/logs. Without this fail-closed path (config.PromptGuardOnErrorAllow,
+// the default), a classify error is indistinguishable from a genuinely safe
+// request - see @adaamko's review on #2760.
+const jailbreakClassificationErrorType = "classification_error"
```
adaamko: collides if a deployment's `jailbreak_mapping.json` has a `classification_error` label - then it isn't distinguishable from a real detected type after all. cheap to reject at mapping load.

src/semantic-router/pkg/config/model_config_types.go
```diff
@@ -155,6 +155,11 @@ type PromptGuardConfig struct {
 	// exclusive with Variant. Requires an external model configured with
 	// model_role="guardrail".
 	Protocol string `yaml:"protocol,omitempty"`
+
+	// OnError selects what a classifier-backend failure does to the rule
+	// that failed to evaluate: PromptGuardOnErrorAllow (default) or
+	// PromptGuardOnErrorBlock. See the constants' doc comments.
+	OnError string `yaml:"on_error,omitempty"`
```
adaamko: scoping question, more about #2920 than this PR: #2760 point 5 asked for one shared knob covering guardrail plus the new classifiers. on `PromptGuardConfig` it isn't reachable from category (#2920), PII (#2922), or the `classifiers` family, so each ends up declaring its own - the per-signal defaults that point was arguing against. worth extracting to a small embeddable struct before #2920 copies this shape?

website/docs/tutorials/global/safety-models-and-policy.md
```diff
@@ -72,6 +72,28 @@ global:
 `http_chat` uses a chat-completions prompt. Both send request text to the
 configured service.
 
+### On a classifier failure
```
adaamko, line 75: worth a sentence that `block` needs a decision consuming the jailbreak signal - the e2e profile adds `block_on_classifier_error` for exactly that. without one the signal fires at confidence 1.0 and the request still goes through. same thing any demo has to show, or it looks like a no-op.

adaamko (commented):
Hey, took a look.

two of the `skip|fail` comments survived the revert. flagged inline so the trim doesn't just shorten the wrong values.

on the demo ask: coverage is already there, `e2e/testcases/jailbreak_onerror_block.go`, real unreachable endpoint. it's `selection_mode: manual-only` in the profile map so it never runs. flip that or paste a run log.

one scoping question on the shared knob, inline.

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
theohsiung: Fixed in d45dc185 - was just a stale comment example, not a validated value (validator only ever accepted allow|block). Reworded.

(Edited: the SHA I first quoted here, 9a182184, was pre-rebase and never landed upstream - sorry for the dead link.)

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
theohsiung: Same, fixed in d45dc185.

(Edited: was 9a182184 before the rebase.)

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
theohsiung: Good catch. Fixed in 05595471 - LoadJailbreakMapping now rejects any mapping that configures a label equal to jailbreakClassificationErrorType.

(Edited: the SHA I first quoted, ffcee2f3, was pre-rebase and never landed upstream. I also originally wrote "so this can't happen" here - that was wrong, see the follow-up below.)

src/semantic-router/pkg/config/model_config_types.go
theohsiung: Agreed - extracted ClassifierOnErrorConfig in d45dc185 (pkg/config/classifier_on_error.go), embeddable via yaml:",inline" so #2920/#2922/#2923 pick it up instead of redeclaring. Notes left on #2920, #2921, #2922 and #2923 to actually use it.

(Edited: was 9a182184 before the rebase, and my note list originally read #2920/#2922/#2921 - it should have included #2923, the `classifiers` family you asked about. That note is now there, and it flags a wrinkle specific to that issue: the generic classifier signal is the one family that already has a condition-level `on_error: no_match|match`, so a backend-level `allow|block` overlaps there in a way it doesn't for jailbreak.)

website/docs/tutorials/global/safety-models-and-policy.md
theohsiung: Added in 9d4ecfd5, pointing at the e2e profile's block_on_classifier_error decision as the worked example.

(Edited: was 45cc84e9 before the rebase.)

adaamko:
Thanks, i think you forgot to push your commits @theohsiung

theohsiung:
@AayushSaini101 @adaamko ran `jailbreak-onerror` end-to-end in a local kind cluster (3-node, real Envoy Gateway + AI Gateway + semantic-router - the only mock is the deliberately-unreachable guardrail endpoint). Fixes pushed in c0874b7.

Result:
```
[Test] blocked=true selected_decision="block_on_classifier_error" status=200
✅ Test jailbreak-onerror-block passed (2.07s)
Total: 1 | Passed: 1 | Failed: 0
✅ All tests passed!
```

Getting there surfaced 4 real, unrelated bugs the profile itself was hitting (all fixed in that commit):
- `IsJailbreakEnabled`/`LoadJailbreakMapping` require a `jailbreak_type_mapping.json` on disk even when `protocol: http_classify` - the mapping-file requirement isn't conditioned on protocol. Fixed by supplying a minimal stand-in mapping via a prerequisite ConfigMap + `extraVolumes`, instead of baking/downloading the real model (irrelevant here since this profile never runs local candle inference).
- `stores.response_cache` referencing `embedding_model: mmbert` eagerly downloaded an unrelated embedding model at pod startup, once causing a helm-install timeout and once an OOMKill - disabled the cache and bumped memory 2Gi→6Gi.
- the external guardrail model config was missing `llm_model_name`, which the router requires at startup.
- the test was sending `model: "openai/mock-model"` directly - a concrete configured backend model, which the router treats as passthrough and bypasses every recipe-local signal/decision/plugin for (by design, unrelated to this PR). Switched to the `"MoM"` auto-routing alias, matching the existing `jailbreak-detection` test's convention, so the request actually reaches the jailbreak signal.

On `selection_mode: manual-only`: confirmed it's documentation-only - the only thing that actually triggers a profile automatically in CI is the fixed `full_ci_profiles` list in `e2e-profile-map.yaml` (currently `envoy-ai-gateway`/`dashboard`/`remote-embedding`). I didn't add this profile to it myself since that changes CI cost/time for every future PR, not just this one. @AayushSaini101 want it added, or is manual/on-demand fine for a profile like this?

theohsiung:
@AayushSaini101 the demo you asked for - `jailbreak-onerror` end-to-end in a local kind cluster (real Envoy Gateway + AI Gateway + semantic-router; the upstream is `mock-vllm`, and the guardrail endpoint is deliberately unreachable at `127.0.0.1:1`):

```
[Test] blocked=true selected_decision="block_on_classifier_error" status=200
✅ Test jailbreak-onerror-block passed (2.07s)
```

`status=200` because `fast_response` returns the decline as a normal completion - the block shows up as `x-vsr-fast-response: true` plus the decision name, which is what the test asserts on.

@adaamko re-reviewing my own replies turned up that the sentinel-collision fix I reported to you was incomplete; corrected in that thread above. I'm still landing follow-ups from the same pass, so rather than a running commentary I'll post one consolidated summary once it settles. Nothing new here needs re-review yet.

src/semantic-router/pkg/classification/classifier_signal_jailbreak.go
theohsiung: Correcting my reply above: "so this can't happen" was wrong - that first guard only covered one of the four mapping shapes `LoadJailbreakMapping` accepts.

It indexed `LabelToIdx`, but the loader only back-fills `label_to_id -> LabelToIdx`, never `idx_to_label -> LabelToIdx`, while `GetJailbreakTypeFromIndex` resolves labels from `IdxToLabel`/`IDToLabel`. So a mapping declaring only `idx_to_label`, or only `id_to_label`, loaded without error and still resolved its index to `classification_error` - exactly the collision you flagged. Verified with a probe: 2 of the 4 shapes slipped through. (The shipped `mmbert32k` mapping declares `label_to_id`, so the back-fill already covered it - the gap was reachable via a hand-written index->label-only mapping.)

Fixed in `7ee7ddd`: the guard now probes via `GetIndexForJailbreakType`, the same lookup the runtime uses, so it can't drift from what it guards, with the test table covering every shape in both directions.

theohsiung:
Two things need a call at the bottom.

**#2918's prerequisite is in.** `sequenceLabelMapping` generalizes
`alignScoresToMapping`/`assignScoreToMapping` — `CategoryMapping` satisfies it,
validated against a real 14-label category model over the live `http_classify`
contract — and `ClassifierOnErrorConfig` carries `allow|block` as an embeddable
struct so #2920/#2922/#2923 pick it up instead of redeclaring it. This PR wires
only `prompt_guard`; category/PII config stays with those issues.

**`on_error: block` turned out to be honored on one of four paths.** Standing up
the e2e profile surfaced the rest, all fixed here:

<details><summary>the four paths, plus two mapping bugs behind them</summary>

- `35f5904` — the response side swallowed the classify error, so an unreachable guardrail blocked the request while the model's own output was forwarded unverified with 200. `IsBlock()` had exactly one call site in the tree.
- `676e1dc` — `method: contrastive` never consulted it; per-message embedding failures were logged and skipped, leaving `MaxScore` at -1, i.e. "no match".
- `7407dc5` — an argmax index absent from the mapping bypassed the check six lines below it; and the scan returned early on the first error, so a real 0.97 detection alongside one errored chunk was reported as `("classification_error", 1.0)`, erasing what feeds `jailbreak_type` in replay.
- `a3715d8` — `LabelCount()` and `GetJailbreakTypeFromIndex` read different label maps, so a resolvable class could have no distribution slot and read as 0.0 past the completeness guard. All four accepted maps now collapse into one canonical set.
- `7407dc5` — `http_classify` was the only backend without an arity check, and `LabelCount() == 0` made it reject every valid response.

</details>

**`prompt_guard` is silently non-functional under the operator.** The CRD's
`Enabled bool` has `omitempty`, so `enabled: false` emits nothing and the router
falls back to its default `true`; separately
`CanonicalPromptGuardModule.jailbreak_mapping_path` has no `omitempty`, so the
operator writes `""` and blanks the default path — which `IsPromptGuardEnabled()`
turns into a no-op. Neither disabled as asked nor working. Found because a
validator I added made it a hard startup failure; `1161abc` narrows that check
back to the remote-backend fields. Separate issue, or fold in?

**@adaamko — `on_error` scope vs #2779.** The label side lines up:
`label_distribution.v1` carries `preserve_all_declared_labels: true`, exactly the
invariant `alignScoresToMapping` enforces. The failure side doesn't — #2779 puts
`on_error` per classifier instance with `unknown`, this is per-module with
`allow|block`, which is what your point 5 asked for over per-signal defaults.
Worth naming because the sentinel travelling through `JailbreakType` is what a
`Binding` migration would replace, and `SignalErrors` already matches `unknown`
semantics. Small follow-up if you'd rather have that shape.

Not here: #2924's bounded I/O, plus two pre-existing items — the response cache
is written before the guardrail runs, and jailbreak inference still fans out C×K
serial calls.

AayushSaini101 (approved):

adaamko (approved):

PR #2945, opened 2026-08-19
[Router] fix: correct rule_name/category keys to name in helm values.yaml

### LINKED ISSUE #2939: bug: Helm values use unknown rule_name/category keys so thinking_decision never matches
### Describe the bug

`deploy/helm/semantic-router/values.yaml` writes decision rule leaves with a `rule_name` key:

```yaml
- name: thinking_decision
  priority: 20
  rules:
    operator: OR
    conditions:
      - type: keyword
        rule_name: thinking   # <- not a RuleNode field
```

`config.RuleNode` has no `rule_name` field (`src/semantic-router/pkg/config/decision_config.go:163-171`); the signal reference lives in `name`. The key is dropped at load, so the leaf becomes `{type: keyword, name: ""}` and `evalComposerLeaf` looks for a matched keyword rule named `""`.

The keyword signal these decisions point at has the same problem in the same file:

```yaml
signals:
  keywords:
    - category: thinking    # <- KeywordRule has `name`, not `category`
      operator: OR
      keywords: [urgent, immediate, asap, think, careful]
```

`config.KeywordRule` declares `name` (`src/semantic-router/pkg/config/signal_config.go:44-55`), so `category` is dropped too and the rule is registered with an empty name.

Occurrences: `rule_name` at values.yaml lines 1026, 1857, 6445, 7276. The same `thinking_decision` is written correctly elsewhere in the same file (lines 1483, 3004 use `name: thinking`), which is what makes it a typo rather than a supported alias.

### To Reproduce

1. Render the chart from any of the four affected profiles and load the resulting config.
2. Inspect the parsed decision: the `thinking_decision` leaf carries an empty signal name.
3. Send a request containing one of the `thinking` keywords — the decision does not select the model the chart intends.

### Expected behavior

- The chart's values use the documented field names (`name` on both the rule leaf and the keyword signal).
- A leaf that declares `type` without `name` fails config validation instead of loading as a reference to nothing.
- Unknown keys inside a rule node or a signal rule are reported rather than dropped (this half overlaps #2469).

### Affected layer

signal

### Additional context

Found while scanning every maintained rule tree for #2122 / PR #2123. That PR deliberately does not add the "a leaf requires both `type` and `name`" check, because enforcing it would make the shipped chart's own default config fail to load until these values are corrected — the chart fix has to land first (or together with it).

csarushan1729:
Closes #2939

## Purpose

The Helm chart's default `values.yaml` writes decision rule leaves using a `rule_name` key and keyword signals using a `category` key. Neither field exists in the actual config schema- `config.RuleNode` and `config.KeywordRule` both expect `name`. This causes the affected `thinking_decision` leaves to load with an empty signal name, so they silently fail to match instead of routing on the intended keywords.

This PR corrects the four `rule_name` occurrences and two `category` occurrences flagged in #2939 (lines 1026, 1857, 6445, 7276, and 1061, 6480) to use `name`, matching the correct pattern already used elsewhere in the same file (lines 1483, 3004).

## Test Plan

- Confirmed the exact lines matched the issue's reported occurrences before editing.
- Validated the full `values.yaml` (all 41 YAML documents) parses cleanly after the change using `yaml.safe_load_all`.
- Did not have `make`/Docker available locally on Windows to run the repo's Go-based render/test harness, relying on CI to confirm the rendered chart behavior.

## Test Result

YAML parses without error; diff is limited to the 6 flagged lines with no unintended changes.

**Note for maintainers:** while fixing the reported keys, I noticed two of the four affected profiles (around lines 1857 and 7276) don't define a `thinking` keyword signal at all, even under the wrong key name, so `thinking_decision` won't match there even after this fix. I left that out of scope for this PR since it wasn't part of the original report, but happy to add the missing signal block here or in a follow-up if you'd like.

---
<details>

</details>

### DIFF
--- a/deploy/helm/semantic-router/values.yaml
+++ b/deploy/helm/semantic-router/values.yaml
@@ -1023,7 +1023,7 @@ config:
           operator: OR
           conditions:
             - type: keyword
-              rule_name: thinking
+              name: thinking
         modelRefs:
           - model: base-model
             lora_name: general-expert
@@ -1058,7 +1058,7 @@ config:
               mode: replace
     signals:
       keywords:
-        - category: thinking
+        - name: thinking
           operator: OR
           keywords:
             - urgent
@@ -1854,7 +1854,7 @@ config:
           operator: OR
           conditions:
             - type: keyword
-              rule_name: thinking
+              name: thinking
         modelRefs:
           - model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
             use_reasoning: true
@@ -6442,7 +6442,7 @@ config:
           operator: OR
           conditions:
             - type: keyword
-              rule_name: thinking
+              name: thinking
         modelRefs:
           - model: base-model
             lora_name: general-expert
@@ -6477,7 +6477,7 @@ config:
               mode: replace
     signals:
       keywords:
-        - category: thinking
+        - name: thinking
           operator: OR
           keywords:
             - urgent
@@ -7273,7 +7273,7 @@ config:
           operator: OR
           conditions:
             - type: keyword
-              rule_name: thinking
+              name: thinking
         modelRefs:
           - model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
             use_reasoning: true


AayushSaini101 (approved):
LGTM Thanks : )

PR #2948, opened 2026-08-19
[Router] fix: preserve Anthropic cache usage on streaming responses

### LINKED ISSUE #2947: [Bug] Anthropic clients see zero cache usage on streamed responses
### Describe the bug

When an Anthropic-protocol client (e.g. Claude Code) calls `/v1/messages`
through the router and an Anthropic backend serves the request, prompt-cache
usage is reported correctly in non-streaming responses — but streaming
responses contradict themselves: `message_start` relays the true upstream cache
counters while the terminal `message_delta` reports them as `0`. Anthropic SDK
accumulators merge `message_delta.usage` as the final word, so streaming
clients end up seeing zero cache usage on cached requests. The router's own
streaming cost metrics undercount identically.

This hits the flagship supported client on its default settings: Claude Code
and the Anthropic SDKs inject `cache_control` markers automatically and stream
by default, so essentially every interactive request through this path gets
cached (and billed as cached) while client-side cost tracking and router
metrics report zero cache activity.

Verified against the live Anthropic API (`claude-haiku-4-5-20251001`).

#### The same missing capture affects OpenAI-protocol clients

When a `/v1/chat/completions` request is routed to an Anthropic backend, the
cache counters are lost there too — streaming for the same reason, and
non-streaming because the step that folds cache counters into `prompt_tokens`
only runs when the `IRExtensions` sidecar exists, and the sidecar is only
created for Anthropic-protocol clients:

```go
// src/semantic-router/pkg/anthropic/client.go:210
if ext == nil ||
    (usage.CacheReadInputTokens == 0 &&
        usage.CacheCreationInputTokens == 0) {
    return encoded, nil
}
```

so `prompt_tokens` reflects only the uncached slice and
`prompt_tokens_details` never carries `cached_tokens`/`cache_write_tokens`.
This scenario requires the backend itself to enable caching: the OpenAI
request shape has no way to set `cache_control`, so `api.anthropic.com` will
not cache these requests. Anthropic-compatible gateways and serving stacks
that add cache markers themselves (or cache implicitly) do report these
counters — and the router drops them. A fix should cover both surfaces.

### To Reproduce

1. Route `/v1/messages` through the router to the real Anthropic API.
2. Send a streaming request (`"stream": true`) with a system prompt marked
   `cache_control: {"type": "ephemeral"}` and long enough to be cacheable
   (Anthropic requires ≥2048 tokens on Haiku models; the example below used
   ~8.5k), then send the identical request again.
3. Each streamed response contradicts itself: `message_start` carries the real
   cache counters (`cache_creation_input_tokens` on the first request,
   `cache_read_input_tokens` on the repeat), while the terminal
   `message_delta` zeroes them. From the repeat request:
   ```
   message_start usage: {..., "cache_read_input_tokens": 8502, "input_tokens": 13, "output_tokens": 4}
   message_delta usage: {"input_tokens": 13, "output_tokens": 4,
                         "cache_creation_input_tokens": 0, "cache_read_input_tokens": 0}
   ```

### Expected behavior

The terminal `message_delta.usage` carries the same cache counters the
upstream reported, consistent with `message_start` and with the non-streaming
path. This is what Anthropic itself sends — the identical streaming request
made directly against `api.anthropic.com` reports the counters in both events:

```
message_start usage: {"input_tokens": 13, "cache_read_input_tokens": 8502, ...}
message_delta usage: {"input_tokens": 13, "cache_creation_input_tokens": 0,
                      "cache_read_input_tokens": 8502, "output_tokens": 4}
```

### Affected layer

global level

### Additional context

The stream is translated Anthropic SSE → OpenAI chunks → Anthropic SSE.
`handleMessageDelta` (`src/semantic-router/pkg/anthropic/client_stream.go:348`)
captures stop reasons into `IRExtensions` but not the `MessageDeltaUsage`
cache fields (despite the comment at
`processor_res_body_streaming_anthropic.go:25` saying it does). The outbound
emitter then sees zero counters, and since the fields lack `omitempty`, writes
explicit zeros into the final `message_delta`:

```go
// src/semantic-router/pkg/anthropic/outbound.go:101
type anthropicUsageResponse struct {
    InputTokens              int64 `json:"input_tokens"`
    OutputTokens             int64 `json:"output_tokens"`
    CacheCreationInputTokens int64 `json:"cache_creation_input_tokens"`
    CacheReadInputTokens     int64 `json:"cache_read_input_tokens"`
    ...
}
```

Suggested fix: capture the cache fields from `MessageDeltaUsage` into
`IRExtensions` in `handleMessageDelta`, and allocate `IRExtensions` whenever
the backend `APIFormat` is anthropic regardless of client protocol. The
existing consumers (`buildAnthropicUsage`, `mergeProviderResponseUsage`,
`mergeProviderStreamingUsage`) already do the right thing once the counters
are present.

Related: #1944 (RFC #1946) fixed the request-side `cache_control`
propagation; #1941 introduced the streaming emitter.

csarushan1729:
Fixes #2947

## Purpose

Streaming responses through the Anthropic path were reporting zero prompt-cache usage to clients, even when the upstream response had non-zero `cache_read_input_tokens` / `cache_creation_input_tokens`.

The stream is translated Anthropic SSE → OpenAI chunks → Anthropic SSE. Cache counters arrived correctly on `message_start`, but were not captured into `IRExtensions` on the inbound streaming path, so the outbound terminal `message_delta` wrote explicit zeros. Anthropic SDKs treat `message_delta.usage` as authoritative, so clients (including Claude Code) saw zero cache activity.

This change:

1. Captures cache usage from `message_start` into `IRExtensions` in `handleMessageStart`.
2. Preserves non-zero cache fields from `message_delta` in `handleMessageDelta` (without overwriting good values with zeros).
3. Allocates `IRExtensions` for Anthropic-backend streaming responses even when the client is OpenAI-protocol, so cache counters are not dropped on that path either.

Affected modules: `pkg/anthropic` (streaming translate), `pkg/extproc` (Anthropic streaming response body handler).

## Test Plan

- [x] `go test ./pkg/anthropic/ -count=1`
- [x] `go test ./pkg/extproc/ -count=1` (or package tests covering Anthropic streaming)
- [ ] Manual / CI: stream a cached Anthropic request through the router and confirm terminal `message_delta.usage` keeps the same cache counters as `message_start` (and as a direct call to the Anthropic API)
- [ ] Confirm OpenAI-protocol client → Anthropic backend streaming path no longer drops cache counters when the backend reports them

## Test Result

Local Go tests were not run in this environment (Go toolchain not installed). Relying on CI for package tests.

Manual verification against a live Anthropic backend with streaming + prompt caching is still recommended after merge/CI green.


AayushSaini101:
Thanks for raising it, can you add some test cases as well @csarushan1729 : )

csarushan1729:
Thanks for the review, @AayushSaini101!

Added test cases in `client_stream_cache_test.go` (`src/semantic-router/pkg/anthropic/`):

- `message_start` correctly captures cache counters into `IRExtensions`
- `message_delta` **preserves** non-zero cache counters instead of zeroing them (the actual bug from #2947), but still **updates** them when a genuinely larger value comes in
- Nil-`ext` path doesn't panic
- Full SSE round trip (`message_start → ... → message_delta → message_stop`) confirming terminal usage matches `message_start`, plus a multi-chunk variant

`go test ./pkg/anthropic/ -count=1` passes locally; `go test ./pkg/extproc/ -count=1` passes in CI.

Still open:
- Live manual verification against a real Anthropic backend with caching; can do before merge if useful
- Explicit `pkg/extproc` coverage for the `IRExtensions` allocation on OpenAI-protocol clients, didn't have enough context on that package's test scaffolding to add it confidently. Happy to if you point me at the right pattern.

FAUST-BENCHOU (commented):
A couple of questions on the remaining cache-usage paths.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic.go
```diff
@@ -22,6 +23,11 @@ func (r *OpenAIRouter) handleAnthropicStreamingResponseBody(
 		ctx.AnthropicStream = anthropic.NewStreamState()
 	}
 
+	// Always allocate IRExtensions when the backend is Anthropic so
+	// cache counters are captured even for OpenAI-protocol clients.
+	if ctx.IRExtensions == nil {
+		ctx.IRExtensions = &ir.IRExtensions{}
```
FAUST-BENCHOU, line 29: Could we allocate IRExtensions for Anthropic backends before the streaming branch instead? The OpenAI-client non-streaming path mentioned in #2947 still reaches ToOpenAIResponseBodyWithExt with a nil sidecar and drops the cache counters.

src/semantic-router/pkg/anthropic/client_stream.go
```diff
@@ -366,6 +371,18 @@ func handleMessageDelta(
 		}
 	}
 
+	// Keep cache counters: prefer non-zero values from message_delta,
+	// otherwise leave whatever message_start already stored in ext.
+	if ext != nil {
```
FAUST-BENCHOU, line 376: For OpenAI streaming clients, are these cache counters actually returned on the wire? buildOpenAIStreamFinishChunk still builds usage only from MessageDeltaUsage, so IRExtensions seems to fix internal accounting but not the client response.

dorota-memex: On this one: #3045 (in review) also touches buildOpenAIStreamFinishChunk — it merges the message_start input tokens into the terminal usage chunk. Once that lands, surfacing the cache counters on the wire could reuse the same mechanism (InitialUsage → prompt_tokens_details.cached_tokens). Happy to coordinate so the two changes don't collide.

Xunzhuo:
Thanks for tracing this and adding focused tests. Current main's unified protocol codec from #3076 replaced the changed path and now preserves Anthropic cache-read and cache-write usage across every target format; the focused regression passes. I'm closing this as already fixed with no unique remaining outcome.

PR #2949, opened 2026-08-19
[Router] Honor router replay Milvus consistency_level

### LINKED ISSUE #2840: bug: Router replay Milvus store ignores the configured consistency_level

drivebyer:
Closes #2840

## Purpose

- The router replay Milvus store parsed `consistency_level` but never applied it: `CreateCollection` and every `Query` ran at the SDK default, so the documented field (and the `DefaultMilvusConsistencyLevel = "Session"` constant) was dead config, and an update's read-modify-write could read a stale snapshot and resurrect old fields on the whole-document upsert.
- Wire the resolved level (case-insensitive; unset defaults to `Session`, unrecognized values warn) into `CreateCollection` and both query paths, with two floors so no read is ever weaker than the implicit Bounded the store ran at before:
  - Session floors to Bounded until this process's first write — the SDK degrades Session to Eventually when no session timestamp exists (fresh process, read-only replica).
  - The read feeding `updateRecordIf` never runs weaker than Session once this process has written, even under a configured Bounded/Eventually, keeping read-your-writes for update chains. `updateMu` only serializes updates in-process; it cannot prevent a stale Milvus read.
- Same bug class as #2821/#2896, applied to the router replay consumer (scoped out of that fix). Affects the `Router` module.
- Second commit (pure refactor, no behavior change): hoist the name→`entity.ConsistencyLevel` mapping duplicated between `pkg/cache` and this store into `pkg/milvus` as `ParseConsistencyLevel`; each package keeps its own fallback policy wrapper (cache: leave the SDK default; replay: default to Session).

## Test Plan

- `go build` / `go vet` / `gofmt` on `pkg/milvus`, `pkg/cache`, `pkg/routerreplay/...`
- `go test ./pkg/milvus/... ./pkg/routerreplay/... -count=1` — new unit tests cover the shared name→level parser, the replay store's fallback policy, and the read/update-read floor matrix over configured level × first-write state.
- `go test ./pkg/cache/ -run 'TestMilvus|TestWarnUnrecognized' -count=1` — the existing cache consistency-level tests pin that the refactor keeps cache behavior unchanged.

## Test Result

- Build, vet, gofmt clean; all router replay package tests pass (including the #2869 lifecycle/async suites). Milvus-backed integration paths run in CI.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/cache/milvus_consistency_level.go
+++ b/src/semantic-router/pkg/cache/milvus_consistency_level.go
@@ -1,33 +1,20 @@
 package cache
 
 import (
-	"strings"
-
 	"github.com/milvus-io/milvus-sdk-go/v2/client"
 	"github.com/milvus-io/milvus-sdk-go/v2/entity"
 
+	milvuslifecycle "github.com/vllm-project/semantic-router/src/semantic-router/pkg/milvus"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
 )
 
 // resolveMilvusConsistencyLevel maps a configured consistency level name to
-// the SDK constant. Names are matched case-insensitively with surrounding
-// whitespace trimmed. ok is false when the name is empty or unrecognized; the
-// caller then leaves the option unset so the Milvus SDK default (Bounded)
-// stays in effect, preserving today's behavior for deployments that do not
-// pin a level.
+// the SDK constant via the shared parser. ok is false when the name is empty
+// or unrecognized; the caller then leaves the option unset so the Milvus SDK
+// default (Bounded) stays in effect, preserving today's behavior for
+// deployments that do not pin a level.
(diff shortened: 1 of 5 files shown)


drivebyer:
Closing in favour of #2842, which targets the same issue (#2840) and was opened first.

I've moved the findings from this PR over there as a review: https://github.com/vllm-project/semantic-router/pull/2842#pullrequestreview-4992090224 — the main one being that a Session read degrades to Eventually until the process's first write, so the unset-config default needs a Bounded floor or the fix is a silent downgrade.

The `pkg/milvus.ParseConsistencyLevel` refactor here is orthogonal to the fix (it de-duplicates the name→level mapping shared with `pkg/cache`); I'll open it separately once #2842 lands.

Issue #2465, opened 2026-07-12
[Security] Remove reusable browser session tokens from Dashboard URLs


Xunzhuo:
## Summary

Use the existing HttpOnly cookie as the normal dashboard session boundary and remove reusable credentials from script-visible storage and URLs. Owner surfaces: `dashboard/backend/auth/middleware.go`, `dashboard/backend/auth/session_cookie.go`, and `dashboard/frontend/src/utils/authFetch.ts`.

## Current behavior

The dashboard accepts a reusable session credential from query strings, stores it in script-visible storage, and appends it to protected iframe, window, streaming, and socket URLs. This expands disclosure through browser and intermediary metadata and weakens the existing HttpOnly cookie control.

## Expected behavior

Same-origin browser transports authenticate with HttpOnly cookies. Ordinary API routes reject query credentials, unsafe cookie-authenticated methods enforce CSRF protection, and any exceptional URL ticket is single-use, short-lived, audience-bound, and route-bound.

## Acceptance

- Remove reusable session credentials from local storage, query strings, generated URLs, history, and Referer-bearing navigation.
- Remove global URL monkey-patching and rely on same-origin cookie behavior.
- Reject query credentials on ordinary API routes.
- Enforce Origin/Referer checks and CSRF tokens for unsafe methods.
- If a transport requires a URL ticket, limit it to one use, one audience/path, and a short expiry.
- Redact the legacy query field during migration.

## Validation

- Add frontend tests proving fetch, iframe, window, WebSocket, and EventSource URLs contain no reusable credential.
- Add backend tests for cookie-only flows, ordinary-route query rejection, CSRF failure, one-time ticket replay, audience mismatch, and expiry.
- Run `cd dashboard/backend && go test -race ./auth ./handlers ./router`.
- Run `cd dashboard/frontend && npm run test:unit && npm run type-check`.

## Related

- Parent audit: #2375
- Security index: #2357

abhinav-m22:
Hi @AayushSaini101 are you working on this issue? I'd like to pick it up. If you haven't started, could I be assigned?

AayushSaini101:
sure you can do thanks @abhinav-m22

abhinav-m22:
@AayushSaini101, @Xunzhuo What I found while digging in:

- The token in `?authToken=` is a full login credential, so it leaks through browser history, proxy logs and the `Referer` header. Pasting a dashboard URL into Slack shares your account.
[...]

Issue #2357, opened 2026-07-05
[Security] Track H2 router security hardening across auth, feedback, and Looper paths


Xunzhuo:
## Summary

Coordinate H2 security hardening across trusted identity, feedback integrity, Looper/plugin boundaries, tool execution, learning inputs, and policy bypass risk.

Parent roadmap: #2287  
Child issues: #1443, #1445, #1452, #1456  
Related: #1146, #1388, #2286, #2336, #2342, #2362, #2463, #2464, #2465, #2466

## Important package boundary

`src/semantic-router/pkg/authz` resolves credentials for upstream/provider access. It is **not** the router's RBAC/governance policy engine.

Authorization-related routing signals and decisions belong in validated config/classification/DSL/ext_proc surfaces. Tool-call final authorization must also be enforced at the external execution gateway (#2547).

## Scope

Track and coordinate:

- trusted, impersonation-resistant caller/tenant identity;
- management/config/feedback API authentication and abuse prevention;
- Looper/Fusion/ReMoM header, breadth, token, concurrency, and plugin-boundary safety;
- replay/diagnostic redaction and content-free security reasons;
- SSRF, reusable browser token, route-bound dashboard auth, and management API hardening;
- poisoning-resistant learning/outcome inputs;
- request-side tool visibility plus execution-boundary RBAC;
- E2E/security regression coverage and release gating.

## Cross-cutting follow-ups

- #2543 — bounded session/caller risk research; depends on trusted identity
- #2547 — tool RBAC across router filtering and execution gateway
- #2548 — privacy-bounded shared reputation research with an explicit trust domain
- #2549 — typed behavioral commitment policy state
- #2558 — governance-as-code ownership and hard-constraint precedence

Router Memory must not become a security reputation/policy database. Router Replay evidence and any security aggregate must be typed, content-minimized, TTL-bound, isolated, and poisoning-aware.

## Non-goals

- No identity derived from prompt text or arbitrary caller headers.
- No router-only claim of pre-execution enforcement for externally executed tools.
- No cost/latency/energy objective may weaken authorization or safety.
- No raw sensitive payload in ordinary metrics or diagnostic reasons.
- No public exploit detail beyond what is required to test and fix the issue.

## Acceptance criteria

- [ ] Every security-relevant attribute has a trusted source and enforcement owner.
- [ ] API, dashboard, feedback, Looper, and execution-gateway boundaries are authenticated and regression-tested.
[...]

PR #2962, opened 2026-08-21
[Dashboard] add CSRF protection to cookie authenticated writes

abhinav-m22:
Related #2465

## Purpose

The dashboard logs browsers in with a session cookie, and browsers attach cookies to any request aimed at this origin, including requests triggered by a hostile page. Previously, the backend did not check `Origin`, `Referer`, or a CSRF token.

On a local dashboard, a cross-site POST using only the session cookie could create an admin account with status 200. It also worked as a simple `Content-Type: text/plain` request, so an auto-submitting HTML form was enough. `SameSite=Lax` is the only protection today, but it does not cover top-level navigations.

This PR adds the missing CSRF checks. Nothing is removed, so it is safe to land on its own.

Step 1 of 2 for #2465. The follow-up removes the token from URLs and localStorage.

- Module: `Dashboard`

### How it works

Two checks run only when the request is authenticated by cookie or the legacy `?authToken=` parameter, and the method is `POST`, `PUT`, `PATCH`, or `DELETE`:

1. `Origin`, or `Referer` as a fallback, must match the dashboard's own origin or an entry in the new `DASHBOARD_ALLOWED_ORIGINS` list.
2. `X-CSRF-Token` must match `HMAC-SHA256(jwtSecret, "csrf:" + sessionID)`, where `sessionID` is the `jti` already inside the session JWT. Comparison is constant-time.

Requests using `Authorization: Bearer` skip both checks because browsers do not send that header automatically, so curl and CI clients are unaffected.

The token reaches the frontend through a readable `vsr_csrf` cookie, and the existing fetch wrapper sends it back. No individual call sites changed. Verification never reads the CSRF cookie. It recomputes the value from the JWT, so planting a cookie does not help an attacker.

Two smaller fixes in the same area:

- The OpenClaw WebSocket upgrader previously accepted every origin with `CheckOrigin: return true`. A WebSocket handshake is not covered by CORS, so this was its only origin check. It now uses the same origin check. A handshake without `Origin` is still accepted because non-browser clients cannot be driven cross-site.
- The proxy router logged the inbound `Referer` as-is, so a page URL containing `?authToken=` could write a live token to dashboard stdout. Credential parameters are now redacted before logging.

### Things worth flagging

- **Anyone already logged in gets one 403.** Existing sessions do not have a `vsr_csrf` cookie. The middleware issues it on the next request, including page loads, so the first write after the upgrade can fail and the retry works. This avoids logging every operator out.
- **New setting: `DASHBOARD_ALLOWED_ORIGINS`** (flag `-allowed-origins`). This is needed when running split dev servers, where the browser is on `:3001` and the backend is on `:8700`. When it is not set, the origin check also trusts `X-Forwarded-Host`, which a client can set on a directly reachable deployment. In that case, the origin check is advisory and the CSRF token is the real protection. Setting the allowlist makes the origin check load-bearing.
- **No stored token.** Deriving it from `jti` means no table, migration, cleanup job, or sibling-subdomain cookie attack.
- **`SameSite` stays `Lax`.** `Strict` would drop the cookie on deep links from alerts and tickets, making operators appear logged out.
- **Left out on purpose:** login CSRF and logout CSRF. Both use public routes that the middleware does not cover. They are considered low impact, and login CSRF would require a separate pre-session token mechanism.
- **Nearby, not fixed here:** `middleware/cors.go` reflects any `Origin` with `Allow-Credentials: true`. It is not exploitable today because auth returns 401 before CORS headers are written and `SameSite=Lax` blocks a credentialed preflight. It should have its own issue.

## Test Plan

- `cd dashboard/backend && go test ./auth/... ./handlers/... ./router/... -race`
- `cd dashboard/frontend && npm run test:unit && npm run type-check`
- `make dashboard-check && make dashboard-build`
- Manual comparison on `main` and this branch using six curl CSRF checks and the `Referer` redaction check
- Manual browser pass: save, delete, and config deploy

## Test Result

Go tests pass with `-race` in `auth`, `handlers`, and `router`. Frontend: 120 files, 485 tests, and `type-check` clean. No existing tests needed changes.

Manual run, same script on both branches:

| Check | main | this branch |
|---|---|---|
| Cross-site POST with the session cookie | 200, admin account created | **403** |
| Same attack as a simple request, no preflight | 200 | **403** |
| Legit write with `Origin` and `X-CSRF-Token` | 200 | 200 |
| Write with `Authorization: Bearer`, no origin, no token | 200 | 200 |
| Read with the session cookie | 200 | 200 |
| `Set-Cookie` on login | `vsr_session` only | `vsr_session` (HttpOnly) + `vsr_csrf` (readable) |
| `Referer` holding `?authToken=` in the API log | Full token in stdout | `authToken=%5BREDACTED%5D` |
| `grep -cF "$token" server.log` | 1 | **0** |

Screenshots:

| What | main (before) | this branch (after) |
|---|---|---|
| CSRF, the six curl steps | 2465-main-before-evidence | 2465-fix-branch-after-evidence |
| `Referer` redaction in the log | 2465-websocket-before | 2465-websocket-after |

**Note:** These Go tests are CI gated. `dashboard-check` includes `dashboard-test-backend` (`tools/make/dashboard.mk:120`). Playwright is still not run by any workflow, so nothing here depends on it.

## Related

- #2465, this is step 1 of 2; the credential removal PR follows
- #2375, parent audit
- #2357, security index

---

<details>

</details>


abhinav-m22:
Hi @AayushSaini101, @wilsonwu the PR is ready for review. Could you please take a look?

dashboard/backend/handlers/openclaw_websocket.go
```diff
@@ -70,9 +72,9 @@ type WSClient struct {
 var wsUpgrader = websocket.Upgrader{
 	ReadBufferSize:  1024,
 	WriteBufferSize: 1024,
-	CheckOrigin: func(r *http.Request) bool {
-		return true // Allow all origins for now
-	},
+	// CORS does not cover handshakes, so this is the only cross-origin control. Every
+	// dashboard client builds its URL from window.location.host. See #2465.
+	CheckOrigin: auth.OriginChecker(nil),
```
FAUST-BENCHOU: The WebSocket upgrader always calls `OriginChecker(nil)`, so `DASHBOARD_ALLOWED_ORIGINS` is ignored here and an explicitly allowed split-origin frontend can still have its WebSocket rejected. Please wire the configured allowed origins into the upgrader as well.

FAUST-BENCHOU (commented):
Others lg

dashboard/backend/handlers/openclaw_websocket.go
abhinav-m22: @FAUST-BENCHOU Good catch, thanks. The upgrader was a package level var built at init, so it could never see `DASHBOARD_ALLOWED_ORIGINS`, which meant a split origin frontend could log in and POST fine but still fail to open the room socket.

Fixed in latest commit by moving the upgrader onto `OpenClawHandler` and wiring in the configured origins during route registration.

abhinav-m22:
Can we please mark the related issue https://github.com/vllm-project/semantic-router/issues/2465 as accepted for the ci to pass? @FAUST-BENCHOU @AayushSaini101

wilsonwu:
> Can we please mark the related issue https://github.com/vllm-project/semantic-router/issues/2465 as accepted for the ci to pass? @FAUST-BENCHOU @AayushSaini101 

Done

wilsonwu:
LGTM

FAUST-BENCHOU (approved):

wilsonwu (approved):

PR #2963, opened 2026-08-21
[Router] Add CAS semantics for distributed router session state

Sanchar127:
Related #2243

## Purpose

This PR adds compare-and-swap (CAS) semantics for distributed router session state.

It makes shared router session updates version-aware so stale concurrent writers cannot overwrite newer state.

### Changes

- Add version tracking to router session snapshots.
- Replace unconditional shared-store saves with version-aware `SaveIfVersion` CAS updates.
- Prevent stale concurrent updates from overwriting newer session state.
- Reconcile stale writers by loading the latest shared state and reapplying their session event.
- Apply the same CAS/reconciliation behavior to session decisions and usage updates.
- Add tests covering stale writes, concurrent Redis CAS updates, and stale-writer reconciliation.
- Update the router learning state store test for the new CAS interface.

## Test Plan

- `go test ./pkg/sessiontelemetry -count=1`
- `go test ./... -count=1`
- `git diff --check`

## Test Result

The focused session telemetry tests pass locally.

The Redis concurrency test runs when `REDIS_ROUTER_SESSION_TEST_ADDR` is configured; otherwise it is skipped.

---

<details>

</details>


Xunzhuo (changes requested):
Thanks for working on the distributed session-state race. Versioned Redis CAS is the right direction, but the current reconciliation path still has correctness issues that can corrupt or lose session telemetry, so I don't think this is ready to merge yet.

1. **Concurrent writes on the same replica can double-apply an event.** `RecordSessionDecision` / `RecordSessionUsage` release the local session lock before persistence, and persistence then reads the current full local snapshot. A later local request can therefore already be included in the snapshot that an earlier writer saves. If the other writer then loses CAS, it reloads that snapshot and reapplies its event even though the event is already present. A deterministic two-goroutine regression produced `TurnCount == 3` and two `model-a` turns from only two decision events. The usage path has the same shape, so token and cost totals can also be double-counted.

2. **A successful rebase does not reconcile the local cache.** `SaveIfVersion` increments `snapshot.Version` only in its value-receiver copy, while the caller passes `rebased.Version` (still `latest.Version`) to `updateRouterSessionVersion`. More importantly, only the version is updated; fields merged from the remote snapshot are never hydrated locally. Since `GetRouterSessionSnapshot` returns the local entry without consulting Redis whenever that entry exists, routing can continue indefinitely with incomplete model-turn, switch, usage, and cost history. A production-path regression left Redis at version 2 with the other replica's `model-a` history, while the local entry remained version 1 and had no `model-a` history.

3. **A second CAS conflict silently drops the event.** Each path performs one initial CAS and one reload/retry. If another replica wins between the reload and the retry, the function returns with no further retry or diagnostic, and the current decision/usage update is lost. This is readily possible with three or more replicas.

Please consider serializing mutation/persistence per session (or persisting an explicit idempotent event/delta), retrying CAS within a bounded timeout, and updating the local cache from the committed merged snapshot without overwriting newer local operations. The tests should cover same-replica concurrent writes, a second conflict with three writers, and full local/shared equality after both decision and usage rebases.

There is also a CI coverage gap: `TestRedisRouterSessionStateStoreConcurrentCAS` skips unless `REDIS_ROUTER_SESSION_TEST_ADDR` is set, but no repository workflow currently sets that variable. The green memory-integration job therefore does not exercise this Redis CAS test. Please wire a real Redis service into the relevant test job.

Finally, this addresses only the shared-hot-state portion of #2243. The parent issue should remain open for the deployment-mode design, sticky-routing guidance, and diagnostics acceptance criteria.

Xunzhuo (changes requested):
The updated branch still implements #2243 while that issue remains needs-acceptance, and the PR only marks it as Related. Please wait for the distributed-state contract to be accepted and make the accepted issue the implementation contract before another review.

Issue #2479, opened 2026-07-12
cli: contain runtime state files under the owned directory


Xunzhuo:
## Summary

Reuse canonical stack-name normalization for runtime state files and enforce direct-child containment and private atomic writes. Owner surfaces: `src/vllm-sr/cli/runtime_stack.py` and `src/vllm-sr/cli/commands/runtime_paths.py`.

## Current behavior

Container and service namespaces normalize the stack name, but runtime configuration filenames use the raw environment value. Names can therefore select inconsistent paths, create missing intermediate directories, or leave the intended state directory.

## Expected behavior

One normalized stack identity is used for every namespace, the final state path is a direct child of the owned directory, collisions are handled deterministically, and writes are atomic with mode `0600`.

## Acceptance

- Reuse `normalize_stack_name` for runtime filenames.
- Resolve and prove direct-child containment beneath the owned state directory.
- Define deterministic behavior for normalization collisions and invalid or empty names.
- Use same-directory atomic replacement and private file mode.

## Validation

- Add tests for spaces, separators, dot segments, Unicode, empty and long values, collisions, and pre-existing links.
- Assert direct-child containment, atomic replacement, and `0600` mode.
- Run `cd src/vllm-sr && python -m pytest tests`.
- Run the CLI changed-surface gates reported by `make agent-report`.

## Related

- Parent audit: #2375
- This is a P2 CLI boundary follow-up.

iroiro147:
I would like to take this. I will reuse normalize_stack_name for runtime filenames, enforce direct-child containment, use same-directory atomic replacement with mode 0600, and add the requested edge-case and symlink coverage before opening a PR.

Issue #2468, opened 2026-07-12
[Security] Constrain OpenClaw image and skill provisioning inputs


Xunzhuo:
## Summary

Make OpenClaw skills catalog-owned, contain every workspace operation, and enforce administrator-controlled image provenance and least privilege. Owner surfaces: `dashboard/backend/handlers/openclaw.go`, `dashboard/backend/handlers/openclaw_skills.go`, `dashboard/backend/handlers/openclaw_provision.go`, and `dashboard/backend/handlers/openclaw_helpers.go`.

## Current behavior

Provisioning combines caller-selected image references with external skill identifiers used in filesystem operations, then starts the selected image with broad runtime privileges. Existing sanitization does not cover the skill boundary.

## Expected behavior

Skills are selected only by strict server-catalog IDs, filesystem destinations cannot leave the owning workspace, images satisfy an administrator policy, and provisioned containers run with least privilege.

## Acceptance

- Use a strict server-owned skill catalog ID and never an external path segment.
- Prove destination containment after symlink-aware resolution and reject aliases or unknown IDs.
- Copy known assets by server identity instead of extracting content through an arbitrary image command.
- Enforce administrator-controlled image allowlist, digest, and provenance policy.
- Run non-root with a read-only root filesystem and minimal capabilities.

## Validation

- Add containment, separator, alias, symlink, unknown-ID, and image-policy negative tests without publishing host-specific paths.
- Add deployment tests for non-root identity, read-only root filesystem, and capability set.
- Run `cd dashboard/backend && go test -race ./handlers`.
- Run the affected OpenClaw E2E profile selected by the repo harness.

## Related

- Parent audit: #2375
- Security index: #2357
- Dashboard state history: #1609

zdy782:
I'd like to take a bounded first slice of this accepted issue.

The current `fetchSkillContent` constructs container paths from a requested skill ID before looking it up in `loadSkills()`. My proposed first PR will reject malformed, path-like, alias, and unknown skill IDs against the server-owned catalog before any provisioning side effects, and cover the public provisioning handler with positive and negative regression tests. Exact catalog IDs remain the public input contract; invalid inputs will not be silently normalized.

This first slice will use `Related #2468`, not close the issue. Symlink-aware workspace containment, catalog-owned content retrieval, administrator image provenance, and least-privilege container deployment remain separate completion requirements. I'll keep the PR's evidence and limitations explicit.

Xunzhuo:
[...]

PR #2985, opened 2026-08-23
[Security] Isolate local stack data services on a dedicated network

### LINKED ISSUE #2485: [Security] Generate per-stack storage credentials and isolate local data services
## Summary

Replace the local split runtime's shared storage credentials with private,
per-stack runtime secrets and isolate data services from unrelated workload
containers.

Parent audit: #2375

## Current behavior

The CLI-owned Postgres service and generated replay configuration use one
checked-in password, while the CLI-owned Redis service starts without
authentication. A stack-scoped Docker network and loopback-only host
publication reduce exposure, but they are not a credential or workload
isolation boundary. Auxiliary containers that join the same network should not
implicitly gain read/write access to replay, workflow, cache, or response data.

The host-publication defect is being fixed by the #2375 proof PR. This issue
owns the independent storage-secret lifecycle and east-west isolation work.

## Required behavior

- Generate independent CSPRNG credentials per stack for Postgres and Redis;
  do not retain a repository default or silently fall back to one.
- Persist recovery material through atomic, mode-`0600`, stack-contained state
  owned by the CLI, with an explicit rotation and migration lifecycle.
- Enable Redis ACL/password authentication and inject both stores' credentials
  into router consumers without printing them, putting them in command output,
  or serializing them into public configuration or report artifacts.
- Keep data services loopback-only at the host and place them on a network that
  unrelated sidecars and user-selected workloads do not join by default.
- Reuse existing volumes without losing data, fail closed on missing or
  malformed secret state, and provide a documented recovery path.
- Treat any previously shared value as compromised during migration and revoke
  it after all consumers have switched.

## Acceptance

- Fresh-stack, restart, multi-stack, port-offset, migration, rotation,
  malformed-state, interrupted-write, and rollback tests pass.
- Command/config/log snapshots contain no secret value and storage files are
  private on disk.
- A container on the application/workload network cannot authenticate to the
  data services; the intended router consumers can.
- Redis and Postgres integration receipts prove authenticated read/write and
  rejection with missing or stale credentials.
- The CPU local feature gate and one AMD split-runtime receipt pass without
  exposing the credentials in CI or issue output.

## Related

- #1146 — Replay confidentiality and access control.
- #2468 — OpenClaw provisioning and workload trust.
- #2479 — CLI runtime state path containment.

drivebyer:
Closes #2485

## Purpose

Publishing the storage ports on `127.0.0.1` closed the north-south half of #2485. East-west stayed open: Redis, Postgres, and Milvus shared one bridge network with Envoy, Dashboard, the simulator, the observability containers, and every user-selected OpenClaw workload.

- **Owner** `wg/enterprise-environment` · **Modules** CLI and docs · **Go core** unchanged
- **PR 2 of 2** for #2485, on top of #2983 (per-stack credentials). Together they cover every bullet under **Required behavior**, so this one closes the issue. Milvus credentials and runtime-socket trust (#2468) stay out of scope.

### Changes

- **A second per-stack bridge network.** `data_network_name` (`vllm-sr-data-network`, prefixed for a named stack) carries Redis, Postgres, and Milvus; `network_name` keeps everything else. Router is the only container on both.
- **Router is created, not run.** A runtime takes one `--network` at creation, so the sequence is create on the application network, connect the data network, start — and the order is forced, since `postgres.OpenDB` pings on open and attaching after start is a race.
- **Reuse is the migration.** A #2983 stack has its stores on the application network and is never rebuilt, so a later `serve` reaches them through reuse. That path now connects the data network and disconnects the application one; a disconnect that genuinely fails stops the serve.
- **`stop` covers both networks**, including OpenClaw workloads — `OPENCLAW_DEFAULT_NETWORK_MODE` is settable, so a leftover attachment would leave a network Podman refuses to remove.
- **Docs**: a network-layering subsection in `security-hardening.md`. `container_data_network.py` and `container_start_runner.py` were extracted to keep `container_start.py` under the 800-line structure limit.

A caller that can reach the container runtime can still attach anything to any network. The split bounds workloads, not socket holders — that is #2468.

### Acceptance items deliberately not covered

Rather than leave the boxes silently unticked: the **interrupted-write test** exercises `_atomic_write_bytes`, which #2479 added and already tests; the **rollback test** cannot run an older CLI from the test suite, so the manual reset is written up in the hardening guide instead; the **AMD split-runtime receipt** covers a path that does not touch storage credentials or networks; and the **integration receipts** below are manual, since the CLI integration workflow does not provision storage. Happy to add any of these.

## Test Plan

```bash
cd src/vllm-sr && python -m pytest tests -q
python tools/agent/scripts/run_agent_precommit_lint.py $(git diff --name-only HEAD~1)
```

Plus a real Docker daemon, since unit tests cannot demonstrate a network boundary: probe both stores from each network, with a control container genuinely on the application network to prove the probe works at all; then migrate a stack built the old way.

## Test Result

**1189 passed, 22 failed, 1 skipped.** The 22 are pre-existing on a clean `main` checkout here (macOS + Python 3.14) and the failing set is byte-identical before and after. The repo's changed-files gate exits 0. On macOS a catch-all resolver makes the blocked probes fail at connect rather than at DNS.

```text
redis / postgres networks              -> ['<stack>-vllm-sr-data-network']   (data only)

from the APPLICATION network
  control redis, genuinely on it       -> PONG                    (the probe works)
  managed redis                        -> Error: Server closed the connection
  managed postgres                     -> server closed the connection unexpectedly

from the DATA network
  managed redis                        -> PONG
  managed postgres                     -> AUTH_OK

migration of a stack built the old way
  before serve  redis / postgres       -> ['<stack>-vllm-sr-network']
  after  serve  redis / postgres       -> ['<stack>-vllm-sr-data-network']
  redis canary                         -> 77
```

> **Known follow-up, not a regression here**: `recipe_topology_clone` uses `run -d` and attaches secondary networks afterwards — the ordering this PR fixes for `serve`. Router now has a secondary network, so Recipe activation reopens that window. The 5-second connect timeout makes it a latent race rather than a break, and fixing it means splitting `_validate_router_management_isolation` across the create/start boundary on a memfd-gated path.

---
<details>

</details>


src/vllm-sr/tests/test_storage_network_isolation.py
```diff
@@ -0,0 +1,550 @@
+"""Which containers can reach this stack's data services over the network.
+
+Publishing the storage ports on loopback only closes the north-south half of
+the exposure. These tests cover the east-west half: the stores sit on a second
+bridge network that Envoy, Dashboard, the simulator, the observability
+containers, and any user-selected OpenClaw workload never join, and Router is
+the single container attached to both.
+
+The ordering assertions are the point of several of these tests, not a
```
ramkrishs, line 9: Since the network boundary here can only be verified with real Docker per the manual test plan, is it worth filing a follow-up issue to get that verification into CI so it doesn't silently regress?

ramkrishs (approved):
LGTM

src/vllm-sr/tests/test_storage_network_isolation.py
drivebyer: Agreed — filed as #3010.

It is not in this PR because the CLI E2E suite never provisions storage: `cli_test_base.py` knows the Redis and Postgres container names only as `AUXILIARY_CONTAINER_NAMES`, for cleanup. Making a CI probe possible is a change to the harness, not an added test case.

PR #2986, opened 2026-08-23
[Bug] Honor the target layer for embedding routing

### LINKED ISSUE #2914: [Bug] target_layer parameter does not work
### Describe the bug

We build the router for us clawith, no matter how we switch the parameter target layer, the latency remains 200+ms in the whole router. However, once we select the onnx to have one-time experiment, model with 7 layers may have about 60+ms latency while the 22layers one have 200+ms so we guess that parameter in router.yaml does not work, since I am not familiar with go, my agent finds some evidence which may be helpful to fix it :
Source-level trace (main @ dcfe31d)
1. Go server routing path (`pkg/classification/embedding_classifier_backend.go:189`) calls
   `getEmbeddingWithModelType(text, modelType, c.optimizationConfig.TargetDimension)` — a **3-arg**
   function that never carries a layer.
2. Go FFI wrapper `candle-binding/semantic-router.go`:
   ```go
   // L235: C.get_embedding_with_model_type(text, model_type, target_dim)          // no layer
   // L236: C.get_embedding_2d_matryoshka(text, model_type, target_layer, target_dim) // with layer
   // L1488-1496:
   func GetEmbeddingWithModelType(text, modelType, targetDim) {
       // modelType == "mmbert" → delegate with hard-coded 0
       return GetEmbedding2DMatryoshka(text, modelType, 0, targetDim) // 0 → full 22 layers
   }
   ```
3. Rust candle binding (`candle-binding/src/mmbert_embedding.rs`):
   ```rust
   let target_layer = target_layer.unwrap_or(num_layers); // None/0 → 22 layers (config.json num_hidden_layers)
   ```
4. `HNSWConfig.TargetLayer` (`pkg/config/model_config_types.go`) is only assigned a default in
   `canonical_defaults.go` and validated in `/embeddings` API handler
   (`pkg/apiserver/route_embeddings.go`) — it is never wired into the router classification path.
5. The only place that passes a real layer today is `pkg/cache/inmemory_cache.go` (hard-coded 6)
   and the request-level `/api/v1/embeddings` API.

So: the capability is fully implemented (4-arg FFI, Rust early-exit, config field exists), but the
routing path drops the parameter before it reaches candle.

### To Reproduce

  model_catalog:
    embeddings:
      semantic:
        mmbert_model_path: models/mmbert-embed-32k-2d-matryoshka
        use_cpu: true
        embedding_config:
          model_type: mmbert
          preload_embeddings: true
          target_dimension: 768
          target_layer: 22
          enable_soft_matching: true
          top_k: 1
          min_score_threshold: 0.5
we use the same parameter as the docs

### Expected behavior

_No response_

### Affected layer

_No response_

### Additional context

_No response_

mikemikimike:
## Summary
- pass the configured `embedding_config.target_layer` through the classifier's Candle embedding path
- use the existing 2D Matryoshka binding so mmBERT early-exit layers are honored
- add a regression test that captures the layer sent to the backend

## Root cause
`EmbeddingClassifier.computeEmbedding` called `GetEmbeddingWithModelType`, which always delegated to `GetEmbedding2DMatryoshka` with layer `0` (full model). `HNSWConfig.TargetLayer` was therefore never used by router classification, even though the configuration and FFI support already existed.

## Validation
A focused Go test will be run against `pkg/classification` once the complete source checkout is available. The regression test uses the package's existing backend seam and does not require a model, GPU, API key, or external service.

Closes #2914


Xunzhuo (commented):
The production change is directionally correct: forwarding HNSWConfig.TargetLayer to the existing 2D Matryoshka binding addresses the reported root cause. This is not merge-ready yet, though:

1. The new backend seam is not migrated everywhere. TestInitializeRuntimeWarmsEmbeddingCandidatesAfterBackendInit in classifier_lifecycle_test.go still stubs getEmbeddingWithModelType, while EmbeddingClassifier now calls getEmbedding2DMatryoshka. That test can fall through to the real Candle backend instead of validating the intended init-before-warmup contract. Please migrate the remaining tests/users of the old seam to the new four-argument seam and run the full classification package tests.

2. Core quality / Full pre-commit checks has failed twice on gofumpt. The two latest style commits are empty, and the PR diff still has the two declarations outside a var block plus no blank line between the new test and the following test. Please apply gofumpt to both changed files and verify the formatting changes are actually staged and pushed.

After these are fixed, please rerun the classification/core tests and wait for all required checks to pass.

FAUST-BENCHOU (commented):
pls fix ci others lg

mikemikimike:
The remaining failure appears to be transient. The Recipe conformance shard failed because Hugging Face returned a 504 Gateway Timeout while downloading \llm-semantic-router/mmbert32k-intent-classifier-merged\; the subsequent missing-file error is only a consequence of that download failure. All core tests, pre-commit checks, DCO, and E2E checks passed. Could a maintainer please re-run the failed jobs when convenient? Thank you!

wilsonwu (approved):

mikemikimike:
Thanks @wilsonwu and @FAUST-BENCHOU for the review and CI follow-up. The remaining seam/formatting issues were resolved on the branch, the required checks completed successfully, and Mergify merged this PR at `26e471b5ba6270c86dd2b23bbce3ac73c4d64dc2`. I’m closing the loop here since I had not replied to the review thread earlier.

PR #2991, opened 2026-08-23
[Bug] Preserve Anthropic backend error responses

### LINKED ISSUE #2956: [Bug] Anthropic backend error bodies replaced with empty success-shaped responses for both client protocols
### Describe the bug

When an Anthropic-format backend returns an error envelope
(`{"type":"error","error":{...}}`), the router forwards the upstream HTTP
status correctly but replaces the body with a syntactically valid, completely
empty success-shaped object. The upstream error message and error type are
unrecoverable from the client side. Any error the backend emits in normal
operation triggers it: 429 rate limits, 529 overloaded, 401 authentication,
unknown/retired model IDs, malformed requests. Streaming requests are
affected identically, because these errors occur before any SSE starts, so the
upstream returns a plain JSON error body that takes the same translation
path.

For OpenAI-protocol clients the body becomes an empty `chat.completion`
(empty `id`, empty `content`, `finish_reason: "stop"`, all-zero usage).
This violates the OpenAI error contract (`{"error": {"message": ..., "type": ..., "code": ...}}`).
Status-based retry logic still works, but any SDK that parses the error body for the failure reason
gets nothing, and operators lose the upstream diagnostic (`invalid x-api-key`,
model not found, rate-limit details).

#### Anthropic-protocol clients are affected too

The error body is translated twice (error envelope → empty OpenAI completion →
Anthropic emit), so a `/v1/messages` client receives an empty `message` object
instead of the error envelope its SDK expects:

```json
{"id":"","type":"message","role":"assistant","model":"claude-haiku-4-5-20251001",
 "content":[],"stop_reason":"end_turn","stop_sequence":null,
 "usage":{"input_tokens":0,"output_tokens":0,
          "cache_creation_input_tokens":0,"cache_read_input_tokens":0}}
```

A Claude Code user with a bad key sees an empty response instead of the
authentication error. For streaming requests the error surfaces as this
plain JSON body rather than SSE. A fix should cover both surfaces.

### To Reproduce

1. Configure a model with `api_format: anthropic` and an Anthropic-type backend.
2. Make the upstream return any error — easiest: start the router with an invalid
   Anthropic API key, or use an invalid upstream model ID.
3. Send any `POST /v1/chat/completions` through the router.
4. With an invalid key the client receives HTTP 401 with body:
   ```json
   {"id":"", "object":"chat.completion",
    "choices":[{"finish_reason":"stop","message":{"content":"","role":"assistant"}}],
    "usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
   ```
   The `authentication_error` message from the upstream envelope is gone.
   Router log confirms the error body went through response translation:
   `Transformed Anthropic response to OpenAI format, original size: 106,
   transformed size: 679`.
5. Send the same request as `POST /v1/messages` (Anthropic shape): the client
   receives HTTP 401 with the empty `message` body quoted above instead of the
   error envelope.

### Expected behavior

The error should survive translation with the upstream status. For the invalid-key example,
an Anthropic-protocol client gets the upstream envelope unchanged (HTTP 401):

```json
{"type":"error","error":{"type":"authentication_error","message":"API key is invalid."}}
```

and an OpenAI-protocol client gets the same error in OpenAI shape (HTTP 401):

```json
{"error":{"message":"API key is invalid.","type":"authentication_error"}}
```

The body of a non-2xx response should never be a success-shaped completion or
message.

### Affected layer

global level

### Additional context

`toOpenAIResponseBody` unmarshals the error envelope into `anthropic.Message`,
which succeeds with zero values, and never checks for `"type": "error"`:

```go
// src/semantic-router/pkg/anthropic/client.go:163
var resp anthropic.Message
if err := json.Unmarshal(anthropicResponse, &resp); err != nil {
    return nil, fmt.Errorf("failed to parse Anthropic response: %w", err)
}
```

The Response API filter already has the guard this path is missing
(`isResponseAPIErrorBody`,
`src/semantic-router/pkg/extproc/req_filter_response_api.go:196`), and
`EmitAnthropicError` exists for emitting errors in the reverse direction.

Related precedent: #2112 (unknown models surfaced a misleading "401 No api
key" — same error-transparency concern on this route). Unlike #2130, the
status code is preserved here, so the impact is lost diagnostics and a
contract-violating error body rather than a false success.

mikemikimike:
## Summary

Fixes #2956.

Anthropic backend error envelopes were decoded as zero-valued successful messages during response normalization. This discarded the upstream error type and message for both protocol clients.

### Changes

- Detect Anthropic `{"type":"error","error":...}` envelopes before success-message decoding.
- Preserve the original envelope for Anthropic-protocol clients.
- Convert the envelope to the OpenAI `{"error":...}` shape for OpenAI-protocol clients.
- Keep the existing successful response conversion path unchanged.

### Validation

- Added helper tests for both response directions.
- Added response-normalization tests using the real `normalizeProviderResponseBody` entry point for both client protocols.
- Added malformed JSON coverage.
- `gofmt` and `git diff --check` pass.
- Anthropic focused tests pass locally.
- Extproc focused tests are blocked locally because this checkout lacks the repository replacement directories `candle-binding`, `nlp-binding`, and `ml-binding`; the PR's remote core test/build, unit test, lint, pre-commit, and E2E checks passed.

### DIFF
--- a/src/semantic-router/pkg/anthropic/client.go
+++ b/src/semantic-router/pkg/anthropic/client.go
@@ -151,6 +151,29 @@ func ToOpenAIResponseBodyWithExt(anthropicResponse []byte, model string, ext *ir
 	return toOpenAIResponseBody(anthropicResponse, model, ext)
 }
 
+// IsErrorBody reports whether body is an Anthropic API error envelope.
+// Error responses must not be decoded as zero-valued successful messages.
+func IsErrorBody(body []byte) bool {
+	var envelope struct {
+		Type  string          `json:"type"`
+		Error json.RawMessage `json:"error"`
+	}
+	return json.Unmarshal(body, &envelope) == nil &&
+		envelope.Type == "error" && len(envelope.Error) > 0
+}
+
+func toOpenAIErrorBody(body []byte) ([]byte, error) {
+	var envelope struct {
+		Error json.RawMessage `json:"error"`
+	}
+	if err := json.Unmarshal(body, &envelope); err != nil {
+		return nil, err
+	}
+	return json.Marshal(struct {
+		Error json.RawMessage `json:"error"`
+	}{Error: envelope.Error})
+}
+
 // toOpenAIResponseBody is the internal form that exposes the
 // IRExtensions side channel. It is called by ToOpenAIResponseBody with a
 // nil ext to preserve the existing inverse cell byte-for-byte; the
@@ -159,6 +182,9 @@ func ToOpenAIResponseBodyWithExt(anthropicResponse []byte, model string, ext *ir
 // flattening that the OpenAI envelope cannot represent.
 func toOpenAIResponseBody(anthropicResponse []byte, model string, ext *ir.IRExtensions) ([]byte, error) {
 	logging.Debugf("Raw Anthropic response: %s", logging.ContentDescriptorBytes(anthropicResponse))
+	if IsErrorBody(anthropicResponse) {
+		return toOpenAIErrorBody(anthropicResponse)
+	}
 
 	var resp anthropic.Message
 	if err := json.Unmarshal(anthropicResponse, &resp); err != nil {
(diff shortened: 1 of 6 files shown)


FAUST-BENCHOU (approved):
LG.

wilsonwu (approved):

PR #3008, opened 2026-08-24
[E2E] pin anthropic-shim Kind image to Never and lock LocalImages

### LINKED ISSUE #2809: [Bug] anthropic-shim e2e profile can fall back to pulling an unpublished image tag
### Proposed Workgroup

Evaluation & Quality

### Describe the bug

**Updated 2026-08-25.** #2835 fixed the substance of the original first claim
(`LocalImages` is registered and `kind load`ed now), and the second — that the profile
never runs in CI — turned out to be a deliberate classification, not an oversight (see
Non-goals). What is still true on `main`:

1. [`backend.yaml`](https://github.com/vllm-project/semantic-router/blob/main/e2e/profiles/anthropic-shim/gateway-resources/backend.yaml#L159-L161)
   sets `imagePullPolicy: IfNotPresent` on the `anthropic-shim` container, but
   `.../anthropic-shim:e2e-test` is only ever built locally (GHCR publishes `:nightly` /
   `:latest`, never `:e2e-test`). A skipped or renamed image load therefore degrades into
   `ImagePullBackOff` and a setup timeout instead of a fast `ErrImageNeverPull`. The
   dashboard profile already pins its fixture with `Never`.
2. Nothing ties the manifest tag to the built tag: `backend.yaml` and the `Tag` in
   [`images.go`](https://github.com/vllm-project/semantic-router/blob/main/e2e/profiles/anthropic-shim/images.go)
   agree only by convention, so a rename on either side reintroduces the failure untested.
3. [`deploy/kubernetes/anthropic-backend/base/deployment.yaml`](https://github.com/vllm-project/semantic-router/blob/main/deploy/kubernetes/anthropic-backend/base/deployment.yaml#L125)
   still carries `image: anthropic-shim:latest` — unqualified, so it resolves to
   `docker.io/library/anthropic-shim` and can never be pulled. Nothing applies this
   kustomize tree; it is an unreferenced copy of the Deployment the profile really uses.
4. `ci-changes.yml` exports an `e2e_anthropic_shim` output
   ([L100-L101](https://github.com/vllm-project/semantic-router/blob/main/.github/workflows/ci-changes.yml#L100-L101),
   [L160](https://github.com/vllm-project/semantic-router/blob/main/.github/workflows/ci-changes.yml#L160))
   that no workflow consumes; profile selection moved to `e2e_profiles`.

### To Reproduce

- (1) Skip `kind load`, or build the shim under another tag, then
  `make e2e-test E2E_PROFILE=anthropic-shim` on a fresh Kind cluster — setup blocks in
  `ImagePullBackOff` until it times out.
- (3) `kubectl apply -k deploy/kubernetes/anthropic-backend/overlays/qwen` — the shim
  container never starts.
- (4) `grep -rn e2e_anthropic_shim .github/` — every hit is the export and its own plumbing.

### Expected behavior

- `imagePullPolicy: Never` on the shim container, matching the dashboard Kind contract.
- A test asserting the `backend.yaml` image equals the `Tag` from
  `anthropicshim.LocalImages()`.
- One anthropic-backend Deployment manifest, or the unreferenced copy corrected or removed.
- No `e2e_anthropic_shim` export that nothing consumes.

### Non-goals

**Running this profile in PR CI is out of scope here.**
[`test-domain-registry.yaml`](https://github.com/vllm-project/semantic-router/blob/main/tools/agent/test-domain-registry.yaml#L446-L455)
classifies it `selection: manual`, `classify_pr_changes.py` lists it under
`NON_PR_E2E_RULES`, and two tests lock that. It downloads a GGUF model and runs
`llama.cpp`, so promoting it is a cost/flake trade-off for the Workgroup, not a bug fix —
and any scheduled run should first pin the floating `ghcr.io/ggml-org/llama.cpp:server`
tag. That belongs in its own issue.

### Impact

E2E only, and the profile does come up on Kind today: this is about failure modes and
drift resistance, not a live outage.

### Additional context

#3008 addresses item (1). Items (2)-(4) remain open; (2) is a suggested change to the test
that PR adds.

Tejaacavale:
Closes #2809

## Purpose

- Kind fixture `.../anthropic-shim:e2e-test` is not published on GHCR. A missed `kind load` with `IfNotPresent` becomes `ImagePullBackOff`. Pin the e2e shim container to `imagePullPolicy: Never` (dashboard Kind contract). Public images (`python:3.11-slim`, `llama.cpp:server`) stay `IfNotPresent`.
- `#2835` already registers `LocalImages()`. This PR does not re-register it.
- Lock `backend.yaml`'s shim `image` to `anthropicshim.LocalImages().Tag` so those two strings cannot drift untested.
- Qualify the unused kustomize copy (`deploy/kubernetes/anthropic-backend/base/deployment.yaml`) from `anthropic-shim:latest` to `ghcr.io/vllm-project/semantic-router/anthropic-shim:latest`. Tree kept (fix, not delete). `apply -k` already rewrote this via `images:`; this makes the source file pullable too.
- Drop the unused `e2e_anthropic_shim` workflow export from `ci-changes.yml`. Classifier tests / `NON_PR_E2E_RULES` unchanged. This profile stays out of PR Kind.
- Affects: `E2E`

## Test Plan

- `cd e2e && go test ./profiles/all/ -count=1 -run 'Test(Dashboard|EnvoyAIGateway|AnthropicShim)'`
- `.venv-agent/bin/python -m unittest tools.ci.tests.test_pr_change_classifier`
- `grep e2e_anthropic_shim .github/workflows/ci-changes.yml` (no hits)

## Test Result

- Profile unit tests passed.
- Classifier tests passed; `make agent-validate` passed.
- Full `make e2e-test E2E_PROFILE=anthropic-shim` not re-run here (extproc/GGUF). Item 1 was already checked with `docker manifest inspect` (`:e2e-test` unknown) and a Kind event that the local tag was already on the node.

---

<details>

- [x] PR title uses module-aligned prefixes such as `[E2E]`
- [x] Purpose, Test Plan, and Test Result match this diff

</details>

### DIFF
--- a/.github/workflows/ci-changes.yml
+++ b/.github/workflows/ci-changes.yml
@@ -97,8 +97,6 @@ on:
         value: ${{ jobs.classify.outputs.e2e_authz_rbac }}
       e2e_streaming:
         value: ${{ jobs.classify.outputs.e2e_streaming }}
-      e2e_anthropic_shim:
-        value: ${{ jobs.classify.outputs.e2e_anthropic_shim }}
       e2e_common:
         value: ${{ jobs.classify.outputs.e2e_common }}
 
@@ -157,7 +155,6 @@ jobs:
       e2e_multi_endpoint: ${{ steps.matrix.outputs.e2e_multi_endpoint }}
       e2e_authz_rbac: ${{ steps.matrix.outputs.e2e_authz_rbac }}
       e2e_streaming: ${{ steps.matrix.outputs.e2e_streaming }}
-      e2e_anthropic_shim: ${{ steps.matrix.outputs.e2e_anthropic_shim }}
       e2e_common: ${{ steps.matrix.outputs.e2e_common }}
     steps:
       - uses: actions/checkout@v4
--- a/deploy/kubernetes/anthropic-backend/base/deployment.yaml
+++ b/deploy/kubernetes/anthropic-backend/base/deployment.yaml
@@ -122,7 +122,7 @@ spec:
                 - ALL
 
         - name: anthropic-shim
-          image: anthropic-shim:latest
+          image: ghcr.io/vllm-project/semantic-router/anthropic-shim:latest
           imagePullPolicy: IfNotPresent
           ports:
             - name: shim
--- a/e2e/profiles/all/imports_test.go
+++ b/e2e/profiles/all/imports_test.go
@@ -113,44 +113,72 @@ func anthropicShimContainerFromBackendYAML(t *testing.T) (string, corev1.PullPol
 
 	reader := utilyaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(raw)))
 	for {
-		doc, err := reader.Read()
+		doc, err := nextYAMLDocument(t, reader)
 		if err == io.EOF {
 			break
 		}
-		if err != nil {
-			t.Fatalf("read backend.yaml document: %v", err)
+		image, pullPolicy, found := anthropicShimFromYAMLDocument(t, doc)
+		if found {
+			return image, pullPolicy
 		}
-		if len(bytes.TrimSpace(doc)) == 0 {
-			continue
-		}
-
-		jsonDocument, err := utilyaml.ToJSON(doc)
-		if err != nil {
-			t.Fatalf("convert backend.yaml document to JSON: %v", err)
-		}
-
-		var typeMeta struct {
-			Kind string `json:"kind"`
-		}
-		if err := json.Unmarshal(jsonDocument, &typeMeta); err != nil {
-			t.Fatalf("decode backend.yaml type meta: %v", err)
-		}
-		if typeMeta.Kind != "Deployment" {
-			continue
-		}
-
-		var deployment appsv1.Deployment
-		if err := json.Unmarshal(jsonDocument, &deployment); err != nil {
-			t.Fatalf("decode backend.yaml deployment: %v", err)
-		}
-		for _, container := range deployment.Spec.Template.Spec.Containers {
-			if container.Name == "anthropic-shim" {
-				return container.Image, container.ImagePullPolicy
-			}
-		}
-		t.Fatal("anthropic-shim container not found in backend.yaml Deployment")
 	}
 
 	t.Fatal("Deployment not found in backend.yaml")
 	return "", ""
+}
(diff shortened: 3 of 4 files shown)


drivebyer:
Thanks for picking this up — the `docker manifest inspect` check and the Kind event you
captured are exactly the right evidence.

Note I rewrote #2809 today: #2835 had already fixed the substance of the original first
claim, so the description you worked from is out of date. Worth re-reading before the next
push.

The diff is right. `Never` matches the dashboard Kind contract, and `runner.go` always
`kind load`s the tag from `LocalImages()`, so a missing load now fails fast.

`Closes #2809` is broader than the diff, though. The rewritten issue has four items; this
covers the first. I'd rather the PR grow to cover all four than demote `Closes` — the rest
are small:

2. The new test compares `LocalImages` against a literal copied from `images.go`, so it
   only catches edits to `images.go`. The drift #2809 describes is `backend.yaml`'s image
   diverging from the built tag — parse the manifest and compare against `Tag` instead.
3. `deploy/kubernetes/anthropic-backend/base/deployment.yaml#L125` still has
   `image: anthropic-shim:latest` (unqualified, unpullable) in an unreferenced kustomize
   tree. Fix or delete it.
4. `ci-changes.yml` exports a dead `e2e_anthropic_shim` output. Remove only the
   workflow-level export — `tools/ci/tests/` still exercises the classifier signal.

Tejaacavale:
Thanks for the rewrite and the review. I'll keep `Closes #2809` and grow this PR for the remaining items.

Might need another day or two. I'll push once that's in and the branch is rebased onto `main`.

Xunzhuo (changes requested):
Please finish the remaining #2809 items before this PR closes it. The test still compares `LocalImages` against a copied literal instead of parsing `backend.yaml` and checking it against `Tag`; the stale unqualified deployment manifest and dead `e2e_anthropic_shim` workflow output are also still untouched. Ping me after the promised update.

Tejaacavale:
Still working through the remaining items — taking a bit longer than I said. Expect a push within a day.

Tejaacavale:
@Xunzhuo remaining items are in. Ready for another look.

Xunzhuo (approved):
Thanks

PR #3021, opened 2026-08-25
[Bug] Fix Response API streaming for Anthropic backends

### LINKED ISSUE #3013: [Bug] /v1/responses streaming emits Chat Completions SSE on Anthropic backends
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

When a Response API client sends `POST /v1/responses` with `"stream": true`
and routing selects a model with `api_format: anthropic`, the client
receives raw `chat.completion.chunk` SSE frames instead of Response API
events. The stream lacks `response.created`, `response.output_text.delta`,
and the terminal `response.completed`.

The same request without streaming works correctly, so the gap is specific
to the streaming dispatch path.

### To Reproduce

1. Configure the router with `response_api.enabled: true`
   (`store_backend: memory` is enough) and a single model with
   `api_format: anthropic` behind a catch-all decision.
2. Send a streaming Response API request:
   ```bash
   curl -sN http://127.0.0.1:8801/v1/responses \
     -H 'content-type: application/json' \
     -d '{"model":"claude-haiku-4-5-20251001","stream":true,"input":"say hi in three words"}'
   ```

   Observed output — Chat Completions chunks on a /v1/responses stream,
   with a raw untranslated Anthropic frame interleaved:

   ```
   data: {"id":"msg_011CeMZ…","choices":[{"delta":{"content":"Hi there",…}}],…,"object":"chat.completion.chunk",…}

   event: content_block_stop
   data: {"type":"content_block_stop","index":0}

   data: [DONE]
   ```

3. Send the same request without `"stream": true`: a correct Response API
   envelope comes back (`"object":"response"`, `"status":"completed"`,
   populated `output`). Request-side translation, routing, and the backend
   round-trip all work on this route; only the streaming response path is
   broken.
4. Real-client variant: point Codex CLI at the router
   (`model_providers.<x>.wire_api = "responses"`, `base_url` at the
   gateway) and run any prompt. Codex errors after 5 reconnects; the
   router log shows one `Transformed request for Anthropic API` line per
   reconnect — each a billed upstream call.

### Expected behavior

A streaming `/v1/responses` request receives Response API SSE events
(`response.created`, `response.output_item.added`,
`response.output_text.delta`, …, `response.completed`) regardless of the
backend's `api_format`.

The router already does this for OpenAI-format backends. Swapping only the
backend format under the same config, the identical request produces the
full event sequence and the same Codex CLI session completes successfully.

### Impact

Response API clients treat the missing `response.completed` as a transport
failure and retry, and every retry is a fresh billed upstream request. In
our test, one Codex CLI prompt produced 6 upstream Anthropic calls
(initial + 5 reconnects) and still failed:

```
ERROR: stream disconnected before completion: stream closed before response.completed
```

Streaming is unusable on the Response API × Anthropic-backend combination,
and each failure multiplies upstream cost. The only workaround is
disabling streaming client-side.

### Additional context

Root cause is dispatch ordering in `handleResponseBody`
(`src/semantic-router/pkg/extproc/processor_res_body.go:45-47`):

```go
// Legacy branch: OpenAI client hitting an Anthropic backend. …
if ctx.IsStreamingResponse && ctx.APIFormat == config.APIFormatAnthropic &&
    ctx.ClientProtocol != config.ClientProtocolAnthropic {
    return r.handleAnthropicStreamingResponseBody(v.ResponseBody.Body, ctx), nil
}
```

A Response API client has no dedicated `ClientProtocol` value
(Responses-ness lives in `ctx.ResponseAPICtx`), so it matches this branch,
and `handleAnthropicStreamingResponseBody` emits `chat.completion.chunk`
frames directly without ever consulting `isResponseAPIRequest`.

The Response API event translator only runs in the branch below:
`handleStreamingResponseBody` checks `isResponseAPIRequest(ctx)` and
applies `buildResponseAPIStreamingBodyMutation`
(`processor_res_body_streaming.go:23,40`). A likely fix shape: feed the
transformed chunks from the Anthropic streaming handler through that same
Response API streaming mutation instead of emitting them raw.

Related: #2446 reported this same leak generically; #2438 fixed it by
applying `buildResponseAPIStreamingBodyMutation` in
`handleStreamingResponseBody`, but the Anthropic streaming branch
dispatches before that handler and returns early, so the fix never covered
`api_format: anthropic` backends. The contract itself is already accepted
scope: #2492 lists "preventing upstream Chat Completions SSE frames from
leaking to `/v1/responses` clients" in its hardening criteria. Therefore this issue
reports a remaining gap in that contract, not a new behaviour proposal.

No e2e coverage exists for this combination: `e2e/pkg/testmatrix` has
Response API cases only against OpenAI-format backends, and the
Anthropic-shim contract has no `/v1/responses` case.

Verified on `main` @ 7b3764da (2026-08-25) with a live Anthropic backend
(`claude-haiku-4-5-20251001`) and Codex CLI 0.147.0.

csarushan1729:
Closes #3013

## Purpose
Response API clients streaming against `api_format: anthropic` backends were receiving leaked `chat.completion.chunk` SSE frames instead of Response API events (`response.created`, `response.output_text.delta`, `response.completed`, …).

Root cause: `handleAnthropicStreamingResponseBody` returns early with OpenAI-shaped chunks and never runs `buildResponseAPIStreamingBodyMutation`. Response API requests match the legacy Anthropic streaming branch because they are not `ClientProtocolAnthropic`.

This change routes Response API requests through the existing Response API streaming mutation after Anthropic → OpenAI SSE translation. Classic Chat Completions clients on Anthropic backends are unchanged.

Affected: `pkg/extproc` (Data Plane & Networking).

## Test Plan
- Code path review: Response API + Anthropic streaming hits `buildResponseAPIStreamingBodyMutation`; non–Response API path still emits OpenAI chunks.
- Local `go test ./pkg/extproc/` blocked on Windows (CGO / Valkey Glide module issues); rely on CI.
- Manual (if available): streaming `POST /v1/responses` with an Anthropic-format model should emit `response.*` events and end with `response.completed`.

## Test Result
- Change is compile-scoped to `processor_res_body_streaming_anthropic.go`.
- Full unit/e2e not run locally (Windows CGO + dependency constraints).
- Awaiting CI for package tests.
<details>


src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic.go
```diff
@@ -49,6 +49,17 @@ func (r *OpenAIRouter) handleAnthropicStreamingResponseBody(
 	ctx.HasStreamingChunks = true
 	r.parseStreamingChunk(chunkStr, ctx)
 
+	if isResponseAPIRequest(ctx) {
```
FAUST-BENCHOU, line 60: Response API handling happens after the `len(transformed) == 0` early return, so Anthropic-only events like `content_block_stop` can still pass through without a body mutation. Please consume these empty-transformation chunks for Response API requests to prevent raw Anthropic SSE leakage.

FAUST-BENCHOU (changes requested):

AayushSaini101:
@csarushan1729 can you please add test cases as well thanks

dorota-memex:
Heads up — I've opened #3045 for this (as assigned in #3013). It has unit and e2e tests verified against a live Anthropic backend; review over there is welcome if you're interested.

src/semantic-router/pkg/extproc/processor_res_body_streaming_anthropic.go
dorota-memex, line 37: This case has the same issue as fixed below. On `err != nil` the mutation returned being nil lets the raw Anthropic bytes go through unchanged, so /v1/responses clients would see untranslated frames. The same handling you added in the `len(transformed) == 0` branch below should cover it.

Xunzhuo (commented):
Thanks for working on this. This branch no longer applies cleanly, and the fix has since landed in #3045 with raw-frame reassembly before Anthropic translation plus unit and end-to-end coverage. Please close this PR in favor of #3045.

drivebyer:
This PR in favor of https://github.com/vllm-project/semantic-router/pull/3045

PR #3035, opened 2026-08-26
[Bug] Preserve full response_format through request parsing and mutations

### LINKED ISSUE #3024: [Bug] v0.3.0 auto-routing drops json_schema payload during model rewrite
## Summary

When a request names the auto-routing model (`auto_model_name`, e.g. `MoM`) and includes `response_format: {"type": "json_schema", ...}`, the schema contract is not enforced by the routed backend. The same request sent through the same router/Envoy path but naming the backend model explicitly IS schema-enforced. Observationally, `json_schema` through auto-routing behaves exactly like `json_object`: at `temperature: 0`, the auto-routed `json_schema` response is byte-identical to a `json_object` response, and required schema properties are absent. JSON-mode survives; the schema semantics are lost. The response is still HTTP 200 with plausible JSON, so the failure is silent to callers.

## Environment

- semantic-router v0.3.0 (latest release), Envoy + extproc deployment per the repo's compose config
- Backend: llama.cpp llama-server (build b1-f280b26), which enforces OpenAI `response_format: json_schema` correctly when called directly
- Routing: deterministic keyword/context rules; `auto_model_name: MoM`

## Reproduction

Identical request, `temperature: 0`, `max_tokens: 100`, prompt "Write one sentence about mountains.", schema requiring a single distinctive property:

```json
{"type": "json_schema", "json_schema": {"name": "probe", "strict": true, "schema": {"type": "object", "properties": {"zqx_answer": {"type": "string"}}, "required": ["zqx_answer"], "additionalProperties": false}}}
```

| Call | model | response_format | Result |
|---|---|---|---|
| A | `MoM` | `json_schema` (above) | JSON without `zqx_answer`; free-form keys |
| B | `MoM` | `json_object` | Byte-identical to A |
| C | explicit backend model, same router path | `json_schema` (above) | `{"zqx_answer": "..."}` — schema enforced |

A == B (byte-identical at temp 0) and C conforming shows the loss happens on the auto-routing path specifically, not in Envoy transport or the backend.

## Impact

Any consumer relying on OpenAI-strict structured output through the auto-routing alias is silently uncontracted. Downstream code that trusts the schema (strict field access, `additionalProperties: false`) fails unpredictably while every response looks like a success. In our deployment this broke a RAG query-plan classifier and answer-envelope pipeline; we had to pin those callers to explicit backend models as a workaround, losing auto-routing for them.

## Expected

`response_format` (including full `json_schema` semantics) should be preserved when the router substitutes the routed backend model for the auto-routing alias, or the limitation should be documented and/or surfaced as an error rather than a silent downgrade.

I could not find a config option in v0.3.0 that changes this behavior — happy to provide more traces or test a patch.

rishabhsinha17:
Closes #3024

## Purpose

`response_format: {"type": "json_schema", ...}` loses its entire schema payload (name, strict, schema) whenever the router re-serializes the parsed SDK request. Root cause: openai-go v1.12.0's `ChatCompletionNewParamsResponseFormatUnion` unmarshals by picking the first structurally matching variant, and the type-only text variant matches every `response_format` object. `json.Marshal` of the struct then emits `{"type":"json_schema"}` with no schema, and backends such as llama.cpp degrade to unconstrained JSON mode while still returning HTTP 200, which is exactly the silent downgrade reported in #3024.

Per the maintainer direction on the issue, the primary auto-routing rewrite is left untouched: since #2869 it rewrites only the model field in the original body and is already lossless. This PR makes the remaining struct-serialized paths lossless and adds the regression coverage:

- `restoreResponseFormat` (new `pkg/extproc/request_response_format.go`, called from `parseOpenAIRequest`) rebuilds the union from the raw request bytes for the documented `json_schema` / `json_object` / `text` types. Every downstream `json.Marshal` of the parsed request now keeps full `response_format` semantics: the `modifyRequestBodyForAutoRouting` serialize fallback, modality AR/omni request builders, looper candidate dispatch, Anthropic-inbound translation, and the tool-selection full-serialize fallback. Unknown type values keep the current SDK behavior.
- A non-object `json_schema` payload cannot be represented by the SDK variant and was previously dropped silently; it is now rejected at parse (surfaces as a 400) instead of forwarding a downgraded contract.
- Unit tests pin the parse/serialize round trip and every request mutation on the routing path (model rewrite, stream fields, reasoning mode, system prompt, request_params strip_unknown, memory injection) as lossless for the issue's strict json_schema probe, plus the auto-routing raw-body path end to end at the routing layer.
- New E2E contract `chat-completions-structured-output` (added to `BaselineRouterContract`) sends the issue's strict json_schema probe through the `MoM` auto model and verifies it routes to a well-formed completion. The simulator backend does not enforce schemas, so byte fidelity of the forwarded body is asserted by the unit tests; the E2E contract guards the transport behavior of this request shape.

Affected modules: `src/semantic-router/pkg/extproc`, `e2e`.
Workgroup: wg/data-plane-networking (issue #3024, accepted).

## Test Plan

- `go test ./pkg/extproc/ -run 'TestParseOpenAIRequest|TestRequestMutationsPreserveResponseFormat|TestModifyRequestBodyForAutoRoutingSerializePathKeepsJSONSchema|TestHandleAutoModelRoutingPreservesJSONSchemaResponseFormat' -v`
- `go test ./pkg/extproc/ ./pkg/looper/ ./pkg/anthropic/` (full packages, macOS arm64, locally built candle/ml/nlp bindings, SKIP_MILVUS/QDRANT/REDIS/VALKEY/LLAMA_STACK=true, SR_TEST_MODE=true)
- `go build ./...` and `go vet` for the `e2e` module (the new testcase compiles and registers; running the Kind profile was not possible in this environment)
- `gofmt -l` clean on all touched files; `go mod tidy` is a no-op for both modules

## Test Result

New tests:

```
=== RUN   TestParseOpenAIRequestRestoresJSONSchemaResponseFormat
--- PASS: TestParseOpenAIRequestRestoresJSONSchemaResponseFormat (0.00s)
=== RUN   TestParseOpenAIRequestResponseFormatVariants
=== RUN   TestParseOpenAIRequestResponseFormatVariants/json_object
=== RUN   TestParseOpenAIRequestResponseFormatVariants/text
=== RUN   TestParseOpenAIRequestResponseFormatVariants/payload_free_json_schema_stays_type_only
=== RUN   TestParseOpenAIRequestResponseFormatVariants/absent_response_format_untouched
=== RUN   TestParseOpenAIRequestResponseFormatVariants/unknown_type_left_to_sdk_behavior
--- PASS: TestParseOpenAIRequestResponseFormatVariants (0.00s)
=== RUN   TestParseOpenAIRequestRejectsNonObjectJSONSchemaPayload
--- PASS: TestParseOpenAIRequestRejectsNonObjectJSONSchemaPayload (0.00s)
=== RUN   TestRequestMutationsPreserveResponseFormat
=== RUN   TestRequestMutationsPreserveResponseFormat/model_rewrite
=== RUN   TestRequestMutationsPreserveResponseFormat/stream_fields
=== RUN   TestRequestMutationsPreserveResponseFormat/reasoning_mode
=== RUN   TestRequestMutationsPreserveResponseFormat/system_prompt
=== RUN   TestRequestMutationsPreserveResponseFormat/request_params_strip_unknown
=== RUN   TestRequestMutationsPreserveResponseFormat/memory_injection
--- PASS: TestRequestMutationsPreserveResponseFormat (0.00s)
=== RUN   TestModifyRequestBodyForAutoRoutingSerializePathKeepsJSONSchema
--- PASS: TestModifyRequestBodyForAutoRoutingSerializePathKeepsJSONSchema (0.00s)
=== RUN   TestHandleAutoModelRoutingPreservesJSONSchemaResponseFormat
--- PASS: TestHandleAutoModelRoutingPreservesJSONSchemaResponseFormat (0.00s)
PASS
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/extproc	1.640s
```

Full packages:

```
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/extproc	2.867s
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/looper	(cached)
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/anthropic	(cached)
```

Remaining risk: the Kind-based E2E profile was not executed locally (no cluster in this environment); `chat-completions-structured-output` compiles and is registered in `BaselineRouterContract` for CI. Behavior change to flag for review: a non-object `json_schema` payload now returns a 400 instead of being forwarded without its schema.

---


Xunzhuo:
@rishabhsinha17 please fix the CI

rishabhsinha17:
@Xunzhuo fixed in df91c6d0: gocognit flagged TestParseOpenAIRequestResponseFormatVariants in request_response_format_test.go at cognitive complexity 29 (limit 20). Extracted parse/serialize test helpers to drop the repeated error plumbing in the subtests. The agent lint gate, baseline pre-commit hooks, and pkg/extproc tests pass locally.

rishabhsinha17:
Rebased onto main, CI green locally; ready for another look @Xunzhuo

Xunzhuo (approved):
Looks good. The raw-body rehydration is limited to the documented variants, rejects an unrepresentable json_schema payload instead of silently downgrading it, and the mutation/E2E tests cover the reported path.

wilsonwu (approved):

PR #3047, opened 2026-08-26
[Bug] Normalize decision rule-tree operators at config load

### LINKED ISSUE #2937: [Bug] Decision rule-tree operators are not normalized consistently across producers
### Describe the bug

A decision rule-tree operator is not normalized the same way by the components that produce, evaluate and render it, so the same tree means different things depending on which one reads it.

**1. An omitted operator has two different defaults.**

| Component | Meaning of an omitted operator |
|---|---|
| Router — `pkg/decision/engine.go:193`, `trace.go:101` | `default:` branch → **OR** |
| Composer trees — `classification/classifier_composers.go:121` | `default:` branch → **AND** |
| CLI — `src/vllm-sr/cli/models.py`, `Rules.operator` | **AND** at the root; a nested node without an operator is rejected outright |
| DSL compiler — `pkg/dsl/compiler_routes.go:54,59` | emits **AND** |
| DSL decompiler — `decompiler_decisions.go:146` | joins with **AND** |
| Dashboard topology — `topology_response.go:221` | renders the empty string |

Every producer writes or displays `AND`; the component that actually routes traffic evaluates `OR`.

**2. Case is normalized by the evaluators but not by the DSL.**

`engine.go:193`, `trace.go:101` and `classifier_composers.go:106` all upper-case the operator, so `and` is a valid operator. The DSL decompilers switch on the raw string: `decompileRuleNodeToExpr` (`pkg/dsl/decompiler_convert.go`) matches nothing for `operator: or` and returns nil, so `route.When` is dropped and the emitted route matches **every** request; the text decompiler (`decompiler_decisions.go:97`) falls through to `decompileRuleFallback` and joins conditions with a literal ` or `, which is not valid DSL.

**3. Nothing is trimmed, and the DSL compiler validates nothing.**

No evaluator trims, so `operator: " AND"` reaches the default branch and silently becomes OR. And `compileComposerObj` (`pkg/dsl/compiler_routes.go:242-252`) copies `operator` into the config verbatim without validating it, so an invalid operator written in DSL source survives compilation and only fails when the emitted config is loaded.

### To Reproduce

Omitted-operator divergence:

1. Write a decision whose rule node omits the operator and has two conditions that are not both satisfied by one request:

```yaml
rules:
  conditions:
    - type: domain
      name: business
    - type: keyword
      name: urgent
```

2. Send a request matching only `business` — the decision matches (OR), while the docs, the CLI and the DSL all describe this tree as AND.

Case divergence:

3. Take any decision using `operator: or` (valid, the router upper-cases it) and decompile the config to DSL. The AST decompiler emits a route with no `WHEN` clause — a route that matches everything — and the text decompiler emits `a or b`, which does not recompile.

### Expected behavior

One normalization, applied once, that every surface agrees on:

- an omitted operator has a single documented meaning, either by normalizing it at config load so the evaluators' default branch becomes unreachable, or by requiring the operator explicitly on any combination node that has conditions (a scan of every rule tree under `config/**`, `deploy/**`, `e2e/**`, `src/**` found no combination node that omits it, so neither choice breaks a maintained config);
- the operator is trimmed and upper-cased once, at load, rather than by each consumer;
- the DSL compiler rejects an operator outside the closed set instead of deferring the failure to config load;
- decompiling is total: every tree config validation accepts decompiles into DSL text that recompiles to an equivalent tree, and no accepted tree reaches a fallback that emits a raw operator string or drops the `WHEN` clause.

### Affected layer

global level

### Additional context

Found while implementing #2122 (PR #2123). That PR validates the operator set, `NOT` arity and node shape in the Go loader, and adds a rule-tree truth-table corpus (`src/semantic-router/pkg/config/decision_rule_tree_corpus_test.go`) that is the natural fixture for a DSL round-trip test. It deliberately leaves the omitted-operator default alone, because choosing it is a semantic decision rather than a validation gap, and documents the ambiguity instead.

Part of the surface-parity ask in https://github.com/vllm-project/semantic-router/issues/2122#issuecomment-4951915446.

ramkrishs:
## Summary

Fixes #2937.

Rule-tree operators meant different things to different components:

| Component | Meaning of an omitted operator |
|---|---|
| Router evaluator (`pkg/decision`) | defaults to **OR** |
| Composer evaluator (`pkg/classification`) | defaults to **AND** |
| CLI (`src/vllm-sr/cli/models.py`) | defaults to **AND** |
| DSL compiler | emits **AND** |

Case was normalized by the evaluators (`strings.ToUpper`) but not by the DSL decompiler, which switched on the raw operator string. A valid, evaluator-accepted `operator: or` therefore hit no case in `decompileRuleNodeToExpr` and the route's `WHEN` clause was silently dropped — turning a scoped OR route into one that matches **every** request. The text decompiler hit the same gap and emitted the literal lowercase operator into DSL source, which then failed to recompile. Nothing validated the operator value anywhere: an invalid value written in DSL source (e.g. a typo like `XOR`) survived compilation and was silently reinterpreted as OR by the router's default branch.

### Changes

- Add `config.NormalizeRuleOperator`, a single recursive normalization point: trims whitespace, upper-cases, defaults an omitted operator on a non-empty node to `AND` (matching the CLI/DSL/composer default), and rejects any value outside `AND`/`OR`/`NOT` with an error naming the decision and node path. A completely empty node (the canonical match-all root fallback) is left untouched.
- Wire it in as the *first* routing-profile config validator (`validator.go`), applied to every decision's rule tree and every complexity-signal composer tree, so every other rule-tree-consuming validator observes the canonical form.
- Fix both decompiler entry points (`decompiler_convert.go`, `decompiler_decisions.go`) to compare a normalized operator instead of the raw field — defensive independent of the load-time validator, since `Decompile`/`DecompileToAST` can run against a config that never passed through it (e.g. a Kubernetes CR merged for a dashboard preview).
- Have the DSL compiler validate a complexity signal's composer operator through the same normalizer (`compiler_signals.go`), rejecting an invalid value at compile time instead of deferring the failure to config load.

### Out of scope

Per the issue, this is coordinated with but independent of #2123 (node-shape/leaf validation) — that PR is still open/unmerged, so this change doesn't depend on it. Evaluator default branches (`pkg/decision/engine.go`, `trace.go`, `pkg/classification/classifier_composers.go`) are left as-is: normalization makes them unreachable for any validated config, and changing them was unnecessary extra blast radius.

## Test plan

- [x] `go test ./pkg/config/... ./pkg/dsl/... ./pkg/decision/...` — all pass
- [x] New tests in `pkg/config/validator_rule_operator_test.go`: omitted-operator defaulting, trim/case normalization, empty-root preservation, invalid-operator rejection (root and nested), and `validateRuleOperatorContracts` wiring across decisions + complexity composers
- [x] New tests in `pkg/dsl/decompiler_rule_operator_test.go`: confirmed (by reverting the decompiler fix locally) that `TestDecompileToAST_RecognizesUnnormalizedOperatorCase` and `TestDecompile_RecognizesUnnormalizedOperatorCase` reproduce the dropped-`WHEN`-clause and invalid-DSL-text bugs, then pass after the fix; `TestCompileComplexitySignal_RejectsUnsupportedComposerOperator` covers compile-time rejection
- [x] `gofmt -l` and `go vet` clean on all changed files
- [x] `make precommit-branch-gate AGENT_BASE_REF=origin/main` — clean through `codespell-tracked`, changed-file lint, `recipe-conformance-static`, and `test-semantic-router`; the only failure (`dashboard-check` → two `dashboard/backend` tests requiring a live `vllm-sr-dashboard-container` Docker container) reproduces identically on a clean `origin/main` worktree with none of this PR's changes, confirming it's a pre-existing environment gap


Xunzhuo (approved):
Looks good. Normalization now happens before rule consumers, so omitted and mixed-case operators no longer change meaning across evaluation and DSL round trips. The nested/default/invalid-operator coverage is solid.

wilsonwu (approved):

PR #3057, opened 2026-08-27
[Bug] Make vector-store shutdown generation-safe

### LINKED ISSUE #2602: [Feature] Add a lifecycle root and bounded idempotent Stop(ctx) to vector storage

Nikhi00718:
Closes #2602

## Purpose

Make vector-store ingestion shutdown generation-safe. A timed-out `Stop(ctx)` previously marked the pipeline as stopped even while a worker was still blocked, so a later `Stop` could falsely return success and a restart could reuse lifecycle resources. This change isolates queues, contexts, stop signals, and wait groups per generation, bounds metadata cleanup, preserves cancellation classification, and prevents runtime dependencies from being closed while workers may still be alive.

The implementation preserves the current context-aware embedding API and adapts the lifecycle design documented in closed PR #2517 without copying its outdated implementation.

## Test Plan

- Updated lifecycle tests verify that a second bounded stop still returns `context.DeadlineExceeded` while the original worker is blocked.
- Added restart coverage proving `Start` waits for the old generation and new jobs run on fresh resources.
- Preserved queued-shutdown and context-aware embedder cancellation coverage.
- `git diff --check` passes.
- `go test ./pkg/config` passes.
- Vector-store/runtime Go test compilation reached native linking, but execution requires the repository's built Candle/ML/NLP Rust libraries (`make build-router`); those libraries are not available in this checkout.

## Test Result

- Passed: `go test ./pkg/config`
- Passed: `git diff --check`
- Not run to completion: `go test -race ./pkg/vectorstore ./pkg/routerruntime` (native libraries unavailable; linker reported missing `-lcandle_semantic_router` and related artifacts).

## Checklist

- [x] PR title uses one category prefix: `[Bug]`
- [x] Commit is DCO signed with `git commit -s`
- [x] Changes are limited to vector-store lifecycle, runtime shutdown, configuration defaults, dashboard exposure, and regression tests

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1961,7 +1961,7 @@ global:
       embedding_model: multimodal
       embedding_dimension: 384
       ingestion_workers: 4
-      ingestion_drain_timeout_seconds: 30
+      ingestion_drain_timeout_seconds: 25
       supported_formats: [.txt, .md, .json, .csv, .html, .pdf]
       milvus:
         connection:
--- a/dashboard/frontend/src/pages/configPageRouterDefaultsCatalog.ts
+++ b/dashboard/frontend/src/pages/configPageRouterDefaultsCatalog.ts
@@ -111,6 +111,7 @@ export const DEFAULT_SECTIONS: Record<RouterSystemKey, unknown> = {
     embedding_model: 'mmbert',
     embedding_dimension: 384,
     ingestion_workers: 2,
+    ingestion_drain_timeout_seconds: 25,
     supported_formats: ['.txt', '.md', '.json', '.csv', '.html'],
     memory: {
       max_entries_per_store: 100000,
--- a/dashboard/frontend/src/pages/configPageRouterDefaultsSupport.ts
+++ b/dashboard/frontend/src/pages/configPageRouterDefaultsSupport.ts
@@ -758,6 +758,7 @@ function fieldsForKey(key: RouterSystemKey): FieldConfig[] {
           placeholder: '384',
         },
         { name: 'ingestion_workers', label: 'Ingestion Workers', type: 'number', placeholder: '2' },
+        { name: 'ingestion_drain_timeout_seconds', label: 'Ingestion Drain Timeout (s)', type: 'number', placeholder: '25' },
         routerStructuredField(key, 'supported_formats'),
         routerStructuredField(key, 'memory'),
         routerStructuredField(key, 'milvus'),
--- a/dashboard/frontend/src/pages/configPageSupport.ts
+++ b/dashboard/frontend/src/pages/configPageSupport.ts
@@ -554,6 +554,7 @@ export interface VectorStoreConfig {
   embedding_model?: string
   embedding_dimension?: number
   ingestion_workers?: number
+  ingestion_drain_timeout_seconds?: number
   supported_formats?: string[]
   milvus?: VectorStoreMilvusConfig
   memory?: VectorStoreMemoryConfig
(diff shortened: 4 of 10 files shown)


Xunzhuo (changes requested):
`Stop(ctx)` is still not bounded when jobs are queued. It calls `failJob` for every queued job before checking `ctx.Done()`, and each call gets its own 5s background timeout. With a full queue, shutdown can therefore block for roughly 500s despite the configured 25s deadline. Please bound the whole queued cleanup by the stop context and add coverage with queued jobs plus a stalled registry.

Nikhi00718:
Addressed the requested shutdown-bound fix in aaad0e6f:

- queued cleanup now uses the same Stop(ctx) budget for every metadata update instead of creating a fresh 5s background timeout per job;
- cleanup checks the shared context before each queued job and again before waiting for workers, cancelling the generation and returning ctx.Err() when the budget is exhausted;
- added a regression test with one blocked in-flight job, three queued jobs, and a registry whose SaveStore stalls until the stop deadline.

The repository structure check passes (file-length warnings only), formatting/diff checks pass, and DCO is green. Full native-linked Go tests are running in PR CI.

Xunzhuo (commented):
The shutdown-budget blocker is fixed in `aaad0e6`: queued cleanup now shares the caller's `Stop(ctx)` budget, and the stalled-registry test covers it. This branch is also closer to #2602 than the overlapping #3120 because restart joins the old generation before creating a new one.

Please rebase onto current `main` and resolve the vector-store/config conflicts, then ping me for the final pass.

Nikhi00718:
Follow-up fix pushed in 65d7f555.

The queue-drain path now marks every drained job failed before attempting bounded metadata persistence and batches count transitions per store. This prevents later queued jobs from being discarded while remaining in_progress when a registry write consumes the Stop deadline.

Regression coverage now checks all three queued jobs plus count invariants (one wedged in-flight job remains in progress; three queued jobs are failed; Total=4).

Local validation: gofmt, git diff --check, and isolated production-file compilation pass. The full vectorstore tests are blocked in this Windows environment by CGO/valkey-glide dependency issues and application-control blocking the generated test binary.

Nikhi00718:
Rebased and pushed on top of the latest main in b767212d (including the newly merged #3115).

The three vector-store commits replayed without conflicts and git range-diff confirms the patch is unchanged. The earlier recipe-conformance shard 2 failure was transient runner infrastructure: Docker returned HTTP 500 while pulling postgres:16-alpine; the core build/tests, pre-commit, dashboard, E2E, security, and DCO checks passed. Fresh CI is running on the updated head.

Please take the final pass when convenient.

Nikhi00718:
@Xunzhuo, the branch is now rebased onto the latest main at b767212d and is 0 commits behind. All fresh CI checks are green, including core build/tests, full pre-commit, Envoy AI Gateway E2E, and all recipe-conformance shards. Please take the final pass when convenient.

Xunzhuo (approved):
Thanks for tightening this up. The current head fixes the shutdown-budget blocker: every drained job is transitioned in memory before a single bounded persistence pass, and the stalled-registry test now verifies all queued statuses and counts. The exact head is green across core tests, dashboard, E2E, recipe conformance, and DCO. Approved; only the normal rebase/merge queue remains.

wilsonwu (approved):
LGTM, thanks!

Issue #2992, opened 2026-08-24
[Epic] Optimize data-plane performance, streaming, and failure recovery


Xunzhuo:
## Parent Direction

Data Plane & Networking — #2967

## Goal

Optimize and harden the live request path across standalone and gateway-integrated modes while keeping measurements reproducible, resources bounded, and recovery behavior deterministic.

## Background

Request parsing, routing, dispatch, streaming, cancellation, and telemetry form one user-visible lifecycle. Improvements must preserve protocol correctness and distinguish Router costs from gateway-, provider-, and deployment-specific overhead.

## Scope

- Establish latency, throughput, streaming, resource, and failure baselines for maintained modes.
- Profile request parsing, signal execution, policy evaluation, dispatch, streaming, telemetry, and immediate responses.
- Bound allocations, buffering, queues, goroutines, retries, and telemetry cardinality.
- Define deterministic cancellation, backend-error, partial-response, overload, and fallback behavior.
- Preserve native upstream deltas for final user-visible Looper stages where the final model is known before bytes are emitted.
- Connect material performance changes to Evaluation & Quality benchmark contracts.

### Resource and lifecycle hardening

- #2222 — reproduce and isolate unbounded request-path memory retention.
- #2470 — generation-owned construction, reload, and graceful shutdown.
- #2474 — make vector-store ingestion cancellable, bounded, and transactional.
  - #2602 — add a lifecycle root and bounded, idempotent `Stop(ctx)`.
- #2478 — define bounded typed external RAG requests and responses.

### Streaming and failure recovery

- #2294 — add bounded upstream-error fallback across supported protocols.
- #3378 — stream eligible final Looper stages through native upstream deltas.

Protocol codec correctness and backend dispatch semantics are owned by #2358. Backend replica telemetry and endpoint selection are owned by #2332.

## Non-goals

- Move backend replica scheduling or capacity solvers into semantic routing.
- Own cross-platform or hardware qualification; Evaluation & Quality owns that evidence through #1510.
- Trade correctness, isolation, safety, or observability for benchmark numbers.
- Claim parity across unlike environments or publish private infrastructure details.

## Completion criteria

- [ ] Maintained data-plane modes have reproducible baselines.
- [ ] The highest-impact request-path costs are measured and prioritized.
- [ ] Optimizations include before-and-after evidence.
- [ ] Streaming, cancellation, retry, fallback, overload, and resource lifecycle have regression coverage.
[...]

PR #3059, opened 2026-08-27
[Bug] Treat omitted tool_choice as auto when tools are present

### LINKED ISSUE #3052: [Bug] Treat omitted tool_choice as auto for configured tool selection
Parent Epic: #2358

## Objective

Treat an omitted `tool_choice` as automatic selection when tools are present and semantic tool selection is enabled.

## Current behavior

The current tool-selection path recognizes an explicit `tool_choice: "auto"`, but an omitted `tool_choice` can bypass selection. For OpenAI-compatible requests, omission with a non-empty tool list has automatic semantics, so the router should not silently change behavior.

## Scope

- Normalize omitted `tool_choice` to automatic selection when tools are present.
- Preserve explicit `none`, `required`, and named-tool behavior.
- Define behavior when no tools are present or the selection plugin is disabled.
- Add an observable diagnostic when selection cannot run instead of silently skipping it.

## Acceptance criteria

- Omitted and explicit `auto` produce equivalent selection behavior.
- Explicit tool modes remain unchanged.
- OpenAI-compatible and Anthropic translation paths are covered.
- Unit and integration tests cover omitted, `auto`, `none`, `required`, and named-tool requests.

## Ownership

**Parent:** #2992  
**Related direction:** #2361

FAUST-BENCHOU:
Close #3052
Related #2992

## Purpose

When semantic tool selection is enabled, the router currently runs selection only if `tool_choice` is the literal string `"auto"`. Many OpenAI-compatible clients (and Anthropic clients) omit `tool_choice` entirely when they pass a non-empty `tools` list. Per the OpenAI Chat Completions contract, that omission **is** automatic selection: `auto` is the default when tools are present, and `none` is the default when they are not. Anthropic has the same default when tools are provided.

The result is a silent behavior change: an explicit `tool_choice: "auto"` request is semantically filtered, but the equivalent omitted request bypasses selection and forwards the full tool list upstream. This PR closes that gap without rewriting the wire body.

**Workgroup:** Data Plane & Networking (`wg/data-plane-networking`)
**Affected modules:** `pkg/extproc` tool-selection gate, Anthropic inbound IR (tests only), tool-selection E2E contract

### Root cause

`handleEarlyToolModes` gated on:

```go
openAIRequest.ToolChoice.OfAuto.Value == "auto"
```

When `tool_choice` is omitted, the OpenAI Go SDK leaves `OfAuto` omitted and `Value` is `""`, so the check fails and selection is skipped. Passthrough mode returned `false` with no log.

### Fix (small, reuse existing helpers)

No new helper and no request-body rewrite. The existing `hasToolChoice` check (already used by `clearToolChoiceWhenNoTools`) is combined with the current `"auto"` comparison:

- Explicit `"auto"` → run selection (unchanged)
- Omitted `tool_choice` **and** `len(tools) > 0` → treat as auto and run selection
- Explicit `"none"`, `"required"`, or named-tool → skip selection (unchanged)
- Omitted `tool_choice` **and** no remaining tools → skip selection (OpenAI default is `none`; do not inject tools)

The same rule is applied after allow/block filtering in `filtered` mode, so an omitted request whose tools are all stripped does not accidentally proceed to semantic selection.

Anthropic inbound translation is intentionally left unchanged: omitted Anthropic `tool_choice` stays omitted on the IR. The IR-level gate then treats omitted + tools as auto, so OpenAI-compatible, Anthropic, and Response API (`ToolChoice == nil`) paths share one rule.

### Observability

Previously, passthrough skipped selection silently. It now logs:

- `Skipping semantic tool selection: tool_choice is not automatic` when the gate declines
- `Skipping semantic tool selection: semantic selection disabled` when the plugin is on but `SelectionEnabled()` is false

No new response headers were added; skip diagnostics follow the existing `logging.Infof` pattern used by the tools plugin.

### Behavior matrix

| Request | Semantic selection |
| --- | --- |
| `tool_choice: "auto"` + tools | runs |
| omitted `tool_choice` + tools | runs (this was the bug) |
| omitted `tool_choice`, no tools | skipped (default `none`) |
| `tool_choice: "none"` / `"required"` | skipped |
| named tool (`{"type":"function",...}`) | skipped |
| Anthropic omitted + tools | runs (IR omitted, gate treats as auto) |
| Anthropic `type: auto` / `none` / `any` / `tool` | same as OpenAI auto / none / required / named |
| `filtered` mode, omitted, tools remain after allow/block | runs |
| `filtered` mode, omitted, no tools remain | skipped |

The upstream JSON is not rewritten to insert `"auto"`. Backends that already treat omitted + tools as auto keep that wire shape.

## Test Plan

- [x] Unit: `TestHandleEarlyToolModes_ToolChoiceGate` covers OpenAI omitted / auto / none / required / named, filtered remaining vs stripped tools, and the matching Anthropic translation cases
- [x] Unit: `TestParseAnthropicRequest_OmittedToolChoiceWithTools` asserts inbound IR keeps omitted `tool_choice` (no silent rewrite)
- [x] Integration: Ginkgo cases in `req_filter_tools_test.go` for omitted+tools (selection replaces noise tools), omitted without tools, and explicit none/required
- [x] E2E: `add_mode_omitted_tool_choice_matches_auto` sends tools without `tool_choice` and expects the same decision/strategy as explicit auto; existing explicit-auto case is unchanged
- [ ] `go test ./pkg/extproc/ -run TestHandleEarlyToolModes_ToolChoiceGate` (needs local candle/ml/nlp bindings)
- [ ] `go test ./pkg/anthropic/ -run TestParseAnthropicRequest_OmittedToolChoiceWithTools`
- [ ] CI gates on this PR

## Test Result

- `go test ./pkg/anthropic/ -run 'TestParseAnthropicRequest_OmittedToolChoiceWithTools|TestParseAnthropicRequest_ToolChoice'` → **pass**
- `go test ./e2e/testcases/ -count=0` → **pass** (compile)
- `go test ./pkg/extproc/ -run TestHandleEarlyToolModes_ToolChoiceGate` → **not run locally**; linker failed because `candle-binding` / `nlp-binding` / `ml-binding` release libs are not present on this machine (`ld: library 'candle_semantic_router' not found`). Coverage is in CI.

Remaining risk: E2E `tool-selection` is Kubernetes-profile only and was not executed locally. The new case mirrors the existing weather add-mode contract with `OmitToolChoice: true`.

## Semantic Router PR Checklist

- [ ] Linked issue is accepted with exactly one owner — **#3052 is still `needs-acceptance`**; owner workgroup is Data Plane & Networking
- [x] Purpose, Test Plan, and Test Result match this change

### DIFF
--- a/e2e/testcases/tool_selection_e2e.go
+++ b/e2e/testcases/tool_selection_e2e.go
@@ -31,6 +31,7 @@ type toolSelectionE2ECase struct {
 	ExpectToolsStrategy     string
 	ExpectConfidenceGT      float64
 	ExpectInjectedSysPrompt *bool
+	OmitToolChoice          bool
 }
 
 func defaultContractTools(minObjectParams json.RawMessage) []fixtures.ChatTool {
@@ -48,6 +49,15 @@ func toolSelectionContractCases(minObjectParams json.RawMessage) []toolSelection
 			Name:                "add_mode_weather_query",
 			Prompt:              "__TOOL_SELECTION_ADD_WEATHER__ What is the weather forecast for Boston tomorrow?",
 			Tools:               contractTools,
+			ExpectDecision:      "tool_selection_add_weather_decision",
+			ExpectToolsStrategy: "default",
+			ExpectConfidenceGT:  0.01,
+		},
+		{
+			Name:                "add_mode_omitted_tool_choice_matches_auto",
+			Prompt:              "__TOOL_SELECTION_ADD_WEATHER__ What is the weather forecast for Boston tomorrow?",
+			Tools:               contractTools,
+			OmitToolChoice:      true,
 			ExpectDecision:      "tool_selection_add_weather_decision",
 			ExpectToolsStrategy: "default",
 			ExpectConfidenceGT:  0.01,
@@ -209,7 +219,7 @@ func buildToolSelectionChatRequest(tc toolSelectionE2ECase) fixtures.ChatComplet
 		},
 		Tools: tc.Tools,
 	}
-	if len(tc.Tools) > 0 {
+	if len(tc.Tools) > 0 && !tc.OmitToolChoice {
 		req.ToolChoice = json.RawMessage(`"auto"`)
 	}
 	return req
--- a/src/semantic-router/pkg/anthropic/inbound_test.go
+++ b/src/semantic-router/pkg/anthropic/inbound_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/openai/openai-go"
+	"github.com/openai/openai-go/packages/param"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/ir"
 )
@@ -366,6 +367,20 @@ func TestParseAnthropicRequest_ToolsWithStrict(t *testing.T) {
 	}
 	if ext.ToolStrict["calc"] {
 		t.Fatalf("expected strict=false (or absent) for calc, got %+v", ext.ToolStrict)
+	}
+}
+
+func TestParseAnthropicRequest_OmittedToolChoiceWithTools(t *testing.T) {
+	body := []byte(`{"model":"claude","messages":[{"role":"user","content":"x"}],"tools":[{"name":"search","input_schema":{"type":"object"}}]}`)
+	params, _, err := ParseAnthropicRequest(body)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if len(params.Tools) != 1 {
+		t.Fatalf("expected 1 tool, got %d", len(params.Tools))
+	}
+	if !param.IsOmitted(params.ToolChoice.OfAuto) || params.ToolChoice.OfChatCompletionNamedToolChoice != nil {
+		t.Fatalf("omitted tool_choice should stay omitted in IR, got %+v", params.ToolChoice)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


Xunzhuo (approved):
Looks good. Omitted tool_choice is treated as auto only when tools remain, while explicit none/required/named choices and the no-tools path keep their existing behavior. The OpenAI, Anthropic, filtered-tool, and E2E cases cover the change.

wilsonwu (approved):

drivebyer:
Closing as superseded by #3076.

Since #3076 merged, `applyRequestSemanticDefaults` in `pkg/protocolcodec/engine.go` normalizes an omitted `tool_choice` to `auto` at decode time whenever `tools` is non-empty, for OpenAI Chat, Responses, and Anthropic Messages alike. `handleEarlyToolModes` now gates on `request.ToolChoice.Mode == llmprotocol.ToolChoiceAuto`, so omitted and explicit `auto` already take the same path. `TestOfficialOmittedToolChoiceUsesAutomaticSelectionWhenToolsExist` covers all three formats, and #3052 was closed on that basis.

This branch also no longer applies to main: `hasToolChoice` / `OfAuto` were removed when the tool-selection path moved to `llmprotocol.Request`, and `pkg/anthropic/inbound_test.go` no longer exists, hence the merge conflict.

Thanks for the fix and the thorough behavior matrix. The one piece not carried over is the skip diagnostics (`Skipping semantic tool selection: ...`); if you want to pursue that, a small follow-up PR against the current `handleEarlyToolModes` would be welcome.

PR #3062, opened 2026-08-27
[Bug] Run tool selection before Anthropic backend dispatch

### LINKED ISSUE #3051: [Bug] Run tool selection before Anthropic backend dispatch
Parent Epic: #2358

## Objective

Apply configured semantic tool selection before choosing and translating to an Anthropic backend.

## Current behavior

On current `main`, request processing can enter the Anthropic routing path before normal semantic tool selection runs. The pre-dispatch path handles explicit tool-disable policy, but a request routed to an Anthropic backend can otherwise skip the configured tool-selection stage.

This creates behavior differences based on the selected backend rather than the routing recipe.

## Scope

- Run tool selection once at a protocol-neutral point before backend dispatch.
- Preserve explicit `none`, `required`, and named-tool semantics.
- Keep the translated OpenAI-compatible or Anthropic request valid for the selected backend.
- Avoid duplicate selection or embedding work after provider translation.

## Acceptance criteria

- The same tool-selection policy applies to OpenAI-compatible and Anthropic backends.
- Requests with automatic routing and requests with an explicit model are covered.
- Streaming and non-streaming paths behave consistently.
- Tests verify that selection runs exactly once and that the final provider payload contains the expected tools.

## Ownership

**Parent:** #2992  
**Related direction:** #2361

FAUST-BENCHOU:
Closes #3051
Related #2992

## Purpose

Tool selection is a recipe/plugin policy. It must run on the protocol-neutral IR **before** the router chooses a provider format and translates the body. On current `main`, `handleModelRouting` can enter `handleAnthropicRouting` immediately after the pre-dispatch `mode=none` strip. That path never calls `handleToolSelectionForRequest`.

OpenAI auto/specified routing still ran selection later. The same decision therefore behaved differently depending on whether the selected backend was Anthropic or OpenAI-compatible.

**Workgroup:** Data Plane & Networking (`wg/data-plane-networking`)
**Affected modules:** `pkg/extproc` request-body orchestration, tool-selection filter, and `pkg/anthropic` passthrough remap

### Root cause

```text
applyPreDispatchToolsPolicy (mode=none only)
  -> if Anthropic API format: handleAnthropicRouting() and return
  -> else specified/auto routing, which then call handleToolSelectionForRequest
```

Anthropic translation (`ToAnthropicRequestBodyWithPassthrough`) already reads tools from the OpenAI IR. The missing step was running selection on that IR first.

### Fix

Run selection once in `handleModelRouting` after the direct-loop shortcut and **before** Anthropic / specified / auto dispatch:

1. `applyToolSelectionBeforeDispatch` mutates the IR through the existing `handleToolSelectionForRequest` path (no second retrieval).
2. Persist selected `tools`/`tool_choice` onto `ctx.WorkingRequestBody` **only when selection actually changed them**. Inactive/no-plugin paths leave the original body untouched so provider-specific fields are not dropped.
3. For Anthropic-shaped inbound, snapshot the native wire body, run IR selection, then restore the snapshot so passthrough capture (`cache_control`, tool_result, images) is not corrupted by OpenAI-shaped `tools` JSON.
4. After Anthropic translation, remap `tools[].cache_control` by **tool name** (`RemapToolsCacheControl`). Dropped tools lose their markers; retained tools keep them at the new index.
5. Carry `x-vsr-tools-*` from the temporary ExtProc stub onto the final Anthropic / specified / auto dispatch response (`ctx.ToolObservability`).
6. Remove the duplicate `handleToolSelectionForRequest` calls from `handleAutoModelRouting` and `handleSpecifiedModelRouting`.

`handleAnthropicRouting` then translates the already-selected IR. Semantic retrieval is not repeated after provider translation.

### What this does not change

- Direct loop models (Fusion / ReMoM / Flow) still return before this call, matching previous skip behavior for those aliases.
- Explicit `none`, `required`, and named-tool requests still skip semantic retrieval in passthrough mode.
- `mode=none` still runs first via `applyPreDispatchToolsPolicy`.
- Omitted `tool_choice` vs `"auto"` remains #3052; this PR uses explicit `"auto"` in tests.

### Behavior

| Path | Before | After |
| --- | --- | --- |
| Auto route to Anthropic backend | skipped selection | selection on IR, then Anthropic translate |
| Explicit Anthropic model | skipped selection | same |
| Streaming Anthropic | skipped selection | same, `stream=true` preserved |
| OpenAI specified/auto | selection after building the OpenAI body | selection once before dispatch; working body carries the result only if tools changed |
| `none` / `required` / named (passthrough) | n/a on Anthropic | request tools forwarded unchanged |
| Native Anthropic `cache_control` after filter | replayed by old tool index | remapped by tool name |
| `x-vsr-tools-*` on Anthropic / model rewrite | left on the temporary stub | copied onto the final dispatch response |

## Test Plan

- [x] `TestHandleModelRouting_AnthropicAppliesToolSelectionOnce` — explicit Anthropic model, filtered allow-list, outbound Anthropic `tools` contains only `lookup_weather`
- [x] `TestHandleModelRouting_AnthropicAutoRoutingAppliesToolSelection` — `MoM` auto-route to Anthropic backend
- [x] `TestHandleModelRouting_AnthropicStreamingAppliesToolSelection` — streaming flag retained with selected tools
- [x] `TestHandleModelRouting_AnthropicExplicitToolModesKeepRequestTools` — `none`, `required`, named keep both request tools
- [x] `TestHandleModelRouting_OpenAISpecifiedKeepsSelectedTools` — OpenAI specified-model path still applies the filtered list
- [x] `TestHandleModelRouting_OpenAISpecifiedDoesNotRewriteToolsWhenSelectionInactive` — no tools plugin leaves original `tools`/`tool_choice` bytes unchanged
- [x] `TestHandleModelRouting_AnthropicRemapsToolCacheControl` / `AnthropicKeepsRetainedToolCacheControl` — filtered tools drop markers; retained tools keep them by name
- [x] `TestHandleModelRouting_AnthropicCarriesToolObservabilityHeaders` — final Anthropic response includes `x-vsr-tools-*`
- [x] `TestRemapToolsCacheControl_*` in `pkg/anthropic`
- [x] E2E `tool-selection-anthropic` (`POST /v1/messages`): requires `x-vsr-selected-decision=tool_selection_filter_decision`, `x-vsr-tools-strategy=filter`, and non-empty `x-vsr-tools-confidence` / `x-vsr-tools-latency-ms`
- [ ] `go test ./pkg/extproc/ -run 'TestHandleModelRouting_Anthropic|TestHandleModelRouting_OpenAISpecified'`
- [ ] `go test ./pkg/anthropic/ -run TestRemapToolsCacheControl`
- [ ] CI Core quality / unit tests and Kubernetes E2E `tool-selection-anthropic` on this PR

## Test Result

`go test ./pkg/anthropic/ -run TestRemapToolsCacheControl` and `go test ./e2e/testcases/` compile and pass locally.

Local `go test ./pkg/extproc/...` could not be executed here: the package links candle/ml/nlp bindings, and this environment has neither those release libraries nor a complete module cache. Coverage is the dispatch / observability tests above plus CI.

Remaining risk: Kubernetes E2E can observe ExtProc response headers, not the outbound Anthropic `tools` body sent to the upstream. Filtered tool lists and `cache_control` remap stay covered by unit tests. The E2E profile still uses the envoy-ai-gateway OpenAI-shaped backend after inbound `/v1/messages`; it does not stand up a native Anthropic provider.

## Semantic Router PR Checklist

- [x] Linked issue is accepted with exactly one owner (`wg/data-plane-networking` on #3051)
- [x] Purpose, Test Plan, and Test Result match this change

### DIFF
--- a/e2e/testcases/tool_selection_anthropic_e2e.go
+++ b/e2e/testcases/tool_selection_anthropic_e2e.go
@@ -1,0 +1,112 @@
+package testcases
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+)
+
+func init() {
+	pkgtestcases.Register("tool-selection-anthropic", pkgtestcases.TestCase{
+		Description: "Native Anthropic /v1/messages tool selection applies the same filter decision as OpenAI chat completions",
+		Tags:        []string{"kubernetes", "plugin", "tool-selection", "anthropic"},
+		Fn:          testToolSelectionAnthropicE2E,
+	})
+}
+
+type anthropicE2ETool struct {
+	Name        string          `json:"name"`
+	Description string          `json:"description,omitempty"`
+	InputSchema json.RawMessage `json:"input_schema"`
+}
+
+func testToolSelectionAnthropicE2E(ctx context.Context, client *kubernetes.Clientset, opts pkgtestcases.TestCaseOptions) error {
+	if opts.Verbose {
+		fmt.Println("[Test] tool_selection native Anthropic /v1/messages filter contract")
+	}
+
+	localPort, stop, err := setupServiceConnection(ctx, client, opts)
+	if err != nil {
+		return err
+	}
+	defer stop()
+
+	minObjectParams := json.RawMessage(`{"type":"object","properties":{}}`)
+	payload := anthropicMessagesRequestBody{
+		Model:     "MoM",
+		MaxTokens: 64,
+		Messages: []anthropicMessage{
+			{Role: "user", Content: "__TOOL_SELECTION_FILTER__ Will it rain in Seattle this weekend?"},
+		},
+		Tools: []anthropicE2ETool{
+			{Name: "get_weather", Description: "Get current weather information for a location", InputSchema: minObjectParams},
+			{Name: "contract_noise_alpha", Description: "Unrelated tool for cataloguing antique spoons", InputSchema: minObjectParams},
+			{Name: "contract_noise_beta", Description: "Metadata about underground subway tile patterns", InputSchema: minObjectParams},
+		},
+		ToolChoice: json.RawMessage(`{"type":"auto"}`),
+	}
+
+	resp, err := sendAnthropicToolSelectionRequest(ctx, localPort, payload)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(resp.Body)
+
+	decision := resp.Header.Get("x-vsr-selected-decision")
+	strategy := resp.Header.Get("x-vsr-tools-strategy")
+	if opts.SetDetails != nil {
+		opts.SetDetails(map[string]interface{}{
+			"status_code":    resp.StatusCode,
+			"decision":       decision,
+			"tools_strategy": strategy,
+		})
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("tool-selection-anthropic: expected status 200, got %d: %s",
+			resp.StatusCode, truncateString(string(body), 200))
+	}
+	if decision != "tool_selection_filter_decision" {
+		return fmt.Errorf("tool-selection-anthropic: decision want %q got %q",
+			"tool_selection_filter_decision", decision)
+	}
+	if strategy != "" && strategy != "filter" {
+		return fmt.Errorf("tool-selection-anthropic: x-vsr-tools-strategy want %q got %q", "filter", strategy)
+	}
+	return nil
+}
+
+func sendAnthropicToolSelectionRequest(
+	ctx context.Context,
+	localPort string,
+	body anthropicMessagesRequestBody,
+) (*http.Response, error) {
+	jsonData, err := json.Marshal(body)
+	if err != nil {
+		return nil, fmt.Errorf("marshal: %w", err)
+	}
+	url := fmt.Sprintf("http://localhost:%s/v1/messages", localPort)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(jsonData))
+	if err != nil {
+		return nil, fmt.Errorf("new request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("anthropic-version", "2023-06-01")
+	req.Header.Set("x-vsr-debug", "true")
+
+	httpClient := &http.Client{Timeout: 45 * time.Second}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("do: %w", err)
+	}
+	return resp, nil
+}
--- a/e2e/testcases/tool_selection_anthropic_e2e.go
+++ b/e2e/testcases/tool_selection_anthropic_e2e.go
@@ -63,11 +63,15 @@ func testToolSelectionAnthropicE2E(ctx context.Context, client *kubernetes.Clien
 
 	decision := resp.Header.Get("x-vsr-selected-decision")
 	strategy := resp.Header.Get("x-vsr-tools-strategy")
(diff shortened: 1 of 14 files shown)


Xunzhuo (changes requested):
Not ready yet. `applyToolSelectionBeforeDispatch` still rewrites `tools`/`tool_choice` when no tools plugin applies, which can drop provider-specific fields. Native Anthropic `cache_control` is also replayed by the old tool index after filtering, and the `x-vsr-tools-*` headers are left on the temporary response instead of the final dispatch response. Please add native Anthropic E2E coverage, rebase, and ping me again.

FAUST-BENCHOU:
> Not ready yet. `applyToolSelectionBeforeDispatch` still rewrites `tools`/`tool_choice` when no tools plugin applies, which can drop provider-specific fields. Native Anthropic `cache_control` is also replayed by the old tool index after filtering, and the `x-vsr-tools-*` headers are left on the temporary response instead of the final dispatch response. Please add native Anthropic E2E coverage, rebase, and ping me again.

Done :) thanks for review.ping @Xunzhuo

src/semantic-router/pkg/extproc/req_filter_tools_dispatch.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+type toolSelectionSnapshot struct {
+	names      []string
+	toolChoice []byte
+}
+
+func snapshotToolSelection(openAIRequest *openai.ChatCompletionNewParams) toolSelectionSnapshot {
+	return toolSelectionSnapshot{
+		names:      openaiToolNames(openAIRequest),
```
Xunzhuo, line 83: Comparing only names misses a real mutation. In add mode the catalog can return the same tool names and order with updated descriptions or schemas; this returns false and leaves the old OpenAI wire definitions in the final payload. Please track whether selection actually mutated the tools, or compare the full canonical definitions, and cover the same-name/different-schema case.

src/semantic-router/pkg/extproc/processor_req_body_tool_selection_dispatch_test.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+func TestHandleModelRouting_AnthropicAppliesToolSelectionOnce(t *testing.T) {
+	router := anthropicDispatchRouter(t)
+	req, err := parseOpenAIRequest([]byte(twoToolChatBody))
+	require.NoError(t, err)
+	ctx := &RequestContext{
+		Headers:             map[string]string{},
+		OriginalRequestBody: []byte(twoToolChatBody),
+		VSRSelectedDecision: filteredToolsDecision(t, []string{"lookup_weather"}),
```
Xunzhuo, line 83: This context cannot come from the real explicit-model path. Concrete backend models resolve as passthrough, so VSRSelectedDecision stays nil and selection is a no-op. Please either implement the explicit-model acceptance path end to end, or align the issue scope and replace this synthetic test.

e2e/testcases/tool_selection_anthropic_e2e.go
```diff
@@ -0,0 +1,116 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("tool-selection-anthropic: expected status 200, got %d: %s",
+			resp.StatusCode, truncateString(string(body), 200))
+	}
+	if decision != "tool_selection_filter_decision" {
+		return fmt.Errorf("tool-selection-anthropic: decision want %q got %q",
+			"tool_selection_filter_decision", decision)
+	}
+	if strategy != "" && strategy != "filter" {
```
Xunzhuo, line 86: This E2E still routes MoM to the OpenAI-shaped base-model, never exercises Anthropic backend translation, and passes when the tool headers are absent. It would not catch the original regression. Please route it to a fake Anthropic backend and assert the upstream provider payload contains the filtered tools; any headers claimed by the test should also be required.

Xunzhuo (changes requested):
Still not ready. The previous implementation issues are fixed and CI is green, but the explicit-model path and same-name tool mutations are still incorrect. The new E2E also does not hit an Anthropic backend, so it cannot catch the original regression. Please address the inline comments and ping me again.

drivebyer:
@FAUST-BENCHOU Please resolve the conflict.

FAUST-BENCHOU:
no need in this one i think

PR #3040, opened 2026-08-26
[Docs] landing page section lockup, install card, and video cards

abhinav-m22:
Closes #3002

## Purpose

Phase 6 of the website revamp, following #3039. This is the final phase and brings the landing page onto the same frame and section rhythm as Docs, Blog, Research, and Community.

- **Shared section intro.** All landing sections now use the same two-column layout, with the title on the left and lede on the right.
- **Consistent alignment.** Section intros now follow the shared frame instead of using different widths and margins.
- **Install card.** The explanation and install command now have separate columns and stack on smaller screens.
- **Video cards.** Thumbnails now scale with the card width and use a consistent 16:9 layout.
- **Cleanup.** Removes the unused `HomepageFeatures` component and its old width and centering rules.

One file sits outside this phase's manifest:

- `section-surfaces.css` contains the shared `.site-section-intro` primitive used by multiple landing page sections.

## Test Plan

```bash id="q4l9cv"
cd website && npm run lint
npm start
```

Checked the landing page at 1920 / 1280 / 996 / 640.

- All section intros share the same left edge and two-column layout.
- No section leaves unnecessary empty space on the right at 1920px.
- Install card uses separate columns for the explanation and command, then stacks below 996px.
- Video cards use full-width 16:9 thumbnails that scale with the column.
- Landing page frame edges match Docs, Blog, Research, and Community.
- At 996px and below, section intros collapse to one column with no horizontal scroll.
- Keyboard focus remains visible on video cards, the copy button, and CTA pills.

## Test Result

`npm run lint` clean.

Please check the deployed preview for the full visual details.

### Before

Before 1

Before 2

Before 3

### After

After 1

After 2

After 3

After 4

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title begins with exactly one bracketed category, such as `[Feature]`, `[Bug]`, `[Docs]`, `[Test]`, `[Research]`, `[Community]`, or `[CI/Build]`
- [x] The title does not stack prefixes such as `[Router][Docs]`; affected modules belong in labels and the PR body
- [x] The PR links an `accepted` issue with exactly one owner: one `wg/*` label for project work or `owner/maintainers` for repository governance
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>

See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full contributor workflow and commit guidance.

Issue #3002, opened 2026-08-24
[Feature] Website UI/UX consistency across Docs, Blog, Research, and Community


abhinav-m22:
### Goal

As discussed with @AayushSaini101, opening this to get a consistent UI/UX across the website.

Docs, Blog, Research, and Community look like four separately designed pages rather than one website. Each page type sets its own width and its own sidebar, so the left and right gaps are uneven and nothing lines up when moving between sections.

Measured on `main`:

| Page         | Container                                        | Rail                                     |
| ------------ | ------------------------------------------------ | ---------------------------------------- |
| Docs         | `76.25rem`, grows to `83.875rem` on wide screens | fixed `15.125rem`                        |
| Research     | own value                                        | **proportional** `minmax(13rem, 0.22fr)` |
| Community    | own value                                        | **proportional** `minmax(13rem, 0.22fr)` |
| Blog listing | own value                                        | `17rem`, **on the right**                |
| Blog article | `62rem`, no line-length cap                      | none                                     |

A fixed rail and a proportional rail can never line up at any viewport width. That is the root cause of the uneven gaps. It is not a spacing bug that can be nudged.

Three related problems on the same surface:

* Navbar links are `display:none` below `1339px`, so 1280px laptops lose the desktop nav.
* `--site-muted` (`#677487`) is 4.7:1 on white and serves both navbar links and secondary body text. One token is doing two jobs, so navigation reads washed out in light mode.
* `--site-bg` and `--site-surface` are both `#ffffff`, so cards have no surface of their own. Dark mode's two values differ, so this only shows in light mode.

### Proposed Workgroup

Developer Experience & Ecosystem

### Scope

One shared frame for every page: width, gutters, rail width, spacing scale, type scale, breakpoints. Two arrangements inside it:

* **Browse** (Research, Community, blog listing): title above a left section rail.
* **Read** (Docs, blog article): title beside the rail, running text capped at a readable measure.

Docs keeps its own internals. It only shares the outer frame. This is deliberate. Docs is navigated, the other three are browsed, and forcing them identical would make the browse pages feel like documentation.

## Delivery: 6 PRs

Split by surface. Every PR has a visible on-screen change so it can be approved from before/after screenshots. No file appears in more than one PR.

### PR 1: [Website] Design foundation: typography, palette, shared primitives

**Visible result:** body text changes from Arial to Inter site-wide; navbar and sidebar text becomes legible in light mode; code blocks, admonitions, tables and buttons restyled.

[...]

PR #3069, opened 2026-08-27
[Docs] landing page: closing CTA and blog rail hierarchy

abhinav-m22:
Related #3002

## Purpose

Follow-up to #3040, addressing review feedback.

- **New data sovereignty section.** Shows the mechanism as a four stage pipeline (Request, Hard constraints, Eligible pool, Rank) reusing the stage language from the routing pipeline above it, with the constraint gate highlighted. Wording is grounded in `docs/overview/goals.md` and `docs/overview/use-cases.md`.
- **Final CTA redesigned** as a centred card on the same 46rem box as the install command, so both call to action moments on the page are one component.
- **Blog rail hierarchy.** "Recent" goes 0.875rem to 1.125rem, post titles 0.98rem to 0.92rem, and the rail widens from 18rem to 23rem via a blog scoped `--blog-rail` so post titles stop wrapping to four lines. The shared `--rail` token is untouched, so docs, research and community are unaffected.

Deliberately not claimed: that data never leaves your infrastructure. `docs/overview/use-cases.md` states that "local" is not an end to end privacy guarantee, so the homepage should not say otherwise.

## Test Plan

Checked at 1920 / 1280 / 996 / 640.

- Sovereignty pipeline collapses to two columns below 640px with the arrows removed.
- CTA card stays centred and its buttons stay centred at every width.
- "How routing policy works" resolves to the Routing Pipeline page.
- Blog rail is wider, heading and titles have a clear step.
- Docs, research and community rails are unchanged.

## Test Result

`npm run lint` clean.

Screenshot From 2026-08-27 16-55-12

### DIFF
--- a/website/src/css/blog.css
+++ b/website/src/css/blog.css
@@ -18,6 +18,9 @@ html[data-route-kind='blog'],
   /* One width for prose, code, images and tables. In rem, not ch: ch resolves
    * against the consumer's font-size, so prose and code would differ. */
   --blog-prose: 46rem;
+  /* Wider than the shared --rail: the listing rail carries post titles rather
+   * than short nav labels, so 18rem forced most of them onto three lines. */
+  --blog-rail: 23rem;
 }
 
 html[data-route-kind='blog'] body,
@@ -51,7 +54,7 @@ html[data-route-kind='blog'] body,
 .site-blog-index__columns {
   position: relative;
   display: grid;
-  grid-template-columns: var(--rail) minmax(0, 1fr);
+  grid-template-columns: var(--blog-rail) minmax(0, 1fr);
   gap: var(--rail-gap);
   align-items: start;
   padding-top: var(--space-6);
@@ -62,7 +65,7 @@ html[data-route-kind='blog'] body,
   position: absolute;
   top: var(--space-6);
   bottom: 0;
-  left: var(--rail);
+  left: var(--blog-rail);
   width: 1px;
   background: var(--site-border);
 }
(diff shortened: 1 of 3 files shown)


AayushSaini101:
Thanks for working on this @abhinav-m22 
* Let's follow proper font Style across the website example: "One model API " to One Model API and similar to other areas 
image
* Line break or make smaller text to show the statements: 
image

* "." is present some heading but in the other there is no ".", let's make consistent across the website 
image

* Regarding the data sovereignty can you add an animation/photo that we have is similar to other areas thanks

abhinav-m22:
Thanks @AayushSaini101

On the line break point, could you confirm which you'd prefer before I change it everywhere? We moved to single line text on purpose: with the text wrapped, every section left a large blank area to its right, which was one of the main things this revamp set out to fix. Here is how it looks with the breaks back in:

image

If the full width lines read as too long, a middle option is to cap the text at a readable width rather than the full frame, which shortens the lines without leaving the section half empty. Happy with whichever you prefer, just want to settle it once so it stays consistent.

I'll take care of the rest:

- Font style: "One model API" to "One Model API", and checked the other product terms across the site.
- Headings: adding the missing "." so all section headings are consistent.
- Data sovereignty: adding an animation in the same style as the routing pipeline above it.

abhinav-m22:
Thanks @AayushSaini101

Could you confirm the line break point before I change it everywhere? We went single line on purpose. 

Wrapped text left a large blank area to the right of every section, which is one of the things this revamp set out to fix. Here is how it looks with breaks back in:

image

I'll take care of the rest

ramkrishs (approved):
Lgtm, thanks for closing on this.

wilsonwu (approved):

AayushSaini101:
> Thanks @AayushSaini101
> 
> Could you confirm the line break point before I change it everywhere? We went single line on purpose.
> 
> Wrapped text left a large blank area to the right of every section, which is one of the things this revamp set out to fix. Here is how it looks with breaks back in:
> 
> image
> I'll take care of the rest

you can update the lines for better visiblity thanks @abhinav-m22

AayushSaini101:
image @abhinav-m22 we need to update the format here also it aligned very much to the left side thanks @abhinav-m22

PR #3070, opened 2026-08-27
[Bug] cache: propagate StoredAt and Age through LookupResult

### LINKED ISSUE #2935: [Router] max-age request control is inert on shared cache backends: every L2 exact hit becomes a stale miss
### Describe the bug

The `max-age` request-control directive cannot be satisfied by any shared cache backend. Only the in-process L1 tier can serve it, so a request carrying `max-age` misses on L2 regardless of how old the entry actually is.

The legacy backend contract has no field for entry age:

https://github.com/vllm-project/semantic-router/blob/92d9a04dd7c79d1ad357e15f6beff0f1c96b724a/src/semantic-router/pkg/cache/cache_interface.go#L24-L31

```go
type LookupResult struct {
	ResponseBody []byte
	Found        bool
	Similarity   float32
}
```

So `LegacyBackendAdapter.LookupExact` has nothing to populate `Age` / `AgeKnown` from, and leaves both at their zero values (`legacy_adapter.go:24-46`):

```go
return CacheResult{
	ResponseBody: result.ResponseBody,
	Found:        result.Found,
	HitKind:      HitKindExact,
	Source:       CacheSourceL2,
	Similarity:   result.Similarity,
	// Age, AgeKnown, ExpiresAt are never set
}, nil
```

`LookupExact` then treats "age unknown" as stale (`response_cache_service.go:138-141`):

```go
if result.Found && lookup.MaxAge != nil &&
	(!result.AgeKnown || result.Age > *lookup.MaxAge) {
	s.staleMisses.Add(1)
	result = CacheResult{HitKind: HitKindMiss}
}
```

The `!result.AgeKnown` guard is the right call on its own — refusing to claim freshness you cannot prove is the safe direction. The problem is that its input can never be anything else for redis, valkey, milvus, qdrant or hybrid, because the type in between cannot carry age.

`exactL1` records `storedAt`, so the L1 path is unaffected (`exact_l1.go:36-62`).

### To Reproduce

`pkg/cache` only, no external service required, and no embedding inference:
`InMemoryCache.FindExact` is documented as returning "a response without running
embedding inference", so the shipped in-memory backend reproduces this directly.
Any backend reached through `LegacyBackendAdapter` behaves the same way, the
in-memory one included when it is the configured store.

`L1MaxEntries` must be negative to disable that tier: `NewResponseCacheService`
replaces a zero with the 1024 default, and `newExactL1` clamps a negative to zero.

In a file carrying the package's usual `//go:build !windows && cgo` constraint:

```go
func TestMaxAgeAgainstL2(t *testing.T) {
	ctx := context.Background()

	backend := NewInMemoryCache(InMemoryCacheOptions{
		Enabled:    true,
		MaxEntries: 100,
		TTLSeconds: 3600,
	})
	service := NewResponseCacheService(
		NewLegacyBackendAdapter(backend, InMemoryCacheType),
		ResponseCacheServiceOptions{L1MaxEntries: -1},
	)

	identity := CacheIdentity{
		Partition: CachePartition{
			RequestModel: "model-a",
			Protocol:     "openai:body",
		},
		ExactFingerprint: "fingerprint-1",
	}
	if err := service.StoreExact(ctx, CacheWrite{
		Identity:     identity,
		RequestID:    "req-1",
		ResponseBody: []byte(`{"id":"cached"}`),
		TTL:          TTL(time.Hour),
	}); err != nil {
		t.Fatalf("StoreExact: %v", err)
	}

	// Control: no MaxAge.
	control, err := service.LookupExact(ctx, ExactLookup{Identity: identity})
	if err != nil {
		t.Fatalf("control LookupExact: %v", err)
	}

	// Subject: same lookup, generous MaxAge, entry is milliseconds old.
	maxAge := time.Hour
	subject, err := service.LookupExact(ctx, ExactLookup{
		Identity: identity,
		MaxAge:   &maxAge,
	})
	if err != nil {
		t.Fatalf("subject LookupExact: %v", err)
	}

	stats, err := service.Stats(ctx)
	if err != nil {
		t.Fatalf("Stats: %v", err)
	}

	t.Logf("control found=%v source=%v", control.Found, control.Source)
	t.Logf("subject found=%v hitKind=%v ageKnown=%v",
		subject.Found, subject.HitKind, subject.AgeKnown)
	t.Logf("stale_miss_count=%d", stats.StaleMissCount)

	if subject.Found {
		t.Errorf("max-age lookup hit L2; this issue would be refuted")
	}
}
```

Observed:

```
control found=true source=l2
subject found=false hitKind=miss ageKnown=false
stale_miss_count=1
```

With L1 left at the production default (1024 entries / 5 min), the behaviour splits by tier:

| scenario | result |
|---|---|
| same replica, entry still in L1 | HIT (`source=l1`, `age=13µs`) |
| same replica, entry evicted from L1 by LRU | MISS (control without `max-age`: HIT from l2) |
| second replica, cold L1 | MISS (control without `max-age`: HIT from l2) |

So the practical failure modes are multi-replica deployments, entries older than the 5 minute L1 ceiling, workloads that push past 1024 L1 entries, and router restarts. A single-replica dev setup with a warm L1 looks correct, which is what makes this easy to miss.

### Expected behavior

A `max-age` lookup served from L2 should hit when the entry is younger than the requested bound.

Two ways to get there:

1. Let the legacy contract express age — add `StoredAt` (or `Age`) to `LookupResult` and populate it in each backend. The cost is **not uniform**: the exact path does not share storage with the semantic path, so what each backend already persists differs.

   - **Milvus, Qdrant — plumbing only.** Both exact writers already persist a unix `timestamp` that `FindExact` never reads back: `AddExact` in `exact_cache_milvus.go` writes an `entity.NewColumnInt64("timestamp", …)` column, and `AddExact` in `exact_cache_qdrant.go` writes a `"timestamp": time.Now().Unix()` payload key.
   - **In-memory — one struct field.** `exactMemoryEntry` in `exact_cache_inmemory.go` keeps only `expiresAt`. Age is not recoverable from an expiry without the TTL used at write time, so this needs a `storedAt` — cheap, but still new state rather than plumbing.
   - **Redis, Valkey — new persisted state.** In `exact_cache_redis.go` and `exact_cache_valkey.go`, `AddExact` is a bare `SET <key> <body> EX <ttl>` and `FindExact` a bare `GET`; no timestamp is written on the exact key, and Redis exposes no key-creation time. Deriving age from `PTTL` would need the effective TTL used at write time, which is per-entry via `effectiveExactTTL(ttlSeconds, c.ttlSeconds)`, so it is not reliable. Either store the entry as a hash (`response_body` + `timestamp`, one `HGETALL` on read) or write a sibling key — and either way handle the upgrade, since entries written by the current release are plain strings under the same `exactCacheStorageKey` prefix (`exact_cache.go`), so a post-upgrade `HGETALL` on a surviving key returns `WRONGTYPE`.

   A partial landing is reasonable — the three cheap backends first — as long as the remaining two are called out rather than left silently unfixed.

2. If age is deliberately out of scope for the legacy seam, stop treating "unknown" as stale and document `max-age` as best-effort, honoured only where the backend can report age.

Option 1 keeps the directive meaning what the tutorial says it means. Option 2 is a one-liner but changes the documented contract, so it needs a docs note in `website/docs/tutorials/plugin/response-cache.md`.

**Correction (2026-08-28).** An earlier revision of this issue claimed that every backend already stored the timestamp it needed, citing `redis_cache.go:568`/`:349`, `valkey_cache.go:461`/`:269`, `qdrant_cache.go:243` and `milvus_cache.go:641`/`:352`/`:360`. Those are all semantic-path writers (`addEntry`, `createIndex`, `AddPendingRequest`, `AddEntriesBatch`, `createCollection`) and do not describe the storage the exact path uses — the exact path lives in `exact_cache_*.go` under a separate keyspace. The per-backend breakdown above replaces that claim. The diagnosis, the reproduction and both options are unchanged.

### Affected layer

plugin

### Additional context

**Reachability.** `ctx.CacheMaxAgeSeconds` has a single writer (`req_filter_cache_control.go:61`), gated on `request_controls.enabled` plus `max-age` appearing in `allowed`. The built-in default `allowed` is `[no-cache, no-store, bypass]` (`req_filter_cache_control.go:69-71`), so a minimal config never reaches this. Two shipped examples do enable it, though:

- `config/config.yaml:657`
- `config/fragments/plugin/response-cache/high-recall.yaml:13`

and the tutorial documents the directive as bounding read freshness (`website/docs/tutorials/plugin/response-cache.md:69`), so anyone following the examples turns it on and gets a silent no-op.

**Severity.** This is not a correctness bug in the served data — the failure direction is safe, a miss rather than a stale hit. The cost is hit rate: traffic carrying `max-age` always falls through to the provider, paying for inference while the cache appears to be in use. There is no signal for it either, beyond `stale_miss_count` in `CacheStats`, which is not exported as a metric.

**Is this a known consequence of the migration seam?** `LegacyBackendAdapter` is documented as confining the old API to one boundary until backends move to `TypedCacheStore`, so the age gap may be an accepted intermediate state rather than an oversight. Is age intended to arrive when the backends do? If so this issue is really "track the age half of that migration", and the interesting question is only whether `max-age` should be documented as L1-only until then.

**Relationship to #2919.** #2919 (`Closes #2473`) already changes `FindExact` / `AddExact` / `LookupSimilarWithThreshold` on all six backends to take a `context.Context`, and fixes the discarded `_ context.Context` in this same adapter. It does not touch `LookupResult`'s fields, so it does not address this. It does do the signature churn that option 1 above would otherwise have to do on its own, so the cheapest sequencing is to land #2919 first and add age as a follow-up rather than widening that PR's scope.

Related: #2375 (router quality and API boundary audit), #2473.

ZireaelK:
Propagate StoredAt, ExpiresAt, Age, and AgeKnown through LookupResult and LegacyBackendAdapter across InMemory, Milvus, Qdrant, Redis, and Valkey cache backends.

Previously, LookupResult omitted timestamp information, causing LegacyBackendAdapter to produce CacheResult with AgeKnown=false. In ResponseCacheService.LookupExact, queries with MaxAge evaluated !result.AgeKnown as true, treating every L2 exact hit as a stale miss.

This fix:
- Adds StoredAt, ExpiresAt, Age, and AgeKnown to LookupResult.
- Populates timing metadata on hits across all cache backends.
- Updates LegacyBackendAdapter to compute and propagate Age / AgeKnown / ExpiresAt.
- Adds regression unit tests verifying that L2 exact hits with max-age are honored.

Fixes #2935

Closes #2935

## Purpose

Propagate `StoredAt`, `ExpiresAt`, `Age`, and `AgeKnown` through `LookupResult` and `LegacyBackendAdapter` across `InMemory`, `Milvus`, `Qdrant`, `Redis`, and `Valkey` cache backends.

### Background & Problem
Previously, `LookupResult` omitted entry timestamp metadata, causing `LegacyBackendAdapter` to construct `CacheResult` with `AgeKnown=false` (the default boolean zero-value). In `ResponseCacheService.LookupExact`, queries with `MaxAge` evaluated `!result.AgeKnown` as `true`, treating every L2 exact hit as an immediate stale miss even when freshly stored.

### Solution & Changes
- **Data Model**: Added `StoredAt`, `ExpiresAt`, `Age`, and `AgeKnown` fields to `LookupResult` in `src/semantic-router/pkg/cache/cache_interface.go`, along with the `lookupResultFromTimestamps` helper.
- **Adapter Layer**: Updated `LegacyBackendAdapter` (`LookupExact` and `LookupSemantic`) in `legacy_adapter.go` to compute and inject `Age`, `AgeKnown`, and `ExpiresAt` into `CacheResult`.
- **Backend Implementations**:
  - `InMemory`: Updated `exactMemoryEntry` and `FindExact` / `LookupSimilarWithThreshold` to track entry insertion time and calculate `Age`.
  - `Milvus`: Added `timestamp` and `expires_at` columns to query and vector search outputs; parsed and propagated timing in `FindExact` and `LookupSimilarWithThreshold`.
  - `Qdrant`: Extracted `timestamp` and `expires_at` from payload points in `FindExact` and semantic lookup.
  - `Redis` & `Valkey`: Requested `timestamp` and `ttl_seconds` in `FT.SEARCH` `RETURN` parameters; parsed and populated timing metadata. Extracted Redis search routines into `redis_cache_search.go` to satisfy file-length constraints (< 800 lines).
- **Workgroup / Ownership**: `wg/dataplane` (Data Plane & Networking Working Group) / `router-core`.

## Test Plan

```bash
make agent-validate
make agent-lint
CI=true make agent-ci-gate
make check-go-mod-tidy
```

## Test Result

- `make agent-validate`: **PASS**.
- `make agent-lint`: **PASS**.
- `CI=true make test-semantic-router`: **PASS**.
- `CI=true make agent-ci-gate`: **PASS**.
- `make check-go-mod-tidy`: **PASS**.

---
<details>

</details>


Xunzhuo (changes requested):
Not ready to merge yet.

- Redis and Valkey exact paths are still unchanged. AddExact stores only the body and FindExact returns AgeKnown=false, so max-age still misses every L2 exact hit on those backends. Please carry timing metadata through exact storage and add Redis/Valkey exact regression coverage.
- redis_cache_search.go now hides methods behind !windows && cgo, but redis_cache.go still calls them. Remove the new build constraint or add equivalent excluded-build implementations.

#2935 also still needs /accept, and the PR title should use [Bug], not [Router].

Xunzhuo (changes requested):
The two blockers from the previous review are still present on the current head:

- Redis and Valkey exact entries still persist only the response body, and `FindExact` still returns no timing metadata. That leaves `AgeKnown=false`, so `max-age` cannot validate L2 exact hits. Please persist/read the timing metadata and add Redis + Valkey exact max-age regressions; the current integration tests only cover round-trip/isolation.
- `redis_cache_search.go` still has `//go:build !windows && cgo`, while the remaining Redis implementation calls methods defined only in that file. Please remove the unsupported constraint or provide implementations for excluded builds.

Once these are fixed, this should be ready for another pass.

ZireaelK:
Thanks for the review, now fixed.

FAUST-BENCHOU (commented):
Could we clarify the expected behavior for legacy Redis string entries? They can still be read by the fallback, but without StoredAt they remain AgeKnown=false and won't satisfy MaxAge checks. Is that intentional?

ZireaelK:
Yes, this is intentional. For legacy string entries, the creation timestamp is unknown. If a client requests with a `max-age`, we should treat the entry as a miss, and it will query the upstream and overwrite the key with the new format. For queries without `max-age`, the legacy entries will be served as hits to prevent cache stampedes.

drivebyer (commented):
AddExact HSETs the key directly, but on a surviving legacy string key Redis returns WRONGTYPE — the write fails until the old key expires (forever with TTL=0). Please DEL+HSET+EXPIRE in a pipeline, and add a legacy-overwrite test.

ZireaelK:
Thans for catching the case, updated `AddExact` and added legacy-overwrite test.

Xunzhuo (approved):
LGTM

PR #3072, opened 2026-08-27
[Feature] Cache and batch tool embeddings during request-time tool selection

### LINKED ISSUE #3053: [Performance] Cache and batch tool embeddings during request-time tool selection
## Objective

Remove repeated, sequential embedding work from request-time semantic tool selection.

## Current behavior

Tool descriptions are embedded during request processing. Stable tool definitions may be embedded again on every request, and multiple tools are processed sequentially. This makes tool-selection latency grow with the number of tools even when their schemas have not changed.

## Scope

- Canonicalize tool definitions and cache embeddings by schema, embedding model, and relevant configuration.
- Batch cache misses where the embedding provider supports batching.
- Bound memory use and define invalidation when a tool schema or embedding model changes.
- Emit cache hit, miss, batch size, and selection-latency metrics.
- Preserve deterministic ranking and safe fallback behavior.

## Acceptance criteria

- Repeated requests with unchanged tools reuse cached embeddings.
- Multiple misses are embedded in a batch where supported.
- Cache size and invalidation are configurable and tested.
- A benchmark reports selection latency and embedding calls across representative tool counts.
- Ranking results remain equivalent within documented numeric tolerance.

## Ownership

**Parent:** #2992  
**Related direction:** #2361

drivebyer:
Closes #3053

## Purpose

The `tool_selection` plugin's **filter mode** re-embedded every request-supplied tool definition, one text at a time, on every request — selection latency grew linearly with the tool count even when no schema had changed. This PR makes the hot path cache-first and batch-filling (modules: `pkg/extproc`, `pkg/embedding`, `pkg/observability/metrics`; owned by the accepted issue #3053, wg/data-plane-networking):

- **Cache**: unchanged tools are served from a bounded in-process FIFO memo (2048 entries, ~6MB upper bound at 768 dims). Keys combine the embedding text with the full model identity (backend + endpoint + model + dimension), so invalidation is structural and needs no TTL — an embedding is a deterministic pure function of (text, model).
- **Batch**: cache misses are deduplicated and filled through the remote provider's existing `EmbedBatch` in chunks of 64, the query joining the same call — a cold 100-tool request drops from 101 sequential HTTP round trips to 2 batched calls, a warm one to a single 1-text call. A chunk rejected by the endpoint degrades to per-text embedding, so batching never fails a workload that previously succeeded. The local candle path keeps per-text FFI calls.
- **Metrics**: `vsr_tool_embedding_cache_hits_total` / `vsr_tool_embedding_cache_misses_total` counters and a `vsr_tool_embedding_batch_size` histogram, recorded for completed passes only.
- **Fixed along the way**: the embedding provider is built once per router and shared with the tools database (previously rebuilt per filter request); remote embedding RPCs run under the request context so client disconnects cancel them; a provider construction error at startup degrades filter mode to its fallback instead of failing per request.

The second commit consolidates the repo's three duplicate bounded-FIFO embedding memos (`pkg/cache`, contextcompression's scorer, the new tool memo) into one shared `embedding.Memo` primitive — only the data structure is shared, all semantic choices stay caller-side; public APIs unchanged, net −70 lines.

### Scope decisions relative to the issue text

1. **No canonicalization by full tool schema** — filter-mode scoring embeds only `name + description` (unchanged here), so keying by the embedding text is complete; a parameters hash can be added to the key later if wanted.
2. **No local candle batch inference** — the Go binding has no multi-text embedding FFI; that is Rust-side work better tracked separately.
3. **Cache size is a fixed default (2048), not configurable** — follows the repo's constant-bounded memo precedent; a knob is easy to add later.
4. **Error text changed** — per-tool `embedding tool %q` becomes batch-level errors with positional attribution (`embedding text %d/%d`); no in-repo consumer matched the old strings.
5. **Cache metrics count completed passes only**, so an aborted fill never skews hit/miss or batch-size stats.

## Test Plan

```bash
cd src/semantic-router
go build ./pkg/...
go test ./pkg/extproc/ ./pkg/embedding/ ./pkg/contextcompression/ -count=1
go test ./pkg/cache/ -count=1 -run 'InMemory'   # Milvus/Valkey suites need live backends
go test ./pkg/extproc/ -bench BenchmarkToolSelectionFilterEmbed -run '^$' -benchtime 100x
golangci-lint run -c tools/linter/go/.golangci.agent.yml ./pkg/extproc/... ./pkg/embedding/...
```

New tests cover: warm requests embed exactly one text (the query); one `EmbedBatch` call per ≤64-miss chunk plus the per-text degrade path; memo key identity across backends/endpoints/models; cold/warm/fresh ranking equivalence (exact per-tool score equality).

## Test Result

- All listed builds, tests, and lint pass locally (darwin/arm64, candle library built).
- Benchmark (fake provider, so overhead only — the avoided work is the point): warm passes embed 1 text regardless of tool count vs N+1 cold; at 100 tools, cold makes 2 batched calls for 101 texts, warm makes 1 call for 1 text.
- Remaining risk: remote batch-vs-single floating-point differences on a text's *first* embedding are provider-dependent and unchanged by this PR; ranking after that first embedding is bit-identical by construction.
- `pkg/cache` Milvus/Valkey integration suites were not run (need live services); the in-memory suite covering the touched code is green.

---
<details>

</details>


FAUST-BENCHOU (approved):
more likely to be a refractor.But need to resolve conflicts

wilsonwu (approved):

PR #3075, opened 2026-08-27
[Feature] Optional NLI polarity tier for the in-memory semantic cache

### LINKED ISSUE #2751: [Router] Optional NLI polarity tier (L2) for in-memory semantic cache negation guard
## What would you like to be added?

An **optional L2 semantic tier** for the semantic-cache polarity guard, layered
on top of the L1 lexical guard shipped for #2691. When enabled, a candidate
cache hit that clears the bi-encoder `similarity_threshold` and passes the L1
lexical guard is verified once with the router's **in-repo NLI** model
(`candle.ClassifyNLI`, `tasksource/ModernBERT-base-nli`, added in #1865): if the
incoming query and the matched entry query are classified **contradiction**
above a configurable threshold, the hit is rejected.

Proposed config (opt-in, default keeps L1-only behaviour):

```yaml
semantic_cache:
  polarity_guard:
    mode: lexical | nli | lexical+nli   # default: lexical (L1 only)
    nli:
      model_id: models/ModernBERT-base-nli
      contradiction_threshold: 0.5
```

Design constraints:

- **L1 stays the unconditional, zero-dependency floor.** L2 never replaces it;
  it only adds coverage when the model is configured.
- **NLI runs at most once per lookup**, on the single best candidate right
  before serving — never per-entry inside the scan loop (that would multiply the
  model cost by the cache size).
- **Opt-in, not mandatory.** Enabling `mode: nli|lexical+nli` binds the NLI
  model; leaving the default `lexical` requires no model and does not regress
  existing cache users. Model absence with an NLI mode selected should fail at
  **config-load** (validate early), not at runtime.
- Reuse the existing `SetGroundingBackends`-style function injection so `pkg/cache`
  does not import `pkg/classification` (avoids an import cycle).

## Why is this needed?

The L1 lexical guard (#2691) is a cue/antonym-table guard: it deterministically
catches single-token negation and known antonym flips, but by design misses
cue-less semantic negation, ordering-only polarity changes, and non-English
queries. The router already ships an NLI model that closes exactly this gap.

Measured on `tasksource/ModernBERT-base-nli` via the production candle path,
on the same polarity pairs from #2691:

| pair (question form) | contradiction |
|---|---|
| enable / disable dark mode | 0.983 |
| "reset my password" + inserted **not** | 0.967 |
| open / close the file | 0.993 |
| start / stop the server | 0.997 |
| add / remove a user | 0.987 |
| genuine paraphrase (reset ⇄ reset) | **0.001** |
| synonym (delete ⇄ remove) | **0.001** |

Discrimination margin is ~0.96 vs ≤0.002 — a fixed `0.5` threshold separates
cleanly with no tuning. It also correctly treats `delete`/`remove` as a synonym
(entailment), a case the L1 antonym table does not contain. Latency is
~70 ms/call single-thread CPU (p99 ~97 ms), one-time init ~330 ms; materially
lower on GPU. Since a cache hit avoids a full multi-second LLM generation, a
single ~70 ms verification before serving is net-positive.

## Additional context

- Follow-up to bug #2691 (L1 lexical guard). This issue is the L2 enhancement;
  it does **not** re-open the bug.
- Distinct from the heavier external cross-encoder rerank tracked in
  #2247 / #2251 / #2236, which remains deferred; this reuses the lighter,
  already-present in-repo NLI head.
- Related: #2341 (adaptive thresholding), #2473 / #2488 (per-lookup similarity +
  context on the same cache surface).

yaojiejia:

Closes #2751

## Purpose

The semantic cache picks a cached answer by embedding similarity. Opposite questions like "How do I turn on dark mode?" and "How do I turn off dark mode?" score above the similarity threshold, so the cache can return the wrong answer (#2691). Raising the threshold does not help, because real paraphrases often score lower than these opposites.

This PR adds an optional second check. When it is turned on, the in-memory cache runs the router's NLI model once on the best candidate before serving it. If the model says the two questions contradict each other, the cache treats it as a miss.

New config under global.stores.response_cache:

polarity_guard:
  mode: lexical          # lexical (default) | nli | lexical+nli
  nli:
    contradiction_threshold: 0.5

- Default lexical changes nothing. Only nli and lexical+nli turn the check on.
- The check runs once per lookup, outside the cache lock, only on the winning candidate.
- A rejected candidate is a normal miss. It logs cache_negation_reject with tier: nli and still reports its score on x-vsr-cache-similarity.
- If the model fails at lookup time, the cache serves the hit anyway and logs a cache_polarity_nli_skipped warning (fail open).
- The model is the hallucination explainer already in the model catalog (tasksource/ModernBERT-base-nli). The native binding can only hold one NLI model, so the guard reuses it instead of taking its own model_id. This is the one difference from the issue text.
- Config loading fails if an NLI mode is set without that model. Startup fails if the model cannot load.
- The lexical tier itself is #2728. This PR only adds the NLI tier and the config knob.

## Affected modules:

- pkg/config: PolarityGuardConfig, validation, NeedsLocalNLIForSemanticCache(), reference config
- pkg/cache: SetPolarityVerifier injection, options, the check in finishFindSimilarSearch
- pkg/classification: classifier.semantic_cache_nli runtime task loads the model and injects the verifier
- pkg/extproc: passes the config into the cache
- pkg/modeldownload, pkg/apiserver: download and /models reporting use the new predicate
- docs: stores-and-tools.md, response-cache.md
- e2e: new semantic-cache-polarity case on the production-stack profile
- tools/agent/docs/tech-debt/td-047: the operator CRD and dashboard do not expose polarity_guard yet

Out of scope: hybrid and remote cache backends, CRD and dashboard mirrors.

## Test Plan

Unit and gates:

make agent-validate
make agent-lint CHANGED_FILES="<changed files>"
make agent-ci-gate ENV=cpu CHANGED_FILES="<changed files>"
cd src/semantic-router
go test ./pkg/config/ ./pkg/modeldownload/ -count=1
go test ./pkg/cache/ ./pkg/classification/ ./pkg/extproc/ ./pkg/apiserver/ -count=1
go test ./pkg/cache/ ./pkg/classification/ -race -run 'Polarity|SemanticCache' -count=1

New tests:

- pkg/config: mode and threshold validation, missing-model error, NeedsLocalNLIForSemanticCache
- pkg/cache/polarity_nli_test.go: reject, serve, threshold is strict, fail open on error and on missing verifier, cancelled context, one call per lookup, replacing the verifier while lookups run
- pkg/cache/polarity_nli_regression_test.go: real model, skips if models/mom-halugate-explainer is not present
- pkg/classification: task registration (not best-effort), init error on a bad model path

E2E on a local Kind cluster:

make e2e-test E2E_PROFILE=production-stack E2E_TESTS=semantic-cache,semantic-cache-polarity E2E_VERBOSE=true

Test Result

- make agent-validate, make agent-lint, make agent-ci-gate: pass.
- go test for pkg/config, pkg/modeldownload, pkg/cache, pkg/classification, pkg/extproc, pkg/apiserver: all pass. -race on the polarity tests: pass.
- E2E production-stack: semantic-cache passes; semantic-cache-polarity passes. The router downloaded models/mom-halugate-explainer on startup and logged semantic_cache_nli_initialized. All three opposite questions were rejected with similarity 0.93-0.97 and NLI contradiction 0.95-0.99; all three paraphrases were served.
- Measured on the deployed model: the guard also rejects unrelated questions that the embedder scores as similar (for example "turn on the lights" vs "add a contact", similarity 0.83). It did not catch "Should I commit this change?" vs "Should I not commit this change?" (similarity 0.97, contradiction below 0.5). That negation case is what the lexical tier in #2728 is for.

---
<details>

- [x] PR title begins with exactly one bracketed category, such as [Feature], [Bug], [Docs], [Test], [Research], [Community], or [CI/Build]
- [x] The title does not stack prefixes such as [Router][Docs]; affected modules belong in labels and the PR body
- [x] The PR links an accepted issue with exactly one owner: one wg/* label for project work or owner/maintainers for repository governance

</details>

See CONTRIBUTING.md for the full contributor workflow and commit guidance.

### DIFF
--- a/.github/workflows/integration-test-k8s.yml
+++ b/.github/workflows/integration-test-k8s.yml
@@ -132,7 +132,7 @@ jobs:
           set +e
           if [ "${{ matrix.profile }}" = "envoy-ai-gateway" ]; then
             # Temporarily skip the stress / pressure coverage until the suite is stable again.
-            ENVOY_AI_GATEWAY_CI_TESTS="chat-completions-request,apiserver-runtime-config-endpoints,domain-classify,semantic-cache,pii-detection,jailbreak-detection,decision-priority-selection,plugin-chain-execution,tool-selection,rule-condition-logic,decision-fallback-behavior,plugin-config-variations"
+            ENVOY_AI_GATEWAY_CI_TESTS="chat-completions-request,apiserver-runtime-config-endpoints,domain-classify,semantic-cache,semantic-cache-polarity,pii-detection,jailbreak-detection,decision-priority-selection,plugin-chain-execution,tool-selection,rule-condition-logic,decision-fallback-behavior,plugin-config-variations"
             make e2e-test E2E_PROFILE=${{ matrix.profile }} E2E_TESTS="${ENVOY_AI_GATEWAY_CI_TESTS}" E2E_VERBOSE=true E2E_KEEP_CLUSTER=false
             TEST_EXIT_CODE=$?
             set -e
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1833,6 +1833,14 @@ global:
       ttl_seconds: 7200
       eviction_policy: lru
       embedding_model: mmbert
+      # Negation/antonym guard for the in-memory backend. "lexical" is the
+      # model-free default; "nli" or "lexical+nli" additionally verifies the
+      # winning candidate once with the hallucination explainer NLI model and
+      # rejects it above nli.contradiction_threshold (~70 ms/hit on CPU).
+      polarity_guard:
+        mode: lexical
+        nli:
+          contradiction_threshold: 0.5
       milvus:
         connection:
           host: milvus
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -17,6 +17,8 @@ var BaselineRouterContract = []string{
 	"chat-completions-stress-request",
 	"domain-classify",
 	"semantic-cache",
+	// NLI polarity tier of the semantic cache (issue #2751)
+	"semantic-cache-polarity",
 	"pii-detection",
 	"jailbreak-detection",
 	"decision-priority-selection",
--- a/e2e/profiles/ai-gateway/values.yaml
+++ b/e2e/profiles/ai-gateway/values.yaml
@@ -788,6 +788,12 @@ config:
         ttl_seconds: 3600
         eviction_policy: fifo
         embedding_model: mmbert
+        # Exercised by the semantic-cache-polarity case: the NLI tier reuses the
+        # default hallucination explainer (models/mom-halugate-explainer).
+        polarity_guard:
+          mode: lexical+nli
+          nli:
+            contradiction_threshold: 0.5
       memory:
         embedding_model: mmbert
       vector_store:
--- a/e2e/profiles/production-stack/profile.go
+++ b/e2e/profiles/production-stack/profile.go
@@ -163,6 +163,7 @@ func (p *Profile) GetTestCases() []string {
 			"pii-detection",
 			"jailbreak-detection",
 			"semantic-cache",
+			"semantic-cache-polarity",
 		},
 	)
 }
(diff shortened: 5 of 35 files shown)


ramkrishs (changes requested):
PR diff looking good, Kindly fix the build failures, thanks

yaojiejia:
> PR diff looking good, Kindly fix the build failures, thanks

thanks, can you look at it again

ramkrishs (approved):
LGTM

wilsonwu (approved):

PR #3078, opened 2026-08-28
[Docs] animation and consistent section heading punctuation

abhinav-m22:
Related #3002

Follow-up to #3069, which merged before these review points were addressed:
- https://github.com/vllm-project/semantic-router/pull/3069#issuecomment-5438492055
- https://github.com/vllm-project/semantic-router/pull/3069#issuecomment-5448193271

## Purpose

- **Data sovereignty animation.** A scan bar sweeps each stage, the arrows pulse as the request passes, and the constraint gate glows when reached. Same visual language as the routing pipeline above it, and disabled under `prefers-reduced-motion`.
- **Docs intro lockup.** The brand card stretched to the full prose measure while the logo and tagline filled only its left portion. It now hugs its content, so there is no dead space on the right.
- **Heading punctuation.** Eleven of the fourteen homepage section headings ended with a period and three did not. Removed them everywhere rather than adding the missing three. The eleven matching Chinese headings are updated too.
- **Product term casing.** `One model API` to `One Model API`. Audited the rest of the visible homepage copy: `Semantic Router`, `Mixture-of-Models`, `Playground`, `Dashboard` and `Fusion API` were already correct.

Section text stays full width, as agreed on the first comment above.

## Test Plan

- Sovereignty stages animate in sequence and the gate pulses; animation stops under reduced motion.
- Docs intro card wraps its logo and tagline with no trailing gap, and stacks below 576px.
- No homepage section heading ends with a period, in either locale.

## Test Result

Screenshot From 2026-08-28 09-19-48

Screenshot From 2026-08-28 09-20-04

image

### DIFF
--- a/website/i18n/zh-Hans/code.json
+++ b/website/i18n/zh-Hans/code.json
@@ -478,9 +478,6 @@
   "homepage.capabilities.preference.value": {
     "message": "将偏好表达为路由策略。"
   },
-  "homepage.capabilities.docsCta": {
-    "message": "了解工作原理"
-  },
   "homepage.capabilities.signal.title": {
     "message": "信号提取"
   },
--- a/website/i18n/zh-Hans/code.json
+++ b/website/i18n/zh-Hans/code.json
@@ -425,7 +425,7 @@
     "message": "架构"
   },
   "homepage.capabilities.heading": {
-    "message": "统一异构推理。"
+    "message": "统一异构推理"
   },
   "homepage.capabilities.description": {
     "message": "在四个维度上统一碎片化的模型生态。"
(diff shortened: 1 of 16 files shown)


AayushSaini101:
Thanks for working on it @abhinav-m22 Suggestion, can you move the animation part above 
image after how it works section and also can we make the heading style consistent 
Differentimage
image

Different syles are there let me know your thoughts on this thanks : )

abhinav-m22:
makes sense @AayushSaini101.. 

moved the data sovereignty section up to sit right after "How it works", and made the section eyebrows consistent across the homepage (there were three different styles, all now use the shared one).

Please check

AayushSaini101:
> makes sense @AayushSaini101..
> 
> moved the data sovereignty section up to sit right after "How it works", and made the section eyebrows consistent across the homepage (there were three different styles, all now use the shared one).
> 
> Please check

Amazing thank you for efforts : )

AayushSaini101 (approved):

PR #3080, opened 2026-08-28
[Bug] Rank decisions by confidence only when the whole pool reports scores

### LINKED ISSUE #3079: [Bug] Signals with no reported confidence default to 1.0 and override priority in tiered selection
### Proposed Workgroup

(The selection behavior itself may belong to MoM & Routing, happy to have this re-triaged.)

### Describe the bug

Within a tier (and under `strategy: confidence`), decisions are ranked by confidence first; priority only breaks ties. But in `pkg/decision/engine.go`:

- keyword, language, fact_check, user_feedback, context, modality, authz and pii never write `SignalConfidences`, so their leaves default to 1.0
- predicate leaves and passing NOT guards also return 1.0

So any match through one of these branches outranks a real classifier score, and priority never applies.

### To Reproduce

Shipped `config/recipes/agent/config.yaml`, tier 3. A legal request: `domain:law` at 0.88, both policy projections at 0.90.

| decision | priority | confidence | result |
|---|---|---|---|
| domain_legal_health | 260 | mean(.90, .90, .88) = 0.893 | should win |
| domain_code, when a word trips its `code_request_markers` keyword branch | 250 | mean(.90, .90, keyword 1.0, NOT 1.0, ...) = 0.960 | wins |

Remove the keyword hit and routing is correct.

<details><summary>standalone test, no models needed</summary>

```go
func TestKeywordDefaultInvertsPriority(t *testing.T) {
	raw, _ := os.ReadFile("../../../../config/recipes/agent/config.yaml")
	var doc struct {
		Routing struct {
			Strategy  string            `yaml:"strategy"`
			Decisions []config.Decision `yaml:"decisions"`
		} `yaml:"routing"`
	}
	yaml.Unmarshal(raw, &doc)
	engine := NewDecisionEngine(nil, nil, nil, doc.Routing.Decisions,
		config.RoutingStrategy(doc.Routing.Strategy))

	result, _ := engine.EvaluateDecisionsWithSignals(&SignalMatches{
		ProjectionRules: []string{"policy_privacy_cloud_allowed", "policy_security_standard"},
		DomainRules:     []string{"law"},
		KeywordRules:    []string{"code_request_markers"},
		SignalConfidences: map[string]float64{
			"projection:policy_privacy_cloud_allowed": 0.90,
			"projection:policy_security_standard":     0.90,
			"domain:law":                              0.88,
		},
	})
	// observed: domain_code (250, 0.960)
	// expected: domain_legal_health (260, 0.893)
}
```

</details>

### Expected behavior

A default constant should not outrank a reported score. When a pool's confidences are not comparable, selection should fall back to priority.

### Additional context

Proposed fix: track during evaluation whether a decision's confidence came entirely from reported scores, and rank by confidence only in pools where every member is scored, falling back to priority otherwise. Confidence values and the empty-AND catch-all behavior stay unchanged.

Several of these signals already compute a confidence internally and never report it (language checks it against a per-rule threshold, fact_check and pii do the same, bm25 keyword rules have a score). Reporting those is worth doing on its own, but boolean signals and NOT/predicate gates would still rank as constants, so selection needs the comparability fallback either way.

Calibrating reported scores across models (classifier probability vs cosine similarity) is separate again, probably under #2974.

subin9:
Closes #3079

## Purpose

Within a tier (and under `strategy: confidence`) decisions are ranked by confidence before priority. Signals that never write `SignalConfidences` (keyword, language, fact_check, user_feedback, context, modality, authz, pii) rank as the default 1.0, and predicate leaves and passing NOT guards return 1.0 outright. Any decision matching through those branches outranks a real classifier score, and priority never gets checked. #3079 has the details and a repro against the shipped agent recipe.

The change keeps confidence values, traces and metrics as they are and only touches ranking:

- evaluation tracks whether every value in a decision's confidence came from a reported score (`ConfidenceScored`, also exposed on trace nodes)
- selection ranks by confidence only in pools where every member is scored, and falls back to priority otherwise
- catch-alls (omitted rules / empty AND) keep zero confidence and now rank last explicitly wherever confidence applies

Behavior changes for any pool containing an unscored decision: it now follows priority, which is the ordering the docs describe. Fully scored pools rank as before, with one corner exception: a decision that reported an exact zero confidence now outranks a catch-all instead of falling through to priority.

## Test Plan

```
gofmt -l pkg/decision/
go vet  ./pkg/decision/
go test ./pkg/decision/ -race
```

New tests in `engine_confidence_scored_test.go`: unscored pool falls back to priority, fully scored pool still ranks by confidence, catch-all stays last in an unscored pool, an OR that matched only through an unscored branch stays unscored, priority strategy without tiers is unchanged.

## Test Result

```
gofmt    clean
go vet   clean
go test  ok with -race, 0 failed (contract matrix, tier, empty-AND, predicate and trace tests all pass unmodified)
```

One internal signature changed: `signalConfidence` now also returns whether a score was reported, and its test was adapted.

Also loaded the tier-3 decisions from `config/recipes/agent/config.yaml` and replayed the #3079 scenario against main and this branch: main picks domain_code (priority 250, confidence 0.960) on a domain:law 0.88 request with an incidental code_request_markers keyword hit; this branch picks domain_legal_health (priority 260, 0.893). Without the keyword hit both pick domain_legal_health.

Didn't run make agent-lint / agent-ci-gate locally; CI should cover those.


Xunzhuo:
Thanks, this is the right immediate fix. My understanding is:

- Today, once tiers are present, selection is `tier -> confidence -> priority` regardless of `routing.strategy`. Missing confidence and structural gates become `1.0`, so they can beat real scores.
- With this PR, confidence is used only when every decision in the competing tier is scored; otherwise that tier falls back to priority. That fixes #3079.
- Longer term, we should make the tier/strategy contract explicit and separate structural match evidence from calibrated confidence. Reported scores from different signal models are not necessarily comparable either.

Can you add a maintained agent recipe probe for a legal request that also hits `code_request_markers` and still selects `domain_legal_health`? This is the regression coverage missing from the PR.

After that, please take a first pass at the longer-term contract as a focused follow-up. I would start with whether `routing.strategy` should apply within a tier, and what makes a confidence score eligible for cross-decision ranking.

Xunzhuo (dismissed):

subin9:
Probe added: the indemnity wording with an incidental "implement", asserting code_request_markers actually fired so the collision is real, tagged collision/priority like security_over_specialist_domain. Verified locally that the query trips exactly that keyword rule and that neither existing legal probe does, so the assertion isn't vacuous.

Keeping this PR scoped to the regression fix; picking up the ranking contract in #3081.

FAUST-BENCHOU (approved):

wilsonwu (approved):

PR #3084, opened 2026-08-28
[Bug] Stop reask embedding after all rules fail

### LINKED ISSUE #3048: [Bug] reask makes one embedding for each earlier user message
### Proposed Workgroup

MoM & Routing

### Describe the bug

The `reask` signal does work and then discards it. The amount of discarded work increases with the length of the conversation and no config value limits that.

`Classify` in `src/semantic-router/pkg/classification/reask_classifier.go` does two steps. First, `computeSimilarities` makes one embedding for each different earlier user message. Second, the rules examine the results. However, the second step can stop early. `evaluateReaskStreak` stops at the first similarity below a threshold and those embeddings are not used afterward.

The first step cannot stop early. It does not read `c.rules`, and its loop limit is `len(priorUserTurns)`. A value of `lookback_turns` (how many earlier messages a rule compares) therefore does not change the number of embeddings.

A request with 50 earlier messages makes 51 embeddings. Two embeddings are enough when the most recent prior turn's similarity falls below every rule's threshold. Each embedding is one model forward pass.

### To Reproduce

Add this file to `src/semantic-router/pkg/classification/`. Run `go test ./pkg/classification/ -run TestReaskClassifier_EmbedsEveryPriorTurn -count=1 -v` from `src/semantic-router`.

``` go
package classification

import (
	"fmt"
	"sync/atomic"
	"testing"

	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
)

// The current turn is orthogonal to every prior turn, so similarity is 0,
// below the rule threshold, so the streak ends at the first prior turn.
func countingReaskEmbedder(t *testing.T, calls *int64) {
	t.Helper()
	restore := SetEmbeddingFuncForTests(func(text string, modelType string, targetDim int) (*candle_binding.EmbeddingOutput, error) {
		atomic.AddInt64(calls, 1)
		if text == "current user turn" {
			return &candle_binding.EmbeddingOutput{Embedding: []float32{1, 0}}, nil
		}
		return &candle_binding.EmbeddingOutput{Embedding: []float32{0, 1}}, nil
	})
	t.Cleanup(restore)
}

func distinctPriorTurns(n int) []string {
	prior := make([]string, n)
	for i := range prior {
		prior[i] = fmt.Sprintf("unrelated earlier question %d", i)
	}
	return prior
}

func TestReaskClassifier_EmbedsEveryPriorTurnRegardlessOfLookback(t *testing.T) {
	for _, priorTurns := range []int{1, 2, 5, 20, 50} {
		t.Run(fmt.Sprintf("prior_turns=%d", priorTurns), func(t *testing.T) {
			var calls int64
			countingReaskEmbedder(t, &calls)

			classifier, err := NewReaskClassifier([]config.ReaskRule{{
				Name:          "likely_dissatisfied",
				Threshold:     0.8,
				LookbackTurns: 1,
			}}, "test-model")
			if err != nil {
				t.Fatalf("NewReaskClassifier() error = %v", err)
			}

			matches, err := classifier.Classify("current user turn", distinctPriorTurns(priorTurns))
			if err != nil {
				t.Fatalf("Classify() error = %v", err)
			}
			if len(matches) != 0 {
				t.Fatalf("expected no reask match, got %+v", matches)
			}

			got := atomic.LoadInt64(&calls)
			// Only these two can be read: the current turn, and the prior turn
			// that ends the streak. This count does not grow with priorTurns.
			const want = int64(2)
			t.Logf("prior turns=%d embedding calls=%d needed=%d", priorTurns, got, want)
			if got != want {
				t.Errorf("embedding calls = %d, want %d: the classifier embeds every prior user turn, "+
					"so cost grows with conversation length and not with lookback_turns", got, want)
			}
		})
	}
}
```

Output (summary)

``` 
prior turns=1   embedding calls=2   needed=2
prior turns=2   embedding calls=3   needed=2   FAIL
prior turns=5   embedding calls=6   needed=2   FAIL
prior turns=20  embedding calls=21  needed=2   FAIL
prior turns=50  embedding calls=51  needed=2   FAIL
```

### Expected behavior

Stop the loop at the first similarity below the lowest threshold of all the rules.

### Impact

Every request evaluating the `reask` signal in a long conversation pays O(conversation length) embedding forward passes, but most are discarded

### Additional context

Some parts of this bug report are AI generated

wilsonwu:

Closes #3048

## Purpose

- Stop the reask classifier from generating embeddings for every earlier user
  message after the active reask streak has already failed.
- Resolve each rule's configured or default threshold and stop scanning older
  turns when a similarity falls below the lowest threshold across all rules. At
  that point, no rule can recover its consecutive-match streak.
- Preserve existing matching behavior, including mixed rule thresholds,
  `lookback_turns`, reported confidence, and matched-turn counts.
- Add focused regression coverage proving that:
  - a 50-turn conversation stops after embedding the current and most recent
    user turns when all rules fail; and
  - a lower-threshold rule continues to be evaluated after a stricter rule
    fails, while messages older than the all-rules failure are not embedded.
- Affected module: Router classification runtime.
- Tracking issue and owner: #3048, `wg/mom-routing`.

## Test Plan

```bash
cd src/semantic-router

go test -count=1 ./pkg/classification \
  -run '^TestReaskClassifier_Classify(StopsEmbeddingAfterAllRulesFail|ContinuesUntilLowestThresholdFails)$'

go test -count=1 ./pkg/classification

make recipe-conformance-static

make agent-lint \
  CHANGED_FILES="src/semantic-router/pkg/classification/reask_classifier.go,src/semantic-router/pkg/classification/reask_classifier_optimization_test.go"

make codespell-tracked

make test-semantic-router
```

## Test Result

- Focused reask optimization tests passed.
- The complete `pkg/classification` test package passed.
- `make recipe-conformance-static` passed:
  - 50 harness tests passed;
  - maintained recipe decisions, entrypoints, fallbacks, algorithms, and plugins
    retained complete required coverage.
- Changed-file lint passed:
  - Go formatting passed;
  - Go structural lint passed;
  - supply-chain scanning passed;
  - repository structure checks passed.
- `make codespell-tracked` passed.
- VS Code reported no diagnostics in the changed files.
- The full `make test-semantic-router` gate could not complete in the local
  environment. Tests in `pkg/cache`, `pkg/extproc`, and `pkg/tools` attempted to
  download `sentence-transformers/all-MiniLM-L6-v2` from Hugging Face and timed
  out. The `pkg/extproc` suite eventually reached its test timeout after repeated
  download failures.
- No failure was reported in the changed classification package. Remaining risk
  is limited to the live model-dependent and local smoke paths that could not be
  exercised while the external model download was unavailable.

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/classification/reask_classifier.go
+++ b/src/semantic-router/pkg/classification/reask_classifier.go
@@ -51,7 +51,7 @@ func (c *ReaskClassifier) Classify(currentUserTurn string, priorUserTurns []stri
 		return nil, fmt.Errorf("failed to compute current user turn embedding: %w", err)
 	}
 
-	similarities, err := c.computeSimilarities(currentEmbedding, priorUserTurns)
+	similarities, err := c.computeSimilarities(currentEmbedding, priorUserTurns, minimumReaskThreshold(c.rules))
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +79,7 @@ func (c *ReaskClassifier) Classify(currentUserTurn string, priorUserTurns []stri
 	return retainMaxLookbackReaskMatches(matches), nil
 }
 
-func (c *ReaskClassifier) computeSimilarities(currentEmbedding []float32, priorUserTurns []string) ([]float64, error) {
+func (c *ReaskClassifier) computeSimilarities(currentEmbedding []float32, priorUserTurns []string, minimumThreshold float64) ([]float64, error) {
 	cache := make(map[string][]float32, len(priorUserTurns))
 	similarities := make([]float64, 0, len(priorUserTurns))
 
@@ -99,10 +99,25 @@ func (c *ReaskClassifier) computeSimilarities(currentEmbedding []float32, priorU
 			cache[priorTurn] = priorEmbedding
 		}
 
-		similarities = append(similarities, float64(cosineSimilarity(currentEmbedding, priorEmbedding)))
+		similarity := float64(cosineSimilarity(currentEmbedding, priorEmbedding))
+		similarities = append(similarities, similarity)
+		if similarity < minimumThreshold {
+			break
+		}
 	}
 
 	return similarities, nil
+}
+
+func minimumReaskThreshold(rules []config.ReaskRule) float64 {
+	minimumThreshold := float64(rules[0].WithDefaults().Threshold)
+	for _, rawRule := range rules[1:] {
+		threshold := float64(rawRule.WithDefaults().Threshold)
+		if threshold < minimumThreshold {
+			minimumThreshold = threshold
+		}
+	}
+	return minimumThreshold
 }
 
 func (c *ReaskClassifier) embedText(text string) ([]float32, error) {
--- a/src/semantic-router/pkg/classification/reask_classifier_optimization_test.go
+++ b/src/semantic-router/pkg/classification/reask_classifier_optimization_test.go
@@ -1,0 +1,104 @@
+package classification
+
+import (
+	"fmt"
+	"testing"
+
+	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+func TestReaskClassifier_ClassifyStopsEmbeddingAfterAllRulesFail(t *testing.T) {
+	embeddingCalls := 0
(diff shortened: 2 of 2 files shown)


AayushSaini101 (approved):
LGMT thanks : )

FAUST-BENCHOU (approved):

PR #3087, opened 2026-08-28
[Feature] Add OrcaRouter as a first-class provider

martinzudergaming-a11y:
[Feature] Add OrcaRouter as a first-class provider

## Purpose

This change lets vLLM Semantic Router users route to [OrcaRouter](https://www.orcarouter.ai) as a named provider instead of an anonymous custom base URL. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents — like OpenRouter it exposes a provider/model namespace across many models, but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` to the provider registry means `providers.models[].backend_refs` can carry `provider: orcarouter`, and the router resolves its auth header, chat path, and per-user credential header the same way it does for `openai`, `anthropic`, or `minimax` — no generic any-endpoint passthrough needed.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The implementation mirrors the existing `minimax` integration exactly: a new `LLMProvider` constant (`orcarouter`), a per-user injected credential header (`x-user-orcarouter-key`), a `DefaultHeaderMap` registration, and a `providerTypeRegistry` entry using `Authorization: Bearer` with the `/chat/completions` chat path (OrcaRouter is OpenAI-compatible).

## Test Plan

- `go build ./pkg/authz/... ./pkg/config/... ./pkg/headers/...`
- `go vet ./pkg/authz/... ./pkg/config/... ./pkg/headers/...`
- `gofmt -l` on changed files (clean)
- `go test -count=1 ./pkg/authz/... ./pkg/config/... ./pkg/headers/...`

## Test Result

All three packages pass. New tests cover header injection, credential resolution, fail-closed no-key behavior, custom header maps, provider type validation, auth header resolution, and chat path resolution for `orcarouter`.

Live verification: `GET https://api.orcarouter.ai/v1/models` and `POST https://api.orcarouter.ai/v1/chat/completions` with a `Bearer` key both return HTTP 200 with OpenAI-format responses, confirming the auth scheme and chat path configured in the registry.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.


FAUST-BENCHOU (commented):
not sure whether we need this.I dont really recommand to add this into code.Maybe blog or example is better

AayushSaini101:
> not sure whether we need this.I dont really recommand to add this into code.Maybe blog or example is better

Agree with @FAUST-BENCHOU we can introduce any blog or research paper instead of mentionining hard core naming component thanks, closing the PR thanks : )

PR #3090, opened 2026-08-28
[Bug] Remove multi-document separators from Helm values.yaml (#3071)

### LINKED ISSUE #3071: [Bug] Helm v4 silently sets image.pullPolicy to Never from a malformed values.yaml
### Proposed Workgroup

### Describe the bug

`deploy/helm/semantic-router/values.yaml` is not a single yaml document. It's 41 documents concatenated with `---` separators. So does `ghcr.io/vllm-project/charts/semantic-router` v0.3.0

Whether this causes a problem depends on the installing Helm's major version:

- **Helm v3** loads `values.yaml` with a plain single-document unmarshal, so only the first document (the real, documented defaults) is used and the other 40 are silently discarded.
- **Helm v4** added multi-document values support (`internal/chart/v3/loader/load.go`, `LoadValues`) and deep-merges all documents in order, last value wins per leaf key. Most of the ~40 accidental trailing documents are near-duplicate blocks, but several redefine real leaf keys — most visibly:

  ```
  image:
    pullPolicy: Never      # README documents the default as "IfNotPresent"
    tag: latest
  ```

  The same merge also changes other defaults less visibly (e.g. `resources.limits`, some `config:` sub-fields) wherever a trailing document happens to redefine them.

### To Reproduce

Same chart digest, same version, zero values overrides, only the Helm client differs:

1. `helm upgrade <release-name> oci://ghcr.io/vllm-project/charts/semantic-router --version 0.3.0 --install` with Helm **v3.19.0**, then `helm get values <release-name> --all`:
   ```
   image:
     pullPolicy: IfNotPresent
     tag: ""
   ```
2. The identical command with Helm **v4.2.3**:
   ```
   image:
     pullPolicy: Never
     tag: latest
   ```
3. On a cluster with no locally cached copy of the image (e.g. a fresh `kind` cluster), installing with Helm v4 and a minimal values file that never touches `image:` fails at the pod level:
   ```
   kubectl describe pod -l app.kubernetes.io/instance=sr
   # -> Warning  ErrImageNeverPull  ... image "...extproc:latest" is not present with pull policy of Never
   ```

### Expected behavior

`image.pullPolicy` should default to `IfNotPresent`, like the chart's README says. That shouldn't change based on which Helm version you install with.

### Impact

On Helm v4, every install or upgrade of this chart gets `image.pullPolicy: Never` by default. You'll only see it fail if the node doesn't already have the image cached locally. On Helm v3 this doesn't happen at all.

**Workaround:** explicitly set `image.pullPolicy: IfNotPresent` (or `Always`) in your own values file when installing with Helm v4.

### Additional context

- I found the commit `043cee97` (PR: \#1553) jumped `values.yaml` to 11,531 lines and introduced all 40 `---` separators at once, adding dozens of unlabeled near-duplicate blocks with no matching documented feature. Before that, the file on commit `5b75ee3a` was a normal 609-line, single-document file.
- Helm behavior update: Refer to [Helm 4 Full Changelog](https://helm.sh/docs/changelog/)
  > #13655	2025-02-20	LuBingtan	feat: support multi-document values files
   
    Corresponding helm issue: [helm issue #13907](https://github.com/helm/helm/issues/13097)
- Suggested fix: `values.yaml` should be a single YAML document containing only the real chart defaults. If any content in the ~40 extra blocks is actually needed, it should move to its own file (maybe under `e2e/` or `ci/`), not ship inside the chart's `values.yaml`.

Benlite777:
## Purpose

  Strips the 40 accidental YAML document separators (`---`) and their trailing content from `deploy/helm/semantic-router/values.yaml`, reducing it from 11,580 lines back to the original 742-line single-document defaults.

  Helm v4 added multi-document values support and deep-merges all documents. The trailing blocks override `image.pullPolicy` to `Never` (among other keys), breaking fresh installs on any cluster without a pre-cached image. Helm v3 silently discards extra documents so the problem only surfaces on v4.

- **Which module(s) does this affect?** `Deploy/Helm`

Fixes #3071

## Test Plan

- Verified the remaining `values.yaml` is valid single-document YAML (`python -c "import yaml; yaml.safe_load(open('deploy/helm/semantic-router/values.yaml'))"`)
- Confirmed `image.pullPolicy` defaults to `IfNotPresent` (matching chart README)
- No `---` separators remain in the file
- Checked that no unique, production-relevant config was lost (trailing blocks were near-duplicate dev/test scenarios already covered by `e2e/profiles/`)

## Test Result

- YAML validation passed
- `grep pullPolicy` shows only `IfNotPresent` (2 occurrences: main image + sidecar)
- File reduced from 11,580 lines to 742 lines


FAUST-BENCHOU (dismissed):
LG

FAUST-BENCHOU:
pls fix the DCO with commit -s

FAUST-BENCHOU (approved):

wilsonwu (approved):

PR #3093, opened 2026-08-28
[Feature] Enforce bounded I/O on the http_classify connector

### LINKED ISSUE #2924: [Feature] Enforce bounded request byte-size limits on the http_classify connector

guan404ming:
Closes #2924

## Purpose

- The classifier connector reads external responses with no size limit.
- A huge or malicious response can exhaust router memory every request.
- Add `max_request_bytes_kb` / `max_response_bytes_kb`, default 256KB / 1MB.
- Reject over-limit requests before dialing; cap error bodies at 8KB.
- Share one read guard, extracted from `pkg/looper` into `pkg/utils/http`.
- Status is now read before the body, so read errors cannot mask it.
- Batch limits excluded: the wire contract has no batch dimension.
- Draft until a maintainer accepts #2924.

## Test Plan

- Assert N accepted and N+1 rejected, on request and response.
- Assert truncated error bodies keep the status, and cancellation returns early.
- Run unit tests and the repo lint gates.

## Test Result

```
gofmt          clean
go test        ok  pkg/classification  8.329s
               ok  pkg/looper         13.042s
               ok  pkg/config          1.087s
               ok  pkg/utils/http      1.730s
agent-validate pass
agent-lint     pass; pre-existing WARN, model_config_types.go 419 lines (warn 400)
```

- Removed each guard in turn; the matching test failed every time.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -2199,6 +2199,19 @@ global:
         access_key: external-classifier-key
         max_tokens: 128
         temperature: 0
+      # Reached over the http_classify contract; the byte ceilings apply only there.
+      - name: external-sequence-classifier
+        llm_provider: openai
+        model_role: classification
+        llm_endpoint:
+          address: sequence-classifier
+          port: 8080
+          protocol: http
+          name: openai-sequence-classifier
+        llm_timeout_seconds: 5
+        access_key: external-sequence-classifier-key
+        max_request_bytes: 262144
+        max_response_bytes: 1048576
     kbs:
       - name: privacy_kb
         source:
--- a/src/semantic-router/pkg/classification/http_classifier.go
+++ b/src/semantic-router/pkg/classification/http_classifier.go
@@ -5,13 +5,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"math"
 	"net/http"
 	"reflect"
 	"time"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+	httputil "github.com/vllm-project/semantic-router/src/semantic-router/pkg/utils/http"
 )
 
 // probabilitySumTolerance bounds how far an http_classify response's scores
@@ -57,11 +57,13 @@ type sequenceLabelMapping interface {
 // external model (a 14-label category classifier) during #2918, not just a
 // synthetic mock.
 type HTTPClassifierInference struct {
-	httpClient *http.Client
-	baseURL    string
-	accessKey  string
-	timeout    time.Duration
-	mapping    sequenceLabelMapping
+	httpClient       *http.Client
+	baseURL          string
+	accessKey        string
+	timeout          time.Duration
+	maxRequestBytes  int64
+	maxResponseBytes int64
+	mapping          sequenceLabelMapping
 }
 
 // NewHTTPClassifierInference creates a new http_classify-backed inference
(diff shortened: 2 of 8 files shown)


FAUST-BENCHOU:
pls fix ci thanks

guan404ming:
Just fixed, please take another look. Thanks!

config/config.yaml
```diff
@@ -2199,6 +2199,8 @@ global:
         access_key: external-classifier-key
         max_tokens: 128
         temperature: 0
+        max_request_bytes_kb: 256
+        max_response_bytes_kb: 1024
```
FAUST-BENCHOU: a little wired for me.May confuse the user

wilsonwu: Such as: max_request_size: 256k

guan404ming: Thanks for taking a look!

The name follows `max_response_bytes_mb` in the same file (line 2099, `LooperConfig`). That is existing `<what>_bytes_<unit>` shape. I used `_kb` because these ceilings are much smaller than the looper's 32 MB.

Happy to change it though. Options I see:

- `max_request_size` / `max_response_size`
- `max_request_bytes` / `max_response_bytes` in raw bytes, matching `max_body_bytes` and `max_total_bytes`

Let me know which you'd prefer, or if you had something else in mind.

AayushSaini101: Thanks for the suggestion @guan404ming, we can mention like this max_request_bytes :<number> if we assign max_request_size: 23k we need to filter the value during the parsing let me know your thoughts on this thanks

guan404ming: Done. `max_request_bytes` / `max_response_bytes`, plain byte count. Agreed on avoiding the suffix parsing.

Also moved them onto a dedicated `http_classify` entry: `external-classifier` is a generative example (reached by the `type: llm` rule at line 538), so classify-only keys read as if they applied there.

AayushSaini101 (approved):
LGTM thanks : )

guan404ming:
Thanks!

Issue #2923, opened 2026-08-17
[Feature] Wire the generic ClassifierSignalRule signal to SequenceClassifierBackend


theohsiung:
### Describe the feature

`ClassifierSignalRule` (`pkg/config/classifier_signal.go`) already exposes a reusable, user-named label-score classifier signal with `type: local` (native sequence classifier) or `type: llm` (a generative chat model prompted to pick one label). It's the closest thing this codebase has to a general "bring your own custom classifier" signal, independent of the fixed jailbreak/category/PII signal types.

Two gaps keep it from being a real alternative to those fixed types:

1. `type: llm`'s `Classify` (`generic_classifier.go`) fakes a distribution — the chosen label gets score `1`, every other label gets `0` — instead of returning the model's actual probability distribution.
2. Neither `type` can point at an `http_classify`-style remote sequence classifier; the only remote option is the generative JSON-choice path.

Add a new backend type (e.g. `type: sequence_classifier`) that lets a `ClassifierSignalRule` reuse the same `SequenceClassifierBackend` implementations (Candle, mmBERT-32K, `http_classify`) that #2759/#2760 build for jailbreak/category, returning a real label distribution.

This directly answers the architectural point @adaamko raised on #2760 (2026-08-14): the current mechanism is "model pluggability, not signal extensibility" — you can swap the model behind jailbreak/category/PII, but can't add a brand-new signal type (harm, toxicity, brand policy) without misusing an existing one, mismatching taxonomy/thresholds/metrics. This closes that gap using the generic signal mechanism that already exists, instead of inventing a new one.

### Primary layer

signal

### Why this layer?

`ClassifierSignalRule` is a signal-rule mechanism (`routing.signals.classifiers`) — this issue extends its backend options, not the router-owned model catalog itself.

### Why do you need this feature?

Resolves @adaamko's "model pluggability ≠ signal extensibility" point from his 2026-08-14 review on #2760, using existing infrastructure (`ClassifierSignalRule`) instead of proposing a new generic signal mechanism.

### Additional context

Related to #2760 (raised during its review) but not a subset of its stated category/complexity/PII scope — filed as a standalone, cross-referenced issue. cc @adaamko @Xunzhuo @AayushSaini101.

theohsiung:
[...]

Issue #2966, opened 2026-08-22
[WG] Router Models & Inference Runtime — charter and Lead/Member self-nomination


Xunzhuo:
## Mission

Own the model-facing runtime and built-in Router Model lifecycle: model contracts, local and remote inference adapters, multimodal correctness, capability discovery, runtime qualification, model quality, routing-native model research, and reproducible training artifacts.

## Owned Epics

[View all current Router Models & Inference Runtime Epics](https://github.com/vllm-project/semantic-router/issues?q=is%3Aissue+is%3Aopen+label%3Aepic+label%3Awg%2Frouter-models-inference-runtime)

## Boundaries

This Workgroup owns how Router Models are represented, loaded, invoked, qualified, evaluated as models, and promoted as immutable artifacts. MoM & Routing owns how signals and observations influence routing recipes and model selection. Evaluation & Quality owns shared benchmark infrastructure, coverage, and reports. Networking & Integrations owns transport and provider integration outside the model-runtime contract.

New work should enter through an owned Epic or an explicitly reviewed new Epic. Child issues must have one Workgroup owner and a concrete deliverable.

## Membership

Use this issue for Lead and Member self-nominations. Describe your relevant background, interests, and an initial contribution area.

Xunzhuo:
@WUKUNTAI-0211, @theohsiung and @adaamko will be the lead for this direction.

Looking for workgroup leads and members, who are interested in this direction as well.

subin9:
Hi @Xunzhuo — I'd like to join this WG as a Member.

- Handle: @subin9
- Role: Member
- Merged PR: #3027 — PII entity confidence was a pairwise fold, not the
  mean the comment claimed; 25 of 447 entities crossed the default 0.9
  threshold once fixed.
- Focus: routing inference runtime (#2782 / #2779). I do accelerator deployment and benchmarking day to day, so that's where I'd be most
  useful.
- Availability: 6 hrs/week (KST, UTC+9).

Xunzhuo:
Welcome @subin9

ramkrishs:
Hi @Xunzhuo and @WUKUNTAI-0211  : My Handle @ramkrishs, I would like to join the workgroup as Lead. 

Merged commit/PR: #1865 (feat(api): expose POST /api/v1/nli : Natural Language Inference as a first-class endpoint), #1864 (feat(signals): configurable confidence threshold for language signal), #1848 (feat(signals): Add EventContextSignal for event-driven request routing)

[...]

PR #3096, opened 2026-08-29
[Feature] Add a sequence_classifier backend to the generic classifier signal

guan404ming:
Related #2923

## Purpose

- Add `type: sequence_classifier` to generic classifier signals.
- Reuse the shared `SequenceClassifierBackend`.
- Return the model’s label distribution and reject incomplete responses.
- Validate unsupported types and invalid configurations.
- Defer backend-level `ClassifierSignalRule.OnError` to #3106. Existing condition-level `on_error: no_match|match` behavior remains unchanged.

## Test Plan

- Verify scores are matched by label.
- Verify incomplete responses are rejected.
- Verify invalid configurations are rejected.

## Test Result

- `go test ./pkg/classification ./pkg/config`
- `make agent-validate`
- `make agent-lint`


FAUST-BENCHOU (commented):
Could we address the ClassifierSignalRule.OnError scope added in #2923 here, or explicitly defer it? The new remote sequence_classifier can fail at request time, but the rule still has no backend-level allow/block policy.

guan404ming:
Thanks for flagging this. I would choose to defer `ClassifierSignalRule.OnError` to a follow-up. This PR will stay scoped to wiring `sequence_classifier`, since backend-level `allow|block` needs separate treatment from the existing condition-level `no_match|match`. I’ll update the PR description and track it separately in another [issue](https://github.com/vllm-project/semantic-router/issues/3106).

theohsiung:
Hi @guan404ming thanks for the effort on this pr. Left a comment on [3106](https://github.com/vllm-project/semantic-router/issues/3106) PTAL thx.

theohsiung:
Nice work! One small thing: this adds a user-facing type value but no doc change website/docs/tutorials/signal/learned/classifier.md still describes only local and llm.

BTW If you're interested in the wider design: there's an interface discussion on [#2760](https://github.com/vllm-project/semantic-router/issues/2760) and the on_error half on #3106, both feeding into RFC [#2779.](https://github.com/vllm-project/semantic-router/pull/2779) The WG charter is [#2966](https://github.com/vllm-project/semantic-router/issues/2966) and takes member self-nominations. Would be good to have you there, since you're already working in this area.

guan404ming:
Thank you for catching this and for sharing the related discussions. I’ve added a minimal documentation update covering sequence_classifier, including its `http_classify` contract and configuration constraints.

I also appreciate the invitation. I’ll read through the linked issues and take a closer look at the WG charter.

src/semantic-router/pkg/config/validator_classifier_signal.go
```diff
@@ -203,6 +210,58 @@ func validateLLMClassifierSignal(cfg *RouterConfig, rule ClassifierSignalRule) e
...
+	}
+	if external.ModelRole != ModelRoleClassification {
+		return fmt.Errorf(
+			"routing.signals.classifiers[%q]: external model %q must use model_role %q",
+			rule.Name,
+			rule.Model,
+			ModelRoleClassification,
+		)
+	}
+	if strings.TrimSpace(external.ModelEndpoint.Address) == "" ||
```
theohsiung: This duplicates `validateLLMClassifierExternalDependency` (285-292) but drops its protocol check, so the two paths disagree:

```
llm                  protocol=grpc  -> rejected: ...protocol must be http or https
sequence_classifier  protocol=grpc  -> ACCEPTED
```

`NewHTTPClassifierInference` uses the value as the URL scheme, so this surfaces as `unsupported protocol scheme "grpc"` on every request rather than at startup. Calling the shared helper here would fix this and the duplication above it in one go.

src/semantic-router/pkg/classification/generic_classifier_sequence.go
```diff
@@ -0,0 +1,92 @@
...
+// model actually reported.
+func (c *sequenceLabelClassifier) Classify(
+	ctx context.Context,
+	input string,
+) (labelClassification, error) {
+	result, err := c.backend.Classify(ctx, input)
+	if err != nil {
+		return labelClassification{}, err
+	}
+	if len(result.Probabilities) != len(c.labels) {
```
theohsiung: This length check can't fire through `newSequenceLabelClassifier`: `alignScoresToMapping` allocates exactly `mapping.LabelCount()` entries and errors out unless every declared label is present, and the mapping is built from the same `rule.Labels`.

Coverage over the PR's own tests shows `hits=0` for this block. `TestSequenceLabelClassifierRejectsIncompleteDistribution` passes on the backend's "missing label" error and never reaches here, so the branch reads as covered when it isn't — worth either dropping it or testing it through a substituted backend.

src/semantic-router/pkg/classification/generic_classifier_sequence.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (m *declaredLabelMapping) LabelCount() int {
+	return len(m.labels)
+}
+
+// sequenceLabelClassifier lets a rule reuse the backends jailbreak and
+// category already use, instead of a parallel remote path.
+type sequenceLabelClassifier struct {
+	backend SequenceClassifierBackend
```
theohsiung: `backend` is the general interface, but `Classify` maps `result.Probabilities[i]` to `c.labels[i]` — position in the wrapper's own copy. That holds only because this constructor happens to build the backend from the same `rule.Labels`.

Substituting any backend that carries a model-shipped mapping (candle, mmBERT) compiles cleanly and silently misattributes scores. With a backend whose mapping is `[toxic, benign]`, toxic input comes back as `benign:0.97 toxic:0.03` with no error anywhere.

Holding `*HTTPClassifierInference`, or reading labels back through the mapping this wrapper already owns, would make the coupling explicit.

theohsiung (changes requested):
Nice piece of work: reusing `SequenceClassifierBackend` instead of adding a parallel remote path is the right call, and the `declaredLabelMapping` comment explaining why a generic signal owns its own ordering is genuinely helpful.

Five things I think need to land in this PR, then a batch of non-blocking notes.

## 1. The Python CLI rejects the new type

`src/vllm-sr/cli/models.py:490` still pins `type: Literal["local", "llm"]`, so a config using the new type fails before it reaches the router. Through the real `parse_user_config` path:

```
routing -> signals -> classifiers -> 0 -> type: Input should be 'local' or 'llm'
```

And after patching only that Literal, the cross-object validator still skips the new type — `src/vllm-sr/cli/validator_classifier.py:78` does `if rule.type != "llm": continue`. Pointing a `sequence_classifier` rule at a **nonexistent** external model:

```
sequence_classifier -> validate_user_config -> 0 error(s)   <- passes
llm (same config)   -> validate_user_config -> 1 error(s)   <- "references unknown external model"
```

`tools/agent/skills/signal-end-to-end/SKILL.md` Acceptance: *"Router config, signal extraction, and Python CLI schema stay aligned for the signal contract."*

## 2. The dashboard can't edit a `sequence_classifier` signal

`dashboard/frontend/src/types/config.ts:185` is still `'local' | 'llm'` (confirmed with `npm run type-check`: `Type '"sequence_classifier"' is not assignable to type '"local" | "llm"'`), the backend select at `configPageSignalFormFields.tsx:144` offers only those two, and every dependent field's `shouldHide` keys on them.

Transcribing the save path (`ConfigPageSignalsSection.tsx:1233-1256`) and the `shouldHide` predicates verbatim and running them:

```
type=sequence_classifier:
  select has matching option : false
  classifier_model           : HIDDEN
  classifier_model_path      : HIDDEN
  classifier_instructions    : HIDDEN
  SAVE -> BLOCKED: External classifier model is required.
```

So an existing rule loads into a form with no way to satisfy the save path. It's a small change — union type, select option, three `shouldHide` predicates, and the `else` branch at 1243.

## 3. `llm_endpoint.protocol` is unvalidated on the new path

`validateSequenceClassifierSignal` re-implements the address/port check but not the protocol check the `llm` path gets from `validateLLMClassifierExternalDependency` (line 285-292). Both validators, same input:

```
llm                  protocol=grpc  -> rejected: ...protocol must be http or https
sequence_classifier  protocol=grpc  -> ACCEPTED
sequence_classifier  protocol=file  -> ACCEPTED
```

`NewHTTPClassifierInference` splices it straight into the scheme, so it becomes a per-request runtime failure instead of a startup config error:

```
baseURL = grpc://clf.internal:8080
-> http_classify request failed: Post "grpc://clf.internal:8080/classify": unsupported protocol scheme "grpc"
```

Reusing the shared helper rather than adding a fourth copy of the endpoint checks would fix this and the duplication at 216-263 together.

## 4. Reference config and fragment tree not updated

`config/config.yaml:531,537` still shows only `local` and `llm`, and `config/fragments/signal/classifier/` still contains only `label-score.yaml` (an `llm` example) — which `classifier.md:87` links as *"a complete example"*. There's no runnable YAML for the new type anywhere in the repo.

`src/semantic-router/pkg/config/AGENTS.md:39`: *"If you change supported signals … update the router-owned surface catalog and the repo-owned `config/` fragment tree in the same change."*

Worth noting why CI stayed green: neither `fragment_catalog_test.go` nor `reference_config_contract_test.go` enumerates classifier types, so nothing fails when a type is missing. A small assertion there would prevent the next one slipping too.

## 5. No E2E coverage

`grep -rl sequence_classifier e2e/` returns nothing. This adds a config type that opens an outbound HTTP call on the request path. Same skill's Acceptance and its gotcha: *"New signal semantics need explicit E2E ownership, not just unit-level classifier coverage."*

## Non-blocking

(two more are inline on the diff)

- **The response contract is stricter than the docs say.** `alignScoresToMapping` requires the response labels to equal the declared set exactly *and* the scores to sum to ~1.0, so only softmax single-label heads work. A sigmoid multi-label head is silently unsupported: a 6-head toxicity model returns `http_classify response scores sum to 3.37, want ~1.0`, and declaring a subset gives `response label "insult" is not in the configured label mapping`. The tutorial only says "require at least two labels" — worth one sentence, since this is the constraint users will hit first.
- **`classifier.md:66`** — the new paragraph sits between the two `on_error` paragraphs, so the following *"This condition-level `on_error`"* loses its antecedent. After the `prompt_guard.on_error` contrast at line 74 would read better.
- **`classifier.md:83`** — the data-residency paragraph still names only `local` and `llm`; `sequence_classifier` also POSTs request text off-box.
- **Three bare `"local"` literals** remain after the good `ClassifierSignalType*` change: `validator_classifier_signal.go:362,373` (already in this PR's diff) and `validator_decision.go:148`. Either all three or none — the last one widens the lint surface to a file this PR doesn't otherwise touch, so leaving it for a separate chore is fine.

## Scope note

`Closes #2923` — but #2923's first gap (`type: llm` fakes a distribution: chosen label 1, others 0) is untouched here, which seems right for a separate change. Consider `Refs #2923` so the issue stays open for it, rather than filing a new one.

src/semantic-router/pkg/config/validator_classifier_signal.go
guan404ming: Thanks for catching this. I extracted a shared endpoint validator, so both llm and sequence_classifier now validate address, port, and protocol consistently. I also added an invalid-protocol regression test.

src/semantic-router/pkg/classification/generic_classifier_sequence.go
guan404ming: Agreed. This check was unreachable because the backend already validates the exact label distribution, so I removed it.

src/semantic-router/pkg/classification/generic_classifier_sequence.go
guan404ming: Just changed the backend field to *HTTPClassifierInference, making the label-order coupling explicit and preventing unrelated mapped backends from being substituted.

guan404ming:
Thanks for the detailed review! I've updated the pr and here are simple summarization:

- Aligned the Python CLI validation with `sequence_classifier`.
- Added Dashboard edit and save support.
- Reused the shared external endpoint validation.
- Added reference config, fragment, and drift tests.
- Added real outbound E2E coverage.
- Addressed the documentation and cleanup notes.

Please take another look, thanks again!

theohsiung (dismissed):
LGTM, thanks for the thorough turnaround!

One follow-up before merge: `dashboard/frontend/src/lib/dslPolicySignalSchemas.ts:29` — the Builder has a second classifier schema still on `['local', 'llm']`, so the new type can't be authored there. Same one-line fix as `configPageSignalFormFields.tsx`.

guan404ming:
Thanks for catching this! I added sequence_classifier to the DSL Builder schema with a regression test.

theohsiung (approved):

WUKUNTAI-0211 (approved):

guan404ming:
Thanks for your detailed review.

Issue #3037, opened 2026-08-26
[Epic] Advance bounded multi-model collaboration algorithms


Xunzhuo:
## Parent Direction

MoM & Routing — #2965

## Goal

Make bounded multi-model collaboration a measurable routing capability that can improve the quality-cost frontier beyond single-model selection while keeping latency, spend, amplification, and failure behavior under explicit control.

## Background

The repository already supports several collaboration families, including Looper, Flow, Fusion, ReMoM, confidence escalation, and ratings. They currently expose different execution, accounting, quorum, verifier, and fallback semantics. This Epic aligns those foundations so existing and future algorithms can be evaluated and operated consistently.

## Scope

- Define shared contracts for candidate generation, fallback, cascade, parallel panels, judging or verification, synthesis, and bounded iteration.
- Standardize usable-response checks, aggregate budgets, cancellation, quorum, fallback, per-attempt traces, replay, and protocol behavior.
- Harden and optimize existing Looper, Flow, Fusion, and ReMoM paths before recommending new algorithms for production.
- Support adaptive breadth, early stopping, confidence escalation, repeated sampling, verifier-guided pruning, and budget-aware execution.
- Evaluate learned or rule-based policies for assigning collaboration strategies and compute budgets.
- Define evidence and graduation gates for moving an experimental algorithm into a maintained recipe capability.

### Shared foundations and hardening

- #2336 — harden existing algorithm families.
- #2855 — add common per-attempt traces and metrics.
- #2856 — unify candidate-panel execution semantics.
- #2857 — define a pluggable outcome-verifier contract.
- #1456 — cap fan-out and total upstream-call amplification.
- #1815 — add per-step controls and accounting to static Router Flow workflows.

### Algorithms, evaluation, and failure policy

- #2858 — compare collaboration algorithms under matched quality-cost budgets.
- #2860 — evaluate Grounding-aware Fusion.
- #2862 — add Best-of-N selection with outcome reranking.
- #2863 — add adaptive breadth and staged early stopping.
- #2864 — harden request-level confidence escalation.
- #2865 — make Fusion adaptive and support repeated self-sampling.
- #2866 — add verifier-guided ReMoM pruning and adaptive breadth.
- #2337 — explore additional bounded collaboration algorithms.
- #3374 — enforce Fusion quorum over usable responses.
- #3375 — define explicit Fusion quorum-failure fallback policy.

## Interfaces

[...]

PR #3101, opened 2026-08-29
[Feature] Harden built-in MoM routing boundaries

Xunzhuo:
Related #3037

## Purpose

Harden the maintained built-in Mixture-of-Models path at the boundaries where request intent, candidate eligibility, session continuity, and multi-model execution meet.

- Derive tool execution state from explicit protocol choices, textual execution intent, and the active conversation tail. Declared tools or old tool history alone no longer make agentic routing sticky.
- Enforce known context-window eligibility before direct selection, Looper execution, evaluation, session continuity, and Router Learning.
- Add `algorithm.minimum_candidates` across Go config, DSL, CLI validation, Dashboard materialization, and runtime filtering. Every built-in decision now declares its assignment cardinality contract.
- Reject impossible Fusion and Workflow quorums instead of silently lowering them at runtime.
- Re-check generated Looper prompts and completion reserve against each target model before every backend dispatch.
- Keep session and learning overrides inside the active decision's eligible candidate set, reset stale continuity after decision drift, and retain explicit opt-in for learning.
- Tighten the Balance, Speed, Cost, Accuracy, and Vault signal/projection/decision boundaries, including a terminal long-context lane, privacy retention controls, and deterministic model-pool assignments.
- Keep canonical recipes, packaged CLI assets, Dashboard topology, documentation, and conformance probes synchronized.

## Before / After

| Contract | Before | After |
| --- | --- | --- |
| Built-in candidate cardinality | 0/26 decisions declared a minimum | 27/27 decisions declare and enforce it |
| Fusion quorum above a two-model panel | silently reduced to 2 | rejected as an invalid quorum of 3 |
| Workflow quorum above two calls | silently reduced to 2 | preserved and rejected before execution |
| Generated Looper context growth | checked only at initial routing | checked per generated stage and target model |
| Built-in static corpus | 26 decisions / 226 variants | 27 decisions / 235 variants |
| Tool declarations on ordinary turns | could keep selecting the agentic lane | remain neutral unless execution is required or the tail is active |
| Session / learning continuity | could restore a model outside the new decision | bounded by the current eligible candidate set |

## Validation

- Signed commit hooks: all passed, including Go format/lint, Python, JS/TS, Markdown, YAML, supply-chain scan, and changed-file structural lint.
- `make agent-lint`, `make agent-ci-gate`, and `make agent-pr-gate`: passed on the final source snapshot.
- Semantic Router Go suites, Python CLI contracts, Dashboard frontend/backend tests, Rust/binding checks, Helm validation, and catalog synchronization: passed.
- CLI integration: 47 passed, 1 existing TODO skipped, 0 failures.
- Maintained live recipe conformance: 287/287 variants across all seven maintained recipes.
- Built-in static conformance: 235/235 variants, 27 decisions, with complete signal, projection, algorithm, and plugin coverage.
- Final changed-source receipt is identical across the worktree and two isolated AMD validation environments; focused post-commit Router packages pass in both.
- Runtime health, ROCm device access, metrics, dashboard, and simulator checks passed in both AMD environments. Task-specific containers, volumes, ports, and temporary workspaces were cleaned afterward while shared caches and unrelated services were preserved.

All applicable GitHub checks pass on the final head `988e6492`.


FAUST-BENCHOU (commented):
Could we apply the context-window check to the actual Looper stage request rather than only the original request? Workflow planner/step/final calls append prompts and prior outputs, so a model can pass this check and still exceed its context window during execution.

Xunzhuo:
@FAUST-BENCHOU Good catch. Addressed in 25e86128: every Looper backend dispatch now checks the generated stage request against the target model context window, including appended or replacement messages and any larger output-token reserve. The shared gate covers Fusion, Confidence, Ratings, ReMoM, and Workflows, and returns StageContextWindowError before the backend call. The stage-level cases are covered in stage_context_test.go; the current head is green. Could you take another look?

FAUST-BENCHOU (approved):

PR #3076, opened 2026-08-27
[Feature] Unify inference protocol translation

Xunzhuo:
Related #2947
Related #3063

## Purpose

Replace protocol-specific request and response paths with one protocol-neutral Router boundary and a registry-driven codec matrix.

This change:
- introduces a bounded neutral request, response, error, usage, and streaming model;
- implements pairwise translation for OpenAI Chat Completions, OpenAI Responses, and Anthropic Messages;
- makes unsupported fields and representation loss explicit instead of silently dropping data;
- consolidates ExtProc request dispatch, buffered responses, streaming, immediate responses, cache identity, replay, and usage settlement on that boundary;
- validates model pricing at `providers.models[].pricing` while preserving the existing configuration shape;
- removes superseded provider-specific translation paths rather than retaining compatibility glue.

## Contract coverage

- complete 3×3 client/backend matrix for buffered and streaming traffic;
- human-readable input/output goldens for every matrix edge;
- tool-call and tool-result lifecycle, structured output, multimodal content, reasoning, refusal, citations, metadata, and authoritative usage provenance;
- provider errors, midstream errors, incomplete streams, arbitrary transport chunking, and terminal behavior;
- strict unknown-field handling and typed unsupported-feature errors;
- official OpenAI and Anthropic field inventories pinned to reviewed schema revisions;
- revision-pinned OpenAI-compatible and Anthropic provider simulators with provider-native validation and error envelopes;
- direct E2E assertions on the native request received by every backend simulator;
- six fuzz targets spanning request, response, and streaming translation.

## Validation

- `make agent-ci-gate AGENT_BASE_REF=origin/main`: passed;
- OpenAI-compatible simulator contract tests: 6 passed;
- Anthropic simulator contract and proxy tests: 49 passed;
- deployed 3 clients × 3 backends × buffered/streaming matrix: 18 cells passed;
- protocol, ExtProc, and E2E Go packages: passed;
- protocol and ExtProc race tests: passed;
- all six codec fuzz targets: passed;
- deployed baseline gateway, streaming, Responses API, and Anthropic profiles: passed.

PR #3104, opened 2026-08-29
[Bug] Escapes non-JSON tool arguments so Anthropic outbound emit stays valid

### LINKED ISSUE #3103: [Bug] Anthropic outbound emit fails on non-JSON tool arguments, turning a successful response into api_error
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

When `EmitAnthropicResponse` (OpenAI → Anthropic response translation) processes a non-streaming response whose tool call carries `function.arguments` that is **not valid JSON**, the whole response marshal fails and the error branch replaces the body with an `api_error` envelope.

Root cause: `pkg/anthropic/outbound.go` `buildAnthropicContent` inserts `tc.Function.Arguments` directly as `json.RawMessage`. `json.Marshal` then fails with e.g. `json: error calling MarshalJSON for type json.RawMessage: unexpected end of JSON input`. `processor_res_body_pipeline.go` turns that failure into `EmitAnthropicError("api_error", ...)`, so a turn that upstream completed successfully is delivered to the Anthropic client as a failure with no configuration workaround.

The streaming path (`sse_out.go`) accumulates `input_json_delta` fragments and is unaffected — the bug is asymmetric to non-streaming.

### To Reproduce

1. Route an OpenAI-format backend (e.g. a non-json-mode or max_tokens-truncated response) to an Anthropic protocol client.
2. Have the backend return a tool call whose `function.arguments` is truncated/partial, e.g. `` `{"location": "SF", "unit": ` ``.
3. `EmitAnthropicResponse` returns an error; the Anthropic client receives `api_error`.

Minimal unit-level repro (fails on current main):

```go
out, err := EmitAnthropicResponse(bodyWithBadToolArgs, nil, "model")
// err = anthropic outbound: marshal response: json: ... unexpected end of JSON input
// out = ""
```

### Expected behavior

A single malformed tool-arguments fragment must not fail the entire response. The tool call should still be delivered (e.g. arguments escaped into a valid JSON string) with a lossy translation warning, so an otherwise successful turn reaches the client.

### Impact

Any Anthropic-protocol client in a mixed-format routing setup can receive `api_error` for a response that upstream completed, with no configuration workaround. Triggered by truncation or non-JSON-mode backends.

### Additional context

I have a candidate fix (escape invalid tool arguments into a JSON string plus a lossy warning, preserving current behavior for valid/empty input) and will open the linked PR.

Bevisy:
### Summary

Fixes #3103

When an OpenAI-format backend returns a tool call whose `function.arguments` is not valid JSON (max_tokens truncation, non-json-mode model, partial fragment), the Anthropic outbound emitter inserted the raw bytes as `json.RawMessage` and the whole response `json.Marshal` failed. `translateResponseBodyForClient` then returned an `api_error` to the Anthropic client — a successful turn delivered as a failure, with no configuration workaround.

### Change

Introduce `toolUseArguments` in `pkg/anthropic/outbound.go`:

- empty arguments → `{}` (unchanged)
- valid JSON → pass through unchanged (unchanged)
- invalid JSON → escaped into a JSON string + a `lossy` translation warning

The streaming path (`sse_out.go`) already tolerates fragments via `input_json_delta` and is untouched.

### Test

`TestEmitAnthropicResponse_InvalidToolArgumentsStayValid` — regression covering the truncated-arguments case (previously failed with `unexpected end of JSON input`).

### Validation

- `go vet ./pkg/anthropic/` — ok
- `go build ./pkg/anthropic/... ./pkg/extproc/...` — ok
- `go test ./pkg/anthropic/... ./pkg/ir/...` — pass

### DIFF
--- a/src/semantic-router/pkg/anthropic/outbound.go
+++ b/src/semantic-router/pkg/anthropic/outbound.go
@@ -246,10 +246,7 @@ func buildAnthropicContent(choice openai.ChatCompletionChoice, ext *ir.IRExtensi
 
 	// Tool calls.
 	for _, tc := range choice.Message.ToolCalls {
-		input := json.RawMessage(tc.Function.Arguments)
-		if len(input) == 0 {
-			input = json.RawMessage("{}")
-		}
+		input := toolUseArguments(tc.Function.Arguments, ext)
 		blocks = append(blocks, anthropicContentBlock{
 			Type:  "tool_use",
 			ID:    tc.ID,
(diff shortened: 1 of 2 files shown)


FAUST-BENCHOU (commented):
Could we keep `tool_use.input` as a JSON object here? Anthropic's tool-use contract defines `input` as an object conforming to `input_schema`, so escaping malformed arguments into a JSON string makes the response marshalable but still violates the expected wire shape: https://platform.claude.com/docs/en/agents-and-tools/tool-use/handle-tool-calls

Bevisy:
Thanks @FAUST-BENCHOU — good catch, the bare string would still violate the wire shape. I updated `toolUseArguments` so invalid arguments are wrapped into an object under a reserved `raw_arguments` key. `input` stays a JSON object conforming to the tool-use contract, the response no longer fails to marshal, and the lossy warning is still recorded for observability.

Bevisy:
Superseded by #3076 (protocolcodec refactor adds json.Valid guard). See #3103 comment.

Issue #3097, opened 2026-08-29
[Bug] Request-path connectors read upstream responses without a byte ceiling


guan404ming:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Several remaining request-path connectors read upstream responses without a byte ceiling. `pkg/looper`, the hallucination endpoint, the embedding and RAG error paths and `pkg/vectorstore` already bound theirs. These do not:

- [x] Router-R1 and AutoMix #3118
- [x] External classifier and MCP response limits #3105
- [x] Memory rewrite response limit #3131
- [x] OpenAI-compatible embedding response limit #3133
- [x] External RAG response limit #3134
- [x] OpenAI and vLLM-Omni image generation response limits #3135
- [x] OpenAI vector store RAG #3151

`json.NewDecoder(resp.Body)` is not a ceiling either: `router_r1_client.go:85,192`, `embedding/openai_provider.go:213` and `req_filter_rag_external.go:197` bound their error bodies but decode success bodies unbounded.

`vllm_client.go` also checks the status after the read, then puts the whole body in the error:

```go
body, err := io.ReadAll(resp.Body)          // unbounded
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    return nil, fmt.Errorf("... status %d: %s", resp.StatusCode, string(body))
}
```

### To Reproduce

Point any site at an endpoint returning a large body; the router reads all of it. `vllm_client.go` is the sharpest: three request-path construction sites, so the cost is paid per classified request.

### Expected behavior

Each connector reads at most a configured number of bytes and fails clearly above it; a non-2xx body is truncated, not echoed in full. Defaults should differ by family, since classifier responses are small but MCP tool output and base64 images are not: a shared helper, not a shared limit.

### Impact

Memory pressure from upstreams the router does not control, per request, with no configuration workaround. Classifier failures propagate through `on_error`; with `block` that fails closed for every affected request. Severe pressure may terminate the process before error handling runs. The other sites have no equivalent policy.

### Additional context

Sites span classification, extproc, imagegen and MCP ownership. Suggest keeping this as the tracking issue and splitting into at least "classifier connectors" and "plugin/tool connectors".

#2924 covers `http_classify`. The open proposal #2779 (unmerged, needs rebase) recommends bounded connector I/O and a reusable guard; `pkg/looper` and `pkg/vectorstore` already carry two separate copies of that helper.

[...]

PR #3105, opened 2026-08-29
[Bug] Bound classifier connector responses

guan404ming:
Related #3097

## Purpose

- Bound external LLM classifier and HTTP MCP response bodies.
- Truncate non-2xx error bodies while preserving status.
- Reuse the shared bounded-response reader.
- Default to 1 MiB for external classifiers and 16 MiB for MCP.

## Test Plan

- Assert oversized responses are rejected.
- Assert oversized error bodies are truncated.
- Run focused Go tests and the repository CI gate.

## Test Result

- `go test ./pkg/classification ./pkg/mcp ./pkg/config`
- `make agent-ci-gate CHANGED_FILES="..."`


src/semantic-router/pkg/classification/vllm_client.go
```diff
@@ -14,18 +13,21 @@ import (
...
+	httpClient       *http.Client
+	endpoint         *config.ClassifierVLLMEndpoint
+	baseURL          string
+	accessKey        string // Optional access key for Authorization header
+	maxResponseBytes int64
 }
 
 // NewVLLMClient creates a new vLLM REST API client for classifiers
-func NewVLLMClient(endpoint *config.ClassifierVLLMEndpoint) *VLLMClient {
+func NewVLLMClient(cfg *config.ExternalModelConfig) *VLLMClient {
```
FAUST-BENCHOU: Could we preserve the existing exported constructors or add a separate config-based constructor here? Changing `NewVLLMClient`'s signature and removing `NewVLLMClientWithAuth` makes this bug fix source-incompatible for downstream users of `pkg/classification`.

config/config.yaml
```diff
@@ -2183,6 +2183,7 @@ global:
         access_key: external-guardrail-key
         max_tokens: 2048
         temperature: 0.2
+        max_response_bytes: 1048576
```
FAUST-BENCHOU: Could we omit these explicit `max_response_bytes` values from the reference config and rely on the documented defaults? Since they exactly match the defaults, including them here makes an implementation-level safety limit look like required configuration.

FAUST-BENCHOU (commented):
Two small concerns around compatibility and config surface.

src/semantic-router/pkg/classification/vllm_client.go
guan404ming: Good catch, thanks. I restored both exported constructors with their original signatures. Internal callers now use an unexported config-based constructor, so configured response limits still apply without breaking downstream users.

config/config.yaml
guan404ming: Make sense to me. Removed the explicit defaults from the external model entries. I kept the MCP field because the canonical reference-config contract requires every public MCP key. User configs can still omit it and use the 16 MiB default.

guan404ming:
Hi @AayushSaini101 could you also help take a look at this one? I think this is the last one in this series. Thanks!

src/semantic-router/pkg/classification/generic_classifier.go
```diff
@@ -34,10 +34,7 @@ func newLLMLabelClassifier(
 	if external == nil {
 		return nil, fmt.Errorf("external model %q is not configured", rule.Model)
 	}
-	client := NewVLLMClient(&external.ModelEndpoint)
-	if external.AccessKey != "" {
-		client = NewVLLMClientWithAuth(&external.ModelEndpoint, external.AccessKey)
-	}
```
AayushSaini101, line 43: @guan404ming Why are we removing the AccessKey check and passing the AccessKey in the method? i checked below code "Optional access key for Authorization header" access key is optional is this why we are removing that thanks : )

guan404ming: Thanks for checking! The AccessKey check is now handled inside the client. An empty key still sends no Authorization header; this change just centralizes client configuration and also passes max_response_bytes.

Xunzhuo (dismissed):
Looks good now. The public VLLM client constructors keep their existing behavior, configured classifier clients pick up the 1 MiB ceiling, MCP gets its own 16 MiB default, and non-2xx diagnostics are bounded separately. Tests and current CI cover both one-byte-over-limit and truncated errors. No blocker from me.

guan404ming:
Just pushed a commit to resolve the conflict.

Xunzhuo (approved):
Looks good. The external LLM classifier and HTTP MCP paths now use the shared bounded reader for successful responses and bounded truncation for error bodies, with the configured limits reaching the actual clients. The one-byte-over and error cases are covered. No code blocker from me; merge after the current CI run is green.

wilsonwu (approved):

WUKUNTAI-0211 (approved):

theohsiung (approved):

PR #3108, opened 2026-08-29
[Bug] Count Response API image inputs before translation so file_id images reach the conversation image signal

### LINKED ISSUE #3107: [Bug] Resolve Router-owned Responses file_id images before cross-format dispatch
Parent Epic: #2358

### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Current main already preserves /v1/responses input_image.file_id in the neutral protocol and counts it for image routing signals. That original signal-loss path was fixed by #3076.

The remaining bug is end-to-end dispatch for Router-owned files. A Responses request can reference an image uploaded through the Router /v1/files API. Routing correctly sees the image and may select an image-capable model, but a selected Chat Completions backend cannot consume the unresolved file_id, so cross-format encoding fails before backend invocation.

### To Reproduce

1. Upload an image through the Router /v1/files API.
2. Send /v1/responses input containing {"type":"input_image","file_id":"file-..."}.
3. Route it to a backend that uses the Chat Completions request contract.
4. Confirm the image signal matches, but the request cannot be encoded for the selected backend.

### Expected behavior

Before cross-format dispatch, resolve Router-owned image file IDs through the Router file store, validate the media type and configured byte limits, and inline the image as a bounded data URL.

- Preserve the existing image-signal count.
- Return a clear 4xx error for unknown, expired, oversized, or non-image Router file IDs.
- Keep upstream/provider-owned file IDs out of scope unless a provider-specific contract defines them.
- Preserve file_id on compatible same-format paths where it can be forwarded safely.
- Do not add input_image.file_data; the Responses input_image variant supports image_url and file_id.

### Impact

Responses clients can route an uploaded image correctly but still fail before reaching a cross-format backend. Converting every upload into a client-hosted URL is not a reasonable workaround.

### Acceptance criteria

- A Router-owned input_image.file_id reaches a Chat Completions backend as an equivalent bounded image URL.
- Same-format behavior is preserved.
- Invalid or unresolvable IDs fail explicitly without leaking file contents.
- File reads and generated data URLs obey configured byte and media-type limits.
- Unit and affected E2E coverage prove routing, dispatch, and error behavior.

### Additional context

The original issue was opened before #3076 landed. Related implementation work is in #3108.

yaojiejia:
Closes #3107

## Purpose

`/v1/responses` requests are translated to Chat Completions before routing signals are extracted. The translator only keeps `input_image` parts that carry `image_url`, so an image sent by `file_id` (or `file_data`) is gone before the `conversation` image signal counts it, and image-routing decisions never fire for those requests.

This change counts image parts from the native Response API input before translation, stores the count on the request context, and merges it into the conversation facts (`max` with the translated count, so URL images are not double counted). Only user input items are counted, so images in assistant/system history do not trigger image routing. The upstream request body is unchanged.

Modules: `pkg/responseapi`, `pkg/extproc`. Workgroup: Data Plane & Networking.

## Test Plan

```
cd src/semantic-router
go test ./pkg/responseapi/ ./pkg/extproc/
```

Unit coverage added: `pkg/responseapi/input_image_facts_test.go` (URL / `file_id` / `file_data` / mixed / non-user items) and `pkg/extproc/req_filter_response_api_image_facts_test.go` (`file_id` restores the image count, URL images are not double counted).

Live check on a kind cluster (response-api profile) with a `conversation` `image_content` signal and a `vision_decision` above `default_decision`, sending the same `/v1/responses` request with `image_url` and with `file_id`, reading `x-vsr-selected-decision` and `x-vsr-matched-conversation`.

## Test Result

```
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/responseapi
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/extproc
```

Live check, before (`main`): `image_url` -> `vision_decision`, `file_id` -> `default_decision`, no matched signal.
Live check, after (this branch): `image_url` and `file_id` -> `vision_decision`, `x-vsr-matched-conversation: image_input`; text-only request still -> `default_decision`.

Remaining gap: `input_audio` parts are dropped by the same translation, so audio-based signals on the Response API path are not addressed here.

---
<details>

</details>


FAUST-BENCHOU (commented):
Could we ensure the `file_id`/`file_data` image itself is preserved for the selected backend? This restores the image routing signal, but the Response API translation still drops those image parts, so the vision model may be selected without actually receiving the image.

yaojiejia:
> Could we ensure the `file_id`/`file_data` image itself is preserved for the selected backend? This restores the image routing signal, but the Response API translation still drops those image parts, so the vision model may be selected without actually receiving the image.

Makes sense. `file_data` is now inlined as an image_url data URL. `file_id` is read from the router's own file store (/v1/files, images accepted with purpose=vision) and inlined the same way, so the backend gets the image. An unresolvable `file_id` returns a 400 instead of dropping the image. File IDs that only exist on an upstream provider are out of scope, since the router can't read them and vLLM has no Files API.

AayushSaini101:
@yaojiejia thanks for working on this, can you please resolve the conflicts, QQ: 
* When we trying to upload the files from the dashboard what will happen in that case, does image_url consider or image_file is consider during that process, 
* I guess we are already setting the limit oni the uploading the files, if yes can we show the message during the upload time from the dashboard thanks

yaojiejia:
> @yaojiejia thanks for working on this, can you please resolve the conflicts, QQ:
> 
> * When we trying to upload the files from the dashboard what will happen in that case, does image_url consider or image_file is consider during that process,
> * I guess we are already setting the limit oni the uploading the files, if yes can we show the message during the upload time from the dashboard thanks

Just resolved the conflict, to answer your questions:

1. The dashboard never uses file_id. It has no UI that calls POST /v1/files. The playground reads the image in the browser, base64-encodes it, and sends it inline as an image_url data URL. That path already routed correctly before this PR. The file_id path is for API clients that upload with POST /v1/files and purpose=vision, then reference the id from a Responses input_image part, the router inlines the stored bytes before dispatch.

2. Yes, limits exist and both show a message at upload time. The dashboard playground checks in the browser before sending: 4 MB per image, 10 MB per attachment, 8 MB of encoded images per request, with messages like "cat.png is 6.2 MB. Each attachment must be 4 MB or smaller." POST /v1/files caps uploads at 50 MB (413 "request body too large") and rejects wrong types (400 INVALID_FILE_TYPE). Since the dashboard has no upload screen for /v1/files, there is no dashboard message to add there.

yaojiejia:
> @yaojiejia thanks for working on this, can you please resolve the conflicts, QQ:
> 
> * When we trying to upload the files from the dashboard what will happen in that case, does image_url consider or image_file is consider during that process,
> * I guess we are already setting the limit oni the uploading the files, if yes can we show the message during the upload time from the dashboard thanks

Oh btw is it possible if you can accept this related issue https://github.com/vllm-project/semantic-router/issues/3107 thanks!

Xunzhuo (changes requested):
The latest code fixes the important second half: the selected backend now receives the image instead of only seeing an image routing fact. One blocker remains: this changes both `/v1/files` and the Responses dispatch path, but the new flow only has unit tests. Please add a `response-api` E2E that uploads a vision image, sends its `file_id` in `input_image`, and verifies both the image decision and the image received by the selected backend. That is the contract this bug needs to pin.

yaojiejia:
> The latest code fixes the important second half: the selected backend now receives the image instead of only seeing an image routing fact. One blocker remains: this changes both `/v1/files` and the Responses dispatch path, but the new flow only has unit tests. Please add a `response-api` E2E that uploads a vision image, sends its `file_id` in `input_image`, and verifies both the image decision and the image received by the selected backend. That is the contract this bug needs to pin.

Got it, just added `response-api-image-file-id` to the response-api E2E profile. It uploads a PNG via POST /v1/files (purpose=vision), sends the file_id in an input_image part to POST /v1/responses, and asserts the image decision is selected and the backend's recorded request contains the inlined image bytes with the file_id gone. It also covers the text-only default decision and the 400 for an unknown file_id.

Xunzhuo (dismissed):
Thanks for closing the E2E gap. The current response-api case now uploads the image, verifies the image decision, confirms the selected backend receives the inlined bytes, and covers an unknown file ID.

yaojiejia:
> Thanks for closing the E2E gap. The current response-api case now uploads the image, verifies the image decision, confirms the selected backend receives the inlined bytes, and covers an unknown file ID.

Thxx

wilsonwu (dismissed):

yaojiejia:
seems like there was a merge conflict yesterday while 

> # Merge Queue Status
> * ✅ **Entered queue** — `2026-09-01 05:15 UTC` · Rule: `default` · triggered by @wilsonwu with the [merge queue checkbox](https://github.com/vllm-project/semantic-router/pull/3108#issuecomment-5487782922)
> * 🚫 **Left the queue** — `2026-09-01 08:11 UTC` · at `48c4c5e4ef989a459709b55b357e308e0a827e2d`
> 
> This pull request spent **2 hours 56 minutes 24 seconds** in the queue, with no time running CI.
> 
> ## Reason
> The pull request can't be updated
> 
> > merge conflict between base and head
> 
> ## Hint
> You should update or rebase your pull request manually. If you do, this pull request will automatically be requeued once the queue conditions match again. If you think this was a flaky issue, you can requeue the pull request, without updating it, by posting a `@mergifyio queue` comment.
> 
> Requeued — the merge queue status continues in [this comment ↓](https://github.com/vllm-project/semantic-router/pull/3108#issuecomment-5491424289).

seems like there was a merge conflict yesterday after you guys approve this, can @wilsonwu @Xunzhuo take a look at this again? thanks

Xunzhuo (approved):
Thanks for resolving the merge against the new input-modality contract. I rechecked the combined diff: the file_id image case keeps the vision decision, text control, and backend-inline assertions, and current Response API and AI Gateway CI is green. No blockers from me.

PR #3120, opened 2026-08-30
[Bug] Give the ingestion pipeline per-generation lifecycle ownership

### LINKED ISSUE #2602: [Feature] Add a lifecycle root and bounded idempotent Stop(ctx) to vector storage

Ivelisya:
Closes #2602

## Purpose

Completes the lifecycle slice of #2602, owned by `wg/data-plane-networking`.
Affects `src/semantic-router/pkg/vectorstore` only.

The root context and the bounded `Stop(ctx)` from this issue already landed. What
was still missing is the state that makes a *timed-out* Stop safe:

- **Tri-state lifecycle.** `running bool` becomes `stopped`/`running`/`stopping`.
  A Stop that exhausts its deadline leaves the pipeline in `stopping`, because its
  workers were never reclaimed.
- **Per-generation ownership.** `jobQueue` and `wg` were allocated once in the
  constructor and reused across every Start/Stop. A stop that timed out followed
  by an immediate restart therefore shared them, and an old worker could dequeue a
  new-generation job. Each `Start` now allocates its own queue, WaitGroup and root
  context.
- **Carrying generations forward.** A still-draining generation is retained on
  restart and joined by a later Stop, so shutdown accounts for every worker.
- **Bounded cleanup contexts.** Queue-failure and completed-status count
  persistence now run detached from the job context but bounded by
  `cleanupTimeout`, so durable counts stay coherent with the status just written
  while a stalled metadata registry cannot hold Stop past its deadline.

### One behaviour change worth calling out

A second `Stop` after a timed-out one no longer returns `nil`. It joins the same
generation and reports its own outcome — the issue lists this explicitly ("does
not falsely report `nil`; waits on the same generation").

`pipeline_lifecycle_test.go` asserted the old behaviour:

```go
// A second Stop on an already-stopped pipeline is a no-op returning nil.
Expect(f.pipeline.Stop(context.Background())).To(BeNil())
```

so that assertion is replaced rather than kept. The replacement asserts the second
Stop reports its own deadline, and that Stop reports success only once the wedged
stage unwinds and the generation actually drains.

Note the old assertion also passed an unbounded `context.Background()`; under the
new contract that would wait as long as the wedged stage does, so the replacement
passes a bounded context.

### Validation cases

Of the four cases the issue lists, two were already covered and are unchanged:
the wedged-stage stop deadline (`pipeline_lifecycle_test.go`,
`integration_test.go`) and graceful drain plus goroutine-leak
(`integration_test.go`). This PR changes the second-Stop case and adds the
timeout-then-restart case, which had no coverage.

## Test Plan

- `go test -race ./pkg/vectorstore ./pkg/routerruntime` (the command in the issue)
- `make agent-ci-lint AGENT_BASE_REF=origin/main`
- `go vet ./pkg/vectorstore`, `gofmt -l pkg/`, `go build ./pkg/...`

## Test Result

`go test -race ./pkg/vectorstore ./pkg/routerruntime -count=1` passes:

```
ok  .../pkg/vectorstore      9.289s
ok  .../pkg/routerruntime    3.058s
```

That run includes the two specs this PR touches:

- `IngestionPipeline bounded Stop does not report success from a second Stop while the same generation drains`
- `IngestionPipeline restart after a timed-out Stop serves a new generation while the previous one is still wedged`

`make agent-ci-lint AGENT_BASE_REF=origin/main` exits 0 with no failures across the
suites it selected. `go vet` and `go build ./pkg/...` are clean; `gofmt -l pkg/` is
empty.

Remaining risk: this is concurrent lifecycle code, so the ownership reasoning is
worth a careful read — in particular that a displaced generation keeps its own
queue, so its workers can only ever observe jobs enqueued before the restart. The
change is confined to `pkg/vectorstore`, and no exported signature changed.

### DIFF
--- a/src/semantic-router/pkg/vectorstore/pipeline.go
+++ b/src/semantic-router/pkg/vectorstore/pipeline.go
@@ -46,6 +46,83 @@ type IngestionJob struct {
 // backend or embedder is wedged.
 const defaultStopTimeout = 30 * time.Second
 
+// cleanupTimeout bounds the detached context used to persist status and count
+// updates for a job that is being torn down. Cleanup must outlive the job's own
+// cancelled context so durable counts stay coherent, but it must not be
+// unbounded: a stalled metadata registry would otherwise push Stop past its
+// deadline.
+const cleanupTimeout = 5 * time.Second
+
+// pipelineState is the tri-state lifecycle of an IngestionPipeline.
+//
+// A Stop that exhausts its deadline leaves the pipeline in pipelineStopping
+// rather than pipelineStopped, because its workers have not been reclaimed yet.
+// Keeping that distinction is what lets a second Stop join the same generation
+// instead of reporting a success that never happened.
+type pipelineState int
+
+const (
+	pipelineStopped pipelineState = iota
+	pipelineRunning
+	pipelineStopping
+)
+
+// generation is the worker-owned state of a single Start/Stop cycle.
+//
+// Each Start creates a new generation, so a timed-out Stop cannot leak an old
+// worker into a new one: the old generation keeps its own queue, WaitGroup and
+// root context until its workers actually exit, and a job enqueued after the
+// restart is only ever visible on the new generation's queue.
+type generation struct {
+	jobQueue   chan IngestionJob
+	stopCh     chan struct{}
+	rootCtx    context.Context
+	rootCancel context.CancelFunc
+	wg         sync.WaitGroup
+	// drained is closed by a single watcher goroutine once every worker in this
+	// generation has returned. Closing a channel rather than exposing wg.Wait
+	// lets several concurrent Stop calls wait on the same generation under
+	// their own deadlines.
+	drained   chan struct{}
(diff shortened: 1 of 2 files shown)


FAUST-BENCHOU (commented):
Could we keep queued-job cleanup within the caller's Stop deadline? `failQueuedJobs` runs synchronously before `joinGenerations` and each `failJob` can spend up to `cleanupTimeout` in `UpdateFileCounts`, so a stalled registry (especially with multiple queued jobs) can still make `Stop(ctx)` run well past `ctx`.

Xunzhuo:
Thanks for working on this. This overlaps with #3057, which was opened first and has now addressed the lifecycle review feedback. The current implementation here also makes queued cleanup potentially take `N * 5s` and uses detached background contexts during shutdown, so it is not the safer branch to continue. Closing this one as a duplicate so we can finish the fix in #3057.

PR #3125, opened 2026-08-30
[Feature] Add opt-in loopback pprof endpoint

drivebyer:
Related #2992

## Purpose

First of two PRs for the reproducible memory/perf baseline (criterion 1 of #2992, plan posted in the issue). Adds an opt-in pprof listener so soak runs and live debugging can capture heap/goroutine/CPU profiles:

- New `global.services.observability.profiling` config (`enabled: false` by default, `127.0.0.1:6060`), served from an explicit `http.NewServeMux` in a new `pkg/observability/profiling` package. Explicit `port: 0` requests an ephemeral port (address in the startup log). Bind must be an IP literal or `localhost`; the listener joins the shutdown hooks and sets `ReadHeaderTimeout` only (profile/trace streams must not be write-capped).
- **Hardening / behavior change**: the metrics server previously served `http.DefaultServeMux`, and `net/http/pprof` is transitively imported via controller-runtime — so `:9190` was already exposing `/debug/pprof/*` (and expvar) on all interfaces. The metrics listener now serves an explicit mux with only `/metrics`; a regression test pins that `/debug/pprof/*` returns 404 there. Anyone relying on the unintended `:9190` debug endpoints must enable `profiling` and use the loopback port.
- Docs updated per the config-contract rule (`installation/configuration.md`, `tutorials/global/api-and-observability.md`). Known gap: the operator CRD's `ObservabilityConfig` mirrors neither `Metrics` nor `Profiling`, so CRD-managed deployments cannot enable this yet (tracked as follow-up).

Affected modules: `pkg/config`, `pkg/observability`, `cmd`, k8s converter goldens (additive field), website docs.

## Test Plan

```
go vet ./cmd/ ./pkg/observability/... ./pkg/config/... ./pkg/k8s/...
go test -count=1 ./pkg/config/... ./pkg/observability/... ./pkg/k8s/... ./cmd/
go test -race -count=1 ./pkg/observability/...
```

Manual: enable profiling, `go tool pprof http://127.0.0.1:6060/debug/pprof/heap`.

## Test Result

All suites pass locally (darwin; `./cmd/` linked against locally built Rust bindings). k8s golden rewrite is idempotent. No remaining blockers.


Xunzhuo (approved):
Looks good. The dedicated mux closes the existing metrics-port pprof exposure, the new listener is opt-in and loopback by default, and the startup/shutdown/config paths are covered. No merge blocker from me. Please keep the operator CRD parity as a follow-up.

wilsonwu (approved):

PR #3134, opened 2026-08-30
[Bug] Bound external RAG response reads

guan404ming:
Related #3097

## Purpose

- Add a configurable response-size limit for external RAG APIs.
- Default unset or `0` to 4 MiB for backward compatibility.
- Update the example config and documentation.

## Test Plan

- Test default, exact-limit, over-limit, and invalid configurations.
- Run `make agent-ci-gate`.

## Test Result

- All tests passed.


AayushSaini101 (approved):

WUKUNTAI-0211:
Could we validate `external_api.backend_config.max_response_bytes` in Python too? `vllm-sr validate` currently accepts `-1`, while the Go validator rejects it at startup. A negative-case test would keep both config paths aligned.

WUKUNTAI-0211 (approved):
LGTM. The external RAG response limit, default behavior, boundary handling, and configuration updates look good.

guan404ming:
Thanks!

Issue #3106, opened 2026-08-29
[Feature] Define unknown-result policy for classifier observations in decision trees


guan404ming:
### Goal

Make an unresolved classifier observation a first-class three-valued decision-tree state instead of collapsing it into false or an internal error.

### Proposed Workgroup

Mixture-of-Models & Routing

### Scope

- Define true, false, and unknown evaluation for classifier-backed predicates.
- Preserve the backend's typed unavailable or unresolved result without exposing backend internals to the tree.
- Add an explicit terminal on_unknown action and validation for trees that can reach unknown.
- Cover unavailable, timeout, malformed-result, and partial-evaluation cases.

### Non-goals

- Backend retry, transport, or deployment lifecycle policy.
- Implicit global fail-open or fail-closed behavior.
- Selecting a safe fallback model outside the recipe.

Parent Epic: #3202

theohsiung:
This landed in the Router Models & Inference Runtime WG, and the overlap it describes is one I flagged on #2923 without resolving — so some context and a suggested reframe.

**Suggested reframe: this isn't two knobs that need a precedence rule.**

The scope reads as "backend-level `allow|block` overlaps condition-level `no_match|match`, so define their responsibilities and precedence". I'd argue the overlap only exists because both are currently forced to answer in the same currency — a boolean — when they're actually answering different questions:

| | Question it answers | Natural output |
|---|---|---|
| Classifier (produces) | Did I get a usable answer for this content? | a **state**: answered / not answered |
| Decision condition (consumes) | Given that, does this leaf match? | a **boolean** |

Today the classifier has no way to say "I didn't get an answer", so the failure has to be collapsed into match/no-match immediately — and once you add a second knob that also collapses it, you need a rulebook for which collapse wins. Giving the classifier a third outcome removes the conflict rather than adjudicating it: the classifier reports the state, each consuming condition decides what that state means for it.

[...]

Issue #2921, opened 2026-08-17
[Feature] Define the score.v1 contract for complexity backends


theohsiung:
### Goal

Define a versioned scalar-score result contract, `score.v1`, so a remote regression model can produce the complexity signal — and let complexity accept the existing `label_distribution.v1` for models that classify difficulty directly, so both remote output shapes are reachable.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

Two remote shapes, one new contract:

- **`score.v1` (new).** Add it to the shared `backend.contract` whitelist and implement `ScoringBackend` — declared in `classifier_backend_tiers.go`, still marked reserved with nothing wiring it in — over the shared connector. The router turns the score into a verdict using per-rule boundaries.
- **`label_distribution.v1` (existing, reused).** A model that returns `hard`/`easy`/`medium` needs no new contract: the winning label *is* the verdict and its probability *is* the confidence. Labels are declared inline, as `routing.signals.classifiers` already does via `newDeclaredLabelMapping`, so no mapping file is involved.

Common to both:

- Attach the backend at `global.model_catalog.modules.complexity.backend`, alongside `prototype_scoring`.
- Carry a context into the backend interfaces.
- Add parity tests covering the local path and both remote shapes.

### Design decisions

Settled while scoping, recorded so they are not re-litigated:

- **The contract follows the model's output shape.** A regression model is `score.v1` and needs boundaries; a three-class difficulty model is `label_distribution.v1` and needs none. This is what the `contract` axis is for, and it means the cheaper path does not wait on the harder one.
- **The attachment point is the module, not the rule.** `routing.signals` is replaced wholesale per recipe (`ConfigForRecipe`), so a backend declared there would silently vanish under any recipe that does not repeat it — a signal that still runs but quietly falls back to local. Every other built-in signal already declares its model under `model_catalog.modules.*`; complexity follows. The generic `routing.signals.classifiers` entries are the deliberate exception, since the signal itself is defined there.
- **For `score.v1`, direction is encoded in the field names rather than a separate field.** `hard_above` + `easy_below` where a higher score is harder; `hard_below` + `easy_above` where a lower score is harder. That removes any need for `direction`, `min`, `max` or a neutral-point field, cannot express an overlapping band, and incidentally gives the local path the asymmetric bands it cannot express today. `threshold: X` remains valid as shorthand for the symmetric pair.
[...]

Issue #2782, opened 2026-08-05
[Epic] Establish typed Router Model deployment, lifecycle, and compatibility contracts


Xunzhuo:
### Goal

Represent Router Model identity, deployment, binding, readiness, lifecycle, and support evidence with typed contracts that are resolved before request handling.

### Current baseline

#2738 already provides generation-owned runtime swap and drain behavior on current main. This Epic starts from that baseline; it does not reopen the completed lifecycle primitive.

### Scope

- Define typed manifests, deployments, capabilities, bindings, and resolved bindings.
- Qualify representative local and remote deployments with compatibility receipts.
- Return typed client and not-ready errors.
- Validate model artifacts and alias-form paths before activation.
- Share outbound connector behavior across remote model deployments.
- Gate per-deployment admission and lifecycle transitions.
- Complete the remaining configuration migration without duplicating routing policy.

### Child issues

- #3195
- #3196
- #2120
- #2669
- #2676
- #3128
- #3159

### Non-goals

- Choosing routes, fallback models, or recipe policy.
- Training or automatically promoting models.
- Claiming untested backend, model, and hardware combinations.

Parent Workgroup: #2966

theohsiung:
**Status report on Phase 1's "generalize the current direct `/classify` connector", plus three questions that need a call before four issues copy the same shape.**

- **Shipped:** #2930 (merged) generalized the connector and label-mapping validation off jailbreak-specific types. #3096 (in review) adds a remote classifier to the generic signal.
- **Converged on:** one `backend:` block per built-in classifier signal, carrying `protocol` (how the call is made) and `contract` (what shape comes back) as separate axes — the RFC's connector/contract split, projected onto config.
- **Not built:** `ModelManifest`, `Deployment`, `Capability`, `Binding`, `Receipt`. Those stay gated on this epic's first completion criterion. The intent is that the config surface they would project onto already matches, so the eventual migration is a rename rather than a redesign.
- **Questions at the bottom:** capability-vs-explicit-axes, offset units for `token_spans`, and two roles with no consumer.

## One signal that the RFC's shapes hold up

`alignScoresToMapping` (#2930) requires an `http_classify` response to carry every declared label and sum to ~1.0, rejecting a truncated top-k response rather than defaulting the missing entries to zero. That was derived from a security argument — a top-k response that omits the positive label silently under-reports risk — and only afterwards did it turn out to be exactly `label_distribution.v1`'s `preserve_all_declared_labels: true`.

[...]

PR #3138, opened 2026-08-30
[Feature] Add shared classifier backend for category

### LINKED ISSUE #2920: [Feature] Add a shared remote-backend block for built-in classifier signals

alexsmolya:
## Summary

This adds a shared remote `backend:` block for built-in classifier signals, with category/domain as the first consumer. The remote path is selected explicitly while the existing local category path remains unchanged when `backend` is absent.

Public backend fields are:

- `protocol`
- `contract`
- `model`
- `deadline_ms`

For category, the supported configuration is:

```yaml
backend:
  protocol: http_classify
  contract: label_distribution.v1
  model: domain-service
  deadline_ms: 5000
```

The named model resolves exactly through `global.model_catalog.external[].name`. Initialization validates its classification role/type, model name, endpoint address, port range, endpoint protocol, backend protocol, semantic contract, and deadline. Request context and cancellation are propagated to the HTTP classifier.

Remote category classification preserves the complete probability distribution and existing downstream semantics, including `model_scores`, `mmlu_categories`, `fallback_category`, entropy and multiple matching rules, `MatchedDomainRules`, `x-vsr-selected-category`, and routing priors consumed by static, Elo, and RL-driven selection.

The legacy `use_modernbert` and `use_mmbert_32k` keys remain readable and normalize to canonical `variant` values. Explicit `modernbert`, `mmbert32k`, and `candle` selections use matching runtime inference paths; the empty legacy selector retains auto-detection. An explicit remote backend overrides an inherited canonical local default, while explicitly contradictory selectors remain rejected.

Model-info reporting uses the effective normalized category variant, or the remote protocol for a remote category backend. Classifier failure-policy design is outside this PR and remains delegated to #3106/#3145. Prompt_guard migration and #2921/#2922 runtime contracts are out of scope.

## Validation

Passing: agent manifest/workflow validation, YAML parsing, recipe-conformance static Python checks, and `git diff --check`.

Go tests, gofmt, go vet, race tests, repository Go gates, local image/smoke, and the Envoy AI Gateway E2E are pending CI/environment validation because the local environment has no Go toolchain (`go`/`gofmt`).

The operator `CategoryModelConfig` CRD propagation of `variant`/`backend`, including generated operator bundle/manifests, remains explicit follow-up scope as noted in review.

Fixes #2920


FAUST-BENCHOU (commented):
pls resolve conflicts

theohsiung (commented):
Awesome job thx @alexsmolya! 

Checked against #2920 the structure is aligned: one shared backend type with per-consumer layering, `protocol`/`contract` as separate axes, explicit model naming with the role kept as a type assertion rather than a lookup key, `on_error` left on the module, and the full distribution preserved through the shared connector, which is what keeps `model_scores` and the entropy path intact. Comments inline.

One that can't be anchored inline: `CategoryModelConfig` (`deploy/operator/api/v1alpha1/semanticrouter_types.go`) has neither `variant` nor `backend`, so the API server prunes both and an operator-managed deployment silently keeps the local default — same class as the `on_error` gap fixed in #2759. Regenerating needs `deploy/operator/bundle/manifests/` too, not just `config/crd/bases/`. Fine as a follow-up.

src/semantic-router/pkg/config/classifier_backend.go
```diff
@@ -0,0 +1,217 @@
...
+
+// RemoteClassifierBackend is the shared remote attachment contract for
+// built-in classifier modules. A nil backend means that the module uses its
+// existing local implementation. TimeoutSeconds is a pointer so omitted and
+// an explicitly invalid zero value cannot be confused during validation.
+type RemoteClassifierBackend struct {
+	Protocol       string `yaml:"protocol" json:"protocol"`
+	Contract       string `yaml:"contract,omitempty" json:"contract,omitempty"`
+	Model          string `yaml:"model" json:"model"`
+	TimeoutSeconds *int   `yaml:"timeout_seconds,omitempty" json:"timeout_seconds,omitempty"`
```
theohsiung: #2920's "Vocabulary alignment with RFC #2779" section asks for `deadline_ms` here. Since `UnmarshalYAML` rejects unknown fields, a config copied from the issue's own example fails to load with `backend: unsupported field "deadline_ms"`. The rename costs nothing now and avoids one once #2782's object model lands — the RFC's own signal example uses `deadline_ms`.

src/semantic-router/pkg/config/classifier_backend.go
```diff
@@ -0,0 +1,217 @@
...
+const (
+	RemoteClassifierProtocolHTTPClassify = "http_classify"
+	RemoteClassifierProtocolHTTPChat     = "http_chat"
+)
+
+// Remote classifier contracts describe the semantic product returned by a
+// remote classifier. Category is the first consumer and currently supports
+// the complete label distribution contract.
+const (
+	RemoteClassifierContractLabelDistribution = "label_distribution"
```
theohsiung: Same section asks for versioned contract names. As it stands `contract: label_distribution.v1` is rejected as an unsupported value, so the issue's example config doesn't validate. The RFC's core contracts are versioned from the start.

src/semantic-router/pkg/config/canonical_defaults.go
```diff
@@ -247,7 +247,7 @@ func defaultClassifierModule() CanonicalClassifierModule {
 			CategoryModel: CategoryModel{
 				Threshold:           0.5,
 				UseCPU:              true,
-				UseMmBERT32K:        true,
+				Variant:             CategoryVariantMmBERT32K,
```
theohsiung, line 255: This moves the default off the booleans, and `normalizeCanonicalCategoryVariant` clears both whenever a variant resolves — but `route_model_info_classifiers.go` is unchanged and still derives the reported type from them:

```go
"model_type": resolveInlineModelType(categoryModel.UseMmBERT32K, categoryModel.UseModernBERT, false)
```

Both are false by construction now, so every default deployment reports `model_type: "bert"` for `category_classifier` while actually running mmbert32k. Reading the effective variant fixes it; a remotely-backed classifier probably wants to report its protocol instead of a local type, the way the `prompt_guard` entry below already does.

config/config.yaml
```diff
@@ -2344,10 +2344,10 @@ global:
           model_id: models/mmbert32k-intent-classifier-merged
           threshold: 0.5
           use_cpu: true
-          use_modernbert: false
-          use_mmbert_32k: true
+          variant: mmbert32k
```
theohsiung: With `ValidateCategoryModelBackend` rejecting `variant` alongside `backend`, a user who starts from this file — the documented starting point — and pastes in a `backend:` block has to delete this line first. That's the "un-set something to switch" friction #2920 set out to remove ("Absent means the local path… nothing has to be un-set"). You solved the same trap at the canonical-defaults layer with the raw-override check; this layer needs the same treatment, or an explicit `backend` should win over an inherited `variant`.

Xunzhuo (changes requested):
Two blockers on the current head:

1. The public contract does not match #2920: this adds `timeout_seconds` and unversioned `label_distribution`, while the accepted shape uses `deadline_ms` and `label_distribution.v1`. Please keep the first shared backend contract aligned with the RFC vocabulary instead of introducing a rename immediately afterward.
2. `variant: modernbert` still installs the auto-detecting `CategoryInferenceImpl`. Its top-1 path tries Candle first, so an explicitly selected ModernBERT backend can execute a loaded Candle model on fallback/direct calls. Give the explicit variant an explicit ModernBERT inference path and cover it with a regression test.

Please also rebase the branch and fix DCO before the next pass.

src/semantic-router/pkg/config/classifier_backend.go
alexsmolya: Changed `RemoteClassifierBackend` to `deadline_ms` (milliseconds), including strict YAML decoding, validation, default 5000, HTTP timeout wiring, tests, examples, and the PR description.

src/semantic-router/pkg/config/classifier_backend.go
alexsmolya: Changed the category contract to `label_distribution.v1` throughout validation, examples, E2E config, and tests; the accepted #2920 example now loads with strict unknown-field checking.

src/semantic-router/pkg/config/canonical_defaults.go
alexsmolya: `category_classifier` model info now reports the effective normalized category variant (`mmbert32k`/`modernbert`) and reports the remote protocol for a remote category backend. Added regression coverage for both local variants and the remote convention.

config/config.yaml
alexsmolya: The canonical loader now lets an explicit backend replace the inherited default local variant, while retaining rejection when a raw override explicitly supplies an active local selector alongside the backend. Existing coverage exercises both paths.

theohsiung:
Thanks for the fast turnaround on this — you turned around three review points in about an hour, and I checked each one in the code rather than the description: `deadline_ms`, `label_distribution.v1`, and the model-info reporting all hold up. The strict unknown-field decoding you added is a nice touch beyond what was asked.

### Ask: drop the `on_error` work from this PR

Error policy moved to #3106 (`accepted`, with #3145 in flight).

**This one is our miss, not yours.** #2920's 8/18 comment said `CategoryModel.OnError` "is where it lands", and nothing on the issue pointed at #3106 until I edited that comment this morning — after you opened this PR.

**Two parts are hard to retract once merged:**

- **`classification_error` reserved label** — the check in `LoadCategoryMapping` is unconditional, running regardless of whether a backend or `on_error` is configured. An existing deployment whose mapping uses that label fails to start after upgrading.
- **`CategoryModel.OnError`** — public config surface. #3106's direction is a three-valued `Unknown` plus `on_unknown`, not a per-module `on_error`, so changing it later is a breaking config change.

**#3145 won't supersede it on its own** — it is scoped to "add internal Unknown evaluation while preserving existing behavior", so the two policies would coexist with no defined precedence.

The shared block, validation, remote classification and the E2E are unaffected by dropping it.

config/config.yaml
theohsiung: Reopening — this still reproduces on `aba8a4eb`.

The relaxation only covers an *inherited* selector:

```go
if hasRawKey(rawDomain, "backend") && !hasActiveRawCategoryLocalSelector(rawDomain) {
```

Any non-empty `variant` in the raw override skips the clearing branch — and this PR guarantees every config has one: `reference_config_global_test.go` excepts `backend`, `use_modernbert` and `use_mmbert_32k` from the coverage assertion, but not `variant`, so `config/config.yaml` must carry `variant: mmbert32k`.

So: the test requires the shipped config to state `variant` → user copies it → adds `backend:` → rejected. Anyone starting from the reference config has to delete a line the test insists is there.

Either fix works:

- except `variant` alongside `backend` and drop it from `config/config.yaml`
- or tolerate a `variant` equal to the canonical default, rejecting only one the user set against the backend

alexsmolya:
Removed the category failure-policy scope from #3138: CategoryModel no longer exposes that policy, category mappings no longer reserve `classification_error`, and the related category runtime, tests, config, and documentation were removed. The accepted backend contract, strict decoding, dispatch, model-info, normalization, remote classification, and distribution behavior remain intact. Failure-policy work remains delegated to #3106/#3145.

Xunzhuo (changes requested):
Thanks, the two earlier blockers are fixed on this head: the public contract now uses deadline_ms and label_distribution.v1, and explicit ModernBERT dispatch no longer falls through to Candle. One blocker remains: TestValidateCategoryModelBackend/incorrect_contract now sets label_distribution.v1, which is the valid contract, but still expects validation to reject it. That makes the config test and core gate fail. Please change that fixture to an actually invalid value and rerun the current CI; after that I do not see a remaining blocker from this pass.

theohsiung:
Thx for the removal of the `on_error` part. I reopened the config/config.yaml thread a few minutes after this push, detail are shown in that thread.

config/config.yaml
alexsmolya: Implemented Option A. The category variant is no longer forced into config/config.yaml, and the reference-field coverage test exempts variant; canonical defaults still resolve an unset local category selector to mmbert32k. Added TestReferenceConfigCategoryBackendReplacesDefaultVariant, which reads the actual reference YAML, adds the explicit backend, and validates through ParseYAMLBytes. Explicit local variant plus backend remains rejected. Pushed as 7c825c9e.

alexsmolya:
Updated the stale `TestValidateCategoryModelBackend/incorrect_contract` fixture to use `label_distribution.v0`, which is genuinely unsupported; `label_distribution.v1` remains the valid contract. The signed fix is pushed at `7255f778`. `git diff --check` and repository agent/static validation passed; Go-focused tests and gofmt could not run locally because Go is unavailable. Fresh PR checks are currently pending.

theohsiung:
can u resolve the conflicts, thx~

alexsmolya:
Resolved the conflicts by rebasing the existing branch onto current `main` (`1de48ced`). The accepted #3138 semantics and the `label_distribution.v0` negative fixture are preserved. Updated head: `bfc82094`. GitHub now reports the PR as mergeable; fresh checks are running.

Xunzhuo (dismissed):
Thanks for the update. The invalid-contract fixture is now genuinely invalid, the backend contract and explicit dispatch blockers are resolved, and the rebased head merges cleanly. Approving while the refreshed CI finishes.

wilsonwu (approved):

theohsiung (approved):
All five points are addressed, verified in the code at `e855ecb8` rather than from the description: `deadline_ms` and `label_distribution.v1` throughout, with tests pinning the old spellings as rejected; model-info reporting off the effective variant; the `on_error` scope fully removed with no leftovers; and the reference-config/coverage-test loop broken by Option A, plus a test that loads the actual reference YAML.

One thing tracked rather than fixed here: the Operator CRD still cannot express `variant` or `backend`, so an operator-managed deployment silently keeps the local default while the new tutorial tells users to write `backend:`. I'll take that as a follow-up on top of this — flagging it so it does not get lost, not asking for it in this PR.

theohsiung (dismissed):
huge pr, thx for your contribution! please fix the ci

wilsonwu (dismissed):

Xunzhuo:
@alexsmolya can you fix CI and conflicts

Xunzhuo (changes requested):
The earlier backend contract and explicit-variant blockers remain fixed. One current-head blocker is visible in the dedicated E2E: providers.defaults.default_model is base-model, but routing.modelCards does not declare base-model, so the router exits during config load and the category-remote-backend job fails. Please add the matching model card and rerun that profile; the other red gate contexts are aggregate failures from this E2E.

FAUST-BENCHOU (commented):
Nice improvement to make category classification follow the same backend abstraction pattern. This should reduce the gap between local and remote classifier paths.

One thing to consider is whether the validation rules for `http_classify` backends should be shared across classifier types instead of being introduced category-by-category, so future classifier backends can reuse the same contract.

Xunzhuo (changes requested):
Thanks for the update — the `base-model` card blocker is fixed. One current-head blocker remains in the dedicated E2E: `category_backend_math` references `domain("math")`, but this recipe declares no domain signal, so the router exits with `runtime_config_load_failed` before the test can run. Please add the matching recipe-scoped domain signal declaration and rerun `category-remote-backend`.

Xunzhuo (approved):
Thanks for the quick fix — this resolves the remaining E2E blocker. The dedicated profile now declares the `math` domain used by `category_backend_math`, and the focused maintained-config contract passes. CI is still running and the branch is behind `main`, but I found no remaining code blocker on this head.

PR #681, opened 2025-11-17
[Feat]: Signal-Decision Driven Semantic Routing with Dynamic Plugin Architecture

Xunzhuo:
This commit introduces a comprehensive decision-based routing system with a flexible plugin architecture, replacing the previous category-based approach.

IntelligentRoute
```yaml
apiVersion: vllm.ai/v1alpha1
kind: IntelligentRoute
metadata:
  name: ai-gateway-route
  namespace: default
spec:
  signals:
    keywords:
      - name: "thinking"
        operator: "OR"
        keywords: ["urgent", "immediate", "asap", "think", "careful"]
        caseSensitive: false

    domains:
      - name: "business"
        description: "Business and management related queries"
      - name: "law"
        description: "Legal questions and law-related topics"
      - name: "psychology"
        description: "Psychology and mental health topics"
      - name: "biology"
        description: "Biology and life sciences questions"
      - name: "chemistry"
        description: "Chemistry and chemical sciences questions"
      - name: "history"
        description: "Historical questions and cultural topics"
      - name: "health"
        description: "Health and medical information queries"
      - name: "economics"
        description: "Economics and financial topics"
      - name: "math"
        description: "Mathematics and quantitative reasoning"
      - name: "physics"
        description: "Physics and physical sciences"
      - name: "computer science"
        description: "Computer science and programming"
      - name: "philosophy"
        description: "Philosophy and ethical questions"
      - name: "engineering"
        description: "Engineering and technical problem-solving"
      - name: "other"
        description: "General knowledge and miscellaneous topics"

  decisions:
    - name: "business_decision"
      priority: 10
      description: "Business and management related queries"
      signals:
        operator: "OR"
        conditions:
          - type: "domain"
            name: "business"
      modelRefs:
        - model: "base-model"
          loraName: "social-expert"
          useReasoning: false
      plugins:
        - type: "pii"
          configuration:
            enabled: true
            pii_types_allowed: []
        - type: "system_prompt"
          configuration:
            enabled: true
            system_prompt: "You are a senior business consultant and strategic advisor with expertise in corporate strategy, operations management, financial analysis, marketing, and organizational development. Provide practical, actionable business advice backed by proven methodologies and industry best practices. Consider market dynamics, competitive landscape, and stakeholder interests in your recommendations."
            mode: "replace"

    - name: "law_decision"
      priority: 10
      description: "Legal questions and law-related topics"
      signals:
        operator: "OR"
        conditions:
          - type: "domain"
            name: "law"
      modelRefs:
[...]

PR #3141, opened 2026-08-31
[Bug] Open a PII entity on an orphan I- tag

### LINKED ISSUE #3140: [Bug] PII entities whose first token is labelled I- are dropped
### Proposed Workgroup

### Describe the bug

`merge_bio_entities` (`candle-binding/src/model_architectures/traditional/modernbert.rs:1298`) only starts an entity on a `B-` tag. An `I-` run with nothing open in front of it is thrown away.

The PII model doesn't always emit `B-`. Over 138 texts, argmax per token:

```
EMAIL_ADDRESS      0 B-   193 I-
DOMAIN_NAME        0 B-    49 I-
```

Zero `B-`, so those two types can never be produced. The tokens are there and confident, they just never become an entity:

```
"Please send the report to bob@example.org and cc carol@example.org."
  ' bob' I-EMAIL_ADDRESS 0.669, '@' 0.638, 'example' 0.862, '.' 0.878, 'org' 0.692
  entities returned: []
```

`openvino-binding/cpp/src/classifiers/token_classifier.cpp:108-128` decodes the same models and does the opposite:

```cpp
// Type mismatch or no current entity - treat as new entity
```

So the answer depends on which binding is loaded. The Go layer sides with openvino too: `stripBIOPrefix` (`pkg/classification/mapping.go:226`) and `pkg/utils/pii/policy.go:93-101` both accept a bare `I-` as a real detection.

### To Reproduce

Default config, `main` at `e1c6b8ef`.

Threshold doesn't help, there's no entity to threshold:

```
POST /api/v1/classify/pii  "Email me at alice.wong@acme.co.uk about the invoice."
  0.05 -> [PERSON]    0.30 -> [PERSON]    0.50 -> []    0.90 -> []
```

Masking is the part that bites:

```
in:  "Contact Sarah Chen at sarah.chen@example.org or 415-555-0134."
out: "Contact[PERSON_0] at sarah.chen@example.org or [PHONE_NUMBER_0]."
```

Name and phone masked, email returned in the clear. (The eaten space is the same bug: the span starts on it.)

### Expected behavior

An `I-` with no matching open entity starts one, like openvino already does.

### Impact

14 of 17 types lose detections. EMAIL_ADDRESS and DOMAIN_NAME are undetectable, so `pii_types_allowed` can't say anything about them. Masking hands back unmasked emails, which is worse than not masking, because the response looks clean.

It also blocks external models. `piiranha-v1-detect-personal-information` returns **0 entities** here, because it labels with `I-` only. That is a problem for #2760 / #2922.

### What a fix costs

I patched the merge locally so an orphan `I-` opens an entity, and ran the project's own e2e suite. It does what it should, and it also breaks two other tests:

```
pii-detection             18/25 -> 22/25
domain-classify          171/261 -> 148/261   (gate 60%)
plugin-config-variations  16/18 -> 14/18      (gate 80%)
```

The domain regression is requests losing their category: misclassifications with an empty `x-vsr-selected-category` go 67 -> 92. PII now matches on prompts that used to reach a domain decision.

Running those 261 prompts through both builds, the newly firing entities are all one token of noise:

```
STREET_ADDRESS '0' '89' '32'      DATE_TIME '0' '°' '0°'
PHONE_NUMBER   '1' '0'  '62'      EMAIL_ADDRESS '2' 'Pa' 'theta'
ORGANIZATION   '1' 'polar'        DOMAIN_NAME 'g'
```

Over the PII corpus, orphan-opened entities split by run length:

```
run length   type matches GT   no match   precision
1 token                   79        357       0.181
2+ tokens                145        174       0.455
```

The model does emit isolated `I-` tags on digits, and the strict merge was quietly absorbing them. The 1-token "matches" are mostly fragments of a longer entity (`CREDIT_CARD '9'`, `DATE_TIME '.'`) rather than usable spans; scoring here is per type, not per span.

On the PII side the fix works: 142 labelled sentences at threshold 0.5 go from recall 0.338 / precision 0.680 to recall 0.643 / precision 0.522, with EMAIL_ADDRESS 0/24 -> 21/24 and STREET_ADDRESS 2/9 -> 9/9.

So this is a decision rather than a patch:

- take the faithful merge and re-tune the PII thresholds, since the current ones were calibrated against a decoder that discarded most spans
- add a run-length condition. Requiring two tokens holds EMAIL_ADDRESS at 18/24 and cuts the extra domain hits from +24 to +10, but it also drops genuine single-token entities such as a one-token `GPE`, so it is a heuristic rather than a rule
- keep the strict merge behind a config option, which leaves the shipped config unable to detect emails and so does not really answer this

Worth stating plainly: the domain-classify baseline of 171/261 depends on PII being broken. The model's false positives on isolated digits exist today, and the merge is what hides them.

This doesn't make PII detection good. Recall is still 0.256 at the shipped 0.9 threshold and US_SSN is 12/35. That's model quality, separate issue.

### Additional context

#712 reported the same symptom. #765 closed it by fixing the FFI fallback and adding BIO prefix stripping in Go; neither touched this merge.

The line dates to `0e9b7a18` (2025-08-11) and was correct then: the same PR added the trainer that guarantees a `B-`. The model was swapped later in #1553, a schema refactor, and nothing rechecked the tags.

`test_modernbert_bio_entity_boundaries` pins the current behaviour. I added it in #3027 as a no-behaviour-change test, so it needs updating.

subin9:
Fixes #3140

An `I-` tag with no matching open entity now starts one, instead of being dropped. Same for an `I-` whose type differs from the open entity. `openvino-binding` already decodes the same models this way.

The PII model labels emails and domains with `I-` only, so requiring `B-` made both types undetectable at any threshold.

Also trims whitespace from spans. The tokenizer puts the leading space inside the token, so masking ate it:

```
before: "Contact[PERSON_0] at sarah.chen@example.org or [PHONE_NUMBER_0]."
after:  "Contact [PERSON_0] at [EMAIL_ADDRESS_0] or [PHONE_NUMBER_0]."
```

142 labelled sentences, threshold 0.5:

```
recall     0.338 -> 0.643
precision  0.680 -> 0.522
EMAIL_ADDRESS 0/24 -> 21/24   STREET_ADDRESS 2/9 -> 9/9   DOMAIN_NAME 0/8 -> 6/8
```

### Threshold recalibration

The first revision dropped `domain-classify` and `plugin-config-variations` under their gates. The cause is the `ai-gateway` profile's `pii_deny_all` threshold of 0.5: with the merge fixed, PII fires on 83 of the 261 domain-classify prompts instead of 59, and the entities it newly finds are single tokens of noise, `STREET_ADDRESS '0'`, `DATE_TIME '°'`, `DOMAIN_NAME 'g'`.

That 0.5 was set in #681 for the previous PII model, `mom-pii-classifier`, which emits proper IOB2 (100% `B-` initial on EMAIL_ADDRESS, DOMAIN_NAME, CREDIT_CARD). #1553 swapped the model to `mmbert32k-pii-detector-merged` and carried the threshold across unchanged. Against a decoder that was discarding most spans it looked fine.

Raising it to 0.7 puts all three suites at or above where they are today:

```
                                upstream @0.5   this PR @0.7
domain-classify prompts flagged            59             53
plugin-config queries flagged               1              1
pii-detection cases detected            18/25          21/25
```

I did not add a two-token rule, for the reason you gave: it would discard valid single-token entities from other models.

One thing worth flagging on wording. Every shipped PII threshold is already 0.7 or higher (`config/config.yaml` and the privacy and agent recipes use 0.85, the module default is 0.9, multi-objective uses 0.7), and the deploy values do not set PII at all. `production-stack` runs `domain-classify` at 0.9 and does not regress, and `ml-model-selection` has no PII signal. So there was nothing miscalibrated on the shipped side; the only stale value is the one in the `ai-gateway` E2E profile. Happy to move it elsewhere if you meant something different.

Five new tests. `test_modernbert_bio_entity_boundaries` asserted that orphan and type-changing `I-` tags are discarded and now expects 4 entities instead of 2. I added that test in #3027 as a no-behaviour-change pin, so it was describing the bug.

### DIFF
--- a/candle-binding/src/model_architectures/traditional/modernbert.rs
+++ b/candle-binding/src/model_architectures/traditional/modernbert.rs
@@ -1273,19 +1273,61 @@ impl BioEntity {
     }
 }
 
+/// Trim surrounding whitespace from an entity span.
+///
+/// Sub-word tokenizers attach the preceding space to a word-initial token, so
+/// a `PERSON` span starts on the space before the name. Callers slice the text
+/// with these offsets, and PII masking then eats that space.
+fn trim_entity_span(text: &str, start: usize, end: usize) -> (usize, usize) {
+    let slice = &text[start..end];
+    let trimmed = slice.trim_matches(|c: char| c.is_whitespace());
+    if trimmed.is_empty() {
+        return (start, end);
+    }
+    let lead = slice.len() - slice.trim_start_matches(|c: char| c.is_whitespace()).len();
+    let trail = slice.len() - slice.trim_end_matches(|c: char| c.is_whitespace()).len();
+    (start + lead, end - trail)
+}
+
 /// Merge a sequence of BIO-labelled tokens into entities.
 ///
 /// `tokens` must already have special tokens removed and must be in text order.
 /// Offsets are byte offsets into `text`.
+///
+/// An `I-` tag that does not continue an open entity opens one of its own type.
+/// Models are not obliged to emit `B-`: the mmBERT-32K PII detector labels
+/// emails and domains with `I-` only, so requiring `B-` dropped them entirely.
 pub(crate) fn merge_bio_entities(text: &str, tokens: &[BioToken<'_>]) -> Vec<BioEntity> {
     let mut entities: Vec<BioEntity> = Vec::new();
     let mut current: Option<BioEntity> = None;
 
+    // Close `current`, trim its span and push it.
+    fn flush(text: &str, entities: &mut Vec<BioEntity>, entity: BioEntity) {
+        let mut entity = entity;
+        let (start, end) = trim_entity_span(text, entity.start, entity.end);
+        entity.start = start;
+        entity.end = end;
+        entity.text = text[start..end].to_string();
+        entities.push(entity);
+    }
+
     for token in tokens {
-        if let Some(entity_type) = token.label.strip_prefix("B-") {
-            // Beginning of a new entity.
+        let opens_new_entity = match token.label.strip_prefix("B-") {
+            Some(entity_type) => Some(entity_type),
+            None => match token.label.strip_prefix("I-") {
+                // Continues the open entity only when the type matches,
+                // otherwise it starts a new one.
+                Some(entity_type) => match current {
+                    Some(ref entity) if entity.entity_type == entity_type => None,
+                    _ => Some(entity_type),
+                },
+                None => None,
+            },
+        };
+
+        if let Some(entity_type) = opens_new_entity {
             if let Some(entity) = current.take() {
-                entities.push(entity);
+                flush(text, &mut entities, entity);
             }
             current = Some(BioEntity {
                 entity_type: entity_type.to_string(),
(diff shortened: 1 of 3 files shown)


Xunzhuo (changes requested):
The decoder bug is real, but we cannot merge a change that knowingly drops both `domain-classify` and `plugin-config-variations` below their gates. Please keep the BIO fix, then recalibrate the shipped PII threshold/model behavior so the affected routing E2E suites pass. I would not add a global two-token rule: that would silently discard valid single-token entities for other models.

subin9:
Kept the BIO fix and left the two-token rule out. Raised the `ai-gateway` profile's `pii_deny_all` threshold from 0.5 to 0.7, which was set in #681 for the previous PII model and carried across unchanged when #1553 swapped in mmbert32k. At 0.7 all three suites land at or above where they are today: domain-classify prompts flagged 59 -> 53, plugin-config 1 -> 1, pii-detection 18/25 -> 21/25. Every shipped threshold is already 0.7 or higher, so the profile value was the only stale one; happy to move it if you meant elsewhere.

Xunzhuo (approved):
Thanks, this addresses the blocker cleanly. The BIO fix stays general, there is no two-token heuristic, and the threshold adjustment is limited to the stale ai-gateway profile value while restoring the affected gates. The orphan, type-change, whitespace, and masking cases are covered. No code blocker from me; merge after the /accept workflow has processed #3140 and the current CI run is green.

wilsonwu (approved):

Issue #3128, opened 2026-08-30
[Feature] Extract a shared connector for remote Router Model deployments


theohsiung:
### Goal

Provide one reusable outbound connector for remote Router Model deployments so lifecycle, deadlines, authentication material, and transport errors are handled consistently.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Define connector configuration and secret-reference boundaries.
- Share connection reuse, deadlines, cancellation, retry eligibility, and typed transport errors.
- Integrate representative remote classifier deployments through the connector.
- Keep deployment readiness and diagnostics visible through #2782 contracts.

### Non-goals

- Provider selection or routing fallback policy.
- A generic proxy transport layer.
- Signal-specific score or span semantics.

Parent Epic: #2782

### Completion signal

Complete when the scoped behavior is implemented and its stated validation passes.

Xunzhuo:
@theohsiung lets unify the code here, thanks for picking it up

guan404ming:
Hi @theohsiung I am interested in this issue could I help implement this?

theohsiung:
@guan404ming I've already assinged [@giantprince](https://github.com/GiantPrince) on slack, srry. But you are welcome to share your thought on this issue.

guan404ming:
Sure, thanks for letting me know. I could help review the code. Thanks!

JiaoliangYu:
Implemented by #3328. The separate Looper transport/lifecycle follow-up is now tracked in #3339 so this issue can retain its original boundary.

@Xunzhuo could you close this issue when convenient? I do not have triage permission to close it directly.

theohsiung:
@JaredforReal Thanks

PR #3143, opened 2026-08-31
[Bug] Honor generic classifier max_tokens

### LINKED ISSUE #3099: [Bug] Generic classifier signal ignores configured max_tokens
### Proposed Workgroup

### Describe the bug

- `generic_classifier.go:30` reads only endpoint, key, timeout and model name.
- `Classify` hardcodes the token budget:

```go
&GenerationOptions{
    MaxTokens:   128,
    Temperature: 0,
    JSONMode:    true,
}
```

- 128 must cover `{"label":"...","rationale":"..."}`; `max_tokens` cannot raise it.
- Truncated output then fails the JSON and rationale checks.
- Sibling paths hardcode 512 and 1000, also unexplained.
- `req_filter_memory_rewrite.go:193` reads both from config, with defaults.
- `Temperature: 0` is deliberate everywhere and should stay fixed.

### To Reproduce

- Point a `type: llm` signal at a verbose model.
- Raise `max_tokens` on its external model entry.
- The call still runs at 128 and fails.
- `config.yaml:535` `tool-risk` is such a signal.

### Expected behavior

- Honor `max_tokens` from the external model, or
- Reject it at config load for `type: llm` signals.

### Impact

- Fails on responses the operator cannot shorten.
- No configuration workaround.

### Additional context

- Reference config carries `max_tokens: 128`, the hardcoded value.
- Same entry carries `parser_type: json`, read only by guardrail.
- Surfaced during review on #3093.

guan404ming:
Close #3099

## Purpose

- Honor the configured `max_tokens` for generic classifier requests.
- Preserve `128` as the default when unset.
- Add coverage for configured and default values.

## Test Plan

- Verify configured `max_tokens` is sent to the classifier backend.
- Verify the default remains `128`.

## Test Result

- `go test ./pkg/classification`
- `make agent-ci-gate`
- All maintained recipe conformance tests and local smoke checks passed.


AayushSaini101 (dismissed):
thanks : )

WUKUNTAI-0211 (dismissed):
LGTM

src/semantic-router/pkg/classification/generic_classifier.go
```diff
@@ -42,12 +45,17 @@ func newLLMLabelClassifier(
 	if timeout <= 0 {
 		timeout = 5 * time.Second
 	}
+	maxTokens := external.MaxTokens
```
theohsiung, line 48: `max_tokens` is now unbounded upward, and the hardcoded `128` was incidentally the only thing bounding the response on this path: `vllm_client.go:180` reads the body with a bare `io.ReadAll(resp.Body)`, with no `MaxResponseBytes` — #3093's byte ceilings landed on `http_classifier.go` only.

So `max_tokens: 100000000` now passes config load and comes back through an unbounded read. Worth a ceiling, either validated at config load or clamped here. Alternatively note it on #3128, which already scopes byte limits as a connector-owned concern — this is one more caller that needs them.

theohsiung (commented):
Fix looks right: it honors `external.MaxTokens`, keeps `Temperature: 0` pinned, and the `<= 0` fallback mirrors the `timeout` guard directly above it. Two follow-ups, neither blocking.

**The docs now contradict this change.** `website/docs/tutorials/signal/learned/classifier.md` still tells operators that the runtime fixes "temperature, output schema, **token bounds**, and exact-label validation". After this PR only the other three are fixed. That sentence is exactly where someone would look before setting `max_tokens`, so leaving it as-is means #3099 stays unfixed in practice for anyone who reads the docs first. Dropping "token bounds" from that list covers it.

src/semantic-router/pkg/classification/generic_classifier.go
guan404ming: Thanks! The docs have been updated. I think I would prefer to follow up this in #3128 with a transport-level response-size limit rather than a model-agnostic max_tokens ceiling.

This follows the existing http_classify [bounded-read pattern](https://github.com/vllm-project/semantic-router/blob/main/src/semantic-router/pkg/utils/http/readlimit.go), and [TiDB’s inference runtime](https://github.com/pingcap/tidb/blob/65ac2fad582510b92d3c4b79999e48217c989737/pkg/inference/embedding/base/base.go#L46-L69) uses the same LimitReader(maxBytes + 1) approach with a configurable byte ceiling:

theohsiung (approved):
thx LGTM!

wilsonwu (approved):

WUKUNTAI-0211 (approved):

PR #3145, opened 2026-08-31
[Feature] Model classifier failures as unknown

guan404ming:
Related #3106

## Purpose

- Model classifier backend failures as `Unknown`.
- Apply ternary `AND`, `OR`, and `NOT` evaluation.
- Preserve existing classifier and Prompt Guard failure behavior.

## Test Plan

- `go test ./pkg/decision ./pkg/classification`
- `make agent-ci-gate`

## Test Result

- Passed.


Xunzhuo (approved):
This split looks good. Unknown now propagates through NOT/AND/OR, while an unresolved tree still falls back to the existing generic-classifier and Prompt Guard policies, so this PR does not change the public config behavior before the follow-up lands. The selection extraction is behavior-preserving as well.

No code blocker from me; merge after the current CI run is green.

guan404ming:
Thanks for your review.

theohsiung (approved):
LGTM

wilsonwu (approved):

PR #3147, opened 2026-08-31
[Bug] Report PII entity positions in code points

### LINKED ISSUE #3146: [Bug] PII entity positions are byte offsets reported as character positions
### Proposed Workgroup

### Describe the bug

`candle-binding/semantic-router.go:522` declares:

```go
Start      int     // Start character position in original text
End        int     // End character position in original text
```

The Rust side fills those with byte offsets (`modernbert.rs:1242`, "Byte offset of the token's start in the original text"), and `/api/v1/classify/pii` passes them through to `start_position` / `end_position` unchanged.

Second, smaller thing in the same two fields: they are plain ints with `omitempty` (`classification_pii.go:39-40`), so an offset of 0 is dropped from the response.

### To Reproduce

`main` at `e1c6b8ef`, default config.

```
"こんにちは、私は John Smith です。電話は 415-555-0134"   39 code points, 67 bytes

PERSON        start_position 24  end_position 35
PHONE_NUMBER  start_position 55  end_position 67
```

The phone span ends at 67 on a 39-character string.

```
"John Smith called yesterday."   -> PERSON, end_position 10, no start_position
"415-555-0134 is my number."     -> PHONE_NUMBER, end_position 12, no start_position
"Berlin is where I live."        -> GPE, end_position 6, no start_position
```

### Expected behavior

`start_position` / `end_position` are code-point offsets, and an entity at offset 0 reports it.

### Impact

Any client that indexes by code point gets a wrong span, or an out-of-range one on non-ASCII text. Python, JavaScript and Java callers all do.

Note the internal offsets should stay as bytes: `buildMaskedPIIText` (`classification_pii_response.go:116`) slices the Go string with them, and Go strings are byte-indexed, so masking is correct today. Only the API boundary needs converting.

### Additional context

Relevant to #2922, which asks the `token_spans` contract to state its offset unit and cites `TokenEntity` as already meaning character positions.

subin9:
Fixes #3146

`start_position` / `end_position` now convert the classifier's byte offsets to code-point offsets at the API boundary.

Internal offsets stay as bytes. `buildMaskedPIIText` slices the Go string with them and Go strings are byte-indexed, so masking is unaffected.

The two fields also become pointers, so an entity at offset 0 reports `start_position: 0` instead of dropping it. Absent now means the caller did not ask for positions.

```
"こんにちは、私は John Smith です。電話は 415-555-0134"   39 code points, 67 bytes
before: PHONE_NUMBER [55,67)   past the end for a code-point client
after:  PHONE_NUMBER [27,39)

"John Smith called yesterday."
before: PERSON, end_position 10, no start_position
after:  PERSON [0,10)
```

`PIIEntity` is only used inside `pkg/services`, so the pointer change does not reach other packages, and the JSON stays the same shape apart from the fix: both fields are still omitted when positions are not requested.

Four existing assertions move to the pointer form. Two of them are a real change, since `return_positions: false` used to assert `StartPos == 0` and now asserts `nil`, which is the distinction the pointers exist to make.

Three new tests: the multi-byte case, an ASCII case that has to stay unchanged, and an entity at offset 0.


Xunzhuo (changes requested):
The conversion and the offset-zero fix look correct. One blocker: this changes the public PII response contract, so please add an E2E case that calls `/api/v1/classify/pii` with non-ASCII text and an entity at offset 0, and asserts the returned code-point offsets. The Go unit coverage alone does not exercise the API boundary.

subin9:
Thank you! as you mentioned, added an E2E case, `pii-entity-offsets`, that calls `/api/v1/classify/pii` with multi-byte text and with an entity at offset 0, and asserts the returned positions by slicing the text as code points.

Xunzhuo (approved):
Thanks, this now covers the missing API boundary. The E2E checks non-ASCII code-point slicing and an entity at offset 0 through the real /api/v1/classify/pii endpoint; the conversion and pointer semantics still look correct. No code blocker from me. Merge after the remaining E2E and CI jobs are green.

wilsonwu (approved):

PR #3149, opened 2026-08-31
[Docs] Add Standalone HTTP Gateway Mode proposal (RFC)

Bevisy:
## Summary

Add a **separate, experimental standalone inference gateway** for vLLM
Semantic Router — and **do not add a second semantic orchestration**.

The gateway is an alternative entrypoint to the same routing semantics. It
does not reimplement model selection, policy, plugins, or response handling.
The plan is to make the existing ExtProc service and the new HTTP gateway
**two adapters over one shared semantic engine**:

- the **ExtProc adapter** keeps Envoy protobuf, stream phases, and mutation
  responses — the current deployment is untouched;
- the **HTTP adapter** handles public HTTP, SSE, client cancellation, and
  HTTP error mapping;
- the **shared engine** owns access, quota, plugins, signals, projections,
  decisions, algorithms, logical model selection, Looper, and response
  semantics;
- a **GatewayExecutor** owns the standalone path's physical backend
  execution only.

This is **not** a `--gateway` mode inside the existing binary, **not** HTTP
calls to a local ExtProc, and **not** a copy of the ExtProc orchestration.

### 1. The semantic orchestration must be extracted first

Semantic Router's request and response orchestration today lives inside
`pkg/extproc` (`req_filter_*`, `processor_*`, `RequestContext`): phase
ordering, error handling, Looper scope, and settlement. A gateway that only
re-assembles the transport-neutral leaf packages would create a **second
orchestration** — a second phase ordering, a second error path, a second
Looper scope, a second settlement path. Golden tests can detect drift
between two implementations; they cannot eliminate the duplication.

Extracting `pkg/extproc`'s orchestration into a transport-neutral
`semanticruntime.Engine` is therefore a **precondition** of the gateway,
not a follow-up. The current Envoy + ExtProc deployment, ports, protocols,
defaults, and external behavior stay unchanged throughout.

### 2. A Service cannot pick the physical backend

A Service placed in front of the gateway can load-balance gateway replicas,
but it cannot pick the physical backend for a request whose logical model is
already selected by the engine. Even as an experiment, the gateway must own
an explicit physical traffic layer: backend selection, health, retry,
credential pinning, and egress — it cannot forward everything to one shared
upstream and claim the traffic contract is verified.

### 3. One response pipeline

Two response paths exist today: the ExtProc path rebuilds response
semantics (plugins, cache, memory, replay, metrics, settlement), and
`pkg/looper/client.go` handles Looper provider responses. Neither is a
complete transport-neutral decoder; running both double-decodes, and running
only one could bypass response-side semantic policy. The proposal unifies
them into one neutral response pipeline (provider bytes decode exactly once,
semantic policy runs on the neutral object, client encoding last), with a
dedicated `pkg/backendinvoker` as an explicit P0 extraction target — not an
existing component.

## Architecture

```mermaid
flowchart LR
  existing_clients["existing clients"] --> Envoy
  Envoy --> ExtAdapter["ExtProc adapter"]
  gateway_clients["gateway clients"] --> HTTPAdapter["HTTP adapter"]

  ExtAdapter --> Engine["shared semantic engine"]
  HTTPAdapter --> Engine
  Engine --> Immediate["ImmediateAction"]
  Engine --> Dispatch["DispatchAction"]

  Immediate --> ExtAdapter
  Immediate --> HTTPAdapter
  Dispatch --> ExtAdapter
  ExtAdapter --> Envoy
  Envoy --> existing_backends["existing model backends"]

  HTTPAdapter -. "pinned at request start" .-> Scope["GatewayExecutionScope"]
  Dispatch --> Scope
  Scope --> Executor["GatewayExecutor"]
  Snapshot["BackendSnapshot"] --> Scope
  Executor --> Invoker["backend invoker"]
  Invoker --> gateway_backends["model backends"]
  gateway_backends --> Response["shared response pipeline"]
  Response --> HTTPAdapter

  Engine --> Looper
  Looper --> ModelPort["ModelExecutor interface"]
  ModelPort --> Scope
```

Two adapters over one engine. The ExtProc path keeps Envoy as its physical
transport. The gateway path pins an execution scope at request start and
resolves the logical dispatch plan against an immutable BackendSnapshot
through the backend invoker.

## Component ownership

| Component | Sole responsibility | Not responsible for |
| --- | --- | --- |
| shared engine | neutral request phases, semantic decisions, logical model, Looper, response phases, terminal state | Envoy protobuf, HTTP writers, backend addresses |
| ExtProc adapter | Envoy message ↔ neutral input/action mapping, current mutation contract | second orchestration, physical backend choice |
| HTTP adapter | public endpoints, HTTP/SSE, client disconnect, public errors | semantic policy, credentials, retry decisions |
| GatewayExecutor / ExecutionScope | pin backend/credential/deadline at request start; admission; physical plan; invoking the backend invoker | semantic model choice, attempt lifecycle |
| BackendSnapshot | one immutable revision of physical backends, capabilities, weights, health overlay | tenant/recipe/model policy |
| backend invoker | credential pinning, egress, backend picking, attempts, retry/fallback, codec, journal | entrypoint/decision computation |

Key contracts: `Engine.Begin(Ingress) → Session`; `Action` is a closed union
(`ImmediateAction` / `DispatchAction`); Looper runs through a request-scoped
`ModelExecutor` and never self-curls; logical dispatch plans never carry a
socket address; retry/fallback requires known-zero evidence, never a bare
status code.

## Non-goals (this proposal is not)

- A proxy engine replacing Envoy — the existing ExtProc deployment ships as
  it does today.
- A Kubernetes controller in this step — the gateway first lands as a
  Docker vertical slice.
- A rewrite of the routing language, recipes, or existing classifiers.
- Built-in identity provider features (JWT/OIDC federation maps onto the
  same tenant/grant runtime, not added as a separate authorization engine).

## Status

Experimental. The full design — contracts, lifecycle, traffic control,
security, observability, capability matrices, and the phased landing plan —
is in the proposal file in this PR:

[`website/docs/proposals/standalone-http-gateway.md`](https://github.com/vllm-project/semantic-router/pull/3149/changes)

No public listener ships until the shared engine extraction is proven
against the current path.

Related #3150


Xunzhuo (changes requested):
Thanks for writing the proposal first. I don't think this shape is ready to accept yet.

- `semantic-router --gateway` makes the Router accept public HTTP and forward upstream model traffic. That does change the responsibility boundary: ExtProc should return a logical selection; Envoy or an external gateway owns HTTP/streaming lifecycle and upstream dispatch. Please redesign this as a separate gateway adapter/process that can reuse the neutral codecs and routing contract, instead of a mode implemented inside `semantic-router` / `pkg/extproc`.
- “Same semantic behavior” is not true if this mode skips local classifier models and omits response-side plugins. Please define the supported capability matrix and explicit degradation behavior; routing behavior must not silently change by transport.

Once that boundary is clear, we can review the endpoint, SSE, readiness, and deployment details. I would keep #3150 unaccepted and unassigned until then.

Bevisy:
Thanks — both points are fair, and since this is still at the proposal stage, I've reworked the proposal rather than defend the current shape.

**1. Transport adapter, not a router mode.** `--gateway` is reframed as an optional HTTP transport adapter that reuses the existing pipeline — the neutral codecs, the routing contract, and the same buffered request handler — isolated from the routing core. Envoy + ExtProc stays the default and is unchanged. The adapter itself takes no serving-layer responsibility (no TLS termination, load balancing, retries, circuit breaking, multi-replica — those stay platform-owned, exactly as in the Envoy deployment where the router itself also does not provide them), consistent with the responsibility table in `production-stack-integration`.

**2. Capability matrix, not degradation.** The "same semantic behavior" claim is replaced by an explicit capability matrix. There is no capability degradation: routing semantics run the same code path in both modes. Transport capabilities that Envoy happens to provide today (LB, TLS, retries) are documented as platform-owned in *both* columns; local classification/embedding models are loaded per config in *both* modes — startup simply does not block on unconfigured models.

**3. Precedent for the single-process shape.** Kthena Router (`volcano-sh/kthena`) runs a default direct HTTP inference listener and optional Gateway API / Inference-Extension listeners on the same router process, without reimplementing Envoy — showing that single-process deployment and standard ingress can coexist in a Kubernetes-native inference router.

Updated proposal is being pushed to this PR (docs-only). #3150 stays unaccepted until the revised spec is reviewed — happy to adjust any matrix row whose boundary line reads wrong.

Xunzhuo (changes requested):
Thanks for the update. The matrix helps, but the main boundary blocker is still there: calling it an adapter does not change that the semantic-router process accepts public HTTP, owns SSE and upstream calls, and forwards model traffic. Our production-stack contract assigns that transport work to the gateway. Please make this a separate gateway binary/process that reuses the neutral codec and routing contract. Also, the proposal still says no degradation while response-side plugins are explicitly left for a follow-up; either include them in the contract or mark that capability unsupported instead of claiming parity.

Bevisy:
@Xunzhuo Thanks for the detailed feedback — both points addressed.

**1. Separate binary, not in-process mode.** `--gateway` is gone from the proposal. New shape: a standalone `cmd/gateway` binary that imports the router packages as a Go library — same link relationship the ExtProc binary has with those packages. The gateway owns HTTP/SSE/upstream; the router packages produce the decision only. The diagrams are updated so the router is control flow, not the upstream data path.

**2. Same semantic pipeline, no degradation.** The gateway is a new entry point to the same router capabilities, not a reduced one: responses pass through the same response-side semantic pipeline the ExtProc path runs (res_filter — hallucination mitigation, response logging and caching), via the linked packages. SSE relay is in scope to match the acceptance criteria. Only Envoy-owned transport features (TLS, LB, retries, circuit-breaking) stay platform-owned.

Xunzhuo:
Thanks for the rewrite. The separate gateway process, linked routing packages, response-side `res_filter` parity, and explicit capability matrix address the earlier architecture blockers. #3150 still needs maintainer acceptance before this proposal can be approved, so I am leaving the review pending that decision.

Xunzhuo (approved):
Thanks for the thorough rewrite. The separate gateway process, shared routing packages and response-side pipeline, and explicit platform-owned transport capabilities resolve the architecture concerns. This proposal is scoped and ready to land.

wilsonwu (dismissed):
LGTM

Xunzhuo (changes requested):
/hold sorry i need a second review for this proposal, it is a big change and required a lot of work

FAUST-BENCHOU (commented):
not sure whether we need this

Bevisy (commented):
Duplicate of the referenced comment above.

Bevisy:
@Xunzhuo Thanks for the review — the proposal has been updated and merged with the design direction from your [second review](https://github.com/vllm-project/semantic-router/pull/3149#pullrequestreview-5066211663): shared engine first, two adapters, gateway-owned traffic layer.

Xunzhuo (changes requested):
Thanks for consolidating the design around one shared engine. Two blockers remain on this head:

- The current-state section says `BackendInvoker` already decodes provider responses, but that component does not exist in the repository today. Provider-response handling is still split between the ExtProc path and `pkg/looper/client.go`. Please describe those existing paths accurately and make `pkg/backendinvoker` an explicit P0 extraction/creation target.
- The required quality check fails codespell at `standalone-http-gateway.md:160` on `ClientA`; please rename or allowlist that diagram label.

Once those are corrected, the single-decode migration and dependency plan will be reviewable against the implementation.

Bevisy:
Thanks for the review — both blockers are fixed on the head.

1. **Current-state response paths.** The Problem section and the PR description no longer cite BackendInvoker as existing. They now describe the two existing paths accurately — the ExtProc path (plugins, cache, memory, replay, metrics, settlement) and `pkg/looper/client.go` for Looper provider responses — and state that a dedicated `pkg/backendinvoker` (single authority for attempts, credentials, terminal evidence) is a P0 extraction/creation target from those existing paths, not an existing component.

2. **codespell.** The mermaid diagram node ids in `standalone-http-gateway.md` were renamed (ClientA/ClientB/ModelsA/ModelsB → lowercase underscore ids) so codespell passes.

The single-decode migration and dependency plan are now described against the real implementation. Please take another look.

Xunzhuo (commented):
Thanks for the update — both blockers are fixed on this head. The current-state response paths now match the code, and codespell passes. #3150 is accepted; the proposal still needs the requested second maintainer review, so I am leaving it unapproved.

Bevisy:
> Thanks for the update — both blockers are fixed on this head. The current-state response paths now match the code, and codespell passes. #3150 is accepted; the proposal still needs the requested second maintainer review, so I am leaving it unapproved.

@wilsonwu please take another look at this RFC.

drivebyer (commented):
Compared this against the design gist; it is faithful. Comments are on the proposal itself.

**1. Several "existing" components do not exist on `main`.**
Semantic publication, `CredentialPublication`, access runtime, quota partition, usage/cost settlement, and the attempt journal (GW-002, GW-011, GW-022, D4) come from #2959, which was closed unmerged. `main` has `pkg/authz`, `pkg/ratelimit` (no Valkey backend), and `pkg/admission`. Either list #2959 as a prerequisite, or scope P0 to today's contracts and move the publication model to P1.

**2. #3150 contradicts the RFC.**
It keeps the buffered pipeline and leaves LB/retry to the deployment layer; the RFC requires the engine extraction first and gives the gateway its own LB/retry/circuit. Rewrite the issue before it serves as the completion signal.

**3. P0-A is unsized.**
It rewrites the request/response core of `pkg/extproc` (~137 files) with golden parity, in serial steps, alongside in-flight PRs on the same files. Name an owner and a rough PR count.

**Overall.**
For a change of this size I would prefer an RFC that stays above implementation detail, then land the work through small PRs. That keeps conflicts with concurrent work in the repo manageable, surfaces problems early, and avoids large shifts in codebase behavior in any single step.

Bevisy:
> Compared this against the design gist; it is faithful. Comments are on the proposal itself.
>
> **1. Several "existing" components do not exist on `main`.**
> Semantic publication, `CredentialPublication`, access runtime, quota partition, usage/cost settlement, and the attempt journal (GW-002, GW-011, GW-022, D4) come from #2959, which was closed unmerged. `main` has `pkg/authz`, `pkg/ratelimit` (no Valkey backend), and `pkg/admission`. Either list #2959 as a prerequisite, or scope P0 to today's contracts and move the publication model to P1.
>
> **2. #3150 contradicts the RFC.**
> It keeps the buffered pipeline and leaves LB/retry to the deployment layer; the RFC requires the engine extraction first and gives the gateway its own LB/retry/circuit. Rewrite the issue before it serves as the completion signal.
>
> **3. P0-A is unsized.**
> It rewrites the request/response core of `pkg/extproc` (~137 files) with golden parity, in serial steps, alongside in-flight PRs on the same files. Name an owner and a rough PR count.
>
> **Overall.**
> For a change of this size I would prefer an RFC that stays above implementation detail, then land the work through small PRs. That keeps conflicts with concurrent work in the repo manageable, surfaces problems early, and avoids large shifts in codebase behavior in any single step.

All three points are fair, and the proposal now reflects them:

**1. Existing components** — correct, and I took the second option you offered: scope P0 to today's contracts and move the publication model to P1. The proposal now states semantic publication, `CredentialPublication`, access runtime, quota partition, settlement, and the attempt journal as **P1 target contracts**, not existing components (Problem, Goals, Architecture decisions, capability matrices, and the landing plan all updated), and anchors P0 to `pkg/authz` / `pkg/ratelimit` / `pkg/admission` plus the current classifiers, protocol codec, and Looper.

**2. #3150** — done, the issue now matches the RFC: shared-engine-first, a gateway-owned physical traffic layer (BackendSnapshot + selection/retry/credential pinning), staged rollout, and a completion signal that includes the engine extraction and logical→physical mapping — not just "three endpoints served".

**3. P0-A scale** — agreed the scale isn't pinned down yet; characterizing current behavior with golden fixtures and defining the minimal neutral contract is the first implementation focus, and we'll measure how much orchestration actually moves before writing any PR-by-PR plan. I'm willing to own the implementation and the PR-by-PR plan, and I'll keep the workgroup in the loop along the way so we can course-correct early.

And on the **overall** point — I completely agree, and it's how I want to work this: keep the RFC above the implementation details, then land it in small, reviewable PRs so nothing big shifts the codebase in one step. Appreciate you spelling it out.

Xunzhuo (commented):
The current proposal addresses the second-review concerns: it scopes the unmerged publication, quota, and settlement pieces as P1 contracts, sizes P0-A and keeps the extraction in small serial PRs, and aligns #3150 with the engine-first, gateway-owned traffic boundary. The branch is now behind `main`; please update it and let the resulting checks complete so the hold can be cleared.

Bevisy:
> The current proposal addresses the second-review concerns: it scopes the unmerged publication, quota, and settlement pieces as P1 contracts, sizes P0-A and keeps the extraction in small serial PRs, and aligns #3150 with the engine-first, gateway-owned traffic boundary. The branch is now behind `main`; please update it and let the resulting checks complete so the hold can be cleared.

Done. PTAL

Bevisy:
@Xunzhuo @drivebyer PTAL

Xunzhuo (approved):

Xunzhuo:
Thanks, we should have a PoC later

PR #3155, opened 2026-08-31
[Feature] Support configurable classifier unknown outcomes

guan404ming:
Related #3106

## Purpose

- Add `rules.on_unknown` with `no_match`, `match`, and `fail_request`.
- Preserve existing behavior when `on_unknown` is omitted.
- Update config surfaces, diagnostics, documentation, and E2E coverage.

## Notes

- `createErrorResponse` now types every 5xx as `api_error`, which also retypes the 24 pre-existing 5xx call sites in `pkg/extproc` (9x500, 9x502, 6x503).

## Test Plan

- `make agent-ci-gate ENV=cpu`

## Test Result

- Passed.


theohsiung (commented):
Reviewed at `c88229f1`. The rework since `dfdb6985` already fixed the 503 mapping, the trace/legacy mismatch, the error-swallowing `EvaluateDecisionsWithTrace`, the phantom applied-policy entries, and the stale OLM bundle — nice. Everything below is reproduced against `c88229f1`; repro snippets at the end.

### 1. `fail_request` discards a clean, higher-priority match

The loop now `continue`s, but still `return output, unresolvedErr` at the end, so `results` is dropped.

```
[sanity] good alone          -> decision=good err=<nil>
[guarded first] good+guarded -> decision=<nil> err=... decision unresolved
[good first   ] good+guarded -> decision=<nil> err=... decision unresolved
```

`good` is priority 100 and never touches the failed classifier; `guarded` is priority 1 with `on_unknown: fail_request`. So one guarded decision 503s all traffic when its classifier is down, regardless of priority. That may be the intended global fail-closed, but nothing states it: `no_match`/`match` are per-decision while `fail_request` is global, and the three sit in the same enum. Worth a doc sentence + a test pinning the semantics.

### 2. Root `on_unknown` silently disables every leaf `on_error` — and the shipped example sets both

`resolveUnknown` only takes the legacy branch when `policy == ""`, and `evaluatePredicateLeaf` only reads `node.OnError` under `legacy`:

```
on_unknown=""         on_error="no_match" -> matched=false
on_unknown=""         on_error="match"    -> matched=true    <- on_error decides
on_unknown="no_match" on_error="no_match" -> matched=false
on_unknown="no_match" on_error="match"    -> matched=false    <- on_error is dead
```

`config/config.yaml` sets `on_unknown: no_match` (L1281) and `on_error: no_match` (L1288) on the same tree. Both values agree today so it looks harmless, but someone copying it and flipping the leaf to `on_error: match` to fail closed gets no behavior change, no startup error, no log. Suggest: drop one from the example, and have `validateDecisionRuleNode` reject/warn when both appear in one tree. The docs say what happens *when `on_unknown` is omitted*; the inverse ("setting it disables every `on_error` below") is never stated.

### 3. `llm_decision_match_total` counts decisions on requests that then 503

Because the loop no longer returns early, every matching decision is now counted before the error surfaces:

```
outcome: decision=<nil> err=... decision unresolved   (client gets 503)
llm_decision_match_total{decision="p7-good"}: 0 -> 1
```

Match-rate dashboards over-report during a classifier outage. Moving `RecordDecisionMatch` after the `unresolvedErr` check would fix it.

### 4. `fail_request` diagnostics never reach the replay store

`ctx.VSRAppliedUnknownPolicies` is set correctly, but `runRequestPreRoutingStages` returns the error response before `startRouterReplay` — which is only called on the success paths.

```
envoy status: ServiceUnavailable
ctx.VSRAppliedUnknownPolicies = map[guarded:fail_request]   <- populated
ctx.RouterReplayID            = ""                          <- never persisted
```

`no_match`/`match`/legacy all persist; only `fail_request` doesn't — the one case #3106's completion signal most needs ("record both the backend error and the applied terminal policy in diagnostics").

### 5. The eval-trace fix swallows the error instead

`classification_signal_contract.go:58` is now `if err != nil && len(traces) == 0 { return nil, err }`. With `?trace=true` and a `fail_request` decision, `traces` is non-empty, so the error is dropped entirely and the endpoint returns **200** with `eval_trace` set and `decision_result` nil — it looks like success. Returning the error *with* the traces attached would give both.

### 6. The 503 body still says `invalid_request_error`

```json
{"error":{"code":503,
          "message":"decision evaluation failed: decision \"guarded\" could not be resolved because a signal evaluator failed: decision unresolved",
          "type":"invalid_request_error"}}
```

`createErrorResponse` hardcodes the type, so a backend outage is still labelled a client error, and the internal decision name is echoed to the caller.

### Smaller

- **DSL has no validation for `on_unknown`.** `ROUTE "x" (on_unknown = "allow")` passes `Validate`, `Compile`, and `Decompile` with nothing mentioning it; `Rules.OnUnknown = "allow"` is written verbatim. It *is* caught by the config validator at router startup, so this is a feedback-latency gap rather than a runtime bug — but the YAML path fails at load and the DSL/playground path doesn't fail until deploy.
- **`cli/evaluation/routing_trace.py` reads none of the new fields.** `_trace_node` drops `state`/`signal_error`, `_decision_traces` drops `state`/`on_unknown`, `_signals` never reads `applied_unknown_policies`. The normalized records aren't identical (confidence differs), but nothing shows *why* a decision matched — which is what `classifier.md`'s new "diagnostics include both the signal error and any terminal policy that was applied" promises.
- **`Condition.on_unknown` is a field that can never hold a value**, and its `raise` is the first statement in `validate_node_shape`, so it masks real structural errors: `NOT` with two children plus `on_unknown` reports only the `on_unknown` problem, hiding "NOT operator must have exactly one child condition". It also puts `on_unknown: None` into every nested condition's `model_dump()`.
- **`safety-models-and-policy.md` drops 25 lines that are still accurate**, including the #2918 gotcha (failure reports at confidence 1.0 with `classification_error`, so `block` only closes a request when a decision actually consumes the jailbreak signal — "without one it looks like a no-op"), the "applies to local and remote backends, request and response side, including `method: contrastive`" scope, and the `jailbreak-onerror` example pointer. The `on_error: block` YAML sample is still there but the surrounding text no longer explains `allow`/`block`. Adding to that section rather than replacing it would keep both.

<details><summary>Repro</summary>

```go
// pkg/decision — findings 1, 2, 3
guarded := config.Decision{Name: "guarded", Priority: 1, Rules: config.RuleNode{
    Type: config.SignalTypeClassifier, Name: "risk", Label: "RISKY",
    Predicate: &config.NumericPredicate{GTE: float64Ptr(0.5)},
    OnUnknown: config.RuleOnUnknownFailRequest}}
good := config.Decision{Name: "good", Priority: 100,
    Rules: config.RuleNode{Type: config.SignalTypeKeyword, Name: "present"}}
signals := &SignalMatches{KeywordRules: []string{"present"},
    SignalErrors: map[string]string{"classifier:risk": "timeout"}}

// good alone -> matches. good+guarded -> nil + error, and
// llm_decision_match_total{decision="good"} was already incremented.
NewDecisionEngine(nil, nil, nil, []config.Decision{good, guarded},
    config.RoutingStrategyPriority).EvaluateDecisionsWithSignals(signals)

// finding 2: flip OnError between "match"/"no_match" with and without
// OnUnknown: "no_match" on the same leaf — it only changes the outcome when
// OnUnknown is empty.
```

```go
// pkg/extproc — findings 4, 6 (config: `good` priority 100 + `guarded`
// priority 1 with on_unknown: fail_request, default recipe selected)
result, fallback, err := router.runDecisionEngine(
    config.DefaultVSRAutoModelName, ctx, signals, cfg.Decisions)
// -> nil, "", ErrDecisionUnresolved
// ctx.VSRAppliedUnknownPolicies = map[guarded:fail_request]
// ctx.RouterReplayID            = ""
```

```go
// pkg/dsl — DSL gap
src := `SIGNAL keyword "hack" { operator: "contains" values: ["hack"] }
ROUTE "x" (on_unknown = "allow") { PRIORITY 100 WHEN keyword("hack") MODEL "m" }`
Validate(src) // no diagnostic mentions on_unknown
cfg, errs := Compile(src) // errs empty; cfg.Decisions[0].Rules.OnUnknown == "allow"
config.ValidateKubernetesConfigContracts(...) // only here is it rejected
```

```python
# cli — Condition ordering
bad = dict(operator="NOT", conditions=[Condition(type="keyword", name="a"),
                                       Condition(type="keyword", name="b")])
Condition(**bad)                      # "NOT operator must have exactly one child condition"
Condition(**bad, on_unknown="match")  # "on_unknown is only valid on the root rules node"
```

</details>

guan404ming:
Thanks for the thorough review. All findings reproduced and are fixed in 65d07e61.

1. Kept the global fail-closed semantics for `fail_request`, now pinned by a test and documented in `classifier.md`.
2. Validation now rejects `rules.on_unknown` combined with any condition-level `on_error`; dropped `on_error` from the shipped examples.
3. `RecordDecisionMatch` moved after the unresolved check.
4. The `fail_request` 503 path now persists the replay record and returns the replay header.
5. The eval endpoint returns 503 with both `eval_trace` and a new `decision_error` field.
6. 5xx error bodies are typed `api_error`.

Smaller: DSL `Validate` now flags invalid `on_unknown` values, `routing_trace.py` reads `state` / `signal_error` / `on_unknown` / `applied_unknown_policies`, the `Condition.on_unknown` check runs after structural validation, and the dropped `safety-models-and-policy.md` content is restored.

Not changed: the 503 message still echoes the decision name. Happy to redact it if you prefer.

Xunzhuo (changes requested):
Thanks for the substantial follow-through on #3106. One blocker: `fail_request` changes API behavior based on `?trace=true`. `ClassifyIntentForEval` returns a structured response only when traces exist; without trace it returns nil and maps the same unresolved decision to a generic 500. The ordinary classify endpoint also maps `ErrDecisionUnresolved` to 500, while ExtProc uses 503. Please make `fail_request` return consistent 503 diagnostics independent of trace, with coverage for both eval modes and ordinary classify.

guan404ming:
Fixed in 0d1dcb3e. `fail_request` now returns 503 with structured diagnostics regardless of `?trace=true`:

- `ClassifyIntentForEval` always builds the response on a decision error (`decision_error`, `signal_errors`, `applied_unknown_policies`; `eval_trace` when tracing), so both eval modes return the same 503 body shape.
- `writeClassificationError` maps `ErrDecisionUnresolved` to `503 DECISION_UNRESOLVED`, so ordinary classify matches ExtProc instead of a generic 500.

Coverage: a service-level test runs both eval modes against a real classifier with a failing backend, plus handler tests for the eval and ordinary classify 503 paths.

theohsiung (dismissed):

guan404ming:
Just pushed to handle the ci error

Xunzhuo (dismissed):
Thanks for the thorough follow-through. I rechecked the current head: `fail_request` now returns the same 503 contract for ordinary classify and eval with or without tracing, while preserving structured diagnostics. The focused decision/config tests pass locally and the current CI is green. This is ready from my side.

theohsiung (commented):
(superseded — see inline notes below)

src/semantic-router/pkg/decision/trace.go
```diff
@@ -85,18 +53,40 @@ func (e *DecisionEngine) evalDecisionWithTrace(
...
+func newDecisionTrace(
+	decision *config.Decision,
+	evaluation nodeEvaluation,
+	originalState evaluationState,
+	policy string,
+	trace *TraceNode,
+) DecisionTrace {
+	return DecisionTrace{
+		DecisionName: decision.Name,
+		State:        originalState.String(),
```
theohsiung, line 73: `State` is stamped from the pre-resolution evaluation while `Matched`/`RootTrace` are post-resolution, so the legacy path emits `"state":"unknown"` next to `"matched":true` and a root trace whose own state is `"true"` — a resolved decision reported as unresolved, which is the distinction #3106 exists to make. `TestUnknownLegacyTraceShowsResolvedLeaf` asserts `Matched` and `RootTrace` but never `State`.

src/semantic-router/pkg/dsl/policy_signal_roundtrip_test.go
```diff
@@ -23,7 +23,7 @@ SIGNAL classifier "risk" {
   use_cpu: true
 }
 
-ROUTE "policy-route" {
+ROUTE "policy-route" (on_unknown = "fail_request") {
```
theohsiung, line 26: This route keeps `on_error: "match"` on its classifier leaf, and `assertPolicyDSLSource` asserts both keys in the decompiled output — but `validateDecisionRuleNode` rejects that pair (`condition on_error has no effect when rules.on_unknown is set`). The fixture pins a config the router won't load; worth unpinning before it gets copied.

src/semantic-router/pkg/dsl/validator.go
```diff
@@ -523,6 +523,15 @@ func (v *Validator) checkRouteConstraints(r *RouteDecl) {
 		)
 	}
 
+	switch r.OnUnknown {
+	case "", config.RuleOnUnknownNoMatch, config.RuleOnUnknownMatch, config.RuleOnUnknownFailRequest:
+	default:
+		v.addDiag(DiagConstraint, r.Pos,
```
theohsiung, line 529: Enum check only. The `on_unknown` + condition `on_error` conflict the config loader enforces isn't checked here, so the DSL compiles clean and the router fails at startup.

src/vllm-sr/cli/models.py
```diff
@@ -661,6 +668,7 @@ class Rules(BaseModel):
 
     operator: str = "AND"
     conditions: List[Condition] = Field(default_factory=list)
+    on_unknown: Optional[Literal["no_match", "match", "fail_request"]] = None
```
theohsiung, line 671: Same gap as `dsl.Validate` — no cross-field check against a nested condition's `on_error`, so `vllm-sr` validates green on a config the Go loader rejects. The dashboard's new On Unknown select has it too.

src/vllm-sr/cli/evaluation/routing_trace.py
```diff
@@ -140,6 +158,7 @@ def normalize_routing_diagnostic(
         routing_decision=_safe_token(payload.get("routing_decision")),
         traces=_decision_traces(payload),
         signals=_signals(payload),
+        applied_unknown_policies=_applied_unknown_policies(payload),
```
theohsiung, line 161: `decision_error` isn't picked up, so the artifact records that a policy fired but not why. Related: nothing can read the new 503 yet — `vllm-sr eval` raises on any non-200 and `_format_error_response` only knows `{"error": {...}}`, and `tools/agent/scripts/tuning/client.py:36` has a bare `urlopen` that throws `HTTPError`.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -57,6 +58,16 @@ func (r *OpenAIRouter) runRequestPreRoutingStages(
 			return requestDecisionState{}, r.createErrorResponse(422, decisionErr.Error())
 		}
 		logging.Errorf("[Request Body] Decision evaluation failed: %v", decisionErr)
+		if errors.Is(decisionErr, decision.ErrDecisionUnresolved) {
+			resp := r.createErrorResponse(503, decisionErr.Error())
+			if ctx.RouterReplayPluginConfig == nil {
+				ctx.RouterReplayPluginConfig = r.Config.EffectiveRouterReplayConfig(nil)
```
theohsiung, line 64: No `finalizeRouterReplay` here, so the record settles as `aborted / stream_ended_before_terminal_response` rather than `failed` — the looper-failure path finalizes explicitly to avoid that. Nothing pins this path either: record, header and `VSRAppliedUnknownPolicies` are all correct today, just unguarded.

src/semantic-router/pkg/extproc/router.go
```diff
@@ -186,10 +186,14 @@ func (r *OpenAIRouter) createJSONResponse(statusCode int, data interface{}) *ext
 
 // createErrorResponse creates a direct error response.
 func (r *OpenAIRouter) createErrorResponse(statusCode int, message string) *ext_proc.ProcessingResponse {
+	errorType := "invalid_request_error"
+	if statusCode >= 500 {
+		errorType = "api_error"
```
theohsiung, line 191: This retypes 24 pre-existing 5xx call sites in `pkg/extproc` (9x500, 9x502, 6x503), not just the new 503. Looks correct to me — just worth a line in the PR body.

theohsiung (commented):
Re-reviewed at `7c0dcc64`. All ten earlier findings + @Xunzhuo's blocker verified by running them, thanks for the follow-through. . Not gating — notes inline. The first two (trace `State`, the DSL fixture) are worth landing here; the rest are fine as follow-ups.

src/semantic-router/pkg/decision/trace.go
guan404ming: Good catch, thanks. Fixed in 6eab3516: the legacy path now syncs `State` to the resolved state, and the legacy test asserts it. Policy paths keep the pre-resolution state on purpose, since `on_unknown` records the resolution (pinned by `TestUnknownTraceRecordsErrorAndPolicy`).

src/semantic-router/pkg/dsl/policy_signal_roundtrip_test.go
guan404ming: Fixed in 6eab3516: dropped `on_error: "match"` from the leaf. Its round-trip stays covered by the recipes test.

src/semantic-router/pkg/dsl/validator.go
guan404ming: Agreed, thanks. I will follow-up issue to add the `on_unknown` + `on_error` conflict check to `dsl.Validate`.

src/vllm-sr/cli/models.py
guan404ming: Agreed. The `dsl.Validate` follow-up issue will also cover the CLI model and the dashboard select.

src/vllm-sr/cli/evaluation/routing_trace.py
guan404ming: Thanks for tracing all the consumers. I will file a follow-up issue for them: `decision_error` here, `vllm-sr eval` error handling, and the bare `urlopen` in `tuning/client.py`.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
guan404ming: Nice find. I will file a follow-up issue to finalize the replay as `failed` here and pin the 503 path with an extproc test.

src/semantic-router/pkg/extproc/router.go
guan404ming: Thanks for checking the call sites. This was intentional; I will note the 24 retyped 5xx sites in the PR body.

theohsiung (approved):

guan404ming:
Thx!

PR #3161, opened 2026-08-31
[Test] Accept PathLike inputs in Podman socket test

### LINKED ISSUE #3157: [Bug] CLI pytest suite crashes under Python 3.14 in Podman socket test
### Proposed Workgroup

### Describe the bug

The vLLM-SR CLI pytest suite crashes with a pytest INTERNALERROR on Python 3.14 while running `test_container_start_vllm_sr_warns_when_podman_socket_missing`.

The test monkeypatches `os.path.exists` with a helper that assumes its argument is a string:

```python
def fake_exists(path: str) -> bool:
    if "podman.sock" in path or "docker.sock" in path:
        return False
```

On Python 3.14, pytest/pathlib calls `os.path.exists` with a `PosixPath`. The membership check raises `TypeError`, and because the global helper is still patched, pytest also fails while formatting the error and cleaning temporary directories.

### To Reproduce

Environment:

- macOS
- Python 3.14.6
- pytest 9.1.1
- current `main` at `8a0e5faa2`

Run:

```bash
cd src/vllm-sr
../../.venv-agent/bin/python -m pytest tests/test_openclaw_podman_socket.py -q
```

Observed:

```text
tests/test_openclaw_podman_socket.py ..
TypeError: argument of type 'PosixPath' is not a container or iterable
INTERNALERROR> File "tests/test_openclaw_podman_socket.py", line 135, in fake_exists
```

The same failure aborts the full `src/vllm-sr/tests` suite at roughly 52%.

### Expected behavior

The three Podman socket tests and the full CLI pytest suite should complete normally on every Python version allowed by `src/vllm-sr/pyproject.toml` (`requires-python = ">=3.10"`).

### Impact

Contributors using Python 3.14 cannot complete the CLI unit-test suite, and the failure corrupts pytest reporting/cleanup rather than appearing as a normal failed assertion.

A temporary workaround is to use Python 3.11 or 3.12.

### Additional context

A minimal fix is to normalize the monkeypatched argument with `os.fspath(path)` before substring checks. This is test-only and does not change runtime behavior.

wuli666:
Closes #3157

## Purpose

- Accept both string and `os.PathLike` inputs in the Podman socket test double.
- Normalize paths with `os.fspath()` before checking socket names.
- Add a direct `Path` probe so the Python 3.14 regression stays covered.
- Owned by `wg/developer-experience-ecosystem`.

## Test Plan

- `cd src/vllm-sr && ../../.venv-agent/bin/python -m pytest tests/test_openclaw_podman_socket.py -q`
- `make agent-ci-gate ENV=cpu CHANGED_FILES="src/vllm-sr/tests/test_openclaw_podman_socket.py"`
- Build and start the local CPU Docker stack, then run the smoke-equivalent service, container, dashboard, and startup-log checks.

## Test Result

- Focused Python 3.14 test: 3 passed.
- Agent CI gate passed; the CLI suite ran 39 tests with 9 skips and 0 failures.
- CPU Docker images built successfully; Router, Envoy, Dashboard, and Fleet Sim all reported Running, and the equivalent local smoke checks passed.
- The API-only Docker routing integration passed. Model-loading integration cases exceeded the 5-minute startup window while downloading mmBERT ONNX assets; the canonical smoke helper also still matches the previous CLI status wording. Both are unrelated existing harness/environment blockers.

### DIFF
--- a/src/vllm-sr/tests/test_openclaw_podman_socket.py
+++ b/src/vllm-sr/tests/test_openclaw_podman_socket.py
@@ -131,11 +131,13 @@ def test_container_start_vllm_sr_warns_when_podman_socket_missing(
     monkeypatch.delenv("VLLM_SR_CONTAINER_SOCKET", raising=False)
     real_exists = os.path.exists
 
-    def fake_exists(path: str) -> bool:
-        if "podman.sock" in path or "docker.sock" in path:
+    def fake_exists(path: str | os.PathLike[str]) -> bool:
+        path_text = os.fspath(path)
+        if "podman.sock" in path_text or "docker.sock" in path_text:
             return False
         return real_exists(path)
 
+    assert fake_exists(config_path)
     monkeypatch.setattr(os.path, "exists", fake_exists)
     _stub_runtime_images(monkeypatch)
     captured = _capture_run_commands(monkeypatch)


Xunzhuo (approved):
Thanks, this is nicely scoped. `os.fspath` handles both `str` and `PathLike`, the focused Python 3.13 run passes all three cases, and current CI is green.

wilsonwu (approved):
LGTM, thanks!

PR #3162, opened 2026-08-31
[Feature] Treat semantic-cache NLI verifier failures as cache misses

### LINKED ISSUE #3176: [Feature] Define the runtime failure policy for semantic-cache NLI verification
### Goal

Define the runtime failure contract for the optional in-memory semantic-cache NLI verifier.

Parent Epic: #3036  
Related implementation: #3162  
Earlier fail-open contract: #3075

### Maintainer decision

When the NLI verifier is unavailable or returns an error, treat the candidate as a cache miss and continue the normal upstream model request.

This is fail-closed only for the unverified cache candidate. The user request remains available through the normal data-plane path.

### Proposed Workgroup

Data Plane & Networking

### Scope

For the in-memory semantic cache in nli or lexical+nli mode:

- do not serve the cached response body after a verifier error;
- continue through the ordinary cache-miss path to the upstream model;
- do not count the candidate as a cache hit or avoided inference;
- retain bounded candidate similarity, verifier-error, rejection-reason, and latency diagnostics;
- keep request cancellation as cancellation rather than converting it to verifier failure or cache miss;
- update tests and documentation to reflect the selected contract.

### Non-goals

- Change normal NLI pass or contradiction behavior.
- Change startup or configuration validation.
- Change the lexical polarity guard.
- Change remote or hybrid cache backends.
- Add verifier retries, a circuit breaker, or a configurable failure policy.

### Completion signal

- [x] #3162 implements cache-miss behavior for verifier errors.
- [x] No unverified cached body is served or counted as a hit.
- [x] The ordinary upstream request path continues with bounded diagnostics.
- [x] Cancellation and normal NLI outcomes remain unchanged.
- [ ] Unit and affected E2E coverage pin the failure contract.

luca-888:
Closes #3176
Related #2751
Follow-up to #3075

## Purpose

When semantic-cache NLI verification is unavailable or fails, return a cache miss and continue to the upstream model instead of serving an unverified cached response. Preserve the candidate similarity and propagate request cancellation without counting it as a cache hit or miss.

The implementation uses each cache's `c.polarityGuard.Verifier`, preserving the ownership and lifecycle on current main. Regression coverage exercises the real cache, adapter, service, and ExtProc path, verifier recovery, cancellation before/during verification, and isolation between cache instances. E2E similarity validation accepts a miss at similarity 1.0 because even a full-similarity candidate can fail verification. The documentation explains the correctness-over-latency trade-off.

## Test Plan

After building the CPU native libraries with `make rust-ci` and exporting their build/runtime library paths (Go package commands run from `src/semantic-router/`; external storage-service tests disabled via the repository's `SKIP_*_TESTS` flags):

- `CI=true make check BASE_REF=origin/main CHANGED_FILES="$(git diff --name-only origin/main...HEAD)"`
- `go test ./pkg/cache ./pkg/extproc ./pkg/classification -count=1`
- `go test ./pkg/modelruntime -run 'TestGlobalCacheNLI' -count=1`
- `go test -race ./pkg/cache -run 'TestPolarityNLIGuard|TestValidateCacheConfigPolarity' -count=1`
- `go test -race ./pkg/extproc -run '^TestPolarity' -count=1`
- From `e2e/`: `go test ./testcases -run 'Test.*Cache|Test.*Polarity' -count=1`
- `make verify PROFILE=envoy-ai-gateway`

## Test Result

- `make check` passed, including changed-file lint/pre-commit, structure and dependency checks, documentation/configuration checks, router build and full router Go tests, E2E build/coverage drift checks, and all 92 E2E framework unit tests.
- The cache, ExtProc, and classification package suites passed. Storage integration cases requiring external services and model-gated cases without local artifacts were skipped using the repository's supported test gates.
- Global cache NLI ownership regressions, both race runs, and the cache/polarity E2E assertion tests passed.
- Deployment E2E was attempted after installing Kind, kubectl, and Helm, but could not create the cluster: Docker Hub requests for the pinned `kindest/node:v1.33.7` image repeatedly timed out or were refused. No deployment E2E pass is claimed; the updated PR's GitHub Actions run must supply that result.

### DIFF
--- a/src/semantic-router/pkg/cache/inmemory_cache_polarity.go
+++ b/src/semantic-router/pkg/cache/inmemory_cache_polarity.go
@@ -15,16 +15,17 @@ import (
 type polarityNLIVerdict struct {
 	Reject        bool
 	Contradiction float32
-	// Skipped means the verifier was unavailable or failed; the caller fails
-	// open and serves the candidate.
+	// Skipped means the verifier was unavailable or failed; the caller treats
+	// the unverified semantic candidate as a cache miss.
 	Skipped bool
 }
 
 // applyPolarityNLI runs the NLI tier on the winning candidate when the tier is
 // enabled. It returns handled=true with the lookup outcome when the candidate
-// must not be served — the request was cancelled, or the queries contradict —
-// and handled=false when the caller should serve it. A rejection is a
-// caller-visible miss that still reports the rejected score
+// must not be served: the request was cancelled, the verifier could not verify
+// the candidate, or the queries contradict. It returns handled=false when the
+// caller should serve the candidate. An unverified or rejected candidate is a
+// caller-visible miss that still reports the candidate score
 // (LookupResult.Similarity) so near-threshold rejections stay diagnosable.
 func (c *InMemoryCache) applyPolarityNLI(
 	ctx context.Context,
@@ -41,6 +42,14 @@ func (c *InMemoryCache) applyPolarityNLI(
 		return LookupResult{}, true, err
 	}
 	verdict := c.verifyPolarityNLI(ctx, model, bestEntry.Query, query)
+	if err := ctxErr(ctx); err != nil {
+		metrics.RecordCacheOperation("memory", "find_similar", "canceled", time.Since(start).Seconds())
+		return LookupResult{}, true, err
+	}
+	if verdict.Skipped {
+		c.recordPolaritySkippedMiss(start, bestSimilarity)
+		return LookupResult{Similarity: bestSimilarity}, true, nil
+	}
 	if !verdict.Reject {
 		return LookupResult{}, false, nil
 	}
@@ -48,9 +57,9 @@ func (c *InMemoryCache) applyPolarityNLI(
 	return LookupResult{Similarity: bestSimilarity}, true, nil
 }
 
-// verifyPolarityNLI runs the NLI tier on the winning candidate. It fails open:
-// a missing or failing verifier serves the threshold-verified hit and records
-// the skip, so a model hiccup never turns into a cache outage.
+// verifyPolarityNLI runs the NLI tier on the winning candidate. A missing or
+// failing verifier records a skip; the caller degrades that unverified semantic
+// hit to a cache miss so the request can continue to the model backend.
 func (c *InMemoryCache) verifyPolarityNLI(ctx context.Context, model, cachedQuery, incomingQuery string) polarityNLIVerdict {
 	start := time.Now()
 	verifier := loadPolarityVerifier()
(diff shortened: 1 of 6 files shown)


Xunzhuo (changes requested):
Thanks for the clear tests. One blocker: accepted #2751 and merged #3075 explicitly define NLI verifier failure as fail-open, serving the cache hit to preserve availability. This PR reverses that contract to a cache miss without a focused accepted issue changing the policy. Please open or link that accepted contract change first, then we can review the implementation against it.

luca-888:
> Thanks for the clear tests. One blocker: accepted #2751 and merged #3075 explicitly define NLI verifier failure as fail-open, serving the cache hit to preserve availability. This PR reverses that contract to a cache miss without a focused accepted issue changing the policy. Please open or link that accepted contract change first, then we can review the implementation against it.

Thanks for the clarification. I agree that #3075 deliberately established fail-open behavior, so changing it requires an explicit contract decision.

I opened #3176 to discuss and define the runtime failure policy. I’ll convert this PR to draft while that decision is pending.

If fail-open remains the accepted policy, I’ll close the PR. If degrading unverified candidates to cache misses is accepted, I’ll update the PR to close the issue, mark it ready for review, and request another review.

FAUST-BENCHOU (commented):
The fail-closed behavior here looks reasonable for semantic cache correctness. Treating verifier failures as cache misses avoids returning an unverified cached response.

It may be worth documenting whether this intentionally prioritizes correctness over latency when the verifier becomes temporarily unavailable.

luca-888:
> The fail-closed behavior here looks reasonable for semantic cache correctness. Treating verifier failures as cache misses avoids returning an unverified cached response.
> 
> It may be worth documenting whether this intentionally prioritizes correctness over latency when the verifier becomes temporarily unavailable.

Thanks. I’ve updated `stores-and-tools.md` to make this correctness-over-latency trade-off explicit.

Xunzhuo (commented):
Thanks for the update. The implementation and docs now match the clarified failure policy. One lifecycle gate is still open: #3176 remains marked needs-acceptance and has not received the required /accept marker. Please keep this PR pending until that issue is formally accepted.

Xunzhuo:
@luca-888, #3176 is now formally accepted, so the earlier policy/acceptance blocker is resolved. Please update this PR against current `main` and resolve the conflicts using the cache-owned verifier (`c.polarityGuard.Verifier`) introduced since this branch. Keep verifier failures as cache misses and propagate request cancellation, then rerun the cache and ExtProc regressions plus the affected E2E checks and request another review.

Xunzhuo (approved):
The verifier-error and missing-verifier paths now return a cache miss while preserving similarity and cancellation semantics. Current-head regressions pass through the real cache, adapter, service and ExtProc path, including recovery and cache-instance isolation; the deployed Gateway cache-polarity case also passes.

wilsonwu (approved):
LGTM

PR #3163, opened 2026-08-31
[Bug] Remove stale image_gen route plugin surface (#3129)

### LINKED ISSUE #3129: [Bug] image_gen route plugin has no runtime effect since #3076 (config validates but is silently ignored)
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The `image_gen` route plugin is documented as available and its configuration still validates, but it has **no runtime effect** since the protocol-translation refactor in #3076. The refactor removed `pkg/extproc/req_filter_image_gen.go` (the only execution path that consumed the plugin), yet left in place:

- the entire `pkg/imagegen/` backend package (`OpenAIBackend`, `VLLMOmniBackend`, `CreateBackend` factory), which is no longer imported by any non-test code;
- the config schema and validator for the plugin (`pkg/config/image_gen_plugin.go`, validated at `pkg/config/validator_decision.go:396`);
- the DSL registration (`image_gen` is still listed as a supported inline plugin type in `pkg/dsl/validator.go:264`);
- the official tutorial `website/docs/tutorials/plugin/image-gen.md` and the maintained fragment `config/fragments/plugin/image-gen/basic.yaml`, both of which still instruct users to configure this plugin.

As a result, a user who follows the documentation and configures `type: image_gen` on a decision gets a configuration that passes validation but is silently ignored at runtime: image generation never happens and the request proceeds down the ordinary chat-completions path.

### To Reproduce

1. Add the documented plugin to a decision, e.g.:

```yaml
routing:
  decisions:
    - name: image_route
      modelRefs: [text-model]
      plugins:
        - type: image_gen
          configuration:
            enabled: true
            backend: vllm_omni
            backend_config:
              base_url: http://image-router:8005
              model: Qwen/Qwen-Image
```

2. Start the router. Configuration validation passes (no error about `image_gen`).
3. Send a request that matches the decision. The prompt is not forwarded to the configured image backend; the request is handled like a normal chat request and no image is produced.

### Expected behavior

Either the `image_gen` plugin resumes working as documented (its execution path re-wired), or — if the feature has been intentionally superseded — the plugin, the backend package, the docs, and the fragment should be removed and configuration validation should reject `image_gen` with an explicit unsupported-feature error instead of silently accepting it.

### Impact

Users who configure `image_gen` per the official documentation get a silent no-op. Because configuration validates successfully and there is no e2e coverage, the regression is easy to miss. The change is contained to the plugin and its request-path wiring, so a fix is low-risk: either restore the execution path or remove the stale surface.

### Additional context

The removal happened in #3076 (`e5da889c` removed `req_filter_image_gen.go` and the `req_filter_modality*.go` request filters). `pkg/imagegen.CreateBackend` is referenced only by its own tests; no production code imports `pkg/imagegen/`. The Request-side `image_generation` tool support in `pkg/llmprotocol` is a separate Responses-API hosted-tool path and does not replace this route-local backend plugin.

@Xunzhuo — this looks like it may be an accidental leftover of the #3076 refactor. Could you confirm whether `image_gen` is intended to be removed or to keep working, and how you'd like to proceed?

Bevisy:
Closes #3129

## Purpose

The `image_gen` route plugin has had no runtime effect since #3076 unified
inference protocol translation: the only execution path
(`pkg/extproc/req_filter_image_gen.go`) was deleted, leaving config
validation, `pkg/imagegen` (zero production importers), DSL/nlgen surface,
docs, and fragments advertising a feature that validates but silently does
nothing at runtime.

This PR tears down the stale plugin surface (the issue's second accepted
outcome) — it does **not** remove image-generation capability, which lives
in the #3076 path: the Responses-API hosted `image_generation` tool proxied
and validated by `protocolcodec` under a vllm-omni modality route:

- delete `pkg/imagegen`, `image_gen` metrics, and the plugin config types
  (`ImageGenPluginConfig`, `ImageGenBackendEntry`, `ModalityRoutingConfig`,
  their backend decoders); keep the live modality signal config
  (`ModalityDetectorConfig`/`ModalityDetectionConfig`/`ModalityClassifierConfig`)
- drop `image_gen` from the decision-plugin catalog, DSL (compiler,
  decompiler, emitter, parser, validator), nlgen schema, and CRD enum
- config validation now **rejects** `image_gen` with an explicit
  unsupported-feature error instead of silently accepting it; a regression
  test (`image_gen_removed_test.go`) pins the rejection
- remove the plugin tutorial (EN + zh-Hans current), fragment, e2e config
  asset, and stale references in `config/config.yaml`, bench assets,
  maintained config assets, structure-rules, and skill-registry

## Test Plan

- `go build ./pkg/...` and `go vet ./pkg/config/... ./pkg/dsl/...`
- `go test -count=1 ./pkg/config/...` (includes maintained-asset /
  reference-config / docs-contract enforcement suites) and `go test ./pkg/dsl/...`
- `uv run pytest tests/test_plugin_parsing.py -q` (24 passed)
- `make agent-validate` (manifests, workflows, ci + agent script unit tests)

## Test Result

- config: 314 specs + full Go test suite pass; dsl pass; build/vet clean
- vllm-sr plugin parsing suite: 24 passed
- agent-validate: all four gates pass (verified via a Python 3.12 agent venv;
  the pre-existing local `.venv-agent` was stuck on Python 3.9, unrelated to
  this change)
- `grep` across the repo: no `image_gen`/`imagegen` surface remains outside
  the legitimate Responses-API hosted `image_generation` tool path

Remaining risk: none identified for the routed-behavior surface; upstream
maintainers may alternatively prefer re-wiring the plugin, which would
contradict #3076's protocol-neutral design and is out of scope here.


Xunzhuo (changes requested):
Thanks for taking the teardown end to end. One blocker remains on the current head: live surfaces still advertise or invoke the deleted feature. The Dashboard DSL catalogs and config types still contain `image_gen`, while `tools/make/build-run-test.mk` and `tools/smoke/test-image-gen.sh` still point at the deleted integration test and E2E config. Please remove or update those references and run the affected Dashboard and harness gates.

Bevisy:
> Thanks for taking the teardown end to end. One blocker remains on the current head: live surfaces still advertise or invoke the deleted feature. The Dashboard DSL catalogs and config types still contain `image_gen`, while `tools/make/build-run-test.mk` and `tools/smoke/test-image-gen.sh` still point at the deleted integration test and E2E config. Please remove or update those references and run the affected Dashboard and harness gates.

Thanks for the thorough review! Applied — I cleaned the deleted `image_gen` surface out of the Dashboard (plugin type unions, DSL schema/catalog/language, topology lists) and removed the Make targets pointing at the removed integration test and E2E config.

Xunzhuo (approved):
LGTM, can you take a look https://github.com/vllm-project/semantic-router/issues/3183 as a follow up?

Bevisy:
> LGTM, can you take a look #3183 as a follow up?

Sure, I'll take a look at #3183.

PR #3165, opened 2026-08-31
[Feature] Add Metal (Apple Silicon) support to candle-binding

### LINKED ISSUE #3160: [Feature] Add Metal (Apple Silicon) support to candle-binding
### Goal

On macOS, Router Model inference falls back to CPU; the GPU sits idle. The candle crates support a `metal` feature but `candle-binding` does not expose it. A prototype (metal feature + Metal device selection) on an Apple M-series machine, mmBERT-32K jailbreak classifier, 256 requests per point, original CPU path vs Metal:

| workers | CPU (origin) p50 | Metal p50 | CPU (origin) p99 | Metal p99 | CPU (origin) req/s | Metal req/s | speedup |
|---|---|---|---|---|---|---|---|
| 1 | 89ms | 51ms | 111ms | 66ms | 11.2 | 19.4 | 1.7x |
| 2 | 101ms | 53ms | 119ms | 67ms | 19.7 | 37.6 | 1.9x |
| 4 | 137ms | 67ms | 176ms | 94ms | 28.7 | 58.7 | 2.0x |
| 8 | 257ms | 129ms | 410ms | 169ms | 30.3 | 60.8 | 2.0x |
| 16 | 526ms | 243ms | 724ms | 364ms | 29.9 | 64.1 | 2.1x |
| 32 | 1.12s | 493ms | 2.23s | 763ms | 25.8 | 60.7 | 2.4x |
| 64 | 1.16s | deadlock | 7.66s | deadlock | 28.5 | — | — |

- ~2x throughput and half the latency across the sweep; peak 64 vs 30 req/s.
- At 64 concurrent forwards, candle's Metal backend deadlocks (`MetalDevice::command_buffer` lock), so the Metal path needs a concurrency bound.

### Proposed Workgroup

### Scope

- Add a `metal` cargo feature wiring `candle-core/metal`, `candle-nn/metal`, `candle-transformers/metal` (needs `once_cell` >= 1.21 in the lockfile).
- One shared device-resolution helper: CUDA, then Metal, then CPU; honor the existing `useCPU` flag.
- Bound Metal inference concurrency to avoid the candle command-buffer deadlock.
- Verify built-in models match CPU output within tolerance; Make target + docs.

### Non-goals

- ONNX/OpenVINO Metal providers, flash-attention on Metal.
- Production support claims or macOS GPU CI.

### Completion signal

- Metal build passes Go tests on Apple Silicon; outputs match CPU within tolerance.
- Measured gain over CPU recorded on the issue; no hang under concurrent load.

guan404ming:
Close #3160 

## Purpose

- Add optional Metal support for Candle on Apple Silicon.
- Use Metal for ModernBERT inference when CPU is disabled.
- Keep CPU fallback when the requested accelerator is unavailable.
- Run Metal inference on a fixed thread pool (`METAL_MAX_CONCURRENCY`, default 8, max 32) and drain the loader thread's command buffer, which removes the 64-worker deadlock.

## Benchmark Result

mmBERT-32K jailbreak classifier, 256 requests per point on Apple Silicon:

| Workers | CPU p50 | Metal p50 | CPU p99 | Metal p99 | CPU req/s | Metal req/s | Speedup |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | 89ms | 56ms | 111ms | 68ms | 11.2 | 17.9 | 1.60x |
| 8 | 257ms | 125ms | 410ms | 166ms | 30.3 | 62.6 | 2.07x |
| 32 | 1.12s | 486ms | 2.23s | 570ms | 25.8 | 64.0 | 2.48x |
| 64 | 1.16s | 962ms-1.14s | 7.66s | 1.02-1.34s | 28.5 | 54.4-64.9 | 2.19x |
| 128 | 2.09s | 2.04s | 8.25s | 2.24s | 29.0 | 58.5 | 2.02x |

Metal columns are measured with the inference pool in place. 64 workers previously deadlocked candle's Metal backend deterministically (3/3 runs); with the pool it completes 3/3 runs, and 128 workers is also stable.

## Deadlock root cause

candle's Metal backend keeps one open command buffer per OS thread on the device's single command queue, and the queue caps in-flight buffers at 64. Every distinct thread that ever runs a forward permanently occupies one slot, so unbounded CGo caller threads deadlock the queue regardless of how few run concurrently. Concurrency bounds alone do not fix it; confining inference to a fixed thread pool does. The device's first command buffer is also enqueued at creation, so the loader drains its queue before handing the model over, otherwise the first pooled forward blocks forever.

## Test Plan

- Verify the Metal feature builds on Apple Silicon.
- Verify the non-Metal build still compiles.
- Verify the jailbreak classifier loads and runs on Metal.
- Verify 64 concurrent workers no longer deadlock.

## Test Result

- cargo build --release --no-default-features --features metal
- cargo check --no-default-features
- cargo fmt --check
- Pressure run at 1/8/32/64/128 workers: no deadlock, results above (64 workers repeated 3x)

### DIFF
--- a/candle-binding/Cargo.lock
+++ b/candle-binding/Cargo.lock
@@ -282,9 +282,21 @@ checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
 name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
 version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
+
+[[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 
 [[package]]
 name = "block-buffer"
@@ -293,6 +305,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
 dependencies = [
  "generic-array",
+]
+
+[[package]]
+name = "block2"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cdeb9d870516001442e364c5220d3574d2da8dc765554b4a617230d33fa58ef5"
+dependencies = [
+ "objc2",
 ]
 
 [[package]]
@@ -360,6 +381,7 @@ dependencies = [
  "accelerate-src",
  "byteorder",
  "candle-kernels",
+ "candle-metal-kernels",
  "cudarc",
  "float8 0.4.2",
  "gemm",
@@ -369,6 +391,8 @@ dependencies = [
  "memmap2",
  "num-traits",
  "num_cpus",
+ "objc2-foundation",
+ "objc2-metal",
  "rand",
  "rand_distr",
  "rayon",
@@ -376,6 +400,7 @@ dependencies = [
  "thiserror 1.0.69",
  "ug",
  "ug-cuda",
+ "ug-metal",
  "yoke 0.7.5",
  "zip",
 ]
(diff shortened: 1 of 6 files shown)


Xunzhuo (changes requested):
Thanks for the benchmark and the clean non-Metal build. One blocker: this closes #3160 while leaving its known Metal deadlock at 64 concurrent forwards unresolved. The feature selects Metal whenever `use_cpu=false`, so shipping the path before a per-deployment bound can hang the router. Please keep the concurrency bound in this PR, or otherwise make Metal opt-in behind a safe bound, and get #3160 accepted before the next review.

guan404ming:
Fixed in 1cf24320. The deadlock is resolved in this PR, not just bounded.

Root cause: candle's Metal backend keeps one command buffer per OS thread on a queue capped at 64 in-flight, so unbounded CGo caller threads deadlock it regardless of concurrency. The fix runs Metal inference on a fixed thread pool (`METAL_MAX_CONCURRENCY`, default 8).

64 workers: previously 3/3 deadlock, now 3/3 pass at ~64 req/s; 128 workers also stable. Details and full sweep in the updated PR description.

Xunzhuo (approved):
Thanks for closing the concurrency gap. I rechecked the current head: Metal inference is confined to a fixed pool, the loader queue is drained before pooled forwards, and the 64/128-worker reruns cover the earlier deadlock. The non-Metal build passes locally and the current CI is green. This is ready from my side.

FAUST-BENCHOU (approved):

theohsiung (approved):

wilsonwu (approved):
LGTM

guan404ming:
Thanks!

PR #3186, opened 2026-08-31
[Bug] Make local-up-router.sh work without Authorino

### LINKED ISSUE #3184: [Bug] tools/dev/local-up-router.sh is broken: all requests 403 (ext_authz/Authorino never started) and wrong SR_ROOT after script move
### Proposed Workgroup

### Describe the bug

`tools/dev/local-up-router.sh` (the dockerless local dev harness) is broken in two independent ways:

1. **The script fails to start at all.** It resolves the repository root one directory level too shallow, so it cannot find the router binary or the config files:

   ```
   tools/dev/local-up-router.sh: line 106: bin/router: No such file or directory
   ```

2. **After working around that, every request through the local Envoy returns 403.** The local Envoy config requires an external authorization service (Authorino) that nothing in the local dev flow starts, and the filter is fail-closed, so Envoy rejects every request before it ever reaches the router:

   ```json
   {"request_path":"/v1/chat/completions","response_code":403,"response_flags":"UAEX", ...}
   ```

The same 403 affects other local entry points that use the same Envoy config, e.g. `make run-envoy`.

### To Reproduce

1. From the repo root, on a machine without Authorino running:

   ```bash
   tools/dev/local-up-router.sh
   ```

   Observe the immediate `bin/router: No such file or directory` failure.

2. Work around bug 1 (temporarily fix the root resolution in the script), start any OpenAI-compatible backend on `127.0.0.1:8000` (e.g. `tools/mock-vllm`), rerun the script, then:

   ```bash
   curl -X POST http://127.0.0.1:8801/v1/chat/completions \
     -d '{"model":"qwen3-8b","messages":[{"role":"user","content":"hello"}]}'
   ```

   Every request returns 403, with `response_flags: UAEX` in the Envoy access log.

Environment: macOS (darwin arm64), Envoy 1.35.4 via func-e, current `main`.

### Expected behavior

The script starts the router plus Envoy, and a `POST /v1/chat/completions` request reaches the backend at `127.0.0.1:8000`. Local quick-start should not require an external authorization service.

### Impact

Anyone using the documented dockerless dev workflow: the script is unusable out of the box (first the startup failure, then 100% 403s). No workaround short of hand-editing the script and the Envoy config.

### Additional context

Separately observed while reproducing (pre-existing, likely out of scope here): with the default `ROUTER_CONFIG=config/config.yaml` the router process itself exits fatally on a bare machine, since that config assumes the docker-compose infrastructure stack (writable system paths, Milvus/Redis/Postgres); `ROUTER_CONFIG=e2e/config/config.e2e.yaml` works standalone.

drivebyer:
Closes #3184

## Purpose

`tools/dev/local-up-router.sh` (the dockerless local dev harness) is broken in two independent ways (#3184):

1. It resolves the repository root one directory level too shallow, so it fails immediately with `bin/router: No such file or directory`.
2. The local Envoy config sends every request through a fail-closed `ext_authz` filter backed by an Authorino instance that no local flow starts, so every request returns 403 (`UAEX`) before reaching the router.

This PR makes local quick-start work out of the box:

- `deploy/local/envoy.yaml`: remove the `ext_authz` filter and the `authorino_service` cluster — local quick-start does not need an external authorization service. This also fixes the same 403 for the other consumers of this config (`make run-envoy`, the bench/demo targets, the hallucination demo).
- `tools/dev/local-up-router.sh`: fix `SR_ROOT` resolution (and make it absolute so it stays valid after `cd`); preserve the exit code in the `cleanup` trap instead of always exiting 0.
- `bench/grounded_fusion/run_ab.sh`: drop the line-number `sed` that stripped the ext_authz block, which no longer exists (and would have deleted the wrong lines).

Affected modules: local dev tooling and the local Envoy config only; no router/runtime code changes. Workgroup: Developer Experience & Ecosystem.

## Test Plan

- `bash -n` and repo-configured shellcheck on the changed scripts.
- `func-e run --config-path deploy/local/envoy.yaml --mode validate`.
- Dry run: `tools/dev/local-up-router.sh -d -o bin`.
- End to end: start a mock OpenAI-compatible backend (`tools/mock-vllm`) on `127.0.0.1:8000`, run `LOG_DIR=... ROUTER_CONFIG=e2e/config/config.e2e.yaml tools/dev/local-up-router.sh -o bin`, then `curl -X POST http://127.0.0.1:8801/v1/chat/completions` with a `Model-A` request.

## Test Result

- Script syntax, shellcheck, and Envoy config validation all pass.
- Before: script aborted at startup; after working around that, every request returned 403 with `response_flags: UAEX`.
- After: the request returns 200; the Envoy access log shows `upstream_cluster: vllm_backend_cluster` and a router-injected `selected_model: "Model-A"`, confirming the request was actually processed by the router (ext_proc) rather than passed through fail-open.
- Remaining risk: none identified for the local dev flow. The Authorino-based authz path is unaffected — its e2e/Kind setup uses its own Envoy config (`e2e/config/envoy-authorino-test.yaml`).


Xunzhuo (changes requested):
Thanks for fixing both local-flow failures. One delivery blocker: this changes the shared local Envoy/startup path but adds no automated regression. Current CI only validates syntax and YAML; it never starts the local path without Authorino, so the wrong-root and 403 failures can return unnoticed. Please add a focused smoke/E2E that runs the local flow with a mock backend and asserts the request reaches it without Authorino.

drivebyer:
Added the smoke test in 667291cb0; the test do not trigger because the pr don't modify the related code, so verified locally:
**Fixed config — PASS (exit 0)**
```
[local-up-smoke] Got HTTP 200; response: {"id":"chatcmpl-...","object":"chat.completion",...}
[local-up-smoke] PASS: request reached the mock backend through Envoy and was processed by the router.
```

**Old envoy.yaml from main (with ext_authz) — FAIL (exit 1)**
```
[local-up-smoke] FAIL: expected HTTP 200 through Envoy, got '403' (a 403 here means the request was rejected before reaching the router)
```
Envoy access log: `"response_flags":"UAEX"`

Xunzhuo (changes requested):
The focused smoke now checks the right contract: it starts the dockerless local harness without Authorino, requires a 200 through Envoy, and verifies ext_proc selected the model instead of accepting fail-open alone. One delivery blocker remains: on this exact head the `Core tests` job was skipped, so the new `Local dev smoke test (no Authorino)` step in `test-and-build.yml` never ran. Please update the changed-file classification/trigger so these local startup paths execute that job, then provide a passing smoke log. The existing E2E and summary checks do not exercise this script.

drivebyer:
@Xunzhuo Classifier updated; Core tests now runs the smoke. Passing log: https://github.com/vllm-project/semantic-router/actions/runs/33586212929/job/100114373098

```
[local-up-smoke] Setting up mock vLLM backend venv...
[local-up-smoke] Starting mock vLLM backend on :8000...
[local-up-smoke] Mock backend is healthy.
[local-up-smoke] Starting local-up-router.sh (config: e2e/config/config.e2e.yaml)...
[local-up-smoke] Harness is up.
[local-up-smoke] Sending /v1/chat/completions through Envoy (:8801)...
[local-up-smoke] Got HTTP 200; response: {"id":"cmpl-mock-123","object":"chat.completion",...,"model":"Model-A",...}
[local-up-smoke] PASS: request reached the mock backend through Envoy and was processed by the router.
```

Xunzhuo (approved):
The delivery blocker is resolved on this head: changed-file classification selects Core tests, and the exact-head job ran the local no-Authorino smoke through Envoy to the mock backend successfully. LGTM.

wilsonwu (approved):
LGTM

PR #3201, opened 2026-08-31
[Feature] Add admission package and model_catalog config

### LINKED ISSUE #3159: [Feature] Add a per-deployment admission gate for Router Model inference
### Goal

Prevent a Router Model deployment from accepting work when its configured capacity or lifecycle state cannot safely serve it.

### Scope

- Define per-deployment concurrency and queue bounds.
- Integrate cancellation, deadlines, draining, and readiness transitions.
- Return typed overload and not-ready outcomes.
- Expose bounded diagnostics without making routing decisions.

### Completion signal

Tests show independent deployments enforce their own bounds, drain safely, and do not turn overload into an untyped internal error.

Parent Epic: #2782

guan404ming:
Related #3159

## Purpose

- Add `pkg/admission`: the `Admissioner` interface, `Ticket`, `Do`, a `noop` gate, and a `semaphore` gate with fixed slots, a bounded wait queue, a queue timeout, context-aware waiting, and `shed | wait | fail_open` overflow policies.
- Add the `global.model_catalog.admission` config block keyed by system-model deployment, with validation, canonical round-trip mapping, and a reference-config example.
- No behavior change: nothing acquires a gate yet. Wiring, gate metrics, and docs land in the follow-up PR that closes #3159.

## Test Plan

- Verify the semaphore bounds concurrency and each overflow policy behaves as configured.
- Verify queued waiters honor context deadlines and the queue timeout without leaking slots.
- Verify the admission block parses through the full config pipeline, survives canonical export, and invalid values are rejected at parse time.

## Test Result

- `go test ./pkg/admission/ -race`
- `go test ./pkg/config/` (includes end-to-end `ParseYAMLBytes` cases and a parse of the shipped `config/config.yaml`)
- `go build ./pkg/...`, `gofmt`, `go vet`


src/semantic-router/pkg/admission/semaphore.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	queued := true
+	select {
+	case s.waiters <- struct{}{}:
+	default:
+		switch s.overflow {
+		case OverflowFailOpen:
+			return func() {}, nil
+		case OverflowWait:
```
FAUST-BENCHOU: Could `wait` still respect `max_queue` rather than bypassing it here? Once the queue is full, `queued=false` lets an arbitrary number of callers block directly on `slots`, so `max_queue` no longer bounds waiting work.

FAUST-BENCHOU (commented):
One concern below.

src/semantic-router/pkg/admission/semaphore.go
guan404ming: Make sense to me, respecting `max_queue` preserves better semantic here. Just fixed in 333d8334. `wait` now blocks for a queue slot, so queued work never exceeds `max_queue`. Also validated and covered by tests.

Xunzhuo (changes requested):
The admission package, bounded semaphore, and config tests are coherent. One merge blocker remains: this adds global.model_catalog.admission to the shipped reference config and canonical public contract, but its documentation is deferred. Please document the keys, overflow semantics, and currently unwired behavior in the config or website docs in this PR, as required for behavior-visible config changes.

guan404ming:
Thanks! Documented in fca58ff2: the config.yaml admission block now covers the deployment keys, field meanings, shed/wait/fail_open overflow semantics, and the not-yet-wired status. Can mirror to website docs if preferred.

Xunzhuo (approved):
The requested config documentation now covers the keys, overflow semantics, and unwired status. The current head and CI address my blocker; thank you.

guan404ming:
Thanks!

wilsonwu (approved):
LGTM

guan404ming:
Thanks!

PR #3205, opened 2026-08-31
[Bug] Validate jailbreak signal rules

### LINKED ISSUE #3203: [Bug] Jailbreak signal rules are not validated
### Proposed Workgroup

### Describe the bug

`evaluateJailbreakRule` (`src/semantic-router/pkg/classification/classifier_signal_jailbreak.go:111`) matches `contrastive` and sends every other method to the model. Nothing validates the value.

`method: hybrid` is not implemented. It ships in `config/config.yaml:414`, in `config/fragments/signal/jailbreak/patterns.yaml:5`, and in the tutorial at `website/docs/tutorials/signal/learned/jailbreak.md:40`. Every one of those rules carries `jailbreak_patterns` / `benign_patterns`, and only the contrastive path reads them.

Separately, the match is `riskScore >= threshold`, so `threshold: 0` marks every request as a jailbreak. An omitted threshold gives the same value.

### To Reproduce

A rule with `method: hybrid`, `threshold: 0.99` and `jailbreak_patterns: ["zzq unique canary phrase alpha"]` does not match a prompt containing that exact phrase. Only the model runs.

A second rule with `threshold: 0` matches `"What is the capital of France?"` at a risk score of 0.0000011.

Both configs load without an error or a warning.

### Expected behavior

An unimplemented method is a config error, and a jailbreak rule declares a threshold that can fail.

### Impact

A guardrail loads, reports healthy, and silently ignores half of its configuration. The shipped reference config is in that state today: `prompt_injection` lists six patterns that are never read.

Parent Epic: #2760

subin9:
Fixes #3203

Unknown methods, patterns on a model-backed rule, and a non-positive or `>1` threshold are now config errors.

`evaluateJailbreakRule` matched `contrastive` and sent every other method to the model, so `method: hybrid` loaded fine and its pattern lists were never read.

The shipped configs and both language versions of the tutorial move from `hybrid` to `contrastive`, which is the method that reads the `jailbreak_patterns` / `benign_patterns` those rules already carry.

```
before: method: hybrid, threshold 0.99, jailbreak_patterns: ["zzq unique canary phrase alpha"]
        loads, and a prompt containing that exact phrase does not match

after:  routing.signals.jailbreak "x": unknown method "hybrid"; valid values are
        ["classifier" "model" "contrastive"] (patterns are only consulted by "contrastive")
```

Checked that `config/config.yaml` still loads and that a config with `method: hybrid` or `threshold: 0` now fails at startup. Every shipped config already sets a positive threshold.

Seven new tests cover the accepted methods, an unknown method, patterns on a model-backed rule, a non-positive and an out-of-range threshold, and duplicate or unnamed rules.

### DIFF
--- a/src/semantic-router/pkg/config/validator_jailbreak.go
+++ b/src/semantic-router/pkg/config/validator_jailbreak.go
@@ -1,0 +1,100 @@
+package config
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+)
+
+// Recognised values for JailbreakRule.Method.
+//
+// The runtime dispatches on this string in
+// classification.evaluateJailbreakRule, which matches "contrastive" and sends
+// everything else to the model-backed path. Without validation, an
+// unrecognised value is not an error but a silent downgrade: a rule written as
+// `method: hybrid` with jailbreak_patterns still loads, still reports healthy,
+// and never consults a single pattern, because only the contrastive path reads
+// them. A guardrail that silently ignores half its configuration is worse than
+// one that refuses to start.
+const (
+	// JailbreakMethodContrastive scores the prompt against the configured
+	// jailbreak_patterns / benign_patterns.
+	JailbreakMethodContrastive = "contrastive"
+	// JailbreakMethodClassifier runs the configured sequence classifier.
+	// "" and "model" are accepted spellings of the same path.
+	JailbreakMethodClassifier = "classifier"
+	JailbreakMethodModel      = "model"
+)
+
+var validJailbreakMethods = []string{
+	"",
+	JailbreakMethodClassifier,
+	JailbreakMethodModel,
+	JailbreakMethodContrastive,
+}
+
+// validateJailbreakContracts rejects jailbreak signal rules that would load
+// without doing what they say.
+func validateJailbreakContracts(cfg *RouterConfig) error {
+	if cfg == nil {
+		return nil
+	}
+	seen := make(map[string]struct{}, len(cfg.JailbreakRules))
+	for _, rule := range cfg.JailbreakRules {
+		name := strings.TrimSpace(rule.Name)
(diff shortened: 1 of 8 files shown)


src/semantic-router/pkg/config/validator_jailbreak.go
```diff
@@ -0,0 +1,95 @@
...
+		name := strings.TrimSpace(rule.Name)
+		if name == "" {
+			return fmt.Errorf("routing.signals.jailbreak: every rule needs a name")
+		}
+		if _, dup := seen[name]; dup {
+			return fmt.Errorf("routing.signals.jailbreak: duplicate rule name %q", name)
+		}
+		seen[name] = struct{}{}
+
+		method := strings.ToLower(strings.TrimSpace(rule.Method))
```
FAUST-BENCHOU: Could we either require canonical method spelling or normalize it before runtime dispatch? `CONTRASTIVE` passes this validator, but runtime checks `rule.Method == "contrastive"` exactly, so it still takes the model path and ignores its patterns.

FAUST-BENCHOU (commented):
One concern below.

src/semantic-router/pkg/config/validator_jailbreak.go
subin9: Thanks, fixed. The validator now compares the method as written, so CONTRASTIVE is rejected with the canonical spelling in the message instead of loading and taking the model path.

Xunzhuo (approved):
The current head rejects unknown or non-canonical methods, model rules with ignored patterns, invalid thresholds, and duplicate or unnamed rules; it also corrects the shipped config and docs. Focused config tests passed, and #3203 is accepted.

FAUST-BENCHOU (approved):

PR #3206, opened 2026-08-31
[Bug] Scan the whole prompt in the security classification API

### LINKED ISSUE #3204: [Bug] The security classification API and the routing signal disagree on long prompts
### Proposed Workgroup

### Describe the bug

`CheckForJailbreakWithRisk` (`src/semantic-router/pkg/classification/classifier_jailbreak_risk.go:130`) classifies the text in one call. The model truncates at `MAX_CLASSIFICATION_SEQ_LEN = 512` (`candle-binding/src/model_architectures/traditional/modernbert.rs:20`), so `/api/v1/classify/security` only ever sees the start of a long prompt.

The routing path does not work that way. `evaluateJailbreakSignal` (`classifier_signal_jailbreak.go:62`) splits the text with `jailbreakSignalChunks` and keeps the riskiest chunk.

### To Reproduce

18k characters of benign prose, with an injection placed at three positions, on `main` at `9fbd7d85`:

```
                 API risk   routing risk
no injection       0.6393         0.8829
inject at start    0.8441         0.5291
inject in middle   0.6393         0.7065
inject at end      0.6393         0.8829
```

Past roughly 3,500 characters the API score stops moving, because it is scoring the same opening fragment every time.

### Expected behavior

The same text produces the same risk score from both paths.

### Impact

`/api/v1/classify/security` is the surface you would use to check what the router will do, and it misses anything past its first window. Anything that treats the two as interchangeable, a dashboard or an eval harness, disagrees with the path that actually enforces.

Parent Epic: #2760

subin9:
Fixes #3204

`CheckForJailbreakWithRisk` classified the text in one call, and the model truncates at its own sequence limit, so `/api/v1/classify/security` only saw the start of a long prompt. It now splits the text with `jailbreakSignalChunks` and keeps the riskiest chunk, which is what the routing path already does.

18k characters of benign prose with an injection placed at three positions:

```
                 API before   API after   routing
no injection         0.6393      0.8829    0.8829
inject at start      0.8441      0.5291    0.5291
inject in middle     0.6393      0.7065    0.7065
inject at end        0.6393      0.8829    0.8829
```

Short prompts still take a single call with the text unchanged.

Two new tests: a long prompt has to classify every chunk and return the highest score, and a short prompt has to stay one call.

Note the two paths can still report a different `is_jailbreak` for the same score, because this one thresholds on `prompt_guard.threshold` while the routing signal uses the rule's. That is a separate question from the score itself.


src/semantic-router/pkg/classification/classifier_jailbreak_risk.go
```diff
@@ -127,9 +127,26 @@ func (c *Classifier) CheckForJailbreakWithRisk(ctx context.Context, text string)
...
+	// The model truncates at its own sequence limit, so a single call only ever
+	// sees the start of a long prompt. The routing path already scans the whole
+	// text in chunks and keeps the riskiest one; do the same here so both
+	// surfaces answer the same question.
+	var result SequenceClassificationResult
+	bestRisk := float32(-1)
+	for _, chunk := range jailbreakSignalChunks(text) {
+		chunkResult, err := c.jailbreakInference.Classify(ctx, chunk)
+		if err != nil {
+			return false, "", 0.0, 0.0, fmt.Errorf("jailbreak classification failed: %w", err)
```
FAUST-BENCHOU: Could we preserve the routing path's partial-chunk behavior here? It records failed chunks but can still keep a genuine match from another chunk, while this returns on the first error, so the API can still disagree with routing.

FAUST-BENCHOU (commented):
One concern below.

src/semantic-router/pkg/classification/classifier_jailbreak_risk.go
subin9: Thanks, fixed. A failed chunk is now logged and skipped rather than failing the call, so a match in a later chunk survives; the error is only returned when no chunk could be classified.

Xunzhuo (approved):
The current head scans the same bounded chunks as routing, keeps the highest-risk result, and preserves a valid later match when another chunk fails. The regression cases cover long input, short input, partial failure, and total failure; #3204 is accepted.

FAUST-BENCHOU (approved):

wilsonwu (approved):
LGTM

PR #3152, opened 2026-08-31
[Feature] Add label score distributions to the LLM generic classifier backend

guan404ming:
Closes #2923

## Purpose

- Update `type: llm` generic classifier signals to request every label’s score.
- Return the model-reported distribution instead of synthetic one-hot scores.
- Reject missing labels, invalid scores, and non-normalized distributions.
- Add outbound routing E2E coverage and documentation.

## Test Plan

- Verify scores are matched by label.
- Verify invalid distributions are rejected.
- Verify high and low scores produce the expected routing decisions.

## Test Result

- `make test-semantic-router`
- `make recipe-conformance-live-cpu-all`
- `make e2e-test E2E_PROFILE=envoy-ai-gateway E2E_TESTS=llm-classifier-distribution-routing,sequence-classifier-routing`

Issue #3081, opened 2026-08-28
[Feature] Define explicit decision ranking semantics for tier, priority, and confidence


Xunzhuo:
### Goal

Make decision selection predictable by separating three different concepts:

- rule eligibility: whether a decision matched
- policy ordering: tier and priority
- evidence strength: an optional, comparable confidence score

This follows #3079 and #3080. #3080 is the right immediate safety fix, but the long-term ranking contract should not depend implicitly on whether every matched rule happened to publish a score.

### Proposed Workgroup

MoM & Routing

### Scope

Current behavior has a few overlapping semantics:

- once matched decisions contain a positive tier, selection becomes `tier -> confidence -> priority` regardless of `routing.strategy`
- unreported confidence and structural rule results have historically been represented as `1.0`
- AND, OR, NOT, predicates, and signal scores are combined into one decision confidence even though boolean policy and model evidence have different meanings
- reported scores from different signal families are not necessarily calibrated or comparable

Define and implement one explicit public contract:

- tier remains the hard precedence boundary
- `routing.strategy` applies consistently within the selected tier
- `priority` is deterministic policy ordering
- missing confidence is absent or unknown, never synthetic certainty
- boolean gates such as NOT and predicates affect eligibility without manufacturing ranking evidence
- confidence ranking is allowed only for scores that satisfy an explicit comparability contract, with a deterministic priority fallback
- decision traces report the ranking mode, score provenance, fallback reason, and final tie-break
- built-in recipes, docs, unit tests, and maintained recipe-conformance probes cover the contract

The design also needs a compatibility plan for recipes that currently rely on confidence-first ordering inside a tier.

### Non-goals

- training or improving Router Models
- adaptive threshold fitting and policy calibration workflows tracked by #2341
- choosing among `modelRefs` after a decision has already matched
- expanding #3080 beyond its immediate regression fix

### Completion signal

- [ ] The documented ordering table covers tiered and non-tiered `priority` and `confidence` strategies.
- [ ] Unknown, structural, and comparable scored evidence have distinct runtime semantics.
- [ ] Configuration validation rejects or explicitly falls back from ambiguous confidence ranking.
- [ ] Eval and replay traces explain which ordering mode selected the winner.
- [ ] Unit tests cover mixed scored/unscored pools, priority ties, catch-alls, predicates, NOT, and heterogeneous score provenance.
- [ ] Maintained recipe probes cover at least one cross-route collision and the compatibility behavior is documented.

[...]

PR #3227, opened 2026-08-31
[Proposal] Decision ranking semantics for tier, priority, and confidence

subin9:
Related #3081

@Xunzhuo [asked](https://github.com/vllm-project/semantic-router/issues/3081#issue-5275620564) for a contract proposal and a behavior matrix before any runtime change. The [matrix](https://github.com/vllm-project/semantic-router/issues/3081#issuecomment-5462730242) is in the thread; this is the proposal. `pkg/decision` is untouched.

The doc records the three ordering branches as they run today, five properties that need an explicit decision, and what `SignalConfidences` actually holds. Then the contract, the compatibility surface, and DR-01 to DR-08 as bounded follow-ups.

Two things changed since the matrix.

The generic `type: llm` row is no longer a synthetic one-hot. #3152 replaced it with the model-reported distribution on 08-31.

The matrix also said the pools holding `omni` fall back to priority because `omni` reports no score. That is backwards. `conversation` rules write a confidence, so `omni` reports a scored `1.0`, and `multi-objective` / `privacy-first` tier 2 ranks by confidence today. The three pools that do fall back do so because the competing decision carries a `NOT` or a `keyword`.

Nothing misroutes, since `omni` also holds the higher priority there. But priority is not what selects it:

```
route at priority 900, omni at priority 250

strategy=priority    winner=omni  conf=1.0000 scored=true
strategy=confidence  winner=omni  conf=1.0000 scored=true

same pool, conversation leaf declared policy instead of evidence
                     winner=unified_privacy_sensitive_route  conf=0.8200
```

Both strategies agree because the tiered branch returns before reading `routing.strategy`, which is also the answer to the first starting question. For the second: score eligibility needs three things together rather than a scored or unscored split, namely a declared role per leaf, an aggregate that does not move with leaf count or with which `OR` branch matched, and a declared score kind.

The tables and both probes come from a test-only file against `pkg/decision` that pins current behavior. I can send that as a separate PR if pinning it before any contract change is useful.

Docs only. The Netlify preview covers the page and its sidebar entry.


Xunzhuo (commented):
The proposal accurately describes the current ranking branches and the policy-versus-evidence ambiguity, and its delivery slices are reviewable. The remaining blocker is lifecycle only: #3081 is still needs-acceptance, so the required check cannot pass. Once a maintainer accepts that proposal, this is ready for approval from this pass.

FAUST-BENCHOU (commented):
Thanks for documenting the current ranking behavior before changing the implementation. Having the behavior matrix first makes the contract discussion much clearer.

One thing that may be worth clarifying is whether `priority` and `confidence` represent different concepts or different ranking strategies over the same candidate set.

subin9:
Thank you for the review — I'd say different concepts.

`priority` is defined for every decision, but `confidence` only exists when something actually reported a comparable score. That asymmetry is why #3080 needs the pool fallback and why `signalConfidence` fills an absent key with `1.0`. If they were just two sort orders over the same set, neither would be necessary.

There's also no single confidence axis to sort on. `SignalConfidences` mixes a classifier probability, an embedding similarity, a min-similarity, a projection output and a boolean constant in one `map[string]float64`, which is what the comparability declaration in DR-06 is for.

So the way I'd put it: `priority` is the ordering, `confidence` refines it inside a pool that declared itself comparable, and `strategy` just picks between the two within a tier (DR-04).

I'll add a short section on this ahead of the ordering table. Happy to go the other way if you'd rather keep them as two strategies over one set.

Xunzhuo (approved):
The proposal clearly separates policy priority from evidence confidence and documents the current ranking and fallback branches. The exact-head documentation checks pass.

wilsonwu (approved):
LGTM

Issue #3199, opened 2026-08-31
[Feature] Standardize Router Model dataset, training-run, and artifact manifests


Xunzhuo:
### Goal

Provide one machine-readable provenance contract for Router Model datasets, training runs, evaluation results, and produced artifacts.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Define versioned manifests for dataset source/revision/license/splits, preprocessing, base model, code revision, dependencies, seed, hyperparameters, label or span mapping, metrics, and artifact digests.
- Validate manifest completeness and cross-reference identities across training, evaluation, export, and runtime qualification.
- Keep secrets, private samples, and environment-specific infrastructure out of public artifacts.
- Integrate the contract with existing task-specific training scripts incrementally.

### Non-goals

- Rewriting every training script in one change.
- Hosting datasets or model artifacts in the repository.
- Automatically promoting an artifact from training loss.

### Completion signal

- One existing classifier workflow emits valid dataset, run, evaluation, and artifact manifests.
- Missing or mismatched provenance fails validation.
- The immutable artifact identity can be consumed by runtime qualification and a model card.

Parent Epic: #2975

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

subin9:
Two implementations of this are open: #3247 (mine, 08-31) and #3310 (@guan404ming, 09-02). Neither mentions the other and they land in different paths (`src/training/model_eval/provenance` vs `src/training/provenance`), so flagging it here before acceptance picks one by accident.

They differ in:

- #3247 validates against a JSON Schema per kind (exact `schema_version`, unknown fields rejected) and cross-references the bundle: revision must be a commit sha, label mapping contiguous and identical across manifests, digest recomputed from the file list, evaluation rows checked against the split. The label check is what catches the intent registry order that costs 27 points on MMLU-Pro.
- #3247 wires one training script (`prompt_guard`) so manifests are emitted rather than hand-written, and its evaluation manifest is what the #3248 baseline consumes.
- #3310 validates source licenses, and Xunzhuo's review notes are already applied on it.

[...]

Issue #3197, opened 2026-08-31
[Research] Establish a versioned quality baseline and gap report for built-in Router Models


Xunzhuo:
### Goal

Measure the maintained built-in Router Model portfolio on held-out and operationally relevant slices so follow-up work starts from evidence.

### Scope

- Inventory the exact domain, complexity, jailbreak, PII, fact-check, feedback, and modality artifacts used by maintained configurations.
- Record dataset identity and provenance, task metrics, calibration, abstention/error behavior, relevant slices, robustness, latency, and memory.
- Separate model-quality gaps from integration, threshold, preprocessing, and routing-policy gaps.
- Publish a versioned gap report that recommends bounded follow-up tasks without silently changing defaults.

### Completion signal

One reproducible report covers every maintained built-in Router Model and states which model-specific issues remain justified.

Parent Epic: #3194

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

subin9:
Proposal for this one. #3248 already has a draft of it, paused until this is accepted.

**Approach**

Measure the artifacts config.yaml actually serves (constants.py names a different one for every task), and take class order from the artifact's own label_mapping.json rather than the registry. For each served text classifier, report accuracy, macro-F1, calibration, threshold behaviour and latency on its published held-out split. PII and modality have no usable split, so they show up as coverage gaps.

The output is a versioned gap report that files each finding under identity, runtime, coverage, calibration, threshold or quality, so it's clear which ones mean retrain and which mean fix the integration.

From adaamko's review on #3248: safe-class recall and FPR at the configured threshold go in as summary columns, and a task can list more than one dataset so external held-out sets can be added as specs later.

**Surfaces**

src/training/model_eval/ only. No router, config or default changes.

**Validation**

Unit tests for the inventory, the calibration maths and the label-order check, plus one measured run of the four served classifiers with the report rendered. No CI job runs src/training tests, so that part is local.

adaamko:
[...]

Issue #3200, opened 2026-08-31
[Feature] Prove the Router Model training and promotion pipeline on one built-in model


Xunzhuo:
### Goal

Exercise the reproducible Router Model pipeline end to end on one maintained built-in model.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Select one built-in model with a documented baseline gap and approved dataset.
- Reproduce training or adaptation from declared inputs and manifests.
- Evaluate against the unchanged supported baseline, including calibration, robustness, latency, and runtime compatibility.
- Export an immutable artifact, produce model-card inputs and compatibility evidence, and exercise shadow, explicit promotion, and rollback.
- Record rejection when the candidate does not meet its predeclared gate.

### Non-goals

- Changing every built-in model.
- Automatic online training or promotion.
- Hiding a failed candidate or publishing private data.

### Completion signal

- One model completes dataset -> training -> held-out evaluation -> export -> runtime conformance -> shadow -> promotion or rejection.
- Every step is reproducible from versioned manifests.
- Rollback to the prior artifact is documented and exercised.

Parent Epic: #2975

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

PR #3247, opened 2026-08-31
[Feature] Standardize Router Model dataset, training-run, and artifact manifests

subin9:
Related #3199

## Purpose

Nothing in the model path pins a revision. `tools/make/models.mk` downloads with `hf download $(HF_ORG)/$$model` and no `--revision`, so `models/` holds whatever `main` pointed at that day, and no artifact records which run produced it. The prompt-guard LoRA script seeded `train_test_split` with `random_state=42` but never seeded torch or the classifier head.

This adds four manifest kinds, dataset, run, artifact and evaluation, as versioned YAML with JSON Schema, following `router_calibration_schema.py`: exact `schema_version`, unknown-field rejection, and no YAML anchors, aliases, tags or merge keys.

Schema validation proves each manifest is well formed. It does not prove the four describe the same model, so cross-reference validation fails on a revision that is not a commit sha, a label mapping that is not contiguous `0..n-1`, an `identity.digest` that disagrees with the file list it summarises, a reference that resolves to nothing, an evaluation reporting more rows than the split holds or fewer with no declared `sample_limit`, and a label mapping that differs between any two manifests.

The label check is the one that motivated this. `constants.py` lists the intent classifier's 14 labels in a different order than the artifact's own `label_mapping.json`, which costs 27 accuracy points on MMLU-Pro test, 0.9445 down to 0.6746. Nothing in the metrics looks wrong, so only a cross-reference catches it.

The dataset schema takes a `composite` source type, since the prompt-guard training set is built in code from `lmsys/toxic-chat`, `OpenSafetyLab/Salad-Data` and the in-repo pattern lists. Each upstream stays pinned separately, so a mixture is traceable without hosting it.

`prompt_guard` is the one workflow wired up. `save_training_artifacts` now writes the dataset, run and artifact manifests alongside the adapter, and `--seed` calls `set_seed` before any sampling so the recorded seed describes the run. Emission raises rather than writing a partial bundle.

The evaluation manifest is emitted by whatever measures the artifact, which is the baseline harness in the #3197 branch. A bundle without one does not validate, since an artifact with no measured evaluation is not qualified.

Non-goals from the issue hold: no other training script changes, nothing is hosted in the repo, and nothing promotes an artifact.

## Test Plan

- Unit tests for the schema, cross-reference, redaction and emission behaviour.
- One real training run that emits the manifests, then validate the bundle.
- The changed-file lint and structure gates.

## Test Result

```
pytest src/training/model_eval/test_provenance.py \
       src/training/model_classifier/prompt_guard_fine_tuning_lora/
26 passed

ruff check --config tools/linter/python/.ruff.toml <changed files>   All checks passed
tools/agent/scripts/structure_check.py <changed files>               no errors
codespell <changed files>                                            clean
black --check <changed files>                                        clean
```

One training run on the jailbreak mixture, `--model mmbert-base --epochs 1 --max-samples 200 --seed 42`, then:

```
python -m provenance.cli validate <run>/manifests
ok dataset    prompt-guard-mixture-max200
ok run        prompt-guard-mmbert-base-r8-seed42
ok artifact   prompt-guard-mmbert-base-r8-lora
```

The bundle is reported incomplete until an evaluation manifest joins it, which is the intended behaviour and what the #3197 branch adds.

Permuting the class order in one manifest fails the bundle. So does a digest that disagrees with the file list, and a reference that resolves to nothing.

Remaining: `structure_check.py` still warns that `jailbreak_bert_finetuning_lora.py` is long. It was 764 lines before this change and is 798 now, under the 800 limit. Putting the emission in `save_training_artifacts` rather than in `main()` is what keeps it there, and it is also where the rest of the run's output is written.

No CI job runs the tests under `src/training/`, so the commands above are local. Wiring that up looks like a separate change.


FAUST-BENCHOU (commented):
Since the manifest format is intended to standardize Router Models, should we define a common emission interface for other training scripts instead of wiring only prompt_guard here? Otherwise provenance coverage may depend on individual training workflows.

Xunzhuo:
Thanks for putting this together. #3199 is still awaiting acceptance and assignment, so I’m holding the implementation review for now. Please pause here until the Router Models maintainers accept the bounded contract; after that, rebase and we’ll review the exact manifest surface.

subin9:
@FAUST-BENCHOU The shared surface is already there: `src/training/model_eval/provenance/emit.py` holds the builders, the revision and digest resolution, and `write_manifest`, which validates before it returns. `jailbreak_provenance.py` is a thin adapter over it.

What is left per script is the part that genuinely differs: which upstreams the dataset is built from, and what the hyperparameters are called. Wiring a second script means calling the same builders with its own values.

Agreed that coverage should not depend on each workflow remembering to do it. #3200 is where that becomes enforceable, since it requires one model to complete the path from declared inputs to a promoted artifact. Wiring the rest here would have gone against the non-goal in #3199 of rewriting every training script in one change, but happy to add a second script if you would rather see the interface exercised twice before it settles.

Thank you :)

@Xunzhuo Understood, pausing here for now. Thank you!

subin9:
#3199 accepted, so unpaused for review :)

Xunzhuo (changes requested):
Thanks for wiring provenance into the training lane. The manifest is not bound to the bytes actually trained: the model and datasets load without revisions, then the current Hub SHAs are resolved only after training. A moving ref or stale cache can therefore produce valid-looking false provenance. Please resolve and pin the exact revisions into the actual loads. The workflow also emits only dataset, run, and artifact manifests; it omits the evaluation manifest required for #3199 completion, so please wire that producer before closing the issue.

subin9:
Revisions are resolved once before anything loads and passed into the loads themselves, so the datasets and the base model are read at the commit the manifests name. Emission records those values instead of resolving again, and a test swaps the resolver for one that raises to keep it that way.

The workflow writes the evaluation manifest now: the adapter is measured on the validation split it held out, with the metrics, calibration, abstention curve and latency the schema requires, and split_rule: by_row, since this workflow splits its own mixture positionally. validate_bundle over the four manifests is a test.

Xunzhuo (approved):

Issue #3194, opened 2026-08-31
[Epic] Measure and improve built-in Router Model quality


Xunzhuo:
### Goal

Establish a measured improvement program for the built-in Router Model portfolio instead of changing defaults from isolated examples.

### Scope

- Maintain versioned quality, calibration, robustness, latency, and memory baselines.
- Separate model-quality gaps from preprocessing, integration, threshold, and routing-policy gaps.
- Require held-out evidence, artifact identity, runtime compatibility, and rollback notes before changing a default.
- Open model-specific work only for gaps identified by the shared baseline.

### Child issues

- #3197
- #1485
- #2568
- #2587

### Non-goals

- MoM recipe or provider-pool quality.
- Shared benchmark infrastructure owned by Evaluation & Quality.
- Automatic model promotion.

Parent Workgroup: #2966

adaamko:
Recommending for acceptance, i'd take it as DRI.

Its the half of the old #2974 we can move on now, the models exist and their failures are observable.

How i'd evaluate under this epic, same protocol for every built-in model, jailbreak first:

- Baseline the shipped artifact. `mmbert32k-jailbreak-detector-merged` on Aegis 2.0 and WildGuardMix held-out, macro-F1, next to published guard checkpoints run on the identical split. The repo currently evaluates a different checkpoint than it ships and uses in-distribution weighted F1, so neither number compares to anything public.
- Split by provenance. PolyGuardMix is largely human-verified translations of WildGuardMix into 17 languages plus native multilingual data, so a row split puts the same prompt in train and test twice and the multilingual number lies. Same pattern in several intent sets.
- Recall and benign false-positive rate separately, per language. One F1 hides direction. For a router a false positive is a blocked or rerouted request on every call, thats a cost/latency defect as much as a safety one.
- Model gap vs integration gap, decided before any model change. #3146 (byte offsets reported as characters) and #3333 (first 512 tokens only, argmax thresholded) both presented as bad detections and were code. The report says which kind each gap is.
- Candidate qualified against the unchanged default on the same frozen sets, runtime compat + rollback note, per the completion criterion here. Seeds and error bars under two points.

[...]

PR #3248, opened 2026-08-31
[Feature] Establish a versioned quality baseline for the built-in Router Models

subin9:
Related #3197

Rebased onto `main` after #3247 was squash merged, so the commits here are this change alone.

## Purpose

`quality_baseline.py` measures the artifact a maintained configuration actually loads, resolved from `config/config.yaml`, and `gap_report.py` renders the result into a versioned report that sorts gaps by who has to act on them.

* **One metric implementation.** #3247 landed `provenance/metrics.py` so that "two producers report the same numbers the same way", and `jailbreak_provenance.py` is the second producer for this same task. This branch had its own `calibration.py` filling the same schema fields with different maths, so it is gone and the harness calls `provenance.metrics`.
* **The report measures the decision the router makes.** `jailbreakRiskScore` compares the probability mass on `positive_labels` to the configured threshold (`classifier_jailbreak_risk.go:78`), not the argmax. `operating_points` reports recall, false-positive rate and precision on that score, so safe class recall is one minus the false-positive rate and becomes a column of its own.
* **Two artifacts compare without agreeing on a threshold.** `--artifact-repo` already measured a candidate, but the report ranked it by accuracy, which moves with the threshold and the class balance. `discrimination` adds `roc_auc` and recall inside a false-positive budget, and the comparison reads those.
* **The result names the bytes it measured.** An artifact manifest lends its identity to every number a run publishes, so it now selects those bytes as well. The repository and revision it names are what gets downloaded, its file list is the `allow_patterns`, and every listed file is re-hashed before the model is loaded. A local directory that does not hash to the manifest, and an `--artifact-repo` the manifest does not describe, both fail before scoring. Resolution moved to `baseline_artifact.py` so the binding is covered by tests that need none of the training extras.
* **The short length bands are reported separately.** One band from 0 to 64 characters averaged the region where the shipped detector is weakest, so `LENGTH_BOUNDARIES` gains cuts at 16 and 32. The slice spread finding now ignores a band under 30 rows, which stops a ten row band from becoming the comparison.
* **Every split says how it was held out.** `TaskSpec` carries the `split_rule` the evaluation manifest requires, which also fixes the `TypeError` the rebase exposed. All four wired tasks read their upstream's published split, so the report files a coverage gap per task saying an external `by_source` set is still needed.
* **The run states what it cannot measure.** An artifact published before the training-run manifests has no `run_ref`, so its bundle does not pass cross-reference validation; that is now a gap the run prints and a line in the README beside the command. A `tokenizer_class` the installed `transformers` does not know is reported as a runtime gap instead of a bare `ValueError`. PII is token classification, modality has no wired split, and complexity has no configured artifact at all, so all three are coverage gaps rather than numbers.

One finding worth its own issue: the evaluation registry names the jailbreak classes `safe` and `unsafe`, the artifact's `config.json` and both mapping sidecars name them `benign` and `jailbreak`, and `routing.signals.classifiers[0].labels` writes a third spelling, `[SAFE, JAILBREAK]`. No number is wrong today, because the indices line up, but nothing checks it. Happy to open that separately.

## Test Plan

`make check BASE_REF=origin/main` and `make test-training-contracts`, then `quality_baseline.py` for each task on CPU, `gap_report.py` over the four results, and `python -m provenance.cli validate` on a bundle. For the binding, the harness was also run three more ways against the published jailbreak artifact: referencing the manifest emitted for it, referencing that manifest against a copy of the snapshot with one byte of `config.json` changed, and referencing it while asking for a different `--artifact-repo`.

## Test Result

Both gates pass. The pytest lane goes from 40 to 67: ten in `test_artifact_inventory.py`, which this PR registers on that lane, nine in `test_provenance.py` for `operating_points` and `discrimination`, one more there for the file re-hash, and seven in `test_baseline_artifact.py` for the manifest binding. The lane runs in CI through `make check`, and the seven binding cases need nothing the agent environment does not already have, so they run rather than skip. `discrimination` also agrees with scikit-learn's `roc_auc_score` and `roc_curve` over 400 random splits with heavy score ties.

Measured on the published held-out splits, seed 42, CPU, batch 16:

| task | rows | acc | macro F1 | ECE | AUC | R@1%FPR |
|---|---|---|---|---|---|---|
| jailbreak | 827 | 0.7267 | 0.7092 | 0.2322 | 0.8735 | 0.4165 |
| fact-check | 3349 | 0.8191 | 0.8038 | 0.1464 | | |
| domain | 12032 | 0.9445 | 0.9470 | 0.0349 | | |

AUC and R@1%FPR are blank where a task declares no `positive_labels`. At the configured 0.7 the jailbreak gate catches 43.3% of unsafe rows while flagging 2.4% of safe ones, so safe class recall is 97.6%.

The two new columns are the point of the change, and they do not point the same way. AUC 0.8735 against accuracy 0.7267 says the ordering carries more signal than the accuracy suggests, so part of the loss is where the score sits, which the report files under `calibration` (ECE 0.2322, MCE 0.5247) and `threshold`. R@1%FPR 0.4165 says the rest is not recoverable by moving the threshold, so the `quality` findings stand and are what a retrain would have to answer.

The finer bands separate that loss by prompt length on the same 827 rows:

| band | rows | acc | macro F1 |
|---|---|---|---|
| chars<16 | 10 | 0.9000 | 0.4737 |
| chars<32 | 75 | 0.7867 | 0.6278 |
| chars<64 | 206 | 0.6650 | 0.6107 |
| chars<256 | 281 | 0.6690 | 0.6444 |
| chars<1024 | 132 | 0.7879 | 0.7806 |
| chars>=1024 | 123 | 0.8455 | 0.8358 |

The old first band held those 291 rows as one number. The band under 16 characters carries 10 rows, which is why the spread finding now needs 30, and it reads 0.6650 on `chars<64` against 0.8455 on `chars>=1024`.

On the binding, a copy of the published snapshot with one byte changed is refused by name: `config.json hashes to sha256:0d15fa45..., but mmbert32k-jailbreak-detector-merged records sha256:4bd0d4df...`, exit 1, and no result file is written. The unchanged copy measures normally. Asking for `llm-semantic-router/mmbert32k-factcheck-classifier-merged` while referencing the jailbreak manifest is refused before the download starts. A referencing run reports the same numbers as the run that minted the manifest, and `provenance.cli validate` still reports exactly one problem per bundle, the missing `run_ref`.

Every row measures fit rather than generalisation, since each task reads the split its upstream publishes. `domain` is the weakest of the three in that respect: the served intent classifier predates #3628 (`lastModified: 2026-03-06`) and no run manifest ties the checkpoint to a training run, which is the missing `run_ref` above.

`feedback` produces no row. It declares `tokenizer_class: TokenizersBackend`, which only transformers v5 provides, while its five siblings declare `PreTrainedTokenizerFast`, so no single version loads the whole portfolio. The previous description measured it on v5; this run uses one interpreter for the portfolio, so the artifact is reported as the runtime gap it is.

`FPR` joins the acronyms in `tools/linter/codespell/.codespell.ignorewords`, since codespell reads it as `for`. The per-row latency here counts rows rather than columns because `predict` takes a list of strings; the miscount #3642 reports is in `mom_collection_eval.py` and is untouched.

The measurements above are local, since they need the model weights and the training extras that the contract lane deliberately does not install.


FAUST-BENCHOU (commented):
Could we make the evaluation manifest record the artifact digest explicitly and validate it against the evaluated artifact? Otherwise a valid evaluation manifest could still describe a different artifact.

Xunzhuo:
Thanks for the substantial follow-up. #3197 is still awaiting acceptance and assignment, and this stack also depends on #3247. Please pause until the Router Models maintainers accept the bounded contract and the base PR is settled; then rebase and we’ll review the training surface.

subin9:
@FAUST-BENCHOU I think it does both for now too!

`artifact_ref` requires `id`, `revision` and `digest`, not just the id (`router-model-common-v1.schema.json`, `$defs.artifactRef`). Cross-reference validation then compares the recorded digest against the artifact manifest's `identity.digest` and fails on a mismatch (`crossref.py:198`), and recomputes `identity.digest` from the file list so a manifest cannot claim a digest its own files do not produce (`crossref.py:105`). `test_artifact_digest_must_match_the_file_list` and `test_evaluation_referencing_a_different_artifact_revision_fails` pin both.

The gap in what you are describing is that this is manifest against manifest. `build_artifact_manifest` hashes the real files at emission, but validation never re-reads them, so a bundle whose file list and digest were both edited consistently would pass. Adding an opt-in `--verify-files` to `provenance.cli` validate that re-hashes on disk would close that perhaps. Thank you!

@Xunzhuo Understood, pausing here too :)

adaamko (commented):
Right harness, should land. The identity table in the Purpose section alone justifies it, and taking class order from the artifact's own mapping is the fix that changes numbers.

Three things before the number this produces can serve as the baseline #2587 plugs into:

1. One external held-out set per safety task next to the in-distribution split. For jailbreak: Aegis 2.0 test and WildGuardMix test, macro-F1. `llm-semantic-router/jailbreak-detection-dataset:test` is the distribution the model was trained on, so it measures fit, not generalisation. Both are public, permissive licences, load with the same `load_dataset` call `baseline_tasks.py` already makes.

2. Safe-class recall as its own summary column. `_per_label` already computes it, so this is exposing a number not adding one. For a binary guard, safe-class recall is 1 minus the benign false-positive rate, and for a router that is the fraction of clean requests that get blocked or rerouted. Belongs next to unsafe-class recall, not buried in the per-label block.

3. Slice by language where the set carries one. `length_slices` is the right pattern, PolyGuardMix has a language column and the multilingual behaviour is where these models differ most.

Smaller: the gap categories (identity / coverage / calibration / threshold / quality) are good, keep them. `calibration` and `threshold` are integration gaps, `quality` is the model gap, which is the split #3194 asks for, worth one line in the README saying which category means retrain and which means fix config.

Once this lands i can follow up with the Aegis/WildGuard task specs as a PR on top, so the jailbreak row in #2587 runs here instead of on a private harness.

Xunzhuo (dismissed):

wilsonwu (dismissed):
LGTM

subin9:
@mergifyio dequeue

src/training/model_eval/quality_baseline.py
```diff
@@ -0,0 +1,409 @@
...
+    return parser.parse_args(argv)
+
+
+def resolve_measured_artifact(
+    args: argparse.Namespace, served: ServedArtifact
+) -> MeasuredArtifact:
+    """Decide which bytes this run scores, and warn when they are not the served ones."""
+    referenced = referenced_artifact(args.artifact_manifest)
+
+    if args.artifact_dir is not None:
```
Xunzhuo: **Bind the manifest to the measured bytes**

Impact: `--artifact-manifest` is accepted alongside an arbitrary local directory or separately selected download; the run then uses the manifest label mapping and emits its digest while scoring different bytes. Metrics for artifact A can therefore be attributed to artifact B.

Required change: Resolve measured bytes from the manifest identity or recompute and verify the repository, revision, and digest against `model_dir` before scoring, with local and downloaded mismatch tests.

Xunzhuo (commented):
Thanks for tightening the evidence pipeline. One provenance binding gap still makes the reported artifact identity unreliable.

adaamko (commented):
right shape, and the two new columns (AUC, R@1%FPR) are the ones that let two artifacts be compared without agreeing on a threshold, thats the piece #2587 needed.

one ask before it lands: the length bands are 64/256/1024 characters, and the region where the shipped detector fails is finer than that (subin9 on #2587: 31% benign blocked at 1-2 words, 9% at 3, 2% at 4-5). one band from 0 to 64 chars averages those together. either add cuts at 16 and 32, or make the boundaries word-based (1-2, 3-5, 6-12, 13+), same LENGTH_BOUNDARIES tuple, different numbers.

two follow-ups i'll take on top once this merges, so they are on record: the external held-out sets per safety task (Aegis 2.0 test, WildGuardMix test, with the near-duplicate pass first) as TaskSpecs, and a language slice once a set carries a language column. xunzhuo's manifest binding point stands, that is the last blocker i see.

src/training/model_eval/quality_baseline.py
subin9: Thank you, all fixed and added.

The manifest now picks the bytes. Its `identity` gives the repository and revision that get downloaded, its file list is the `allow_patterns`, and every file it lists is re-hashed from `model_dir` before scoring. A mismatch raises, so a downloaded snapshot and a local `--artifact-dir` are checked the same way, and an `--artifact-repo` the manifest does not describe is refused before the download starts.

`test_baseline_artifact.py` covers both mismatches, plus the revision pin and the wrong repository. Against the published artifact, a copy of the snapshot with one byte of `config.json` changed is refused by name and nothing is written.

Xunzhuo (commented):
Thanks for tightening the manifest binding. One supported checkpoint format is still omitted from the download closure.

src/training/model_eval/baseline_artifact.py
```diff
@@ -0,0 +1,137 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from artifact_inventory import ServedArtifact
+from provenance.crossref import verify_artifact_bytes
+from provenance.emit import resolve_hf_revision
+from provenance.manifest import load_manifest
+
+ARTIFACT_PATTERNS = ["*.json", "*.safetensors", "*.txt", "*.model"]
```
Xunzhuo: **Include `.bin` weights in artifact downloads**

Impact: The default `--artifact-repo` path downloads JSON, safetensors, text, and model files but not `*.bin`, even though the manifest builder and Transformers support sharded `pytorch_model.bin` checkpoints. A valid bin-only artifact therefore gets its index without its weight shards and fails before producing a baseline.

Required change: Include `*.bin` in the artifact download closure, preserving indexed shard resolution, and add a bin-only repository regression.

subin9: Thank you! all added.

Xunzhuo (approved):

Issue #3208, opened 2026-08-31
[Feature] Reject on_unknown + on_error conflict across config producers


guan404ming:
## Goal

Reject the decision-rule combination where rules.on_unknown and a nested condition on_error are both set, consistently across canonical config, DSL, CLI, and Dashboard producers.

## Proposed Workgroup

MoM & Routing

## Scope

- Keep the existing Go config-loader rule as the canonical semantic contract.
- Add the same conflict check to DSL validation.
- Add the cross-field check to the CLI Rules model.
- Prevent or clearly reject the conflicting Dashboard edit.
- Add parity tests proving every producer rejects the same invalid rule tree.

## Non-goals

- Changing the meaning of on_unknown or on_error.
- Adding a new fallback mode.
- Defining separate CLI or Dashboard semantics.
- Expanding the change into general decision-editor refactoring.

## Completion signal

Go config, DSL, CLI, and Dashboard all reject the conflicting combination with actionable errors, and parity tests protect the contract.

guan404ming:
Close this one and tracing in #3267

PR #3249, opened 2026-09-01
[Feature] Reject on_unknown with on_error across producers

guan404ming:
Related #3267

## Purpose

- Reject `rules.on_unknown` combined with a nested condition `on_error` in every config producer, matching the canonical Go config-loader rule from #3155.
- DSL: `Validate` emits a constraint diagnostic for any `on_error` leaf under a route that sets `on_unknown`.
- CLI: the `Rules` model rejects the combination recursively at parse time.
- Dashboard: the decision editor refuses to save the combination, including when a preserved nested rule tree already carries `on_error` and the form adds `on_unknown`.

## Test Plan

- Verify each producer rejects the same invalid rule tree with the shared error wording.
- Verify nested conditions are covered, not only root-level leaves.
- Verify either field alone still passes everywhere.

## Test Result

- `go test ./pkg/dsl/ ./pkg/config/`
- `pytest tests/test_cli_models.py tests/test_recipe_validation.py` (31 passed)
- `vitest run src/pages/configPageDecisionPreservation.test.ts` (6 passed)
- `gofmt`, `golangci-lint`, `prettier`, `black` clean on touched files


FAUST-BENCHOU (commented):
Could we add a runtime validation test as well? Since invalid configs may come from different sources, it would be good to ensure the loaded canonical config always rejects this combination.

Xunzhuo:
Thanks for the focused implementation. #3208 is still awaiting acceptance and assignment, so I’m holding the code review until the feature contract is accepted. Please pause here; once the issue is accepted, rebase and we’ll review the exact builder behavior and tests.

Xunzhuo (commented):
Thanks for carrying this through every producer. The dashboard save path, DSL validator, CLI model, and canonical Go loader now reject the same nested `on_unknown`/`on_error` conflict, with focused coverage for preserved trees and either field alone. This looks good to me. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

guan404ming:
Hi @Xunzhuo I think the main ci looks great now. Could you help take another look, thanks!

guan404ming:
Just pushed commit to resolve the conflict.

src/semantic-router/pkg/dsl/validator.go
```diff
@@ -342,6 +342,26 @@ func (v *Validator) walkBoolExpr(expr BoolExpr) {
...
+	case *BoolAnd:
+		v.checkOnUnknownOnErrorConflict(e.Left, context)
+		v.checkOnUnknownOnErrorConflict(e.Right, context)
+	case *BoolOr:
+		v.checkOnUnknownOnErrorConflict(e.Left, context)
+		v.checkOnUnknownOnErrorConflict(e.Right, context)
+	case *BoolNot:
+		v.checkOnUnknownOnErrorConflict(e.Expr, context)
+	case *SignalRefExpr:
+		if _, ok := getStringField(e.Fields, "on_error"); ok {
```
theohsiung: `getStringField` returns `ok=true` for an empty value, but the canonical rule tests `node.OnError != ""` (validator_decision.go:87). So `on_error: ""` is flagged here while `Compile` of the same source yields a config `ParseYAMLBytes` accepts. And because `hasBlockingDiagnostics` counts `DiagConstraint`, this false positive makes `appendRuntimeValidationDiagnostics` skip every TEST block in the file.

```suggestion
		if v, ok := getStringField(e.Fields, "on_error"); ok && v != "" {
```

src/vllm-sr/cli/models.py
```diff
@@ -731,6 +731,26 @@ def normalise_leaf_or_empty(cls, data):
             return rules
         return data
 
+    @model_validator(mode="after")
+    def validate_on_unknown_conflict(self):
+        if self.on_unknown is not None and _any_condition_sets_on_error(
```
theohsiung: Only `self.conditions` is walked, but `ruleTreeSetsOnError` checks the root node before recursing. `Rules(operator="AND", on_unknown="no_match", on_error="no_match", conditions=[{"type":"keyword","name":"x"}])` is accepted here and rejected by the loader. `Rules` declares no `on_error` field and no `extra="forbid"`, so pydantic drops the key — the user's guard is silently deleted rather than reported.

dashboard/frontend/src/pages/configPageSupport.ts
```diff
@@ -1407,6 +1407,25 @@ export function conditionHasNestedRules(condition: DecisionCondition): boolean {
...
+): DecisionRuleSet {
+  const rules = decisionRulesForSave(existing, next)
+  if (decisionRulesConflict(rules)) {
+    throw new Error('Condition on_error has no effect when on_unknown is set; remove one of them.')
+  }
+  return rules
+}
+
+export function decisionRulesConflict(rules: DecisionRuleSet): boolean {
+  return Boolean(rules.on_unknown) && (rules.conditions || []).some(conditionSetsOnError)
```
theohsiung: Same root-node gap. The preserve branch of `decisionRulesForSave` deep-clones `existing`, keeping a root `on_error` that `DecisionRuleSet` doesn't model, then sets `on_unknown` on it — the save succeeds and the router then refuses the config at startup. Checking the root as well as its children covers both shapes.

dashboard/frontend/src/pages/configPageSupport.ts
```diff
@@ -1407,6 +1407,25 @@ export function conditionHasNestedRules(condition: DecisionCondition): boolean {
   return Boolean(condition.operator || condition.conditions?.length)
 }
 
+export function decisionRulesForSaveChecked(
```
theohsiung: Minor: `decisionRulesForSave` now has no production caller but stays exported next to the checked variant, so a future save path can pick the unguarded name. Folding the check into it drops both the wrapper and the foot-gun.

src/semantic-router/pkg/dsl/policy_signal_roundtrip_test.go
```diff
@@ -188,6 +188,32 @@ ROUTE "x" (on_unknown = "allow") {
 	t.Fatalf("no diagnostic mentions on_unknown: %#v", diagnostics)
 }
 
+func TestValidateFlagsOnUnknownOnErrorConflict(t *testing.T) {
```
theohsiung: This `WHEN` is a single flat `*SignalRefExpr`, so `go test ./pkg/dsl/ -coverprofile` reports `checkOnUnknownOnErrorConflict` at 37.5% — `BoolAnd`/`BoolOr`/`BoolNot` never execute. Worth a nested case (`WHEN a(...) AND b(..., on_error: "no_match")`) plus the mirror negatives the Python and TS tests both have; the latter is what would have caught the empty-string over-firing above.

src/semantic-router/pkg/dsl/validator.go
```diff
@@ -541,6 +561,9 @@ func (v *Validator) checkRouteConstraints(r *RouteDecl) {
 			nil,
 		)
 	}
+	if r.OnUnknown != "" && r.When != nil {
```
theohsiung: `r.When != nil` is redundant — a nil `BoolExpr` matches no arm of the type switch, which is why `walkBoolExpr` is called unguarded.

src/semantic-router/pkg/config/validator_decision_unknown_test.go
```diff
@@ -42,3 +42,54 @@ func TestUnguardedClassifierConditions(t *testing.T) {
 		t.Fatalf("unguarded = %v", got)
 	}
 }
+
+func TestParseYAMLBytesRejectsOnUnknownOnErrorConflict(t *testing.T) {
```
theohsiung, line 46: `buildDecisionRefConfig` (decision_reference_validation_test.go:10) already wraps a decisions block in a valid v0.3 config. The classifier signal isn't needed either: the root conflict check fires before leaf recursion, which is why the table row at :17 uses a plain keyword leaf. Together that takes this to ~8 lines.

theohsiung (dismissed):
Verified the new checks against the canonical `ruleTreeSetsOnError`: two of the three don't match it yet, so the parity this PR is for isn't reached.

- DSL treats `on_error: ""` as set, flagging a tree the loader accepts — and since `DiagConstraint` is blocking, that also skips every TEST block in the file.
- CLI and Dashboard walk only `rules.conditions`, never the root node, so a root `on_error` alongside `on_unknown` passes both and fails at router startup.
- `checkOnUnknownOnErrorConflict` sits at 37.5% coverage with all three recursion arms unexecuted, so "nested conditions are covered" holds for the Python and TS tests only.

Question: `writeValidationDiagnostics` returns `errCount` only, and `Compile` never calls `Validate` — so `dsl validate` exits 0 and `dsl compile` still writes the conflicting YAML. #3208's "all reject" isn't reachable from the DSL side without changing that. Intentional, or a follow-up?

Optional: the rule is now hand-copied four times with three different wordings. `pkg/config` could export the checker plus the message and the DSL run it on `compileBoolExpr(r.When)` — the pattern `checkStructureSignalConstraints` already uses.

src/semantic-router/pkg/dsl/validator.go
guan404ming: Fixed: the DSL now compiles WHEN and runs the shared config checker, so empty strings pass.

src/vllm-sr/cli/models.py
guan404ming: Added root `on_error` to `Rules`, checked with the children, and `extra="forbid"` so extras are reported.

dashboard/frontend/src/pages/configPageSupport.ts
guan404ming: `DecisionRuleSet` now carries `on_error` and the conflict check starts at the root node.

dashboard/frontend/src/pages/configPageSupport.ts
guan404ming: Folded the check into `decisionRulesForSave` and removed the checked wrapper.

src/semantic-router/pkg/dsl/policy_signal_roundtrip_test.go
guan404ming: Added nested AND, OR with NOT, empty on_error, and each field alone; walker replaced by compiled tree.

src/semantic-router/pkg/dsl/validator.go
guan404ming: Removed.

src/semantic-router/pkg/config/validator_decision_unknown_test.go
guan404ming: Uses `buildDecisionRefConfig` with a declared classifier; a bare root `on_error` is caught first by #3323.

guan404ming:
Thanks, all points addressed in `ad37e4b16`. On `Compile` not calling `Validate`: not intentional. Follow-up tracked in #3674; this PR keeps the four producers in parity.

theohsiung (dismissed):
@guan404ming thanks for the instant fix. Re-checked at ad37e4b1. Confirmed fixed: the DSL empty-string case, the dead nil conjunct, the buildDecisionRefConfig fixture, the folded-in save check, and the test table — RuleTreeSetsOnError is now at 87.5% with all of AND / OR / NOT executing. #3674 and #3323 both say what you describe.

theohsiung (changes requested):
Four I could not square with the claims, all reproduced at this head:

cli/models.py:733 — Literal["no_match", "match"] makes an explicit on_error: "" a field-level error, while the Go loader accepts it (the canonical test is node.OnError != ""), and the CLI accepted it before this diff. Being field-level it also pre-empts validate_on_unknown_conflict, so the conflict message never shows.
configPageSupport.ts:1346 — existing?.conditions is not chained past existing, so a leaf-form rules ({type, name} — what Rules accepts and Go emits under conditions,omitempty) throws TypeError: Cannot read properties of undefined (reading 'some') on Save.
configPageSupport.ts:1346 — a root on_error only reaches the conflict check on the clone branch. A flat tree takes rules = next, and next never carries on_error, so the root guard is dropped silently and nothing is thrown.
cli/models.py:737 — normalise_leaf_or_empty projects a fixed key whitelist, so on the leaf path extra="forbid" never applies: unknown keys, a typo'd on_errror (the guard vanishes, no conflict reported), and even an entire conditions subtree are all dropped silently — the last one a shape the Go loader rejects outright.
Three smaller ones: Condition (models.py:655) is the one model in this region still without extra="forbid", so nested extras are dropped the same way; decisionRulesConflict (configPageSupport.ts:1367) gates on on_unknown, but the loader rejects a root on_error unconditionally, so that half still saves a config the router refuses at startup, where models.py:765 checks it unconditionally; and at dsl/validator.go:545 my earlier ask to drop r.When != nil was right for walkBoolExpr but not for this callee — compileBoolExpr reWHEN-less match-all route with on_unknown now manufactures one at 0:0, invisible only because thethrowaway compiler's errors are discarded, real ones included (a bad predicate value yields no diagnostic from Validate at all).

Test-side: configPageDecisionPreservation.test.ts:130 pins the conflict check on a childless OR trejects, so the tree it proves cannot load; and policy_signal_roundtrip_test.go:223 matches onlythe message substring, neverLevel, so a DiagConstraint→DiagWarningchange would keep it green whilehasBlockingDiagnostics` stops gating.

guan404ming:
Thanks, all fixed in e9b7faa6a.

- `models.py:733`: root `on_error` is `Optional[str]`, empty means unset.
- `configPageSupport.ts:1346`: `conditions?.some`, and a composite root's `on_error` now reaches the check on the flat path.
- `models.py:737`: leaf wrapper passes all keys through; `Condition` gets `extra="forbid"`.
- `decisionRulesConflict` → `decisionRulesError`, rejects root `on_error` unconditionally.
- `validator.go:545`: walks the WHEN tree directly, no throwaway compiler.
- Tests: loadable OR subtree, `DiagConstraint` asserted, plus match-all, unknown-key, and leaf-child cases.

Xunzhuo (approved):
The canonical loader, DSL, and CLI now reject the nested on_unknown/on_error conflict consistently while retaining unknown-field validation. The focused contract coverage and exact-head checks pass.

wilsonwu (approved):
LGTM

FAUST-BENCHOU (approved):

guan404ming:
@theohsiung could you help take another look, thanks!

PR #3250, opened 2026-09-01
[Feature] Add route action for prompt attack decisions

### LINKED ISSUE #3175: [Feature] Add a recipe action that routes detected prompt attacks to a safe fallback model
### Goal

Let a routing recipe explicitly send a detected prompt attack to a configured safe fallback model or route.

### Proposed Workgroup

Mixture-of-Models & Routing

### Scope

- Define a route action that references a validated destination.
- Apply it to an explicit prompt-guard predicate in a recipe.
- Define behavior when the destination is unavailable using existing recipe policy.
- Add end-to-end cases for detected and non-detected requests.

### Non-goals

- Implicit global prompt-guard enforcement.
- Backend unknown-result semantics; those belong to #3106.
- Managing provider credentials or model deployments.

Parent Epic: #3202

guan404ming:
Closes #3175

## Purpose

- A detected prompt attack could bypass guard decisions by pinning a model name; the only alternative was blocking.
- Add an explicit `route` decision action that sends matched requests to a validated safe destination, overriding a caller-pinned model.
- Require an explicit jailbreak condition in the decision rules, and validate the destination against `model_config`.
- A context-ineligible destination falls back to the decision's normal candidate policy; without an `action` block nothing changes.
- Mirror the action in the CLI `Decision` model and document keys and semantics in the reference config.

## Test Plan

- `go test ./pkg/config/ ./pkg/extproc/ ./pkg/decision/ ./pkg/dsl/`
- `python -m pytest src/vllm-sr/tests/test_cli_models.py`
- Unit coverage: action validation (7 cases), runtime override with and without action, context-ineligible fallback, CLI mirror.
- Not run: Kubernetes E2E lane (no local cluster); detected and non-detected paths are pinned at the extproc layer instead.

## Test Result

- All listed suites passed locally.

## Notes

- The DSL does not express `action` yet: `Decompile` drops the field, so a config using it is not DSL round-trippable. If that should land here rather than as a follow-up issue, happy to add it; otherwise I will file one.


FAUST-BENCHOU (commented):
Could we keep the DSL round-trip behavior consistent before exposing this action in config? Dropping `action` during decompile may silently remove a security-related routing override.

Xunzhuo (changes requested):
Thanks for taking on this security-sensitive path. Three merge blockers remain:

1. When the safe destination exceeds the context window, `decisionRouteActionDestination` falls through and `finalizeDecisionEvaluation` preserves a caller-pinned model. A detected attack can therefore bypass the route action with a long prompt. This path needs to fail closed or select an eligible safe destination.
2. The CLI validator only searches `rules.conditions`, so it rejects the direct `rules.type: jailbreak` form used by the new canonical YAML example. Please validate the root rule too and cover that exact shape.
3. The DSL decompiler drops `action`, so a config round trip silently removes the security behavior. Please preserve the action and add DSL parity plus the issue-required detected/non-detected E2E coverage.

guan404ming:
Fixed in dc0b23e6.

1. The route action is now terminal: ineligible destination falls back to an eligible candidate, otherwise the request fails closed. A pinned-model bypass test guards it.
2. The root `rules.type: jailbreak` form already passes; tests now pin the root and nested shapes.
3. The DSL round-trips `ACTION route "safe-model"` with validator parity.

E2E: a new `route-action` profile asserts the detected and benign cases.

Xunzhuo (changes requested):
The current head fixes all three blockers from my prior review: route actions now fail closed instead of preserving a pinned model, root-leaf jailbreak rules validate, and the DSL round trip preserves the action. The detected/benign E2E cases are also well scoped. One integration blocker remains: the new `route-action` profile is not registered in `tools/agent/test-domain-registry.yaml`. `make agent-report` selects only `envoy-ai-gateway`, and the exact-head CI likewise ran only that profile, so the new security behavior E2E has never executed. Please add the profile to the affected-E2E registry/matrix and provide a passing `route-action` run; then this should be ready for a final pass.

guan404ming:
Registered `route-action` with `selection: pr` in `tools/agent/test-domain-registry.yaml` and added it to the `e2e/README.md` profile inventory (681af7ca). `make agent-report` on this PR's changed files now selects `envoy-ai-gateway, route-action` for CI E2E.

Local `make e2e-test E2E_PROFILE=route-action` run on a fresh kind cluster:

```
[Test] decision="route_prompt_attack" model="openai/safe-model" status=200
✅ PASSED - route-action-detected (2.045886625s)
[Test] decision="default_decision" model="openai/mock-model" status=200
✅ PASSED - route-action-benign (2.045032792s)
Total: 2 | Passed: 2 | Failed: 0
```

The pushed head should also run the `route-action` profile in the affected E2E matrix.

Xunzhuo (approved):
The route-action profile is now registered for affected E2E, so the security behavior is selected and running on this head. The prior code blockers remain resolved. LGTM from this pass.

wilsonwu (approved):
LGTM

guan404ming:
Thanks for your review!

PR #3251, opened 2026-09-01
[Feature] Teach CLI tooling to read the decision-unresolved 503

### LINKED ISSUE #3209: [Feature] Teach CLI tooling to read the decision-unresolved 503
### Goal

Make the CLI evaluation, routing-trace, and tuning clients consume the Router's typed decision-unresolved HTTP 503 instead of crashing or discarding the decision error.

### Proposed Workgroup

### Scope

- Preserve decision_error in routing trace artifacts.
- Teach vllm-sr eval to render the maintained 503 error shape.
- Handle HTTPError in the tuning client and return a bounded typed failure.
- Add focused tests for the CLI message, trace artifact, and tuning-client behavior.
- Keep the Router-owned response shape unchanged.

##### Proposed Workgroup

### Non-goals

- Changing decision unknown-result semantics.
- Changing the Router's status code or response payload.
- Retrying or selecting a fallback model inside client tooling.
- Defining evaluation policy in the CLI.

### Completion signal

A fail-request decision 503 produces a readable CLI error, a trace containing decision_error, and a bounded tuning-client result without an uncaught exception.

guan404ming:
Closes #3209

## Purpose

- Preserve `decision_error` in routing trace artifacts as a sanitized, length-bounded field.
- Render the decision-unresolved 503 shape in `RouterManagementClient`, including the applied unknown policies, so `vllm-sr route preview` reports it.
- Handle `HTTPError` in the tuning `RouterClient`: parse the bounded error body, surface the payload, and report an `UNRESOLVED` typed probe result instead of an uncaught exception.
- Router-owned response shapes are unchanged.

## Test Plan

- Verify the trace artifact keeps a bounded `decision_error` and drops absent or invalid values.
- Verify the management client renders the 503 payload shape and keeps the legacy error envelope.
- Verify the tuning client returns the payload on HTTP errors and reports unresolved probes as bounded failures.

## Test Result

- `pytest tests/test_router_management_client.py tests/test_routing_trace.py tests/test_route_command.py` (41 passed)
- `pytest tools/agent/scripts/tuning/tests/test_client.py` (6 passed)
- `black` clean on touched files


FAUST-BENCHOU (commented):
Could we also add a backward compatibility test for older 503 responses without the new decision fields? It would help ensure existing router error handling remains unchanged.

Xunzhuo:
Thanks for the focused implementation. #3209 is still awaiting acceptance and assignment, so I’m holding the code review until the feature contract is accepted. Please pause here; once the issue is accepted, rebase and we’ll review the exact CLI behavior and tests.

FAUST-BENCHOU (commented):
LGTM. Nice to keep the CLI behavior aligned with the new decision-unresolved response contract.

guan404ming:
Added in d15e3bf7.

- This caught a real regression: the tuning client returned legacy envelope 503s as payloads, so probes reported `NONE` instead of the old `ERROR` row.
- Payloads now surface only when they carry `decision_error`; legacy 503s keep the previous typed-error path.
- Three new compatibility tests pin the legacy envelope behavior across the tuning client, `run_probes`, and the eval command.

Xunzhuo (changes requested):
run_probes still invokes result_adapter before checking decision_error. The calibration adapter always returns a result, so a decision-unresolved 503 becomes NONE and can even be marked correct as keep_7b instead of UNRESOLVED. Please handle decision_error before the adapter and add a regression using the calibration adapter, or another adapter that always returns a result.

guan404ming:
Thanks, good catch. run_probes now handles decision_error before the adapter, and a regression with the calibration adapter checks that a keep_7b probe on a 503 reports UNRESOLVED, not keep_7b.

guan404ming:
Rebased onto main in c9d061c06. #3724 renamed the eval command to route preview and moved error rendering into RouterManagementClient, so the 503 rendering now lives in _error_message with the same two tests (decision_error payload and legacy envelope).

Xunzhuo (approved):
The unresolved-decision check now runs before the calibration adapter, and the regression covers an adapter that always returns a result. Legacy error envelopes remain supported, and the exact-head checks pass.

wilsonwu (approved):
LGTM

Issue #3211, opened 2026-08-31
[Bug] vllm-sr CLI still ships unused serve.py and models_memory.py


FAUST-BENCHOU:
### Proposed Workgroup
Developer Experience & Ecosystem

### Describe the bug
The live `vllm-sr serve` command is the Click entry in `src/vllm-sr/cli/commands/runtime.py`, registered from `src/vllm-sr/cli/main.py`. A leftover module `src/vllm-sr/cli/commands/serve.py` is still in the tree but is never imported.

That leftover `serve_command()` parses config, writes a defaults note, optionally generates Envoy YAML, then stops at `# TODO: Start services` and prints "Next steps (to be implemented)". It is not a second serve path and must not be wired up.

`src/vllm-sr/cli/models_memory.py` is the same class of leftover: Pydantic memory-storage models that nothing imports. Live CLI schema for the memory plugin is `MemoryPluginConfig` in `src/vllm-sr/cli/models.py`.

### To Reproduce
1. Open `src/vllm-sr/cli/main.py` and confirm `REGISTERED_COMMANDS` imports `serve` from `cli.commands.runtime`, not `cli.commands.serve`.
2. Search the repo for `from cli.commands.serve`, `cli.commands.serve`, and `serve_command(`. The only definitions are inside `commands/serve.py` itself.
3. Search for `models_memory`, `MemoryMilvusConfig`, and `MemoryValkeyConfig` outside `models_memory.py`. There are no imports.
4. Run `vllm-sr serve --help` (or the CLI unit tests that exercise `cli.commands.runtime.serve`) and confirm help/text comes from `runtime.py`.

### Expected behavior
The CLI package should only keep modules that the live command graph or tests import. Dead leftovers should be deleted. `vllm-sr serve` must keep using `commands/runtime.py` and the local image flow.

Suggested change:
- Delete `src/vllm-sr/cli/commands/serve.py`.
- Delete `src/vllm-sr/cli/models_memory.py`.
- Re-search to confirm no remaining imports or path references.
- Add or adjust a CLI test that the registered `serve` command is `cli.commands.runtime.serve`.

### Impact
New contributors can mistake `serve.py` for the real serve implementation. The unused files add review noise and a fake second entrypoint.

Workaround: ignore those files and edit `commands/runtime.py` instead.

### Additional context
Out of scope:
[...]

PR #3260, opened 2026-09-01
[Bug] Use an effective complexity threshold in signal examples

### LINKED ISSUE #3246: [Bug] Complexity signal examples use an ineffective threshold of 0.70–0.75
### Proposed Workgroup

MoM & Routing

### Describe the bug

The `threshold` on a `complexity` signal rule is compared against the **margin** between the two prototype banks, not against a similarity score.

`src/semantic-router/pkg/classification/complexity_rule_scoring.go:41`:

```go
return hardScore, easyScore, hardScore.Score - easyScore.Score
```

`src/semantic-router/pkg/classification/complexity_rule_scoring.go:73-81`:

```go
func classifyComplexityDifficulty(threshold float32, signal float64) string {
	if signal > float64(threshold) {
		return "hard"
	}
	if signal < -float64(threshold) {
		return "easy"
	}
	return "medium"
}
```
In testing with the shipped candidate banks, none of the measured prompts, including prompts closely matching the hard candidates, exceeded 0.75. They were therefore classified as medium rather than hard.

The failure is silent. Configuration validates, the router starts with `embedding_ready:true`, every request returns `200`, and no error or warning is logged. The only symptom is that the escalation model is never selected.

Affected files, all reference/example surfaces:

| File | Line | Value |
| --- | --- | --- |
| `website/docs/tutorials/signal/learned/complexity.md` | 34 | `0.75` |
| `config/fragments/signal/complexity/escalation.yaml` | 5 | `0.75` |
| `config/config.yaml` | 366 | `0.75` |
| `deploy/kserve/example-multi-model-config.yaml` | 63 | `0.7` |

For contrast, every complexity threshold in a maintained recipe is in the reachable band:

| Recipe | Thresholds |
| --- | --- |
| `agent` | 0.14, 0.12, 0.12, 0.12 |
| `balance` | 0.14, 0.12, 0.12, 0.12, 0.12, 0.12 |
| `privacy` | 0.12, 0.12 |
| `multi-objective` | 0.08, 0.15 |
| `mom-v1` (config + CLI assets) | 0.08, 0.10, 0.10, 0.15 |

### To Reproduce

Deploy two backends and use the shipped example signal unchanged, i.e. the rule from `config/fragments/signal/complexity/escalation.yaml` with a decision on `needs_reasoning:hard` at a higher priority and an unconditional fallback decision at a lower priority.

Send a prompt that quotes the hard candidates verbatim:

```bash
curl -s -D - -o /dev/null http://127.0.0.1:8899/v1/chat/completions \
  -H 'content-type: application/json' \
  -H 'x-vsr-debug: true' \
  -d '{
    "model": "vllm-sr/auto",
    "messages": [{"role": "user", "content": "Solve this step by step and compare multiple tradeoffs: analyze the root cause of cascading failures in a distributed queue."}],
    "max_tokens": 3000
  }'
```

Observed with `threshold: 0.75`:

```http
x-vsr-selected-decision: routine-traffic
x-vsr-selected-model: routine-model
x-vsr-matched-complexity: needs_reasoning:medium
```

Changing only the threshold to `0.10` and restarting:

```http
x-vsr-selected-decision: escalate-hard-prompts
x-vsr-selected-model: reasoning-model
x-vsr-matched-complexity: needs_reasoning:hard
```

### Measured signals

Eight prompts through a running router, scores read from `vllm-sr logs router | grep "Complexity rule"`. Candidate banks are the shipped example banks, unmodified.

| Prompt | hard | easy | signal | @0.10 | @0.75 |
| --- | --- | --- | --- | --- | --- |
| Give me a quick summary of what HTTP is. | 0.434 | 0.595 | -0.160 | easy | medium |
| What port does SSH use? | 0.361 | 0.361 | -0.001 | medium | medium |
| Rewrite this sentence to be shorter. | 0.428 | 0.625 | -0.197 | easy | medium |
| hey calculate my assets based on all the tax reports I have | 0.469 | 0.404 | +0.064 | medium | medium |
| Explain how a database index works. | 0.494 | 0.470 | +0.025 | medium | medium |
| Solve this step by step and compare multiple tradeoffs: analyze the root cause ... | 0.578 | 0.422 | +0.156 | **hard** | medium |
| Analyze the root cause of this outage and compare multiple tradeoffs for the fix. | 0.642 | 0.491 | +0.151 | **hard** | medium |
| Design a rate limiter, solve this step by step, and analyze the root cause of contention ... | 0.586 | 0.496 | +0.090 | medium | medium |

- Largest absolute signal observed: `0.197`
- Threshold required to fire: `0.750`
- Escalation rate at `0.10`: 2/8
- Escalation rate at `0.75`: **0/8**

Individual bank scores span `0.36`-`0.64`, but the margin never exceeds `0.197`. Row 2 shows the shared floor directly: a prompt unrelated to both banks scores `0.361` against each, and the margin cancels to `-0.001`.

Rows 6 and 7 are the relevant cases. They contain the hard-bank candidate phrases verbatim and still do not escalate at `0.75`.

### Environment

- Semantic Router `0.3.0.dev20260809115245`, `vllm-sr serve`
- Two vLLM backends, `Qwen3.6-27B-FP8` and `Qwen3.8-27B-FP8`
- Single AMD Instinct MI300X, ROCm
- Default prototype scoring: `best_weight` 0.75, `top_m` 2
  (`src/semantic-router/pkg/config/prototype_scoring_config.go:27,33`)

### Expected behavior

The shipped examples should use a threshold within the reachable range of the margin, consistent with the maintained recipes, so that a user who copies the documented rule sees escalation occur.

### Impact

Affects anyone starting from the complexity signal documentation, the config fragment, the reference config, or the KServe example. These are the entry points for new users of the signal.

The consequence is a router that appears healthy while never escalating, so a provisioned escalation model serves no traffic. Because there is no error, the condition is likely to go unnoticed without inspecting `x-vsr-matched-complexity` under `x-vsr-debug: true` or per-model request counts.

Workaround: set the threshold into the `0.08`-`0.15` band used by the recipes, then calibrate against representative traffic by reading logged `signal` values.

### Additional context

The complexity `threshold` is applied to the margin `hard_score - easy_score`, unlike similarly named similarity thresholds elsewhere in the configuration. Documenting this distinction would help prevent example thresholds from being confused with similarity cutoffs.

The proposed fix should update the four affected examples and briefly explain that users need to calibrate the margin threshold against representative traffic. 

I’m happy to submit the fix after the issue is accepted.

Mahdi-CV:
Closes #3246

## Purpose

The `complexity` signal derives its difficulty band from the **margin** between
the hard and easy prototype banks, not from an absolute similarity score:

```go
// src/semantic-router/pkg/classification/complexity_rule_scoring.go:41
return hardScore, easyScore, hardScore.Score - easyScore.Score
```

```go
// src/semantic-router/pkg/classification/complexity_rule_scoring.go:73-81
func classifyComplexityDifficulty(threshold float32, signal float64) string {
	if signal > float64(threshold) {
		return "hard"
	}
	if signal < -float64(threshold) {
		return "easy"
	}
	return "medium"
}
```

Because the two banks can assign similar baseline scores, subtracting them can
produce a margin much smaller than either individual score.

The shipped examples set `threshold` to `0.70` or `0.75`. In the calibration run
recorded in #3246, the largest absolute margin observed across eight prompts was
`0.197`, and no prompt reached the `hard` or `easy` band at `0.75`. Prompts
containing the hard candidate phrases verbatim were classified `medium`, so the
`needs_reasoning:hard` decision condition did not match and the escalation model
was never selected. A reader who copies one of these examples gets a complexity
rule that appears healthy while never escalating.

For contrast, maintained recipe configurations use complexity thresholds in the
`0.08`-`0.15` range, including the agent, balance, privacy, multi-objective,
and built-in mom-v1 recipes.

This PR sets the example thresholds to `0.10` and documents that `threshold` is
a margin, presented as a starting point to calibrate rather than a universal
default.

**Scope note.** A repository-wide scan after filing #3246 found two additional
active occurrences in `bench/cpu-vs-gpu/config-bench-candle.yaml`. This PR
updates all active configuration and documentation surfaces.

Historical v0.3 English and translated documentation snapshots are intentionally
unchanged because they describe the configuration shipped with that release.

Affected modules: `config`, `deploy`, `bench`, `website`.
Owning Workgroup: `wg/mom-routing` (per issue #3246).

### Files changed

| File | Change |
| --- | --- |
| `config/config.yaml` | `needs_reasoning` threshold `0.75` -> `0.10` |
| `config/fragments/signal/complexity/escalation.yaml` | `needs_reasoning` threshold `0.75` -> `0.10` |
| `deploy/kserve/example-multi-model-config.yaml` | `needs_reasoning` threshold `0.7` -> `0.10` |
| `bench/cpu-vs-gpu/config-bench-candle.yaml` | `needs_reasoning` threshold `0.7` -> `0.10` (two occurrences) |
| `website/docs/tutorials/signal/learned/complexity.md` | threshold `0.75` -> `0.10`, plus a section explaining the margin semantics and the need to calibrate |

## Test Plan

1. Confirm no active complexity rule retains an ineffective threshold, using a
   YAML parser that also descends into ConfigMap-embedded `config.yaml` strings
   and Markdown fenced blocks (a plain `grep` misreads nearby
   `similarity_threshold`, `best_weight`, and `embeddings` values).
2. Verify the changed YAML still parses.
3. Run the reference config contract test:
   `go test ./pkg/config/... -run TestReferenceConfig -count=1`
4. Run the harness lint gate for the changed files:
   `make agent-ci-lint CHANGED_FILES="..."`
5. Run the docs gate: `make agent-docs-ci-gate AGENT_BASE_REF=origin/main`
6. Ask the harness what this change surface requires:
   `make agent-report ENV=cpu CHANGED_FILES="..."`
7. Check whitespace: `git diff --check origin/main..HEAD`

## Test Result

**1. Structural scan** — 27 complexity rules found across all YAML (including
ConfigMap-embedded configs). After the change, **0** active rules have a
threshold above `0.5`. The only remaining occurrences are the three intentionally
excluded historical snapshots under `website/versioned_docs/version-v0.3/` and
`website/i18n/zh-Hans/`.

**2. YAML parses**

```text
OK  config/config.yaml (1 docs)
OK  config/fragments/signal/complexity/escalation.yaml (1 docs)
OK  deploy/kserve/example-multi-model-config.yaml (1 docs)
OK  bench/cpu-vs-gpu/config-bench-candle.yaml (49 docs)
```

**3. Reference config contract test** — passes

```text
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/config  0.075s
```

**4. `make agent-ci-lint`** — exit `0`

```text
md fmt.......................................Passed
yaml/yml fmt.................................Passed
```

**5. `make agent-docs-ci-gate AGENT_BASE_REF=origin/main`** — exit `2`, blocked
by a pre-existing failure unrelated to this change. See "Blocker" below.

**6. `make agent-report`** — classifies the change as `documentation-only`;
impacted surfaces `router_config_contract, docs_examples`; required surfaces
`none`; fast tests `none`; feature tests `none`; local smoke `not required`.

**7. `git diff --check origin/main..HEAD`** — clean.

**Measured behavior.** The calibration data is the eight-prompt table recorded in
#3246, measured on a running router with the shipped candidate banks and read
from the router's own `complexity_rule_scoring.go:101` log line:

- Largest absolute margin observed: `0.197`
- Bands at `0.75`: **0 of 8** non-medium
- Bands at `0.10`: **4 of 8** non-medium (2 `easy`, 2 `hard`)

Rows 6 and 7 of that table are the decisive cases: they contain the hard-bank
candidate phrases verbatim, score `+0.156` and `+0.151`, and still do not reach
the `hard` band at `0.75`. Row 2 shows the mechanism directly, with `hard` and
`easy` both `0.361` cancelling to `-0.001`.

**Remaining risk.** `0.10` matches the calibrated range already shipped in the
maintained recipes, but the correct threshold depends on the candidate banks,
the embedding model, and the traffic. The documentation change states this and
directs readers to measure rather than implying `0.10` is universal.

**Blocker encountered (pre-existing, unrelated to this PR).**
`make agent-docs-ci-gate AGENT_BASE_REF=origin/main` fails during
`agent-validate` with:

```text
ERROR: Pattern has no matches: src/vllm-sr/**/*memory*
```

The pattern is `tools/agent/test-domain-registry.yaml:187`. Commit `2e79a811`
("Remove unused cli serve.py and models_memory.py (#3211)") deleted
`models_memory.py`, the only file under `src/vllm-sr/` matching that glob, so the
registry entry no longer resolves. This reproduces on a pristine `origin/main`
with no local changes and is not introduced by this PR; fixing the registry entry
belongs to a separate harness change. The focused gate for the changed files,
`make agent-ci-lint`, passes.

---
<details>

- [x] PR title begins with exactly one bracketed category: `[Bug]`
- [x] Title does not stack prefixes
- [x] Links accepted issue #3246 with exactly one owner (`wg/mom-routing`)
- [x] Commit signed off with `git commit -s`
- [x] Purpose, Test Plan, and Test Result reflect actual scope, commands, and blockers

</details>

### DIFF
--- a/bench/cpu-vs-gpu/config-bench-candle.yaml
+++ b/bench/cpu-vs-gpu/config-bench-candle.yaml
@@ -898,7 +898,7 @@ data:
               - other
         complexity:
           - name: needs_reasoning
-            threshold: 0.7
+            threshold: 0.10
             description: Escalate difficult or multi-step prompts.
             hard:
               candidates:
@@ -5190,7 +5190,7 @@ data:
               - other
         complexity:
           - name: needs_reasoning
-            threshold: 0.7
+            threshold: 0.10
             description: Escalate difficult or multi-step prompts.
             hard:
               candidates:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -363,7 +363,7 @@ routing:
 
     complexity:
       - name: needs_reasoning
-        threshold: 0.75
+        threshold: 0.10
         description: Escalate multi-step reasoning or synthesis-heavy prompts.
         hard:
           candidates:
--- a/config/fragments/signal/complexity/escalation.yaml
+++ b/config/fragments/signal/complexity/escalation.yaml
@@ -2,7 +2,7 @@ routing:
   signals:
     complexity:
       - name: needs_reasoning
-        threshold: 0.75
+        threshold: 0.10
         description: Escalate multi-step reasoning or synthesis-heavy prompts.
         hard:
           candidates:
--- a/deploy/kserve/example-multi-model-config.yaml
+++ b/deploy/kserve/example-multi-model-config.yaml
@@ -60,7 +60,7 @@ data:
               - other
         complexity:
           - name: needs_reasoning
-            threshold: 0.7
+            threshold: 0.10
             description: Escalate difficult or multi-step prompts.
             hard:
               candidates:
--- a/website/docs/tutorials/signal/learned/complexity.md
+++ b/website/docs/tutorials/signal/learned/complexity.md
@@ -31,7 +31,7 @@ routing:
   signals:
     complexity:
       - name: needs_reasoning
-        threshold: 0.75
+        threshold: 0.10
         description: Escalate multi-step reasoning or synthesis-heavy prompts.
         hard:
           candidates:
@@ -44,6 +44,27 @@ routing:
             - quick summary
             - simple rewrite
 ```
+
+`threshold` is a margin, not a similarity cutoff. The router scores the request
+against the `hard` and the `easy` candidate banks, then compares the two:
+
+```text
+signal = hard_bank_score - easy_bank_score
+
+signal >  threshold  -> hard
+signal < -threshold  -> easy
+otherwise            -> medium
+```
+
+Because the two banks can assign similar baseline scores, subtracting their
+scores may produce a margin much smaller than either individual score. In one
+calibration run using the candidate banks above, the largest observed absolute
+margin across eight prompts was `0.197`; none reached the `hard` or `easy` band
+with a threshold of `0.75`.
+
+Treat `0.10` as a starting point rather than a universal default. Measure the
+margin on representative traffic and tune the threshold for the configured
+candidate banks and embedding model.
 
 A rule emits a suffixed name. Decisions must reference
 `<rule>:easy`, `<rule>:medium`, or `<rule>:hard`:


Xunzhuo (approved):
LGTM, thanks @Mahdi-CV

PR #3266, opened 2026-09-01
[Bug] Apply shared Candle semantic input validation across native and stub builds

### LINKED ISSUE #2675: [Bug] Apply shared Candle semantic input validation across native and stub builds
### Proposed Workgroup

### Describe the bug

Semantic-rule inputs do not consistently enforce target dimension, top-k bounds, and tensor-shape checks across Candle-enabled and stub builds.

### To Reproduce

### Expected behavior

All build variants should apply the same validation before inference and return the same typed client error for invalid input.

### Impact

##### Completion signal

Shared contract tests cover valid and invalid dimensions, top-k values, shapes, and empty inputs across build variants.

Parent Epic: #2396

ZireaelK:
Closes #2675

> **Note**: This PR is a follow-up that builds on top of #2674 (which introduced the build-tag-free validator for presence/NUL checks). Once #2674 is merged into `main`, this branch will be rebased onto `main` and marked as Ready for Review.

## Purpose

Semantic-rule inputs (target dimensions, top-k bounds, tensor shapes, and similarity candidate slices) previously lacked consistent pre-dispatch validation across Candle CGO native and non-CGO stub builds. As a result, malformed inputs could either trigger unexpected backend FFI errors or be handled inconsistently depending on whether the native backend was compiled in.

This PR extends the build-tag-free shared validator (`candle-binding/validation.go`) to enforce semantic input constraints uniformly across both build modes before backend dispatch:
- **Target Dimension (`targetDim`)**: rejects negative dimensions (`targetDim < 0`), while preserving `0` as the default/auto dimension sentinel across `MultiModalEncode*`, `GetEmbedding*`, and similarity APIs.
- **Top-K Bounds (`topK`)**: rejects negative top-k values (`topK < 0`), while allowing `0` for return-all/default behavior in `CalculateSimilarityBatch`.
- **Tensor Shape Constraints**:
  - Image tensors: validates `pixelData` non-empty, `height > 0`, `width > 0`, and ensures `len(pixelData) == 3 * height * width`.
  - Audio Mel tensors: validates `melData` non-empty, `nMels > 0`, `timeFrames > 0`, and ensures `len(melData) == nMels * timeFrames`.
  - Image bytes: validates non-empty slice and non-negative target dimension.
- **Candidate Slice & Model Type**: validates candidate arrays are non-empty and contain no empty or NUL-containing strings; validates allowed model types in batch and pairwise similarity APIs.
- **Native & Stub Alignment**: wires validation helpers into both `candle-binding/semantic-router.go` and `candle-binding/semantic-router_mock.go` (the fail-closed stub now validates before returning `ErrBackendUnavailable`).

Affected module: `candle-binding` (owned by `wg/router-models-inference-runtime`, under parent epic #2396).

## Test Plan

```bash
# 1. Non-CGO stub mode tests (verifies pre-dispatch validation in fail-closed build)
cd candle-binding && CGO_ENABLED=0 go test -v .

# 2. Native CGO build validation tests
cd candle-binding && cargo build --no-default-features
cd candle-binding && LD_LIBRARY_PATH="./target/debug" CGO_ENABLED=1 CGO_LDFLAGS="-L./target/debug" go test -v -run "TestValidate|TestMultiModalValidation|TestSharedEmbeddingValidation" .

# 3. Formatting & Linting
cd candle-binding && go vet . && gofmt -l .
```

## Test Result

- `CGO_ENABLED=0 go test -v .`: All test suites (including `TestValidateSemanticHelpers`, `TestMultiModalValidationRunsInBothModes`, and `TestSharedEmbeddingValidationRunsInBothModes`) pass.
- `CGO_ENABLED=1 go test`: Pre-dispatch validation correctly intercepts malformed dimensions, shapes, and inputs before native FFI dispatch.
- `go vet .` and `gofmt -l .`: Clean with 0 issues.
- Commit has been signed off with `git commit -s`.

---
<details>

</details>


FAUST-BENCHOU (commented):
Could we keep the validation rules documented close to the public API contract? Some constraints are important behavior and may be easy to change accidentally later.

ZireaelK:
> Could we keep the validation rules documented close to the public API contract? Some constraints are important behavior and may be easy to change accidentally later.

Just to make sure I understand correctly: did you mean adding the explicit parameter constraints and edge-case behaviors to the exported doc comments in semantic-router.go?

Xunzhuo (changes requested):
Yes, the exported doc comments are the right place for these constraints. One blocker also remains in the validation itself: positive Go `int` values outside the `C.int` range are accepted and then silently narrowed at the FFI boundary, and the tensor-size products can overflow before the length check. For example, `targetDim == 1<<32` reaches native code as the default sentinel `0`. Please reject values that cannot be represented by the native ABI, use overflow-safe shape multiplication, and cover those boundaries before this is marked ready.

candle-binding/validation_test.go
```diff
@@ -0,0 +1,492 @@
...
+			t.Fatalf("expected 0 to be valid, got %v", err)
+		}
+		if err := validateTargetDim(128); err != nil {
+			t.Fatalf("expected 128 to be valid, got %v", err)
+		}
+		if err := validateTargetDim(-1); err == nil || !strings.Contains(err.Error(), "targetDim cannot be negative") {
+			t.Fatalf("expected negative error, got %v", err)
+		}
+		if math.MaxInt > math.MaxInt32 {
+			huge := int(int64(1) << 32)
```
Xunzhuo: **Keep the cross-build test source representable**

`int(int64(1) << 32)` is still a compile-time constant conversion, so a 32-bit Go build rejects this file before the runtime `math.MaxInt` guard can skip the case. That defeats this no-build-tag suite's cross-build contract. Please make the 64-bit value non-constant before converting (or isolate the fixture with an architecture constraint), and add a 32-bit compile gate.

candle-binding/semantic-router.go
```diff
@@ -871,14 +871,21 @@ func InitEmbeddingModelsBatched(qwen3ModelPath string, maxBatchSize int, maxWait
...
 //
 // Returns:
 //   - *EmbeddingOutput: Embedding output with metadata
-//   - error: Non-nil if embedding generation fails
+//   - error: Non-nil if validation fails (empty/NUL text, negative or out-of-range targetDim) or embedding generation fails
 func GetEmbeddingBatched(text string, modelType string, targetDim int) (*EmbeddingOutput, error) {
+	if err := validateRequiredText("text", text); err != nil {
+		return nil, err
+	}
+	if err := validateTargetDim(targetDim); err != nil {
```
Xunzhuo, line 888: **Validate the batched model type before dispatch**

`GetEmbeddingBatched` documents and the Rust FFI enforces that only `qwen3` is supported, but shared validation stops after `targetDim`. An invalid or NUL-containing model type therefore reaches native FFI and becomes a generic generation failure, while the stub returns `ErrBackendUnavailable`, so malformed input still differs across build modes. Please add a build-neutral batched-model validator and cover invalid, empty, and NUL model types through this public API in both variants.

Xunzhuo (commented):
Thanks for tightening the shared validation path and fixing the earlier ABI-width and shape-overflow issues. Two cross-build gaps remain on this head.

candle-binding/validation_test.go
```diff
@@ -0,0 +1,547 @@
...
+	if testing.Short() {
+		t.Skip("skipping cross-architecture compilation gate in short mode")
+	}
+	goBin, err := exec.LookPath("go")
+	if err != nil {
+		t.Skip("go binary not found in PATH")
+	}
+	outputBinary := filepath.Join(t.TempDir(), "candle_binding_32bit.test")
+	cmd := exec.Command(goBin, "test", "-c", "-o", outputBinary, ".")
+	cmd.Env = append(os.Environ(), "CGO_ENABLED=0", "GOARCH=386")
```
Xunzhuo: **[P2] Pin a supported OS for the 32-bit compile gate**

This forces `GOARCH=386` while inheriting `GOOS`. On macOS, a normal non-short package test therefore invokes the unsupported `darwin/386` target and fails. Running this test with an inherited Darwin target reproduces `go: unsupported GOOS/GOARCH pair darwin/386`. Since the binary is only compiled, set a supported target such as `GOOS=linux` with `GOARCH=386`, and cover the inherited-environment case.

Xunzhuo (changes requested):
The shared validation fixes address the earlier ABI and model-type findings. The new compile gate still fails on macOS; please make its target OS explicit and add the environment regression.

Xunzhuo (approved):
The shared validation and compile regressions are resolved. Focused tests now pass in both CGO and non-CGO builds, including the Linux/386 check with an inherited Darwin environment; the current-head Core and required checks are green. No remaining code blocker found.

wilsonwu (approved):
LGTM

PR #3268, opened 2026-09-01
[Feature] Wire admission gate into Router Model inference

### LINKED ISSUE #3159: [Feature] Add a per-deployment admission gate for Router Model inference

guan404ming:
Closes #3159

## Purpose

- Wire the admission gate from #3201 into Router Model inference via a deployment-keyed registry; all seven system deployments are decorated, unconfigured ones admit through a no-op.
- A shed inference surfaces `admission.ErrQueueFull` as a classify error under the existing `on_error` / `on_unknown` policy.
- Remove the superseded signal-evaluation load gate and emit per-deployment admission metrics.
- Deprecate `batch_classification.concurrency_threshold` / `max_concurrency`: a non-zero `max_concurrency` migrates at load time into wait-mode admission defaults for deployments without an explicit entry, with a deprecation warning; explicit admission entries win. `concurrency_threshold` has no equivalent (the gate no longer bypasses low load).

## Notes

- The operator CRD and dashboard config schema still expose the deprecated fields; cleaning those versioned surfaces is left to their owners as a follow-up.

## Test Plan

- `go test ./pkg/admission/ ./pkg/classification/ ./pkg/extproc/ ./pkg/config/`
- Not run: load behavior against real models.

## Test Result

- Passed.


FAUST-BENCHOU (commented):
Could we document why the legacy max_concurrency value is safe to migrate to every deployment? Different Router Models may have very different memory/latency characteristics, so a single global limit might not be an equivalent replacement.

guan404ming:
Thanks for asking. `max_concurrency` was already one global value for every model. Migration only preserves it; explicit `admission` entries override per deployment. Will note in description.

Xunzhuo (dismissed):
Thanks for carrying the admission contract through the model paths. One blocker remains: `evaluateDomainSignal` falls back to `Classify` after any `ClassifyWithProbabilities` error. The first call can now return `admission.ErrQueueFull` or a context cancellation, so this path immediately reacquires the same deployment gate; a shed request can run if a slot opens between attempts, and wait-mode can incur a second acquisition instead of surfacing the typed error to `on_error` / `on_unknown`. Please short-circuit typed admission/context errors before the capability fallback and add a test that proves overload causes only one admission/backend attempt. Reviewed at `e8581c380ac00ea249dfc9f5a77c21807ab2910c`.

guan404ming:
Thanks for the catch. Fixed in 798ccc79: admission and context errors now skip the `Classify` fallback. `TestDomainSignalDoesNotRetryAdmissionErrors` proves one gate acquisition and no backend call.

FAUST-BENCHOU:
pls resolve confilts

guan404ming:
Just resolved, thanks!

Xunzhuo (dismissed):
The typed admission and context errors now short-circuit the probability fallback, and the regression pins one gate acquisition with no backend retry. The prior blocker is resolved. LGTM.

wilsonwu (dismissed):
LGTM

wilsonwu:
@guan404ming please resolve conflicts, thanks!

guan404ming:
Just resolved, thanks!

Xunzhuo (dismissed):
The migration runs only when legacy `max_concurrency` is present, but this PR removes that field from Helm defaults and maintained configs without adding equivalent admission entries. Those generated/default configs skip migration and silently lose their previous concurrency bound. Please preserve the deprecated field for migration or translate every maintained config, with a generated/default-config regression.

guan404ming:
Thanks! Restored the deprecated `max_concurrency` in every maintained config so migration keeps the bound, plus a regression test that parses each default config and checks every deployment is bounded.

Xunzhuo (commented):
Thanks—the maintained configs now retain `max_concurrency`, and the asset regression proves migration preserves a bound for every deployment. That resolves my blocker on this head; I do not see another code issue in the follow-up. The branch is still behind current main, so please rebase and let the exact new head go green for the final pass.

Xunzhuo (changes requested):
One blocker remains: several newly gated inference paths discard the caller context. Nil contexts become context.Background(), while direct intent and PII routes and internal fact-check, feedback, and NLI paths still call the classifiers without request context. With migrated on_overflow: wait gates and a zero queue timeout, a canceled request can wait indefinitely and later consume a model slot. Please thread request context through every gated inference path and add cancellation and deadline coverage for a direct API path and an internal signal path.

guan404ming:
Thanks, agreed. Dropped the `context.Background()` fallback and threaded request context through all gated paths. Added cancel and deadline tests for the direct PII route and the internal fact-check signal.

Xunzhuo (approved):

wilsonwu (approved):
LGTM

PR #3287, opened 2026-09-01
[Bug] roll back applied blocks on recovery-store failure in context_compression

### LINKED ISSUE #3278: [Bug] context_compression leaves partially compressed body on recovery-store failure
### Describe the bug

`context_compression` with the default `failure_mode: fail_open` forwards a partially compressed request body when a recoverable target's recovery-store operation fails after earlier candidates were already applied. The compression result reports `Applied = false` and `Failure != nil`, but the body on the wire has been mutated.

Parent task: #2984
Related epic criterion: #2984 — "Unsupported capabilities and runtime failures have an explicit safe fallback"; "Lossy transformation must not be silent or bypass protected content".

### Proposed Workgroup

Agentic & Context

### Current-main evidence

`Service.executePlan` (`src/semantic-router/pkg/contextcompression/service.go`) iterates compression candidates and calls `applyCandidate` per block. On any `applyErr` it returns the error without restoring blocks mutated by earlier candidates in the same batch.

`applyCandidate` mutates a block before the recovery step can fail:

1. Extractive target — `block.SetText(compressed.Content)` at `service.go:418`, returns `(true, …, nil)`.
2. A later recoverable target — `applyRecoverableCandidate` fails: store `Put` error (`service.go:476`), store nil/disabled (`service.go:460`), or `recovery.MaxBytesPerRequest` exceeded (`service.go:455`).
3. `executePlan` returns the error. Blocks applied in step 1 are not restored.

`Service.failureResult` (`service.go:496`) sets `TokensAfter = OriginalTokens` and `Failure = err` but performs no rollback. Block restoration (`candidate.block.SetText(candidate.originalText)`) only exists in `finalizeResult` (`service.go:155`) for the budget-not-reduced branch.

`req_filter_context_compression.go:86`: `result.Failure != nil` → `semanticCompressionFailure` → `fail_open` (default, `plugin_config.go:164`) → returns `nil` → request forwarded with the partially mutated body.

`SkipRecoveryUnavailable` (`contracts.go:140`) is defined but never assigned. The skip-not-abort intent for a nil/disabled recovery store was never wired.

Present since PR #2813 (merged 2026-08-10), the first version of the compression service.

### To Reproduce

1. Configure `context_compression` with `failure_mode: fail_open` (default) and at least one `tool_outputs` target in `recoverable` mode.
2. Send a request with two compressible tool-output messages. Set `recovery.max_bytes_per_request` to the first block's size so the first stores successfully and the second exceeds the limit.
3. The first tool-output block is replaced by the compressed summary and recovery-key placeholder in the forwarded body, but the compression result reports `Failure != nil`, `Applied = false`.

A unit test reproducing this (against current main, before any fix) fails with the first block left as the compressed placeholder instead of its original text.

### Expected behavior

On a recovery-store failure mid-batch, already-applied blocks are restored to their original text so the forwarded body matches the compression result. The receipt (`Applied`, `Failure`, `SkipReason`) describes the body actually forwarded.

### Impact

Deployments using `recoverable` compression with the default `fail_open`. The forwarded request carries compressed content the receipt says was not applied, and the original tool-output text is lost (the recovery key is only injected on the success path, so the first block's key is never added). `fail_closed` rejects the request but still leaves the body partially mutated before the error surfaces.

### Scope

- Restore already-applied candidates in `executePlan` when a later candidate's recovery-store operation returns an error, reusing the existing `finalizeResult` rollback pattern.
- Open: wire `SkipRecoveryUnavailable` so a nil/disabled recovery store skips the recoverable target instead of aborting the batch.
- Open: whether store `Put` failure and byte-limit-exceeded should skip-and-continue vs abort-and-rollback. Proposal: abort + rollback — a failing store should not silently drop recovery content.

### Completion signal

- [ ] A recovery-store failure mid-batch restores all already-mutated blocks so the forwarded body matches the compression receipt.
- [ ] `fail_open` returns a body identical to the original request on recovery-store failure.
- [ ] `fail_closed` rejects the request with the original body intact.
- [ ] Unit test covers the mixed extractive/recoverable + store-failure + `fail_open` path.

Bevisy:
Closes #3278

## Purpose

`executePlan` returned the error from a failing recoverable candidate without restoring blocks already mutated by earlier candidates in the same batch. Under the default `failure_mode: fail_open`, the request was forwarded with a partially compressed body while the receipt reported `Applied = false` / `Failure != nil` — the body and receipt disagreed, and the original text of already-applied blocks was lost (recovery keys are only injected on the success path).

Restore all already-applied candidates in the `applyErr` branch of `executePlan`, reusing the existing rollback pattern from `finalizeResult`.

Two related design questions (wiring `SkipRecoveryUnavailable`, skip-vs-abort on store `Put` failure) are left as open items on #3278 for the Workgroup; this PR only fixes the rollback gap.

Module: `pkg/contextcompression`.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/contextcompression/... -count=1
```

`TestServiceRollsBackAppliedBlocksOnRecoveryFailure`: two recoverable tool-output blocks; the first stores successfully and is applied, the second exceeds `recovery.max_bytes_per_request` and errors. Verifies the first block is restored to its original text and the committed store entry is retained.

## Test Result

```
ok   github.com/vllm-project/semantic-router/src/semantic-router/pkg/contextcompression   2.880s
```

Verified the test fails on current main (block left as compressed placeholder) and passes with this fix. `fail_open` now forwards a body identical to the original request on recovery-store failure.

---

<details>

- [x] The PR links an `accepted` issue with exactly one owner (#3278, `wg/agentic-context`)

</details>

### DIFF
--- a/src/semantic-router/pkg/contextcompression/service.go
+++ b/src/semantic-router/pkg/contextcompression/service.go
@@ -118,6 +118,9 @@ func (s *Service) executePlan(
 			&recoveryBytes,
 		)
 		if applyErr != nil {
+			for _, applied := range appliedCandidates {
+				applied.block.SetText(applied.originalText)
+			}
 			return result, appliedCandidates, applyErr
 		}
 		if !applied {
--- a/src/semantic-router/pkg/contextcompression/service_test.go
+++ b/src/semantic-router/pkg/contextcompression/service_test.go
@@ -419,3 +419,55 @@ func TestServiceReportsRecoveryBackendFailure(t *testing.T) {
 		t.Fatalf("backend failure result = %#v", result)
 	}
 }
+
+func TestServiceRollsBackAppliedBlocksOnRecoveryFailure(t *testing.T) {
+	firstContent := strings.Repeat("first recoverable output ", 200)
+	secondContent := strings.Repeat("second output ", 200)
+	body := map[string]interface{}{
+		"messages": []interface{}{
+			map[string]interface{}{"role": "user", "content": "question"},
+			map[string]interface{}{
+				"role":         "tool",
+				"tool_call_id": "call_1",
+				"content":      firstContent,
+			},
+			map[string]interface{}{
+				"role":         "tool",
+				"tool_call_id": "call_2",
+				"content":      secondContent,
+			},
+		},
+	}
+	ir := ParseRequestIR(body, Provenance{})
+	store := &testRecoveryStore{}
+	result := NewService().Apply(context.Background(), Request{
+		Scope:    "scope",
+		Request:  ir,
+		Recovery: store,
+		Policy: Policy{
+			Mode: ModeAlways,
+			Targets: Targets{
+				ToolOutputs: TargetPolicy{
+					Mode:         TargetRecoverable,
+					MinTokens:    10,
+					TargetTokens: 20,
+				},
+			},
+			Recovery: RecoveryPolicy{
+				Enabled:            true,
+				MaxBytesPerRequest: len(firstContent),
+			},
+		},
+		TokenCounter: HeuristicTokenCounter{},
+	})
+	if result.Failure == nil {
+		t.Fatalf("expected failure from recovery byte-limit, got %#v", result)
+	}
+	firstBlock := ir.Messages[1].Blocks[0]
+	if firstBlock.Text != firstContent {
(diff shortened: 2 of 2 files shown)


FAUST-BENCHOU (approved):
LGTM. The rollback fix keeps the request body and recovery result consistent when a later candidate fails, which closes an important partial-application edge case.

wilsonwu (approved):
LGTM, thanks!

PR #3293, opened 2026-09-01
[Security] Replace hardcoded Grafana admin credential with per-stack secret

### LINKED ISSUE #3276: [Security] Grafana admin credentials are hardcoded to admin/admin with no way to override them
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

The local support stack always starts Grafana with `GF_SECURITY_ADMIN_USER=admin` and `GF_SECURITY_ADMIN_PASSWORD=admin`. The credentials are hardcoded in `src/vllm-sr/cli/container_support_services.py` and cannot be overridden.

### To Reproduce

1. Start the maintained support stack.
2. Open the Grafana endpoint.
3. Sign in with `admin` / `admin`.
4. Observe that the credential is identical for every generated stack and has no supported override.

### Expected behavior

Grafana should receive a generated per-stack credential through a secret file, or at minimum require an explicit environment-provided password. A fixed shared administrator credential must not be the maintained default.

### Impact

Any process that can reach the loopback Grafana port can use the administrator account to change dashboards and data sources or inspect connected telemetry. Loopback reduces exposure but is not an authorization boundary on a shared host or after a user adds a tunnel or reverse proxy.

### Additional context

Current main still passes the fixed values directly to the container. #2485 already moved Postgres and Redis toward generated per-stack credentials; Grafana is the remaining support service with a source-controlled administrator password. Prefer the same generated-secret pattern and avoid printing the secret itself.

littlewuuu:
Closes #3276

## Purpose

Fixes a security weakness in the local support stack: Grafana was always started with a source-controlled `GF_SECURITY_ADMIN_PASSWORD=admin` and no way to override it. The mounted production `grafana.serve.ini` also granted anonymous users the Admin role, so a generated password would not have blocked administrator access.

Mirrors the generated-per-stack secret pattern already used for Postgres/Redis (#2485):

- **Explicit override wins**: an operator-set host `GF_SECURITY_ADMIN_PASSWORD` is used verbatim and materialized into the per-stack secret file, so the container always mounts a real source.
- **Otherwise generated per stack**: a 256-bit CSPRNG value minted once, persisted per stack, and reused on later `serve` calls so a restart keeps the same credential.
- **Never exposed**: the value is handed to Grafana only through a bind-mounted secret file (`GF_SECURITY_ADMIN_PASSWORD__FILE`), so it never reaches argv, `docker inspect .Config.Env`, a config file, or a log.
- **Well-defined lifecycle**: restarting keeps the same credential; the env overwrites that one file, and unsetting it reuses whatever is already persisted.
- **Anonymous is Viewer**: `grafana.serve.ini` keeps anonymous access for the dashboard iframe but not Admin; login form and basic auth stay enabled so the password-file admin can authenticate.
- **Loopback publish**: Grafana's host port binds to `127.0.0.1`.
- The serve summary prints the password file path, never the secret, and no longer advertises `(admin/admin)`.

Affected modules (Python CLI): 
- `src/vllm-sr/cli/grafana_credentials.py` (new) — per-stack Grafana admin password file (env overwrite or generated reuse)
- `src/vllm-sr/cli/commands/runtime_paths.py` — container-readable state-file reader that does not revoke 0644 on read-back
- `src/vllm-sr/cli/container_support_services.py` — `container_start_grafana` mounts the materialized secret file instead of injecting the credential directly, and publishes Grafana on loopback
- `src/vllm-sr/cli/templates/grafana.serve.ini` — anonymous Viewer, authenticated admin path
- `src/vllm-sr/cli/runtime_lifecycle.py` — print the password file path instead of `(admin/admin)`
- `src/vllm-sr/tests/test_grafana_credentials.py` (new) — regression coverage
- `e2e/testing/vllm-sr-cli/test_integration_grafana_credentials.py` (new) — real Grafana with the production ini; unauthenticated org-admin requests are denied

## Test Plan

```bash
cd src/vllm-sr
python3 -m pytest tests/test_grafana_credentials.py \
  tests/test_runtime_lifecycle.py \
  tests/test_container_services.py \
  tests/test_storage_secrets.py \
  tests/test_storage_backends.py \
  tests/test_runtime_support.py \
  tests/test_runtime_stack_naming.py \
  tests/test_grafana_dashboard_template.py -q
```

## Test Result

All 114 related tests pass. `test_grafana_credentials.py` adds nine cases covering generation/reuse, explicit-env materialization into the same per-stack file, per-stack isolation, container regressions proving both auto and explicit paths mount a real secret file while the password never appears in argv, production ini restricting anonymous users to Viewer with loopback publish, and the serve summary printing the password file path rather than the secret.

### DIFF
--- a/src/vllm-sr/cli/container_support_services.py
+++ b/src/vllm-sr/cli/container_support_services.py
@@ -20,6 +20,11 @@ from cli.container_observability import (
 )
 from cli.container_runtime import get_container_runtime
 from cli.container_services import _replace_existing_container
+from cli.grafana_credentials import (
+    CONTAINER_GRAFANA_PASSWORD_PATH,
+    GRAFANA_ADMIN_PASSWORD_FILE_ENV,
+    ensure_grafana_admin_password_file,
+)
 from cli.runtime_stack import RuntimeStackLayout, resolve_runtime_stack
 from cli.utils import get_logger
 
@@ -143,6 +148,12 @@ def container_start_grafana(
             stack_layout,
         )
 
+    # Resolves and materializes the admin password into a file the value is
+    # bind-mounted from, so the mount source always exists (never in argv/env).
+    password_file = ensure_grafana_admin_password_file(
+        config_dir, stack_layout=stack_layout
+    )
+
     cmd = [
         runtime,
         "run",
@@ -152,9 +163,12 @@ def container_start_grafana(
         "--network",
         network_name,
         "-e",
-        "GF_SECURITY_ADMIN_USER=admin",
-        "-e",
-        "GF_SECURITY_ADMIN_PASSWORD=admin",
+        f"{GRAFANA_ADMIN_PASSWORD_FILE_ENV}={CONTAINER_GRAFANA_PASSWORD_PATH}",
+        "-v",
+        (
+            f"{os.path.abspath(password_file)}:"
+            f"{CONTAINER_GRAFANA_PASSWORD_PATH}:ro,z"
+        ),
         "-e",
         f"PROMETHEUS_URL={stack_layout.prometheus_container_name}:9090",
         "-v",
@@ -168,7 +182,7 @@ def container_start_grafana(
         "-v",
         f"{os.path.abspath(os.path.join(grafana_dir, 'llm-router-dashboard.serve.json'))}:/etc/grafana/provisioning/dashboards/llm-router-dashboard.json:ro",
         "-p",
-        f"{stack_layout.grafana_port}:3000",
+        f"127.0.0.1:{stack_layout.grafana_port}:3000",
         "docker.io/grafana/grafana:11.5.1",
     ]
     return _run_service_start(cmd, "Grafana")
--- a/e2e/testing/vllm-sr-cli/test_integration_grafana_credentials.py
+++ b/e2e/testing/vllm-sr-cli/test_integration_grafana_credentials.py
@@ -1,0 +1,231 @@
+#!/usr/bin/env python3
+"""Real-daemon regression: a real Grafana container must boot and become ready
+when started exactly like ``container_start_grafana`` with the CLI's mounted
+password file and the rendered production ``grafana.serve.ini``.
+"""
+
+import base64
+import os
+import stat
+import time
+import unittest
+from pathlib import Path
+from urllib import error as urllib_error
+from urllib import request as urllib_request
+
+from cli import grafana_credentials as gc
+from cli.container_observability import render_observability_template
+from cli_test_base import (
+    HTTP_STATUS_OK,
+    CLITestBase,
+    stack_scoped_test_container_name,
+)
+
+# Keep in sync with `container_start_grafana` in
+# src/vllm-sr/cli/container_support_services.py.
+GRAFANA_IMAGE = "docker.io/grafana/grafana:11.5.1"
+GRAFANA_SERVE_INI_TEMPLATE = (
+    Path(gc.__file__).resolve().parent / "templates" / "grafana.serve.ini"
+)
+CONTAINER_GRAFANA_INI_PATH = "/etc/grafana/grafana.ini"
+# Org-admin API: anonymous Viewer is 403, anonymous disabled is 401, Admin is 200.
+GRAFANA_ORG_ADMIN_API = "/api/org/users"
+HTTP_STATUS_UNAUTHORIZED = 401
+HTTP_STATUS_FORBIDDEN = 403
+
+GRAFANA_CONTAINER_SUFFIX = "vllm-sr-cli-test-grafana"
+GRAFANA_READY_TIMEOUT = 180
+GRAFANA_START_COMMAND_TIMEOUT = 600
+
+integration_only = unittest.skipUnless(
+    os.environ.get("RUN_INTEGRATION_TESTS", "").lower() == "true",
+    "Integration tests disabled. Set RUN_INTEGRATION_TESTS=true to enable.",
+)
+
+
+class TestGrafanaPasswordFileContainer(CLITestBase):
+    """A real Grafana container must boot with the CLI's mounted password file."""
+
+    @classmethod
+    def setUpClass(cls):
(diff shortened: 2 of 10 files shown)


FAUST-BENCHOU (dismissed):
LGTM. Generating a per-stack Grafana credential and keeping it out of container environment/log output is a good improvement over the previous fixed admin password approach.

wilsonwu (dismissed):
LGTM, thanks!

wilsonwu (dismissed):

Xunzhuo (changes requested):
Thanks for removing the fixed credential and keeping the generated secret out of argv/env inspection. One blocker remains in the explicit override path: `resolve_grafana_admin_password()` returns `GF_SECURITY_ADMIN_PASSWORD` without creating the password file, while `container_start_grafana()` discards that return value and still bind-mounts `grafana_password_path(...)`. With the documented host override set, the mount source therefore does not exist and Grafana cannot read the supplied password. Please materialize the explicit value through a secure file-backed path (without leaking it into argv/logs), define its restart/removal behavior, and add a container-command regression test for the explicit-env case. The current focused suite passes 6 tests but does not exercise that path end to end.

Xunzhuo (dismissed):
The explicit override is now materialized through a private file-backed path, mounted without exposing the value in argv, with restart and unset behavior covered. The prior blocker is resolved. LGTM.

wilsonwu (dismissed):
LGTM

littlewuuu:
Rebased to latest main (adapting the new test to upstream's removal of log_runtime_summary's fleet_sim_enabled). Only change since the last approve, please re-approve, thanks!

Xunzhuo (commented):
Thanks for keeping the password out of argv and the container environment. One startup blocker remains: `ensure_grafana_admin_password_file` writes both mounted files as host-owner-only `0600`, while the Grafana image runs as an unprivileged user, so it cannot read the bind-mounted `GF_SECURITY_ADMIN_PASSWORD__FILE`. Please use the existing container-readable state-file mode (the parent directory remains private) or otherwise make the mounted file readable by Grafana, and add a real container startup/readiness regression.

littlewuuu:
> Thanks for keeping the password out of argv and the container environment. One startup blocker remains: `ensure_grafana_admin_password_file` writes both mounted files as host-owner-only `0600`, while the Grafana image runs as an unprivileged user, so it cannot read the bind-mounted `GF_SECURITY_ADMIN_PASSWORD__FILE`. Please use the existing container-readable state-file mode (the parent directory remains private) or otherwise make the mounted file readable by Grafana, and add a real container startup/readiness regression.

@Xunzhuo Got it on the permission issue, and I just find out I didn't create a separate branch for this and just pushed from my fork's main, so the PR is tied to main. do i need to close this and open a fresh PR from a dedicated branch (e.g. fix/3276-grafana-secret), or keep this one and keep updating main?

Xunzhuo (commented):
No need to close or reopen it—you can keep this PR and continue updating the same branch, even if that branch is the main branch of your fork. Please fix the Grafana-readable file-mode blocker here, then sync current upstream `main`, let the exact-head checks finish, and re-request review.

Xunzhuo (commented):
Thanks—the Grafana password is now container-readable only inside the owner-only state directory, while remaining absent from argv and environment. The exact-head real-daemon regression verifies wrong-password 401 and generated-password 200, so the prior secret-permission blocker is resolved. Please rebase onto current `main` for the final high-risk approval pass.

WUKUNTAI-0211 (changes requested):
The password-file plumbing is correct, but the shipped Grafana configuration still enables anonymous users as Admin while the container publishes its port. The generated/explicit password therefore does not protect administrator access.

The new real-daemon test does not mount the production grafana.serve.ini, so its 401/200 assertions exercise image defaults rather than the vllm-sr serve configuration.

Please disable anonymous auth, or restrict it to Viewer while retaining authenticated admin access through the password file. Update the real-daemon test to mount the rendered production ini and prove unauthenticated admin access is denied.

src/vllm-sr/cli/container_support_services.py
```diff
@@ -152,9 +166,16 @@ def container_start_grafana(
         "--network",
         network_name,
         "-e",
-        "GF_SECURITY_ADMIN_USER=admin",
-        "-e",
-        "GF_SECURITY_ADMIN_PASSWORD=admin",
+        f"{GRAFANA_ADMIN_PASSWORD_FILE_ENV}={CONTAINER_GRAFANA_PASSWORD_PATH}",
```
Xunzhuo, line 166: **Anonymous Admin bypasses the generated Grafana password**

Impact: Anyone who can reach the published Grafana port can still obtain administrator access without credentials, so replacing admin/admin with a generated secret does not remove the vulnerability this PR claims to fix.

Required change: Disable anonymous access or remove its Admin role and enable an authenticated login path in the mounted production grafana.serve.ini; make the integration test mount that exact configuration and assert unauthenticated administrative requests are denied.

Xunzhuo (commented):

Xunzhuo (commented):
Thanks for tightening the Grafana credential path. The production flow is covered now, but the optional state-directory contract still regresses.

src/vllm-sr/cli/container_support_services.py
```diff
@@ -143,6 +148,12 @@ def container_start_grafana(
             stack_layout,
         )
 
+    # Resolves and materializes the admin password into a file the value is
+    # bind-mounted from, so the mount source always exists (never in argv/env).
+    password_file = ensure_grafana_admin_password_file(
```
Xunzhuo, line 153: **Preserve the default Grafana state path**

Impact: `container_start_grafana` still advertises `config_dir` as optional, and its existing template path normalizes an omitted value to `$PWD/.vllm-sr`. This new credential call passes the original `None` into `Path`, so a direct/default restart removes the existing Grafana container and then raises `TypeError` before starting a replacement.

Required change: Normalize the effective state root (`config_dir` or the current working directory) before both template and credential setup, and add an omitted-`config_dir` regression.

PR #3294, opened 2026-09-01
[Bug] Detect existing Podman runtime in install.sh

### LINKED ISSUE #3286: [Bug] install.sh ignores an existing Podman runtime and always installs Docker
### Proposed Workgroup

### Describe the bug

`install.sh` only detects Docker as an existing container runtime and never checks for Podman. When Docker is not present, the installer proceeds to install Docker (via `apt-get install docker.io`, Homebrew `docker`+`colima`, etc.) even when Podman is already installed and usable.

This contradicts `src/vllm-sr/cli/container_runtime.py`, which treats Podman as a first-class runtime and auto-detects it when Docker is absent (#2954 previously fixed Podman support in `vllm-sr serve`).

Three functions are affected:
- `detect_existing_runtime()` only checks `docker_ready`
- `docker_ready()` only checks the `docker` command
- `ensure_runtime()` falls straight into Docker installation without a Podman branch

### To Reproduce

1. Prepare a host with Podman installed but no Docker (common on RHEL/Fedora/Rocky, rootless Linux servers, or macOS with `podman machine`).
2. Confirm Podman works: `podman info` succeeds.
3. Run `curl -fsSL https://vllm-sr.ai/install.sh | bash`.
4. Observe the installer does not recognize Podman and attempts to install Docker instead.

### Expected behavior

`install.sh` should detect an existing Podman runtime the same way the CLI (`container_runtime.py`) does, and skip Docker installation when Podman is available. The detected runtime should be persisted so `vllm-sr serve` picks it up.

Suggested approach:
- `docker_ready()` / `detect_existing_runtime()` should also check `has_cmd podman && podman info >/dev/null 2>&1`.
- `ensure_runtime()` should accept an existing Podman runtime and write `CONTAINER_RUNTIME=podman` into `$INSTALL_ROOT/runtime.env` instead of installing Docker.
- `describe_runtime_dependency_plan()` and `choose_runtime_preference()` should reflect Podman as a valid auto-detected option.

### Impact

- Rootless Linux servers where only Podman is available: the Docker install step fails (no root) or is undesirable.
- RHEL/Fedora/Rocky default environments ship Podman, not Docker — these users are all affected.
- macOS users running `podman machine` are pushed toward installing Colima/Docker unnecessarily.

Workaround: manually set `VLLM_SR_RUNTIME=skip` (or `--runtime skip`) to bypass runtime bootstrap, then ensure `CONTAINER_RUNTIME=podman` is set before running `vllm-sr serve`. This is not obvious from the installer output.

### Additional context

- Podman support in `vllm-sr serve` was added in v0.2.0 and further fixed in #2954.
- `write_runtime_env()` currently just removes `runtime.env` instead of persisting the detected runtime — it could write the detected runtime so the CLI picks it up automatically.

Zi-Yi-Ming:
## Purpose

install.sh only detected Docker as an existing container runtime and never checked for Podman. When Docker was absent, the installer proceeded to install Docker even when Podman was already installed and usable. This contradicts `container_runtime.py`, which treats Podman as a first-class runtime.

This PR adds Podman detection so the installer recognizes an existing Podman runtime, skips Docker installation, and persists the detected runtime to `runtime.env` for `vllm-sr serve` to pick up.

Owned by `wg/developer-experience-ecosystem` (accepted issue #3286).

## Test Plan

- `bash -n install.sh` — syntax check
- Manual: on a host with Podman only (no Docker), verify `detect_existing_runtime` returns "podman" and `ensure_runtime` skips Docker installation.
- Manual: on a host with Docker, verify Docker path unchanged.
- Manual: on a host with neither, verify fallback to Docker install unchanged.

## Test Result

`bash -n install.sh` passes. Docker-only and neither-present paths are unchanged by inspection (pure additions, no modification to existing Docker branches). Podman-only path verified by code review against the symmetric Docker logic.

---

Closes #3286

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -775,7 +775,7 @@ install_linux_docker_runtime() {
 }
 
 write_runtime_env() {
-  local runtime="${1:-}"
+  local runtime="${1:-${SELECTED_RUNTIME:-}}"
   rm -f "$INSTALL_ROOT/runtime.env"
   if [ -n "$runtime" ]; then
     printf 'CONTAINER_RUNTIME=%s\n' "$runtime" > "$INSTALL_ROOT/runtime.env"
--- a/install.sh
+++ b/install.sh
@@ -498,6 +498,11 @@ detect_existing_runtime() {
     return
   fi
 
+  if podman_ready; then
+    printf 'podman\n'
+    return
+  fi
+
   return 1
 }
 
@@ -710,6 +715,10 @@ docker_ready() {
   has_cmd docker && docker info >/dev/null 2>&1
 }
 
+podman_ready() {
+  has_cmd podman && podman info >/dev/null 2>&1
+}
+
 choose_runtime_preference() {
   if [ "$REQUESTED_RUNTIME" != "auto" ]; then
     printf '%s\n' "$REQUESTED_RUNTIME"
@@ -766,7 +775,11 @@ install_linux_docker_runtime() {
 }
 
 write_runtime_env() {
+  local runtime="${1:-}"
   rm -f "$INSTALL_ROOT/runtime.env"
+  if [ -n "$runtime" ]; then
+    printf 'CONTAINER_RUNTIME=%s\n' "$runtime" > "$INSTALL_ROOT/runtime.env"
+  fi
 }
 
 ensure_runtime() {
@@ -781,6 +794,13 @@ ensure_runtime() {
     SELECTED_RUNTIME="docker"
     write_runtime_env
     done_step "Using existing Docker runtime"
+    return
+  fi
+
+  if podman_ready; then
+    SELECTED_RUNTIME="podman"
+    write_runtime_env "podman"
+    done_step "Using existing Podman runtime"
     return
   fi
 
--- a/install.sh
+++ b/install.sh
@@ -797,7 +797,7 @@ ensure_runtime() {
     return
   fi
 
-  if podman_ready; then
+  if [ "$REQUESTED_RUNTIME" = "auto" ] && podman_ready; then
     SELECTED_RUNTIME="podman"
     write_runtime_env "podman"
     done_step "Using existing Podman runtime"
--- a/src/vllm-sr/tests/install_runtime_harness.sh
+++ b/src/vllm-sr/tests/install_runtime_harness.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# Harness: exercise install.sh runtime selection with stubbed Docker/Podman
+# so the behavior tests can inspect the real code path and the persisted
+# runtime.env. Invoked by test_install_runtime_behavior.py; not meant to be
+# run by hand.
+#
+# Usage: install_runtime_harness.sh <scenario> <repo-root>
+#
+# Scenarios:
+#   auto-both-ready              Docker + Podman present, --runtime auto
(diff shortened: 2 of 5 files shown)


Xunzhuo (changes requested):
Thanks for taking this on, and welcome! I found one blocker: this head changes the installer contract without updating its tests or user-facing runtime contract. `test_install_script_only_mentions_docker_runtime` now fails because it still asserts that `install.sh` must not mention Podman, and there is no coverage for Docker-preferred auto detection, Podman-only detection, explicit Docker/skip, or the persisted `runtime.env`. Please update the focused installer tests and documentation for the new behavior, then rerun that suite.

Zi-Yi-Ming:
@Xunzhuo  Thanks for the clear writeup. Follow-up pushed in
6be65ebd:

- test_install_script_surface.py: replaced the Docker-only surface test
  with test_install_script_runtime_contract_supports_podman_fallback,
  asserting Docker-first / Podman-fallback gated on --runtime auto and
  unchanged user-facing --runtime choices; added
  test_install_script_persists_selected_runtime covering the
  runtime.env write; renamed the doc test accordingly.
- website/docs/installation/installation.md: Requirements now reads
  "Docker, or Podman as a fallback on Linux" and the installer
  paragraph documents the --runtime auto preference order and the
  persisted runtime.env.

Xunzhuo (changes requested):
Thanks for the quick follow-up—the documentation now describes the intended fallback, and the focused surface suite passes 8 tests. The test blocker is only partially addressed, though: the new tests assert that strings such as `podman_ready` and `CONTAINER_RUNTIME=` appear in the script; they never execute runtime selection or inspect the generated file. They would still pass if Docker/Podman precedence, explicit `docker`/`skip`, or the written value regressed. Please add a small shell/function harness with stubbed Docker/Podman commands that exercises: both ready (Docker wins), Podman only under auto, explicit Docker, skip, and the exact `runtime.env` contents. Then rerun that focused suite.

Zi-Yi-Ming:
@Xunzhuo  Follow-up pushed:

- Added src/vllm-sr/tests/install_runtime_harness.sh: sources install.sh
  with `main` stripped and stubs docker/podman through a temp PATH, then
  invokes ensure_runtime so the real runtime-selection code path runs.
- Added src/vllm-sr/tests/test_install_runtime_behavior.py with 4
  scenarios asserting both behavior and the exact runtime.env contents:
  - auto + both ready       -> docker wins, CONTAINER_RUNTIME=docker
  - auto + podman only      -> podman fallback, CONTAINER_RUNTIME=podman
  - explicit --runtime docker + both ready -> docker, podman branch not
    exercised
  - --runtime skip          -> SELECTED_RUNTIME empty, no runtime.env
    written

The surface suite still covers the script/doc contract; the new suite
covers executable behavior. Verified all 4 scenarios locally.

Xunzhuo (approved):
Thanks for the follow-up — this resolves the test blocker. The new harness exercises the real `ensure_runtime` path across Docker-first selection, Podman fallback, explicit Docker, and skip scenarios; all 12 focused installer tests pass. CI is still running, but I found no remaining code blocker on this head.

FAUST-BENCHOU (approved):

wilsonwu (approved):
LGTM

PR #3296, opened 2026-09-01
[CLI] Accept digest-pinned images in the local-presence check

### LINKED ISSUE #3277: [Bug] --image-pull-policy never always fails for digest-pinned images
### Proposed Workgroup

### Describe the bug

`--image-pull-policy never` rejects a digest-pinned image even when that exact image is present locally. `container_image_exists` calls `<runtime> images -q <reference>`, which does not reliably resolve `repo@sha256:...` references, and `_ensure_image_available` then exits as if the image were missing.

### To Reproduce

1. Pull a vLLM-SR image locally.
2. Resolve and use its `repo@sha256:<digest>` reference.
3. Confirm `<runtime> image inspect <digest-reference>` succeeds while `<runtime> images -q <digest-reference>` returns no ID.
4. Run `vllm-sr serve` with the digest reference and `--image-pull-policy never`.

### Expected behavior

The local-presence check should accept both tag and digest references, allowing a reproducible digest-pinned image to run without network access.

### Impact

Air-gapped and reproducible deployments cannot combine the two intended guarantees: immutable image identity and no pull during startup.

### Additional context

Current main still uses `images -q`. A bounded fix is to use `<runtime> image inspect <reference>` and check its return code for both Docker and Podman, with tests for tag, digest, missing image, and runtime failure paths.

iroiro147:
## Summary

`container_image_exists` used `<runtime> images -q <reference>`, which does not reliably resolve `repo@sha256:...` references. A digest-pinned image that is present locally yields no ID, so `_ensure_image_available` treated it as missing and `--image-pull-policy never` refused to start it.

That made the two guarantees air-gapped deployments need — immutable image identity and no pull at startup — mutually exclusive, which is the impact described in #3277.

The check now uses `image inspect`, which resolves tag and digest references alike, and reports presence through the **exit status** rather than stdout. That second part matters: `images -q` communicates presence by printing an ID, `image inspect` by exiting 0, so keying off `stdout` would have kept the bug for any runtime that prints diagnostics.

This is the bounded fix the issue asks for — one call site, no change to the pull-policy logic around it.

## Tests

Two regression tests in `tests/test_container_runtime.py`, both of which **fail against the old implementation** (verified by reverting the change and re-running):

- `test_container_image_exists_accepts_digest_pinned_images` — a `repo@sha256:` reference reads as present, and the command issued is `image inspect`. Its fake returns the manifest on stdout with exit 0, mirroring the real runtime.
- `test_container_image_exists_reports_missing_on_nonzero_exit` — a genuinely absent image still reads as absent even when the runtime writes `[]` to stdout, so the fix cannot turn every miss into a false positive.

## Verification

- `pytest tests/test_container_runtime.py` — 15 passed
- `black --check` and `ruff check --config tools/linter/python/.ruff.toml` — clean on both files

One note for reviewers: I did not exercise this against a real container runtime with a genuinely digest-pinned local image, so the behavioural claim rests on the documented difference between `images -q` and `image inspect` plus the reproduction in the issue, not on my own end-to-end run.

Fixes #3277


Xunzhuo (approved):
Thanks for the focused fix. `image inspect` correctly uses the runtime exit status for both tag and digest references, and the tests cover the digest-present and diagnostic-output/missing cases. I also ran the focused container-runtime suite at this exact head: 15 passed. Approved; the branch only needs its routine rebase onto current main.

wilsonwu (approved):
LGTM, thanks!

FAUST-BENCHOU (approved):

Issue #3256, opened 2026-09-01
[Bug] model_fa_fp16.onnx holds FP32 weights in five of six published 32K heads


subin9:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

Five of the six published 32K heads ship `onnx/model_fa_fp16.onnx` with FP32 weights. Only `mmbert32k-modality-router-merged` is actually FP16.

```
repo                                   onnx/model.onnx  onnx/model_fa_fp16.onnx  onnx/model_sdpa_fp16.onnx
mmbert32k-jailbreak-detector-merged            1173.7                   1173.6                      586.7
mmbert32k-intent-classifier-merged             1173.7                   1173.7                      586.7
mmbert32k-pii-detector-merged                  1173.8                   1173.7                      586.8
mmbert32k-factcheck-classifier-merged          1173.7                   1173.6                          -
mmbert32k-feedback-detector-merged             1173.7                   1173.7                          -
mmbert32k-modality-router-merged               1173.7                    587.1                      587.1
```

Sizes in MB, from the Hub API. Checked the weights directly for jailbreak, intent and factcheck: all three hold 138 `FLOAT` initializers totalling 307.5M parameters. Modality holds 138 `FLOAT16`.

The custom op is not the reason. All three files carry the same 22 `com.ck::CKFlashAttention` nodes in the `com.ck` domain, and the modality export is FP16 with those same nodes, so CK Flash Attention does not require FP32 weights.

`mmbert32k-factcheck-classifier-merged` and `mmbert32k-feedback-detector-merged` also publish no `model_sdpa_fp16.onnx` at all.

### To Reproduce

```
python - <<'EOF'
from collections import Counter
from huggingface_hub import hf_hub_download
import onnx
from onnx import TensorProto
name = {v: k for k, v in TensorProto.DataType.items()}
for repo in ["llm-semantic-router/mmbert32k-intent-classifier-merged",
             "llm-semantic-router/mmbert32k-modality-router-merged"]:
    m = onnx.load(hf_hub_download(repo, "onnx/model_fa_fp16.onnx"), load_external_data=False)
    print(repo, Counter(name[i.data_type] for i in m.graph.initializer))
EOF
```

```
.../mmbert32k-intent-classifier-merged Counter({'FLOAT': 138})
.../mmbert32k-modality-router-merged   Counter({'FLOAT16': 138})
```

### Expected behavior

[...]

PR #3303, opened 2026-09-02
[Bug] Rank the FP32 graph first on the CPU provider

### LINKED ISSUE #3302: [Bug] find_onnx_models ignores the execution provider, so CPU signals run 10x slower on the FP16 graph
### Proposed Workgroup

### Describe the bug

A signal pinned to CPU loads `model_sdpa_fp16.onnx` and runs 10 to 12 times slower than the same head on `model.onnx`. The CPU provider has no FP16 kernels for these operators, so it casts around every one of them.

`mmbert32k-jailbreak-detector-merged`, p50 ms:

```
file                          size MB     len32    len128    len512
model.onnx                     1173.7      18.9      52.1     212.4
model_sdpa_fp16.onnx            586.7     200.8     617.6    2545.5
```

It picks that file because `find_onnx_models` ranks candidates by filename and never sees the provider. It takes only a path (`mmbert_classifier.rs:236`); the provider goes separately to `create_session_with_fallback` (`:322`). The one thing that reorders the list is `ORT_CK_FLASH_ATTN_LIB`, which is AMD CK Flash Attention. Unset, the FP16 graph is first:

```rust
// mmbert_classifier.rs:272
} else {
    &[
        "model_sdpa_fp16.onnx",
        "model.onnx",
        "classifier.onnx",
        "model_optimized.onnx",
    ]
};
```

`mmbert_embedding.rs:407` has the same list. Only the CPU provider is affected; FP16 first is right on CUDA and ROCm and this does not change them.

### To Reproduce

Run a signal with `use_cpu: true` against a model directory holding both files. `DownloadModelWithProgress` (`downloader.go:79`) runs `hf download <repo> --local-dir <path>` with no include or exclude patterns, so `onnx/` arrives whole on a normal install. Startup prints the mismatch:

```
INFO: Selected classifier ONNX file: .../onnx/model_sdpa_fp16.onnx
INFO: Using CPU execution provider
```

### Environment

ONNX Runtime 1.29, CPU provider, batch 1, 8 intra-op threads. Intel i7-14700K, x86_64, Linux 6.18 under WSL2. The ratio is the claim, not the absolute latency.

### Expected behavior

`find_onnx_models` takes the provider. Under `ClassifierExecutionProvider::Cpu`, `model.onnx` ranks first.

### Impact

`config/config.yaml` sets `use_cpu: true` in ten places, including `prompt_guard`, `classifier.domain`, `classifier.pii`, `hallucination_mitigation.fact_check` and `routing.signals.classifiers[0]`. Every one of them pays the 10x on every request.

`Dockerfile.cuda:281` and `Dockerfile.rocm:281` build with `-tags=onnx` and produce `router` as the only binary, so this is not an opt-in path in those images.

Timed on one head; the others publish the same file pair. Table from my comment on #2408, which changes AMD provider order and does not touch candidate ranking. #3256 is a separate defect in the same list.

subin9:
Closes #3302

`find_onnx_models` took only a path, so the ranking never saw which execution provider the session would use. With `ORT_CK_FLASH_ATTN_LIB` unset it put `model_sdpa_fp16.onnx` first, including for a signal pinned to CPU, where the provider has no FP16 kernels and casts around every operator.

The provider now reaches the ranking. Under `Cpu` the order starts at `model.onnx` and the FP16 graph moves to last rather than being dropped, so a model that publishes only that file still loads. `Auto` counts as CPU when the binary has no `cuda` or `rocm` feature, since it cannot resolve to anything else there.

CK Flash Attention and GPU orderings are untouched. FP16 first is the right choice on CUDA and ROCm.

`mmbert_embedding.rs` had the same list and takes `use_cpu` already, so it gets the same treatment.

## Test Plan

- `cargo test --lib` in `onnx-binding`
- three new cases: CPU ranks `model.onnx` first, CPU still loads an FP16-only model, `Cuda` and `Rocm` keep `model_sdpa_fp16.onnx` first

## Test Result

- 50 passed, 0 failed.

Not covered here: the `Auto` case on a GPU build where EP registration fails at runtime and the session falls back to CPU. Ranking happens before that is known, so it would need the fallback to re-rank. #3256 is a separate defect in the same candidate list.


FAUST-BENCHOU (approved):

wilsonwu (approved):
LGTM

Issue #2999, opened 2026-08-24
[Feature] Implement Conversations API state and history semantics


VadimEisenberg:
### Goal

Implement bounded OpenAI-compatible Conversations API state and history semantics and make explicit conversation continuation behave consistently with previous_response_id continuation.

Parent Epic: #2973

### Proposed Workgroup

Agentic & Context

### Current-main baseline

Current main already provides memory and Redis response/conversation store primitives, Router-owned response identities, previous_response_id chain loading, and conversation ID assignment.

An explicit conversation_id on a Responses request is currently recorded as identity but does not load that conversation's retained history. There is also no complete maintained Conversations API HTTP surface.

### Scope

- Define the supported create, retrieve, update, delete, item, and pagination subset for Router-owned Conversations objects.
- Materialize bounded retained history when a Responses request references an explicit conversation.
- Define validation and precedence for conversation and previous_response_id, including whether combinations are rejected and how branching is represented.
- Preserve Router-owned object identity, chronological ordering, metadata, TTL, deletion, and store-disabled or unavailable behavior.
- Bound history depth, item count, bytes, store calls, and response size.
- Keep memory and Redis behavior contract-compatible while documenting their consistency limits.
- Add protocol, persistence, restart, multi-replica, and E2E coverage for the maintained subset.

### Non-goals

- Treat provider-owned conversation state as Router-owned portable history.
- Promise full OpenAI platform behavior outside the documented maintained subset.
- Build a general application database, agent memory platform, or workflow engine.
- Silently merge incompatible branches or accept ambiguous continuation state.

### Completion signal

- [ ] The maintained Conversations API subset has explicit request, response, pagination, retention, and error contracts.
- [ ] Explicit conversation continuation loads the correct bounded history.
- [ ] conversation and previous_response_id combinations have deterministic branch or rejection behavior.
- [ ] Memory and Redis stores preserve ordering, isolation, TTL, and deletion semantics.
- [ ] Store-disabled, unavailable, partial, restart, and multi-replica behavior is covered.
- [ ] Maintained E2E tests exercise object lifecycle and conversation continuation.

Zheng-Lu:
Thanks for raising this issue! I agree with your goal and I would like to contribute to it. 

[...]

Issue #2814, opened 2026-08-07
[Bug] Remove full-keyspace Redis scans from conversation response lookup


gagandhakrey:
### Proposed Workgroup

Agentic & Context

### Describe the bug

RedisStore.ListResponsesByConversation scans every stored response key and performs an individual GET before filtering by conversation ID. Lookup cost therefore grows with the entire response dataset rather than the requested conversation.

Parent task: #2999  
Implementation candidate: #2817

### To Reproduce

Current main in src/semantic-router/pkg/responsestore/redis_store.go uses SCAN over response:* followed by one GET and JSON decode per key. The memory backend does not have this full-keyspace Redis cost.

The operation will become a production conversation-history path under #2999, so both performance and consistency must be fixed before relying on it.

### Expected behavior

Conversation response lookup should scale with the number of responses in the requested conversation, preserve deterministic ordering and pagination, and remain correct for existing data, expiry, updates, deletion, partial index writes, and Redis Cluster deployment.

### Impact

Not separately specified in the original report.

##### Proposed Workgroup

Agentic & Context

##### Scope

- Replace the full response-key scan with a bounded conversation-specific lookup contract.
- Maintain any secondary state consistently across create, update, move, delete, expiry, and conversation cascade deletion.
- Preserve or explicitly migrate responses written before the new lookup mechanism; existing retained responses must not silently disappear from listings or cascade deletion.
- Define repair behavior for missing, stale, or partial secondary state.
- Preserve chronological ordering, ListOptions semantics, duplicate-create behavior, and Redis Cluster compatibility.
- Add dataset-size-independent benchmarks and failure-injection coverage.

##### Completion signal

- [ ] No full response-keyspace scan remains in conversation lookup.
- [ ] Lookup work scales with the requested conversation plus a documented bound.
- [ ] Legacy and partially indexed data has a safe migration or fallback path.
- [ ] Create, update, move, expiry, delete, and cascade semantics remain consistent.
- [ ] Ordering and pagination match the documented store contract.
- [ ] Unit, integration, failure, and benchmark coverage demonstrates the result.

gagandhakrey:
I have drafted the pr for the same, once completed will change it to ready for review :  #2817

PR #3304, opened 2026-09-02
[Bug] Stop previous_response_id continuations from inheriting conversation_id membership

### LINKED ISSUE #3389: [Bug] previous_response_id continuations inherit conversation_id membership they never requested
### Summary

`determineConversationID` (in `extproc/req_filter_response_api.go`) inherits
the `conversation_id` of the earliest response in a `previous_response_id`
chain, and auto-mints a fresh one when neither field is present. A response
continued only via `previous_response_id` — never having explicitly opted into
a conversation — ends up tagged as a member of whatever conversation its
lineage happens to trace back through.

### Key problem

Given turns A, B, C sent with an explicit `conversation_id` (X), a later
request D sent with `previous_response_id` pointing at B (no `conversation_id`)
currently inherits `conversation_id: X` — both in its response body and in
what gets persisted to the store. A subsequent request E using
`conversation_id: X` is then positioned to pull D into X's history alongside
A/B/C, even though D represents a divergent branch the client never attached
to X.

Per OpenAI's Conversations API model (checked against the docs during design
discussion on #2999), a `conversation` is a durable, explicitly-joined
append-only log — membership comes from a request explicitly supplying
`conversation`, not from `previous_response_id` lineage. `previous_response_id`
is an orthogonal chaining mechanism and must not imply conversation
membership.

### Scope

- Make `ConversationID` strict external membership: populated only from the
  request's own explicit `conversation_id`, never inherited from
  `previous_response_id` history, never auto-minted.
- Preserve session-tracking/telemetry continuity through the change — today's
  code uses the (soon-strict) `ConversationID` as the router's internal
  session-correlation key too, so removing inheritance without a replacement
  would blank out cache-warmth estimation and turn telemetry for any
  `previous_response_id`-only continuation.
- Add test coverage: explicit `conversation_id` (membership populated),
  `previous_response_id`-only (membership empty, no leak), neither field
  (standalone, no membership).

### Completion criteria

- [ ] A `previous_response_id`-only continuation of a conversation-tagged
      response no longer inherits that conversation's ID, in both the
      response body and the persisted `StoredResponse`.
- [ ] Session telemetry / cache-warmth estimation / model-transition detection
      are unaffected by the now-common empty `ConversationID` case.
- [ ] `conversation_id` + `previous_response_id` together on one request
      remains rejected (already enforced in `llmprotocol/validate.go`).

### Related issues

- **Parent issue:** #2999
- **PR:** #3304

Zheng-Lu:
Fixes #3389

## Summary

`ResponseObjectState.ConversationID` (and its predecessor `determineConversationID`) previously inherited the `conversation_id` of the earliest response in a `previous_response_id` chain, and auto-minted a fresh one when neither field was present. That meant a response continued only via `previous_response_id` — never having explicitly opted into a conversation — would silently be treated as a member of whatever conversation its lineage happened to trace back through.

Concretely: if a client sends turns A, B, C under an explicit `conversation_id` (call it X), then later sends D with `previous_response_id` pointing at B (no `conversation_id`), D would inherit `conversation_id: X` in its response body and in storage — even though the client never attached D to X. A later request E using `conversation_id: X` would then be positioned to pull D into X's history alongside A/B/C, despite D representing a divergent, unrelated branch.

This fix makes `conversation_id` strict, spec-aligned external membership: populated only when a request explicitly supplies it, never inferred from `previous_response_id` lineage.

## Background

Found and worked through on #2999 (checked against OpenAI's Conversations API docs — a `conversation` object is a durable, explicitly-joined append-only log; `previous_response_id` is an orthogonal lineage mechanism and doesn't imply conversation membership). Per @VadimEisenberg's request on that thread, this lands as a standalone bug-fix PR ahead of the rest of the Conversations API work, specifically so the interpretation above gets reviewer confirmation before more is built on top of it.

Unrelated to #2814 / PR #2817 (the Redis O(N) scan fix) — this is a request-time semantics fix in `extproc`, not a store-layer change.

## What changed

*   **`req_filter_response_api.go`:** `determineConversationID` removed. `ConversationID` is now set directly from the request's own `conversation_id` — no chain inheritance, no auto-mint. A new `determineSessionTrackingID` computes a separate, always-non-empty, namespaced internal key (`respapi:conversation:`, `respapi:lineage:`, `respapi:standalone:`) so session/telemetry continuity doesn't regress now that the public field is often empty.
*   **`session_transition.go`, `session_telemetry.go`:** Session-ID derivation and the turn-recording gate switched from the public `ConversationID` to the new internal `SessionTrackingID`.
*   **`sessiontelemetry/telemetry.go`:** `resolve()` had its own independent gate and used the raw conversation ID as both the internal cumulative-state key and the logged `session_id` — same conflation, one level deeper. Fixed to key off `SessionTrackingID`; the real `conversation_id` is now logged as its own field instead of doubling as the correlation key.
*   **Test coverage:** Provided for all three `ConversationID` sources (explicit / lineage-only / standalone), the empty-`ConversationID`-must-not-blank-`SessionID` regression, and an end-to-end lineage-only-turn-still-records-telemetry test.

## Deliberately not in this PR

Tracked as follow-ups on #2999, not built here:
*   **Full branch isolation for the internal tracking key:** Two independent `previous_response_id`-only forks off the same parent still share `SessionTrackingID` today — no worse than before this fix, but not fully precise; needs a lineage ID propagated parent-to-child on `StoredResponse`.
*   **`POST /v1/conversations`:** Bounded/whole-history materialization for `conversation_id`-only requests — the actual feature work this bug-fix was a prerequisite for.
*   **`responseapi.GenerateConversationID()`:** Is now unused (auto-mint removed) and left in place rather than deleted here — small enough to fold into a future cleanup rather than widen this diff.

## Testing

No `go` toolchain locally; verified via `golang:1.25` in Docker against the real module:
*   `go build`, `go vet`, `gofmt -l` — clean across `extproc`, `sessiontelemetry`, `llmprotocol`.
*   `sessiontelemetry` full suite — passing, including new/updated tests.
*   `llmprotocol` mutual-exclusion test (`conversation_id` + `previous_response_id` together, already enforced pre-existing) — passing.
*   **`extproc` package tests could not be executed** in this environment: the package links against `candle-binding`/`ml-binding`/`nlp-binding` (Rust cgo artifacts not built here) and `valkey-glide` (requires cgo, no pure-Go fallback). `go vet` fully type-checks those test files, but CI or a full local dev environment needs to actually run them before merge — specifically `TestPrepareObjectStateConversationMembershipIsStrictlyExplicit`, `TestPopulateSessionTransitionFieldsUsesTrackingIDWhenConversationEmpty`, and `TestRecordSessionTurnRecordsResponseAPILineageOnlyTurn`.

## Validation

- [x] `gofmt -l` via `golang:1.25`
- [x] `git diff --check`
- [x] `go test ./pkg/llmprotocol ./pkg/sessiontelemetry`
- [x] `go vet ./pkg/llmprotocol ./pkg/sessiontelemetry`
- [x] `go vet ./pkg/extproc`
- [ ] `go test ./pkg/extproc`


FAUST-BENCHOU:
pls create a seperate issue to link

FAUST-BENCHOU (approved):
LGTM. Keeping this change focused on previous_response_id handling makes sense, and the follow-up work can address broader branch isolation semantics.

wilsonwu (approved):
LGTM, thanks!

Issue #2238, opened 2026-06-18
[Epic] Optimize routing recipes through an offline-to-online lifecycle


Xunzhuo:
## Parent Direction

MoM & Routing — #2965

## Goal

Define a data-driven MoM learning lifecycle spanning bounded online Router Learning, offline recipe improvement, candidate evaluation, approval, gray release, promotion, and rollback.

## Background

Routing improvements need a path from evidence to production that is reproducible and reversible. Online adaptation must stay inside validated candidates and protection rules; larger taxonomy, signal, policy, and recipe changes should remain offline and reviewable.

### Online adaptation

- Keep online changes inside validated candidate sets, protection envelopes, and typed adaptation contracts.
- Consume versioned, privacy-bounded outcomes and emit deterministic reasons.
- Define restart, persistence, and multi-replica state semantics.

### Offline recipe learning

- Build reproducible datasets from approved references, observations, and outcomes.
- Generate candidate recipes against explicit objectives and compare them with fixed/current baselines.
- Use held-out evaluation, uncertainty, and failure slices before proposing promotion.

### Promotion lifecycle

- Use candidate → evaluation → shadow → gray/canary → approval → active stages.
- Retain a known-good generation for rollback.
- Record artifact provenance, evaluation evidence, decision, and operational outcome.

## Current child work

- #2308 — benchmark contextual Thompson and LinUCB strategies against maintained baselines.
- #3376 — compare a bounded set of shadow arms with blinded judging while leaving the primary response unchanged.

Related evaluation protocol: #2346. Portable model-pool, recipe identity, evaluation-result, and model-card contracts: #2971.

## Boundaries and non-goals

- Router Memory is prompt-visible user or conversation memory, not a learning-experience store.
- Training and durable global optimization do not run synchronously on the request path.
- Router Model artifact training is owned by Router Models & Inference Runtime; this Epic consumes qualified artifacts.
- Do not mutate production recipes without reviewable evidence and rollback.
- Do not build a general-purpose MLOps platform or replace simple baselines without demonstrated benefit.

## Completion criteria

- [ ] Typed outcome, observation, and candidate contracts are documented.
- [ ] An offline baseline and evaluation workflow is reproducible.
- [ ] Approval, shadow, gray release, promotion, and rollback have explicit gates.
- [ ] Restart and multi-replica semantics are tested.
- [ ] Metrics, diagnostics, documentation, and E2E coverage make each decision auditable.

ramkrishs:
[...]

Issue #3280, opened 2026-09-01
[Feature] Export calibrated shadow comparison datasets with immutable provenance


Xunzhuo:
## Goal

Export calibrated primary-versus-shadow comparison datasets with immutable provenance for reviewed Router Model experiments.

## Proposed Workgroup

Router Models & Inference Runtime

## Background

Shadow observations are useful training evidence only when their inputs, outputs, judge policy, privacy state, and inclusion decisions are reproducible. Provider/model identity and presentation order can bias model-based judging, so those effects must be controlled and recorded.

Parent Epic: #2975. #3279 and its active implementation #3308 own the versioned live shadow-capture and Replay input contract; this issue must consume that contract rather than create another observation envelope.

## Work item shape and dependency

Manifest, blinded-judging, inclusion, split, and fixture work can proceed against versioned test observations. Integration and completion wait for #3279's capture/Replay contract to land. Model training and promotion remain separate work.

## Scope

- Convert approved #3279 primary-versus-shadow observations into a versioned dataset manifest.
- Preserve input/reference identity, primary and shadow model identities, output hashes, sampling policy, redaction state, and immutable artifact lineage.
- Present candidates to model-based judges in randomized order under stable opaque arm IDs and retain the private mapping only in authorized provenance.
- Record judge model/version, rubric/prompt version, bounded score/confidence, tie/abstention, malformed-output reason, and budget use without hidden reasoning.
- Define deterministic inclusion, exclusion, deduplication, balancing, and train/evaluation split rules.
- Treat low-confidence or invalid judgments as evidence, never as an automatic label or routing override.
- Validate schema, lineage, privacy, calibration, position/identity leakage, and split leakage before export.
- Support pluggable storage destinations without committing datasets to the source repository.

## Configuration and API impact

Add versioned offline export and manifest configuration only. No request-time selector, live traffic mutation, training trigger, or automatic promotion is introduced.

## Non-goals

- Capture live shadow traffic, define a second observation schema, train or promote a model, or treat an unreviewed judge score as ground truth.
- Store hidden judge reasoning or publish the private arm mapping without authorization.
- Feed exported evidence into the active Router automatically.

## Completion signal

- [ ] The exporter consumes the landed #3279 capture/Replay contract without a parallel envelope.
- [ ] The same approved inputs reproduce the same manifest and split identities.
- [ ] Every example traces to immutable, privacy-reviewed observation and judge evidence.
[...]

PR #3308, opened 2026-09-02
[Feature] Add bounded shadow model dispatch plugin with replay capture

### LINKED ISSUE #3279: [Feature] Add bounded shadow model dispatch and replay capture
### Goal

Add a bounded shadow-model call path that observes a secondary model without changing or delaying the primary response, providing the next-release foundation for the model flywheel.

### Proposed Workgroup

MoM & Routing

### Scope

- Add an explicit request-time shadow dispatch action that receives the same approved request context as the primary route.
- Keep the primary response independent from shadow success, failure, latency, and output.
- Bound sampling, concurrency, queue depth, timeout, response bytes, retries, and total resource use.
- Apply the configured redaction and trust-boundary policy before capture.
- Record immutable primary/shadow request identity, model identity, timing, outcome, and bounded comparison provenance for replay and audit.
- Define fail-open behavior for shadow execution and observable dropped/failed-shadow reasons.
- Cover standalone and maintained gateway paths with deterministic unit and E2E tests.

### Non-goals

- Judging output quality, producing a training dataset, or training a model.
- Letting a shadow response affect the live request.
- Sending unredacted prompts or credentials to an untrusted endpoint.
- Building a general asynchronous workflow engine.

### Completion signal

- Operators can enable sampled shadow calls with explicit resource and privacy bounds.
- The primary response remains unchanged when the shadow endpoint is slow, unavailable, malformed, or overloaded.
- Replay/audit evidence can correlate the primary and shadow observations without exposing protected content.
- Maintained E2E coverage proves failure isolation and bounded execution.

Parent Epic: #2238

yaojiejia:
Closes #3279

## Purpose

Add a `shadow_dispatch` route plugin. After the primary dispatch is finalized, it sends a sampled copy of the same approved request to a second configured model in the background and records the result on the request's replay record. The primary response never waits on the shadow or changes because of it.

Why: the MoM promotion lifecycle (#2238) needs evidence from live traffic before a candidate model reaches gray release. This is the shadow step.

What it does:

- Bounds sampling, concurrency, queue depth, timeout, response bytes, and retries per decision.
- Renders the shadow request through the same encode and provider-adaptation steps as a primary dispatch to that model.
- Uses only static router credentials for the shadow backend. Client headers and keys never travel with it. The shadow model must have a configured backend.
- Appends one append-only outcome to the primary replay record with request identity, models, backends, timing, attempts, status, sizes, stop reason, token counts, and a SHA-256 of the shadow text. Response text is stored only when `capture_response_body` is on, and then truncated.
- Emits `sr_shadow_dispatch_total{decision,result,reason}`, latency, in-flight, and queued metrics. Resource drops are reported through metrics and logs, not replay writes.
- Rejects the plugin on looper-executed algorithms at config load.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/config/ ./pkg/dsl/ ./pkg/k8s/ ./pkg/observability/metrics/
LD_LIBRARY_PATH=<rust libs> go test ./pkg/extproc/ -run ShadowDispatch -race
LD_LIBRARY_PATH=<rust libs> go test ./pkg/extproc/

cd src/vllm-sr
uv run --extra dev python -m pytest tests/test_plugin_shadow_dispatch.py tests/test_plugin_parsing.py

cd dashboard/frontend
npx vitest run src/lib/dslCapabilityPluginSchemas.test.ts && npx tsc --noEmit -p .

cd website && npm run config:check

.venv-agent/bin/python tools/agent/scripts/run_agent_precommit_lint.py <changed files>

# e2e (kind)
make e2e-test E2E_PROFILE=router-replay E2E_TESTS=shadow-dispatch-observes-candidate-model,shadow-dispatch-fail-open-unreachable-backend
```

## Test Result

- Go: config, dsl, k8s, metrics, and full extproc suites pass. 16 shadow tests pass under `-race`.
- Python: 34 tests pass.
- Dashboard: vitest and type check pass.
- Docs: configuration catalog check passes.
- Agent lint gate passes on all changed files.
- e2e: both new cases build and vet. Not yet run on a kind cluster. The failure case relies on `127.0.0.1:9` refusing connections.

---
<details>

</details>


FAUST-BENCHOU (commented):
/assign @Xunzhuo

subin9 (commented):
Nice to see this as a bounded plugin — the fail-open path and the lane bounds read well.

One ask from downstream. #3280 turns these outcomes into comparison datasets, and almost everything it needs is already in `shadowOutcomeMetadata` (`shadow_dispatch.go:507`).

Three things aren't, and they can only be recorded at capture time:

- **Primary response hash.** `response_sha256` covers the shadow text only. The dataset is a primary-vs-shadow pair, so without the same hash on the primary side an exported example isn't self-contained.

- **A slot for judge or label provenance.** Judging is out of scope here and I don't think it belongs in this PR, but #3280 has to separate reviewed from unreviewed evidence, and there's currently nowhere for that to attach. An optional key now is cheaper than migrating records already written.

- **Record-level redaction state.** The doc treats redaction as a viewer permission, which works inside the replay store but doesn't travel with an exported dataset. If the record carried what was applied, an export could prove it rather than relying on `capture_response_body` having been left off.

Raising it here rather than in #3280 only because records written before these exist can't be backfilled.

Happy to take it as a follow-up if you'd rather keep this PR as is.

Xunzhuo (approved):
The shadow dispatch stays off the primary latency path, applies per-decision bounds, reuses the protocol and provider contracts, and records bounded replay outcomes with focused unit and E2E coverage. LGTM from this pass.

Xunzhuo (changes requested):
i got one concern is that we have other client, can we unify them? like i can merge https://github.com/vllm-project/semantic-router/pull/3328 in first and try to see how we can have one clean code here and dig if we have any other dirty codes

FAUST-BENCHOU (commented):
Could we clarify how shadow request failures are recorded in replay data? Timeout, backend error, and rejected requests may have different meanings when evaluating candidate models.

yaojiejia:
> i got one concern is that we have other client, can we unify them? like i can merge #3328 in first and try to see how we can have one clean code here and dig if we have any other dirty codes

I see. Since #3328 is merged, I can rebase and move the shadow HTTP path (retries, status checks, bounded reads, the two `http.Client`s) onto `pkg/modelruntime/connector`, keeping only target resolution, encoding, response translation, and outcome recording in extproc.

The connector needs three small additive changes, which I'll include with tests: per-call headers, an optional TLS config, and a result that also returns status code and attempt count.

yaojiejia:
> Could we clarify how shadow request failures are recorded in replay data? Timeout, backend error, and rejected requests may have different meanings when evaluating candidate models.

Sure just added "Interpreting failures" section to the tutorial, it groups the reasons by what they say about the candidate: rejected request (4xx), health or capacity (5xx, timeout, transport), output problem (malformed, too large), and router-side config errors that should be excluded from any comparison. Dropped and sampled-out shadows never write to replay, so they can't be mistaken for candidate failures.

Xunzhuo (changes requested):
`shadowExtraHeaders` forwards arbitrary decision header mutations to the shadow connector. A primary `Authorization`, `x-api-key`, token, or custom secret can survive when the shadow is keyless or uses different auth, crossing the untrusted shadow boundary. Please filter sensitive mutations and add a wire-level regression; the maintained router-replay E2E also needs an exact-head run.

yaojiejia:
> `shadowExtraHeaders` forwards arbitrary decision header mutations to the shadow connector. A primary `Authorization`, `x-api-key`, token, or custom secret can survive when the shadow is keyless or uses different auth, crossing the untrusted shadow boundary. Please filter sensitive mutations and add a wire-level regression; the maintained router-replay E2E also needs an exact-head run.

Thanks for the review. Whats fixed in 31da41e4: credential-bearing header mutations (`Authorization`, `x-api-key`, `api-key`, `Cookie`, the `x-user-*-key` headers, and either profile's configured `auth_header`) are now stripped before the shadow call. Added a wire-level regression against a keyless shadow backend that fails on the old code, plus unit tests for the filter. Full `router-replay` E2E on kind passes 7/7 at this head (1d6426cf).

Xunzhuo (changes requested):
Thanks for adding the two-stage filter and wire-level coverage. The trust-boundary blocker is still open for arbitrary decision secrets: `shadowHeaderIsSensitive` denies only a fixed set plus the two provider `auth_header` names, so a mutation such as `X-Internal-Token: <secret>` is still copied into `job.extraHeaders` and sent to the shadow backend. Please make shadow-forwarded decision headers opt-in or allowlisted, or otherwise conservatively reject custom credentials, and add that wire case before rerunning the maintained router-replay E2E.

yaojiejia:
> Thanks for adding the two-stage filter and wire-level coverage. The trust-boundary blocker is still open for arbitrary decision secrets: `shadowHeaderIsSensitive` denies only a fixed set plus the two provider `auth_header` names, so a mutation such as `X-Internal-Token: <secret>` is still copied into `job.extraHeaders` and sent to the shadow backend. Please make shadow-forwarded decision headers opt-in or allowlisted, or otherwise conservatively reject custom credentials, and add that wire case before rerunning the maintained router-replay E2E.

Thanks for the review and fixed in 70ff32a3. Decision header mutations are now opt-in: `shadow_dispatch.forward_headers` is an allowlist (empty by default), so an unlisted mutation such as `X-Internal-Token` never reaches the shadow. Known credential headers and either profile's `auth_header` are rejected by config validation and still dropped at run time even if listed. Added wire cases for the default (nothing forwarded), an allowlisted `x-tenant`, and credentials forced past validation, plus config, CLI, DSL, and dashboard coverage. Reran the maintained `router-replay` E2E on kind at this head: 7/7 passed.

Xunzhuo (changes requested):
Thanks for making decision-header forwarding opt-in. One trust-boundary leak remains: with tracing enabled, request processing extracts client baggage into ctx.TraceContext, the global propagator includes propagation.Baggage{}, and shadowExtraHeaders injects every propagated header before the forward_headers filter. A client-supplied baggage: token=... therefore reaches the untrusted shadow even with an empty allowlist. Please restrict shadow propagation to non-baggage trace headers, or sanitize or drop baggage, and add a wire-level regression.

yaojiejia:
> Thanks for making decision-header forwarding opt-in. One trust-boundary leak remains: with tracing enabled, request processing extracts client baggage into ctx.TraceContext, the global propagator includes propagation.Baggage{}, and shadowExtraHeaders injects every propagated header before the forward_headers filter. A client-supplied baggage: token=... therefore reaches the untrusted shadow even with an empty allowlist. Please restrict shadow propagation to non-baggage trace headers, or sanitize or drop baggage, and add a wire-level regression.

Good catch. The shadow now injects through a fixed W3C TraceContext propagator, so only `traceparent` and `tracestate` cross and client `baggage` never does, whatever the global propagator carries. Added a wire-level regression that sends client baggage through the real request-header phase and asserts the shadow backend never receives it, and the maintained router-replay E2E passed 7/7 at 4186850a.

Xunzhuo (commented):
Thanks—shadow dispatch now uses a fixed W3C TraceContext propagator, so client baggage cannot cross the shadow boundary while trace continuity is preserved. The exact-head wire regression covers the original leak. Please rebase onto current `main` for the final high-risk approval pass.

yaojiejia:
> Thanks—shadow dispatch now uses a fixed W3C TraceContext propagator, so client baggage cannot cross the shadow boundary while trace continuity is preserved. The exact-head wire regression covers the original leak. Please rebase onto current `main` for the final high-risk approval pass.

Thanks, just rebased earlier today and seems like all ci are green :)

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
```diff
@@ -0,0 +1,307 @@
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if client, ok := d.clients[key]; ok {
+		return client, "", nil
+	}
+	authorize, err := shadowAuthorizer(job.routerConfig, target.profile, target.logicalModel)
+	if err != nil {
+		return nil, shadowReasonCredentialUnresolved, err
+	}
+	client, err := connector.New(target.baseURL, authorize, shadowConnectorOptions(job.cfg))
```
Xunzhuo, line 134: **Reject cross-origin redirects from shadow backends**

Impact: A shadow backend can return 307 or 308 and cause the production prompt body, plus provider headers Go does not strip cross-origin such as api-key or x-api-key, to be sent to an unconfigured origin, violating the configured-target trust boundary. The connector created here uses the default redirect policy in src/semantic-router/pkg/modelruntime/connector/connector.go:91.

Required change: Reject redirects or permit only explicitly validated same-origin redirects without forwarding sensitive headers or bodies cross-origin; classify rejections visibly and add a cross-origin 307 regression proving prompts and credentials never leave the configured origin.

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
```diff
@@ -0,0 +1,307 @@
...
+
+// shadowEndpointPath mirrors setProviderRequestPath so the shadow reaches the
+// same provider path the primary dispatch would use for that wire format.
+func shadowEndpointPath(profile *config.ProviderProfile, format llmprotocol.WireFormat) string {
+	path := requestWirePath(format)
+	if profile == nil {
+		return path
+	}
+	if format == llmprotocol.OpenAIChatV1 {
+		if configured, err := profile.ResolveChatPath(); err == nil && configured != "" {
```
Xunzhuo: **Preserve Azure API-version query parameters**

Impact: Azure OpenAI shadow targets configured with api_version fail locally before any request is sent, so this supported provider cannot produce shadow observations.

Required change: Parse the resolved endpoint into a validated path and RawQuery, or otherwise construct a validated operation URL, rather than passing query text through Operation.Path; add an Azure shadow-target regression.

Xunzhuo (commented):
Thanks for the update—I found 2 current-head issues that need to be addressed before this can merge.

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
yaojiejia: Thanks! This is fixed in 625897936. The connector now never follows a redirect: CheckRedirect returns http.ErrUseLastResponse, so a 301/302/303/307/308 is returned as a non-retryable KindRedirect error and the shadow records it as redirect_rejected with the 3xx status_code. The error names only the redirect target's scheme and host, never its path. Regressions: a connector test covers cross-origin 307/308/302, same-origin 307, and relative 301, asserting the redirect target receives zero requests; a shadow-level test sends a 307 to a second origin and proves the prompt and Bearer credential reached only the configured backend, once, with no retry. Docs list the new reason.

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
yaojiejia: Thanks, this is fixed in e47950536. connector.Operation now carries a separate validated Query field, and the shadow target parses the resolved endpoint with splitShadowEndpoint, which requires an absolute path and rejects a scheme, host, user info, fragment, or unparsable query before any request is sent. Regressions: a shadow test with an azure-openai profile and api_version asserts the backend receives /openai/deployments/gpt-4o/chat/completions?api-version=2024-02-01 with the api-key header, plus table tests for the splitter and for connector query validation. Verified the test fails with encode_failed on the previous head.

Xunzhuo (approved):

Issue #2975, opened 2026-08-22
[Epic] Build a reproducible Router Model training and artifact pipeline


Xunzhuo:
### Goal

Make Router Model datasets, training runs, evaluation outputs, exports, qualification, promotion, and rollback reproducible and traceable.

### Scope

- Define versioned dataset, run, evaluation, and artifact manifests.
- Integrate provenance validation into existing task-specific workflows incrementally.
- Prove the complete pipeline on one maintained built-in model with a documented quality gap.
- Produce immutable artifact identity and inputs for qualification and model cards.

### Child issues

- #3199
- #3200

### Non-goals

- Rewriting every training script at once.
- Hosting datasets or model artifacts in the repository.
- Automatic online training or promotion.

Parent Workgroup: #2966

guan404ming:
Hi @Xunzhuo, could I help with this epic and also its child issues #3199, #3200, and #3225?

Here is my rough plan:

- #3199: versioned manifest schemas 
  1. dataset, training-run, evaluation, and artifact manifest schemas, with validation wired into one existing task-specific workflow, plus docs.
- #3200: prove the pipeline on one built-in model
  1. export and qualification receipt tooling with immutable artifact identity.
  2. run one maintained built-in model through the full path (dataset -> train -> eval -> export -> qualification receipt), with the documented quality gap and model card.
- #3225: int8 variants as manifest-declared artifacts
  1. fix the provider-unaware model-file selection (int8 on CPU, FP16 only on GPU, per-classifier precision override); independent of quantization, can land first.
  2. dynamic int8 quantization step in the export path and models.mk targets, with the fp32-vs-int8 benchmark extension.
  3. quality gate on the measured fp32-vs-int8 delta for jailbreak and modality on VNNI and non-VNNI x86, publishing recalibrated thresholds where needed.

Ordering: #3199 first, then #3200 and the #3225 quantization PRs on top of it; the #3225 selection fix has no dependency and can go anytime. Please feel free to let me know if there is anything that needs to be refined, thanks!

PR #3310, opened 2026-09-02
[Feature] Add Router Model provenance manifest contract

guan404ming:
Related #2975
Related #3199

## Purpose

- Add `src/training/provenance`: versioned dataset, run, evaluation, and artifact manifests with sha256 identities.
- Validate completeness, source licenses, cross-references by identity, artifact file digests, and secret-like fields.
- Ship `python -m provenance` for bundle validation. Training-script integration and CI gating follow in later PRs.

## Test Plan

- `cd src/training && python -m pytest provenance/test_provenance.py`
- `make agent-ci-lint CHANGED_FILES="src/training/provenance/*"`

## Test Result

- 14 passed. Lint and structure gates pass.


FAUST-BENCHOU (commented):
The provenance manifest contract is a good foundation for reproducible Router Models. One thing to consider is making the relationship between artifact identity and evaluation identity explicit, so an evaluation result cannot accidentally be attached to a different artifact revision.

guan404ming:
Thanks. Today the link is one-directional: `artifact.evaluation_id` pins the evaluation manifest by identity, and validation rejects an artifact that points at a different evaluation. The reverse binding (evaluation records the artifact digest it ran against) needs the eval step to consume the exported artifact, which lands in the next PR that wires eval/export emission. I will add `artifact_digest` to the evaluation manifest there and cross-check it in `validate_bundle`.

Xunzhuo (dismissed):
Thanks for putting the provenance contract together. Two validation blockers remain on this head:

- `sample_count` and `command` are described as required contract fields but are absent from `REQUIRED_FIELDS`, so zero-sample or empty-command evaluation manifests pass validation.
- Loaded JSON is only checked for field presence and digest self-consistency. Negative or string split counts, non-finite or string metrics, and malformed IDs/digests can still be accepted. Please validate the v1 field types, ranges, identifier formats, and digest formats before accepting a manifest.

Please also link a bounded accepted issue for this implementation; #3199 still needs acceptance, while #2975 is the parent epic.

FAUST-BENCHOU (commented):
The manifest validation looks good. Could we document where in the artifact lifecycle this validation should be invoked (for example before publishing or loading artifacts) so provenance checks are consistently applied?

guan404ming:
Thanks both. Fixed in 804391ce.

- `sample_count` and `command` now required; zero or empty fails.
- Added type, range, and format checks: `sha256:<64 hex>` ids and digests, non-negative split counts, finite metrics, typed strings and maps. 13 negative tests.
- README adds a "When to validate" table: after training, after eval, before publish, on artifact load.
- #3199 is still `needs-acceptance` and #2975 is the epic. Please accept #3199 or name the bounded issue and I will relink.

Xunzhuo (dismissed):
Thanks for the thorough validation follow-up. `_type_errors` records wrong nested types, but `_nested_errors` then calls `.get()`, so inputs such as `sources: [1]` escape with `AttributeError` instead of a bounded validation error. Please gate nested validation and cover wrong containers/members. Also validate serialized `schema_version` and source field types; #3199 still needs acceptance and coordination with #3247.

guan404ming:
Thanks! Nested checks now run only when the container type is valid, source fields must be strings, and serialized `schema_version` is covered, with tests for wrong containers and members. #3199 acceptance and #3247 coordination still pending on your side.

Xunzhuo (commented):
Thanks—the container guards plus source and schema-version checks resolve my validation blocker on this head. I do not see another code issue in the update. #3199 still needs acceptance and coordination with #3247, the branch is behind main, and the refreshed baseline jobs are queued; once those gates settle, I will take the final pass.

Xunzhuo (commented):
One malformed-input path still escapes the provenance CLI's validation boundary.

src/training/provenance/manifests.py
```diff
@@ -0,0 +1,109 @@
...
+
+def manifest_id(manifest: Manifest) -> str:
+    return (
+        "sha256:"
+        + hashlib.sha256(canonical_json(manifest.to_dict()).encode()).hexdigest()
+    )
+
+
+def manifest_from_dict(data: dict[str, Any]) -> Manifest:
+    kind = data.get("kind")
```
Xunzhuo, line 96: **Validate the decoded JSON shape before dispatch**

Impact: `load_manifest` passes arbitrary JSON here, so an array or scalar raises `AttributeError` at `.get`, while a list/dict `kind` raises `TypeError` during membership. Neither exception is caught by the CLI, turning malformed manifests into tracebacks instead of the promised controlled validation failure.

Required change: reject non-object roots and non-string `kind` values with the handled validation error, and cover those malformed shapes in the CLI tests.

guan404ming: Thanks, good catch.
`manifest_from_dict` now rejects non-object roots and non-string `kind` values with the handled `ValueError`, and a parametrized CLI test covers array, scalar, list-kind and dict-kind manifests.

guan404ming:
Closing in favor of #3247, which landed first. The per-source license check moves to a follow-up PR on that module.

PR #3311, opened 2026-09-02
[Bug] Support NVIDIA Dynamo nvext in requests and responses

### LINKED ISSUE #3136: [Bug] Support NVIDIA Dynamo nvext in requests and responses
Parent Epic: #2358

### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The strict protocol codec rejects NVIDIA Dynamo top-level nvext data in both directions.

A request containing nvext fails with invalid_json because the request contains a non-canonical field. A Dynamo response containing nvext is rejected as invalid upstream JSON and becomes a 502.

Dynamo documents nvext as an extension on its OpenAI-compatible request and response contract:
https://github.com/ai-dynamo/dynamo/blob/main/docs/fern/pages/developer-guide/additional-resources/nvidia-request-extensions-nvext.md

### To Reproduce

1. Send an OpenAI-compatible request containing top-level nvext through the Router to a Dynamo backend.
2. Or configure Dynamo to return nvext in its response.
3. Observe that the Router rejects the request or upstream response before it can be forwarded.

### Expected behavior

Support nvext as an explicit, bounded Dynamo provider extension on request and response paths, including streaming where Dynamo can emit it.

This must not globally relax canonical unknown-field rejection:

- Model nvext explicitly and enforce size and nesting limits.
- Preserve it on compatible same-format Dynamo paths.
- Define cross-format and non-Dynamo behavior explicitly; reject unsupported use instead of silently dropping fields.
- Respect the documented header-versus-body precedence.
- Treat isolation and routing fields such as cache_salt as untrusted input, not authorization.

### Impact

A maintained Dynamo integration cannot carry the documented request extension or consume Dynamo responses that include it.

### Acceptance criteria

- Buffered request and response paths accept and preserve valid bounded nvext.
- Streaming response handling is covered.
- Unsupported cross-format use fails explicitly.
- Arbitrary unknown fields remain rejected.
- Unit and Dynamo E2E coverage include valid, oversized, malformed, and trust-boundary cases.

### Additional context

Verified against current main: both request and upstream response decoding reject top-level nvext.

cr-zhao:
Closes #3136

## Purpose

Support NVIDIA Dynamo top-level `nvext` as an explicit, bounded provider extension without relaxing canonical unknown-field rejection globally.

This change:
- adds typed request, response, and streaming `nvext` contracts aligned with the current Dynamo documentation;
- preserves `nvext` through Chat Completions and Responses codec paths;
- enforces size and nesting limits at the provider boundary;
- enables the extension only for backends explicitly configured with `backend_refs[].type: dynamo`;
- extends mock-vLLM and the response-api E2E profile with buffered and SSE contract coverage.

The implementation intentionally uses `backend_refs[].type: dynamo` as the provider boundary. Maintainer confirmation that `type` is the preferred long-term configuration surface would be appreciated.

## Test Plan

- `make agent-ci-gate CHANGED_FILES="<changed files>"`
- `python -m pytest -q tests/test_provider_contract.py` from `tools/mock-vllm`
- `make e2e-test E2E_PROFILE=response-api E2E_TESTS=dynamo-nvext-chat-contract`

## Test Result

- Agent CI gate passed after synchronizing with current `origin/main`, including lint, the Semantic Router core test suite, and E2E binary build.
- mock-vLLM provider contract tests: 12 passed.
- Targeted Kind/Envoy E2E: 1 passed, 0 failed; both buffered and SSE `nvext` paths were exercised.

---
<details>

- [x] The PR links accepted issue #3136
- [x] Purpose, Test Plan, and Test Result reflect the actual scope and results

</details>


FAUST-BENCHOU (commented):
1. Could we clarify the expected behavior when `nvext` is used with multiple backends?

Currently it looks like the request is allowed only if all configured backends are Dynamo. But if routing selects a Dynamo backend at runtime, should we only validate the selected backend instead? This may affect fallback scenarios where a pool contains both Dynamo and non-Dynamo backends.

2. Should we add an explicit version field for `nvext`?

Since this is a provider-specific extension, Dynamo may add or change fields in the future. A version marker could help the router handle future schema changes more safely.

cr-zhao:
Thanks for the suggestions! @FAUST-BENCHOU For these two issues:

1. Considering that the Router first selects a model and Envoy then selects an endpoint from that model’s backend pool, selecting a model with a mixed Dynamo/non-Dynamo backend pool cannot guarantee that Envoy will ultimately choose a Dynamo endpoint.
To avoid changing the existing backend-selection logic, the current implementation takes the most conservative and minimal approach: requests containing nvext are only routed to models whose backend pools consist entirely of Dynamo endpoints.
I agree that this may be too restrictive as a long-term solution. A possible better design would allow the Router to select any model with at least one Dynamo endpoint and then constrain Envoy to choose the best endpoint only from the Dynamo subset. However, that would require changes to the existing filtering and endpoint-selection behavior, so I'm thinking of handling that it in a follow-up PR.

2. I agree that a version marker could be useful. However, the current official nvext contract does not define a version field. Using the Dynamo version instead would also be difficult because the nvext payload is generated by the client and cannot reliably indicate the version of the Dynamo backend that will eventually be selected. A capability-based approach may be more appropriate—for example, matching the capabilities required by the nvext payload against those declared by the backend—but the exact contract and matching mechanism would need further discussion and design.

Xunzhuo (changes requested):
Thanks for taking this on, and welcome! I found one blocker: the current Dynamo nvext.extra_fields contract includes prompt_token_ids, but supportedDynamoExtraFields rejects it and DynamoResponseNVExt cannot preserve the response field. Please add bounded request, response, and stream handling plus regression coverage for prompt_token_ids so this path covers the documented extension set.

cr-zhao:
@Xunzhuo Thanks for catching this. I’ve added `prompt_token_ids` to the supported `nvext.extra_fields` set and implemented bounded preservation across buffered and streaming response paths. I also added unit, mock-provider, and E2E regression coverage. The fix includes in commit a327270.

Xunzhuo (dismissed):
Thanks for the thorough follow-up. The current head now preserves `prompt_token_ids` across the bounded nvext request, response, and streaming contracts, with unit, mock, and E2E coverage, so my earlier blocker is resolved. CI is green and local agent validation passes. Approved; please rebase onto current main before merge.

FAUST-BENCHOU (dismissed):
cool

wilsonwu (dismissed):
LGTM

cr-zhao:
Thanks @Xunzhuo @FAUST-BENCHOU @wilsonwu for the apporvals. I rebased the branch but looks like the repository’s stale-review policy automatically dismissed the previous approvals.

Xunzhuo (commented):
Thanks for the thorough update. I found three compatibility gaps at this head:

- Add typed, bounded `nvext` handling to strict Responses request and response codecs; same-format `/v1/responses` payloads currently reject it as unknown.
- Apply the Dynamo-only backend check to header-only requests, and reject response `nvext` unless the selected backend is Dynamo.
- Handle Dynamo's annotated `event: request_id` SSE frame before strict Chat-chunk decoding; its data is not a Chat completion chunk and currently fails decoding.

Could you add focused coverage for these paths?

Xunzhuo (changes requested):
Thanks—the Responses, header-only, and annotated request-id gaps are now covered. One backend-boundary issue remains: captureUpstreamBackendIdentity reads xds.upstream_host_metadata as a nested Struct, but Envoy delivers that requested Metadata value in its serialized attribute representation. The exact-head Dynamo E2E therefore leaves the backend identity empty and rejects a valid nvext response with unexpected_dynamo_nvext_backend. Please parse the real Envoy representation, or request scalar metadata leaf attributes, cover that shape, and rerun the required checks.

cr-zhao:
> Thanks—the Responses, header-only, and annotated request-id gaps are now covered. One backend-boundary issue remains: captureUpstreamBackendIdentity reads xds.upstream_host_metadata as a nested Struct, but Envoy delivers that requested Metadata value in its serialized attribute representation. The exact-head Dynamo E2E therefore leaves the backend identity empty and rejects a valid nvext response with unexpected_dynamo_nvext_backend. Please parse the real Envoy representation, or request scalar metadata leaf attributes, cover that shape, and rerun the required checks.

I adopted the second approach and now request the scalar metadata leaf attributes for `backend_name` and `backend_type` directly. I also updated the focused unit/config-generation coverage and the response-api E2E fixture to exercise the actual selected-backend metadata path. CI failure fixed. Thanks for the suggestion.

Xunzhuo (commented):
Thanks—the scalar response attributes now preserve the actually selected backend identity and fail closed when either identity field is missing. The exact-head response-api and streaming checks cover the Dynamo path, so my prior backend-identity blocker is resolved. Please rebase onto current `main`; after that this is ready for the final approval pass.

cr-zhao:
rebased

Xunzhuo (changes requested):
Thanks for the thorough update. One blocker remains: Dynamo backend validation still runs after the response-cache fast path. Cache identity deliberately excludes the Dynamo envelope, so a cached ordinary request can let an otherwise identical nvext request to a non-Dynamo model return 200 instead of being rejected; distinct `cache_salt` values can also collide. Please move this boundary check before every cache or fast response and add a cache-hit regression. Also wire the maintained Dynamo values and selected-upstream response attributes into this contract.

cr-zhao:
> Thanks for the thorough update. One blocker remains: Dynamo backend validation still runs after the response-cache fast path. Cache identity deliberately excludes the Dynamo envelope, so a cached ordinary request can let an otherwise identical nvext request to a non-Dynamo model return 200 instead of being rejected; distinct `cache_salt` values can also collide. Please move this boundary check before every cache or fast response and add a cache-hit regression. Also wire the maintained Dynamo values and selected-upstream response attributes into this contract.


fixed and rebased

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -67,6 +68,43 @@ func (r *OpenAIRouter) runRequestPreRoutingStages(
...
+	originalModel string,
+	selectedModel string,
+	decisionName string,
+	ctx *RequestContext,
+) *ext_proc.ProcessingResponse {
+	targetModel := selectedModel
+	if targetModel == "" {
+		targetModel = originalModel
+	}
+	if err := validateDynamoBackendPool(r.Config, targetModel, ctx, ctx.ProtocolEnvelope); err != nil {
```
Xunzhuo, line 103: **Partition or bypass response cache for Dynamo extensions**

Impact: Requests with identical neutral messages and models but different token_data, use_raw_prompt, routing headers, or cache_salt can share exact or semantic cache entries, returning a response produced for different model input or routing without contacting Dynamo.

Required change: Bypass response-cache reads and writes whenever hasDynamoRequestExtension is true, or define an extension-aware exact key containing every response-affecting envelope and header field while disabling semantic reuse when raw token input can differ; add same-message/different-extension regressions.

Xunzhuo (commented):

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
cr-zhao: now bypass cache reads and writes when request has dynamo extensions. tests added. Thanks for catching this.

cr-zhao:
The three E2E failures appear to be unrelated to this PR. They fail during the ONNX-backed router build because the ONNX binding is missing the recently introduced TextWindows API, before any tests for this PR are executed. I have opened #3680 to address that issue separately.

src/semantic-router/pkg/extproc/req_filter_dynamo_nvext.go
```diff
@@ -0,0 +1,198 @@
...
+		GetEndpointsForModel(string) []config.VLLMEndpoint
+	},
+	model string,
+	ctx *RequestContext,
+	envelope llmprotocol.Envelope,
+) error {
+	if !hasDynamoRequestExtension(ctx, envelope) {
+		return nil
+	}
+	if !modelHasOnlyDynamoBackends(cfg, model) {
```
Xunzhuo, line 91: **Keep the maintained Dynamo values inside this provider boundary**\n\nImpact: The documented `deploy/kubernetes/dynamo/semantic-router-values/values.yaml` still sets the frontend backend to `provider: vllm`. Canonical materialization copies that provider ID into `VLLMEndpoint.Type`, so every `nvext`, top-level `cache_salt`, or Dynamo-routing-header request through the maintained Dynamo installation fails this guard with 400.\n\nRequired change: Use the new `dynamo` provider in the maintained values, update the installation guidance as needed, and cover this actual configuration path.

src/semantic-router/pkg/extproc/processor_backend_identity.go
```diff
@@ -0,0 +1,46 @@
...
+		return
+	}
+	// Response attributes are authoritative for the endpoint Envoy actually
+	// selected. Clear any earlier value before parsing so missing or malformed
+	// metadata fails closed, including after an upstream retry.
+	ctx.UpstreamBackendName = ""
+	ctx.UpstreamBackendType = ""
+	ctx.AllowDynamoExtensions = false
+
+	attributes := req.GetAttributes()[extProcAttributesNamespace]
```
Xunzhuo, line 30: **Expose selected-backend identity in the maintained Gateway path**\n\nImpact: The documented Dynamo `EnvoyPatchPolicy` requests no response attributes, and its routed upstream endpoints carry none of the required `semantic-router` metadata. This lookup therefore always finds no attributes, leaves `AllowDynamoExtensions` false, and rejects valid Dynamo `nvext` responses; the standalone Envoy-template change does not affect this Gateway API path.\n\nRequired change: Update the maintained Gateway manifest to request the scalar backend identity leaves and attach matching metadata to the selected upstream, then cover that deployed path.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -72,6 +73,43 @@ func (r *OpenAIRouter) runRequestPreRoutingStages(
...
```
Xunzhuo, line 103: **Enforce the Dynamo boundary for shadow dispatch**\n\nImpact: This guard validates only the primary model. The later shadow path re-encodes `ctx.ProtocolEnvelope` for an independently configured target without checking its provider: a same-format non-Dynamo shadow receives the client’s `nvext`/`cache_salt`, while a cross-format shadow fails encoding and silently stops shadowing.\n\nRequired change: Validate the resolved shadow target’s provider and format before dispatch (or disable shadowing for Dynamo-extension requests), and add Dynamo-primary/non-Dynamo-shadow coverage.

Xunzhuo (commented):
Thanks—the cache bypass is fixed, but three current-head Dynamo integration paths still need correction before this can merge.

src/semantic-router/pkg/extproc/req_filter_dynamo_nvext.go
cr-zhao: Updated the maintained Dynamo values to use `provider: dynamo` and added regression coverage for the generated backend identity in aa81c108a.

cr-zhao: Changed the provider to dynamo and add related test. Fixed in 157a172

src/semantic-router/pkg/extproc/processor_backend_identity.go
cr-zhao: b70d617 This commit requests the selected upstream backend identity through the EnvoyPatchPolicy and attaches the corresponding backend_name and backend_type metadata to every Dynamo lb_endpoint discovered through EDS.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
cr-zhao: Validates Dynamo-extension requests during shadow dispatch preparation, ensuring that the resolved shadow backend is Dynamo and that the source wire format matches the target format before dispatch. Add coverage test. Fixed in 3ffc0a6

src/semantic-router/pkg/extproc/shadow_dispatch.go
```diff
@@ -404,6 +408,8 @@ func (d *shadowDispatcher) submit(
 		engine:          deps.engine,
 		encode:          deps.encode,
 		request:         *ctx.SemanticRequest,
+		hasDynamoExt:    hasDynamoRequestExtension(ctx, ctx.ProtocolEnvelope),
```
Xunzhuo, line 411: **Preserve header-only Dynamo routing in shadow calls**\n\nImpact: hasDynamoRequestExtension also matches x-dynamo-* and x-tenant-id inputs, so this permits a same-format Dynamo shadow, but the job retains only a boolean and shadowCallHeaders deliberately omits client headers. Header-only routing and header-over-body precedence therefore disappear, so the shadow can run under different tenant, worker, rank, or priority semantics.\n\nRequired change: Snapshot and forward the validated Dynamo routing headers only to an allowed same-format Dynamo target, or explicitly reject shadowing when Dynamo semantics come from headers; cover header-only and conflicting header/body cases.

Xunzhuo (commented):
Thanks for closing the earlier Dynamo boundaries. One shadow-routing gap remains on this head.

src/semantic-router/pkg/extproc/shadow_dispatch.go
cr-zhao: Snapshot the validated Dynamo routing headers when creating the shadow job. Done in f65c8e3. Thanks for catching this.

Xunzhuo (commented):
Thanks for fixing header-only shadow routing; one current-head boundary gap still needs to be closed.

src/semantic-router/pkg/extproc/shadow_dispatch.go
```diff
@@ -404,6 +409,9 @@ func (d *shadowDispatcher) submit(
...
+		sourceFormat:    ctx.ProtocolEnvelope.Format,
+		dynamoHeaders:   snapshotDynamoRoutingHeaders(ctx),
```
Xunzhuo: **Snapshot the effective validated routing state**

Impact: This captures raw client Dynamo headers. A Responses request never runs `validateDynamoRoutingHeaders`, and later shadow assembly applies this map after profile and decision sets while ignoring decision deletes. The shadow can therefore receive malformed or operator-replaced/deleted tenant, worker, rank, or priority values that the primary path would not use.

Required change: Build the shadow routing-header snapshot from validated effective outbound state after profile and decision set/delete semantics (or reject unsupported source formats), then cover malformed Responses headers, a profile override, and a decision update/delete.

cr-zhao: Moved the snapshot after the profile and decision. Thanks.

src/semantic-router/pkg/extproc/shadow_dispatch.go
```diff
@@ -404,6 +435,9 @@ func (d *shadowDispatcher) submit(
 		engine:          deps.engine,
 		encode:          deps.encode,
 		request:         *ctx.SemanticRequest,
+		hasDynamoExt:    hasDynamoRequestExtension(ctx, ctx.ProtocolEnvelope),
```
Xunzhuo: **Gate the effective Dynamo routing state**

Impact: When a provider profile or decision injects the only Dynamo routing header, `dynamoHeaders` contains that effective value but `hasDynamoExt` remains false because it examines only the original request. The shadow path then skips the Dynamo-only backend and same-format guards and can forward tenant or routing state to a non-Dynamo or cross-format target.

Required change: Derive this gate from the effective validated snapshot (or otherwise couple forwarding to the same boundary), and add profile/decision-only header regressions with non-Dynamo and cross-format shadow targets.

Xunzhuo (commented):
Thanks for the careful follow-up. One effective-state boundary still needs to be closed before this can merge.

src/semantic-router/pkg/extproc/shadow_dispatch.go
cr-zhao: Added a parallel check for whether the effective outbound headers contain any Dynamo routing state.

src/semantic-router/pkg/llmprotocol/dynamo_nvext.go
```diff
@@ -0,0 +1,326 @@
...
+type DynamoWorkerInfo struct {
+	PrefillWorkerID *uint64
+	PrefillDPRank   *uint32
+	DecodeWorkerID  *uint64
+	DecodeDPRank    *uint32
+}
+
+var supportedDynamoExtraFields = map[string]struct{}{
+	"worker_id": {}, "timing": {}, "routed_experts": {}, "engine_data": {},
+	"stop_reason": {}, "prompt_token_ids": {}, "completion_token_ids": {}, "prompt_logprobs": {},
```
Xunzhuo: **Preserve Dynamo's detailed finish reason**

Impact: Dynamo's current documented `extra_fields` set includes `detailed_finish_reason`, but this allowlist rejects that valid request value. The response contract and buffered/stream codecs also have no field for the returned reason, so the extension cannot be preserved even if the request passes elsewhere.

Required change: Add bounded request, response, and streaming handling for `detailed_finish_reason`, with same-format Chat Completions and Responses regressions.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -74,6 +75,43 @@ func (r *OpenAIRouter) runRequestPreRoutingStages(
...
```
Xunzhuo, line 117: **Enforce the Dynamo boundary across Looper fan-out**

Impact: This validates only the initially selected target. A Looper decision then encodes the neutral request with an empty envelope and dispatches every eligible `ModelRef`, silently dropping `nvext`, top-level `cache_salt`, and effective Dynamo routing headers; mixed non-Dynamo or cross-format candidates therefore bypass this boundary and run with different inputs.

Required change: Either reject Dynamo-extension requests before entering Looper, or preserve the validated envelope/headers while checking every candidate's backend and wire format; cover token-data, header-only, mixed-backend, and cross-format decisions.

Xunzhuo (commented):
Thanks for closing the effective shadow-state boundary. Two Dynamo compatibility paths still need fixes on this head.

src/semantic-router/pkg/llmprotocol/dynamo_nvext.go
cr-zhao: done

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
cr-zhao: Disabled Looper fan-out for Dynamo-extension requests. Since nvext and Dynamo routing headers may contain backend-specific routing semantics, distributing them across multiple backends would be unsafe. Thanks for catching.

Xunzhuo (commented):
Thank you for the sustained work on these Dynamo boundaries. The previous detailed-finish-reason and Looper blockers are fixed, but five current-head trust and lifecycle issues still need attention. The exact-head response-api E2E and PR Gate are also failing.

src/semantic-router/pkg/protocolcodec/dynamo_nvext_wire.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	return extension, nil
+}
+
+func decodeDynamoRequestNVExtWire(wire dynamoRequestNVExtWire) *llmprotocol.DynamoRequestNVExt {
+	extension := &llmprotocol.DynamoRequestNVExt{
+		GreedSampling: wire.GreedSampling, UseRawPrompt: wire.UseRawPrompt,
+		Annotations:       append([]string(nil), wire.Annotations...),
+		BackendInstanceID: wire.BackendInstanceID,
+		TokenData:         append([]uint32(nil), wire.TokenData...),
```
Xunzhuo, line 116: Dynamo treats a present `token_data` as the input to execute and skips prompt tokenization, while the router's routing, PII, jailbreak, RAG, and tool-policy stages still inspect the unrelated neutral text. A caller can therefore present benign text to the router but send different tokens to the model. This conversion also collapses an explicitly present empty array to absence, which changes Dynamo's behavior after any mutation-triggered encode. Please reject this field at the public boundary unless its provenance is trusted and the inspected input matches it, and preserve presence separately if the field remains supported; add mismatched-text/tokens and present-empty regressions.

src/semantic-router/pkg/llmprotocol/dynamo_nvext.go
```diff
@@ -0,0 +1,333 @@
...
+		bytes += len(field)
+	}
+	return bytes, nil
+}
+
+func validateDynamoMetadataUpload(upload *DynamoMetadataUpload, limits Limits) (int, error) {
+	if upload == nil {
+		return 0, nil
+	}
+	if strings.TrimSpace(upload.URL) == "" {
```
Xunzhuo, line 202: This permits any client-supplied upload destination after only length and non-empty checks. Dynamo opens this URL through its storage layer and can write to local, S3, GCS, or Azure destinations with worker credentials, so exposing it here gives an untrusted request a backend-side write primitive. Please disable `metadata_upload` at the public boundary or constrain it to an operator-configured destination allowlist, with rejection tests for client-chosen schemes/locations.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -74,6 +75,56 @@ func (r *OpenAIRouter) runRequestPreRoutingStages(
...
+			"Dynamo target-bound request state cannot be safely fanned out by Looper",
+			nil,
+		)
+		ctx.ImmediateProtocolError = protocolError
+		metrics.RecordRequestError(targetModel, "unsupported_dynamo_nvext_looper")
+		inflight.End(selectedModel, ctx.InflightToken)
+		ctx.InflightToken = 0
+		return r.createErrorResponse(http.StatusBadRequest, protocolError.Error())
+	}
+	if err := validateDynamoBackendPool(r.Config, targetModel, ctx, ctx.ProtocolEnvelope); err != nil {
```
Xunzhuo, line 117: This checks the initially selected model and the original request headers, but `prepareProviderDispatch` can subsequently capability-reroute to a different model and provider/decision mutations can add the only Dynamo routing header. There is no equivalent check after either change, so target-bound state can reach a non-Dynamo final dispatch. Please validate the finalized envelope and effective outbound headers against `dispatch.logicalModel` immediately before dispatch, with reroute and profile/decision-only header regressions.

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
```diff
@@ -102,6 +103,19 @@ func (d *shadowDispatcher) prepareShadowCall(job *shadowJob) (*preparedShadowCal
 	if err != nil {
 		return nil, shadowReasonBackendUnresolved, err
 	}
+	if job.hasDynamoExt {
```
Xunzhuo, line 106: A same-format Dynamo target is not enough to make the whole envelope portable. This path preserves target-bound `token_data`, worker/rank IDs, and the side-effecting `metadata_upload`: the shadow may use a different tokenizer or worker pool, and both calls can race writing the same upload object. Please sanitize explicitly portable fields for the shadow target or reject shadowing when target-bound/side-effecting nvext fields are present, and cover these cases.

src/semantic-router/pkg/extproc/processor_res_semantic_stream.go
```diff
@@ -85,17 +85,33 @@ func (buffers *semanticStreamBuffers) push(responseBody []byte, ctx *RequestCont
 	}
 	if ctx.ProtocolResponseStream != nil {
 		frames, events, diagnostics, err := ctx.ProtocolResponseStream.Push(responseBody)
-		buffers.translated = appendProtocolFrames(buffers.translated, frames)
 		observeProtocolStream(ctx, events, diagnostics)
+		var boundaryErr error
+		if err == nil {
+			boundaryErr = validateDynamoResponseEvents(ctx, events)
```
Xunzhuo, line 91: The events have already been passed to `observeProtocolStream` before this boundary check. On rejection, `StreamingAborted` is set, but finalization can still reconstruct the observed terminal state and write cache, memory, response-object, and replay records because those stores do not gate on the abort. A forbidden Dynamo event can thus be hidden from the current client yet persist into later replay. Please validate before observation, retain the boundary failure request-wide, and prevent every store after an aborted stream; add a terminal rejected-event regression.

src/semantic-router/pkg/protocolcodec/dynamo_nvext_wire.go
cr-zhao: Thanks, I addressed this by rejecting token_data at the public boundary. Since client-provided tokens may come from a different tokenizer/configuration, we can't reliably reconstruct and match them against messages

src/semantic-router/pkg/llmprotocol/dynamo_nvext.go
cr-zhao: Thanks, agreed. The official guidance also states, “Do not allow untrusted inference callers to set this URL.” I've therefore disabled client-supplied metadata_upload at the public boundary for now. We can consider supporting it later with an operator-configured destination allowlist.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
cr-zhao: Thanks, I've added an additional validation immediately before dispatch, using the finalized envelope, effective outbound headers, and final dispatch.logicalModel.

src/semantic-router/pkg/extproc/shadow_dispatch_call.go
cr-zhao: Most Dynamo extension fields depend on the target model, worker pool, or scheduling context, and few are clearly portable. I chose the conservative approach: requests carrying any Dynamo `nvext` or routing headers now skip shadow dispatch entirely, even for another Dynamo backend using the same wire format. This includes headers added by provider profiles or decision mutations. The primary request remains unaffected, and the skip is recorded as `dynamo_request_state`. Added regression coverage for these cases.

src/semantic-router/pkg/extproc/processor_res_semantic_stream.go
cr-zhao: Moved validation before `observeProtocolStream()` in both `Push()` and `Finalize()`, so rejected events never enter the accumulated response state. `StreamingAborted` remains set for the rest of the request, and finalization checks it before response processing and cache, memory, response-object, or replay writes. Cleanup and usage accounting still run. Added regressions for a rejected terminal event and for an earlier abort followed by valid completion.

src/semantic-router/pkg/extproc/processor_res_semantic_stream.go
```diff
@@ -85,17 +85,41 @@ func (buffers *semanticStreamBuffers) push(responseBody []byte, ctx *RequestCont
 	}
 	if ctx.ProtocolResponseStream != nil {
 		frames, events, diagnostics, err := ctx.ProtocolResponseStream.Push(responseBody)
-		buffers.translated = appendProtocolFrames(buffers.translated, frames)
-		observeProtocolStream(ctx, events, diagnostics)
+		var boundaryErr error
+		if err == nil {
```
Xunzhuo: [P2] Validate returned stream events even when decoding also fails

A transport read can contain a valid Dynamo extension frame followed by malformed JSON. StreamEngine.Push deliberately returns the accepted prefix frames/events together with the trailing error. In that case this err == nil guard skips the provider check, boundaryErr stays nil, and the frames are appended below; processingResponse then sends them in the aborted-stream BodyMutation. A non-Dynamo backend can therefore still forward nvext or request_id. Validate the returned events even on the error path before releasing their frames, and cover a coalesced extension plus malformed frame.

Xunzhuo (changes requested):
The request, shadow and aborted-storage fixes address the earlier cases. One response-boundary issue remains: a valid extension prefix followed by a decoding error skips provider validation and is still forwarded. Both prefix cases reproduce in the codec; 80 focused Go tests and 16 mock-provider tests pass. The complete router/Envoy path was inspected but not run. Please validate returned events before emitting frames on the error path and add that regression.

src/semantic-router/pkg/extproc/processor_res_semantic_stream.go
cr-zhao: Updated to run the provider/boundary validation regardless of the decoding error. Thanks.

cr-zhao:
The only independent failure in the latest CI run is the Dashboard image build. This is unrelated to the Dynamo changes in this PR. The CI/PR gate failures are downstream consequences; Router Contracts and the previously failing E2E shard now pass.

src/semantic-router/pkg/extproc/processor_res_semantic_stream.go
```diff
@@ -85,17 +85,35 @@ func (buffers *semanticStreamBuffers) push(responseBody []byte, ctx *RequestCont
...
+		boundaryErr := validateDynamoResponseEvents(ctx, events)
+		if err == nil && boundaryErr == nil {
+			observeProtocolStream(ctx, events, diagnostics)
+		} else {
+			ctx.ProtocolDiagnostics = append(ctx.ProtocolDiagnostics, diagnostics...)
+		}
+		if boundaryErr == nil {
+			buffers.translated = appendProtocolFrames(buffers.translated, frames)
+		} else {
+			err = boundaryErr
```
Xunzhuo: This puts the provider-boundary error only in this chunk's `semanticStreamBuffers`; the next response-body call creates a new buffer. With a non-Dynamo `nvext` chunk whose `finish_reason` is `stop`, followed by a separate `data: [DONE]`, the first body is empty but the next emits a normal finish frame and `[DONE]` even though `ctx.StreamingAborted` is true. Keep the rejection in request-scoped stream state (or poison the codec), and check the second `BodyMutation` in the regression test.

Xunzhuo (changes requested):
Thanks for the follow-up. One response-boundary gap remains: a non-Dynamo backend can send an `nvext` terminal chunk followed by a separate `[DONE]` chunk. The first chunk is suppressed, but a focused two-chunk check shows the next response still emits a normal finish frame and `[DONE]` while the request is marked aborted. Please carry the rejection across chunks and assert that the later body mutation cannot finish successfully. The exact-head Router Contracts and Dynamo E2E checks pass, but they do not cover this split response; the overall PR Gate is failing on the Dashboard image build.

Issue #3183, opened 2026-08-31
[Feature] Define a capability-driven plugin and backend codec pipeline for vLLM-Omni


Xunzhuo:
### Goal

Define and implement a capability-driven semantic-plugin and backend-codec pipeline for vLLM-Omni models so the Router can select and execute qualified text, image, audio, video, and omni tasks without restoring the removed `image_gen` route plugin or leaking model-server dialects into routing orchestration.

Parent Epic: #2358

Cross-WG dependency: #3030

Follow-up to #3163 and #3129. Related provider capability contract: #2358. Related protocol-neutral translation work: #3076.

### Proposed Workgroup

Data Plane & Networking

### Scope

- Define model/backend task and endpoint bindings for applicable Chat Completions, Responses, Images, Audio, Video, and realtime or duplex contracts instead of assuming that one model has one undifferentiated `api_format`.
- Record the execution boundary explicitly: ingress codec -> neutral request -> routing and semantic plugins -> capability/task negotiation -> provider adapter/backend codec -> backend -> neutral response/events -> response plugins -> egress codec.
- Keep official client/provider protocol semantics in `llmprotocol` and `protocolcodec`; keep vLLM-Omni extensions and endpoint-specific request shapes in bounded provider adapters.
- Define when a Responses hosted operation such as `image_generation` can translate losslessly to a qualified vLLM-Omni Chat or dedicated Images endpoint, and return a typed unsupported-capability error when it cannot.
- Preserve multimodal inputs, generated artifacts, usage, streaming progress, terminal state, cancellation, backend errors, retries, fallback, and Router Replay diagnostics.
- Define plugin applicability and ordering for non-text request/response content so semantic plugins neither erase media blocks nor run on unsupported task types.
- Land the implementation in vertical slices. Start with one maintained image-generation path, then split additional audio, video, and duplex paths into bounded child work as their contracts are accepted.
- Add deterministic codec/provider fixtures and at least one maintained end-to-end vLLM-Omni execution contract.

### Non-goals

- Restore the removed `image_gen` decision plugin or its direct backend client.
- Add provider- or model-specific branches to the main routing orchestrator.
- Promise support for every vLLM-Omni model, endpoint, modality combination, or hardware target.
- Implement or fork the vLLM-Omni runtime.
- Treat model-pool eligibility and recipe quality as Data Plane ownership; those remain with #3030 and MoM & Routing.

### Completion signal

[...]

PR #3301, opened 2026-09-01
[CI/Build] Unify Kind bootstrap for Operator and E2E

mikemikimike:
Closes #3290

## Purpose

Unify Operator and Kubernetes E2E Kind bootstrap under one composite action owned by the Evaluation & Quality Workgroup. The shared path reclaims runner disk, optionally relocates Docker for E2E, installs Kind v0.31.0 for both jobs, pre-pulls one digest-pinned kindest/node image, and exports that image for the cluster creation paths. Operator integration and E2E profile execution remain separate receipts.

## Test Plan

- `go test ./pkg/cluster -count=1` (from `e2e/`)
- `python -m unittest tools/ci/tests/test_kind_bootstrap.py tools/ci/tests/test_pr_change_classifier.py tools/ci/tests/test_workflow_permissions.py`
- `python tools/ci/validate_workflows.py`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 -shellcheck=`
- `ruff check tools/ci/tests/test_kind_bootstrap.py`
- `git diff --check`

## Test Result

All listed local checks pass. Full Kind/Operator and Kubernetes E2E execution was not run on this Windows host because Docker/Kubernetes CI infrastructure is unavailable locally; the changed E2E command path is covered by the focused Go tests and the workflow/action validators.

PR #3313, opened 2026-09-02
[Feature] Capability-driven image-generation vertical: reroute + DALL-E images codec

Bevisy:
## What

First vertical slice for #3183: a capability-driven image-generation path through the Router, without restoring the removed `image_gen` plugin. Hosted `image_generation` operations on Responses are dispatched to a qualified image model and sunk to a DALL-E dialect backend (vLLM-Omni), then normalized back to the Responses wire.

## Changes

- **codec**: new `protocolcodec` ImagesCodec mapping the DALL-E (`/v1/images/generations`) dialect to/from the neutral IR — `openai.images.v1` wire, `api_format: "images"`, encode sinks `image_generation` tool input to `{model,prompt,size,response_format:b64_json}`, decode renders `ContentGeneratedImage`; streaming entrypoints fail closed (no image streaming contract yet).
- **dispatch**: on capability mismatch the router re-routes the decision to a qualified sibling modelRef (`reroute dispatch`), then steers re-route candidates by their **declared model capabilities** (`declared capabilities ∩ required`).
- **typed failure**: a backend that cannot satisfy the required capabilities returns `400 unsupported_capability` (protocol error) instead of a 500.
- **test infra**: `mock-vllm` gains a responses-wire `image_generation` tool handler (12/12 pytest), plus deterministic codec provider fixtures.

## Verification

- Unit: new ImagesCodec tests (encode/decode round-trip, capability declaration, engine translate DALL-E → `image_generation_call`); full `protocolcodec`/`llmprotocol`/`config` green; extproc regression suite green (Linux VM, go build/vet/test).
- E2E (A100 cluster, real vLLM-Omni / Z-Image-Turbo behind Envoy+ExtProc): Responses request requiring `image_generation` → reroute to image model → real **1024×1024 PNG** generated (magic bytes verified) → normalized `image_generation_call` response (base64) → HTTP 200. Capability-mismatch cases return `unsupported_capability` 400.

## Out of scope (bounded child work)

- Capability-gated semantic plugins + multi-step composition (issue's plugin pillar).
- Conformance coverage for image streaming / cancellation / usage / Replay (codec is fail-closed for streaming).
- Audio/video/duplex dialects and public config guidance.

Related #3183 — first vertical slice; remaining slices tracked in the plan comment


FAUST-BENCHOU (commented):
The capability-driven routing direction looks promising. One detail worth clarifying is how capability declarations are validated against the actual backend implementation, so a model does not advertise support for a capability that the provider cannot execute.

FAUST-BENCHOU (dismissed):
LGTM. The capability-driven design keeps the image generation support isolated and the typed capability handling makes unsupported cases easier to reason about.

wilsonwu (dismissed):
LGTM, thanks!

Bevisy:
@FAUST-BENCHOU @wilsonwu — rebased onto latest main and pushed; the earlier force-push dismissed your reviews, so this is a re-request.

What changed since your LGTM:
- Rebased onto current main (caught up on #3301/#3321).
- One fix: the capability reroute rebase had dropped the `resolveImageFileReferences` call in `prepareProviderRequest`, silently disabling file-backed image inlining (file_id references were forwarded in place of file content). Restored it (commit `5ca1e8ebc`); the retained-history image test now guards it.

Verified locally (Linux VM): `go build`, `go vet`, and `pkg/extproc` / `pkg/protocolcodec` / `pkg/llmprotocol` tests all green. No other code changes.

wilsonwu:
@Bevisy please resolve conflicts, thanks!

Bevisy:
> @Bevisy please resolve conflicts, thanks!

Done. PTAL

Xunzhuo (dismissed):
Thanks for carrying this through the image vertical. The sink still drops the provider model, does not request `b64_json` although its decoder rejects URL responses, and silently loses neutral image options. Please make supported conversion lossless and fail closed for unsupported options. Also restrict `lastRequestText` to the intended user turn and add multi-message coverage.

Xunzhuo (changes requested):
Thanks—the earlier lossless-conversion issues are fixed. One blocker remains: ImagesCodec advertises generic tools support, so a Responses request that combines image_generation with an ordinary function tool is accepted and routed here, but EncodeRequest drops Tools and ToolChoice. That silently changes the request. Please either narrow the advertised capability or reject/encode mixed tools explicitly, with a regression test. Non-blocking: align CandidateCount with CapabilityMultipleCandidates so multi-image behavior is consistent.

Bevisy:
Both addressed.

For the blocker, we kept `CapabilityTools` advertised (the capability-driven selection requires it for tool-bearing requests) and instead fail closed in `EncodeRequest`: any ordinary function tool in `request.Tools`, or a tool choice mode of `auto`/`required`/`named`, is now rejected with `unsupportedDownstreamTranslation` rather than silently dropped. The pure `image_generation` tool choice still encodes normally. A regression test covers the mixed-tools and named-tool-choice rejection paths plus the accepted hosted-tool path.

For the non-blocking point, `ImagesCodec` now advertises `CapabilityMultipleCandidates`, matching the existing `CandidateCount` <-> `n` mapping on both encode and decode, so multi-image requests can select this codec consistently.

Xunzhuo (changes requested):
Thanks—the mixed-tool loss is fixed, but one fail-closed blocker remains. A Responses request that declares `image_generation` with `tool_choice: "none"` still requires image generation; `clearSemanticToolChoiceWhenNoTools` erases that explicit choice, and `rejectUnsupportedTools` also accepts `ToolChoiceNone`, so the request can be rerouted to `/v1/images/generations` and generate an image despite the caller forbidding all tools. Please preserve the explicit no-tool choice and reject or avoid the Images dispatch, with an end-to-end regression proving no image backend call occurs.

Bevisy:
> Thanks—the mixed-tool loss is fixed, but one fail-closed blocker remains. A Responses request that declares `image_generation` with `tool_choice: "none"` still requires image generation; `clearSemanticToolChoiceWhenNoTools` erases that explicit choice, and `rejectUnsupportedTools` also accepts `ToolChoiceNone`, so the request can be rerouted to `/v1/images/generations` and generate an image despite the caller forbidding all tools. Please preserve the explicit no-tool choice and reject or avoid the Images dispatch, with an end-to-end regression proving no image backend call occurs.

Addressed on both fronts.

`clearSemanticToolChoiceWhenNoTools` now preserves an explicit `tool_choice: "none"` whenever a hosted image_generation operation is declared (hosted tools don't occupy `request.Tools`, which is why the choice was previously erased). We took the "avoid" path for dispatch: `RequiredCapabilities` no longer demands the images capability when a declared `ImageGeneration` is paired with `tool_choice: "none"`, so the request stays on the text wire and the images backend is never selected. As a backstop, `ImagesCodec` also rejects `ToolChoiceNone` at encode time.

Regression coverage: an end-to-end dispatch test proves a `none`-pairing request with an images sibling in the decision is NOT rerouted to it — `logicalModel` stays on the primary chat wire, target format is not images, and the no-tool choice is preserved. Unit tests cover the capability requirement shift, the `clearSemanticToolChoiceWhenNoTools` preservation, and the codec rejection.

Xunzhuo (commented):
Thanks—the explicit no-tool choice now stays intact, avoids image-capability rerouting, and is rejected by the Images codec as a backstop. That resolves my blocker. Please sync current main and let the required checks return green before re-requesting review.

Bevisy:
> Thanks—the explicit no-tool choice now stays intact, avoids image-capability rerouting, and is rejected by the Images codec as a backstop. That resolves my blocker. Please sync current main and let the required checks return green before re-requesting review.

Synced current main (merged up to `ffe2bb64`, including #3512) and fixed the two CI failures the merge surfaced:

1. `chore: tidy dashboard backend go.sum for router dependency hashes` — the dashboard merge gate (`dashboard-go-mod-tidy`) failed because `dashboard/backend/go.sum` was missing transitive hashes for the router module (dashboard replaces `src/semantic-router`). Re-ran `go mod tidy` in `dashboard/backend` only; no version changes.
2. `fix: carry fusion depth on request context instead of mutable client field` — `TestFusionLooperPanelQuorumSkipsSlowWorker` failed intermittently under CI scheduling. The `x-vsr-fusion-depth` header came from a mutable `Client` field set at `Execute` start and reset via `defer`; a slow panel worker whose request was dispatched after `Execute` returned raced the reset and sent the header empty. The depth now travels on the request context (same pattern as the existing routing-recipe context), so every request carries its own depth regardless of send timing.

The required checks are green on the new head (`6ab2ba5f`). Ready for re-review.

Xunzhuo (changes requested):
Thanks—the explicit `tool_choice: none` path is fixed, but omitted image-tool choice is still defaulted to `auto` and then rejected by the Images codec, while capability rerouting leaves request preparation and all selected-model accounting on the original backend. Please carry the final dispatch model through the full request state and add an ingress-to-dispatch regression for the default choice.

Bevisy:
> Thanks—the explicit `tool_choice: none` path is fixed, but omitted image-tool choice is still defaulted to `auto` and then rejected by the Images codec, while capability rerouting leaves request preparation and all selected-model accounting on the original backend. Please carry the final dispatch model through the full request state and add an ingress-to-dispatch regression for the default choice.

Addressed both points on `feefdb3c`:

1. **Defaulted `auto` now encodes on the Images wire.** `applyRequestSemanticDefaults` fills `ToolChoiceAuto` when a pure image request omits `tool_choice`, and `rejectUnsupportedTools` was rejecting it. `auto` (with no named function tool) is now accepted — it can only invoke the hosted `image_generation` operation. A named function choice still fails closed.

2. **The final dispatch model now flows through the full request state.** After a capability reroute, the reasoning mode is re-applied against the rerouted model (system prompt and request params are decision-scoped and were already applied, so they are deliberately not re-applied), the dispatch log records the rerouted model, and `ctx.SemanticRequest`/`request.Model`/`ctx.TargetFormat` all reflect the final dispatch.

Regressions added: pure `auto` encodes on the images codec, and an omitted-`tool_choice` image request reroutes to the images sibling with the final model asserted in request and context state.

Xunzhuo (approved):

PR #3314, opened 2026-09-02
[Bug] Surface backend capability mismatch as a typed protocol error

Bevisy:
## What

When a request requires capabilities no backend can satisfy, the Router returns a typed **`400 unsupported_capability`** protocol error instead of an opaque 500. Adds the routing-layer capability query seam (`Registry.CapabilitiesFor(format)`).

Part of #3183 (capability-driven pipeline). This is the standalone bug-fix half; the capability-driven reroute/codec vertical lives in #3313.

## Changes

- `protocolcodec` registry: expose per-format capability sets (`CapabilitiesFor`).
- extproc request-body routing: pre-check dispatch against backend capabilities → `ProtocolError` with `unsupported_capability` (client-visible 400).
- Unit tests for the capability registry and the rejection path.

Related #3183


FAUST-BENCHOU (commented):
The typed protocol error is a good improvement over surfacing backend capability mismatches as generic failures. It may be worth ensuring the capability registry remains the single source of truth as more codecs/backends add capabilities, otherwise validation behavior could diverge again.

FAUST-BENCHOU (approved):
LGTM. Introducing a typed capability mismatch error improves API clarity compared with returning a generic server error.

wilsonwu (approved):
LGTM, thanks!

Issue #2668, opened 2026-07-25
[Bug] Honor declared embedding backends instead of silently using Candle


shraderdm:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

A configuration that declares the OpenVINO embedding backend can be accepted while the runtime silently executes Candle, making configuration and observed behavior disagree.

### To Reproduce

Use the reproduction or current-code evidence documented in this issue to exercise the affected path.

### Expected behavior

The resolved binding should use the declared qualified backend or reject the configuration as unsupported before activation.

### Impact

Not separately specified in the original report.

##### Completion signal

Resolution and runtime tests prove that backend identity is honored and visible in diagnostics.

Parent Epic: #2396

shraderdm:
Additional detail from tracing the image build, which makes the gap more concrete than the original report.

The shipped image advertises this backend but never builds a binary for it. In `tools/docker/Dockerfile.extproc`, the only router binaries produced are `router-candle` (lines 440 and 445, no build tags) and `router-onnx` (line 470, `-tags=onnx`). There is no `go build` for `router-openvino` anywhere in `tools/docker/`, `tools/make/`, or `.github/workflows/`.

The image nonetheless stages the OpenVINO shared library into `/app/lib` (line 459) and documents the backend as selectable: line 4 says "set AI_BINDING=candle (default), onnx, or openvino to choose the inference backend", and line 500 repeats it above `ENV AI_BINDING=candle`.

`scripts/entrypoint.sh` then maps `AI_BINDING=openvino` to `BINARY=/app/router-openvino`, finds no such file, prints "Binary not found" plus "Falling back to candle binding...", and runs candle instead. So an operator who follows the documented instruction gets candle, and the only signal is a stderr line at startup.

This is the runtime half of the same declaration-versus-reality gap: `initOpenVINOModel` sits behind the `openvino` build tag and no build target sets it, so a config declaring `backend: openvino` is a contract about a backend the shipped image cannot run.

Worth noting for scope: `tools/make/openvino.mk` and `.github/workflows/openvino-binding-ci.yml` do build and test the OpenVINO binding itself. It is specifically the tagged router binary that is never produced.

nanasis:
@theohsiung @Xunzhuo please assign this to me, I've done some investigations and would like to help on this area

theohsiung:
sure thx

[...]

PR #3316, opened 2026-09-02
[BugFix] Ship the OpenVINO router backend in image build

nanasis:
Related to #2668

## Purpose

Ship the OpenVINO router path already advertised by the amd64 extproc image instead of leaving `AI_BINDING=openvino` without a corresponding binary.

- build `router-openvino` with the real `openvino` CGO implementation
- package the minimal OpenVINO CPU runtime, tokenizer extension, IR frontend, and wrapper SONAME links
- fail the image build when the OpenVINO binary or dynamic dependencies are incomplete
- keep the existing missing-binary fallback while explicitly reporting that the effective backend changed to Candle
- prevent configurations declaring `backend: openvino` from entering the shared Candle-only embedding initialization path

OpenVINO remains unsupported on arm64. Shared tools/cache consumers remain unready rather than being falsely advertised as OpenVINO-capable; routing embedding classifiers use the existing OpenVINO runtime path.

Owned by the Router Models & Inference Runtime Workgroup.

## Test Plan

- `make agent-validate`
- `make agent-lint`
- `make test-semantic-router`
- `make agent-dev ENV=cpu`
- `make agent-serve-local ENV=cpu`
- `make agent-smoke-local ENV=cpu`
- `make agent-ci-gate CHANGED_FILES="tools/docker/Dockerfile.extproc tools/docker/entrypoint.sh src/semantic-router/pkg/modelruntime/router_runtime.go src/semantic-router/pkg/modelruntime/router_runtime_test.go"`
- build the amd64 extproc image and inspect `ldd` for `router-openvino` and `libopenvino_semantic_router.so`
- start the image with an `all-MiniLM-L6-v2` OpenVINO fixture and preload one embedding candidate

## Test Result

- all repository validation commands above passed
- final image contains an executable `router-openvino`; Docker build-time dependency assertions passed
- `ldd` reported no unresolved dependencies for the OpenVINO binary or wrapper
- real CPU initialization loaded the OpenVINO tokenizer extension, CPU plugin, model, and infer-request pool
- candidate preload returned a 384-dimensional embedding through `backend=openvino`
- clean OpenVINO smoke performed no Candle embedding initialization, no fallback, and no model download
- canonical Candle CPU build/serve/smoke remained healthy

Remaining boundary: OpenVINO is packaged only for amd64, matching the existing project support boundary.

---
<details>

- [x] The PR links accepted issue #2668 with exactly one Workgroup owner
- [x] Purpose, Test Plan, and Test Result reflect the actual scope and results

</details>

### DIFF
--- a/src/semantic-router/pkg/modelruntime/router_runtime.go
+++ b/src/semantic-router/pkg/modelruntime/router_runtime.go
@@ -182,6 +182,11 @@ func embeddingRuntimeTasks(
 		logEmbeddingRuntimeStart(component, cfg, paths)
 		return tracker.snapshot(), remoteEmbeddingRuntimeTask(cfg, component, tracker), tracker
 	}
+	// OpenVINO model initialization belongs to the classifier runtime. Do not
+	// send its model paths through the Candle-only shared runtime.
+	if cfg.EmbeddingModels.EmbeddingBackend() == config.EmbeddingBackendOpenVINO {
+		return EmbeddingRuntimeState{}, nil, nil
+	}
 	if !paths.hasConfiguredModels() {
 		return EmbeddingRuntimeState{}, nil, nil
 	}
--- a/src/semantic-router/pkg/modelruntime/router_runtime_test.go
+++ b/src/semantic-router/pkg/modelruntime/router_runtime_test.go
@@ -71,6 +71,21 @@ func TestEmbeddingRuntimeTasksUseOnlyRemoteProviderWhenConfigured(t *testing.T) 
 	}
 }
 
+func TestEmbeddingRuntimeTasksDoNotUseCandleForOpenVINO(t *testing.T) {
+	cfg := &config.RouterConfig{InlineModels: config.InlineModels{EmbeddingModels: config.EmbeddingModels{
+		Qwen3ModelPath: "models/openvino-embedding",
+		EmbeddingConfig: config.HNSWConfig{
+			Backend:   config.EmbeddingBackendOpenVINO,
+			ModelType: config.EmbeddingModelTypeQwen3,
+		},
+	}}}
+
+	_, tasks, _ := embeddingRuntimeTasks(cfg, "test", resolveEmbeddingPaths(cfg))
+	if len(tasks) != 0 {
+		t.Fatalf("embeddingRuntimeTasks() returned %d task(s), want no Candle tasks for OpenVINO", len(tasks))
+	}
+}
+
 func TestPrepareRouterRuntimeProbesRemoteEmbeddingProvider(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
 		writeRuntimeEmbeddingResponse(t, w, []float64{0.1, 0.2})
--- a/tools/docker/Dockerfile.extproc
+++ b/tools/docker/Dockerfile.extproc
@@ -310,9 +310,25 @@ RUN if [ "$TARGETARCH" = "arm64" ]; then \
         ls -la libopenvino_semantic_router.so; \
     fi
 
+# Collect the CPU runtime needed by the tagged router binary. Keep the
+# directory present on arm64 so later cross-stage copies remain valid while
+# OpenVINO stays unsupported there.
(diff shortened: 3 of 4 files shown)


FAUST-BENCHOU (commented):
The OpenVINO packaging path looks well covered, especially the runtime dependency checks and fallback behavior. One thing to keep in mind is keeping the advertised backend capabilities aligned with actual image contents, so future backend additions do not expose a capability before the corresponding runtime assets are shipped.

nanasis:
CI follow-up: the initial E2E failure occurred before tests because the upstream `quay.io/centos/centos:stream10` tag had moved from the runnable container image to a bootc OCI artifact without `/bin/sh`. The extproc final stage now uses `stream10-development`, which retains the CentOS Stream 10 shell and dnf environment required by the Dockerfile. I reproduced the failure with a fresh pull, rebuilt the complete image against the replacement tag, reran the OpenVINO CPU preload smoke, and the subsequent PR run passed E2E, image, core, quality, security, DCO, and all compatibility gates.

FAUST-BENCHOU:
has been fixed , pls rebase

nanasis:
> has been fixed , pls rebase

I've rebased with the latest master, please help review

WUKUNTAI-0211 (commented):
Thanks for making the advertised backend real — the build-time assertion on `router-openvino` and its deps is what closes the gap between `AI_BINDING=openvino` and what ships, and not sending an `openvino` config through the Candle-only path is the substantive half of #2668.

Correcting the weight of my original note: this is a nit, not a blocker. The early return at `src/semantic-router/pkg/modelruntime/router_runtime.go:185-189` sits above `logEmbeddingRuntimeStart` (`:621`), so an `openvino` config no longer emits `embedding_models_init_started` at all. Nothing in tree consumes that event — no dashboard, e2e, or test reads it — so it costs an operator one log line and breaks no consumer. Restoring it is a one-liner if you want it, and it should not hold this up.

`tools/docker/entrypoint.sh:28` still activates candle when the declared binary is missing, which is the arm64 path now that amd64 is build-asserted. Whether that leaves the diagnostics half of #2668's completion signal satisfied is the issue owner's call rather than a merge gate.

Xunzhuo (approved):
Thanks for closing the OpenVINO image gap, and welcome! The packaging contract, runtime wiring, and focused tests are coherent. LGTM from this pass.

nanasis:
> Thanks for making the advertised backend real — the build-time assertion on `router-openvino` and its deps is what closes the gap between `AI_BINDING=openvino` and what ships. One blocker: the early return at `src/semantic-router/pkg/modelruntime/router_runtime.go:185-189` sits above `logEmbeddingRuntimeStart`, the only emitter of `"backend": EmbeddingBackend()` (`:624`), so an `openvino` config that used to log `embedding_models_init_started` with `backend=openvino` now logs nothing. This drops the one diagnostic covering the configuration #2668 is about, whose completion signal is that backend identity stays visible. The new test asserts only `len(tasks) == 0`, so it passes either way; logging before the return covers both. Also, `tools/docker/entrypoint.sh:28` still activates candle when the declared binary is missing — arm64, now that amd64 is build-asserted — so if that fallback stays by design, this reads as `Part of #2668` rather than `Closes`.

Sure, we can merge this PR as a part of #2668 and discuss more about the current by-design issue later

WUKUNTAI-0211 (approved):
Thanks for the OpenVINO packaging work. I rechecked head `63ca2eee`: `go test ./pkg/modelruntime/ ./pkg/config/` passes locally, including the new `TestEmbeddingRuntimeTasksDoNotUseCandleForOpenVINO`; `entrypoint.sh` parses clean; and the `libopenvino_ir_frontend.so.2510` assertion matches the `openvino==2025.1.0` pin at `Dockerfile.extproc:291`, so the final-stage check is self-consistent.

CI is not green yet, and neither reason is this PR: the branch needs a rebase, and `E2E / integration-test (envoy-ai-gateway)` is failing repo-wide because the `quay.io/centos/centos:stream10` base image no longer has `/bin/sh` (identical failure signature on #3268). Ready from my side once it is rebased.

PR #3319, opened 2026-09-02
[Bug] Align ONNX and Candle image preprocessing

### LINKED ISSUE #2166: [Bug] Align ONNX and Candle image preprocessing

karthikeyan1592:
Closes #2166

## Purpose

onnx-binding resized images via nearest-neighbor sampling; candle-binding
used a Catmull-Rom cubic filter — different algorithms on the same input
meant different embeddings for the same routing decision. See the doc
comment on `decode_resize_to_chw_f32` in `onnx-binding/src/ffi/multimodal.rs`
for the exact mechanism. Owned by `wg/router-models-inference-runtime`.

First attempt: ported the resize onto `golang.org/x/image/draw`'s own
CatmullRom scaler, staying entirely in Go. Measured it against
candle-binding's real output on a real fixture before trusting it:

    cosine similarity:   0.999904
    mean abs diff:       0.003066
    median abs diff:     0.003922
    max abs diff:        0.058824
    pixels diverged:     55.34% (diff >= 1e-6)

Two different libraries both calling their kernel "CatmullRom" is not the
same as one library — 55% of pixels differing, even at a high cosine score,
is a real divergence, not floating-point noise.

Reused candle-binding's actual implementation instead: moved the same
decode+resize into onnx-binding's own Rust FFI layer (`image = "0.25"`,
same version/features as candle-binding), so both bindings run the
identical code path. Verified bit-exact (`max_diff = 0.0`) against
candle-binding's real output before finalizing.

**Known tradeoffs** (filed as
[TD056](tools/agent/docs/tech-debt/td-056-onnx-candle-image-resize-duplication.md)):
- `decode_resize_to_chw_f32` now exists as two copies (candle-binding,
  onnx-binding), kept in sync only by a code comment and matching `image`
  crate version pins. Nothing enforces they stay identical beyond a human
  noticing.
- Every call through this path crosses from Go into Rust (goroutine parked,
  resumed on return) on every request that hits this code. The crossing
  cost is a fixed, tiny toll next to the millisecond-scale decode+resize
  work it wraps — but it's a real, continuous cost, not a free abstraction.

**Note on CI coverage:** onnx-binding is not currently built or tested in
the core CI receipt (a pre-existing gap, documented as TD046). That means
the tests added here are currently verified locally only, not by CI. Filed
#3320 to track closing that gap.

## Test Plan

    cd onnx-binding
    cargo check --locked --lib
    cargo test --lib ffi::multimodal
    cargo fmt --check
    gofmt -l semantic-router.go
    go build ./...

## Test Result

All green, run locally (not currently exercised by CI — see the note
above):
- `cargo check --locked --lib`: clean
- `cargo test --lib ffi::multimodal`: 5 passed (2 invalid-input tests
  ported from candle-binding's existing coverage; 3 golden-image parity
  tests against real fixtures, declared 1e-4 tolerance)
- `cargo fmt --check`: clean
- `gofmt -l`: clean
- `go build ./...`: clean

Remaining risk: no automated enforcement yet that the two
`decode_resize_to_chw_f32` copies stay in sync (TD056), and no CI coverage
for onnx-binding generally (TD046, tracked as #3320).

---
<details>

</details>


karthikeyan1592:
Status update on CI for this PR:

**Core quality / Full pre-commit checks: now passing.** Fixed three issues surfaced by touching `onnx-binding/src/ffi/multimodal.rs` for the first time in CI:
- markdownlint (MD018) on the new TD056 tech-debt file
- pre-existing clippy findings on 4 functions in this file that were never linted before (`not_unsafe_ptr_arg_deref`, `cast_slice_from_raw_parts`) — fixed to match candle-binding's existing conventions for the same lints
- TD056 registration in `tools/agent/repo-manifest.yaml` (both the flat docs list and the `doc_governance` stewardship section), required by `agent-validate`

**The remaining red check (E2E / integration-test envoy-ai-gateway) is unrelated to this PR.** Root cause: the final stage of `tools/docker/Dockerfile.extproc` builds `FROM quay.io/centos/centos:stream10`, and that image tag is currently broken on Quay's registry — `/bin/sh` (and even `ls`) don't exist inside it. Reproduced independently of any code in this PR:

```
$ docker pull quay.io/centos/centos:stream10
Status: Downloaded newer image for quay.io/centos/centos:stream10

$ docker run --rm quay.io/centos/centos:stream10 /bin/sh -c "echo hello"
docker: Error response from daemon: failed to create task for container: failed to
create shim task: OCI runtime create failed: runc create failed: unable to start
container process: error during container init: exec: "/bin/sh": stat /bin/sh: no
such file or directory
```

Architecture matches (amd64/amd64, no arch mismatch). This Dockerfile is untouched by this PR's diff and would fail identically on any PR right now, including a no-op PR against `main`. Not something fixable from this branch — needs either Quay to fix the published image, or a re-pin to a different tag/digest in `tools/docker/Dockerfile.extproc`.

Xunzhuo (approved):
Thanks for aligning the ONNX resize path, and welcome! The implementation matches the Candle contract and the golden and invalid-input coverage is well scoped. LGTM.

PR #3321, opened 2026-09-02
[Bug] Honor configured llm_timeout_seconds in VLLMClient HTTP timeout

### LINKED ISSUE #3315: [Bug] VLLMClient ignores configured llm_timeout_seconds, HTTP timeout hardcoded at 30s
### Proposed Workgroup

### Describe the bug

`VLLMClient` (`pkg/classification/vllm_client.go:51`) constructs its `http.Client` with a hardcoded `Timeout: 30 * time.Second`. The configured `llm_timeout_seconds` on `ExternalModelConfig` is never passed to the client, so the HTTP round-trip timeout is always 30s regardless of configuration.

All three remote classifier bindings route through `newVLLMClientFromConfig` → `newVLLMClient` and inherit this hardcoded client timeout:

- `VLLMJailbreakInference` (http_chat guardrail, `vllm_classifier.go`)
- `llmLabelClassifier` (generic LLM label classifier, `generic_classifier.go`)
- `PreferenceClassifier.classifyExternal` (external preference router, `preference_classifier_external.go`)

Each caller separately computes a `context.WithTimeout(ctx, timeout)` from `llm_timeout_seconds`, so two independent deadline chains run in parallel. Go applies whichever fires first:

- `llm_timeout_seconds > 30`: the client's 30s fires first, the configured timeout is silently ignored.
- `llm_timeout_seconds < 30`: the request context fires first, which happens to mask the bug.
- `llm_timeout_seconds = 30`: both fire together, no visible symptom.

The sibling `HTTPClassifierInference` (http_classify, `http_classifier.go:104`) constructs its own `&http.Client{Timeout: timeout}` from config and does not have this bug — only the `VLLMClient` path is affected.

A secondary issue: the three callers hardcode different default timeouts when `llm_timeout_seconds` is unset, so the same remote endpoint sees different behavior depending on which classifier consumes it:

| caller | default |
|---|---|
| `vllm_classifier.go:54` | 30s |
| `preference_classifier_external.go:50` | 30s |
| `generic_classifier.go:45` | 5s |

### To Reproduce

1. Configure any remote classifier binding with `llm_timeout_seconds: 60` against a backend whose inference latency exceeds 30s.
2. Send a classification request that triggers the slow path.
3. Observe the request is terminated at ~30s with `context deadline exceeded` (wrapped as `HTTP request failed`), not at the configured 60s.

The masking case (`llm_timeout_seconds < 30`) can be reproduced in isolation against an `httptest.Server` whose handler sleeps longer than the configured timeout: pass `context.Background()` (no caller deadline) and confirm only `http.Client.Timeout` can interrupt. The pre-fix client waits the full hardcoded 30s instead of the configured value.

### Expected behavior

`llm_timeout_seconds` bounds the full HTTP round-trip (including response body read) for every remote classifier binding, and the default is resolved once rather than hardcoded separately in each caller.

### Impact

- Deployments that raise `llm_timeout_seconds` above 30s to tolerate slow generative guardrail models get no benefit; requests are cut at 30s and may surface as classifier-unavailable, triggering `on_error` policy (e.g. `block` mis-flags traffic, `allow` silently drops guardrail coverage).
- The three callers' inconsistent defaults make remote-classifier timeout behavior depend on which binding consumed the config, not on a single documented default.
- This is a precondition gap for the Epic [#2782](https://github.com/vllm-project/semantic-router/issues/2782) direction: [#3196](https://github.com/vllm-project/semantic-router/issues/3196) asks each qualified binding to validate deadline behavior, but the current deadline behavior is config-not-honored plus divergent defaults.

### Additional context

Root cause is the shared `newVLLMClient` constructor, so the fix belongs there rather than per-caller guards. The repo already has the `GetMaxRequestBytes` / `GetMaxResponseBytes` default-resolver pattern on `ExternalModelConfig` (`pkg/config/model_config_limits.go`); adding `GetTimeout()` alongside and threading it through `newVLLMClient` resolves both the hardcoded client timeout and the divergent caller defaults in one source of truth.

Bevisy:
Closes #3315

## Purpose

`VLLMClient` (`pkg/classification/vllm_client.go`) constructed its `http.Client` with a hardcoded `Timeout: 30 * time.Second`, ignoring the configured `llm_timeout_seconds`. All three remote classifier bindings that route through `newVLLMClientFromConfig` — jailbreak (`vllm_classifier.go`), generic LLM label (`generic_classifier.go`), and external preference (`preference_classifier_external.go`) — inherited this hardcoded client timeout, so a configured `llm_timeout_seconds` above 30s was silently capped at 30s by the client. Below 30s the caller's own `context.WithTimeout` fired first, masking the bug.

The sibling `HTTPClassifierInference` (`http_classifier.go`) already builds its client timeout from config and was unaffected — only the `VLLMClient` path was broken.

**Scope of this fix:** add `ExternalModelConfig.GetTimeout()` (matching the existing `GetMaxRequestBytes`/`GetMaxResponseBytes` pattern) and thread it through `newVLLMClient`, so the client timeout honors `llm_timeout_seconds` for every caller. The vLLM jailbreak and external preference callers also adopt `GetTimeout()` for their caller-side `context.WithTimeout`, so both deadline chains agree.

The generic LLM label caller is **not** changed: it keeps its existing 5s caller-side default, but still gets the config-driven client timeout via `newVLLMClientFromConfig`, so a configured `llm_timeout_seconds` is honored on the client chain regardless. Converging the generic 5s caller default to the shared 30s is a user-visible behavior change and is left for a separate change.

Affected modules: `pkg/config` (new `GetTimeout`), `pkg/classification` (client + two callers). Owned by `wg/router-models-inference-runtime` (#3315).

## Test Plan

- `go test ./pkg/classification/` — covers the callers plus the new `vllm_client_timeout_test.go`
- `go test ./pkg/config/` — covers `GetTimeout`
- `gofmt -l` + `go vet ./pkg/classification/ ./pkg/config/`

The new test `TestVLLMClientTimeoutHonorsConfig` isolates the client timeout by passing `context.Background()` (no caller deadline) against an `httptest.Server` whose handler sleeps 2s; with `llm_timeout_seconds: 1` the request must terminate near 1s, not the pre-fix 30s. `TestVLLMClientTimeoutDefaultsWhenUnset` checks the 30s default still applies when unset. `TestGetTimeoutCoversCallers` checks the three resolution states (unset / configured / negative).

## Test Result

```
=== classification ===
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/classification	4.717s
=== config ===
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/config	0.322s
```

`gofmt` + `go vet` clean. No remaining risk: the fix only changes how the client timeout value is resolved (config → client), not the deadline semantics. The generic caller's 5s default is untouched, so deployments using generic LLM label classifiers without an explicit `llm_timeout_seconds` see no behavior change. Configs that set `llm_timeout_seconds` now get that value on the client chain for all three callers; configs above 30s are no longer silently capped.

---

<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/classification/preference_classifier_external.go
+++ b/src/semantic-router/pkg/classification/preference_classifier_external.go
@@ -47,10 +47,7 @@ func newExternalPreferenceClassifier(
 
 	client := newVLLMClientFromConfig(externalCfg)
 
-	timeout := 30 * time.Second
-	if externalCfg.TimeoutSeconds > 0 {
-		timeout = time.Duration(externalCfg.TimeoutSeconds) * time.Second
-	}
+	timeout := externalCfg.GetTimeout()
 
 	return &PreferenceClassifier{
 		client:             client,
--- a/src/semantic-router/pkg/classification/vllm_classifier.go
+++ b/src/semantic-router/pkg/classification/vllm_classifier.go
@@ -51,10 +51,7 @@ func NewVLLMJailbreakInference(cfg *config.ExternalModelConfig, defaultThreshold
 	client := newVLLMClientFromConfig(cfg)
 
 	// Use timeout from config, default to 30 seconds
-	timeout := 30 * time.Second
-	if cfg.TimeoutSeconds > 0 {
-		timeout = time.Duration(cfg.TimeoutSeconds) * time.Second
-	}
+	timeout := cfg.GetTimeout()
 
 	// Use threshold from config, fallback to default
 	threshold := defaultThreshold
--- a/src/semantic-router/pkg/classification/vllm_client.go
+++ b/src/semantic-router/pkg/classification/vllm_client.go
@@ -27,19 +27,19 @@ type VLLMClient struct {
 
 // NewVLLMClient creates a new vLLM REST API client for classifiers
 func NewVLLMClient(endpoint *config.ClassifierVLLMEndpoint) *VLLMClient {
-	return newVLLMClient(endpoint, "", config.DefaultClassifyMaxResponseBytes)
+	return newVLLMClient(endpoint, "", config.DefaultClassifyMaxResponseBytes, time.Duration(config.DefaultClassifierTimeoutSeconds)*time.Second)
 }
 
 // NewVLLMClientWithAuth creates a new vLLM REST API client with access key
 func NewVLLMClientWithAuth(endpoint *config.ClassifierVLLMEndpoint, accessKey string) *VLLMClient {
-	return newVLLMClient(endpoint, accessKey, config.DefaultClassifyMaxResponseBytes)
+	return newVLLMClient(endpoint, accessKey, config.DefaultClassifyMaxResponseBytes, time.Duration(config.DefaultClassifierTimeoutSeconds)*time.Second)
 }
 
 func newVLLMClientFromConfig(cfg *config.ExternalModelConfig) *VLLMClient {
-	return newVLLMClient(&cfg.ModelEndpoint, cfg.AccessKey, cfg.GetMaxResponseBytes())
+	return newVLLMClient(&cfg.ModelEndpoint, cfg.AccessKey, cfg.GetMaxResponseBytes(), cfg.GetTimeout())
 }
 
-func newVLLMClient(endpoint *config.ClassifierVLLMEndpoint, accessKey string, maxResponseBytes int64) *VLLMClient {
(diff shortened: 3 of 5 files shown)


Xunzhuo (approved):
The configured timeout now reaches the HTTP client as well as the caller contexts, with the 30-second compatibility default and a focused round-trip regression. LGTM.

wilsonwu (approved):
LGTM

AayushSaini101 (approved):

Issue #3292, opened 2026-09-01
[Bug] Make Envoy base URL path rewrites idempotent for /v1-prefixed targets


Asthenia0412:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Envoy base URL path rewrites are not idempotent when the configured backend path itself begins with the `/v1` segment. The current generated `regex_rewrite` can run again after route recomputation and prepend the backend path twice.

For example, a backend with `base_url: https://api.example.com/v1/provider` can transform `/v1/chat/completions` into `/v1/provider/chat/completions` on the first pass, then into `/v1/provider/v1/provider/chat/completions` on a subsequent pass. The resulting upstream request is incorrect even though model selection succeeds.

The repository already contains focused unit and CLI integration tests documenting this remaining gap, but both are skipped as TODO(issue-2885). The merged fix for #2885 only prevents the `/v1beta` prefix case from matching again; it does not guarantee idempotency when the output still begins with `/v1`.

### To Reproduce

1. Configure an OpenAI-compatible backend whose `base_url` path is `/v1/provider`.
2. Generate the Envoy configuration with `vllm-sr serve`.
3. Trigger route recomputation by allowing the ext_proc filter to select the model.
4. Inspect the mock upstream request path.

Expected upstream path: `/v1/provider/chat/completions`.

Observed path: `/v1/provider/v1/provider/chat/completions`.

### Expected behavior

Applying the generated rewrite more than once must be a no-op after the first application. The upstream path must be derived exactly once from the immutable ingress path and the selected backend base path.

### Impact

Not separately specified in the original report.

##### Proposed Workgroup

Data Plane & Networking

##### Scope and non-goals

- Preserve existing behavior for `/v1beta`, `/openai/v1`, `/api/v1`, and paths without a base prefix.
- Keep the internal original-path carrier out of the upstream request.
- Do not change model selection, authentication, or unrelated Envoy routing behavior.
- Do not rely on enumerating a fixed set of endpoint names.

##### Validation

- Unit tests for ordinary prefixes, `/v1`-prefixed targets, query strings, empty paths, and repeated application stability.
- Restore the skipped Python config-generation regression test.
- Restore the skipped CLI integration E2E test using the mock upstream path recorder.
- Run affected CLI/Envoy tests, repository agent gates, and the relevant regression suite.

Asthenia0412:
[...]

PR #3324, opened 2026-09-02
[Bug] Collapse repeated version segment when resolving backend chat path

### LINKED ISSUE #3065: [Bug] `backend_refs.base_url` uses inconsistent API-root conventions for OpenAI and Anthropic backends
Parent Epic: #2358

### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The expected value of `providers.models[].backend_refs[].base_url` differs depending on the configured API format.

For an OpenAI-compatible backend, `/v1` must be included in `base_url`:

```yaml
base_url: https://backend.example.com/v1
```

so that requests are sent to:

```text
https://backend.example.com/v1/chat/completions
```

For an Anthropic-compatible backend, `/v1` must not be included:

```yaml
base_url: https://backend.example.com
```

because vLLM-SR appends `/v1/messages`.

If `/v1` is included for the Anthropic backend as well, the resulting upstream path contains `/v1` twice.

This makes the semantics of the same `base_url` configuration field dependent on the API format.

### To Reproduce

Configure equivalent OpenAI and Anthropic-compatible backends.

OpenAI:

```yaml
providers:
  models:
    - name: test-openai
      api_format: openai
      backend_refs:
        - name: backend
          base_url: https://backend.example.com/v1
```

This correctly sends requests to:

```text
https://backend.example.com/v1/chat/completions
```

Configure the Anthropic backend using the same `base_url` convention:

```yaml
providers:
  models:
    - name: test-anthropic
      api_format: anthropic
      backend_refs:
        - name: backend
          base_url: https://backend.example.com/v1
```

The Anthropic request is then constructed with `/v1/messages` appended to the configured URL, resulting in an incorrect path containing `/v1` twice.

To make the Anthropic request work, the configuration instead has to be:

```yaml
base_url: https://backend.example.com
```

so that the final path becomes:

```text
https://backend.example.com/v1/messages
```

### Expected behavior

`backend_refs.base_url` should have consistent semantics regardless of API format.

For example, if `base_url` represents the versioned API root, both providers could use:

```yaml
base_url: https://backend.example.com/v1
```

with provider-specific paths:

```text
OpenAI:    /chat/completions
Anthropic: /messages
```

Alternatively, if `base_url` represents the server root, both could use:

```yaml
base_url: https://backend.example.com
```

with:

```text
OpenAI:    /v1/chat/completions
Anthropic: /v1/messages
```

The important part is that users should not need to know different `base_url` conventions depending on `api_format`.

### Impact

This makes provider configuration unnecessarily confusing and error-prone, especially when the same backend exposes both OpenAI-compatible and Anthropic-compatible APIs, for example:

```text
https://backend.example.com/v1/chat/completions
https://backend.example.com/v1/messages
```

It is natural to expect both model definitions to use the same `base_url`, but currently they require different values.

This can easily result in invalid upstream paths such as:

```text
/v1/v1/messages
```

and makes it harder to generate or manage provider configurations consistently.

### Additional context

It appears possible to work around the inconsistency by explicitly configuring `chat_path`, for example:

```yaml
backend_refs:
  - name: backend
    base_url: https://backend.example.com/v1
    chat_path: /v1/messages
```

However, the built-in defaults still use different URL conventions.

If changing the defaults would introduce backward-compatibility issues, documenting this difference explicitly for OpenAI and Anthropic backends would at least make the behavior clearer.

rk9595:
Closes #3065

## Purpose

`providers.models[].backend_refs[].base_url` needed `/v1` for OpenAI-format backends but had to omit it for Anthropic-format ones. Giving an `anthropic` backend the complete API root that [`common-errors.md`](https://github.com/vllm-project/semantic-router/blob/main/website/docs/troubleshooting/common-errors.md) documents produced a doubled version segment upstream.

Reproduced on `main` @ `6ffc9c68` against `ProviderProfile.ResolveChatPath` (`src/semantic-router/pkg/config/helper.go`):

```
type=anthropic  base_url=https://backend.example.com/v1  -> /v1/v1/messages   <- bug
type=anthropic  base_url=https://backend.example.com     -> /v1/messages
type=openai     base_url=https://backend.example.com/v1  -> /v1/chat/completions
```

**Root cause.** `providerTypeRegistry` stores the version inside the per-type suffix for some types and not others:

```go
"openai":    {..., ChatPath: "/chat/completions"},      // version comes from base_url
"anthropic": {..., ChatPath: "/v1/messages"},           // version is in the suffix
"minimax":   {..., ChatPath: "/v1/chat/completions"},   // same shape as anthropic
```

`ResolveChatPath` concatenated the `base_url` path and the suffix with no reconciliation, so whether `/v1` belonged in `base_url` silently depended on the type. `minimax` shares anthropic's shape, so this was never a two-provider issue.

**The rule already existed on the other upstream-URL path.** `providerProtocolPath` (`src/semantic-router/pkg/extproc/processor_req_body_routing.go`) already collapses the duplicate for protocol dispatch. The two builders simply disagreed with each other.

**Fix.** Apply that same de-duplication in `ResolveChatPath` via a new `joinProviderBasePath` helper, so a trailing `/v1` on `base_url` is never doubled and the documented spelling works for every provider type.

Collapsing was chosen over declaring one convention canonical deliberately: redefining `base_url` to mean server-root for OpenAI (or versioned-root for Anthropic) would silently re-route every deployment already using the current convention, whereas de-duplication accepts the documented spelling and changes no working deployment's upstream path. If you would rather declare a single convention with a deprecation warning for the other spelling, I am happy to rewrite it that way — that is a config-semantics call that should be yours.

The helper extraction is not cosmetic: the inline form pushed `ResolveChatPath` past the repo's `cyclop` (13 > 12) and `nestif` limits under `make agent-lint`.

### Out of scope

`type=openai` with an unversioned `base_url` still resolves to `/chat/completions` with no version segment. Injecting `/v1` there would require declaring the convention this PR avoids declaring, and would break `azure-openai`, `bedrock`, `gemini`, and `vertex-ai`, which share the same suffix but not the same API root shape. Since the docs already define `base_url` as the complete API root, that case is a misconfiguration rather than a defect. Happy to file it separately if you disagree.

Consolidating `joinProviderBasePath` and `providerProtocolPath` into one shared helper is also left out — `pkg/extproc` imports `pkg/config` and not the reverse, so it is a wider refactor than this bug needs, and it would overlap #3292, which is reworking the Envoy-side rewrite path.

## Test Plan

- New table-driven spec in `src/semantic-router/pkg/config/config_test.go` covering `anthropic` / `minimax` / `openai` against both versioned and unversioned `base_url`, plus the cases that must **not** collapse (`/openai/v1`, `/v1beta`) and a trailing-slash root.
- Existing `ResolveChatPath` specs, including the `azure-openai` `?api-version` case and the explicit `ChatPath` override, kept unchanged as regression cover.
- `make agent-report ENV=cpu CHANGED_FILES="..."` → primary skill `config-platform-change`.
- `make agent-ci-gate CHANGED_FILES="src/semantic-router/pkg/config/helper.go src/semantic-router/pkg/config/config_test.go website/docs/troubleshooting/common-errors.md"`.

## Test Result

`make agent-ci-gate` passes (exit 0): baseline pre-commit, Go structural lint, `recipe-conformance-static`, `test-semantic-router`, and `dashboard-check` all green. The two `golangci-lint` findings on the first attempt (`cyclop` 13 > 12 and `nestif` on the `base_url` block) are resolved by the helper extraction.

The new spec was confirmed to actually cover the bug: with the `helper.go` change stashed it fails with `/v1/v1/messages` for `anthropic` + `https://api.anthropic.com/v1`, and passes with the change applied.

No E2E update: `ResolveChatPath` behavior changes only for a `base_url` spelling that is currently broken, and no repo-owned config, recipe, or E2E profile uses it. The `dashboard/backend` caller (`handlers/model_verification_service.go`) picks the fix up unchanged and its suite passes.

No remaining blocker.

### DIFF
--- a/src/semantic-router/pkg/config/config_test.go
+++ b/src/semantic-router/pkg/config/config_test.go
@@ -3649,6 +3649,27 @@ model_config:
 				Expect(path).To(Equal("/v1/chat/completions"))
 			})
 
+			It("should not double the version segment for a versioned base_url", func() {
+				for _, tc := range []struct {
+					providerType string
+					baseURL      string
+					expected     string
+				}{
+					{"anthropic", "https://api.anthropic.com/v1", "/v1/messages"},
+					{"anthropic", "https://api.anthropic.com", "/v1/messages"},
+					{"minimax", "https://api.minimax.io/v1", "/v1/chat/completions"},
+					{"minimax", "https://api.minimax.io", "/v1/chat/completions"},
+					{"openai", "https://api.openai.com/v1", "/v1/chat/completions"},
+					{"anthropic", "https://gateway.example.com/openai/v1", "/openai/v1/messages"},
+					{"anthropic", "https://gateway.example.com/v1beta", "/v1beta/v1/messages"},
+					{"anthropic", "https://api.anthropic.com/v1/", "/v1/messages"},
+				} {
+					path, err := (&ProviderProfile{Type: tc.providerType, BaseURL: tc.baseURL}).ResolveChatPath()
+					Expect(err).NotTo(HaveOccurred(), "%s %s", tc.providerType, tc.baseURL)
+					Expect(path).To(Equal(tc.expected), "%s %s", tc.providerType, tc.baseURL)
+				}
+			})
+
 			It("should append api-version for azure-openai", func() {
 				profile := &ProviderProfile{
 					Type:       "azure-openai",
--- a/src/semantic-router/pkg/config/helper.go
+++ b/src/semantic-router/pkg/config/helper.go
@@ -738,7 +738,8 @@ func (p *ProviderProfile) ResolveAuthHeader() (string, string, error) {
 //
 // Resolution order (no silent fallback):
 //  1. Explicit ChatPath field on the profile (used as-is, plus ?api-version for azure-openai).
-//  2. base_url path + type-default suffix from providerTypeRegistry.
+//  2. base_url path + type-default suffix from providerTypeRegistry, with a
+//     version segment repeated by both sides collapsed to one.
 //  3. Type-default suffix alone if base_url has no path component.
 //
 // Returns error if the type is not recognised or base_url is unparsable.
@@ -761,10 +762,7 @@ func (p *ProviderProfile) ResolveChatPath() (string, error) {
 		return path, nil
 	}
 
-	suffix := info.ChatPath
-	if p.Type == "azure-openai" && p.APIVersion != "" {
-		suffix += "?api-version=" + p.APIVersion
-	}
+	path := info.ChatPath
 
 	// Prepend base_url path component if present
 	if p.BaseURL != "" {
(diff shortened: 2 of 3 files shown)


Xunzhuo (approved):
Thanks for fixing this configuration edge case, and welcome! The shared suffix helper removes the duplicate slash path and the regressions cover both endpoints. LGTM.

PR #3327, opened 2026-09-02
[Bug] Pin CentOS base image to a resolvable stream10 digest with shell and dnf

### LINKED ISSUE #3312: [Bug] Extproc image build fails on every PR since quay.io retagged centos:stream10 (no /bin/sh)
### Proposed Workgroup

### Describe the bug

Since 2026-09-02 00:54 UTC every CI job that builds `tools/docker/Dockerfile.extproc` fails.

The final stage is `FROM quay.io/centos/centos:stream10`. That is a floating tag. CentOS re-pushed the tag at 00:54 UTC to a new digest, and the new image has no `/bin/sh`, so the first `RUN` step in that stage cannot start:

```
runc run failed: unable to start container process: error during container init:
exec: "/bin/sh": stat /bin/sh: no such file or directory
ERROR: process "/bin/sh -c dnf update -y && dnf install -y ..." did not complete successfully
```

This hits `E2E / integration-test (envoy-ai-gateway)` and `Operator / Operator smoke images`, and through them `PR Gate`, `Validate OLM Bundle`, and `test-and-build`.

Nothing in this repo changed. No merge to `main` in that window touches a Dockerfile.

| | Digest |
| --- | --- |
| Last passing build (PR #3304, 00:52 UTC) | `quay.io/centos/centos:stream10@sha256:ad14f7d919c9b9995a236abace06e888d37d6145e048a6ae26770e1f5dc718a8` |
| Every failing build | `quay.io/centos/centos:stream10@sha256:97fb91c65037aa5a97c223a230e82b40d271d720e43f8544c45855c71ef76265` |

### To Reproduce

1. Push any commit to any open PR, or rerun a failed image job.
2. Watch the extproc image build reach the `FROM quay.io/centos/centos:stream10` stage.
3. The `RUN dnf update -y ...` step fails with the `/bin/sh` error above.

Timeline from open PRs (image job completion time, UTC):

| Result | PRs |
| --- | --- |
| Passed | #3301 00:14, #3303 00:27, #3304 00:52 |
| Failed | #3268 01:23, #3305 01:43, #3307 01:57, #3308 01:59, #3250 02:03, #3306 02:11, #3249 02:11, #3251 02:19, #3299 02:23, #3162 02:34 |

Every build after 00:54 fails. Every build before passes.

### Expected behavior

The extproc image builds the same way regardless of what CentOS publishes under a moving tag.

### Impact

Ten open PRs are red right now and every new push will fail the same way. Authors cannot fix it from their branches. `main` will fail too on its next image build.

Workaround: none from a PR. Fix options:

1. Pin the base image to the last good digest (`sha256:ad14f7d9...`) in `Dockerfile.extproc`. One line, unblocks every PR, and stops the next silent retag from doing this again.
2. Wait for CentOS to fix or re-push the tag. No control over timing.

### Additional context

Example failing runs:

- https://github.com/vllm-project/semantic-router/actions/runs/33581025928/job/100095198794
- https://github.com/vllm-project/semantic-router/actions/runs/33581025928/job/100095198800

PeiwenPZ:
Fixes #3312

### Problem

quay.io/centos/centos:stream10 was re-pushed to a manifest whose only
outer layer expands to blobs/, index.json and oci-layout instead of a
root filesystem, so /bin/sh is absent and the first RUN in the final
stage cannot start. Both occurrences in this repo are immediately
followed by a RUN dnf ..., so both break the same way:

  tools/docker/Dockerfile:1
  tools/docker/Dockerfile.extproc:478

The digest suggested in the issue (sha256:ad14f7d9...) is no longer
resolvable — Quay returns not found, while other digests in the same
repository pull fine, so the manifest appears to have been garbage
collected after the tag moved off it. Pinning to it would not unblock
CI.

### Change

Pin both occurrences to stream10-development by digest:

  quay.io/centos/centos:stream10-development@sha256:556c6f55...

### Verification

Per the request in the issue, both /bin/sh and dnf are confirmed present
in the pinned image:

  $ docker create --platform linux/amd64 --name probe \
      quay.io/centos/centos:stream10-development
  $ docker export probe | tar -tf - | grep -E '^(usr/)?bin/(sh|dnf)$'
  usr/bin/dnf
  usr/bin/sh

Filesystem probe rather than an exec check because every Stream 10 / RHEL
10 image requires x86-64-v3, which my arm64 emulation layer doesn't
provide — Red Hat's own ubi10 fails identically on my machine, so this is
a local limitation, not a property of the image. CI runners are x86 and
support v3.

For reference, what else I checked:

  stream10            no /bin/sh (current, broken)
  10                  same digest as stream10, no /bin/sh
  stream10-minimal    no /bin/sh — so the bad push hit the minimal tag too
  stream10-development /bin/sh + dnf present  <- chosen
  stream9             /bin/sh + dnf, fully verified by exec
  ubi10               /bin/sh + dnf present

### Why this option

Smallest change that unblocks: same version line, one digest per file, no
package version churn.

  - stream9 would work and is the only one I could verify by execution,
    but it's a major version drop and would shift gcc / python3 / openssl
    versions.
  - ubi10 is arguably the better long-term base, but tools/docker/Dockerfile
    does `dnf -y install epel-release`, which needs different handling on
    UBI. That's not a change I can validate locally.

stream10-development is a development stream, so this is an unblock rather
than the end state — switching back to stream10 once upstream fixes the
published artifact would be the natural follow-up. Happy to switch to
ubi10 or stream9 instead if you'd prefer.

### Follow-up worth filing separately

Automated digest bumps (Renovate/Dependabot) for base images, so a silent
retag lands in one reviewable PR with CI on it instead of reddening every
open PR at once.

### DIFF
--- a/tools/docker/Dockerfile.extproc
+++ b/tools/docker/Dockerfile.extproc
@@ -475,7 +475,7 @@ RUN cd src/semantic-router && \
 # ============================================================================
 # Final stage: runtime image (candle on all platforms, onnx + openvino on amd64 only)
 # ============================================================================
-FROM quay.io/centos/centos:stream10
+FROM quay.io/centos/centos:stream10-development@sha256:556c6f559c63a24e52843778ed7ee0887be572f7df4219a971d2b2820bbb6675
 
 # Install runtime dependencies including OpenSSL, Python and HuggingFace CLI
 RUN dnf update -y && \
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos/centos:stream10
+FROM quay.io/centos/centos:stream10-development@sha256:556c6f559c63a24e52843778ed7ee0887be572f7df4219a971d2b2820bbb6675
 
 RUN dnf -y update && \
     dnf -y install epel-release && \


FAUST-BENCHOU (approved):

FAUST-BENCHOU:
Sorry for ealry merge but i think this is in a hurry

tools/docker/Dockerfile.extproc
```diff
@@ -475,7 +475,7 @@ RUN cd src/semantic-router && \
 # ============================================================================
 # Final stage: runtime image (candle on all platforms, onnx + openvino on amd64 only)
 # ============================================================================
-FROM quay.io/centos/centos:stream10
+FROM quay.io/centos/centos:stream10-development@sha256:556c6f559c63a24e52843778ed7ee0887be572f7df4219a971d2b2820bbb6675
```
yaojiejia, line 478: Thanks for fix this, all of my pr were red and now they are fixed

PR #3328, opened 2026-09-02
[Feature] Add shared connector for remote Router Model calls

JiaoliangYu:
related #3128

## Purpose

Introduce a small shared outbound connector for remote Router Model deployments.

The connector owns transport concerns that should be consistent across remote model protocols:

- endpoint and authentication setup
- per-attempt deadlines and caller cancellation
- opt-in retries for retry-safe operations
- bounded request, response, and error bodies
- typed errors and connection lifecycle

This PR migrates `HTTPClassifierInference` as the representative production caller while keeping the `/classify` request/response schema, label mapping, and score validation in the classification adapter. Both jailbreak and generic/category sequence-classifier consumers continue to use their existing interfaces.

vLLM chat completion is intentionally left unchanged. Protocol neutrality is covered by connector contract tests rather than by expanding the production migration surface.

## Test Plan

- `gofmt -d pkg/modelruntime/connector pkg/classification/...`
- `go vet ./pkg/modelruntime/connector ./pkg/classification`
- `go test -race ./pkg/modelruntime/connector ./pkg/classification`
- `make agent-ci-gate AGENT_BOOTSTRAP_DONE=1 ENV=cpu CHANGED_FILES="<14 changed files>"`
- `make precommit-branch-gate AGENT_BASE_REF=origin/main`

## Test Result

- Formatting check passed.
- Targeted vet and race tests passed.
- Agent CI gate passed, including pre-commit, security scan, Go structural lint, structure checks, recipe conformance, Rust builds, and the complete `test-semantic-router` suite.
- The repository pre-commit branch gate passed in the official pre-commit image.

---
<details>

- [x] The PR links accepted issue #3128 with the Router Models and Inference Runtime owner
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope and validation

</details>

### DIFF
--- a/src/semantic-router/pkg/classification/category_remote_backend_test.go
+++ b/src/semantic-router/pkg/classification/category_remote_backend_test.go
@@ -24,14 +24,12 @@ func TestCategoryHTTPBackendPreservesNamedFullDistribution(t *testing.T) {
 
 	deadline := 1500
 	backend, err := newCategoryHTTPBackend(&config.ExternalModelConfig{
-		ModelEndpoint: config.ClassifierVLLMEndpoint{Address: "127.0.0.1", Port: 8080},
+		ModelEndpoint: endpointForTestServer(t, server),
 		ModelName:     "named-category-service",
 	}, mapping, time.Duration(deadline)*time.Millisecond)
 	if err != nil {
 		t.Fatalf("newCategoryHTTPBackend: %v", err)
 	}
-	backend.(*categoryHTTPBackend).backend.(*HTTPClassifierInference).baseURL = server.URL
-
 	result, err := backend.ClassifyWithProbabilities(context.Background(), "solve this")
 	if err != nil {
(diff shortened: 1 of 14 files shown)


Xunzhuo (approved):
Thanks for extracting the shared connector, and welcome! The ownership boundary and connector tests preserve the classification behavior cleanly. LGTM.

Xunzhuo:
as a follow up, i think we should dig into other parts like loopers and see if we have any more dirty codes here

JiaoliangYu:
Thanks — I audited the Looper model-call path and opened #3339 as a bounded follow-up.

The proposed boundary keeps OpenAI request/response and algorithm semantics in `pkg/looper`, moves only shared transport concerns to the connector, removes mutable request metadata so the client can be reused concurrently, and ties the connector lifecycle to one router generation. I will implement it after the issue is accepted.

PR #3335, opened 2026-09-02
[Feature] Move response jailbreak detection into a response-stage signal

subin9:
Related #3272

## Purpose

- Make response jailbreak detection a response-stage signal: a `jailbreak` rule declares `direction: response` to score the model's output. Same `jailbreak:<name>` key and reporting; CLI schema, Dashboard and DSL carry `direction` as one more rule field. `method: contrastive`, the pattern lists and `include_history` are rejected on it.
- Detection is driven by the rules and scoped to the recipe the request resolved to, read through its classifier like request-stage classification. Scored once per response before any plugin runs; detected, not-detected and unavailable are distinct, the last landing in `SignalErrors` with its own code.
- The `response_jailbreak` plugin of the decision selected for the request consumes the observation and applies its action. Nothing is selected again after the model answers; a decision that reads a response-direction rule, directly in its rules or through a projection, is rejected at load. A response is clean only when every chunk of it was scored: a chunk the backend failed on leaves the rule unresolved unless another chunk matched, and `prompt_guard.on_error: block` fails closed on it.
- The plugin's own threshold is reported as ignored once a rule is declared; a plugin that still owns detection (no rule) is reported at load as the compatibility path. Each detector runs once.
- Declaring a response-direction rule, or enabling the plugin, provisions `prompt_guard` for the recipe (label mapping, model download, runtime task) even when no decision rule reads a jailbreak signal. Before, the mapping only loaded for a decision rule, so an `http_classify` guard refused to build and the plugin-only path never ran.
- Router Replay gets one outcome per response-direction rule (verdict, score or failure code, plugin action), recorded before a block returns. With `x-vsr-debug`, `x-vsr-matched-jailbreak` carries the response matches after the request ones.
- `dashboard/backend/go.sum` is tidied: `main` has needed it since #3445 and the Dashboard job is skipped on `main` pushes, so it surfaced on this PR's quality and dashboard gates.

## Test Plan

- `cd src/semantic-router && LD_LIBRARY_PATH=<rust libs> go test ./pkg/config/ ./pkg/extproc/ ./pkg/classification/ ./pkg/dsl/ ./pkg/routerreplay/...`
- `make e2e-test E2E_PROFILE=response-jailbreak`

## Test Result

- `TestResponseJailbreakSignalReadsTheSelectedRecipeRules` is the named-entrypoint regression; it fails when rules come from the root config.
- `TestResponseJailbreakSignalDrivesTheSelectedDecisionPlugin`, `TestResponseJailbreakBackendFailureIsNotHidden`, `TestResponseJailbreakSignalRecordsReplayOutcome`, `TestResponseJailbreakMatchedHeaderIncludesResponseRules`, `TestValidateRejectsDecisionReadingResponseDirectionRule`, and the three `TestJailbreakDirection*` DSL round-trips.
- `TestResponseJailbreakPartialScanFailureIsNotClean` and `TestCheckForJailbreakErrorsWhenACleanVerdictNeedsAFailedChunk` (a failed chunk beside clean ones is unresolved and blocks under `on_error: block`; a match in a scored chunk still counts), `TestValidateRejectsProjectionReadingResponseDirectionRule` (score, score of scores, confidence-valued output), `TestNeedsJailbreakMappingForResponseStageConsumers` with its recipe-reachability variant, and `TestInitializeRuntimeInitializesJailbreakClassifierForResponseStageConsumers`; each fails on the previous code.
- E2E `response-jailbreak`: window block and warning through the signal-driven path (the warning case also checks `x-vsr-matched-jailbreak`), plus the streaming pass-through contract.

## Not covered here

- Hallucination follows as a response-only signal on the same stage machinery.
- Streaming responses are still not scored; the E2E case pins the pass-through so scoring streams is a deliberate change.


Xunzhuo:
the inital idea for https://github.com/vllm-project/semantic-router/issues/3272 is that i think the design of halu and response jailbreak is not aligned with the signal decision architecture. We should move them from plugin to signal, same as other signals which can be composed as input of projection or decisions.

Just make sure we are aligning with the target

subin9:
@Xunzhuo Agreed, and this PR is not the move yet — execution still sits in the plugins. It only separates a response that was cleared from one that was never scanned, which both detectors currently report as the same thing.

Once they are signals, unavailable should land on `SignalErrors`, not a field of its own.

One question before I cut the move. `SignalResults` is only built in `req_filter_*`, and the non-goals rule out a response observation changing a selection that already executed. So what consumes a response-stage signal — the next turn, an enforcement-only decision, or an escalation path?

Happy to fold this into that PR instead.

Xunzhuo:
i dont have clear design but have some ideas,

for the response direction signal, like when we detected a response jailbreak or haluciation sigal, we can have some decisions like:
1. block the response with a plugin, or direct respond with a message: "this message contains xxxx..."
2. forward to a stronger or safer models to have response

and we can combine the request direction signal with response direction signal, this signal will only be activated at the response stage since we have not response signal when it is at request stage

Xunzhuo:
with the response direction signals working with request direction signals, we can express stronger routing recipes. like we can also detect the system stats as reponse direction signals as well as the structure validations for model structure output etc

subin9:
Thank you, that makes sense. I'll rework this PR in that direction. `response_jailbreak` as a response-stage signal type first, with unavailable on `SignalErrors` instead of the separate field I added here. I'll push it to this branch rather than opening a new one :)

subin9:
@Xunzhuo Reworked and pushed. Title and description updated too.

Signal types now declare a stage, and `response_jailbreak` is registered as the first response-stage one with a declarable `routing.signals.response_jailbreak` surface. A decision that reads it is skipped while the request is still being routed, kept separate from `on_unknown` so a detector that never ran cannot decide a request through `on_unknown: match`. Unavailable is on `SignalErrors` now, and the separate field is gone.

Detector execution is still in the plugin. That is the next change, along with hallucination in the same shape.

Three things I left in the description rather than guessing at: bounding the re-dispatch for "forward to a stronger model" and whether it should reuse `pkg/looper`, whether response-stage decisions share the tier and priority pool with request-stage ones, and whether a streamed response is scored once on completion or per chunk.

subin9:
@Xunzhuo Pushed the actual move rather than just the state, since the earlier version would have shipped a config surface nothing evaluated.

`response_jailbreak` is now a signal type. Rules are declared under `routing.signals.response_jailbreak`, evaluated once the model answers, and published to `SignalConfidences` and `SignalErrors` where every other signal publishes. A decision reads one as `{type: response_jailbreak, name: <rule>}` and can combine it with request-stage signals in the same rule, since those matches are still in scope by then.

Stage sits on the signal type rather than the rule, so `type:name` and the condition shape are unchanged, and system stats or structured-output validation would be response-stage types under the same rule with no further mechanism.

Each stage evaluates only the decisions whose signals exist by then, kept separate from `on_unknown` so a detector that never ran cannot decide a request through `on_unknown: match`.

Enforcement has not moved. The plugin still blocks and warns exactly as before, and this publishes the evidence it acts on, so the acting side can move onto the decision next without the observation moving with it. Hallucination is the same shape and follows.

The three things in the description are where I would like your read: bounding the re-dispatch for "forward to a stronger model" and whether it should reuse `pkg/looper`, whether response-stage decisions share the tier and priority pool with request-stage ones, and whether a streamed response is scored once on completion or per chunk.

subin9:
@Xunzhuo Pushed. Detection is out of the plugin — the declared rules drive it, so the observation exists whether or not `response_jailbreak` is enabled anywhere, and the plugin only applies its action.

I didn't wire a decision to resolve from it. Checking the issue again, the split it asks for is the plugin consuming evidence, so that felt like it belongs with the enforcement work — happy to add it now if you'd rather.

halu follows, and the block / replace / forward outcomes come with moving enforcement onto the decision :)

Xunzhuo (changes requested):
Thanks for pushing the stage split. I found three blockers on this head:

- response-stage decisions are skipped during request evaluation, but never evaluated after the response signal is published; the plugin still acts only on the already selected request decision, so request+response composition cannot drive a response action;
- the new signal is absent from the CLI schema, Dashboard signal/condition types, and DSL compiler/decompiler, so supported config surfaces cannot preserve it;
- this behavior-visible config/routing change has no buffered/streaming E2E coverage.

Please complete the response-stage decision seam and the required cross-surface/E2E contract before this lands.

config/config.yaml
```diff
@@ -409,6 +409,11 @@ routing:
...
             name: premium-tier
 
+    response_jailbreak:
+      - name: unsafe_completion
+        threshold: 0.85
+        description: Detect jailbreak content in the model's own output.
+
     jailbreak:
       - name: prompt_injection
         method: contrastive
```
Xunzhuo, line 414: can we add a field under existed jailbreak siganl like direction or something else?

subin9: Of coarse! working on the fix now, I'll include it :)

FAUST-BENCHOU (commented):
Could we add a test case covering the case where the backend is unavailable or returns an unexpected error during replay? It would help ensure the new fallback behavior does not hide real backend failures.

config/config.yaml
subin9: Added :)

subin9:
@Xunzhuo All three are on this head, and the description now matches it.

- Response-stage decisions are evaluated after the response. `EvaluateResponseStageDecision` runs from the request-stage matches plus the response observation once the signal is published, and the plugin enforces on whichever decision matched, so a request keyword AND the response-direction rule can drive a response action. `TestResponseStageDecisionDrivesTheResponseAction` pins that path to a 403, and `response-jailbreak-compose-block` covers it on a cluster.
- The stage is `direction: response` on the existing `jailbreak` rule, as you asked inline, so the CLI schema, Dashboard, and DSL compiler/decompiler carry it as one more rule field rather than a new type.
- E2E: the buffered block and warning cases now run through the signal-driven path, plus the compose case above and `response-jailbreak-streaming-passthrough`, which pins that a stream is delivered unscored.

@FAUST-BENCHOU `TestResponseJailbreakBackendFailureIsNotHidden` drives the guardrail backend to a 500: the failure lands in `SignalErrors` under the rule's key, no score is reported, no response-stage decision matches, and `on_error: block` still fails closed.

The lane's first run failed on the compose case sending a probe the profile's keyword signal does not match, the same defect the window cases had; fixed, and CI is rerunning on this head.

Xunzhuo:
@subin9 thanks for the work here! please rebase the code and fix conflicts

Xunzhuo (changes requested):
Thanks for the substantial follow-up. This still sends the response observation through a new `DecisionEngine` and can select a second decision after the model answers, contrary to #3272's current lifecycle: the already-selected response plugin should consume the typed observation. Response rules also come from root config while classification is recipe-scoped. Please use the selected recipe/config and add a named-entrypoint regression.

subin9:
@Xunzhuo Reworked it the way you described, thanks for the steer. The observation no longer goes through the decision engine and nothing is selected after the model answers; the plugin on the request-time decision consumes it. Rules come from the selected recipe now, and `TestResponseJailbreakSignalReadsTheSelectedRecipeRules` is the named-entrypoint regression (it fails when rules are read from the root config). While checking Replay I noticed the response-stage result never reached it, so it's recorded as one outcome per rule now, and `x-vsr-matched-jailbreak` picks up the response match in the body phase under the debug gate. Rebased on main and PR body updated.

Xunzhuo (changes requested):
Two response-safety blockers remain on this head:

- `scanJailbreakChunks` drops a chunk error once any other chunk succeeds, and the response path then publishes `resolved=true`. A clean result can therefore bypass `on_error: block` even though part of the response was never inspected. Please preserve positive matches, but return an unresolved/error outcome when any chunk needed for a clean verdict fails, with a partial-failure regression.
- Stage validation only walks direct decision leaves. A projection can still consume a response-direction jailbreak rule, and request-time evaluation turns its absent result into the configured miss value before a response exists. Please reject response-stage inputs anywhere in the decision dependency graph and cover the projection-indirection case.

subin9:
@Xunzhuo Both Fixed :) a partly scanned response is unresolved (match kept, on_error: block fails closed) and a response-direction rule behind a projection is rejected at load, each with a regression. 

The Memory Integration lane failed before the tests started, on an HF 429 while downloading the embedding model; could you re-run it?

Xunzhuo (commented):
Thanks—the partial-scan path now remains unresolved unless a scored chunk genuinely matches, and the projection walk covers direct and nested score/output references, so both blockers are resolved. The exact head only needs the Memory Integration lane rerun after its external Hugging Face 429 before approval.

Xunzhuo (changes requested):
Thanks for the update. One fail-closed gap remains. Response evaluation scans once at the lowest rule threshold; if a lower-threshold rule matches, `CheckForJailbreakRiskWithThreshold` suppresses a partial-scan error, then every rule is published as resolved. With thresholds 0.4 and 0.9, a 0.5-scored chunk plus a failed chunk incorrectly marks the 0.9 rule clean. Please preserve partial-scan failure per rule and add the multi-threshold regression.

Xunzhuo (approved):

subin9:
Fixed :)

Xunzhuo (commented):
lets add halu into the signal layer

subin9:
Okay, proceeding in #3531!

Issue #3195, opened 2026-08-31
[Feature] Define typed Router Model manifests, deployments, capabilities, and resolved bindings


Xunzhuo:
### Goal

Introduce the minimum typed runtime objects needed to resolve Router Model configuration before initialization or request handling.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Define immutable ModelManifest, Deployment, Capability, Binding, and ResolvedBinding representations.
- Separate model identity, artifact identity, task/result contract, runtime adapter, connector, device/provider policy, and recipe-local binding.
- Resolve references and reject duplicate, unknown, incompatible, or incomplete combinations before model preparation.
- Preserve current configuration through an explicit compatibility adapter.
- Expose normalized resolved state without secrets or mutable runtime handles.

### Non-goals

- Implementing a new tensor engine.
- Migrating every model family in one change.
- Defining routing decisions, model pools, or recipe policy.
- Adding hot training or automatic promotion.

### Completion signal

- A representative local classifier and remote classifier resolve through the typed objects.
- Invalid references and unsupported capability/backend combinations fail before initialization.
- Request-time callers receive a compiled ResolvedBinding rather than scanning raw configuration.

Parent Epic: #2782

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #3317, opened 2026-09-02
[Bug] Validate the vector-store collection dimension against the effective embedding dimension at startup


WUKUNTAI-0211:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

Nothing checks that a vector-store collection's vector dimension matches the embedding dimension the Router will actually produce. Both Milvus-backed stores create a collection from their own configuration and then load an existing one without inspecting its schema.

- `src/semantic-router/pkg/cache/milvus_cache.go:241` creates the semantic-cache collection from `config.Collection.VectorField.Dimension`, resolved through `semanticCacheEmbeddingDimension` (`src/semantic-router/pkg/cache/cache.go:183`) — a hardcoded `model -> width` map (`qwen3` 1024, `gemma` 768, `mmbert` 768, `multimodal` 384, default 384) that is not derived from the loaded model's contract.
- `src/semantic-router/pkg/memory/milvus_store_collection.go:16` creates the memory collection from an unrelated `config.Milvus.Dimension`, with no derivation at all.
- `EnsureCollectionLoadedWithHooks` (`src/semantic-router/pkg/milvus/lifecycle.go:105-119`) already takes an `onExists` hook for validating an existing collection, but both call sites reach it through `EnsureCollectionLoadedWithRetry`, which passes `nil` (`lifecycle.go:136`). An existing collection is therefore loaded unvalidated.

So a collection created under one embedding configuration keeps being loaded after that configuration changes, and the mismatch only surfaces as an insert or search failure on a live request.

### To Reproduce

1. Start the Router with a semantic cache backed by Milvus and an embedding model resolving to 768, and let the collection be auto-created.
2. Stop the Router, change the embedding model (or `target_dimension`) so the effective width becomes 384.
3. Start the Router again against the same Milvus instance.

The collection already exists, so `createCollection` is skipped and `LoadCollection` succeeds. Startup reports the cache as ready. The dimension mismatch appears later, per request, on insert or search.

### Expected behavior

Startup fails with a message naming the collection, its stored dimension, and the effective embedding dimension, instead of reporting a ready cache that cannot serve a request.

The effective dimension should come from the model contract established in #2407, not from a second hardcoded table.

### Impact

A configuration change that looks successful leaves the cache and memory paths broken at request time, with an error that points at Milvus rather than at the configuration that caused it. The hardcoded width map is also a fourth dimension table to keep in sync with the model contract.

### Additional context

[...]

PR #3354, opened 2026-09-02
[Feature] Expose typed embedding capabilities from native runtime bindings

### LINKED ISSUE #3336: [Feature] Expose typed embedding capabilities from native runtime bindings
### Goal

Give callers one typed query for native embedding capabilities, answered by the binding that owns the model implementation, so batching and model compatibility are no longer inferred from model-type string comparisons in the Go router.

### Proposed Workgroup

### Problem

The router currently has no authoritative interface for asking a native binding what an embedding model supports.

The immediate example is batching:

- `candle-binding/semantic-router.go:846-852` implements `SupportsBatchedEmbedding` as a normalized comparison against `"qwen3"`.
- `onnx-binding/semantic-router.go:742-747` exports the same symbol but always returns `false`, because ONNX does not use Candle's continuous-batching path.
- `src/semantic-router/pkg/modelruntime/router_runtime.go:377-385` independently compares the configured model type against `"qwen3"`.
- `src/semantic-router/pkg/extproc/router_selection.go:118-134` contains a normalization helper and a comment explaining that the tolerant capability check and strict embedding calls otherwise disagree about casing.

This means the Go layer still carries model-specific knowledge even though the capability belongs to the native runtime.

#2194 originally proposed moving the batching check behind the binding interface. The current implementation added the Go-facing symbol, but the Candle answer is still decided in the Go wrapper rather than by the Rust runtime. It also reduces an unsupported or misspelled model type to `false`, making it indistinguishable from a known model that legitimately lacks batching support.

Candle and ONNX expose a compatible Go surface through both build selection and Go module replacement. The ONNX router is built with `-tags=onnx` and `src/semantic-router/go.onnx.mod`, whose replacement at line 6 maps the imported Candle binding module to `../../onnx-binding`; see `tools/make/build-run-test.mk:51-60`.

OpenVINO does not currently participate in that replacement seam. It is compiled as a coexisting build-tagged module and selected separately by the Go extproc layer. Therefore, the `Backend` returned by this initial contract identifies the native binding that implements the shared Candle/ONNX surface; it must not be interpreted as the effective embedding executor for the existing OpenVINO path. Extending the same contract to OpenVINO requires a separate adapter decision.

### Relationship to existing work

This issue is a bounded child of #2396. It replaces the narrow batching probe from #2194 with a typed native embedding capability query.

The returned information may later be consumed by the deployment and binding objects proposed in #2782 and #3195. This issue does not define those objects or their lifecycle semantics.

Related work remains independently owned:

- #2318 owns deployment-facing discovery of loaded multimodal embedding capabilities and readiness. This issue defines lower-level static facts reported by a native binding; it does not replace the loaded/configured/unavailable state model or discovery API owned by #2318.
- #2668 and PR #3316 own configured-versus-effective backend selection, unavailable binary handling, backend validation, and related diagnostics.
- #3302 owns execution-provider-aware ONNX artifact ranking.
- #2926 owns Arm64 ONNX image support.
- #2394 owns research into base-plus-LoRA serving.

### Scope

- Define a versioned native-binding capability query for embedding model types.
- Return the same Go-facing capability shape from Candle and ONNX, with each binding supplying its own values.
- Report:
  - native binding identity for the shared Candle/ONNX binding seam;
  - canonical model type;
  - batching support;
  - supported modalities;
  - supported embedding dimensions; and
  - supported device families.
- Normalize the requested model type inside the capability query and return its canonical value.
- Return a typed unsupported-model error for an unknown model type instead of reporting `batching=false`.
- Marshal capability data through the Go binding without duplicating model-specific capability decisions in Go.
- Replace `SupportsBatchedEmbedding` and migrate its current router call sites to the capability query.
- Provide common conformance cases that Candle and ONNX must both satisfy.

This issue defines facts reported by a native binding. Callers may consume those facts, but deployment policy and compatibility resolution remain outside this issue.

### Interface sketch

Illustrative, not final. The field semantics and error behavior are the reviewable contract.

```go
type Backend string

const (
    BackendCandle Backend = "candle"
    BackendONNX   Backend = "onnx"
)

type ModelType string

type Modality string

const (
    ModalityText  Modality = "text"
    ModalityImage Modality = "image"
)

type Device string

const (
    DeviceCPU   Device = "cpu"
    DeviceCUDA Device = "cuda"
    DeviceROCm Device = "rocm"
    DeviceMetal Device = "metal"
)

type EmbeddingCapabilities struct {
    Version             uint32
    Backend             Backend
    ModelType           ModelType
    SupportsBatching    bool
    Modalities          []Modality
    SupportedDimensions []int
    SupportedDevices    []Device
}

func EmbeddingCapabilitiesFor(modelType string) (EmbeddingCapabilities, error)
```

`ModelType` in the result is canonical. For example, an accepted case-insensitive input such as `" Qwen3 "` returns `"qwen3"`.

`SupportedDevices` describes device families supported by the built binding and model implementation. It does not identify the active device, guarantee availability on the current host, or expose backend-specific execution providers such as MIGraphX, DirectML, or OpenVINO EP.

An empty `SupportedDimensions` means the binding does not impose a finite allowlist through this contract. It must not be interpreted as supporting no dimensions.

The native representation must define stable versioning, enum/value semantics, memory ownership, and error codes. The Go wrapper is responsible for marshalling that representation, not for deciding capabilities.

### Error behavior

Unknown and unsupported model types must be distinguishable from known models that do not support a feature.

For example:

```go
var ErrUnsupportedModelType = errors.New("unsupported embedding model type")
```

Callers must be able to use `errors.Is` or `errors.As` rather than parsing an error message.

The following cases are distinct:

- known model with `SupportsBatching=false`;
- model type unknown to the binding;
- capability query unavailable because the native binding is not built; and
- malformed capability data returned across the FFI.

### Non-goals

- Defining deployment-level `Capability`, `Binding`, `ResolvedBinding`, readiness, lifecycle, or compatibility receipts; those belong to #2782 and #3195.
- Providing deployment-facing loaded-model discovery; tracked by #2318.
- Validating configured backend identity or changing backend fallback behavior; tracked by #2668 and PR #3316.
- Changing ONNX artifact selection or candidate ranking; tracked by #3302.
- Adding Arm64 ONNX image support; tracked by #2926.
- Defining routing decisions, signal composition, or model-selection policy.
- Adding support for a new backend, model family, modality, device, or execution provider.
- Defining dynamic adapter or base-plus-LoRA capability semantics; tracked by #2394.
- Adding runtime readiness to the static capability descriptor.
- Changing existing embedding behavior for supported model types.

### Completion signal

- Candle and ONNX expose the same versioned Go-facing embedding capability interface.
- Capability values originate from the native runtime implementation rather than model-name tables in the Go wrappers.
- Known model types return a canonical model type and backend-specific capability data.
- Unknown model types return a typed unsupported-model error.
- `SupportsBatchedEmbedding` is removed from Candle, ONNX, and mock bindings.
- Existing callers use the capability result rather than comparing model-type strings to decide whether the batched embedding path is available.
- The casing workaround and explanatory comment in `router_selection.go:118-134` are removed because normalization has one owner.
- Candle and ONNX pass equivalent conformance cases covering canonical model normalization, known non-batched models, backend-specific batching behavior, supported dimensions and modalities, unknown model errors, and unavailable native bindings.
- Existing embedding behavior remains unchanged for supported model types.

### Affected code and surfaces

Expected primary surfaces:

- `candle-binding/src/ffi/**`
- `candle-binding/semantic-router.go`
- `candle-binding/semantic-router_mock.go`
- `onnx-binding/src/**`
- `onnx-binding/semantic-router.go`
- `src/semantic-router/pkg/extproc/router_selection.go`
- `src/semantic-router/pkg/modelruntime/router_runtime.go`
- associated binding and router tests

The final changed-file set may identify additional required surfaces through the repository harness.

### Validation

Resolve the final validation set from the files actually changed:

```bash
make agent-report ENV=cpu CHANGED_FILES="<changed files>"
```

At minimum, run the binding and router gates selected by that report. Because the change affects a native binding interface and runtime dispatch, final validation must include the applicable binding conformance tests and semantic-router tests.

### Sequencing

This issue can land independently of #2668, #3302, and #2926.

It should define only the native capability-reporting seam. #3195 may later consume the returned facts when resolving deployment-level capabilities and bindings, without redefining model-specific support tables in the Go layer.

Parent Epic: #2396

bugkeep:
Closes #3336

## Purpose

Expose `EmbeddingCapabilitiesFor` with the same typed Go API in Candle and ONNX. Native bindings own backend identity, model normalization, batching, modalities, compiled device families, and dimension metadata. Legacy model-runtime initialization resolves batching from this contract before loading. Extproc selection consumes the generation-owned prepared providers introduced by #3768.

The dimension lifecycle follows the observed-after-load option in the [owner's clarification](https://github.com/vllm-project/semantic-router/pull/3354#issuecomment-5522312157):

- Static capabilities remain available before initialization. Dimensions have `DimensionStateNotLoaded`, native width zero, and an empty list.
- After a public legacy embedding initializer prepares a model, a fresh query returns `DimensionStateAvailable`, the native width, and deduplicated model-declared widths that do not exceed it. An empty list never means unrestricted support; list order does not choose the default.
- Both lifecycle surfaces use the native model's existing config/getters, including Gemma's inference dimension constant and the loaded multimodal config. These are the same authoritative facts used by #3355. This PR does not call #3355's loaded-model lookup or depend on its branch.

Qwen3 dimension discovery also recognizes the public batched-only initializer. Its context retains metadata read from the loaded base model before the scheduler takes ownership; capability lookup uses it when no standard Qwen3 model is loaded.

Generation-owned bindings continue to expose instance-specific metadata through their handle contract. The model-type capability query observes dimensions retained by the legacy initializers. ONNX retains its model-loaded Matryoshka configuration and applies the upstream `legacy-ffi` export guards to the existing unsafe FFI functions so Candle and ONNX instance ABIs can coexist.

The native ABI has explicit status codes, and the Go API distinguishes unsupported models, unavailable native runtimes, and malformed FFI data. Dimension buffers are owned by native code, copied into Go, and released through the matching backend's free function. `Version` identifies the ABI descriptor; task-contract identity/version remains #3195 work.

Existing inference and backend-selection behavior is preserved. Dimension enforcement and public API/default changes remain #2407 work. The OpenVINO trim/lowercase fix and Docker source-copy regression coverage are included. ONNX keeps the pointer safety declarations and two private, same-file batch helpers required by the touched-file lint/structure checks; no separate similarity module or result-packaging abstraction is added.

The accepted issue has one owner, `wg/router-models-inference-runtime`. No config schema or user-visible routing contract is added.

ONNX capability discovery filters out widths above the loaded native dimension. The checked-in 3-wide fixture now advertises only `[3]`. Loaded conformance checks execute every advertised width through `GetEmbeddingWithModelType` and verify the output length. This correction does not change inference enforcement.

## Test Plan

- Run the repository gate against the merged baseline:

  ```bash
  CHANGED_FILES="$(git diff --name-only b4be720cd6d511469ba3995dc2b60e3a2ab1fb02...HEAD | tr '\n' ' ')"
  CI=true CI_MINIMAL_MODELS=true make check ENV=cpu BASE_REF=b4be720cd6d511469ba3995dc2b60e3a2ab1fb02 CHANGED_FILES="$CHANGED_FILES"
  ```

- With CPU ONNX Runtime 1.22.0 selected through `ORT_DYLIB_PATH` and native libraries on `LD_LIBRARY_PATH`, run `CI=true make test-owned-native ENV=cpu`.
- In `onnx-binding`, build with `cargo build --release --no-default-features --features legacy-ffi,dynamic --lib`, then run:

  ```bash
  cargo test --release --no-default-features --features legacy-ffi,dynamic --lib ffi::
  go test -count=1 -race -v .
  ```

  `TestNativeEmbeddingCapabilitiesLoadedDimensions` loads the checked-in `instance/testdata/embedding` fixture through the public legacy initializer in a separate process. It checks the unloaded-to-loaded transition, the default output width, and actual inference for every advertised width. The existing model-backed capability assertion now checks every advertised width too.

## Test Result

Validated on `3b27ba65`, incorporating main at `b4be720c`:

- Reproduced the review failure before the descriptor fix: the 3-wide fixture advertised `[3, 768, 512, 256, 128, 64]`, and all five oversized widths failed the inference round trip on the merged baseline.
- PASS after the fix: the fixture reports native width 3 and supported widths `[3]`; every advertised width produces exactly that many floats. Rust regressions cover oversized defaults, mixed valid/invalid widths, duplicate removal, and native-only support.
- PASS: all 34 ONNX FFI unit tests and the legacy Go suite with the race detector. Tests needing an external checkpoint retain their skip conditions; the new fixture regression runs without a checkpoint download.
- PASS: `make check`, including the repository's formatting/lint, changed-file Rust lint, harness contracts, router tests, and minimal Candle binding checks.
- PASS: `make test-owned-native` with CPU ONNX Runtime 1.22.0.
- The main merge resolves three textual conflicts and reuses main's `ErrBackendUnavailable` sentinel instead of redeclaring it. The dimension correction is confined to the ONNX descriptor collector and its tests; it does not change inference enforcement.
- An additional `CGO_ENABLED=0 go test -count=1 .` check in `onnx-binding` fails because main's `classifier_context_mock.go` references six undefined legacy classifier initializers. The same missing symbols were reproduced on unmodified main `b4be720c`; no unrelated mock implementation is included in this fix.
- GPU inference and full image builds were not run locally. Rust lint results refer to the repository's changed-file gate.
- [CI for `3b27ba65`](https://github.com/vllm-project/semantic-router/actions/runs/35172292625) is pending.

---
<details>

- [x] PR title begins with exactly one allowed bracketed category
- [x] The title uses one prefix; affected modules are described in the body
- [x] The PR links accepted issue #3336 with exactly one `wg/*` owner
- [x] Purpose, Test Plan, and Test Result reflect the actual scope and local limitations

</details>


FAUST-BENCHOU (commented):
Could we clarify the expected behavior when multiple ModelRoutes share the same limiter state? Removing or changing the shared state handling may affect existing routes that rely on the current aggregation behavior.

WUKUNTAI-0211 (changes requested):
Thanks for working on this. The overall native-owned capability design looks good, but I think these points should be addressed before merging:

1. **Move capability lookup out of the request path.** `resolveSelectionEmbeddingFunc` currently calls `EmbeddingCapabilitiesFor` for every embedding request. Since the capability is static for a configured binding/model, please resolve it once when constructing the embedding function and capture the canonical model type and batching decision in the closure.

2. **Preserve the existing BERT path.** `batchedEmbeddingNeeds` may query capabilities for `bert`, but BERT uses a separate embedding path and is outside this capability inventory. This can prevent otherwise valid unified models from initializing. Please restrict the query to model types owned by the unified embedding seam.

3. **Resolve ONNX compatibility explicitly.** The existing ONNX wrapper routes non-multimodal model names to mmBERT, while the new capability query accepts only `mmbert` and `multimodal`. Please migrate the ONNX callers/defaults to the canonical model type, with regression coverage. Preserving arbitrary-name fallback would conflict with #3336’s requirement that unknown model types return a typed error.

Separately, please clarify whether unavailable ONNX builds are part of the supported build matrix. If they are, the capability query needs the corresponding typed unavailable behavior; otherwise this limitation should be explicit.

I’m not asking for a shared abstraction or broader refactor here. These changes are limited to lifecycle placement, preserving existing behavior, and clarifying the supported build contract.

bugkeep:
Addressed the requested changes in 47bb6bc. Capability lookup for model selection now happens once during registry construction and the closure captures the canonical model type and batching decision. The runtime batching check only queries unified embedding models, so the separate BERT path is preserved. ONNX now uses mmbert as its explicit default, rejects arbitrary model names with ErrUnsupportedModelType, and returns ErrBackendUnavailable in non-native builds. Added regression coverage for construction-time lookup, recipe model plans, BERT exclusion, ONNX compatibility, and unavailable builds. Local agent-ci-lint and agent-ci-gate pass; the native ONNX and CGO-disabled focused tests also pass. @WUKUNTAI-0211 please take another look when convenient.

bugkeep:
Follow-up pushed in 695330b.

This completes the requested review follow-up:
- capability selection is resolved once while constructing the embedding function and captured by the closure;
- the independent BERT path remains outside the unified capability inventory;
- ONNX callers and defaults now use the canonical mmbert model type, with regression coverage for typed unknown-model behavior;
- the ONNX native versus capability-only build contract is documented, including typed backend-unavailable behavior on unsupported builds.

While validating the linked ONNX router build, I also found that Candle and ONNX exported the same private C capability symbol. ELF symbol interposition could therefore make an ONNX build read the Candle inventory. The private C symbols are now binding-namespaced, while the shared Go-facing EmbeddingCapabilitiesFor API required by #3336 remains unchanged.

Validation passed: make agent-ci-gate for all 8 changed files; make test-semantic-router; make test-binding-minimal; Candle and ONNX Rust capability tests; ONNX native and CGO-disabled Go packages; and targeted router tests under go.onnx.mod. test-binding-lora remains environment-limited because the restricted Qwen3, Gemma, and classifier model artifacts are unavailable. Local image smoke was not counted as valid because the router image rebuild was blocked by an external TLS failure while fetching NVIDIA cutlass.

@WUKUNTAI-0211 could you please re-review?

WUKUNTAI-0211 (changes requested):
Thanks, the previous review items are addressed. One regression remains: OpenVINO is excluded from the new capability normalization, while `openvinoEmbeddingFunc` still dispatches using exact `"mmbert"` / `"modernbert"` matches. Values such as `" MMBERT "` previously normalized correctly but now fall through to the default BERT path.\n\nPlease preserve trim/lowercase normalization before OpenVINO dispatch and add a focused regression test. This does not require adding OpenVINO to the capability contract.

WUKUNTAI-0211:
Coordination note after closing superseded PR #2639: #3355 is now the active binding-owned dimension-contract implementation for its #3317 consumers. `EmbeddingCapabilitiesFor` should project that same authoritative native/default + supported-dimensions source once available, rather than preserve a second model table or permanently define an empty list as unrestricted prefix support.

The remaining #2407 Candle/ONNX modality enforcement and API behavior have been recorded separately on #2407; they should not expand this PR. For this PR, please finish the outstanding OpenVINO normalization regression, keep the capability lookup outside the request path, and coordinate/rebase against #3355 before finalizing `SupportedDimensions` semantics.

WUKUNTAI-0211:
Clarification to my previous coordination note: #3354 should not be forced to call #3355’s loaded-model lookup directly. The two PRs currently operate at different lifecycle points: #3354 supplies static/preparation-time capability facts, while #3355 observes the loaded model for #3317 consumers. What must be shared is the authoritative model metadata and invariants, not necessarily one query surface.

For `SupportedDimensions`, please coordinate one of these explicit models with #3355/#2407: declared dimensions available before load and verified after preparation, or observed dimensions available only after load with typed not-loaded/unavailable behavior. Avoid a parallel model-name table and avoid defining an empty list as unrestricted if #2407 will permit only model-declared dimensions.

Also, I see the OpenVINO trim/lowercase regression fix has now been pushed on the current head, so that part of my earlier note is no longer an outstanding author request. The FFI `Version` here remains an ABI descriptor version; a future embedding task-contract identity/version is separate #3195 work.

bugkeep:
@WUKUNTAI-0211 I've addressed your review follow-ups. OpenVINO trim/lowercase normalization is preserved, and `SupportedDimensions` now has an explicit not-loaded/available lifecycle using authoritative native model metadata. This PR remains independent of #3355's loaded-model lookup, and capability lookup stays out of the request path.

CI is green on `cdc686b`, including the CUDA and ROCm image builds: [CI run](https://github.com/vllm-project/semantic-router/actions/runs/34206180553).

Could you please take another look when convenient? Thanks!

candle-binding/src/ffi/capability_dimensions.rs
```diff
@@ -0,0 +1,112 @@
...
+
+fn loaded_metadata(model_type: u32) -> Option<(usize, Vec<usize>)> {
+    if model_type == MODEL_TYPE_MULTIMODAL {
+        let (model, _) = get_multimodal_refs()?;
+        return Some((
+            model.get_embedding_dimension(),
+            model.get_matryoshka_dimensions(),
+        ));
+    }
+    let factory = GLOBAL_MODEL_FACTORY.get()?;
```
Xunzhuo, line 73: **Recognize batched-only Qwen initialization**

Impact: A caller can successfully initialize and use the public batched Qwen API, yet EmbeddingCapabilitiesFor(qwen3) reports DimensionStateNotLoaded, preventing consumers from obtaining or validating the loaded model's dimensions through the new capability contract.

Required change: Have the capability query read dimension metadata from GLOBAL_BATCHED_MODEL or register shared loaded metadata, and add a conformance test that invokes only InitEmbeddingModelsBatched before querying capabilities.

Xunzhuo (commented):

bugkeep:
> Thanks for the update—I found one current-head issue that needs to be addressed before this can merge.

ok，i do it

onnx-binding/src/ffi/capability_dimensions.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl EmbeddingDimensions {
+    pub fn from_model(native: usize, declared: &[usize]) -> Result<Self, ()> {
+        let native = positive_dimension(native)?;
+        let mut supported = vec![native];
+        for &dimension in declared {
+            let dimension = positive_dimension(dimension)?;
+            if !supported.contains(&dimension) {
+                supported.push(dimension);
```
Xunzhuo, line 19: **[P2] Only advertise dimensions the loaded ONNX model can produce**

Loading the checked-in `instance/testdata/embedding` ONNX fixture (hidden size 3) reports `NativeDimension: 3` but `SupportedDimensions: [3, 768, 512, 256, 128, 64]`. I ran each advertised width through `GetEmbeddingWithModelType`; all five larger widths return 3 floats. `matryoshka_dimensions()` supplies the fixed mmBERT defaults and this collector appends them without checking the loaded width, so a consumer using the new capability descriptor can select an impossible output shape. Filter or reject incompatible dimension metadata and add a loaded-model conformance test that verifies the advertised widths; this can be fixed in the descriptor without changing inference enforcement.

Xunzhuo (changes requested):
The loaded-model capability descriptor still advertises output widths that the model cannot produce. The checked-in ONNX fixture reproduces this with the native library: its 3-dimensional output is reported as supporting 64–768 dimensions. Please filter or reject incompatible dimension metadata and add a loaded-model conformance test for each advertised width. The existing native capability tests pass; the added round-trip check fails for all five larger widths.

Xunzhuo (dismissed):
The loaded ONNX capability contract now excludes widths larger than the actual model output. Rebuilding this head and running the checked-in 3-wide model confirms that it advertises only [3], and every advertised width produces the promised dimension. The seven Rust dimension tests and five focused Go test groups pass.

wilsonwu (dismissed):
LGTM

bugkeep:
> LGTM

Thanks for approving this PR earlier! The merge was blocked by conflicts, which I’ve now resolved. The PR requires re-approval after the update. Could you please take another look and approve it again when you have a chance? Thanks!

candle-binding/embedding_capabilities_native.go
```diff
@@ -0,0 +1,214 @@
+//go:build !windows && cgo && (amd64 || arm64)
```
Xunzhuo: **[P1] Include the capability implementation in RISC-V builds**

The current native Candle files support `linux/riscv64`, but this new file still excludes that architecture. With `CGO_ENABLED=1 GOARCH=riscv64`, neither this implementation nor the `windows || !cgo` stub provides `EmbeddingCapabilitiesFor`; the new capability test helpers are excluded the same way. The current QEMU job fails compiling `semantic-router_test.go` with these undefined symbols before it can run the binding tests, and the router also calls this API from `modelruntime`. Align the capability implementation and helper build constraints with the supported native architectures, then rerun the QEMU binding/router checks.

Xunzhuo (changes requested):
The capability dimension fixes remain covered by the passing native Core tests. The updated main integration exposes a RISC-V build failure: the new capability implementation and test helpers exclude an architecture supported by the surrounding Candle binding. Please align those build constraints and rerun the QEMU binding/router checks. The current QEMU job stops during Go test compilation; it does not validate router startup.

tools/make/rust.mk
```diff
@@ -136,7 +143,7 @@ test-rust-flash-attn-module: rust-flash-attn
 # Hermetic Go/C ABI contracts. Published checkpoint inference is a separate
 # required suite (make test-models), not a whitelist of optional legacy tests.
 # The same contracts run under RISC-V QEMU without the unsupported race detector.
-BINDING_MINIMAL_GO_TESTS ?= ^Test(Owned.*|NewRegexProvider|RegexProvider_.*|UtilityFunctions)$$
+BINDING_MINIMAL_GO_TESTS ?= ^Test(Owned.*|NewRegexProvider|RegexProvider_.*|UtilityFunctions|EmbeddingCapabilitiesConformance|EmbeddingDimensionStateValidation)$$
```
Xunzhuo, line 146: **[P2] Update the RISC-V evidence contract with this test selection**

This adds two capability test groups to `BINDING_MINIMAL_GO_TESTS`, but `tools/ci/riscv_evidence.py` still accepts only the previous exact pattern. The current QEMU job successfully runs the binding tests and router smoke checks, then fails evidence extraction with `RISC-V minimal bindings must use the maintained owned fixture selection`; no qualified result is written, so the required gate fails on every run. Update the evidence inventory and its tests together with this selection, then rerun the QEMU check.

Xunzhuo (changes requested):
The previous RISC-V build failure is resolved: current-head QEMU CI runs the binding, classifier and router smoke checks successfully. Qualification still fails because the expanded Go test selection does not match the evidence parser’s exact inventory. Please update that contract and its tests together, then rerun the RISC-V gate.

Xunzhuo (approved):
The capability and RISC-V follow-ups are addressed. The native ONNX fixture now advertises only its actual 3-dimensional output, and every advertised width round-trips correctly. Batched-only Qwen, native CPU, RISC-V QEMU, and Clippy checks also pass.

I rebuilt the ONNX binding and ran all four focused capability test groups against this head. I found no remaining blocking issue. GPU execution, physical RISC-V hardware, and a full service deployment were not tested.

bugkeep:
@Mergifyio queue

bugkeep:
@shraderdm @drivebyer @ramkrishs @AayushSaini101 @theohsiung @WUKUNTAI-0211 @FAUST-BENCHOU @Peterren @wilsonwu

The capability and RISC-V review follow-ups have been addressed on the current head (`a13cd33d`). Could you please review the updated PR and update your review when you have a chance? Thank you!

bugkeep:
All requested changes have been addressed, and Xunzhuo has approved the current head. Please proceed with merging once the remaining required approvals and CI checks are complete. Thanks for reviewing!

wilsonwu (approved):
LGTM

bugkeep:
@Mergifyio queue

PR #3355, opened 2026-09-02
[Bug] Validate Milvus collection dimensions against embedding contracts

### LINKED ISSUE #3317: [Bug] Validate the vector-store collection dimension against the effective embedding dimension at startup

Alanxtl:
Closes #3317

## Purpose

Fix Milvus startup failures caused by mismatched vector dimensions.

- Resolve the effective embedding dimension from the loaded Candle embedding model contract, following #2407.
- Remove the cache-side hardcoded model-to-dimension mapping.
- Reuse the resolved dimension for Milvus collection creation and embedding generation.
- Validate existing Milvus collection schemas before loading them.
- Fail startup with the collection name, stored dimension, and expected dimension when they differ.
- Apply the same validation to both `MilvusCache` and memory `MilvusStore`.

## Test Plan

- `cargo fmt --manifest-path candle-binding/Cargo.toml -- --check`
- `cargo check --manifest-path candle-binding/Cargo.toml --no-default-features`
- `cd semantic-router/src/semantic-router && go test ./pkg/milvus"`
- `git diff --check`

## Test Result

- Rust formatting check passed.
- CPU-only Candle Rust check passed.
- Milvus lifecycle and schema tests passed.
- Full memory tests are currently blocked because WSL does not have the Linux `libcandle_semantic_router` library available for linking.
- CUDA-enabled `cargo check` is blocked because `nvcc` is not installed.

---
<details>

- [x] The PR links accepted issue #3317 with exactly one owner: `wg/router-models-inference-runtime`

</details>


FAUST-BENCHOU (commented):
Could we add a test covering the transition between the old and new behavior? In particular, it would be useful to verify that existing configurations keep the same routing result after this change.

Alanxtl:
> Could we add a test covering the transition between the old and new behavior? In particular, it would be useful to verify that existing configurations keep the same routing result after this change.

done, ptal

WUKUNTAI-0211:
Coordination note: I closed #2639 to remove the overlapping old implementation. Please treat the binding-owned model -> native/default dimension + supported-dimensions lookup in this PR as the single source consumed by the #3317 Milvus/cache/memory work.

Two scope constraints matter before merge:
1. Do not duplicate capability metadata that #3354 will project; expose one authoritative contract that it can consume.
2. This PR does not replace the remaining #2407 multimodal correctness work: Candle/ONNX modality enforcement, no silent clamp/substitution, and API model-resolved error/default behavior remain follow-ups recorded on #2407.

Please rebase onto current main and keep the contract surface narrow enough for #3354 to consume without maintaining a second dimension table.

bugkeep:
Hi @Alanxtl — I’m coordinating #3354 with this PR before finalizing the dimension fields in `EmbeddingCapabilitiesFor`.

#3354 currently exposes a static binding capability query that is resolved once during embedding-function construction and then captured outside request handling. In the current #3355 diff, `GetEmbeddingDimensionContract` is backed by the loaded Candle model and returns `NativeDimension` plus `SupportedDimensions`.

I want #3354 to project that exact authoritative source, without introducing another model table or permanently treating an empty dimension list as unrestricted support. Could you please confirm:

1. Which lifecycle is intended for consumers? The #3336 capability query is currently safe before model-factory initialization, while the #3355 contract requires a loaded model. Should #3355 expose an immutable contract seam usable before factory initialization, or should #3354 query and cache it once after model load? If the latter, what typed result should capability discovery return before the model is loaded?

2. Is an equivalent `EmbeddingDimensionContract` / `GetEmbeddingDimensionContract` surface planned for ONNX in this PR? `go.onnx.mod` replaces the Candle module with ONNX, so ONNX builds need the same Go-facing contract.

3. Should capability V1 add an explicit `NativeDimension` or `DefaultDimension` field and populate `SupportedDimensions` from this contract, rather than infer the default from list ordering?

4. After #3355 is rebased onto current main, would you prefer #3354 to rebase directly onto this PR’s head, or wait until #3355 merges?

I have already pushed the OpenVINO trim/lowercase regression fix to #3354, and capability lookup remains outside the request path. I will keep the remaining #2407 modality-enforcement and API behavior work out of #3354.

WUKUNTAI-0211:
Clarification to my previous coordination note: by “single source” I meant one authoritative set of model dimension facts, not necessarily one API or one lifecycle seam. This PR currently exposes an observed contract from a loaded model for #3317 consumers. #3354 needs capability facts during construction/preparation and may require a declared/pre-load representation.

The design decision is therefore:
- either expose immutable declared metadata before load and verify it against the prepared model; or
- expose the observed contract after load and make callers handle typed not-loaded/unavailable states.

Please do not broaden #3355 solely to satisfy #3354 until that lifecycle is agreed. `EmbeddingDimensionContract` can remain a meaningful typed value object; it does not need to be renamed merely because the future task contract identity/version belongs to #3195. The invariant is that declared and observed facts cannot drift or be maintained in parallel model-name tables.

Alanxtl:
Thanks for the clarification @WUKUNTAI-0211 @bugkeep  . I’ll keep #3355 focused on an observed, post-load embedding dimension contract for the #3317 Milvus/cache/memory consumers.

1. `GetEmbeddingDimensionContract` is valid only after the corresponding native model has been loaded. Before model initialization, the current surface returns an error and does not fabricate capability metadata. I will not add a second pre-load model-to-dimension table in #3355. A declared/pre-load representation and its typed unavailable state should be decided separately for #3354/#3336.
2. The equivalent Go-facing contract has now been added for ONNX, because `go.onnx.mod` replaces the Candle binding. Its values are read from the loaded ONNX model rather than from another hardcoded table.
3. I agree that capability V1 should expose an explicit `NativeDimension` together with `SupportedDimensions`. Consumers should not infer the native/default dimension from list ordering, and an empty list must not mean unrestricted support.
4. I’ll rebase #3355 onto current main first. To minimize repeated rebases, I suggest that #3354 wait for #3355 to merge. If earlier integration is required, it can temporarily stack on the rebased #3355 head without duplicating the model dimension facts.

The remaining #2407 work—modality enforcement, eliminating silent dimension substitution/clamping, and API-level model-resolved behavior—will remain outside this PR.

bugkeep:
OK. I’ll wait for #3355 to be rebased and merged first, then continue with #3354 on top of the updated main.

Alanxtl:
> OK. I’ll wait for #3355 to be rebased and merged first, then continue with #3354 on top of the updated main.

already rebased

Xunzhuo (dismissed):
Thanks for the substantial first contribution and for adding the ONNX equivalent. One current-head blocker remains: the full pre-commit job reports `onnx-binding/embedding_contract.go:64` at cyclomatic complexity 13, above the repository limit of 12. Please split the model-specific lookup/validation out of `GetEmbeddingDimensionContract` (without adding a second dimension table) and rerun the PR gate. The focused agent validation and the substantive core, memory-integration, E2E, and image jobs otherwise pass.

Alanxtl:
> Thanks for the substantial first contribution and for adding the ONNX equivalent. One current-head blocker remains: the full pre-commit job reports `onnx-binding/embedding_contract.go:64` at cyclomatic complexity 13, above the repository limit of 12. Please split the model-specific lookup/validation out of `GetEmbeddingDimensionContract` (without adding a second dimension table) and rerun the PR gate. The focused agent validation and the substantive core, memory-integration, E2E, and image jobs otherwise pass.

done, ptal

Alanxtl:
the ci passes, PTAL

Xunzhuo (commented):
Thanks for the update. There is still a memory-safety issue in the new Candle FFI contract: supported_dimensions is built as Vec<usize>, then exposed as *mut i32 and freed as Box<[i32]>. On amd64/arm64 this both corrupts the dimensions read by Go and deallocates with the wrong layout. Please build the exported buffer as Vec<i32> (as the ONNX path does) and add a real FFI round-trip test that covers reading and freeing it.

Alanxtl:
> Thanks for the update. There is still a memory-safety issue in the new Candle FFI contract: supported_dimensions is built as Vec, then exposed as *mut i32 and freed as Box<[i32]>. On amd64/arm64 this both corrupts the dimensions read by Go and deallocates with the wrong layout. Please build the exported buffer as Vec (as the ONNX path does) and add a real FFI round-trip test that covers reading and freeing it.

done

Xunzhuo (commented):
Welcome, and thanks for the thorough follow-up. The current head fixes the dimension-contract ABI by allocating and freeing an `i32` slice that matches the C/Go `int*` contract, and the cache/memory paths now validate existing collection dimensions against the loaded model contract. The exact-head core, image, E2E, Operator, and memory-integration gates are green. I don't see a remaining code blocker. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

WUKUNTAI-0211 (changes requested):
Thanks for updating the branch. Two blockers remain:

1. #2407 explicitly defines only model-declared Matryoshka dimensions as supported, so enforcing the declared list is intentional even where the current inference path can technically truncate to other prefix widths. The compatibility migration is still missing: `config/runtime/response-cache/milvus.yaml:49-50` ships `dimension: 384` and documents it as ignored/auto-detected, while `deploy/helm/semantic-router/values.yaml:345` selects mmBERT for response-cache embeddings. This PR now makes that 384 value authoritative and rejects it. The two E2E fixtures were changed from 384 to 256 rather than covering the existing configuration. Please update the shipped configuration and document the release/migration path for existing 384-dimensional collections.

2. `NewMilvusStore` unconditionally queries the loaded native-model contract before `VLLM_SR_DETERMINISTIC_EMBEDDINGS` can take effect. Deterministic integration mode intentionally skips native model loading, so memory initialization now fails before deterministic embeddings can run. Please preserve this mode and add focused coverage.

Separately, `semanticCacheEmbeddingDimension` converts contract lookup failures to dimension `0` for the remaining Redis, Valkey, and Qdrant callers. Broader backend lifecycle unification can be tracked separately, but this PR should avoid introducing that silent failure.

The earlier Candle FFI allocation/free issue is resolved.

Alanxtl:
> Thanks for updating the branch. Two blockers remain:
> 
> 1. #2407 explicitly defines only model-declared Matryoshka dimensions as supported, so enforcing the declared list is intentional even where the current inference path can technically truncate to other prefix widths. The compatibility migration is still missing: `config/runtime/response-cache/milvus.yaml:49-50` ships `dimension: 384` and documents it as ignored/auto-detected, while `deploy/helm/semantic-router/values.yaml:345` selects mmBERT for response-cache embeddings. This PR now makes that 384 value authoritative and rejects it. The two E2E fixtures were changed from 384 to 256 rather than covering the existing configuration. Please update the shipped configuration and document the release/migration path for existing 384-dimensional collections.
> 2. `NewMilvusStore` unconditionally queries the loaded native-model contract before `VLLM_SR_DETERMINISTIC_EMBEDDINGS` can take effect. Deterministic integration mode intentionally skips native model loading, so memory initialization now fails before deterministic embeddings can run. Please preserve this mode and add focused coverage.
> 
> Separately, `semanticCacheEmbeddingDimension` converts contract lookup failures to dimension `0` for the remaining Redis, Valkey, and Qdrant callers. Broader backend lifecycle unification can be tracked separately, but this PR should avoid introducing that silent failure.
> 
> The earlier Candle FFI allocation/free issue is resolved.

Thanks for the detailed review. I've updated it, PTAL, specifically:

1. Milvus configuration and migration:

   - Removed the misleading `dimension: 384` from `config/runtime/response-cache/milvus.yaml`; an omitted dimension now follows the loaded model’s native dimension.
   - Explicit positive dimensions remain authoritative and must be declared by the model contract.
   - Changed the sample’s `drop_collection_on_startup` default to `false`.
   - Added an upgrade path for existing 384-dimensional response-cache collections in the response-cache and Milvus documentation, including recreation/re-warming guidance and a warning not to drop agentic-memory collections.
   - Restored both memory E2E fixtures to 384 so deterministic integration mode continues covering the existing configuration instead of changing it to 256.

2. Deterministic memory mode:

   - `NewMilvusStore` now skips native contract lookup when `VLLM_SR_DETERMINISTIC_EMBEDDINGS` is enabled.
   - It preserves the configured vector width in that mode.
   - Restored the previous deterministic dimension behavior to avoid introducing another model capability table.
   - Added focused coverage using an unloaded model with a 384-dimensional deterministic embedding, verifying store initialization and embedding generation without a native contract.

3. Other cache backends:

   - `semanticCacheEmbeddingDimension` now returns lookup errors instead of converting them to dimension `0`.
   - Redis, Valkey, Qdrant, and exact Qdrant callers now propagate those errors.

Xunzhuo (dismissed):
Thanks for the comprehensive follow-up. One current-head build blocker remains: `pkg/cache/valkey_cache.go:318` redeclares `err` with `:=` after `createIndex` already declared it at the dimension lookup, so the exact-head core, pre-commit, E2E, and image jobs all stop with “no new variables on left side of :=”. Please assign to the existing `err` (or use a new name) and rerun the gates.

Alanxtl:
> Thanks for the comprehensive follow-up. One current-head build blocker remains: `pkg/cache/valkey_cache.go:318` redeclares `err` with `:=` after `createIndex` already declared it at the dimension lookup, so the exact-head core, pre-commit, E2E, and image jobs all stop with “no new variables on left side of :=”. Please assign to the existing `err` (or use a new name) and rerun the gates.

thanks for pointing it out, already fixed :)

Xunzhuo (changes requested):
Thanks for fixing the compile error. One current-head integration blocker remains: Memory Integration runs e2e/config/config.memory-user.yaml with dimension 384, but mmBERT declares only 768, 512, 256, 128, and 64. The router disables Memory and the exact-head suite fails. Please migrate the maintained memory fixture/config to a model-declared width, preserve the migration guidance, and rerun the gates.

Alanxtl:
> Thanks for fixing the compile error. One current-head integration blocker remains: Memory Integration runs e2e/config/config.memory-user.yaml with dimension 384, but mmBERT declares only 768, 512, 256, 128, and 64. The router disables Memory and the exact-head suite fails. Please migrate the maintained memory fixture/config to a model-declared width, preserve the migration guidance, and rerun the gates.

fixed

Alanxtl:
The Memory Integration, Dashboard checks, core build, E2E, and full pre-commit all pass. The remaining failures are limited to the vllm-sr CUDA and ROCm image jobs. Both fail during apt installation while fetching Debian bullseye-security package libssl1.1_1.1.1w-0+deb11u8 with HTTP 404 from deb.debian.org; no repository source or test failure is reported. @Xunzhuo

Alanxtl:
The latest run 33956115994 confirms the 256-dimension fixture migration is working: Memory Integration no longer reports the mmBERT dimension-contract error. The Memory Integration job now stops while downloading the required mmBERT model because Hugging Face returns HTTP 429 Too Many Requests, with deterministic fallback disabled by CI. The CUDA and ROCm image jobs still fail while fetching Debian bullseye-security libssl1.1_1.1.1w-0+deb11u8 with HTTP 404. Core build, Dashboard, E2E, pre-commit, and the other integration jobs pass. The remaining failures are external registry/package-service availability issues; please rerun these jobs after the rate limit and mirror state recover.

Xunzhuo (commented):
Thanks for the thorough follow-up. The maintained memory fixtures now use the model-declared 256-dimension mmBERT contract, so the previous integration blocker is resolved. I don’t see a new code blocker in this patch. Approval is waiting on a clean required-check rerun; the current Memory and image jobs failed on external downloads.

Alanxtl:
the same external image issue still persists. Memory Integration, core build, Dashboard, E2E, and full pre-commit all pass. Only the vllm-sr CUDA and ROCm image jobs fail: both hit the same Debian bullseye-security download failure @Xunzhuo

WUKUNTAI-0211 (changes requested):
Thanks for all the fixes and for working through the earlier feedback. One #3317 issue remains: `NewMilvusStore` returns the collection dimension mismatch, but `createMemoryRuntime` catches the error and starts the Router with Memory disabled.

Since #3317 says this mismatch should fail startup, could you propagate this specific incompatibility through router initialization and add a focused test? Please keep unrelated Memory connection failures on the existing fail-open path.

Alanxtl:
> Thanks for all the fixes and for working through the earlier feedback. One #3317 issue remains: `NewMilvusStore` returns the collection dimension mismatch, but `createMemoryRuntime` catches the error and starts the Router with Memory disabled.
> 
> Since #3317 says this mismatch should fail startup, could you propagate this specific incompatibility through router initialization and add a focused test? Please keep unrelated Memory connection failures on the existing fail-open path.

done, PTAL

Alanxtl:
@Xunzhuo @WUKUNTAI-0211 @FAUST-BENCHOU , the CI finally all pass, PTAL. If any further comment, please let me know.

WUKUNTAI-0211 (dismissed):
Thanks for working through this one. It lands the startup check #3317 asked for: the effective embedding width now comes from the loaded model contract instead of a hardcoded table, and an existing Milvus collection whose stored dimension does not match is rejected at startup rather than silently starting with Memory disabled. I rechecked the current head and the required checks are green. Ready from my side.

FAUST-BENCHOU (dismissed):

wilsonwu (dismissed):
LGTM

src/semantic-router/pkg/cache/cache.go
```diff
@@ -180,21 +182,25 @@ func normalizeEmbeddingModel(model string) string {
...
-		return 1024
-	case "gemma":
-		return 768
-	case "mmbert":
-		return 768
-	case "multimodal":
-		return 384
-	default:
-		return 384
+	dimension, err := candle_binding.ResolveEmbeddingDimension(
```
Xunzhuo, line 194: **Resolve cache dimensions after embedding initialization**

Impact: On a cold process, an enabled semantic cache with an omitted dimension fails during router construction before its embedding model is initialized; Milvus also fails with an explicit dimension because its constructor always queries the unloaded native contract.

Required change: Initialize the embedding runtime before constructing semantic-cache backends, or defer contract resolution and validation until after classifier runtime initialization while preserving rollback ownership; add a cold-start regression with no preinitialized native model.

Xunzhuo (commented):

src/semantic-router/pkg/cache/cache.go
Alanxtl: fixed, now classifier runtime init first before semantic cache construct

Alanxtl:
blocked by https://github.com/vllm-project/semantic-router/issues/3681

src/semantic-router/pkg/cache/cache.go
Alanxtl: @Xunzhuo PTAL

Alanxtl:
all complete @Xunzhuo please take a look

Xunzhuo (commented):
Thanks for resolving the earlier initialization-order blocker. Two current-head issues still prevent this dimension-contract change from working end to end.

src/semantic-router/pkg/cache/cache.go
```diff
@@ -180,21 +182,34 @@ func normalizeEmbeddingModel(model string) string {
 	return normalized
 }
 
-func semanticCacheEmbeddingDimension(configured int, embeddingModel string) int {
+func semanticCacheEmbeddingDimension(provider embedding.Provider, configured int, embeddingModel string) (int, error) {
```
Xunzhuo, line 186: **Restore the embedding package import**

Impact: this new signature references `embedding.Provider` and `DimensionContractProvider` without importing `pkg/embedding`, so exact-head Core CI fails with `undefined: embedding` and the dependent images and integration jobs cannot build.

Required change: add the missing import, run gofumpt (which also currently flags `candle-binding/instances.go:179`), and rerun the exact-head gates.

src/semantic-router/pkg/memory/milvus_store.go
```diff
@@ -105,12 +126,50 @@ func NewMilvusStore(options MilvusStoreOptions) (*MilvusStore, error) {
...
+		if embeddingCfg.Provider != nil {
+			contractProvider, ok := embeddingCfg.Provider.(embedding.DimensionContractProvider)
+			if !ok {
+				return 0, fmt.Errorf("prepared embedding provider does not expose a dimension contract")
+			}
+			contract, err := contractProvider.EmbeddingDimensionContract()
+			if err != nil {
+				return 0, fmt.Errorf("failed to get embedding dimension contract: %w", err)
+			}
+			return contract.Resolve(embeddingCfg.Dimension)
```
Xunzhuo, line 149: **Apply the resolved dimension during memory inference**

Impact: `contract.Resolve` accepts reduced Qwen3/Gemma widths and uses that width for the Milvus schema, but `GenerateEmbeddingWithContext` passes empty options for those models and still produces native-width vectors. A reduced-width configuration therefore passes startup and then fails inserts and searches.

Required change: either pass the resolved width into inference or reject non-native widths, with an end-to-end reduced-width test.

src/semantic-router/pkg/memory/embedding.go
```diff
@@ -48,7 +48,10 @@ func GenerateEmbeddingWithContext(ctx context.Context, text string, cfg Embeddin
 	options := embedding.Options{}
 	switch modelName {
 	case "qwen3", "gemma", "bert", "":
-		// These paths have always requested the provider's full output.
+		// The store resolves Dimension from the model contract before inference.
+		// Pass that width through so reduced Qwen3/Gemma vectors match the
+		// Milvus schema instead of silently returning the native width.
+		options.Dimension = cfg.Dimension
```
Xunzhuo, line 54: **Align the provider option contract**

Impact: This now sends `Dimension` for BERT, Qwen3, and Gemma, while `TestOwnedMemoryProviderPreservesModelOptionsAndRawVector` still requires empty options for all three; exact-head Core CI fails on the first BERT case. The branch therefore does not currently establish whether native-width BERT behavior is preserved while reduced Qwen3/Gemma widths are forwarded.

Required change: Encode one explicit contract—either scope dimension pass-through to the models that support reduced widths or update the maintained expectations—and cover native BERT plus reduced Qwen3/Gemma before rerunning the gates.

Xunzhuo (commented):
Thanks for wiring the resolved width into inference; the two earlier blockers are fixed. The exact-head Core test still fails on the resulting option contract, and pre-commit also flags the two changed Go files for gofumpt.

Alanxtl:
all complete @Xunzhuo please take a look

src/semantic-router/pkg/cache/redis_cache.go
```diff
@@ -239,18 +248,21 @@ func (c *RedisCache) getEmbedding(ctx context.Context, text string) ([]float32,
 	return computeCacheEmbedding(ctx, c.embeddingProvider, text)
 }
 
-func (c *RedisCache) embeddingDimension() int {
+func (c *RedisCache) embeddingDimension() (int, error) {
```
Xunzhuo, line 251: **Update the cache identity callers for the new return signature**

`LocalEmbeddingSettings` in `pkg/cache/embedding_identity.go:28,30,34` still uses the Redis, Valkey, and Qdrant `embeddingDimension()` calls as single integers. These methods now return `(int, error)`, so the current Core check fails with `multiple-value value.embeddingDimension() ... in single-value context`; the Memory image build fails on the same three calls before its tests run. Update those callers to consume the resolved dimension and handle errors, and cover the cache identity/namespace path before rerunning Core and Memory integration.

Xunzhuo (changes requested):
The dimension-contract merge leaves one build blocker: the cache identity code still uses the old single-value dimension API. The current-head Core and Memory checks both report the same compiler errors, so Memory integration does not reach its runtime tests. Please update the callers and error handling, cover the cache identity path, and rerun the affected checks.

src/semantic-router/pkg/cache/milvus_cache_collection.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if c.config == nil {
+		return 0
+	}
+	dimension, err := resolveMilvusCacheEmbeddingDimension(
+		c.embeddingProvider,
+		c.embeddingModel,
+		c.config.Collection.VectorField.Dimension,
+	)
+	if err != nil {
```
Xunzhuo, line 100: [P2] Preserve the prepared dimension when naming Milvus collections

`PrepareEmbeddingNamespace` builds its Milvus/Hybrid view without `cfg.EmbeddingProvider`, so a generation-owned mmBERT provider can reach this global-model lookup before any global factory exists. Returning zero hides that error and makes the identity resolver use the native dimension, although `NewMilvusCache` later resolves the configured width from the supplied provider. With a prepared provider supporting 256/512/768, I verified that 256 and 512 both receive the same physical collection name while their constructor dimensions differ. A dimension change then reopens the old fixed-width collection and fails validation. Forward the prepared provider into both namespace views and propagate resolution errors instead of converting them to the default dimension.

src/semantic-router/pkg/cache/embedding_identity_test.go
```diff
@@ -105,16 +109,60 @@ func TestEmbeddingIdentityIsolatesPersistentSemanticCacheWithoutDeletingLegacy(t
...
-		t.Fatalf("inmemory actual settings: %#v %v", settings, ok)
+	settings, ok, err := LocalEmbeddingSettings(memory)
+	if err != nil || !ok || settings.Layer != 6 || settings.Dimension != 256 {
+		t.Fatalf("inmemory actual settings: %#v %v %v", settings, ok, err)
 	}
 	persistent := &QdrantCache{embeddingModel: "mmbert"}
-	settings, ok = LocalEmbeddingSettings(persistent)
-	if !ok || settings.Layer != 0 || settings.Dimension != 768 {
-		t.Fatalf("persistent actual settings: %#v %v", settings, ok)
+	settings, ok, err = LocalEmbeddingSettings(persistent)
```
Xunzhuo, line 117: **[P2] Give the identity fixtures a prepared dimension provider**

This fixture creates a Qdrant cache with only `embeddingModel: "mmbert"`, but the new error-returning dimension path queries the loaded native contract when no provider is supplied. In a fresh process, `TestCacheEmbeddingSettingsReflectActualBackend` now fails here with `embedding model factory is not initialized` instead of the expected 768; the Qdrant namespace fixture fails for the same reason. Inject a prepared provider into these contract-only fixtures (and the namespace matrix) so the tests cover resolved dimensions without depending on another test loading a global model.

Xunzhuo (changes requested):
The cache identity callers now compile with the error-returning API. Two issues remain: Milvus/Hybrid namespace preparation loses the prepared provider and collapses distinct widths onto one physical collection, and the providerless identity fixtures fail in a fresh process. Both failures reproduce on this updated head with the branch native binding. Focused cache, memory, Milvus lifecycle, and configuration checks were also inspected or run; no live Milvus server or full serve stack was exercised.

src/semantic-router/pkg/cache/embedding_identity.go
```diff
@@ -18,32 +18,47 @@ const (
...
+		var err error
+		dimension, err = value.embeddingDimension()
+		if err != nil {
+			return embedding.ConsumerSettings{}, false, fmt.Errorf("resolve redis cache embedding dimension: %w", err)
+		}
 	case *ValkeyCache:
-		model, dimension = value.embeddingModel, value.embeddingDimension()
+		model = value.embeddingModel
+		var err error
+		dimension, err = value.embeddingDimension()
```
Xunzhuo, line 37: **[P1] Update the RISC-V Valkey implementation for the new contract**

Current main now supplies `valkey_cache_unavailable.go` on RISC-V. Its `embeddingDimension()` still returns one integer and calls `semanticCacheEmbeddingDimension` with two arguments, while this assignment expects two results and the shared helper now also requires a provider. The clean integration fails to compile for RISC-V even when Valkey is disabled. Update the unavailable implementation alongside the native one and run the maintained RISC-V build/QEMU gate.

src/semantic-router/pkg/cache/milvus_cache.go
```diff
@@ -72,6 +76,16 @@ func NewMilvusCache(options MilvusCacheOptions) (*MilvusCache, error) {
...
 		milvusConfig = options.Config
 	}
+	embeddingModel := normalizeEmbeddingModel(options.EmbeddingModel)
+	effectiveDimension, err := resolveMilvusCacheEmbeddingDimension(
+		options.EmbeddingProvider,
+		embeddingModel,
+		milvusConfig.Collection.VectorField.Dimension,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Milvus cache embedding dimension: %w", err)
```
Xunzhuo, line 86: **[P1] Preserve model-free startup for exact-only caches**

For reachable `response_cache` plugins using only exact mode, `ResponseCacheDemand()` returns `(true, false)`, so startup intentionally prepares no embedding provider but still opens the cache store. This new Milvus resolver queries the loaded native model contract even when a storage dimension is configured; Qdrant now does the same before opening its client. With no other model consumer, both return an unloaded-model error and abort startup. Carry the semantic-demand distinction into initialization so exact-only storage retains its schema width without requiring an embedding model, and cover startup with no model loaded.

Xunzhuo (changes requested):
Two additional blockers need fixes: integrating current main breaks the RISC-V Valkey fallback signature, and the new dimension lookup makes exact-only Milvus/Qdrant caches require an embedding model that their startup path intentionally does not load. The RISC-V mismatch is confirmed by a partial cross-compilation probe; the exact-only failure follows the demand-to-constructor call path. The earlier namespace and fixture findings remain open. Full CGO/QEMU, live cache services and a deployed startup run were not exercised.

src/semantic-router/pkg/cache/cache.go
Alanxtl: done

src/semantic-router/pkg/memory/milvus_store.go
Alanxtl: done

src/semantic-router/pkg/memory/embedding.go
Alanxtl: done

src/semantic-router/pkg/cache/milvus_cache_collection.go
Alanxtl: done

src/semantic-router/pkg/cache/embedding_identity_test.go
Alanxtl: done

src/semantic-router/pkg/cache/embedding_identity.go
Alanxtl: done

src/semantic-router/pkg/cache/milvus_cache.go
Alanxtl: done

src/semantic-router/pkg/cache/valkey_cache.go
```diff
@@ -174,8 +195,25 @@ func (c *ValkeyCache) initializeIndex() error {
 			"reason":  "development_mode",
 		})
 	}
+	if indexExists && c.config.Index.VectorField.Dimension > 0 {
+		storedDimension, dimensionErr := valkeyIndexVectorDimension(info, c.config.Index.VectorField.Name)
```
Xunzhuo, line 199: [P1] Parse Valkey Search index.dimensions before rejecting an existing index

This new restart check rejects the metadata returned by the supported Valkey Search server. Its [FT.INFO response](https://valkey.io/commands/ft.info/) places the vector width under the field's `index` array as `dimensions`; both this parser and the new memory parser only recognize a flat `dim` next to `type: VECTOR`. I reproduced this against the official arm64 `valkey/valkey-bundle:unstable` server/Search module 1.3.0: at this head, `NewValkeyCache` and `memory.NewValkeyStore` each create a 384-dimensional index successfully, then fail when called again with the identical configuration (`no positive dimension` / `no positive vector DIM`). Thus a valid persisted index prevents cache initialization on restart, and the memory constructor also fails. Please parse the nested `index.dimensions` while retaining the owning field name/type, and cover create-then-reopen against an actual Valkey Search response in both packages.

Xunzhuo (changes requested):
Reviewed the current patch and the affected embedding preparation, cache identity, memory startup and native dimension-contract callers. The prior import/caller-signature, prepared-provider, reduced-width and exact-only initialization issues have corresponding fixes; one new Valkey restart blocker remains below.

Validation on this exact head in credential-free Linux arm64 containers: `make rust-ci` built Candle, ML, NLP and ONNX; `make impact` and `make config-schema-check` passed; all 52 selected dimension/cache/memory/startup tests passed, alongside the full config and embedding package suites (one existing Helm-dependent skip). Real Valkey Search create/reopen reproduced the blocker in both production constructors. Full router E2E, external Milvus/Qdrant/Redis services, large-model dimension inference were not run; current main has content conflicts, so no merged-runtime pass is claimed.

Issue #3182, opened 2026-08-31
[Test] Cover cross-surface routing pipelines and failure semantics end to end


Xunzhuo:
### Goal

Protect representative complete routing pipelines so individually covered signals, projections, algorithms, plugins, and learning controls are also verified at their real composition boundaries.

Parent Epic: #1519

### Proposed Workgroup

Evaluation & Quality

### Scope

- Define a risk-based composition matrix for `signal -> projection -> decision -> algorithm -> plugin -> learning -> backend` rather than an exhaustive Cartesian product.
- Cover recipe and entrypoint isolation, decision priority/tier behavior, selector and Looper boundaries, plugin short-circuiting, learning apply/observe/bypass, and final backend dispatch.
- Include representative buffered and streaming requests, protocol translation, backend errors, cancellation, retry/fallback, and typed unsupported-capability behavior.
- Assert consistency among the public response, selected backend, Router Replay/debug evidence, usage, and failure diagnostics.
- Add this work after or alongside the focused family coverage so cross-surface failures can be assigned to the layer that owns them.
- File each implementation or API defect under its actual Workgroup and keep this task focused on the integrated regression contract.

### Non-goals

- Test every possible combination of registered surfaces.
- Replace focused family, unit, integration, performance, or quality evaluation.
- Hide an unclear ownership boundary inside a broad E2E pull request.
- Require private infrastructure or external credentials for the maintained baseline.

### Completion signal

- The coverage matrix identifies and runs a bounded set of high-risk complete pipelines.
- Success, short-circuit, unsupported capability, backend failure, streaming termination, cancellation, and fallback are represented where applicable.
- Response, routing, replay/debug, and usage facts agree across each maintained scenario.
- Cross-layer bugs found by the work have focused owning issues and regression coverage.

rishabhsinha17:
Claimed above. Posting the plan before code so the slicing can be corrected early.

I audited the existing baseline coverage against the composition matrix in the scope. Keeping selector algorithms out (#3179 owns those) and Looper out (#2859 owns those).

## What is already covered

Decision priority, plugin chain execution, fallback, entrypoint recipe routing, retention directives, and per-protocol request/response shape all have baseline cases. The `streaming` profile covers streamed request bodies including `fast_response` blocking.

## The gap this slice targets

[...]

PR #3356, opened 2026-09-02
[Test] Prove guardrail short-circuit does not dispatch upstream

rishabhsinha17:
## Summary

First slice of the cross-surface coverage in #3182, scoped to one contract: a `fast_response` guardrail must answer the client without dispatching to the upstream backend.

The existing guardrail cases assert that the client received a fast response (`e2e/testcases/streaming_safety.go:60-67`), but nothing asserts the other half of a short-circuit. A regression that blocked the client while still dispatching upstream, or that dispatched and then discarded the response, passes every case we have today.

The simulator already has the seam this needs: `GET /debug/last-request` returns the last body it saw for an `x-vsr-test-session-id` and 404s when that session never reached it (`tools/mock-vllm/provider_boundary.py:11,72-86`). No profile or simulator change is required.

## Changes

- Add `plugin-short-circuit-no-dispatch` with three sub-cases on the existing `envoy-ai-gateway` baseline, each with its own session id:
  - control, first: asserts `x-vsr-response-path: upstream`, a non-empty `x-vsr-selected-model`, and that the simulator saw the request with an upstream model equal to that header.
  - jailbreak: `x-vsr-fast-response: true`, `x-vsr-response-path: fast_response`, `x-vsr-selected-decision: block_jailbreak`, and the simulator returns 404.
  - PII: the same shape against `block_pii`, simulator 404.
- Add `fixtures.OpenServiceEndpointSession` so a testcase can port-forward to a dependency it needs to observe, rather than only the profile service.
- Register the case in `BaselineRouterContract` and in the curated `ENVOY_AI_GATEWAY_CI_TESTS` list, since cases outside that list do not run in PR CI.

The control sub-case is load-bearing rather than decorative. Without it a 404 could mean the router dropped the session header rather than that no dispatch happened, and both negative assertions would pass for the wrong reason. It runs first and fails with an explicit message saying the negative assertions cannot be trusted.

The guardrail prompts are reused verbatim from the existing guardrail cases so both tests move together. The control prompt is digit-free, since the profile's deny-all PII signal tags long digit strings as ID-like entities (the first CI run of this case proved that by blocking its own control), and it classifies to `biology_decision`, which has no `response_cache` plugin, so a repeat run against a long-lived cluster cannot be intercepted by the semantic cache. The per-run id now scopes only the simulator session ids, which no classifier sees.

## Validation

- `go build ./...` and `go vet ./...` in the `e2e` module: clean.
- `gofmt -l`: clean.
- `golangci-lint run ./testcases/...` with `tools/linter/go/.golangci.yml`, built against Go 1.25.0 to match CI: no findings in the added file. Pre-existing findings elsewhere in the module are untouched.
- `tools/ci/workflow_policy_validation.py`: passes.
- `tools/ci/classify_pr_changes.py --base origin/main --head HEAD`: selects the `envoy-ai-gateway` profile, so the new case runs on this PR.
- `go test ./pkg/testcases/...`: passes.

The E2E case itself is validated by this PR's own `envoy-ai-gateway` run; a full Kind profile is not reproducible on my machine.

Related #3182


FAUST-BENCHOU (dismissed):
LGTM. The change keeps the scope focused and improves the current behavior without introducing additional complexity.

wilsonwu (dismissed):
LGTM, thanks!

rishabhsinha17:
CI is green on c024f4be. What the red run was:

The only real failure was the envoy-ai-gateway E2E leg. The seven other red checks (Lint, PR Gate, pre-commit file lint, Unit Tests, test-and-build, Verify Manifests, Validate OLM Bundle) are domain gate jobs reporting "Affected PR domain failed or was cancelled, e2e=failure", all cascading from that one leg.

The E2E failure was the new case blocking itself. The control prompt embedded a UnixNano run id to stay out of the semantic cache, and this profile's deny-all PII signal tags a long digit string after "id" as an ID-like entity, so the control came back as a block_pii fast_response instead of routing upstream.

Fix in c024f4be:

- The control prompt is now digit-free and classifies to biology_decision, which has no response_cache plugin, so the cache cannot intercept a repeat run and the run id is only needed for the simulator session ids, which no classifier sees.
- Fixed an err shadow in runShortCircuitProbe flagged by the repo govet config.

The rerun proves the full contract on a live cluster: control routes with response_path=upstream and the simulator observes the request; jailbreak and pii return response_path=fast_response and the simulator answers 404 for their sessions.

Xunzhuo (dismissed):
The control probe makes the negative no-dispatch assertions trustworthy, and the case is wired into both the baseline contract and the curated CI shard. I also checked the observable backend session helper and the green rerun after the control-prompt fix; local agent validation passes. Approved; please rebase onto current main before merge.

FAUST-BENCHOU (dismissed):

wilsonwu (dismissed):
LGTM

rishabhsinha17:
Rebased onto main to clear the queue's blocked branch update. Mergify could not push the update because the branch carries a workflow file change and the app lacks workflows permission, same situation as #3360. The only conflict was the CI test list in `.github/workflows/integration-test-k8s.yml`; resolved by keeping both sides, `event-routing` from main plus this PR's `plugin-short-circuit-no-dispatch`. No other content changed. The force-push dismissed the three approvals, so once checks are green this needs a re-approval to re-enter the queue.

rishabhsinha17:
Merged current main into the branch to clear a new conflict in the CI test list, keeping main's pii-entity-offsets, pii-long-text, and security-long-text entries plus this PR's plugin-short-circuit-no-dispatch. No other content changed. The 09-05 rebase was needed because Mergify cannot update branches that carry workflow file changes, and that push is what dismissed the approvals. The reviewed test content is unchanged since then, so one renewed approval should let this re-enter the merge queue. @Xunzhuo @FAUST-BENCHOU @wilsonwu

Xunzhuo (approved):
The no-dispatch contract remains unchanged after the main merge, and the exact-head Envoy gateway E2E proves the control reaches upstream while both guardrail cases do not. The helper, test registration, and current integration are clean. Thanks for the careful coverage.

Issue #3233, opened 2026-08-31
[Test] Cover Dashboard and CLI user journeys with behavioral E2E contracts


Xunzhuo:
### Goal

Protect the maintained Dashboard and CLI user journeys with behavioral end-to-end contracts rather than HTTP-success or configuration-load smoke tests.

### Proposed Workgroup

Evaluation & Quality

### Scope

- Cover first-run setup, configuration editing and validation, deployment, status and diagnostics, evaluation, rollback, and safe failure.
- Exercise canonical Router and deployment contracts through both Dashboard and CLI surfaces.
- Reuse bounded deterministic fixtures and record profile ownership, prerequisites, and failure diagnostics.
- Open focused implementation-owner issues for product or API defects discovered by the tests.

### Non-goals

- Duplicating frontend unit tests or every CLI argument permutation.
- Treating obsolete or unreachable UI surfaces as supported behavior.

### Completion signal

Every maintained Dashboard and CLI journey maps to a runnable E2E contract or an explicit owned gap.

Parent Epic: #1519

rishabhsinha17:
Claimed above. Plan and a first slice, which is up as #3360.

## Where the current coverage stands

The `dashboard` profile covers health, status, config read, deploy preview, config versions, evaluation, and restart recovery. Measured against the journeys in the scope, the gaps are:

- safe failure: no case asserts that a rejected deploy leaves the serving config untouched. `dashboard-deploy-invalid-yaml` asserts only a 400 status.
- rollback: `POST /api/router/config/rollback` is registered (`dashboard/backend/router/core_routes.go:123`) and never exercised end to end.
- validation diagnostics: `config_validation_error` has no test anywhere in the repo.
- negative auth: no case asserts a mutating endpoint rejects an anonymous caller.
- diagnostics: `GET /api/logs` is not covered.

## Slices, in the order I would take them

1. Safe failure, up now as #3360.
2. Deploy then rollback, asserting the config comes back byte-for-byte and the version list grows by exactly one.
3. Status and diagnostics as behavior rather than HTTP success.

## One finding from slice 1 that affects the rest

The natural payload for a safe-failure case does the opposite of what it looks like. The unknown-model check only runs when the merged document declares a model surface (`src/semantic-router/pkg/config/canonical_config.go:264-272`), and `e2e/profiles/dashboard/values.yaml` is still pre-v0.3 and declares no `routing.modelCards`.

Reproduced locally against a base mirroring that fixture:

[...]

PR #3360, opened 2026-09-02
[Test] Prove a rejected deploy leaves the active config serving

rishabhsinha17:
## Summary

First slice of #3233, scoped to one journey: safe failure on deploy.

The dashboard already has a negative deploy case, but it asserts only that malformed YAML returns 400 (`e2e/testcases/dashboard_deploy_invalid_yaml.go:63-65`). Nothing asserts what an operator actually depends on: that a deploy the router refuses leaves the config that is currently serving completely untouched.

`deployDirectWrite` validates the merged document before it creates a backup or writes (`dashboard/backend/handlers/deploy.go:254-271`), so a semantically invalid deploy must change neither the active config nor the backup history. That invariant had no coverage at either level, and `config_validation_error` had no test anywhere.

## Changes

- `dashboard-deploy-safe-failure` E2E: sends a fragment that parses but cannot validate, then asserts
  - the rejection is `config_validation_error`, not `yaml_parse_error`, so validation actually ran,
  - `GET /api/router/config/yaml` is byte-for-byte unchanged,
  - no backup version was created,
  - the same request without an `Authorization` header is rejected with 401.
- A handler-level regression pinning the same invariant deterministically, including that the rejection names the unknown model.

## A finding worth flagging

The obvious payload for this case is wrong, and quietly so. The unknown-model check only runs when the merged document declares a model surface (`src/semantic-router/pkg/config/canonical_config.go:264-272`), and the dashboard profile's router config (`e2e/profiles/dashboard/values.yaml`) still uses the pre-v0.3 schema and declares no `routing.modelCards`.

Reproduced locally against a base mirroring that fixture:

- fragment without its own `modelCards`: `200`, and the active config **was rewritten**.
- fragment carrying its own `modelCards`: `400 config_validation_error`, config unchanged.

So the first form would have silently rewritten the deployed config mid-suite and asserted nothing. The fragment here carries its own `modelCards` for that reason, and the comment in the file explains it so the next person does not "simplify" it back.

That fixture drift is worth its own issue: because the profile config is pre-v0.3, the dashboard profile never exercises the canonical schema, which is also why `dashboard-config-read` asserts chart placeholders. I did not fix it here to keep this PR to one contract. Happy to file it if you agree.

## Validation

- `go build ./...` and `go vet ./...` in both the `e2e` and `dashboard/backend` modules: clean.
- `gofmt -l`: clean.
- `golangci-lint run` with `tools/linter/go/.golangci.yml`, built against Go 1.25.0 to match CI: no findings in either added file.
- `go test ./handlers/` in `dashboard/backend`: passes, including the new regression.
- `tools/ci/classify_pr_changes.py --base origin/main --head HEAD`: selects the `dashboard` profile, so the new E2E case runs on this PR.

The E2E case itself is validated by this PR's `dashboard` profile run; a full Kind profile is not reproducible on my machine.

Related #3233

### DIFF
--- a/dashboard/backend/handlers/deploy_safe_failure_test.go
+++ b/dashboard/backend/handlers/deploy_safe_failure_test.go
@@ -1,0 +1,85 @@
+package handlers
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+)
+
+// safeFailureFragment references a model the merged config does not define.
+// The YAML is syntactically valid, so it clears the parse stage and can only be
+// rejected by config validation.
+const safeFailureFragment = `routing:
+  decisions:
+    - name: e2e-safe-failure
+      description: Deploy probe that must be rejected by config validation
+      priority: 3
+      rules:
+        operator: OR
+        conditions:
+          - type: domain
+            name: business
+      modelRefs:
+        - model: e2e-nonexistent-model
+          use_reasoning: false
+`
+
+// TestDeployHandler_SemanticRejectionLeavesConfigUnchanged pins the safe-failure
+// half of the deploy contract: a config that parses but does not validate must
+// be rejected before the active config is touched. deployDirectWrite validates
+// the merged document before it creates a backup or writes, so a rejected
+// deploy must leave the serving config byte-for-byte identical.
+func TestDeployHandler_SemanticRejectionLeavesConfigUnchanged(t *testing.T) {
+	tempDir := t.TempDir()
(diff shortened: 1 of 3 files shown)


FAUST-BENCHOU (approved):
LGTM. The implementation looks consistent with the existing architecture and the added checks improve reliability.

wilsonwu (approved):
LGTM, thanks!

FAUST-BENCHOU (approved):
LGTM. Nice to see the regression coverage added here. Verifying that rejected deployments keep serving the existing active config makes the failure behavior much clearer and safer.

PR #3332, opened 2026-09-02
[Test] Failing test for missing include_usage on replay-eligible streaming dispatch

dorota-memex:
Related #3331

## Purpose

This PR is a draft that documents the bug in #3331 and gives whoever picks it up a red test to start from after issue is accepted.

## Test Plan

Two tests in `pkg/extproc`, both running a request through `prepareProtocolRequest` and `encodeDispatchRequest`, the same code production dispatch uses.

- `TestExtProcDispatchForcesIncludeUsageOnReplayEligibleStream`: a streaming Chat request without `stream_options` stays replay-eligible, and the dispatch body reaches the backend without the forced `include_usage`. Fails on main by design; this is the bug.
- `TestExtProcDispatchPreservesReplayWhenClientRequestsUsage`: when the client already asked for usage, byte replay is preserved. Passes today, and guards the optimization for any future fix.

The failing test also pins two side contracts: the dispatch body stays valid Chat Completions JSON (it is decoded back through the codec engine), and forcing the flag never leaks into the retained client preference.

## Test Result

`go test ./pkg/extproc/` on this branch: the first test fails (expected red until a fix lands), the second passes, the rest of the suite is unaffected.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title begins with exactly one bracketed category, such as `[Feature]`, `[Bug]`, `[Docs]`, `[Test]`, `[Research]`, `[Community]`, or `[CI/Build]`
- [x] The title does not stack prefixes such as `[Router][Docs]`; affected modules belong in labels and the PR body
- [ ] The PR links an `accepted` issue with exactly one owner: one `wg/*` label for project work or `owner/maintainers` for repository governance
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>

See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full contributor workflow and commit guidance.

PR #3368, opened 2026-09-02
[Bug] Retire the replay claim when forcing include_usage on dispatch

### LINKED ISSUE #3331: [Bug] Byte-replay bypasses forced include_usage; streamed usage goes unrecorded
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The router forces `stream_options.include_usage: true` on streaming Chat Completions requests so it can observe authoritative token usage. The flag is set on a copy of the request without bumping its `Generation` counter. When nothing else has modified the request, the byte-replay optimization concludes the original client bytes can be forwarded unchanged, and the flag never reaches the backend. The backend then emits no usage chunk, and the router records no usage for the request. No warning or fallback is logged.

This happens on ordinary traffic. A request replays when nothing mutates it: the requested model name equals the upstream model id, the stream flag already matches, no memory injection or context compression applies, the body fits the envelope bound, and the formats match. An OpenAI SDK streaming call with a pinned model and default settings qualifies, since SDKs do not set `stream_options` unless asked.

Code path: the injection is in `src/semantic-router/pkg/extproc/processor_protocol_contract.go:184-191`; `EncodeRequest` returns the original bytes when `CanReplay` passes (`pkg/protocolcodec/codec_openai_chat_encode.go:11-13`, `pkg/llmprotocol/policy.go:139`). The comment above the injection says "The Router always asks Chat backends for the final usage chunk so accounting observes authoritative tokens"; the replay path falsifies "always".

### To Reproduce

1. Configure a model whose public name equals its upstream model id, backed by an OpenAI-format backend. Enable streaming SSE logging on the backend (or use a mock that logs request bodies verbatim).
2. Send `POST /v1/chat/completions` with `"stream": true`, no `stream_options`, and a distinctive formatting quirk in the JSON (irregular spacing, unusual key order).
3. The backend receives the body byte-identical, quirk preserved, proving replay; `stream_options` is absent; the backend emits no usage chunk.
4. Router side: no `llm_usage` log line for the request, and the `llm_model_prompt_tokens_total` / `llm_model_completion_tokens_total` series for the model are never populated. No warning is logged.
5. Control: repeat with a model alias that maps to a different upstream id. The model rewrite invalidates replay, the re-encoded body carries `"stream_options":{"include_usage":true}`, the usage chunk flows, and `llm_usage` plus the token metrics appear.

### Expected behavior

Streaming Chat dispatch always requests the final usage chunk, as the code comment states, regardless of whether the body is replayed or re-encoded. Usage settlement then records authoritative tokens for replayed requests exactly as for re-encoded ones.

The client-facing side needs no change: the router already strips the forced usage chunk from the client stream when the client did not ask for it (verified in the control), so fixing the injection does not leak anything to clients.

### Impact

Token accounting and cost metrics silently undercount. Requests affected by this bug meter zero tokens: the per-model token counters on the router's metrics endpoint stay empty, `llm_usage` records are missing, and anything built on them (cost dashboards, pricing settlement, capacity planning) sees a gap with no signal that data is missing. Clients are unaffected, which is why the loss is easy to miss.

### Additional context

Fix shape: force the flag and bump `Generation` only when the request does not already carry `stream_options.include_usage: true`. When the client already asked for usage, nothing changes and byte-replay stays valid, since the flag is in the original bytes. When the injection is actually needed, the bump invalidates replay and the re-encoded body carries the flag.

Verified on `main` @ af9437b1 (2026-09-02, router built at that commit) through a live Envoy+extproc stack with a logging mock backend.

rk9595:
Closes #3331

## Purpose

`encodeDispatchRequest` (`src/semantic-router/pkg/extproc/processor_protocol_contract.go`) forces `stream_options.include_usage` on a copy of the neutral request so accounting can observe authoritative tokens, but it never bumped `Generation`. The source envelope therefore still matched, `CanReplay` (`pkg/llmprotocol/policy.go:141`) held, and `EncodeRequest` (`pkg/protocolcodec/codec_openai_chat_encode.go:11-13`) forwarded the original client bytes. The flag never reached the backend, the backend emitted no usage chunk, the Router recorded no usage, and nothing was logged.

This hits ordinary traffic: any streaming Chat request that nothing else mutates — a pinned model whose public name equals its upstream id, default SDK settings, no `stream_options` — stays replay-eligible.

The fix bumps `Generation` when the injection actually changes the effective value, retiring the replay claim on bytes that no longer describe the dispatch request.

**The bump is deliberately conditional.** `Generation++` is the established idiom for a request mutation here (`req_filter_tools.go:59,64,85,98`, `req_filter_cache.go:394`), but doing it unconditionally would regress requests that already carry `include_usage`: they replay correctly today, and bumping their generation drops them out of the replay path and silently strips preserved unknown fields under `UnknownPreserveSameFormat` — the fidelity contract the envelope exists to hold. This matches the fix shape described in the issue's Additional context.

Owner: `wg/data-plane-networking` (issue #3331, `accepted`, assigned).

## Test Plan

Tests are taken unchanged from @dorota-memex's draft #3332, which pinned this bug before the fix existed; the commit carries a `Co-authored-by` trailer. Both drive the real `prepareProtocolRequest` → `encodeDispatchRequest` path:

- `TestExtProcDispatchForcesIncludeUsageOnReplayEligibleStream` — red on `main` by design
- `TestExtProcDispatchPreservesReplayWhenClientRequestsUsage` — green on `main`, guards the replay optimization against exactly the unconditional-bump regression

Commands:

- `go test ./pkg/extproc/` and the two tests individually, with the fix reverted and re-applied
- `make test-semantic-router`
- `make agent-dev ENV=cpu` → `make agent-serve-local ENV=cpu` → `make agent-smoke-local`
- Live stack driven with a logging mock backend on the smoke config's `backend_refs` endpoint (`host.docker.internal:8000`); `smoke-model` there has `provider_model_id: "smoke-model"`, i.e. the replay-eligible shape

## Test Result

Red-then-green confirmed by reverting `processor_protocol_contract.go` to `main` and re-running:

```
--- FAIL: TestExtProcDispatchForcesIncludeUsageOnReplayEligibleStream
    dispatch dropped the forced usage request on replay:
    {"model":"public-model","messages":[{"role":"user","content":"hello"}],"stream":true}
```

`TestExtProcDispatchPreservesReplayWhenClientRequestsUsage` passes on `main` too, so it is a real guard rather than a test that only passes with the change.

Through the live Envoy + ExtProc stack:

Client omits `stream_options` (previously silently unrecorded):

```
backend received: {"model":"smoke-model","messages":[...],"stream":true,"stream_options":{"include_usage":true}}
client stream:    no usage chunk (client preference untouched)
```

Client already asked for usage:

```
backend received: {"model":"smoke-model","stream":true,"stream_options":{"include_usage":true},"messages":[...]}
client stream:    "usage": {"prompt_tokens":7,"completion_tokens":2,"total_tokens":9}
```

The field ordering is the evidence: the first body comes back in the encoder's normalized order because it was re-encoded, the second preserves the client's original ordering because byte replay survived. `llm_completion_tokens_per_request{model="smoke-model"}` recorded both requests.

Gates: `make test-semantic-router` exit 0; `make agent-smoke-local` reports "Local smoke checks passed"; `gofmt` clean; `golangci-lint run ./pkg/extproc/...` reports the same 14 pre-existing issues as `main` with none in the changed files.

Remaining risk: the live before/after was captured green-with-fix only — the red side is the unit test plus the reporter's own live verification at `af9437b1` in the issue, not a rebuilt image on `main`. Non-Chat target formats are untouched, as is the retained client preference used for public stream rendering.


dorota-memex (commented):
Looks great! Thanks for quick turnaround on preparing a  fix for the issue.

FAUST-BENCHOU (approved):
LGTM. Making the generation bump conditional preserves replay when the client already requested usage while correctly invalidating stale source bytes when the router mutates the dispatch request.

wilsonwu (approved):
LGTM

rk9595:
The `anthropic-shim` E2E job failed on the latest run, but it looks like infrastructure rather than this change — @wilsonwu could someone with write access rerun it?

From the job's `test-report.json`: `total_tests: 0, passed_tests: 0, failed_tests: 0` with `exit_code: 1`. No tests executed. The timeline from the collected pod logs:

- harness started `12:44:48Z`
- the `semantic-router` container didn't start until `12:58:11Z` — image pull plus an unauthenticated HF model download (`Warning: You are sending unauthenticated requests to the HF Hub`)
- `startup_complete` at `12:58:47Z`
- harness gave up at `12:59:55Z`, 68s later

So the readiness wait expired while the model download was still in flight. The same job passed on both previous runs of this PR (20m12s, then again on the merged head), and the diff here only touches `encodeDispatchRequest` in `pkg/extproc`. Might be worth a longer readiness budget or a cached/authenticated model pull for that profile — it looks like it would flake for anyone whose runner is slow on the pull.

The other seven red checks are all the `Preserve stable merge-gate context` aggregate step reporting downstream of this one.

PR #3369, opened 2026-09-02
[Bug] Scan the whole text in the PII classification API

### LINKED ISSUE #3364: [Bug] PII API drops entities past the first 512 tokens
### Proposed Workgroup

### Describe the bug

`/api/v1/classify/pii` calls `ClassifyPIIWithDetailsAndThreshold` (`classifier_pii_ops.go:77`), one `ClassifyTokens` call on the whole text. The token classifier caps at 512 (`modernbert.rs:1399`). The routing PII signal chunks (`classifier_signal_pii.go:48`); the API, `AnalyzeContentForPII`, and `DetectPIIInContent` do not. Same shape as #3333, which covers the response jailbreak surface.

### To Reproduce

Model `mmbert32k-pii-detector-merged`, filler then `john.doe@example.com, 555-123-4567, SSN 123-45-6789`:

| filler | entities |
|---|---|
| 409 chars | 2 |
| 2,084 chars | 3 |
| 4,094 chars | 0 |

### Expected behavior

The API chunks like the routing signal and returns offsets in the original text.

### Impact

Documents, transcripts, or logs over ~2,800 characters return no PII past the window.

### Additional context

Verified on `origin/main` (2026-09-02) with a real model on CPU via candle-binding. Related: #3333.

subin9:
Fixes #3364

`ClassifyPIIWithDetailsAndThreshold` classified in one call, and the PII token classifier is built with `max_length = MAX_CLASSIFICATION_SEQ_LEN` (512) and right truncation, so `/api/v1/classify/pii` only saw the start of a long text. PII past that point came back absent, and `masked_text` came back with it unmasked. #3206 fixed the same split on the jailbreak side, where the routing signal already chunked and the classification API did not.

`piiSignalChunks` could not be reused as it stands. It returns `[]string` with no offsets and de-duplicates identical chunks, which discards the mapping from a chunk back to its position in the original text. The routing signal never needed that mapping, because `evaluatePIIRule` reads entity types and confidences only. This surface reports positions, and `buildMaskedPIIText` slices the original text with them. So `securitySignalChunkSpans` returns each chunk with its byte offset, `securitySignalChunks` is expressed in terms of it so there is still one walk, and every entity maps back with `span.StartByte + entity.Start`.

Chunks overlap, so an entity inside the overlap window is classified twice. Detections are de-duplicated on `(type, start, end)` keeping the higher confidence, then sorted by position, which is the order a single call produced.

6898 bytes of prose with `Contact John Doe at john.doe@example.com or 555-123-4567.` appended past the limit, against the real mmBERT-32K PII model, 18 chunks:

```
before   0 detections

after    PERSON         [6849-6857] "John Doe"     0.963
         EMAIL_ADDRESS  [6861-6869] "john.doe"     0.623
         DOMAIN_NAME    [6869-6870] "@"            0.576
         EMAIL_ADDRESS  [6870-6881] "example.com"  0.757
         PHONE_NUMBER   [6885-6897] "555-123-4567" 0.975
```

Every detection satisfies `text[start:end] == entity text` against the original string, so the remap holds under the real tokenizer and not only under a stub. The same entity placed before the limit is detected either way. Short texts still take a single call with the text unchanged.

The defect is in the tokenizer, so a mocked backend cannot reproduce it. One test runs the real model and skips when it is absent, following `classifier_jailbreak_risk_realmodel_test.go`. Four more run against a stub that truncates the same way: an entity past the window is found while one inside it still is and both stay ordered, offsets survive a multi-byte text, an entity in the overlap window is reported once, and a short text stays one call. A new e2e case `pii-long-text` drives `/api/v1/classify/pii` with the entity past the limit and checks the code-point offsets. It is registered in `BaselineRouterContract` next to `pii-entity-offsets`, and neither is on the fixed list the PR lane runs (`ENVOY_AI_GATEWAY_CI_TESTS` in `integration-test-k8s.yml`), so CI has not executed it. I could not run it against a cluster either; `make e2e-test E2E_PROFILE=envoy-ai-gateway E2E_TESTS=pii-long-text` is the command.

No existing test changed.

#3364 also names `AnalyzeContentForPII` and `DetectPIIInContent`. They carry the same single-call shape, but no production caller reaches them today — only `ClassifyPIIWithDetails{,AndThreshold}`, from `services/classification_pii.go`. Same asymmetry, different change, so they are untouched here.

One limit worth writing down rather than rediscovering. An entity is only guaranteed to fall inside a single chunk if it is no longer than `piiSignalChunkOverlapRunes` (64). Sweeping one across the text a rune at a time, 64 and 65 runes miss nothing in 800 consecutive positions, 70 runes misses 8, 80 runes misses 25. Real PII entities are far shorter, so the constant is unchanged here. Raising it is a separate question, because the routing signal shares the walk.

### DIFF
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -24,6 +24,8 @@ var BaselineRouterContract = []string{
 	"pii-detection",
 	// PII entity positions are code-point offsets (issue #3146)
 	"pii-entity-offsets",
+	// PII past the classifier's sequence limit is still detected (issue #3364)
+	"pii-long-text",
 	"jailbreak-detection",
 	"decision-priority-selection",
 	"plugin-chain-execution",
--- a/e2e/testcases/pii_long_text.go
+++ b/e2e/testcases/pii_long_text.go
@@ -1,0 +1,80 @@
+package testcases
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/vllm-project/semantic-router/e2e/pkg/fixtures"
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+	"k8s.io/client-go/kubernetes"
+)
+
+func init() {
+	pkgtestcases.Register("pii-long-text", pkgtestcases.TestCase{
+		Description: "Verify /api/v1/classify/pii detects entities past the classifier's sequence limit",
+		Tags:        []string{"kubernetes", "apiserver", "classification", "pii", "api"},
+		Fn:          testPIILongText,
+	})
+}
+
+// The PII token classifier truncates at its own sequence limit, so a single
+// classification call only ever scores the start of a long text. This drives the
+// endpoint with the entity placed past that limit, which no mocked backend can
+// reproduce: the truncation lives in the tokenizer.
+func testPIILongText(
+	ctx context.Context,
+	client *kubernetes.Clientset,
+	opts pkgtestcases.TestCaseOptions,
+) error {
+	session, err := fixtures.OpenRouterAPISession(ctx, client, opts)
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	httpClient := session.HTTPClient(60 * time.Second)
+	url := session.URL("/api/v1/classify/pii")
+
+	// Filler carries no names, numbers or dates, so any entity reported here
+	// belongs to the trailing secret.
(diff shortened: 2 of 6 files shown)


FAUST-BENCHOU (approved):
Looks good to me. The offset remapping and overlap deduplication cover the tricky parts of chunking here, and the multibyte regression test is especially useful.

wilsonwu (approved):
LGTM

guan404ming (approved):
LGTM, thanks for your contribution!

Issue #3367, opened 2026-09-02
[Bug] Signal windows overflow the 512-token cap on dense and CJK text


guan404ming:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

`classifier_signal_text_window.go` sizes every model-facing window by runes: `securitySignalChunks` (line 84) with a per-rune heuristic, and `representativeSignalText` (line 52) with a flat 2,048-rune head/middle/tail sample. Every ModernBERT / mmBERT classifier truncates at 512 tokens (`modernbert.rs:20`). Wherever runes tokenize denser than the heuristic assumes, the end of the window is never scored: hex, base64, URL, and emoji text for the security chunks, and CJK for the representative window. This affects the routing jailbreak and PII signals, the #3206 classification API, and every bounded signal (domain, fact-check, feedback, reask, preference).

### To Reproduce

Security chunk, model `mmbert32k-jailbreak-detector-merged`. `strings.Repeat("a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6 ", 40) + attack` gives 1 chunk of ~1,300 tokens:

| input | P(jailbreak) |
|---|---|
| attack alone | 1.000 |
| chunked scan | 0.003 |
| same attack after English prose of equal length | 0.699 |

Representative window, model `mmbert32k-intent-classifier-merged`, 40 Chinese legal questions then 40 Chinese calculus questions (2,162 runes = 1,565 tokens; the English control is 459 tokens):

| input | category |
|---|---|
| head only | law 1.00 |
| tail only | math 0.51 |
| representative window | law 1.00 |

### Expected behavior

Every window fits in 512 tokens for any script: size by the tokenizer, or calibrate the rune weights so dense text and CJK cannot exceed the cap.

### Impact

A prompt attack after ~512 tokens inside dense text is invisible to all jailbreak surfaces, and CJK prompts over ~700 characters are routed on their opening only.

### Additional context

Verified on `origin/main` (2026-09-02) with real models on CPU via candle-binding. Related: #3333 (same 512-token cap on the response jailbreak surface).

Issue #3366, opened 2026-09-02
[Bug] Semantic cache collides on long prompts sharing a prefix


guan404ming:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

Default `embedding_model: bert` (`cache.go:171`) calls `GetEmbedding(text, 0)`, which Rust resolves to `max_length = 512` (`core/similarity.rs:161`). Prompts sharing their first 512 tokens get identical embeddings. `multimodal` has the same cap (`ffi/embedding.rs:850`).

### To Reproduce

Model `all-MiniLM-L6-v2`, A = "write a poem about the ocean", B = "SQL to drop the production users table":

| shared prefix | cosine(prefix+A, prefix+B) |
|---|---|
| 335 chars | 0.7156 |
| 4,020 chars | 1.0000 |

### Expected behavior

Do not cache on a truncated embedding, or embed the whole prompt.

### Impact

A long system prompt or history ahead of the user turn returns the answer to a different question. No threshold fixes similarity 1.0.

### Additional context

Verified on `origin/main` (2026-09-02) with a real model on CPU via candle-binding.

Issue #3373, opened 2026-09-02
[Feature] State which model-facing surfaces sample and which scan


subin9:
### Goal

Write down which model-facing surfaces may discard part of their input and which must read all of it, so that a new call site has a rule to follow instead of a precedent to copy.

Three treatments exist today and nothing states when each applies:

- `representativeSignalText` keeps a head, middle and tail sample of 2 048 runes and drops everything between (`classifier_signal_text_window.go:53`). The eleven types in `boundedSemanticSignalTypes` take this path.
- `securitySignalChunks` scans the whole input in overlapping chunks, because "security detection cannot safely discard the middle of a long prompt" (`classifier_signal_text_window.go:82`). The PII and jailbreak routing signals take this path.
- Several call sites took neither and handed the whole text to a model that truncates it: #3204 on the security classification API, #3333 on the response jailbreak filter, #3364 on the PII classification API.

The first two are deliberate and the reasoning for each is in the file. What is missing is the rule that tells a new surface which one it owes, which is how the third group keeps appearing — three times in the last week, each found by reading the call site.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- State the contract: which categories of surface may sample, which must scan, and what a caller that reports positions owes on top of that. The last part is not theoretical. A chunked scan that reports offsets has to map them back to the original text, which is why `piiSignalChunks` could not be reused as it stands in #3364 — it returns `[]string` with no offsets and de-duplicates identical chunks.
- Document the budgets. The comment at `classifier_signal_text_window.go:17-27` explains why PII and jailbreak have different budgets, but not how the specific values were chosen or what margin they leave: `piiSignalChunkBudget` is 128 tokens and `jailbreakSignalChunkBudget` is 384, both against a 512-token cap, so one has about four times the headroom and the other about a third. Since the budget is spent with a per-rune heuristic, that margin is what absorbs text tokenizing denser than the heuristic assumes.
- Record the bound the overlap implies. A span is only guaranteed to sit inside a single chunk if it is no longer than the overlap, 64 runes for both security signals. Measured against a stub that truncates the way the tokenizer does, a 70-rune span is missed at 8 of 800 consecutive offsets and an 80-rune span at 25.
- Decide whether the contract should be enforced rather than documented — for example a model-facing helper a call site cannot bypass without saying which treatment it wants.

### Non-goals

- Changing any existing sampling or chunking behaviour. The head, middle and tail sample for intent signals is deliberate and should stay.
[...]

PR #3383, opened 2026-09-02
[Bug] Name the position-embedding limit when multimodal text input exceeds it

### LINKED ISSUE #3381: [Bug] Multimodal text encoding fails past 512 tokens with an opaque index error
### Proposed Workgroup

### Describe the bug

`multimodal_encode_text` (`ffi/embedding.rs:2296`) tokenizes and uses the result as-is:

```rust
let ids: Vec<u32> = encoding.get_ids().to_vec();
let mask: Vec<u32> = encoding.get_attention_mask().to_vec();
let seq_len = ids.len();
```

The text encoder has 512 position embeddings (`multimodal_embedding.rs:73`), so a longer input fails in the position lookup. The batch path in the same model clamps — `max_length.min(...)` at `multimodal_embedding.rs:1279` — and `ffi/embedding.rs:850` passes 512 into it. Only the single-text FFI skips it.

### To Reproduce

`mom-embedding-multimodal` on CPU, `"navigation "` repeated, through `MultiModalEncodeText`:

```
words  runes  result
509    5599   ok, 364ms
510    5610   ok, 355ms
511    5621   index-select invalid index 512 with dim size 512
```

510 words plus the two special tokens is exactly 512.

CJK reaches the limit far sooner, and the routing signal's own bound does not help, because `representativeSignalText` caps at 2048 runes and runes are not tokens:

```
CJK filler   512 runes   error
            1024 runes   error
            2048 runes   error
ASCII prose 2046 runes   ok
```

So an embedding signal configured with `modelType: multimodal` fails on any CJK query longer than about 510 characters, and the sampler's cap sits four times above the failing point.

### Expected behavior

Clamp to the position-embedding limit the way the batch path does, or fail with an error that names the limit and the input length instead of a candle index-select error.

### Impact

Any deployment using the multimodal embedding model. The embedding signal returns an error rather than a result, and the message names neither the limit nor the input, so the cause is not obvious from logs.

### Additional context

Which of the two behaviours is right is worth deciding rather than defaulting to the clamp. Silent truncation is what produced #3204, #3333 and #3364, so matching the other models is not automatically the better answer here.

Related: #3372 on making truncation observable, #3373 on which surfaces may discard input.

subin9:
Fixes #3381

## Purpose

The multimodal text encoder holds 512 position embeddings. Two single-text FFI paths tokenized and used the result as-is, so a longer input failed inside the position lookup with `index-select invalid index 512 with dim size 512`, a message that names neither the limit nor the input:

- `multimodal_encode_text`, reached from `MultiModalEncodeText`
- `generate_multimodal_text_embedding`, reached from `get_embedding_2d_matryoshka` and `get_embedding_with_model_type`

Both now go through `check_position_limit`, which rejects the sequence with an error naming the token count and the limit. Nothing is truncated.

An earlier revision clamped to the first 512 ids to match the batch path. That was the wrong default for the second path: `generate_multimodal_text_embedding` is also what every semantic cache backend calls for `embedding_model: multimodal` (`cache/inmemory_cache.go:257` and the milvus, qdrant, redis and valkey equivalents), so a clamp there would give the cache the same vector for any two long prompts sharing a 512-token prefix. That is the collision #3366 reports on the other embedding models; today the multimodal cache fails instead of colliding, and this keeps it that way with a readable reason. #3381 lists a descriptive error as one of its two acceptable outcomes.

On the routing embedding signal the error is logged and the signal skipped (`classifier_signal_embedding_helpers.go:61`), so the request is unaffected and only the log line changes. Long CJK reaching this limit at all is #3367: `representativeSignalText` measures its 2048-rune window in runes, not tokens. If the maintainers decide this surface should score a prefix instead, the helper is the one place a clamp would go.

Affects `candle-binding` only. Owned by Router Models & Inference Runtime via #3381.

`generate_mmbert_embedding` has the same unclamped shape but 32K position embeddings and chunked attention, so it does not fail here and is left alone.

## Test Plan

```bash
cd candle-binding
cargo fmt --all -- --check
cargo test --release --no-default-features --lib position_limit
cargo test --release --no-default-features --lib

# real model on CPU, both FFI paths, the receipt for a multimodal FFI change (#2319)
MULTIMODAL_MODEL_PATH=<path>/mom-embedding-multimodal make test-binding-multimodal
```

Plus a throwaway Go test (not committed) that drives `MultiModalEncodeText`, `GetEmbeddingWithModelType("multimodal")` and `GetEmbedding2DMatryoshka("multimodal")` with `"navigation "` repeated 510, 511 and 1024 times and with 512 and 2048 runes of CJK, and checks what the binding logs.

## Test Result

`cargo fmt --all -- --check` is clean (rustfmt 1.8.0-stable, the rewrite the PR gate was applying).

`cargo test --release --no-default-features --lib`: 526 passed, 69 failed, 45 ignored. The six `check_position_limit` cases pass. The 69 are the model-loading tests (`core::similarity_test`, `core::tokenization_test`, gemma, modernbert, `get_embedding_smart`) that fail on any machine without those models and fail identically on `main`.

`mom-embedding-multimodal` on CPU through all three Go entry points, `MultiModalEncodeText`, `GetEmbeddingWithModelType("multimodal")` and `GetEmbedding2DMatryoshka("multimodal", 3, 256)`. `before` is a build of the tree without this change. Every row agrees across the three paths, so each is listed once:

| input | tokens | before | after |
|---|---|---|---|
| `"navigation "` x510 | 512 | ok | ok |
| `"navigation "` x511 | 513 | `index-select invalid index 512 with dim size 512` | `text tokenizes to 513 tokens, over the text encoder's 512-position limit` |
| `"navigation "` x1024 | 1026 | same index error | same shape, `1026 tokens` |
| Korean prose, 512 runes | 624 | same index error | same shape, `624 tokens` |
| Korean prose, 2048 runes | 2477 | same index error | same shape, `2477 tokens` |
| Chinese prose, 512 runes | 514 | same index error | same shape, `514 tokens` |
| Chinese prose, 2048 runes | 2050 | same index error | same shape, `2050 tokens` |

The Go-side error strings are unchanged (`multi-modal text encoding failed`, `failed to generate embedding with model type multimodal (status: -1)`). The binding has always put the reason on stderr, and that is the line that changed. One note for anyone reproducing the CJK rows: a run of CJK with no spaces is not a valid probe, because WordPiece maps a word over its character cap to a single `[UNK]`, so 2048 runes of one repeated word encodes fine.

`make test-binding-multimodal` receipt (#2319), both halves, against `models/mom-embedding-multimodal`:

```
candle-binding, -race, ^TestMultiModal
  TestMultiModalEmbeddingInit            PASS
  TestMultiModalEncodeText               PASS  (BasicText, MRL_Dimensions, SemanticSimilarity, Consistency)
  TestMultiModalEncodeImageFromBytes     PASS
  TestMultiModalEncodeImageFromBase64    PASS
  TestMultiModalEncodeImageFromURL       PASS
  TestMultiModalCrossModalRetrieval      PASS
  TestMultiModalInputValidation          PASS
  ok  119.247s

pkg/classification, ^TestEmbeddingClassifier_Integration
  TestEmbeddingClassifier_IntegrationImageQueryEndToEnd           PASS
  TestEmbeddingClassifier_IntegrationTextRulesIgnoredOnImagePath  PASS
  ok  7.249s
```

`test-binding-multimodal-rust-baseline` is the known-red exploratory lane and is deliberately not part of that receipt, so it was not run.

Remaining question for the WG: whether this surface should clamp instead. Nothing in this PR forecloses that; it is a change to `check_position_limit` and its two callers.

---
<details>

</details>


FAUST-BENCHOU (commented):
Could we confirm the intended contract here before settling on silent truncation? #3381 explicitly leaves clamping vs. a descriptive error open, so matching the existing batch path alone doesn’t seem to resolve that design choice.

The current PR gate is also still red because `cargo fmt` rewrites both changed Rust files. Could we fix the formatting and align the PR description with the repository’s `Purpose / Test Plan / Test Result` template before merge?

subin9:
Thank you @FAUST-BENCHOU, all three seem fair :)

fmt and the template are fixed. While in there I moved the helper out of the middle of `multimodal_encode_text`'s doc comment, which the earlier placement had split in two.

On the contract, I traced the callers and switched sides: the PR now returns an error naming the token count and the limit, and truncates nothing. `generate_multimodal_text_embedding` is also the semantic cache's path for `embedding_model: multimodal`, so a clamp there would hand the cache one vector for any two long prompts sharing a 512-token prefix, which is #3366. On the routing signal the error is already logged and the signal skipped, so only the log line changes; long CJK hitting the limit at all is #3367. Before/after on the real model is in the description.

@Xunzhuo #3381 left clamp vs. error open; could you confirm an error is the contract you want for this surface? If it should clamp instead, it's a small change in `check_position_limit` and its two callers. The wider question, which surfaces may discard input at all, is filed as #3373; that can stay separate from this PR.

Xunzhuo (approved):
This keeps the existing no-truncation behavior while replacing the opaque position-lookup failure with the actual token count and encoder limit on both affected FFI paths. The boundary tests, real-model receipt, green CI, and local agent validation cover the change well. Approved; please rebase onto current main before merge.

FAUST-BENCHOU (approved):
Thanks for your work

wilsonwu (approved):
LGTM

PR #3299, opened 2026-09-01
[Feature] Support open-ended context signal ranges and overflow-safe routing

yaojiejia:
Closes #3054

## Purpose

Context signal rules could not express "and above": leaving out `max_tokens` parsed as 0, so the rule never matched, and the only workaround was a huge literal like `999M`. Nothing validated the ranges either, so `min > max`, negative values, and typos loaded fine and silently dropped the signal.

This PR makes context bands explicit and safe:

- Omitting `max_tokens` is now an open-ended band that matches every count at or above `min_tokens`. Omitting `min_tokens` means 0.
- Bounds stay inclusive, and `min_tokens == max_tokens` is a supported exact-match band.
- Config and DSL validation reject invalid ranges with actionable messages and warn on gaps and partial overlaps. Overlaps still report every matching rule in config order.
- The classifier parses bounds once, logs invalid rules instead of skipping them silently, and leaves a debug trace when an open-ended band matches.
- The CRD, dashboard, DSL schema, and docs now agree that both limits are optional.

Modules: `pkg/config`, `pkg/classification`, `pkg/dsl`, `pkg/apis`, `pkg/k8s`, `deploy/*/crds`, `dashboard/frontend`, `website/docs`.
Owner: `wg/mom-routing`.

## Test Plan

```bash
cd src/semantic-router
go vet ./pkg/config/ ./pkg/dsl/ ./pkg/classification/ ./pkg/k8s/
go test ./pkg/config/ ./pkg/dsl/ ./pkg/classification/ ./pkg/k8s/ ./pkg/services/ ./internal/nlgen/
make generate-api   # regenerates CRDs and syncs the helm copy

cd dashboard/frontend
npx tsc --noEmit -p .
npx vitest run src/lib src/pages/topology src/pages/ConfigPage

```
Test Result

All listed Go packages, tsc, eslint, and the vitest suites pass. CRD regeneration only changed the contextRules field descriptions and dropped minTokens/maxTokens from required. The built-in mom-v1 recipe loads with no new diagnostics.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title begins with exactly one bracketed category, such as [Feature], [Bug], [Docs], [Test], [Research], [Community], or [CI/Build]
- [x] The title does not stack prefixes such as [Router][Docs]; affected modules belong in labels and the PR body
- [x] The PR links an accepted issue with exactly one owner: one wg/* label for project work or owner/maintainers for repository governance
[...]

PR #3386, opened 2026-09-02
[Bug] Accept open-ended context bands in the CLI config schema

### LINKED ISSUE #3054: [Feature] Support open-ended context signal ranges and overflow-safe routing
### Goal

Make context-signal matching explicit and safe for exact token counts, bounded ranges, open-ended ranges, and overflow.

### Proposed Workgroup

MoM & Routing

### Scope

- Preserve inclusive bounded matching for `min_tokens <= token_count <= max_tokens`.
- Treat `min_tokens == max_tokens` as a supported exact-match band, not an invalid or empty range.
- Add an unbounded upper range, or an equivalent explicit overflow rule, for the final band.
- Preserve the meaning of existing bounded configurations.
- Define deterministic validation or precedence for gaps and overlaps.
- Keep model context-window limits distinct from semantic routing bands.
- Emit bounded diagnostics when fallback, gap, overlap resolution, or overflow behavior is used.
- Align Router configuration, CLI, Dashboard, documentation, and replay/debug evidence.

### Non-goals

- Changing tokenization semantics or model context-window limits.
- Adding a second context signal implementation.
- Using fuzzy or semantic matching for numeric token bands.

### Completion signal

- A singleton band with equal minimum and maximum matches exactly that token count and has focused tests.
- Requests below, inside, exactly on both boundaries, between, and above configured bands have deterministic results.
- A final open-ended band can match all values above its minimum without integer-overflow tricks.
- Existing valid configurations continue to load with unchanged meaning.
- Invalid or ambiguous ranges receive actionable validation, and routing cannot silently lose the context signal.

Parent Epic: #3202

yaojiejia:
Closes #3054

## Purpose

#3299 made an omitted `max_tokens` an open-ended context band in the Router, Dashboard, and docs, but `ContextRule` in `src/vllm-sr/cli/models.py` still required `max_tokens: str`. A valid open-ended config loaded in the Router and failed in the CLI with `max_tokens: Field required`. The `str` field also rejected unquoted numbers like `min_tokens: 8001`, which the Router and docs examples use.

This aligns the CLI with the Router:

- New `cli/context_bands.py` mirrors the Router's rules: strings or numbers, `K`/`M` suffixes, at least one limit, equal limits allowed, `min_tokens` above `max_tokens` rejected with the Router's message.
- `ContextRule` makes both limits optional and runs that check.
- New `tests/test_context_signal_bands.py` covers every band shape and rejection.
- Docs note the Router, CLI, and Dashboard enforce the same rules.

Out of scope: the built-in `mom-v1` recipe still uses the `max_tokens: "2147483647"` sentinel. It touches DSL round-trip tests, so it is left for a follow-up.

## Test Plan

```bash
# config with a final band that omits max_tokens
PYTHONPATH=src/vllm-sr .venv-agent/bin/vllm-sr config router --config open-ended.yaml
# load the emitted file through the Router's config.Load and check Unbounded

cd src/vllm-sr
python -m pytest tests/test_context_signal_bands.py tests/test_router_signal_extensions.py tests/test_cli_models.py tests/test_config_contract.py tests/test_validator_workflows.py -q
python -m pytest tests -q
black --check cli/context_bands.py cli/models.py tests/test_context_signal_bands.py
ruff check cli/context_bands.py cli/models.py tests/test_context_signal_bands.py
``` 
Test Result

- Reproduction fails before the change with context -> 1 -> max_tokens: Field required and exits 0 after.
- Router loads the emitted file: long min=4001 unbounded=true.
- Targeted tests: 83 passed (26 new).

---
<details>

- [x] PR title begins with exactly one bracketed category, such as [Feature], [Bug], [Docs], [Test], [Research], [Community], or [CI/Build]
- [x] The title does not stack prefixes such as [Router][Docs]; affected modules belong in labels and the PR body
- [x] The PR links an accepted issue with exactly one owner: one wg/* label for project work or owner/maintainers for repository governance

</details>

See CONTRIBUTING.md for the full contributor workflow and commit guidance.


FAUST-BENCHOU (commented):
Could we mirror the Router's overflow check here as well? `TokenCount.Value()` rejects scaled values at `math.MaxInt`, while `parse_token_count()` only checks finiteness and negativity, so something like `1e100` can pass the CLI and then fail in the Router. A parity test for this case would keep the “same rules” contract true.

yaojiejia:
> Could we mirror the Router's overflow check here as well? `TokenCount.Value()` rejects scaled values at `math.MaxInt`, while `parse_token_count()` only checks finiteness and negativity, so something like `1e100` can pass the CLI and then fail in the Router. A parity test for this case would keep the “same rules” contract true.

gotcha, just implemented

Xunzhuo (commented):
Thanks for following through on the CLI parity. The new validator matches the Router's numeric parsing, overflow boundary, and exact/open-ended band semantics, including YAML-number coercion. The exact-head CLI, core, and E2E gates are green, and I don't see a remaining blocker. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

Xunzhuo (changes requested):
Thanks—the overflow and open-ended fixes are in place. One parity blocker remains: parse_token_count() delegates to Python float(), while Router uses Go strconv.ParseFloat; those grammars differ. Full-width digits such as "１２８" pass CLI validation but fail Router loading, while an existing bounded config such as "0x1p10"/"0x1p20" is accepted by Router (and previously passed through the CLI) but is newly rejected here. Please align the parser grammar with Router and add cross-language regression cases.

yaojiejia:
> Thanks—the overflow and open-ended fixes are in place. One parity blocker remains: parse_token_count() delegates to Python float(), while Router uses Go strconv.ParseFloat; those grammars differ. Full-width digits such as "１２８" pass CLI validation but fail Router loading, while an existing bounded config such as "0x1p10"/"0x1p20" is accepted by Router (and previously passed through the CLI) but is newly rejected here. Please align the parser grammar with Router and add cross-language regression cases.

Fixed in 8bf20825. The CLI now parses token counts with the same grammar as Go's `strconv.ParseFloat`, so `１２８` is rejected and `0x1p10`/`0x1p20` load again. Both the Router and CLI tests now read the same case file, `pkg/config/testdata/token_count_cases.json`, so any future drift fails on both sides.

Xunzhuo (changes requested):
Thanks for working through the grammar cases. The parity check still runs after PyYAML has discarded the original scalar spelling: unquoted 0x10 and 0b101 become Python integers and pass the CLI even though the Router rejects those token-count strings, while 0123 becomes 83 in the CLI but 123 in the Router. Because config router forwards the original YAML, this permits CLI-valid/Router-invalid or semantically different configs. Please preserve or validate the raw scalar lexeme at the YAML boundary and add YAML-to-Pydantic-to-Router parity cases for these implicit-resolution forms.

yaojiejia:
> Thanks for working through the grammar cases. The parity check still runs after PyYAML has discarded the original scalar spelling: unquoted 0x10 and 0b101 become Python integers and pass the CLI even though the Router rejects those token-count strings, while 0123 becomes 83 in the CLI but 123 in the Router. Because config router forwards the original YAML, this permits CLI-valid/Router-invalid or semantically different configs. Please preserve or validate the raw scalar lexeme at the YAML boundary and add YAML-to-Pydantic-to-Router parity cases for these implicit-resolution forms.

Thanks, just checked with `config.Parse`: the Router loader decodes untyped with yaml.v2 and re-marshals before the typed decode, so unquoted `0x10`, `0b101` and `0123` already load as 16, 5 and 83 on both sides. The real divergences were forms where PyYAML and yaml.v2 typing differ, such as `1:30` (accepted by the CLI, rejected by the Router) and `0o17`, `0X10`, `1_e3` (the reverse). Fixed in dded077b and 0c849245 by applying yaml.v2's scalar typing to token counts at the CLI YAML boundary (`cli/config_yaml.py`, `cli/goyaml_scalars.py`), deferring `${ENV}` references to the Router with a warning, and pinning 92 YAML-to-Pydantic-to-Router cases in the shared fixture.

Xunzhuo (approved):

PR #3390, opened 2026-09-02
[Bug] Cut ReMoM last_n_tokens compaction on rune boundaries using observed bytes per token

### LINKED ISSUE #3387: [Bug] ReMoM last_n_tokens compaction cuts mid-rune and misjudges CJK budget
### Proposed Workgroup

MoM & Routing

### Describe the bug

ReMoM's `last_n_tokens` compaction keeps the tail of each intermediate response for the next round. It does this by cutting the string at a byte offset (`looper/remom.go:653`):

```go
maxChars := maxTokens * 4
return content[len(content)-maxChars:]
```

Go strings are bytes, and a Chinese character is 3 bytes. Cutting at a byte offset breaks in two ways.

**1. The cut can land in the middle of a character.**

Two out of three offsets fall inside a 3-byte character, so the kept text starts with a partial character. Go's JSON encoder replaces the invalid bytes with `�` when the prompt is sent upstream. The model sees `�...` instead of clean text at the start of every reference response, and the cut character is lost.

**2. The budget is measured in bytes, but the config is in tokens.**
The code assumes 4 bytes per token. That holds for English. For Chinese, 4000 bytes is only 1333 characters, which is about 600 Qwen tokens. So `compaction_tokens: 1000` actually keeps 600 tokens of Chinese. `estimateTokens` (`remom.go:770`) uses the same assumption, so the `token_count` in the trace is off by the same factor.

### To Reproduce

Config: `compaction_strategy: last_n_tokens`, `compaction_tokens: 1000`

```go
content := strings.Repeat("量子计算机的工作原理是", 400)
out := l.compactResponse(cfg, content)
// len(out) = 4000 bytes, 1334 runes, utf8.ValidString(out) = false
// head bytes: \x9f理\xe6\x98
// after json.Marshal: "\ufffd理是\ufffd\ufffd"
```

Token count of the kept tail (tokenizers crate, truncation disabled):

| Tokenizer | 4000-byte CJK | 4000-rune CJK | 4000-byte ASCII |
|---|---|---|---|
| Qwen3 | 607 | 1819 | 969 |
| all-MiniLM | 1333 | 4000 | 904 |

Configured budget is 1000. With an upstream LLM tokenizer the code keeps about 600.

### Expected behavior

- Compacted text is valid UTF-8 and starts on a character boundary.
- Kept length tracks `compaction_tokens` regardless of script.

### Impact

- Any decision using `remom` with `compaction_strategy: last_n_tokens` on non-ASCII responses.
- Every reference response in the next round starts with replacement characters.
- CJK responses keep about 40% less context than configured.
- Workaround: `compaction_strategy: full`.

### Additional context

Two separate fixes:

- **Boundary:** back off to a rune start (`utf8.RuneStart`) before slicing. Keeps the byte budget unchanged.
- **Budget:** rune slicing alone makes it worse (1819 Qwen tokens for 4000 runes). Needs a tokenizer-aware or calibrated count. The classification package already has a calibrated bytes-per-token counter.

yaojiejia:
Closes #3387

## Purpose

ReMoM's `last_n_tokens` compaction sliced responses at a raw byte offset of `compaction_tokens * 4`. This broke on non-ASCII text in two ways. The cut could land inside a multi-byte character, so the kept text started with `\ufffd` after JSON encoding. The budget assumed 4 bytes per token, so CJK kept about 600 Qwen tokens when 1000 were configured. The trace `token_count` was off by the same factor.

This change is confined to `pkg/looper/remom.go`:

- Advance the cut to the next rune start. Compacted text is now valid UTF-8 and stays within the byte budget.
- Derive bytes per token from the response's own completion usage. The looper already stores this on `ModelResponse.Usage`. Completion usage covers content plus reasoning, so the ratio uses their combined byte length.
- Fall back to 4 bytes per token when the backend reports no usage. Behaviour for those responses is unchanged.
- Use the same observed ratio for the trace `token_count`.

This avoids wiring the classification package's calibrated counter into the looper. Each response carries an exact ratio for its own model and script.

## Test Plan

New tests in `pkg/looper/remom_compaction_test.go`:

- Issue repro: 400 repeats of a CJK phrase with budget 1000. Asserts valid UTF-8, clean first rune, tail preserved, budget respected, and no `\ufffd` after `json.Marshal`.
- ASCII and CJK responses with reported usage each keep exactly 100 tokens' worth of their own script.
- Ratio derivation: nil response, missing usage, content only, and content plus reasoning.
- `lastNTokens` edge cases: under and exact budget, fractional ratio, cut inside a rune, cut on a boundary, non-positive ratio, and tiny budget.
- `full` strategy is untouched. Round responses use the observed ratio.

Commands, run from `src/semantic-router` with the Rust bindings on `LD_LIBRARY_PATH`:

```
gofmt -l pkg/looper
go vet ./pkg/looper/
go test ./pkg/looper/ -run 'Compact|LastNTokens|BytesPerToken|BuildReMoMRoundResponse' -count=1 -v
go test ./pkg/looper/ -count=1
golangci-lint run ./pkg/looper/
```

## Test Result

- Targeted compaction tests: pass.
- Full `pkg/looper` package: `ok` in 11.3s.
- gofmt and go vet: clean.
- golangci-lint: no findings in changed files. Three pre-existing findings in `workflows_state_store.go` and `client.go` are unrelated.

---
<details>

</details>


FAUST-BENCHOU (commented):
I think this should wait on the repository lifecycle first: #3387 is still `needs-acceptance`, and the current gates are red on the accepted-issue check, with `remom.go` also exceeding the 800-line limit.

Separately, could we revisit the observed ratio when both content and reasoning are present? `reMoMSynthesisText()` uses only content when it exists, while `reMoMBytesPerToken()` derives the ratio from content + reasoning and total completion tokens, so the budget can be inaccurate for the text actually being compacted.

yaojiejia:
> I think this should wait on the repository lifecycle first: #3387 is still `needs-acceptance`, and the current gates are red on the accepted-issue check, with `remom.go` also exceeding the 800-line limit.
> 
> Separately, could we revisit the observed ratio when both content and reasoning are present? `reMoMSynthesisText()` uses only content when it exists, while `reMoMBytesPerToken()` derives the ratio from content + reasoning and total completion tokens, so the budget can be inaccurate for the text actually being compacted.

Thanks for the review, I think the issue just got accepted and I fixed the 800 line limit ci failure. 

To answer the other concern:
`completion_tokens` counts content and reasoning together and `TokenUsage` doesn't keep a breakdown, so if I divide only the content bytes by it, the ratio comes out too low and the budget shrinks whenever reasoning is present. The combined ratio is the model's average bytes per token for that call, and it tracks the content closely unless reasoning and content differ in script or density, or the backend bills hidden reasoning it never returns. vLLM and OpenAI both report `completion_tokens_details.reasoning_tokens` and the SDK already parses it, so a content-only ratio only needs a new field on `TokenUsage` and its two parsers. Maybe we can do that as a follow up but Im also happy to address here as well.

Xunzhuo (changes requested):
The observed ratio still does not reach the next-round prompt. `buildReMoMRoundResponse` stores `CompactedContent`, but `prepareReMoMRoundMessages` reconstructs each `ModelResponse` with only content, reasoning, and model. `buildSynthesisPrompt` then sees zero usage and falls back to 4 bytes/token, so the visible intermediate metadata changes while actual CJK synthesis compaction keeps the old behavior. Please carry the ratio/usage through or build from the stored compacted content, and add a multiround regression that asserts the prompt sent to the next round.

yaojiejia:
> The observed ratio still does not reach the next-round prompt. `buildReMoMRoundResponse` stores `CompactedContent`, but `prepareReMoMRoundMessages` reconstructs each `ModelResponse` with only content, reasoning, and model. `buildSynthesisPrompt` then sees zero usage and falls back to 4 bytes/token, so the visible intermediate metadata changes while actual CJK synthesis compaction keeps the old behavior. Please carry the ratio/usage through or build from the stored compacted content, and add a multiround regression that asserts the prompt sent to the next round.

Thanks for the review, the reconstructed `ModelResponse` dropped `Usage`, so the synthesis prompt fell back to 4 bytes/token. Fixed in cd76bccd: `IntermediateResp` now carries the backend usage (`json:"-"`, so the published payload is unchanged) and `intermediateResponsesToModelResponses` copies it back, which also covers `remomRoundResponsesToModelResponses`. Added `TestReMoMMultiRoundSynthesisPromptUsesObservedRatio`, a two-round run against a mock backend that asserts the reference in the next-round prompt is the 100-rune tail and matches the published `compacted_content`. It fails before the fix with 133 runes in the prompt.

Xunzhuo (approved):

PR #3393, opened 2026-09-03
[Bug] Prevent duplicate v1 paths in Envoy rewrites

### LINKED ISSUE #3292: [Bug] Make Envoy base URL path rewrites idempotent for /v1-prefixed targets

Apricooooot:
Closes #3292

## Summary

Prevent provider base paths from being applied twice when Envoy reevaluates routing after Semantic Router selects a model.

The root cause was two components writing the upstream path:

1. `ext_proc` resolved and wrote the complete provider path from the selected provider profile.
2. The selected-model Envoy route then applied its own `regex_rewrite` to that already-resolved path.

For example:

```text
Ingress path:    /v1/chat/completions
Provider prefix: /v1/provider

Expected:        /v1/provider/chat/completions
Incorrect:       /v1/provider/v1/provider/chat/completions
```

This change makes `ext_proc` the single owner of provider-resolved paths while preserving the existing fallback-route rewrite needed by requests that skip semantic processing.

## Solution

The generated Envoy configuration now follows these ownership rules:

| Route | Path owner | `regex_rewrite` |
| --- | --- | --- |
| Selected-model route | `ext_proc` | Removed |
| Default/fallback route | Envoy | Preserved |
| External-gateway route | Existing external-gateway path handling | Unchanged |

After model or provider selection, `ext_proc` writes the complete upstream `:path` and sets the selected-model routing header. Envoy then reevaluates routing and selects a model-specific route that no longer rewrites the path.

Requests that skip semantic processing remain on the default route, where the existing `regex_rewrite` continues to apply the configured direct-endpoint prefix.

No Lua finalizer, capability negotiation, or internal path-carrier headers are introduced.

## Request flow

```mermaid
flowchart TD
    A["Ingress request<br/>/v1/chat/completions"] --> B["Initial fallback route"]
    B --> C{"Semantic processing?"}

    C -->|"Model/provider selected"| D["ext_proc resolves complete provider path"]
    D --> E["Set x-selected-model"]
    E --> F["Clear Envoy route cache"]
    F --> G["Selected-model route"]
    G --> H["Forward path unchanged"]

    C -->|"Skip processing"| I["Remain on default route"]
    I --> J["Existing regex_rewrite applies endpoint prefix"]
    J --> H
```

## Route-cache handling

Most provider-dispatch paths already cleared the Envoy route cache after setting `x-selected-model`.

This change also covers two paths that can enter through the fallback route and then materialize a provider path:

- Looper internal provider dispatch;
- same-name entrypoint provider dispatch.

Both now request route reevaluation after supplying the complete provider path and selected-model routing header. This moves the request from the rewriting fallback route to the non-rewriting selected-model route.

External-gateway mode intentionally keeps its existing selected route and direct `:path` mutation behavior.

## Preserved behavior

This change preserves:

- provider-profile path resolution in `ext_proc`;
- OpenAI, Anthropic, Responses, and custom provider path handling;
- query-string handling;
- host and cluster selection;
- authentication and provider headers;
- direct endpoints with non-empty path prefixes;
- skip-processing behavior through the default route;
- external-gateway protocol conversion;
- existing retry and outlier-detection configuration.

The change only removes the second path writer from selected-model routes. The default route retains its existing rewrite behavior.

## Regression coverage

| Case | Expected result |
| --- | --- |
| `/v1/chat/completions` with prefix `/v1/provider` | `/v1/provider/chat/completions` |
| `/v1/chat/completions` with prefix `/v1beta/openai` | `/v1beta/openai/chat/completions` |
| Overlapping prefix `/v1/chat` | `/v1/chat/chat/completions` |
| Prefix beginning with a `/v1` segment | Applied exactly once |
| Query string such as `?api-version=test` | Preserved |
| Direct endpoint with `/compatible-mode/v1` | Provider-resolved path is forwarded once |
| Skip-processing direct endpoint | Default-route prefix rewrite remains active |
| Looper provider dispatch | Route is reevaluated and the provider path is not rewritten again |
| Same-name entrypoint provider dispatch | Route is reevaluated onto the selected-model route |
| External-gateway dispatch | Existing direct-path behavior remains unchanged |
| Repeated route evaluation | Selected-model route cannot apply a second path rewrite |

## Tests

The following validations passed after rebasing onto the latest `main`:

- `go test ./pkg/extproc`
- `go vet ./pkg/extproc`
- selected-model and default-route configuration contract tests:
  - 43 relevant Python tests passed
  - one unrelated existing `listener.api_keys` schema test was deselected locally
- production router image build
- six focused CLI integration tests:
  - `test_base_url_path_rewrite_is_idempotent`
  - `test_base_url_path_rewrite_idempotency_for_v1_segment_prefix`
  - `test_base_url_path_rewrite_with_overlapping_prefix`
  - `test_direct_endpoint_path_is_resolved_by_ext_proc`
  - `test_skip_processing_keeps_direct_endpoint_prefix_rewrite`
  - `test_looper_custom_prefix_uses_selected_route_without_rewrite`
- Black formatting check
- `git diff --check`

Focused CLI integration result:

```text
Ran 6 tests in 166.035s

OK
```

## Scope

The production behavior change is intentionally small:

- remove `regex_rewrite` from selected-model routes;
- preserve it on the default route;
- clear the route cache in the two additional provider-dispatch paths that previously remained on the fallback route.

The remaining changes are focused unit, configuration-contract, and CLI integration coverage.

## Checklist

- [x] The PR title contains exactly one category: `[Bug]`.
- [x] The accepted issue is linked.
- [x] The commit includes a DCO `Signed-off-by` line.
- [x] Selected-model and default-route behavior are covered by unit tests.
- [x] Overlapping-prefix behavior is covered by an integration regression.
- [x] Route-cache recomputation is covered by integration tests.
- [x] Direct-endpoint and skip-processing behavior are covered.


FAUST-BENCHOU (commented):
Could we cover the case where the backend prefix overlaps the original API path, e.g. `path_prefix=/v1/chat` with `/v1/chat/completions`? The new alternation treats that first-pass request as already rewritten and leaves it unchanged, so prefix matching alone may not reliably distinguish ingress from a rewritten path—the ambiguity #3292's immutable ingress-path requirement was meant to avoid.

Apricooooot:
> Could we cover the case where the backend prefix overlaps the original API path, e.g. `path_prefix=/v1/chat` with `/v1/chat/completions`? The new alternation treats that first-pass request as already rewritten and leaves it unchanged, so prefix matching alone may not reliably distinguish ingress from a rewritten path—the ambiguity #3292's immutable ingress-path requirement was meant to avoid.

Thanks @FAUST-BENCHOU , this is a good catch. I dug further into the request flow, and prefix matching alone cannot distinguish an original ingress path from an already rewritten path when the two overlap.
One important detail is that clearing Envoy’s route cache only invalidates the cached route selection; it does not restore the original :path. Currently, ext_proc can set a provider-specific final :path, and the newly selected Envoy route may then apply another regex_rewrite.
I see two possible directions:
1. Preserve a trusted, immutable ingress path before ext_proc mutations and always derive Envoy’s rewrite from that path during route evaluation.
2. Make ext_proc the single owner of the final provider path, since it already resolves the provider profile and wire format, and have the selected-model Envoy routes only choose the cluster/host without rewriting again. This would also need to preserve the behavior of default and skip-processing routes, as well as branches where the route cache is not cleared.
I’m currently leaning toward the second option because the provider-specific path logic already lives in ext_proc, but it has a somewhat broader configuration impact. Does the first option better match the intended immutable-ingress-path design for #3292, or would the single-owner approach be preferable?
Either way, I’ll add the overlapping /v1/chat case as a regression test in addition to the existing /v1/provider case.

Xunzhuo (changes requested):
Welcome, and thanks for picking this up. There is one correctness blocker: the alternation cannot distinguish an ingress path from an already rewritten path when the configured prefix overlaps it. For example, with `path_prefix=/v1/chat`, `/v1/chat/completions` matches the prefix branch on the first pass and is left unchanged, although the first rewrite should produce `/v1/chat/chat/completions`. The added test only covers the non-overlapping `/v1/provider` case. Please implement the accepted issue’s immutable-ingress-path/single-owner approach (and strip any internal carrier before upstream), then add an overlapping-prefix regression plus repeated-application coverage.

Apricooooot:
> Welcome, and thanks for picking this up. There is one correctness blocker: the alternation cannot distinguish an ingress path from an already rewritten path when the configured prefix overlaps it. For example, with `path_prefix=/v1/chat`, `/v1/chat/completions` matches the prefix branch on the first pass and is left unchanged, although the first rewrite should produce `/v1/chat/chat/completions`. The added test only covers the non-overlapping `/v1/provider` case. Please implement the accepted issue’s immutable-ingress-path/single-owner approach (and strip any internal carrier before upstream), then add an overlapping-prefix regression plus repeated-application coverage.

Thanks for clarifying this — the overlapping-prefix example made the limitation of the alternation approach clear.
I’m updating the implementation to follow the immutable-ingress-path / single-owner approach:
- capture the original ingress :path before routing;
- keep it unchanged across route-cache recomputation;
- compute the provider upstream path from that immutable value;
- apply the final path exactly once after ext_proc;
- preserve the existing direct-endpoint prefix behavior; and
- remove all internal path-carrier headers before the request reaches the upstream.

I’ve also added coverage for overlapping prefixes, repeated application, query preservation, direct endpoints, and client-supplied carrier headers. I’m finishing the repository-required validation now and will push the update once the checks are complete. In the meantime, I’d be happy to incorporate any additional feedback on this approach.

Apricooooot:
@FAUST-BENCHOU @Xunzhuo Thanks again for the feedback. I’ve replaced the previous alternation approach with the requested immutable-ingress-path and single-owner design.

The update includes the overlapping-prefix and repeated-application regressions, strips all internal carriers before upstream, and preserves the existing direct-endpoint behavior. The PR body has the full design and validation details.

The branch is rebased, and all local unit, integration, build, and smoke checks passed. Could you please take another look when you have time? I’m happy to make any further adjustments.

Xunzhuo (commented):
Thanks for the update—the prior overlapping-prefix blocker is addressed. The current integration with `main` still does not compile: `processor_req_body_external_gateway.go` calls the removed `setProviderRequestPath` at line 67. That path intentionally mutates `:path` directly because external-gateway mode keeps the selected route and does not use the CLI Lua finalizer, so switching it blindly to the carrier helper would change behavior. Please retain a direct-path helper for external-gateway protocol conversion (or otherwise preserve that contract), update the new caller, and rerun the failed checks.

Xunzhuo (commented):
Welcome, and thanks for the update. The immutable ingress-path carrier now handles overlapping prefixes, the external-gateway direct-path contract is preserved, and the carrier headers are stripped before upstream. The focused unit, CLI, and E2E coverage closes the earlier blockers. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

Apricooooot:
@Xunzhuo Thanks for the follow-up. The branch is now rebased on the latest `main`, and all CI checks are green. Could you please take another look when you have time?

Apricooooot:
@Xunzhuo Thanks for the follow-up. The branch is now rebased on the latest `main`, and all CI checks are green. Could you please take another look when you have time?

Apricooooot:
@Xunzhuo @FAUST-BENCHOU  I’ve rebased onto the latest `main` and resolved the conflicts while preserving the updated upstream path-resolution behavior. All post-rebase checks are green. Could you please take another look when convenient? Thank you so much!

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -375,20 +375,52 @@ func overwriteRequestHeader(key, value string) *core.HeaderValueOption {
...
 			requestPath = configured
 		}
 	}
-	*headersOut = append(*headersOut, &core.HeaderValueOption{Header: &core.HeaderValue{
-		Key: ":path", RawValue: []byte(requestPath),
-	}})
+	requestPath = appendIngressQuery(requestPath, ctx)
+	*headersOut = append(*headersOut, &core.HeaderValueOption{
+		Header: &core.HeaderValue{
+			Key: headers.VSRUpstreamPath, RawValue: []byte(requestPath),
```
Xunzhuo: **Install the path finalizer in every supported Envoy chain**

Impact: Operator, local, KServe, Istio, OpenShift, benchmark, and Authorino deployments retain the ingress endpoint instead of the provider-resolved path. Cross-protocol or custom provider paths therefore reach the wrong endpoint, and internal carrier headers leak upstream.

Required change: Retain direct :path mutation unless a finalizer capability is negotiated, or add the Lua finalizer and carrier stripping to every supported ext_proc chain and generator, with representative non-CLI coverage.

Xunzhuo (commented):

src/semantic-router/pkg/extproc/processor_req_body_routing.go
Apricooooot: Good Catch! Thanks! I’ve addressed this by keeping direct `:path` mutation for external-gateway mode, while using the carrier only when the filter chain explicitly advertises finalizer support.

All checks are now green except `Core test and build`. Its local smoke test reached a healthy router, but `func-e` received an [HTTP 500 while downloading Envoy v1.35.4](https://github.com/vllm-project/semantic-router/actions/runs/34291293701/job/102278838583?pr=3393#step:24:48). I don’t appear to have permission to rerun upstream Actions jobs. Could you please rerun the failed job when convenient?

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -532,8 +579,24 @@ func (r *OpenAIRouter) applyDecisionHeaderMutations(state *routeHeaderState, ctx
...
+	}
+	for _, header := range removeHeaders {
+		if isReservedPathCarrier(header) {
+			continue
+		}
+		state.removeHeaders = append(state.removeHeaders, header)
+	}
+}
+
+func isReservedPathCarrier(header string) bool {
```
Xunzhuo: **Reserve path carriers across backend headers**

Impact: Public `backend_refs[].extra_headers` and `auth_header` accept these names without validation. With a non-empty `base_url` and `x-vsr-internal-path-needs-prefix: true`, profile headers preserve the spoofed marker while the new path code emits an already-resolved provider path, so Lua prefixes it a second time. The same extra header can also be projected as an Envoy route header after Lua removes carriers, leaking it upstream. This decision-only filter does not protect either config source.

Required change: Reject or filter all three reserved carrier names case-insensitively at every public provider/backend header boundary, including route projection and auth headers, and add focused double-prefix and upstream-leak regressions.

Xunzhuo (commented):
Thanks for fixing capability negotiation across the non-CLI chains. One remaining public header path can still spoof the new internal carriers.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
Apricooooot: Thanks, you’re right. The current sanitization covers request and decision-plugin headers, but not config-sourced `backend_refs[].extra_headers`, `auth_header`, or Envoy route-header projection. I’ll reserve all three carrier names case-insensitively at those public boundaries, keep runtime filtering as defense in depth, and add focused double-prefix and upstream-leak regressions.

Apricooooot: Fixed in [`33e3127`](https://github.com/vllm-project/semantic-router/commit/33e3127fc3f64188923d52c992c8eb4da01fec1c).

All three internal path-carrier names are now handled case-insensitively across:

- `backend_refs[].extra_headers`
- `backend_refs[].auth_header`
- runtime provider and decision-header mutation
- Envoy route-header projection

Invalid configured carriers are rejected where possible, with runtime and projection filtering retained as defense in depth. The added regressions cover mixed-case carrier names, the non-empty `base_url` plus spoofed prefix-marker double-rewrite case, and upstream carrier leakage while confirming ordinary provider headers remain unaffected.

The relevant unit, config-generator, Dashboard build, and targeted integration checks passed. The current unrelated E2E failure occurs before test execution due to the ONNX `TextWindows` regression tracked in #3672 and fixed by #3673.

Apricooooot:
Heads-up: after rebasing onto the latest `main`, the `external-gateway-responses` E2E job now fails while building the ONNX router, before the integration tests execute:

- [`candle_binding.TextWindows` / `TextWindow` are undefined in the ONNX build](https://github.com/vllm-project/semantic-router/actions/runs/34421296112/job/102697139562?pr=3393#step:9:4575)
- This is the current-`main` regression tracked by [#3672](https://github.com/vllm-project/semantic-router/issues/3672)
- The upstream fix is in [#3673](https://github.com/vllm-project/semantic-router/pull/3673)

This PR does not modify the RAG windowing or native binding code. Once #3673 is merged, I’ll rebase onto the fixed `main` and rerun the checks.

src/semantic-router/pkg/extproc/processor_req_header.go
```diff
@@ -43,7 +43,12 @@ func (r *OpenAIRouter) handleRequestHeaders(v *ext_proc.ProcessingRequest_Reques
 	// also short-circuit in the no-op path.
 	if ctx.SkipProcessing {
 		detectStreamingExpectation(ctx)
-		return newContinueRequestHeadersResponse(buildLooperInternalHeaderRemovalMutation()), nil
+		mutation := withoutPathCarriers(buildLooperInternalHeaderRemovalMutation())
```
Xunzhuo: **Preserve route prefixes for skipped requests**

Impact: A valid skip-processing request in the finalizer-capable CLI chain removes all path carriers here. Because this patch also removes route-level `regex_rewrite` and the Lua finalizer needs an original or upstream carrier to apply `path_prefix`, a direct endpoint such as `/compatible-mode/v1` now receives the ingress `/v1/chat/completions` instead of its configured prefixed path.

Required change: Preserve a sanitized trusted ingress path for finalizer-capable skipped requests, or otherwise let the finalizer apply the selected route prefix. Add a CLI integration regression for skip processing with a non-empty direct-endpoint prefix.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -532,8 +587,18 @@ func (r *OpenAIRouter) applyDecisionHeaderMutations(state *routeHeaderState, ctx
 		return
 	}
 	setHeaders, removeHeaders := r.buildHeaderMutations(ctx.VSRSelectedDecision)
-	state.setHeaders = append(state.setHeaders, setHeaders...)
-	state.removeHeaders = append(state.removeHeaders, removeHeaders...)
+	for _, header := range setHeaders {
+		if header.GetHeader() == nil || headers.IsInternalPathCarrier(header.GetHeader().GetKey()) {
```
Xunzhuo: **Keep path carriers out of shadow requests**

Impact: This filter protects the primary request, but `shadowExtraHeaders` rebuilds the same decision mutations separately and forwards allowlisted values. Since `forward_headers` currently accepts the three internal path-carrier names, a valid configuration can send private path state to the shadow upstream.

Required change: Reject all internal path carriers in shadow `forward_headers` validation and keep a defense-in-depth runtime filter. Add a regression with an allowlisted carrier mutation.

Xunzhuo (commented):
Thanks for the thorough rewrite. I found two remaining path-boundary gaps on the current head.

Xunzhuo (commented):
Thanks for the thorough fixes. One query-ownership gap remains on this head.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -496,20 +512,59 @@ func overwriteRequestHeader(key, value string) *core.HeaderValueOption {
...
+	ingressPath := ctx.Headers[headers.PseudoHeaderPath]
+	queryStart := strings.IndexByte(ingressPath, '?')
+	if queryStart == -1 || queryStart == len(ingressPath)-1 {
+		return requestPath
+	}
+	separator := "?"
+	if strings.Contains(requestPath, "?") {
+		separator = "&"
+	}
+	return requestPath + separator + ingressPath[queryStart+1:]
```
Xunzhuo: **Keep configured query keys authoritative**

Impact: An ingress request such as `?api-version=attacker` is appended after Azure's configured `api-version`, so the upstream receives two values and may select the client-controlled version or reject the ambiguous request. The same collision is possible for query keys embedded in a configured chat path.

Required change: Merge queries structurally, preserving provider/configured keys over duplicate ingress keys (or reject collisions), and cover both the finalizer-carrier and direct `:path` branches.

drivebyer (commented):
The root cause is two writers, not route recomputation: ext_proc already writes the full provider path from profile.BaseURL, then the route's regex_rewrite prefixes it again. A minimal fix is to drop regex_rewrite only on the x-selected-model routes and keep it on the default route for skip-processing. Benefits: a ~10-line template change, ext_proc stays the single path owner as in every other chain, no Lua finalizer, no capability negotiation, no new internal headers to guard, and no query-string behavior change.

Apricooooot:
> The root cause is two writers, not route recomputation: ext_proc already writes the full provider path from profile.BaseURL, then the route's regex_rewrite prefixes it again. A minimal fix is to drop regex_rewrite only on the x-selected-model routes and keep it on the default route for skip-processing. Benefits: a ~10-line template change, ext_proc stays the single path owner as in every other chain, no Lua finalizer, no capability negotiation, no new internal headers to guard, and no query-string behavior change.

Thanks for clarifying the rewrite timing and the actual two-writer root cause. I agree that keeping `ext_proc` as the sole path owner for selected-model routes is much simpler.
I also confirmed on the latest `main` that canonical endpoint URLs are materialized into the provider profile, so this approach preserves normal direct-endpoint paths as well. I’ll reduce the patch to remove `regex_rewrite` only from selected-model routes, retain it on the default route for skip-processing, revert the Lua/capability/carrier and query-handling changes, and keep focused regressions for the `/v1/provider`, overlapping-prefix, and skip-processing cases.

drivebyer (dismissed):
LGTM

src/vllm-sr/cli/templates/envoy.template.yaml
```diff
@@ -68,15 +68,9 @@ static_resources:
...
-                  {% if model.path_prefix %}
-                  # Prepend path prefix to /v1 requests (e.g., /compatible-mode + /v1/chat/completions = /compatible-mode/v1/chat/completions).
-                  # Match only the complete /v1 segment so route recomputation cannot rewrite /v1beta/... again.
-                  regex_rewrite:
-                    pattern:
-                      google_re2: {}
-                      regex: "^/v1([/?].*)?$"
-                    substitution: "{{ model.path_prefix }}\\1"
-                  {% endif %}
+                  # ext_proc has already resolved the complete provider path for
```
Xunzhuo, line 71: **Update the stale selected-route tests**\n\nImpact: Four existing catalog projection tests still index regex_rewrite on the selected route (test_catalog_provider_projection.py lines 111, 282, 351, and 570), so the declared pytest tests/ -v suite now raises KeyError even though the runtime rewrite is correct.\n\nRequired change: Update all four contracts to assert that selected routes have no rewrite and check the expected substitution on _default_route(...), matching the updated generator tests.

Xunzhuo (commented):
Thanks for fixing the duplicate rewrite. One repository test contract still needs to follow the new route ownership.

Xunzhuo (dismissed):

Xunzhuo:
Can you fix the CI @Apricooooot

Apricooooot:
Fixed and rebased — all CI checks are green now. When you have time, could you please review and approve the latest revision so it can be merged? Thanks! @Xunzhuo @drivebyer

src/vllm-sr/cli/templates/envoy.template.yaml
```diff
@@ -68,15 +68,9 @@ static_resources:
...
+                  # requests carrying x-selected-model, so this route must not
```
Xunzhuo, line 72: **Keep Looper dispatches off the rewriting fallback**

Impact: Looper calls to a backend with a `/v1`-prefixed base path still receive the prefix twice and reach the wrong upstream URL. The internal Looper request has no `x-selected-model` before ExtProc, and `buildLooperBackendDispatchResponse` writes the selected header and complete `:path` without clearing the route cache, so Envoy remains on this default rewriting route.

Required change: Route every provider-owned path mutation, including Looper hops, to a non-rewriting route (or restrict the fallback rewrite to a true bypass marker), and add a Looper/custom-prefix integration regression.

Xunzhuo (commented):
Thanks for updating the route contracts. One internal dispatch path still reaches the rewriting fallback.

src/vllm-sr/cli/templates/envoy.template.yaml
Apricooooot: @Xunzhuo Thanks again for the review. I’ve addressed the latest routing gap and rebased the PR onto the current `main`.

The final implementation keeps `ext_proc` as the single owner of provider-resolved paths, removes `regex_rewrite` only from selected-model routes, and preserves it on the default route for skip-processing requests. Looper and same-name entrypoint provider dispatch now also trigger route reevaluation.

The focused Go, config-contract, image-build, and six CLI integration checks all pass. Could you please take another look when you have time?

Xunzhuo (dismissed):
The selected-model routes preserve the complete provider path, and both entrypoint and internal Looper dispatch clear the cached fallback route. The earlier internal-dispatch gap is resolved; exact-head CLI and protocol E2E checks pass.

wilsonwu (dismissed):
LGTM

Xunzhuo (approved):
The provider-path and fallback-route contracts now agree, including skipped requests with a selected-model header and internal Looper hops. The exact-head Core regressions and all six path-routing integration cases pass in the 57-test CLI run.

Apricooooot:
@mergifyio queue

Apricooooot:
@Xunzhuo All checks are green and the PR is merge-ready, but Mergify does not allow me to run the `queue` command. Could you please requeue it when you have a chance? Thanks!

PR #3395, opened 2026-09-03
[Bug] Window hallucination context so the answer survives truncation

### LINKED ISSUE #3365: [Bug] Hallucination detection goes silent past 512 tokens of context
### Proposed Workgroup

### Describe the bug

`detect_hallucinations` (`ffi/classify.rs:1690`) runs `"{context} Question: {q} [SEP] {answer}"` through the token classifier with right truncation at 512. `context` is the unbounded `ctx.ToolResultsContext` (`req_filter_rag.go:323`, `req_filter_fact_check.go:69`). Once it fills the window, zero answer tokens survive, no span is found, and the result is `has_hallucination = false`. The NLI premise (`"{context} {question}"`) is truncated the same way.

### To Reproduce

Model `KRLabsOrg/lettucedect-base-modernbert-en-v1` (the registry target for `models/mom-halugate-detector`), threshold 0.5, true context repeated, answer with wrong year and height:

| context | has_hallucination | confidence |
|---|---|---|
| 208 chars | true | 0.993 |
| 2,080 chars | false | 1.000 |
| 4,160 chars | false | 1.000 |

### Expected behavior

The answer always reaches the model. Window the context, not the answer.

### Impact

Behind a real RAG or tool-calling pipeline, longer context means detection returns "no hallucination" with full confidence, so `action: block` never fires.

### Additional context

Verified on `origin/main` (2026-09-02) with a real model on CPU via candle-binding. Same tokenizer shows zero answer tokens survive truncation at that context length.

guan404ming:
Closes #3365

## Purpose

- Long context filled the 512-token window, so no answer token reached the detector.
- Add `fit_prefix_to_window` and window the context in `detect_hallucinations` and `classify_nli`.
- Touches `candle-binding`, a `pkg/classification` test, and the hallucination doc.

## Test Plan

- `cd candle-binding && cargo test --release --no-default-features --lib test_fit_prefix_to_window_keeps_suffix`
- `cd src/semantic-router && go test ./pkg/classification/ -run TestHallucinationDetector`

## Test Result

- Passed. The new long-context test fails on `main` and passes with the fix.

### DIFF
--- a/.github/workflows/community-labels.yml
+++ b/.github/workflows/community-labels.yml
@@ -19,12 +19,13 @@ jobs:
   pull-request-state:
     name: Synchronize pull request state
     if: >-
-      (github.event_name == 'workflow_run'
+      github.repository == 'vllm-project/semantic-router'
+      && ((github.event_name == 'workflow_run'
       && (github.event.workflow_run.event == 'pull_request'
       || github.event.workflow_run.event == 'pull_request_review')
       && github.event.workflow_run.pull_requests[0] != null)
       || (github.event_name == 'check_suite'
-      && github.event.check_suite.pull_requests[0] != null)
+      && github.event.check_suite.pull_requests[0] != null))
     # Serialize duplicate lifecycle events for one PR without blocking other PRs
     # or the scheduled queue reconciliation job.
     concurrency:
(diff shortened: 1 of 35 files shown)


FAUST-BENCHOU (dismissed):
LGTM. Nice catch on the long-context truncation issue.

Keeping the relevant suffix within the detector window makes the hallucination checks more reliable, and the added regression test covers the previous failure mode well.

wilsonwu (dismissed):
LGTM

AayushSaini101 (approved):

wilsonwu (approved):
LGTM

Issue #3329, opened 2026-09-02
[Bug] Config fields parsed and validated but never honored at runtime


guan404ming:
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

#3099 and #3315 each found one field the loader accepts and the runtime ignores. A sweep of every `yaml`-tagged field in `pkg/config` finds more:

- [x] RAG connector timeouts (apply #3321's `GetTimeout()` pattern) - #3396
  - [x] External API: `timeout_seconds` sets only the caller context (`req_filter_rag_external.go:170`); the shared client hardcodes 30s (`:33`)
  - [x] OpenAI: `timeout_seconds` (`rag_plugin_backends.go:43`) never read; `openai/vectorstore.go:39` hardcodes 60s, `filestore.go:27` 300s
- [x] `api.batch_classification.max_batch_size`: `route_classify.go` never bounds `len(req.Texts)` - #3457
- [x] Cache tuning fields, honor or remove - #3466, #3488
  - [x] Milvus `performance.*`, `data_management.*`, `enable_query_log`, `enable_metrics`, `verbose_errors` (`runtime_config.go:181-209`)
  - [x] Redis and Valkey `verbose_errors`, `enable_query_log`, `enable_metrics` (`runtime_config.go:84-133`)
- [x] Memory `quality_scoring.*` and `extraction_batch_size` (`runtime_config.go:254-277`) - #3466, #3488
- [x] Generic `type: llm` signal ignores `parser_type`; only guardrail reads it (`vllm_classifier.go:66`). Noted in #3099, not in #3143 - #3353
- [x] Single fields, honor or remove: `image_gen.prompt_prefixes`, `feedback_mapping_path`, `windowed_metrics.model_metrics`, `output_contract.on_violation/field_order/repair`, Milvus RAG `metadata_field`, `model_selection.momentum.*`, `ml.feature_weights` - #3467, #3458, #3488, #3422

Method: static grep for `.Field` reads outside `pkg/config`. Excluded: fields serialized whole (`response_cache.revision.*`), fields the vllm-sr Envoy template consumes (`reliability.health_check_*`), and `model_admission.*` (#3268 wires it).

### To Reproduce

Set any field above and observe no change. Sharpest: `rag.backend: external_api` with `timeout_seconds: 60` against a 40s endpoint fails at 30s with `context deadline exceeded`.

### Expected behavior

[...]

PR #3396, opened 2026-09-03
[Bug] Honor configured timeout_seconds in RAG connectors

guan404ming:
Related #3329

## Purpose

- External API RAG capped `timeout_seconds` at a hardcoded 30s client; OpenAI RAG never read it.
- Add `GetTimeout()` on both RAG configs and thread it into the HTTP clients.
- Touches `pkg/config`, `pkg/extproc`, and `pkg/openai`.

## Test Plan

- `cd src/semantic-router && go test ./pkg/config/ ./pkg/openai/ ./pkg/extproc/`

## Test Result

- Passed.


src/semantic-router/pkg/config/rag_plugin_backends.go
```diff
@@ -1,6 +1,20 @@
...
+	"fmt"
+	"time"
+)
+
+const DefaultExternalRAGTimeoutSeconds = 30
+
+const DefaultOpenAIRAGTimeoutSeconds = 60
+
+func ragTimeout(seconds *int, defaultSeconds int) time.Duration {
+	if seconds == nil || *seconds <= 0 {
```
abhinav-m22, line 16: If a user explicitly sets `timeout_seconds: 0` in their config because they want an infinite timeout, this logic overrides it to the 30s default. Is preventing a 0s (unbounded) timeout intentional here to protect against hanging requests, or should we respect `0` as 'no timeout'?

guan404ming: This is Intentional. 0 never meant unbounded here: before this change it produced an already-expired context under the 30s client cap, and #3321's GetTimeout treats non-positive the same way. An unbounded upstream call would reopen the hang this family of fixes closes, so a value that large should be set explicitly.

abhinav-m22: Makes sense, thanks for the context on this :)

Xunzhuo (dismissed):
Two compatibility blockers remain. `ragTimeout` accepts arbitrarily large positive seconds, so `time.Duration(seconds) * time.Second` can overflow negative and disable `http.Client.Timeout`; please validate or safely clamp the configured range before conversion. Also, changing the exported three-argument `NewVectorStoreClientWithSearchResponseLimit` signature breaks downstream Go callers. Please preserve that constructor as the compatibility wrapper and add a separately named timeout-aware constructor or options path.

guan404ming:
Thanks! `ragTimeout` now clamps to `MaxRAGTimeoutSeconds` before conversion. The three-arg constructor is back as a wrapper; the timeout-aware path is `NewVectorStoreClientWithTimeout`.

Xunzhuo (commented):
The current head safely clamps the duration before conversion and preserves the exported three-argument constructor as a compatibility wrapper. Both RAG request paths now use the configured timeout. I do not see a remaining code blocker; please update onto current main for the final integration check.

Xunzhuo (approved):

wilsonwu (approved):
LGTM, thanks!

Issue #3178, opened 2026-08-31
[Test] Cover supported routing signals and projections end to end


Xunzhuo:
### Goal

Protect every supported routing signal and projection family with a meaningful end-to-end behavior contract, while using the coverage work to surface implementation bugs and unreasonable public API behavior.

Parent Epic: #1519

### Proposed Workgroup

Evaluation & Quality

### Scope

- Derive the inventory from the runtime catalogs rather than a hand-maintained list, including every type returned by `config.SupportedSignalTypes()`.
- Add focused positive and negative cases for each supported signal. A configuration-load or HTTP-200 smoke does not count as behavior coverage.
- Cover projection partitions, weighted scores, value sources, dependency ordering, threshold-band and multi-emit mappings, defaults, and calibration behavior.
- Assert externally useful outcomes such as the selected decision/model and the corresponding debug or Router Replay evidence without expanding the default public header surface.
- Use deterministic fixtures and land coverage in bounded slices so one signal or projection family can be reviewed independently.
- When a case exposes incorrect implementation or an unreasonable API contract, open a focused issue under the Workgroup that owns that behavior, link it here, and add the regression case with the fix.

### Non-goals

- Exhaustively test the Cartesian product of every signal and projection.
- Turn routing-model accuracy evaluation into E2E acceptance coverage.
- Fix unrelated product bugs inside a broad test-only pull request.
- Duplicate focused unit tests that do not need a deployed request path.

### Completion signal

- Every current supported signal and each projection family has at least one runnable behavior contract in the catalog-driven coverage matrix from #2379.
- Applicable negative, unsupported, and failure paths are represented.
- No configuration-only or report-only probe is counted as acceptance coverage.
- Discovered implementation and API gaps have focused owning issues and regression coverage.
- Adding a new signal or projection without a coverage mapping is rejected by the maintained inventory gate.

yu3zhang1:
As the first PR for this issue, I'll add end-to-end test for the `event` signal (config.SignalTypeEvent), which currently has no e2e coverage. The EventClassifier is pure regex/rule matching against EventRule config (event_types, severities, action_codes, temporal) with no ML model dependency, so I'll add a new e2e/testcases/event_routing.go file following the existing pkgtestcases.Register(...) pattern used in keyword_routing.go, covering: 
* One positive case (a request whose text matches a configured event_type/severity/action_code, confirmed via the router's decision and matched-rule evidence) and
[...]

PR #3399, opened 2026-09-03
[Test] Add E2E coverage for the event signal

yu3zhang1:
Related #3178

## Purpose

The `event` routing signal (`config.SignalTypeEvent`) currently has only
Go unit tests for its classifier logic (`event_classifier_test.go`) and
no end-to-end behavioral coverage through a deployed request path, as
required by #3178.

This PR adds:
- A `critical_payment_event` event rule (`event_types`, `severities`,
  `action_codes`, `temporal`) and a `critical_event` decision (priority 35)
  to `e2e/profiles/ai-gateway/values.yaml`.
- `e2e/testcases/event_routing.go`, following the existing
  `pkgtestcases.Register(...)` pattern used by `keyword_routing.go`,
  asserting the `x-vsr-selected-decision` and `x-vsr-matched-event`
  response headers.
- `e2e/testcases/testdata/event_routing_cases.json` with two positive
  cases (event_type-only match; severity + action_code match) and one
  negative case (no configured criteria present).

Owning Workgroup: Evaluation & Quality (`wg/evaluation-quality`).

## Test Plan

- `gofmt -l e2e/testcases/event_routing.go`
- `go vet ./testcases/...` (run from `e2e/`)
- `go build ./...` (run from `e2e/`)
- `make e2e-test-specific E2E_TESTS="event-routing"`

## Test Result

- gofmt / go vet / go build: pass.
- `make e2e-test-specific`: blocked locally by a Docker Desktop
  compatibility issue with the `quay.io/centos/centos:stream10` base
  image used by `Dockerfile.extproc` (reproduces even with a plain
  `docker run` on that image, independent of this change, and appears
  to be a local/CDN-cache-specific issue — the same image builds
  successfully in this repo's own recent GitHub Actions runs). Deferring
  final e2e confirmation to CI on this PR.
- Manually verified the priority ordering in `values.yaml`: the new
  `critical_event` decision (priority 35) sits above all domain
  decisions (priority 10) and below the decisions that require
  unrelated unique trigger keywords (sensitive_data, tool_selection_*,
  jailbreak, PII), so the new test cases can't be shadowed by an
  unrelated decision.

---
AI assistance (Claude Code) was used to draft this change, under my review;
I read and understand the diff and take responsibility for it.

### DIFF
--- a/.github/workflows/integration-test-k8s.yml
+++ b/.github/workflows/integration-test-k8s.yml
@@ -87,7 +87,7 @@ jobs:
           set +e
           if [ "${{ matrix.profile }}" = "envoy-ai-gateway" ]; then
             # Temporarily skip the stress / pressure coverage until the suite is stable again.
-            ENVOY_AI_GATEWAY_CI_TESTS="chat-completions-request,apiserver-runtime-config-endpoints,domain-classify,semantic-cache,semantic-cache-polarity,pii-detection,jailbreak-detection,decision-priority-selection,plugin-chain-execution,tool-selection,rule-condition-logic,decision-fallback-behavior,plugin-config-variations"
+            ENVOY_AI_GATEWAY_CI_TESTS="chat-completions-request,apiserver-runtime-config-endpoints,domain-classify,semantic-cache,semantic-cache-polarity,pii-detection,jailbreak-detection,decision-priority-selection,plugin-chain-execution,tool-selection,rule-condition-logic,decision-fallback-behavior,plugin-config-variations,event-routing"
             make e2e-test E2E_PROFILE=${{ matrix.profile }} E2E_TESTS="${ENVOY_AI_GATEWAY_CI_TESTS}" E2E_VERBOSE=true E2E_KEEP_CLUSTER=false
             TEST_EXIT_CODE=$?
             set -e
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -47,6 +47,8 @@ var BaselineRouterContract = []string{
 	"session-telemetry-metrics",
 	"session-pricing-chat-completions",
 	"session-pricing-response-api",
+	// Event signal rule matching and routing (issue #3178)
+	"event-routing",
 }
 
 // DashboardContract is the canonical E2E contract for the dashboard API surface.
--- a/e2e/profiles/ai-gateway/values.yaml
+++ b/e2e/profiles/ai-gateway/values.yaml
@@ -680,7 +680,6 @@ config:
             - critical
           action_codes:
             - TXN_DECLINE
-          temporal: true
       domains:
         - name: business
           description: Business, corporate strategy, management, finance, marketing
--- a/e2e/profiles/ai-gateway/values.yaml
+++ b/e2e/profiles/ai-gateway/values.yaml
@@ -574,6 +574,24 @@ config:
               enabled: true
               system_prompt: You are handling a query with sensitive data. Be cautious and provide security-focused guidance.
               mode: replace
+      - name: critical_event
+        description: Critical operational events requiring specialized handling
+        priority: 35
+        rules:
+          operator: OR
+          conditions:
+            - type: event
+              name: critical_payment_event
+        modelRefs:
+          - model: base-model
+            lora_name: general-expert
+            use_reasoning: false
+        plugins:
+          - type: system_prompt
+            configuration:
+              enabled: true
+              system_prompt: You are handling a critical event alert. Prioritize a fast, precise, and factual response.
+              mode: replace
       - name: other_decision
         description: General knowledge and miscellaneous topics
         priority: 1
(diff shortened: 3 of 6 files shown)


FAUST-BENCHOU (commented):
Should this fail whenever `correctTests != totalTests` instead? As written, the E2E still passes if only one of the test cases succeeds, which could hide routing regressions.

yu3zhang1:
Good catch, it should be totalTests. I copied from `keyword_routing.go` without checking it applied here, it seems keyword matching is fuzzy and tolerance makes sense, but event matching is deterministic.

Fixed in 2dfe2dac: now requires `correctTests == totalTests`. Waiting on the new E2E run to confirm, will update here once it's done.

Xunzhuo (changes requested):
Welcome, and thanks for adding the event-signal cases and fixing the all-cases assertion. One blocker remains: `event-routing` is registered, but it is not included in `testmatrix.BaselineRouterContract` or any profile’s `GetTestCases()`. The changed `ai-gateway` profile still returns only the baseline contract, so normal E2E runs never execute this test and the green E2E check does not validate it. Please add the case to the intended profile/test matrix and add a contract test that proves the profile selects it.

yu3zhang1:
> Welcome, and thanks for adding the event-signal cases and fixing the all-cases assertion. One blocker remains: `event-routing` is registered, but it is not included in `testmatrix.BaselineRouterContract` or any profile’s `GetTestCases()`. The changed `ai-gateway` profile still returns only the baseline contract, so normal E2E runs never execute this test and the green E2E check does not validate it. Please add the case to the intended profile/test matrix and add a contract test that proves the profile selects it.

Done. Added `event-routing` to `BaselineRouterContract` and a new contract test proving envoy-ai-gateway selects it.

One question on placement:  I put the new contract test in a separate file other than `imports_test.go`, since more signal tests will need the same check, and I didn't want imports_test.go to become a catch-all test file. I can move it if you'd rather keep this pattern — let me know which you prefer.

yu3zhang1:
I checked the CI still not ran event-routing:`.github/workflows/integration-test-k8s.yml` hardcodes a separate `ENVOY_AI_GATEWAY_CI_TESTS` list that doesn't read from BaselineRouterContract at all. 

It looks like it was intended for CI stability (the comment above it "Temporarily skip the stress / pressure coverage until the suite is stable again."). Should event-routing be added there too, or is there a different BaselineRouterContract is meant to be consumed?

FAUST-BENCHOU (dismissed):
For the negative cases, should we also assert the expected fallback decision instead of only checking that it is not `critical_event`?

This would make the E2E coverage catch incorrect routing to another event-related decision in the future.
others lg

Xunzhuo (commented):
Thanks for the update. The profile-selection blocker is fixed, but `event-routing` still is not exercised by the required CI path: `.github/workflows/integration-test-k8s.yml` overrides the profile contract with `ENVOY_AI_GATEWAY_CI_TESTS`, and that list omits the new case. The green E2E job therefore did not run the behavior this PR adds. Please add `event-routing` to that list (or remove the override in a separately justified change) and rerun the E2E job. The negative-case fallback assertion can stay a non-blocking follow-up.

yu3zhang1:
> Thanks for the update. The profile-selection blocker is fixed, but `event-routing` still is not exercised by the required CI path: `.github/workflows/integration-test-k8s.yml` overrides the profile contract with `ENVOY_AI_GATEWAY_CI_TESTS`, and that list omits the new case. The green E2E job therefore did not run the behavior this PR adds. Please add `event-routing` to that list (or remove the override in a separately justified change) and rerun the E2E job. The negative-case fallback assertion can stay a non-blocking follow-up.

Done, CI re-ran with event-routing in `ENVOY_AI_GATEWAY_CI_TESTS`, and it's now executed:
```
EVENT ROUTING TEST RESULTS
Total Tests: 3
Correct: 3 (100.00%)
```

Run link: https://github.com/vllm-project/semantic-router/actions/runs/33840917546/job/100923150173

Will follow up on @FAUST-BENCHOU's negative-case fallback suggestion separately.

yu3zhang1:
> For the negative cases, should we also assert the expected fallback decision instead of only checking that it is not `critical_event`?
> 
> This would make the E2E coverage catch incorrect routing to another event-related decision in the future. 

Thanks for the suggestion! 

Because it already has `expected_matched_event: ""`, which checks `x-vsr-matched-event` is empty. If this query got mis-routed to any event rule, then header would no longer be empty, so we can catch it regardless of which decision it ended up. 

My concern to use an exact decision name: this query's actual routing depends on the `domain` classifier, which is not deterministic, it only returns `other` when its confidence is low. If we assert an exact decision here, the test could fail just because the classifier picked a different domain.

(FWIW, in this latest CI run, it did land on other_decision, but I am not sure that's guaranteed.)

So I'll leave it as-is, let me know if you feel strongly otherwise.

Xunzhuo (approved):
LGTM. Thanks!

Issue #3382, opened 2026-09-02
[Feature] Migrate the remaining candle models to the shared chunked attention kernel


subin9:
### Goal

`chunked_sdpa` already exists (`model_architectures/attention/chunked_sdpa.rs`) and mmBERT embedding is the only model that calls it. Migrate the rest so attention memory stops growing with the square of the input.

#2014 asked for exactly this and proposed the API that later landed. It was closed by the stale bot without review. #2007 is the mmBERT fix, #1957 the crash that prompted it.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

Measured on CPU, one embedding call per row, peak RSS above the loaded weights:

```
              Qwen3 (dense, F64)        mmBERT (chunked)
words       delta MiB   elapsed       delta MiB   elapsed
512                 0     5.3s                0      1.1s
2048             2235    60.7s                0      8.9s
4096             9461   231.0s              116     31.1s
8192    stopped above 15 GiB                502    108.5s
```

Qwen3 casts Q and K to F64 before the matmul (`qwen3_embedding.rs:1232-1244`), which doubles it. The two models differ in size and head count, so the absolute numbers are not a controlled comparison of the kernel; the scaling shape is.

Sites still on dense attention: `qwen3_embedding.rs:1184,1244`, `candle_models/modernbert.rs:209`, `gemma3_model.rs:881,948`, `multimodal_embedding.rs:265,533,671,890`, `qwen3_with_lora.rs:265`, `base_model.rs:480`. Only Qwen3 and multimodal were measured; the rest are read from the source.

`candle_models/modernbert.rs:209` is the one worth starting from. Every ModernBERT and mmBERT classifier loads that model (`traditional/modernbert.rs:722,935,1422`), and `MAX_CLASSIFICATION_SEQ_LEN = 512` exists because of it — the comment there says 4000 tokens OOM-kills the container with no logs. That cap is what #3204, #3333 and #3364 each work around.

### Non-goals

- Raising or removing `MAX_CLASSIFICATION_SEQ_LEN`. That needs its own measurement once attention is bounded, and it changes classifier behaviour.
- The per-surface chunking fixes. #3206, #3334 and #3369 own those.
- The multimodal text FFI failing past its position limit, which is #3381.

### Completion signal

- Each migrated model returns the same output as its dense path, the way `chunked_sdpa_test.rs` already checks for mmBERT.
- Peak memory for a long input grows with the input rather than its square, measured the same way as above.

PR #3402, opened 2026-09-03
[Feature] Move ModernBERT attention onto the shared chunked kernel

subin9:
Related #3382

## Purpose

Move `candle_models/modernbert.rs` onto the shared `chunked_sdpa` kernel, which until now only mmBERT embedding used. Every ModernBERT and mmBERT classifier loads this backbone, and its quadratic attention is why `MAX_CLASSIFICATION_SEQ_LEN` is 512.

Three things were `O(seq²)` per forward and are now gone:

- the full `(b, heads, seq, seq)` score matrix in `compute_standard_attention`
- the `(b, 1, seq, seq)` padding mask expansion in `prepare_4d_attention_mask`
- the `(seq, seq)` sliding-window band in `get_local_attention_mask`

The kernel walks the query dimension in blocks, the padding mask becomes `(b, 1, 1, seq)`, and a local layer becomes `window: Some(w)` applied per block. Output is unchanged and no config, API or FFI surface moves — `ModernBert::forward(xs, mask)` keeps its signature.

One detail: `q` was scaled once for both the flash and dense paths. The kernel folds the scale in itself, so that binding now lives inside the flash arm as `q_scaled` — sharing it would have scaled `q` twice on the flash-failure fallback.

Qwen3, Gemma3, multimodal, LoRA and `base_model` stay dense and remain with #3382.

## Test Plan

```
cd candle-binding && cargo test --release --no-default-features --lib modernbert::tests
make test-rust-ci
cd candle-binding && cargo test --release --no-default-features --lib   # compared against main
cd candle-binding && cargo fmt --check && cargo check --no-default-features
```

Memory measured with a throwaway harness (not committed): one attention block at hidden 768 (12 heads × 64), batch 1, CPU, F32, reporting `VmHWM` above the pre-forward `VmRSS`.

## Test Result

Two new tests compare the new path against the dense code this PR deletes, over global and local layers, sequence lengths 1/5/16/40, block sizes 1/3/8/16/512, with and without padding. `max|Δ| < 1e-4`, both pass. Both are registered in `RUST_CI_LIB_TESTS` so `make test-rust-ci` runs them — 7/7 pass.

Regression sweep: 522 passed / 69 failed, and the failing set is identical to `main` measured the same way (all 69 need `../models`, absent here). `cargo fmt --check` and `cargo check --no-default-features` clean.

Three runs per cell; memory varied by at most 1 MiB.

```
                 peak RSS (MiB)        elapsed (s)
seq    kind     dense   chunked      dense  chunked
2048   global    1037       253       0.88     0.88
4096   global    3992       503       3.08     3.20
8192   global   15789      1003      11.67    12.44
2048   local     1057       114       0.89     0.26
4096   local     4056       162       3.22     0.51
8192   local    16045       258      11.86     0.99
```

Dense quadruples per doubling, chunked doubles. Global layers cost about 3% more wall clock (the block loop's `cat` and per-block `contiguous`); local layers get faster because the kernel reads only the `±window` band instead of masking a full matrix.

The flash-attn arm is CUDA-gated and cannot be built here or in PR CI (`cargo check --features flash-attn` fails at `nvcc`), so the `q_scaled` change is verified by reading. This does not raise `MAX_CLASSIFICATION_SEQ_LEN`, a non-goal in #3382.


FAUST-BENCHOU (commented):
It looks like the flash path scales `q` manually and also passes the same `softmax_scale` to `flash_attn`. Could this apply the `1/sqrt(d)` scaling twice here?

subin9:
Yes — though it predates this PR: q is already scaled before the flash call and softmax_scale is passed as well. This PR only stopped the chunked fallback from sharing that scaled binding. I'll drop the manual scale and check it if you need it :)

subin9:
oh, found where this came from — upstream candle-transformers only has the dense path and pre-scales `q`, which is right there since nothing scales again after. 

This file is a local copy with flash added on top, and the pre-scale sits above the branch, so both paths inherited it.

Dropped it on the flash side so `softmax_scale` does the scaling. dense/chunked unchanged.

Xunzhuo (approved):
The chunked path preserves global/local and padding semantics against the deleted dense reference, removes the quadratic mask/score allocations, and the latest head also fixes the flash path’s inherited double scaling. The focused Rust matrix is registered in CI, all checks are green, and local agent validation passes. Approved; please rebase onto current main before merge.

FAUST-BENCHOU (approved):

wilsonwu (approved):
LGTM

PR #3416, opened 2026-09-03
[Bug] Model context_management and forward it to Anthropic backends

### LINKED ISSUE #3413: [Bug] Anthropic requests carrying context_management fail decode, breaking Claude Code
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Claude Code sends a top-level `context_management` field on `/v1/messages` requests (Anthropic's context-editing beta: the client asks the API to trim old thinking blocks). The neutral request model does not name the field, and ingress decode is strict — one unknown field refuses the whole request. The router answers 400 `invalid inference request` and the session fails before any routing happens.

### To Reproduce

1. Route a model through the router and point Claude Code at it via `ANTHROPIC_BASE_URL`.
2. Run a session with thinking enabled (the client default). Requests carry `"context_management": {"edits": [{"type": "clear_thinking_20251015", ...}]}` and the router rejects them.

### Expected behavior

The request decodes, and the field reaches backends that can use it. This field tells the API to drop old thinking blocks from the billed prompt, so it changes what the user pays. When the dispatch target speaks the Anthropic protocol, the router should model the field and emit it on the dispatch request. A request that byte-replays already carries it in the original bytes; modeling it makes the re-encoded path behave the same instead of silently disabling the cost saving on every routed request. When the target speaks a different protocol there is nothing to carry it in, and dropping it there is normal translation loss.

### Impact

Claude Code sessions fail with a 400 on any request that carries the field. Both versions I tested send it (2.1.191 from June and current 2.1.258), so this is not tied to one release. Any Anthropic SDK user of the context-editing beta hits the same wall.

With simple naive accept-and-drop fix sessions would work, but every routed request would silently lose a cost-reduction instruction the client explicitly sent, and nothing would tell the user their prompts stopped being trimmed.

### Additional context

The strict decode is doing what it is designed to do — refuse what the neutral model does not name. The issue is that this field is routine traffic from a stock client, and one whose loss has a billing effect, which is why the proposal is model-and-forward rather than accept-and-ignore. Preservation-shaped handling of provider-specific data has precedent in the nvext work (#3136).

Verified with real Claude Code sessions against the live Anthropic API through the codec, and with a minimal hand-written request. Router at upstream main @ c47e1ebfb (2026-09-03).

dorota-memex:
Closes #3413 

## Purpose

Claude Code sends a top-level `context_management` field on `/v1/messages` requests (Anthropic's context-editing beta: it asks the API to trim old thinking from the billed prompt, so it affects what the user pays). The strict request decode refuses the unknown field, so the router answers 400 `invalid inference request` before routing and the session fails.

The field asks the API to trim old thinking blocks from the billed prompt, so it affects what the user pays — accept-and-drop would stop the 400 but silently disable that saving on every routed request. The fix instead models the field on the neutral request as an opaque carry and emits it on Anthropic-format dispatch, so backends that understand it still receive it whether the request replays or re-encodes. Cross-format targets cannot carry it and omit it as normal translation loss.

The field inventories record this as a new `preserved` disposition: accepted and carried verbatim to same-format targets, never interpreted and never invented cross-format. It is reserved for fields whose omission has a real consequence (here billing); the closure tests keep membership deliberate.

## Test Plan

Two layers of tests: `pkg/protocolcodec` pins that the request decodes, that an Anthropic-format encode carries the field through, and that an OpenAI-format encode omits it; `pkg/extproc` drives the production dispatch path (`prepareProtocolRequest` -> `encodeDispatchRequest`) and pins that the dispatch body headed to an Anthropic backend still carries the field, on both the byte-replay and the forced re-encode routes.

- `go test ./pkg/protocolcodec/`
- `go test ./pkg/extproc/`
- `go test ./pkg/llmprotocol/`
- gofmt / go vet clean.

## Test Result

- All three suites green; the draft's regression tests went red-on-main to green with the fix commit.
- gofmt / go vet clean.

---
<details>

</details>


FAUST-BENCHOU (dismissed):
LGTM. Nice fix for preserving Anthropic-specific request semantics through the routing pipeline.

Keeping `context_management` as an opaque carried field avoids coupling the router to provider-specific details, while the replay and re-encode tests ensure the field is preserved only when the target protocol supports it.

wilsonwu (dismissed):
LGTM

Xunzhuo (dismissed):
Thanks for following up on the earlier thread, but the strict-policy behavior is still unchanged. Both OpenAI encoders call `appendDroppedDiagnostic`, which does not inspect `LossyFeatures`; under `DefaultPolicy()` the Anthropic context/billing directive is therefore still silently removed even though lossy translation is configured to reject. The new test currently cements that success path. Please route this through `appendLossy`, propagate the error, and cover default rejection plus permissive translation with a diagnostic.

dorota-memex:
@Xunzhuo, thank you for the review. Changes are made, and I can see how this is a consistent boundary: the directive is request semantics the client asked for, so whether its loss is tolerable belongs to the policy, not to a per-field decision in the codec. The cross-format omission now goes through appendLossy and the error propagates out of both OpenAI encoders. The default policy rejects the route with lossy_translation, and LossyAllowWithDiagnostic keeps the translation with the omission diagnostic. The request tests pin both branches and the capability goldens record the default rejection.

One consequence I want to flag clearly: the policy seems to not be configurable, so with this change Claude Code remains incompatible with the router for OpenAI-format routing. The extproc engine always uses DefaultPolicy() (processor_protocol_contract.go). Claude Code sends context_management on effectively every request, and every such request routed to an OpenAI-format backend is now rejected before dispatch. It would work if an operator could explicitly opt into LossyAllowWithDiagnostic, but today there is no way to do that. If you agree, I can file a follow-up issue for making the lossy policy configurable; that feels out of scope for this PR.

Xunzhuo (commented):
The current head now applies the lossy policy to both OpenAI encoders, handles the response field, and verifies buffered and streaming forwarding through the Anthropic shim. I do not see a remaining code blocker. Please update onto current main for the final integration check.

Xunzhuo (changes requested):
Thanks—the lossy-policy fixes are solid. One protocol blocker remains: Anthropic puts streaming `context_management` at the top level of the final `message_delta`, but `anthropicEventWire` does not model it, so strict decoding rejects a valid stream as `invalid_upstream_json`. The new test nests the field under `message_start.message`, so it misses the production shape. Please handle the final-delta field and cover that event shape.

drivebyer (approved):
LGTM

FAUST-BENCHOU (approved):

wilsonwu (approved):

Xunzhuo (approved):
The request, response, and streaming mappings handle context_management consistently, including the terminal message_delta field. The earlier strict-decoding blocker is resolved, and the exact-head protocol integration checks pass.

src/semantic-router/pkg/protocolcodec/testdata/golden/capability/018-anthropic-official-response-fields-chat-out.json
```diff
@@ -72,6 +72,45 @@
...
+          "model": "routed-model",
+          "object": "chat.completion",
+          "usage": {
+            "completion_tokens": 1,
+            "completion_tokens_details": {
+              "reasoning_tokens": 0
+            },
+            "prompt_tokens": 2,
+            "prompt_tokens_details": {
+              "cached_tokens": 0
```
Xunzhuo, line 101: [P2] Refresh the context-management golden for current main

With current main merged, `go test ./pkg/llmprotocol ./pkg/protocolcodec` fails in `TestGoldenCapabilityMatrix/018-anthropic-official-response-fields_to_chat`: this new case expects only `cached_tokens`, while the encoder now also emits `cache_write_tokens: 0`. The same case passes at this PR's standalone head. Please regenerate this golden after updating onto main so the protocol test gate passes.

Xunzhuo (changes requested):
The context-management and Anthropic presence checks pass, but integrating this branch with current main breaks one capability golden. Please refresh the new Chat Completions expectation for the current cache-usage output.

PR #2741, opened 2026-08-01
[Router][Dashboard] Add first-class Mixture-of-Models recipes

Xunzhuo:
Closes #2331

## Summary
- make entrypoints and named recipes a complete public routing contract, including entrypoint-only catalogs, recipe-aware eval, CLI migration, and validation
- apply selector weights, reasoning controls, replay, learning, memory, and eval snapshots consistently to named-recipe decisions
- add a maintained five-profile Mixture-of-Models recipe while keeping provider models and machine-local AMD YAML out of the public catalog and PR
- add a first-class Dashboard Mixture-of-Models manager, centered topology view, accessible catalog-driven Playground picker, and guarded recipe/entrypoint lifecycle

## Test plan
- [x] `make dashboard-check` (91 frontend files / 290 tests plus type-check, lint, and backend checks)
- [x] `go test ./pkg/config -count=1`
- [x] native-linked Router regression: services race tests plus named-recipe selector, reasoning, and replay tests
- [x] `make vllm-sr-test` (39/39 CLI E2E tests after isolated container cleanup)
- [x] changed-files lint: pre-commit, Ruff, golangci-lint, YAML, JS/TS, structure, and AST security checks
- [x] `make vllm-sr-build`
- [x] local router eval matrix: 12/12 routing cases without backend calls
- [x] local AMD smoke: five-model catalog, health, Dashboard, and 5/5 live entrypoint routes
- [x] named-recipe Router Replay API smoke
- [x] three independent Router, CLI, and Dashboard defect reviews, followed by closure reviews with no remaining functional findings

## Notes
- `pytest -q src/vllm-sr/tests` passes 475 tests; four pre-existing Dockerfile surface assertions fail against unchanged main Dockerfiles because they expect older literal base-image strings.
- The full Router gate completed functional suites; two existing classifier integration performance thresholds exceeded their timing limits on this loaded host.
- `.local/**`, the AMD subscription key, generated runtime state, and the machine-local AMD YAML are ignored and are not part of this PR.

PR #3421, opened 2026-09-03
[Bug] Size signal windows by estimated tokens

### LINKED ISSUE #3367: [Bug] Signal windows overflow the 512-token cap on dense and CJK text

guan404ming:
Closes #3367

## Purpose

- Estimate tokens per rune, calibrated on the mmBERT tokenizer, instead of counting runes.
- Cut security chunks at whitespace via prefix sums so no chunk exceeds 512 tokens.
- Size the head/middle/tail window by the same estimate so CJK keeps its tail.

## Test Plan

- `go test ./pkg/classification/`
- Real tokenizer over 10 scripts: every chunk and window ≤ 512 tokens.
- Real models on hex+attack and CJK law+math prompts.

## Test Result

- Passed. Max chunk 349 tokens, max window 459, 0 violations.
- Model still dilutes a short attack inside a 300-token noise chunk; not a truncation issue.


src/semantic-router/pkg/classification/classifier_signal_text_window.go
```diff
@@ -7,21 +7,14 @@ import (
...
-// is local token classification and loses entity confidence in long windows;
-// jailbreak detection benefits from broader semantic context. Keep separate
-// budgets instead of forcing both security signals through one compromise.
-// Units are quarter-token estimates: ordinary runes cost one, word starts add
-// one, and CJK runes cost six in total (about 1.5 tokens).
+// allocate against that same unbounded window. Units are quarter-token
+// estimates calibrated on the mmBERT tokenizer.
 const (
+	semanticSignalUnitLimit     = 440 * 4
 	piiSignalChunkBudget        = 128 * 4
```
AayushSaini101, line 18: @guan404ming On what basis we are deciding these numbers ? thanks : )

guan404ming: 440 is the 512 cap minus:

- the two omission markers, ~20 tokens each
- `[CLS]` / `[SEP]`
- a ~30-token buffer for estimate error

128, 384, and the 64-rune overlap are unchanged from #2741. Per-rune costs were fitted on the real mmBERT tokenizer over 10 scripts so estimates never undercount. Measured over 30 inputs: max window 459 tokens, max chunk 349, none over 512.

src/semantic-router/pkg/classification/classifier_signal_text_window.go
```diff
@@ -142,44 +139,75 @@ func securitySignalChunkSpans(text string, budget, overlapRunes int) []signalChu
...
-		return 1, true
+		return units, true, false
+	case unicode.IsSymbol(r):
+		return 5, false, false
+	default:
+		return 4, false, false
+	}
+}
+
+func signalUnitsForward(prefix []int, start, budget int) int {
```
abhinav-m22, line 197: Since the prefix array contains cumulative sums and is strictly increasing, would it make sense to use `sort.Search` here for an $O(\log N)$ binary search instead of an $O(N)$ linear scan? 

I know this likely won't be a bottleneck for typical text sizes, but could it help save some CPU time when parsing large prompts?

guan404ming: Thanks for the suggestion. I think the scan is bounded by one chunk (a few hundred runes) and each rune is visited once across all chunks, so chunking is already O(N) and the prefix build itself is O(N). sort.Search would work on the non-decreasing sums, but it saves nothing measurable here, so I would keep the simpler loop.

Feel free to correct me if there is any misunderstanding, thanks!

abhinav-m22: Ah, you're right! Because building the prefix array is already O(N). A simple linear scan is perfectly fine here and probably much more cache-friendly. Thanks :slightly_smiling_face:

AayushSaini101 (approved):

wilsonwu (approved):
LGTM, thanks!

Issue #2857, opened 2026-08-11
[Feature] Define a pluggable Looper outcome-verifier contract


Xunzhuo:
## Goal

Define one shared outcome-verifier contract so Looper algorithms can score, accept, rerank, reject, tie, or abstain on candidate outcomes without embedding a separate verifier protocol in every algorithm.

## Proposed Workgroup

MoM & Routing

## Background

Verification is fragmented across Confidence AutoMix, Fusion grounding hooks, and prompt-only workflow roles. Panel agreement is not a correctness oracle, and model-based judges can be biased by provider identity or presentation order. A common typed result is required before downstream algorithms such as Best-of-N can make deterministic, auditable choices.

Parent delivery area: #2336. Related attempt tracing: #2855. #3478 is the first planned outcome-reranking consumer after this contract lands.

## Work item shape

This leaf owns the verifier interface, typed result and failure semantics, one deterministic test verifier, one bounded model or HTTP adapter, and a behavior-preserving Confidence AutoMix integration. Best-of-N selection remains in #3478 and must consume this contract rather than extend this issue.

## Contract and scope

- Accept the original task, normalized policy-valid candidates, and optional explicitly trusted context.
- Return typed `approve`, `redo`, `reject`, `tie`, or `abstain` disposition; bounded score/confidence and reason codes; verifier type and version; usage and latency; and timeout, unavailable, or malformed-output status.
- Randomize candidate order and replace provider/model identity with opaque IDs for comparative model judges; retain the blind mapping only in authorized provenance.
- Keep deterministic and executable verifiers limited to their declared task-specific inputs and capabilities.
- Integrate verifier compute with common attempt tracing and request-wide budget accounting.
- Adapt existing Confidence AutoMix behavior to the interface without changing its public policy.
- Cover calibration identity, trusted context, timeouts, malformed output, partial panels, cancellation, and `on_error` behavior.

## Configuration and API impact

Add a versioned internal verifier contract and only the validated adapter configuration required by an existing consumer. Existing Confidence configuration and behavior remain compatible. No verifier result becomes a client-visible hidden-reasoning surface.

## Non-goals

- Treat NLI, peer agreement, or an uncalibrated judge score as ground truth.
- Add Best-of-N, retry loops, or token-level process-reward search.
- Persist hidden judge reasoning or expose private blind mappings.

## Completion signal

- [ ] Confidence AutoMix consumes the shared contract without behavior regression.
[...]

PR #3426, opened 2026-09-03
[Feature] looper: shared outcome-verifier contract (#2857)

Bevisy:
Implements the shared Looper outcome-verifier contract (#2857, slice 1) and routes two existing verifier paths through it: `confidence.AutoMix` HTTP entailment and `fusion` grounding (faithfulness + peer consistency). Public routing behavior is unchanged; both paths keep their scoring math and decision semantics byte-for-byte.

Related #2857

## Purpose

- Adds `pkg/looper/verifier.go`: a shared `Verifier` contract with a typed semantics taxonomy (`VerifierKind` — deterministic outcome, trusted-context faithfulness, ORM/PRM, LLM scalar judge, pairwise, peer consistency, …), a 5-state disposition (approve/redo/reject/tie/abstain), nullable confidence (nil = no score/abstain, distinct from 0), per-candidate scores (`CandidateScore` with `Flags`) so future outcome reranking needs no second algorithm-specific interface, and typed failure codes (timeout/unavailable/malformed/no-candidate) via `errors.As`.
- Rewires ConfidenceAutoMix entailment (`confidence.go`) and Fusion grounding (`grounding.go`) to consume the contract through `verifier_automix.go` / `verifier_grounding.go` adapters. Absence of peer evidence (<2 candidates) now yields a determinate `abstain` instead of an implicit tie.
- No config surface, enums, or public API changed.

## Test Plan

- `go test ./pkg/looper/ ./pkg/extproc/ ./pkg/protocolcodec/ ./pkg/llmprotocol/` (colima VM, cgo-ready)
- `go vet ./pkg/looper/...`; `gofmt -l` clean; `go mod tidy -diff` clean
- golangci-lint v2.5.0 (CI-pinned) + go1.25: `golangci-lint run ./pkg/looper/... --config ../../tools/linter/go/.golangci.yml --new-from-rev origin/main`
- New unit tests: contract semantics (deterministic verifier, nil-vs-zero confidence, disposition mapping, error classification), AutoMix adapter (approve/redo/wire/timeout/malformed/no-candidate via httptest), grounding verifiers (faithfulness `1/(1+spans)`, peer-consistency penalty math, tie/abstain with <2 candidates)

## Test Result

- All four packages pass; pre-existing `confidence`/`grounding` tests green → rewires are behavior-preserving.
- Lint: 0 issues on `./pkg/looper/...` with the CI-pinned toolchain combo.
- Remaining risk: calibration/benchmark evidence for the contract, and Replay tracing integration — tracked as follow-ups in #2857 (this slice does not change routing outcomes, so no E2E change required).

---
<details>

- [x] PR title begins with exactly one bracketed category: `[Feature]`
- [x] The title does not stack prefixes; affected modules (looper) belong in labels and the PR body
- [x] The PR links an `accepted` issue with exactly one owner (`Related #2857`)

</details>


FAUST-BENCHOU (dismissed):
LGTM. The shared verifier contract keeps the existing scoring behavior intact while giving the different verifier semantics a much clearer typed boundary.

wilsonwu (dismissed):
LGTM

Xunzhuo (changes requested):
Thanks for extracting the shared verifier contract. One behavior regression remains: PeerConsistencyVerifier emits synthetic panel-N IDs into flags, grounding spans, and synthesis notes, where the existing path used model names. The downstream synthesis prompt can no longer associate those notes with the actual model responses. Please map internal panel IDs back to caller-visible model labels before emitting the result and pin the prior label behavior in a regression test.

Bevisy:
@Xunzhuo  Good catch — fixed in the latest push.

Peer-consistency flags are remapped back to caller-visible model labels before the result is emitted, so traces, grounding spans, and the synthesis notes stay associated with the real responses, matching the legacy behavior. The prior label behavior is pinned in a regression test.

Bevisy:
The `E2E / integration-test (looper)` failure on the latest commit is in the E2E image build stage, before any test runs: the `openvino-builder` stage of `tools/docker/Dockerfile.extproc` fails to copy the OpenVINO runtime libs (`cp: cannot stat '.../site-packages/openvino/libs/libopenvino.so.*'`), i.e. the installed `openvino==2025.1.0` wheel no longer ships the layout the Dockerfile expects. This is unrelated to this PR's changes (only `pkg/looper` is touched; the same job passed on the previous commit and fails during `docker buildx`, before the Kind cluster is even created). All other checks, including `Full pre-commit checks` and `Core test and build`, are green.

Xunzhuo (changes requested):
Thanks—the synthetic panel-ID leak is fixed, but this remapping now corrupts faithfulness evidence. `groundingScoresFromVerifier` applies `remapFlaggedSpans` to every verifier result, while faithfulness flags are arbitrary unsupported text spans. A genuine span equal to `panel-0` is therefore replaced with that panel model name before tracing and synthesis. Please restrict ID remapping to peer-consistency results (or otherwise type the flags) and add a collision regression that preserves literal faithfulness spans.

Bevisy:
Fixed — remapping is now scoped to peer-consistency results only, keyed off the verifier kind. Faithfulness flags pass through verbatim, and a collision regression pins that a literal span like `panel-0` is preserved untouched.

Xunzhuo (commented):
Thanks—the remap is now limited to peer-consistency verification, and the literal `panel-0` regression closes the collision case I flagged. I don’t see a remaining code blocker; please sync current main and re-request review.

Bevisy:
> Thanks—the remap is now limited to peer-consistency verification, and the literal `panel-0` regression closes the collision case I flagged. I don’t see a remaining code blocker; please sync current main and re-request review.

Done, PTAL.

Xunzhuo (changes requested):
Thanks—the earlier routing collision is fixed, but two contract blockers remain: AutoMix accepts any numeric confidence without a finite or [0,1] check, so values such as 2.0 can approve, and `context.Canceled` falls through to unavailable instead of the deterministic cancellation outcome required by #2857. Please validate the score domain and map cancellation explicitly, with contract regressions.

Bevisy:
Both fixed in the latest commit:

- **Score domain**: AutoMix now rejects non-finite or out-of-[0,1] confidence as `malformed_output` — a value like `2.0` can no longer approve. Contract regression covers `2.0`, `-0.1`, `NaN`, and `±Inf`.
- **Cancellation**: `classifyVerifierHTTPError` now maps `context.Canceled` to a dedicated `canceled` failure code (deadline expiry stays `timeout`), so on_error policy is fully deterministic per #2857 — the completion signal's cancellation case. Regressions cover plain and wrapped cancellation, deadline expiry, network timeout, and unavailable.

Xunzhuo (approved):

PR #3428, opened 2026-09-03
[Feature] Add per-model request, stream idle, and connect timeouts to the reliability block

### LINKED ISSUE #3210: [Feature] Add per-model request, stream idle, and connect timeouts to the reliability block
### Goal

Let each provider model define its own total request deadline, stream idle timeout, and connect timeout. Today every model behind a listener shares one `listener.timeout`, so the value must fit the slowest reasoning model and a hung fast-tier request holds a stream, a client connection, and an in-flight slot for the full duration. A reasoning model legitimately needs minutes while a fast-tier model should fail in seconds, and today they share one number. The generated per-model Envoy routes also hardcode route `idleTimeout: 1200s` and cluster `connect_timeout: 1200s`, so the inter-chunk guard is effectively absent and connect failures take 20 minutes to trigger `retry_on: connect-failure`.

### Proposed Workgroup

Data Plane & Networking

### Scope

- Add `request_timeout`, `stream_idle_timeout`, and `connect_timeout` to the per-model reliability block in both config producers: Router config and the vLLM-SR CLI.
- Render them on generated per-model routes and clusters, replacing the hardcoded route idle timeout and cluster connect timeout.
- Fall back to listener values where unset; allow an explicit zero total deadline only when a stream idle timeout still bounds the stream.
- Define pre-response and mid-stream timeout surfaces for clients.
- Record timeouts as typed outcomes in metrics and telemetry instead of an unlabeled local-reply 504.
- Stop the default route from implicitly borrowing `models[0]` reliability.
- Document the reliability block and compatibility behavior.

### Non-goals

- Changing retry, hedging, or circuit-breaker policy beyond the timeout integration needed here.
- Defining provider-specific timeout fields outside the shared reliability block.
- Making an unbounded stream valid without another explicit idle or lifecycle bound.

### Completion signal

E2E coverage shows two models on one listener enforcing different total deadlines, a stalled stream cut by its idle timeout, a connect failure surfacing in seconds, and every timeout recorded as a typed outcome.

Parent Epic: #2992

ZireaelK:
Closes #3210

## Purpose
Add `request_timeout`, `stream_idle_timeout`, and `connect_timeout` to the per-model `reliability` block across Go and Python config producers, render them into generated Envoy routes and clusters, and classify timeout responses as typed metric outcomes.

### Background & Problem
Previously, all models behind a listener shared a single `listener.timeout`, requiring slow reasoning models and fast-tier models to use the same deadline. Additionally, generated Envoy routes and clusters hardcoded `idleTimeout: 1200s` and `connect_timeout: 1200s`/`10s`, and default routes implicitly inherited `models[0]` reliability policies.

### Solution & Changes
- **Config Models & Validation**:
  - Added `request_timeout`, `stream_idle_timeout`, and `connect_timeout` to `ProviderReliability` in both Go (`pkg/config/canonical_providers.go`) and Python CLI (`vllm-sr/cli/models.py`).
  - Added validation enforcing strictly positive `connect_timeout`, non-negative durations, and requiring a positive `stream_idle_timeout` if `request_timeout` is explicitly `0s`.
- **Envoy Template Rendering**:
  - Updated `envoy.template.yaml` to render per-model `timeout`, `idleTimeout`, and cluster `connect_timeout`.
  - Stopped the default route from implicitly borrowing `models[0]` reliability and retry policy.
- **Typed Telemetry & Error Surfaces**:
  - Updated `processor_res_header_runtime.go` and `metrics.go` to record 504/408 status codes under typed `"timeout"` error metric reason instead of generic `"upstream_5xx"`.
  - Updated `processor_protocol_contract.go` immediate protocol errors for 408/504 to return `ErrorUpstreamTimeout`.
- **Dashboard & Docs**:
  - Updated frontend forms, structured editors, and `config.yaml` reference sample.
- **Tests**:
  - Unit tests in `provider_reliability_test.go` and `test_config_generator.py`.
  - Metrics assertions in `extproc_test.go`.

## Test Plan

```bash
make agent-validate
make agent-lint
make check-go-mod-tidy
CI=true make agent-ci-gate
pytest src/vllm-sr/tests/test_config_generator.py
```
## Test Result

- `make agent-validate`: PASS
- `make agent-lint`: PASS
- `make check-go-mod-tidy`: PASS
- `CI=true make test-semantic-router`: PASS
- `CI=true make agent-ci-gate`: PASS
- `pytest src/vllm-sr/tests/test_config_generator.py`: PASS

---
<details>

- [X] PR title begins with exactly one bracketed category, such as `[Feature]`, `[Bug]`, `[Docs]`, `[Test]`, `[Research]`, `[Community]`, or `[CI/Build]`
- [X] The title does not stack prefixes such as `[Router][Docs]`; affected modules belong in labels and the PR body
- [X] The PR links an `accepted` issue with exactly one owner: one `wg/*` label for project work or `owner/maintainers` for repository governance
- [X] Commits in this PR are signed off with `git commit -s`
- [X] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>


FAUST-BENCHOU (commented):
Should the unset `stream_idle_timeout` keep the previous `1200s` default instead of falling back to `listener.timeout`? Otherwise existing configs that only set a listener request timeout will also get a shorter stream idle timeout after upgrading.

ZireaelK:
> Should the unset `stream_idle_timeout` keep the previous `1200s` default instead of falling back to `listener.timeout`? Otherwise existing configs that only set a listener request timeout will also get a shorter stream idle timeout after upgrading.

Updated, thanks for the review.

FAUST-BENCHOU (dismissed):
LGTM. Nice improvement on moving timeout configuration from listener-level defaults to per-model reliability policies.

The added validation and typed timeout metrics make the behavior clearer, and the tests cover both configuration safety and runtime error classification well.

wilsonwu (dismissed):
LGTM

Xunzhuo (changes requested):
Shared Anthropic routes carry per-model reliability settings but all point to one fixed-10s cluster, so a configured per-model connect timeout is accepted and silently ignored. Please render a policy-correct cluster per model/timeout or reject that shape. Also align unset stream-idle with #3210's listener-default contract and add runtime E2E for distinct deadlines, stalled streams, and short connect failures.

Xunzhuo (changes requested):
The new E2E still does not exercise the timeout configuration. It never sends the slow-model request or compares two deadlines, the stalled-stream path has no idle-timeout assertion, and the connect case only dials a local closed port instead of the generated Envoy cluster. Please make the maintained profile configure and verify all three runtime behaviors, and add the reliability-block compatibility documentation required by #3210.

Xunzhuo (changes requested):
Thanks for the follow-up. The generated timeout path is still not demonstrated: the request probes hit responsive backends and assert only upper bounds, the incomplete stream closes immediately after one chunk, and the connect target refuses immediately behind a separately hard-coded route timeout. These checks pass even if all three generated reliability settings are ignored, and connect or mid-stream failures are not recorded as typed timeout outcomes. Please use controllable header delay, frame stall, and connection establishment through the CLI-generated path, assert lower and upper timing plus the exact status or reset surface and metrics, fix the current gci failure, and align the documented 1200-second listener default with the actual 300-second default.

Xunzhuo (changes requested):
Thanks for extending the runtime probes. Two blockers remain. handleProcessReceiveError records timeout for every incomplete pre-header cancellation, EOF, or reset, while connect-timeout 503 responses remain upstream_5xx, so timeout telemetry is not tied to an observable timeout signal. Also the AI Gateway E2E applies separate hard-coded route and connect deadlines, so passing it does not prove the generated reliability configuration. Please classify only verified timeout outcomes and exercise the generated path without duplicate enforcement, with non-timeout abort coverage.

Xunzhuo (commented):
Thanks for extending the timeout coverage. One supported deployment path still cannot enforce the advertised per-model stream idle timeout.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -521,6 +524,35 @@ func appendRoutingHeaders(headersOut *[]*core.HeaderValueOption, model string) {
...
+				Key:      "x-envoy-upstream-rq-timeout-ms",
+				RawValue: []byte(strconv.FormatInt(timeoutMs, 10)),
+			}})
+		}
+	}
+	if rel.StreamIdleTimeout != "" {
+		if dur, err := time.ParseDuration(rel.StreamIdleTimeout); err == nil && dur > 0 {
+			timeoutMs := dur.Milliseconds()
+			*headersOut = append(*headersOut, &core.HeaderValueOption{Header: &core.HeaderValue{
+				Key:      "x-envoy-upstream-stream-idle-timeout-ms",
```
Xunzhuo: **Use an idle-timeout mechanism Envoy understands**

Impact: The external-Gateway path relies on this injected header, but Envoy does not consume `x-envoy-upstream-stream-idle-timeout-ms`; stream idle timeout is a route configuration field. This mutation is therefore only forwarded upstream, so a configured per-model stall deadline is not enforced on that path.

Required change: Configure `idle_timeout` through a supported Gateway or route policy, or reject and clearly scope the option where per-request enforcement is unavailable. Add an external-Gateway stall regression that fails without the real enforcement mechanism.

src/semantic-router/pkg/extproc/processor_core.go
```diff
@@ -130,10 +132,113 @@ func (r *OpenAIRouter) handleProcessReceiveError(ctx *RequestContext, err error)
...
+		start = ctx.StartTime
+	}
+	var elapsed time.Duration
+	if !start.IsZero() {
+		elapsed = time.Since(start)
+	}
+
+	// Mid-stream stall termination (stream idle timeout)
+	if ctx.IsStreamingResponse && !ctx.StreamingComplete {
+		if idleDur, ok := parseDurationSafe(rel.StreamIdleTimeout); ok && elapsed >= idleDur-300*time.Millisecond {
```
Xunzhuo, line 212: **Measure the actual idle gap before recording a timeout**

Impact: This compares total request lifetime with `stream_idle_timeout`. A stream can keep producing chunks beyond that duration and then be canceled or reset for a non-timeout reason; because this runs before cancellation handling, it is still labeled and counted as a stream idle timeout.

Required change: Base attribution on explicit Envoy timeout/reset evidence or track the time since the last response frame and exclude cancellations. Add a regression with a long-lived continuously active stream ending in a non-timeout cancellation or reset.

Xunzhuo (commented):
Thanks for replacing the unsupported header with a real Gateway route policy. One timeout-attribution path still conflates total request age with stream idleness; the exact-head CI is also failing on the stale branch, so please refresh and rerun the relevant checks after the fix.

ZireaelK:
@Xunzhuo Thanks for the detailed reviews and guidance above. I've addressed all the feedback and fixed them. However, the CI is failing now due to some unrelated issues. Could you please take another look when you have a moment? Thanks!

src/vllm-sr/cli/templates/envoy.template.yaml
```diff
@@ -55,7 +55,10 @@ static_resources:
                 {% endif %}
                 route:
                   cluster: {{ model.cluster_name }}_cluster
-                  timeout: {{ listener.timeout | default('1200s') }}
+                  timeout: {{ model.reliability.get('request_timeout') if (model.reliability is defined and model.reliability and model.reliability.get('request_timeout')) else (listener.timeout | default('300s')) }}
```
Xunzhuo, line 58: **Normalize the timeout duration before rendering Envoy config**

The new validators accept values such as `request_timeout: "300ms"`, but this template writes the Go-style string directly into a protobuf `Duration`. Rendering that accepted configuration and running `envoy --mode validate` fails with `duration must end with a single s`; the same configuration with `"0.3s"` validates. A valid reliability setting therefore prevents the generated proxy from starting. Normalize all three timeout fields to protobuf seconds (or reject unsupported forms before generation), and cover this with an Envoy configuration-validation regression.

src/vllm-sr/tests/test_config_generator_timeouts.py
```diff
@@ -0,0 +1,331 @@
...
+""",
+        extproc_host="localhost",
+        router_api_host="localhost",
+    )
+
+    # Fast model route and cluster timeouts
+    fast_route = _model_route(rendered, "model-fast")["route"]
+    assert fast_route["timeout"] == "10s"
+    assert fast_route["idleTimeout"] == "3s"
+    fast_cluster = _cluster_by_name(rendered, "model_fast_cluster")
```
Xunzhuo: **Update and run the new generator regressions against the current generator**

Running this module gives two failures: the generator now emits names such as `model_model_2dfast_cluster` and `model_claude_2dfast_cluster`, while these tests expect the old names. The fallback checks below also require an `idleTimeout` key even though the template inherits it from the connection manager when unset. Update those assertions to the actual resource and inheritance contracts, and include this module in `make vllm-sr-test`; its explicit pytest list currently omits these new regressions.

Xunzhuo (changes requested):
The earlier stream-idle findings are fixed. Two new issues remain: accepted duration values can prevent Envoy from starting, and two generator regressions fail. Please normalize all three timeout fields and fix and include the regression module in the CLI test target. Envoy configuration parsing and focused mock controls were checked; full runtime timeout coverage remains pending.

ZireaelK:
@Xunzhuo Thanks for the detailed review, I've addressed both issues in the latest commit and fixed them. BTW, since `google.protobuf` is not a dependency in `vllm-sr`, I implemented a standalone validator `_assert_valid_protobuf_duration` instead of using `from google.protobuf.duration_pb2 import Duration`. Looking forward to another review, thanks!

Xunzhuo (commented):
The two previous findings are resolved: Go-style durations render as protobuf durations, and the generator assertions match the route and idle-timeout contracts. On this head, all 12 focused Python tests, five generated configurations validated by Envoy 1.34, and the full repository harness checks pass. The preceding 9ca3251 integration with current main also passed all four deployed reliability-timeout scenarios; this last commit only replaces equivalent exception handling for lint. I have not repeated that deployment on 181382b. Current-main integration is clean; the remaining approval gate is the required current-head CI.

e2e/pkg/testmatrix/testcases.go
```diff
@@ -61,6 +61,8 @@ var BaselineRouterContract = []string{
 	"language-routing",
 	// Reask signal rule matching and routing (issue #3178)
 	"reask-routing",
+	// Reliability timeouts runtime contract (issue #3210)
+	"reliability-timeouts",
```
Xunzhuo, line 65: **[P2] Update the baseline inventory regression for the added case**

Adding `reliability-timeouts` here makes `BaselineCases("standard")` return 37 cases and `BaselineCases("full")` return 39, but `TestBaselineScopesPreserveEveryFunctionalCase` still requires 36 and 38. The current Router Tools job fails deterministically on that assertion; the same test passes on current main and fails on this head. Update the inventory assertion alongside this addition and rerun `make test-e2e-unit`.

Xunzhuo (changes requested):
The deployed Gateway timeout scenarios now pass on this head. One integration issue remains: adding the reliability case breaks the fixed baseline inventory assertion, so Router Tools fails deterministically. Please update that regression and rerun the checks. Production Stack also remains red on jailbreak accuracy.

.github/workflows/integration-test-k8s.yml
```diff
@@ -104,6 +104,15 @@ jobs:
           trap cleanup_router_log_stream EXIT
 
           set +e
+          if [ "${{ matrix.profile }}" = "envoy-ai-gateway" ]; then
+            ENVOY_AI_GATEWAY_CI_TESTS="chat-completions-request,apiserver-runtime-config-endpoints,domain-classify,semantic-cache,semantic-cache-polarity,pii-detection,pii-entity-offsets,pii-long-text,jailbreak-detection,security-long-text,decision-priority-selection,plugin-chain-execution,plugin-request-mutations,tool-selection,rule-condition-logic,decision-fallback-behavior,plugin-config-variations,event-routing,plugin-short-circuit-no-dispatch,language-routing,reask-routing,llm-classifier-distribution-routing,sequence-classifier-routing,reliability-timeouts"
+            make e2e-test E2E_PROFILE=${{ matrix.profile }} E2E_TESTS="${ENVOY_AI_GATEWAY_CI_TESTS}" E2E_VERBOSE=true E2E_KEEP_CLUSTER=false
```
Xunzhuo: **[P2] Keep the Gateway CI selection aligned with its runtime profile**

This explicit list omits 15 cases selected by the Gateway profile, including entrypoint-recipe-routing, protocol-codec-openai-regression and the Anthropic request/response cases. On current main, TestGateDCIOverridesMatchKnownExclusions rejects those unrecorded omissions, so e2e-coverage-check fails during build-e2e; current main alone passes the same check. Please use the maintained baseline selection or align the explicit list and its bounded exclusions with the runtime profile.

e2e/testcases/reliability_timeouts.go
```diff
@@ -0,0 +1,491 @@
...
+	pkgtestcases.Register("reliability-timeouts", pkgtestcases.TestCase{
+		Description: "Verify runtime enforcement of distinct request deadlines, stalled stream idle timeouts, and short connect timeouts (issue #3210)",
+		Tags:        []string{"reliability", "timeout", "streaming", "connect", "kubernetes"},
+		Fn:          testReliabilityTimeouts,
+	})
+	pkgtestcases.Register("reliability-distinct-deadlines", pkgtestcases.TestCase{
+		Description: "Verify per-model request timeouts enforce distinct deadlines across models",
+		Tags:        []string{"reliability", "timeout", "deadline"},
+		Fn:          testReliabilityDistinctDeadlines,
+	})
```
Xunzhuo: **[P2] Resolve the unselected standalone testcase registrations**

These four standalone reliability registrations are selected by no profile: the baseline selects only reliability-timeouts. The aggregate does call their functions, but the new current-main runtime inventory gate checks registered testcase identities and rejects all four as unreachable, causing e2e-coverage-check to fail during build-e2e. Please remove redundant standalone registrations, select them through a maintained profile, or record an explicit bounded disposition, then rerun the gate.

Xunzhuo (changes requested):
The current-main integration fails the mandatory E2E coverage checks: the Gateway CI list omits 15 selected cases, and four standalone reliability registrations are unreachable by profile selection. The same checks pass on main alone. Please align both selections with the current inventory contract and rerun the checks. This validation was cluster-free; the deployed timeout scenarios were not rerun.

PR #3429, opened 2026-09-03
[CI/Build] Add unassign-inactive-assignees workflow

### LINKED ISSUE #2586: [CI/Build] Unassign inactive issue participants

abhinav-m22:
Closes #2586

## Purpose

Adds a daily workflow that warns and unassigns inactive contributors from accepted issues, filling the gap that `stale.yml` intentionally leaves by exempting accepted and in-progress issues.

Activity is measured from human Timeline API events only. Bot comments, label changes, and automation do not reset the clock. Each assignee is evaluated independently. Wired as a reusable workflow through `maintenance.yml`.

Affected: `.github/workflows/`, `.github/scripts/`, `tools/agent/docs/maintainer-ops.md`. Owner: `owner/maintainers`.

## Test Plan

```bash
cd .github/scripts && npm test
```

## Test Result

All 34 unit tests pass. Covers warn, unassign, recovery, per-assignee isolation, fail-closed API errors, idempotency, dry-run, and pagination edge cases.

---
<details>

</details>


FAUST-BENCHOU (commented):
Could this issue-level `WARN_LABEL` break the per-assignee isolation? Once one assignee is warned, `isWarned` becomes true for every other assignee on the issue, so another inactive assignee may never receive their own warning before being unassigned.

abhinav-m22:
@FAUST-BENCHOU That's right. The issue-level label broke per-assignee isolation completely (e.g., an active assignee could inadvertently suppress the warning for an inactive co-assignee).

I've just pushed a commit that completely removes the `WARN_LABEL` dependency. We now use the existence of a recent warning comment mentioning the specific assignee (checked against their lastHumanActivity timestamp) as the sole, isolated source of truth for the warning state. All assignees on an issue are now tracked independently.

FAUST-BENCHOU (commented):
The workflow logic looks good overall. The current CI failure seems to come from the pre-commit checks rather than the implementation itself.

Please fix the lint issue before merging.

abhinav-m22:
@FAUST-BENCHOU thanks for the review. The pre-commit issue is fixed and CI is green. Please take a look

Xunzhuo (dismissed):
The workflow can remove an assignee without a recorded warning, treats warning lookup failures as safe to mutate, derives activity from aggregate linked-PR updates rather than that assignee, and removes before a best-effort notification. Please make this transition warned, fail-closed, assignee-specific, retryable/idempotent, and align the maintainer docs. Exact-head CI must also execute the new Jest tests.

abhinav-m22:
@Xunzhuo addressed all the points.

- **Warned and fail-closed.** Removal now requires a warning addressed to that assignee plus a full grace window since it was posted, so nobody is dropped unwarned. A failed timeline, linked-PR, or comment read leaves the assignment untouched and fails the job instead of mutating.
- **Activity is per-assignee.** Dropped `search.issuesAndPullRequests` and `pr.updated_at`, which moves on anyone's comment or a CI write. Linked PRs come from paginated `cross-referenced` events, and each PR timeline is read for that assignee's own commits, reviews, and force-pushes.
- **Retryable and idempotent.** The removal notice is posted before the mutation, and the assignment is kept if it fails. Per-assignee markers replace the shared one, which let a removal notice count as a warning and let `@alice` match `@alice-b`.
- **Docs and CI.** `maintainer-ops.md` documents the policy, activity rules, and override paths. The `unassign-warned` label is gone entirely (@FAUST-BENCHOU's isolation issue). A new `github-scripts-tests` pre-commit hook runs Jest at the exact PR head via the always-on `quality` job. Tests 34 to 62.

Xunzhuo (changes requested):
Thanks for addressing the earlier transition issues; those fixes look good. Two assignee-specific correctness gaps remain: a committed timeline event is credited to the PR opener without checking the commit author, so someone else’s commit can reset that assignee’s clock; and limiting linked PRs to 20 can ignore qualifying activity on an older linked PR and still remove the assignee. Please attribute commit activity to the evaluated assignee and evaluate the full paginated linked-PR set, with coverage for both cases.

abhinav-m22:
@Xunzhuo Both gaps were real, fixed now. Commits are now attributed via `pulls.listCommits` on the GitHub author/committer login instead of the timeline's git identity, so someone else's commit on the assignee's PR no longer resets their clock. The 20-PR cap is gone: every same-repo linked PR is evaluated, and the scan only stops once activity newer than the warn threshold has already settled the outcome as active.

Tests cover foreign commits, committer-only attribution, unattributable commits, a fail-closed commit read, and qualifying activity found on the oldest of 40 linked PRs.

Xunzhuo (changes requested):
This state machine can still run concurrently. The caller gives the scheduled run, `task=all`, and `task=unassign` different concurrency groups, while the reusable job has no shared group. Two runs can both observe an absent marker and post duplicate warning or removal notices, violating #2586’s idempotency and at-most-one-warning contract. Please serialize this job with one repository-wide concurrency group and `cancel-in-progress: false`.

abhinav-m22:
@Xunzhuo Fixed. Added a job-level concurrency group to the reusable workflow so the scheduled run, `task=all`, and `task=unassign` all queue behind a single sweep, with `cancel-in-progress: false` so in-flight writes still land.

The two points from the earlier review are already on the head: commit activity is attributed by GitHub login rather than the PR opener, and there is no 20-PR cap on linked PRs. Both have test coverage. PTAL.

Xunzhuo (changes requested):
Thanks—the shared job-level concurrency group resolves the race from my last review. One operational blocker remains: the docs say maintainers can run the reusable workflow directly with `dry_run` and threshold or exemption inputs, but it exposes only `workflow_call` and `maintenance.yml` forwards none of them. Please add a real manual dispatch surface or wire those inputs through the caller before this mutation workflow lands.

abhinav-m22:
@Xunzhuo Good catch, the workflow only exposed `workflow_call`, so there was no way to dry-run it. Added `workflow_dispatch` with the same four inputs, with `dry_run` defaulting to true so a manual run previews by default and only mutates when you pass `-f dry_run=false`.

Left the inputs off `maintenance.yml` since they are unassign-specific and would not apply to the other tasks it dispatches. Docs now carry the real command. PTAL.

Xunzhuo (commented):
Thanks—the workflow now exposes a direct manual dispatch with dry-run as the safe default, forwards all override inputs into the same serialized job, and the maintainer documentation matches that surface. My prior reachability blocker is resolved. Please rebase onto current `main` for the final high-risk approval pass.

Xunzhuo (approved):
The inactivity workflow now has a reachable dry-run dispatch, serialized mutations, and fresh warning/grace checks before removal. The prior reachability concern is resolved, and the exact-head helper-script tests pass.

wilsonwu (approved):
LGTM

Issue #3377, opened 2026-09-02
[Feature] Gate session model switches on calibrated recent-window progress evidence


Xunzhuo:
## Goal

Prevent session-level model thrashing by requiring calibrated recent-window evidence before escalating or de-escalating the selected logical model.

## Proposed Workgroup

Agentic & Context

## Background

Current `main` already has `global.router.learning.protection`, decision-local `adaptations.protection`, observe and apply modes, trusted session and conversation identity, idle and minimum-turn controls, switch margin and stability, tool-loop and context-portability hard locks, switch-history penalties, and Replay. It does not yet have a count-and-age-bounded outcome window, calibrated progress and degradation semantics, consecutive evidence, cooldown, or explicit hysteresis.

Parent Epic: #2973. Reuse those Router Learning protection seams; the legacy `global.router.model_selection.model_switch_gate` surface remains rejected, and calibration is a versioned external input rather than work invented by this issue.

## Work item shape

This is one safety leaf. The typed evidence window, candidate and hard-constraint validation, cooldown/hysteresis decision, bounded state update, Replay, and regression coverage form one end-to-end behavior. They may land through reviewable commits or PRs but are not independent claimable tasks.

## Scope

- Define content-minimized typed outcome samples for progress or quality, repeated failure classes, confidence trend, cost and latency trend, switch count, and time since the last switch.
- Bound the recent window by count and age and define cold-start, stale, missing, conflicting, and partially available evidence.
- Evaluate calibrated improvement or degradation with consecutive-evidence thresholds, cooldown, and hysteresis.
- Re-run authorization, safety, residency, context, capability, candidate-set, portability, and budget constraints before committing a switch.
- Record calibration version, evaluated evidence, decision or suppression reason, switch provenance, and bounded state transition in Replay.
- Add deterministic unit, state, concurrency, restart, and maintained E2E coverage.

## Configuration and API impact

Extend `global.router.learning.protection.tuning` with explicit, disabled-by-default recent-window, threshold, cooldown, hysteresis, and calibration-identity fields and use decision-local `adaptations.protection` only where an existing override seam is required. Do not add another top-level switch surface. Existing sessions keep current behavior when the fields are absent.

## Non-goals

- Store raw conversation content solely for routing.
- Invoke models or tools from the gate.
- Create an online calibrator or mutate production thresholds on the request path.
- Replace Router Learning outcome and protection contracts.

## Completion signal

[...]

PR #3436, opened 2026-09-03
[Feature] Add recent-outcome evidence window for session switch gate

ayrnb:
Related #3377

## Purpose

Implement the evidence-to-switch vertical from #3377 within the existing
Router Learning protection flow. A model switch proposed by the selector or
rescue path is evaluated against bounded, content-minimized session evidence;
the gate does not become another selector or a top-level routing surface.

### What changes

- **Bounded evidence:** terminal outcomes and owned feedback feed one
  count/age-bounded window, applied from the first turn and persisted
  across restarts.
- **Explicit sample semantics:** duplicates merge by request+model only;
  failures stay order-independent; missing metrics stay unknown (an omitted
  score is not a measured zero).
- **Deterministic gating:** coverage, same-class streaks, trend, cooldown, and
  windowed switch history; de-escalation uses the catalog quality index at
  exact reasoning effort.
- **Hard constraints stay authoritative:** candidate boundary, backend,
  context, capabilities, portability, and SLO/quality filters are rechecked
  before any override; suppression never restores an ineligible model.
- **Explainable application:** Replay records calibration, evidence, direction,
  current/proposed/final models, and whether a suppression applied; rejected
  rescues remain visible in `rescue_switch_gate`.
- **Maintained coverage:** a dedicated `progress-gate` profile and one shared
  HTTP assertion cover capture, idempotent feedback, suppression, switches,
  cooldown, oscillation, and Replay; `router-replay` is unchanged.

### Implementation flow

With the gate enabled and outer protection in `apply` mode:

```mermaid
flowchart TD
    Outcomes["Prior responses + accepted feedback"] --> Window["Bounded evidence window"]
    Window --> Gate["Evaluate evidence, thresholds and switch history"]
    Proposal["Selector / rescue proposes a switch"] --> Gate
    Gate --> Mode{"Gate mode?"}
    Mode -->|observe| Observe["Record soft verdict only"]
    Mode -->|enforce| Enforce["Use verdict to allow or suppress the switch"]
    Observe --> Final["Recheck hard constraints<br/>Finalize routing result + Replay"]
    Enforce --> Final
```

Hard constraints remain authoritative in both modes; suppression never restores
an ineligible model.

### Configuration and compatibility

The only configuration surface remains:

```yaml
global:
  router:
    learning:
      protection:
        tuning:
          progress_gate:
            enabled: false
            mode: observe
            calibration_id: ""
            window_size: 8
            window_ttl_seconds: 900
            min_window_outcomes: 3
            min_consecutive_regressions: 2
            min_consecutive_recoveries: 2
            cooldown_seconds: 120
            max_switches_per_window: 2
```

Omitting the section leaves the gate disabled; `enabled` and `mode` are
independent, `observe` never applies verdicts, and neither mode relaxes hard
constraints. Existing protection, candidate boundaries, auth, ownership, and
API contracts are unchanged.

`calibration_id` is an external `name@version` identity required by `enforce`;
only its format is validated, not calibration quality. Defaults stay
disabled/observe.

Snapshot additions are backward-readable (missing fields read as
unknown/default, never fabricated). The save-order and recovery guards apply
only to gate-configured sessions — no cross-session locks or new storage — with
regressions covering both.

### Scope boundaries

This change adds no online calibrator, no in-gate model/tool calls, no
request-time threshold mutation, no raw-text storage, no Prometheus metrics,
and no change to the learning/protection contracts. Existing authorization,
safety, and residency policy still define the candidate boundary; the gate
cannot introduce candidates outside it.

Positive completion tokens are a transport-level progress proxy, not a
semantic quality judgment. Explicit tool errors stay non-attributable and are
never inferred from message shape; costs depend on configured pricing and the
observed latency metric is TTFT.

## Test Plan

Build the native bindings from the same source revision and use their normal
linker/runtime paths before running Go tests.

```bash
cd src/semantic-router
go test -count=1 ./pkg/config ./pkg/sessiontelemetry ./pkg/selection \
  ./pkg/extproc ./pkg/apiserver ./pkg/configschema
go test -race -count=1 ./pkg/sessiontelemetry
go vet ./pkg/config ./pkg/sessiontelemetry ./pkg/selection ./pkg/extproc ./pkg/apiserver
go build ./cmd
```

The regressions cover default-off behavior, effective config validation,
missing/zero score distinction, future/stale samples, merge ordering and model
identity, restart-first-write preservation, exact-effort quality comparison,
SLO/context exclusions, hard locks in observe, and rejected-rescue Replay.

```bash
cd e2e
go test ./pkg/... ./profiles/...
go test -c -o /tmp/progress-gate-e2e.test ./testcases
```

Against a running stack configured from `e2e/profiles/progress-gate/values.yaml`,
set `GATE_E2E_INFERENCE_URL`, `GATE_E2E_MANAGEMENT_URL`, and
`GATE_E2E_MANAGEMENT_TOKEN`, then run the maintained assertion:

```bash
GATE_E2E_MODE=enforce /tmp/progress-gate-e2e.test \
  -test.run '^TestProgressGateLive$' -test.v -test.count=1 -test.timeout=4m
# Restart that test router with the same fixture in observe mode, then:
GATE_E2E_MODE=observe /tmp/progress-gate-e2e.test \
  -test.run '^TestProgressGateLive$' -test.v -test.count=1 -test.timeout=4m
```

The normal Kubernetes deployment entrypoint is:

```bash
make verify PROFILE=progress-gate
```

Repository contract checks:

```bash
make harness-check
make config-schema-check
make docs-check-translation-coverage
make bench-router-learning
```

## Test Result

| Check | Result |
| --- | --- |
| Six affected Go packages with `-count=1` | Passed on a host with current native bindings |
| Session-state race tests | Passed |
| Go vet and router build | Passed for the checked packages |
| E2E framework/profile Go tests | Passed |
| Harness registry/workflow checks | registry, 33 workflow contracts and 161 Python tests passed; |
| Changed-file pre-commit and Go lint | Passed after fixing shadow/import issues |
| Generated schema check | Passed |
| Translation coverage check | Passed; no coverage regression |
| Existing Router Learning architecture evaluation | Passed 7/7 checks; not a production calibration result |
| Maintained HTTP assertion, two independent vLLM backends, enforce | Passed: seven turns, 22.89 s; backend counters +5/+2 match the script |
| Maintained HTTP assertion, two independent vLLM backends, observe | Passed: same seven turns, 22.92 s; backend counters +3/+4 match the divergent sequence |

Backend: real vLLM 0.27.1 serving Qwen3-0.6B under two fixture names (16
tokens/request), on **two independent instances** each registering only its
own name. Envoy routes on the ExtProc-emitted `x-selected-model` header, so a
misroute fails with 404 and backend counters independently show where each
turn landed. Feedback and `fixture/session-gate@1.0.0` are synthetic inputs;
this validates routing and state transitions, **not** weight quality.

Enforce run (`--- PASS: TestProgressGateLive (22.89s)`):

| # | Model | Verdict | Reason | Suppression applied | Window | Streak | Switches in window |
| ---: | :---: | --- | --- | :---: | ---: | ---: | ---: |
| 1 | A | — (seed) | — | — | 0 | 0 | 0 |
| 2 | A | suppress | `insufficient_evidence` | yes | 1 | 1 | 0 |
| 3 | A | suppress | `insufficient_evidence` | yes | 2 | 2 | 0 |
| 4 | B | switch | — | — | 3 | 3 | 0 |
| 5 | B | suppress | `cooldown` | yes | 4 | 4 | 1 |
| 6 | A | switch | — | — | 5 | 5 | 1 |
| 7 | A | suppress | `oscillation_guard` | yes | 6 | 6 | 2 |

A = `openai/gpt-oss-20b`, B = `openai/shadow-candidate`. `calibration_id`,
`evidence_version`, and `enforced=true` are identical on every turn and are not
repeated per row; "Suppression applied" is `—` on switch turns because no
suppression was pending.

Backend request counters (`vllm:request_success_total`, `finished_reason="length"`):

| Backend | Logical model | Before | After | Delta | Scripted turns |
| --- | --- | ---: | ---: | ---: | --- |
| :18001 | `openai/gpt-oss-20b` | 51 | 56 | +5 | 1,2,3,6,7 |
| :18002 | `openai/shadow-candidate` | 11 | 13 | +2 | 4,5 |

The observe run below starts from these same counters (56 / 13), so the two
tables are consecutive measurements of one environment.

Observe run — the same seven-turn script (`--- PASS: TestProgressGateLive (22.92s)`):

| # | Model | Verdict | Reason | Suppression applied | Window | Streak | Switches in window |
| ---: | :---: | --- | --- | :---: | ---: | ---: | ---: |
| 1 | A | — (seed) | — | — | 0 | 0 | 0 |
| 2 | B | suppress | `insufficient_evidence` | no | 1 | 1 | 0 |
| 3 | B | — (no proposal) | — | — | — | — | — |
| 4 | B | — (no proposal) | — | — | — | — | — |
| 5 | A | suppress | `cooldown` | no | 4 | 4 | 1 |
| 6 | A | — (no proposal) | — | — | — | — | — |
| 7 | B | suppress | `oscillation_guard` | no | 6 | 6 | 2 |

Verdicts are recorded but never applied, so every proposal executes: observed
`A→B→B→B→A→A→B` vs enforced `A→A→A→B→B→A→A` (turns 3, 4, 6 carried no switch
proposal). Turns 5 and 7 still switch despite recorded `cooldown` /
`oscillation_guard` verdicts.

Backend counters: :18001 +3, :18002 +4 — opposite of the enforced (+5/+2),
matching the divergent sequences.

---
<details>

- [x] PR title begins with exactly one bracketed category, such as `[Feature]`
- [x] The title does not stack prefixes; affected modules belong in labels and the PR body
- [x] The PR links an `accepted` issue with exactly one owner (`wg/agentic-context`)

</details>


Xunzhuo (changes requested):
Thanks for building out the session outcome state. One ordering blocker remains: the window is arrival-ordered even though its contract is timestamp-ordered, and TTL pruning removes only an expired prefix. A delayed older outcome inserted behind a newer one can therefore survive expiry and affect switching. Please order by event time or prune all expired evidence, add a delayed-arrival regression, and complete the accepted issue's remaining vertical before this is ready.

src/semantic-router/pkg/extproc/req_filter_classification_selector.go
```diff
@@ -145,6 +145,9 @@ func (r *OpenAIRouter) selectWithSelector(
 		selectedModel,
 		ctx,
 	)
+	if ctx.VSRProgressGateError != nil {
```
Xunzhuo: **Propagate gate rejection across every selector path**

Impact: This is the only caller that returns `VSRProgressGateError`. The single-candidate and fallback paths also call `applyRouterLearning`, but ignore the error and return the selected model, so a gate-enabled decision can dispatch a model rejected by context, capability, SLO, or quality hard filters.

Required change: Centralize this fail-closed propagation after every learning application (or return the error through the single/fallback signatures), and add regressions for hard rejection on both paths.

src/semantic-router/pkg/extproc/processor_res_body.go
```diff
@@ -17,6 +17,9 @@ func (r *OpenAIRouter) handleResponseBody(v *ext_proc.ProcessingRequest_Response
 	}
 
 	completionLatency := time.Since(ctx.StartTime)
+	if !ctx.IsStreamingResponse || v.ResponseBody.GetEndOfStream() {
```
Xunzhuo, line 20: **Record an outcome when a stream terminates before EOS**

Impact: Streaming turns enter the evidence window only from this EOS branch. If the ExtProc stream ends, is canceled, or errors before EOS, `handleProcessReceiveError` marks the stream aborted but records no non-attributable fact. Those newer failed turns therefore do not displace older regressions in the count-bounded window, so a later request can authorize a switch using evidence that is no longer from the latest turns.

Required change: Idempotently record a `missing`/non-attributable terminal outcome on pre-EOS termination and cover cancellation, EOF, and error paths with a window regression.

src/semantic-router/pkg/sessiontelemetry/router_memory_shared_store.go
```diff
@@ -193,6 +193,15 @@ func persistRouterSessionState(sessionID string) {
 		return
 	}
 	defer release()
+	s := globalRouterSessionMemory
+	s.mu.Lock()
+	st := s.sessions[sessionID]
+	gateConfigured := st != nil && st.outcomeWindowSize > 0
+	s.mu.Unlock()
+	if gateConfigured {
+		st.outcomeSaveMu.Lock()
```
Xunzhuo: **Make outcome persistence atomic across router replicas**

Impact: This mutex serializes snapshot writes only within one process, while Redis `Save` remains a whole-snapshot `SET`. Two replicas that loaded the same session can append different outcomes and then overwrite each other, losing evidence and switch history that the shared store is meant to recover.

Required change: Use a distributed atomic merge/CAS (or an append-oriented representation) that preserves concurrent facts and switch timestamps, and add a multi-writer persistence regression.

Xunzhuo (commented):
Thanks for completing the vertical and fixing the ordering issue. Three current-head integrity gaps still need to be closed before this is merge-safe. The exact-head progress-gate E2E is also red before reaching its first gate assertion, so please rerun it after the fixes.

ayrnb:
> Thanks for completing the vertical and fixing the ordering issue. Three current-head integrity gaps still need to be closed before this is merge-safe. The exact-head progress-gate E2E is also red before reaching its first gate assertion, so please rerun it after the fixes.

Have fixed it. The base moved and brought the management API changes with it — replay list and outcome ingest are now under /api/v1/observability, so the test's old /v1/router_replay and /v1/router/outcomes were hitting the mux's default 404 before any gate assertion got a chance to run. Rebased onto main and updated the paths.

src/semantic-router/pkg/extproc/processor_res_body.go
ayrnb: Done

src/semantic-router/pkg/extproc/router_learning_progress_gate.go
```diff
@@ -0,0 +1,180 @@
...
+		})
+	}
+	return facts
+}
+
+func progressEvidenceStateKey(ctx *RequestContext) string {
+	if ctx != nil && !ctx.Routing.IsPassthrough() && ctx.ResponseObjectState != nil {
+		return config.RoutingNamespaceKey(ctx.Routing.RecipeName(), ctx.ResponseObjectState.SessionTrackingID)
+	}
+	return routingSessionStateKey(ctx)
```
Xunzhuo, line 46: **Scope evidence to the protected conversation**

With the default conversation protection scope, the selected/current model is stored under `VSRLearningSessionID`, but this fallback keys outcome evidence by the broader `SessionID`. Two conversations sharing one session ID can therefore contribute to the same regression window, letting one conversation authorize or suppress a switch in the other. The new E2E keeps those headers equal, so it cannot expose the leak. Please key evidence with the same protection-scoped identity as `routingLearningStateKey` and add a regression with two conversation IDs under one session ID.

src/semantic-router/pkg/extproc/req_filter_classification_selector.go
```diff
@@ -26,10 +26,28 @@ const (
...
 func (r *OpenAIRouter) selectModelFromCandidates(
 	selCtx *selection.SelectionContext,
 	algorithm *config.AlgorithmConfig,
 	ctx *RequestContext,
+) (*config.ModelRef, string, error) {
+	selected, method, err := r.selectModelFromCandidatesInner(selCtx, algorithm, ctx)
+	if err != nil {
+		return nil, method, err
+	}
+	if ctx != nil && ctx.VSRProgressGateError != nil {
```
Xunzhuo, line 41: **Do not persist a rejected selection**

This check happens only after the inner selector, single-candidate, or fallback path has called `recordAgenticSessionDecision`. When a hard filter sets `VSRProgressGateError`, the request returns no model but has already persisted the proposed/current selection plus switch timestamp and cooldown, so a rejected request changes later routing. Please propagate the rejection before recording session state (or explicitly skip/roll back that write), and verify the snapshot remains unchanged for all three paths.

Xunzhuo (commented):
Thanks for the latest fixes—the fail-closed propagation and shared-store merge are much closer. Two inline state-integrity blockers remain before this is safe to merge; the earlier pre-EOS termination blocker also still applies and is not repeated here.

One shared-state blocker also remains in `src/semantic-router/pkg/sessiontelemetry/session_evidence.go:174`: after CAS merges concurrent writers in Redis, a process that already has any local session skips the shared read and continues evaluating only its local outcomes. Replicas can therefore make different coverage/streak decisions despite the stored union. Please reconcile merged shared state into the live gate read path (or install the CAS result locally), and add a two-replica regression proving both readers observe the merged window.

src/semantic-router/pkg/routerreplay/store/learning_diagnostics.go
```diff
@@ -7,6 +7,11 @@ type LearningDiagnostics struct {
 	ProtectionPreflight *LearningProtectionDiagnostics `json:"protection_preflight,omitempty"`
 	Adaptation          *LearningAdaptationDiagnostics `json:"adaptation,omitempty"`
 	Protection          *LearningProtectionDiagnostics `json:"protection,omitempty"`
+	// ProtectionStateKey is the window key the gate resolved for this request.
+	// Derived from client correlation headers, no more sensitive than the
+	// record's session id; ingest lands feedback here, records without it
+	// fall back to the bare session key.
+	ProtectionStateKey string `json:"protection_state_key,omitempty"`
```
Xunzhuo, line 14: Please regenerate `website/static/openapi/apiserver/apiserver.openapi.json` with this serialized field. Both `Core test and build` and `Full pre-commit checks` fail `api-docs-check` on this head: the generated diff adds `protection_state_key` in two response schemas, but the checked-in contract omits it. Run `make api-docs-openapi` and commit the result so the published schema and required checks match the new replay payload.

Xunzhuo (changes requested):
The scoped outcome-ingest and shared-window fixes are present, and the focused session telemetry and progress-gate regressions pass. The new replay field still leaves the generated API contract stale; please regenerate the OpenAPI document and commit it with this change.

src/semantic-router/pkg/routerreplay/store/learning_diagnostics.go
ayrnb: Done — rebased onto latest main.

Xunzhuo (dismissed):
The generated API now includes protection_state_key in both replay schemas, resolving the remaining contract drift. The prior session-state fixes are preserved, and the current-head progress-gate E2E and generated-contract checks pass.

src/semantic-router/pkg/extproc/router_learning_protection.go
```diff
@@ -377,9 +424,79 @@ func (r *OpenAIRouter) selectProtectionResult(
...
+	result *selection.SelectionResult,
+	currentModel string,
+) bool {
+	if result == nil || currentModel == "" {
+		return false
+	}
+	if selectedModelRefFromResult(learningCtx, &selection.SelectionResult{SelectedModel: currentModel}) == nil {
+		return false
+	}
+	result.SelectedModel = currentModel
```
Xunzhuo: [P1] Keep the held candidate consistent with the held model

`SessionAwareSelector.Select` now returns an exact `SelectedCandidate`. When it proposes `frontier` and the gate suppresses that switch to keep `cheap`, this assignment leaves the candidate pointing to `frontier`. `ResolveSelectionCandidate` rejects the mismatch in `protectionDecisionFromResult`, which falls back to `input.baseResult`; when the base chose `frontier`, the supposedly suppressed switch is dispatched anyway. Resolve the current session candidate and update the complete result identity together, including `SelectedCandidate` and LoRA, and cover a selector result that carries an exact candidate.

src/semantic-router/pkg/extproc/router_learning_gate_constraints.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/llmprotocol"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/selection"
+)
+
+// progressCandidateReason reuses the request's candidate boundary and hard filters.
+func (r *OpenAIRouter) progressCandidateReason(ctx *RequestContext, selCtx *selection.SelectionContext, model string) string {
+	if selCtx == nil || !selectionContextContainsModel(selCtx, model) {
+		return "candidate_set"
+	}
+	ref := selectedModelRefFromResult(selCtx, &selection.SelectionResult{SelectedModel: model})
```
Xunzhuo: [P2] Check eligibility using the exact candidate reference

A supported candidate pool can contain the same model with different reasoning efforts. Resolving this model-only result is then intentionally ambiguous, so both otherwise eligible refs receive `model_unavailable`. `progressEligibleContext` calls this once per ref but passes only `ref.Model`, drops the whole pool, and makes requests fail with `ErrNoEligibleCandidates` as soon as the progress gate is enabled. Pass each exact `ModelRef` through the eligibility check, and preserve the recorded/proposed exact ref when checking a hold or switch.

src/semantic-router/pkg/extproc/router_learning_protection.go
```diff
@@ -283,6 +323,9 @@ func (r *OpenAIRouter) protectionDecisionFromResult(
 	baseModel := selectedModelName(input.baseResult)
 	proposalModel := selectedModelName(enteredResult)
 	if preflight.mode == config.DecisionAdaptationModeObserve {
+		if result.SessionPolicy != nil {
```
Xunzhuo: [P2] Handle the nil result on the observe-mode error path

`applyProtectionSwitch` explicitly calls `protectionDecisionFromResult(..., nil, ...)` when selection fails in outer protection `observe` mode. For example, the session selector returns `ErrNoEligibleCandidates` when an active tool loop owns an exact candidate that is no longer in the pool. This new dereference panics instead of preserving the incoming selection and recording the diagnostic policy. Guard `result` before reading its session policy and add coverage for the observe-mode selector-error branch.

Xunzhuo (changes requested):
Three focused regression tests reproduce the candidate-identity and observe-mode failures below. Please preserve exact candidate identity through filtering and suppression, handle the nil observe result, and rerun the affected tests and deployed progress-gate profile. The live-test inventory fix is confirmed; the deployed profile remains unverified.

ayrnb:
> Three focused regression tests reproduce the candidate-identity and observe-mode failures below. Please preserve exact candidate identity through filtering and suppression, handle the nil observe result, and rerun the affected tests and deployed progress-gate profile. The live-test inventory fix is confirmed; the deployed profile remains unverified.

#3685 moved candidate identity from a model name to an exact `SelectedCandidate` (model + reasoning effort), and these sites were written before it.  Have rebased and fixed it.

src/semantic-router/pkg/selection/session_switch_gate.go
```diff
@@ -0,0 +1,193 @@
...
+	proposedParams, proposedOK := s.modelParams[proposed]
+	if !currentOK || !proposedOK || currentParams.QualityIndex == "" ||
+		currentParams.QualityIndex != proposedParams.QualityIndex {
+		return false
+	}
+	effort := func(model string) string {
+		if selCtx != nil {
+			for _, candidate := range selCtx.CandidateModels {
+				if candidate.Model == model {
+					return candidate.ReasoningEffort
```
Xunzhuo: **[P1] Compare the exact current and proposed candidate efforts**

This lookup takes the first candidate with the same model name, ignoring the recorded `PreviousCandidate` and the exact proposal. With candidates `[A/low, A/high, B]`, same-index quality scores `40/90/60`, and the session held on `A/high`, switching to `B` is a downgrade. This code compares `A/low` instead and reports an escalation, so two regression outcomes authorize moving from quality90 to60 when the downgrade gate should require recovery. A focused regression reproduces that enforced switch. Pass the exact current/proposed `ModelRef`s into this comparison and cover duplicate model names and candidate-order invariance.

Xunzhuo (changes requested):
The held-candidate, exact eligibility and nil observe-result fixes are covered by passing exact-head Core regressions. One candidate-identity bug remains in switch-direction classification: a focused regression reproduces an enforced quality downgrade during regression evidence, and an independent code-path review confirms both protection callers are affected. Deployed progress-gate and multi-replica runtime checks were not run in this review.

Xunzhuo (approved):
Both protection callers now compare the exact current and proposed candidate efforts. The focused direction, candidate-identity and outcome-window regressions pass in current-head Core CI, and the deployed progress-gate-evidence-to-switch case passes.

wilsonwu (approved):
LGTM

PR #3450, opened 2026-09-03
[Bug] Sanitize Postgres-hostile text in Router Replay writes

### LINKED ISSUE #3427: [Bug] Replay writer silently drops records with invalid UTF-8 content
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

The Router Replay writer sends request and response content to Postgres without normalizing text that Postgres cannot store. A replay record containing a NUL byte or invalid UTF-8 can therefore fail at insert time while the routed completion still succeeds, leaving the record silently absent from `router_replay_records`.

This was observed on a `nightly-20260830` deployment carrying PDF-derived text. Two failures from the same session were:

```text
msg: router_replay_persist_failed
decision: Large Context Decision
error: failed to insert record: pq: unsupported Unicode escape sequence
error: failed to insert record: pq: invalid byte sequence for encoding "UTF8": 0xef 0xbf
```

### To Reproduce

Send PDF-derived content containing either a NUL byte / `\^@` escape or a truncated UTF-8 sequence through a deployment with Postgres-backed Router Replay enabled, then inspect the replay table and router logs. The completion returns normally, `router_replay_persist_failed` is logged, and the corresponding replay row is absent.

At the store boundary, the same failure can be exercised by inserting a replay record whose request or response body contains one of those sequences into the Postgres store.

### Expected behavior

Router Replay should preserve the record by normalizing unsupported text before insert, or otherwise expose a bounded, record-identifying failure that makes the loss countable. An accepted implementation may strip NUL bytes and replace invalid UTF-8, for example with Go's `strings.ToValidUTF8`, while retaining the rest of the record.

### Impact

The routed completion is unaffected, but document-bearing traffic can disappear from replay data. This silently under-counts the cases for which replay is especially useful and can make downstream denominators incorrect.

### Additional context

Postgres text columns cannot store NUL bytes, and truncated UTF-8 is rejected by the driver. Current `main` passes replay request and response strings directly into the parameterized Postgres insert and logs the resulting persistence error without retaining the record. The report comes from a deployment consumer who is not carrying a downstream patch and can confirm behavior against a fix.

chiliec:
Closes #3427

## Purpose

Router Replay wrote request/response content into Postgres without normalizing
text that Postgres cannot store. A replay record containing a NUL byte or
invalid UTF-8 (as PDF-derived text carries) failed at insert time while the
routed completion still succeeded, so the row was silently absent from
`router_replay_records`:

```
error: failed to insert record: pq: unsupported Unicode escape sequence
error: failed to insert record: pq: invalid byte sequence for encoding "UTF8": 0xef 0xbf
```

This sanitizes untrusted text at the Postgres store boundary (owned by
wg/enterprise-environment, the accepted issue):

- text columns (`request_body`, `response_body`, `prompt`, `tool_definitions`,
  `terminal_reason`): drop NUL bytes and replace invalid UTF-8 with the Unicode
  replacement character via `strings.ToValidUTF8`, preserving the rest.
- jsonb columns: strip `\u0000` escapes from the marshaled JSON, which Postgres
  rejects with "unsupported Unicode escape sequence" (`json.Marshal` already
  replaces invalid UTF-8, so `\u0000` is the only jsonb-hostile artifact left).

Applied in `preparePostgresInsertRecord`/`marshalPostgresInsertJSON` (the `Add`
path) and in the incremental `AttachRequest`/`AttachResponse`/`UpdateLifecycle`
writers. Clean input is returned unchanged.

## Test Plan

```
cd src/semantic-router
go test ./pkg/routerreplay/store/
go test ./pkg/routerreplay/store/ -run 'Sanitize|InsertRecordSanitizes' -v
```

RED→GREEN: reverting the sanitizers to identity makes the new tests fail
(`sanitizePostgresText("a\x00b") = "a\x00b", want "ab"`; `request body still
contains a NUL byte`); restoring passes.

## Test Result

```
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/routerreplay/store  0.432s
```

New tests pass; full `store` package suite green; `go build`, `go vet`, and
`gofmt -l` clean on the touched files. Not run: a live-Postgres integration
insert (no Postgres in this environment) — the fix is verified at the store
boundary where the bad bytes are prepared for binding.

Happy to adjust naming or scope.

---

🤖 Generated with assistance from an AI coding tool; authored, reviewed, and
verified by me.

### DIFF
--- a/src/semantic-router/pkg/routerreplay/store/postgres.go
+++ b/src/semantic-router/pkg/routerreplay/store/postgres.go
@@ -375,6 +375,7 @@ func (p *PostgresStore) UpdateLifecycle(
 		return err
 	}
 	defer release()
+	reason = sanitizePostgresText(reason)
 	//nolint:gosec // tableName is validated during store creation
 	query := fmt.Sprintf(`
 		UPDATE %s
@@ -410,6 +411,7 @@ func (p *PostgresStore) AttachRequest(ctx context.Context, id string, body strin
 		return err
 	}
 	defer release()
+	body = sanitizePostgresText(body)
 	//nolint:gosec // tableName is validated during store creation
 	query := fmt.Sprintf(`
 		UPDATE %s
@@ -449,6 +451,7 @@ func (p *PostgresStore) AttachResponse(ctx context.Context, id string, body stri
 		return err
 	}
 	defer release()
+	body = sanitizePostgresText(body)
 	//nolint:gosec // tableName is validated during store creation
 	query := fmt.Sprintf(`
 		UPDATE %s
--- a/src/semantic-router/pkg/routerreplay/store/postgres_record_row.go
+++ b/src/semantic-router/pkg/routerreplay/store/postgres_record_row.go
@@ -118,7 +118,7 @@ func marshalPostgresInsertJSON(record Record, out *postgresInsertRecord) error {
 		if err != nil {
 			return fmt.Errorf("failed to marshal %s: %w", field.name, err)
 		}
-		*field.target = encoded
+		*field.target = sanitizePostgresJSON(encoded)
 	}
 	return nil
 }
@@ -137,6 +137,11 @@ func preparePostgresInsertRecord(record Record) (Record, error) {
 	if record.LifecycleState == "" {
 		record.LifecycleState = LifecycleInProgress
 	}
+	record.RequestBody = sanitizePostgresText(record.RequestBody)
+	record.ResponseBody = sanitizePostgresText(record.ResponseBody)
+	record.Prompt = sanitizePostgresText(record.Prompt)
+	record.ToolDefinitions = sanitizePostgresText(record.ToolDefinitions)
+	record.TerminalReason = sanitizePostgresText(record.TerminalReason)
 	return record, nil
 }
 
--- a/src/semantic-router/pkg/routerreplay/store/postgres_text_sanitize.go
+++ b/src/semantic-router/pkg/routerreplay/store/postgres_text_sanitize.go
@@ -26,6 +26,13 @@ func sanitizePostgresText(s string) string {
 // "unsupported Unicode escape sequence". json.Marshal already replaces invalid
 // UTF-8 with the replacement character, so \u0000 is the only jsonb-hostile
 // artifact left.
+//
+// Only a \u0000 that is itself a JSON escape (an odd number of backslashes
+// precedes it) encodes an actual NUL byte and is removed. A literal string
+// value containing the text "\u0000" marshals with an escaped backslash
+// (\\u0000, an even run), so the match falls on an escaped backslash and must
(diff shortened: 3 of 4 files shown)


AayushSaini101 (dismissed):
LGTM thanks for the finding : )

FAUST-BENCHOU (dismissed):

wilsonwu (dismissed):
LGTM

Xunzhuo (commented):
Thanks for tackling this. I found one JSON escaping edge case that can still make PostgreSQL reject an otherwise valid replay write; the precise case and fix are in my follow-up review.

Xunzhuo (commented):
Thanks again. `sanitizePostgresJSON` also matches inside valid marshaled JSON for a literal backslash followed by `u0000`: the encoded value has two backslashes, so the replacement starts at the second one and leaves the first escaping the closing quote. That makes the JSON invalid, and a valid replay value can still fail the PostgreSQL write. Please sanitize decoded string values before marshaling, or otherwise preserve JSON escaping, and add this case to the regression tests.

chiliec:
Good catch — you're exactly right. `sanitizePostgresJSON` did a blind `ReplaceAll` of `\u0000` on the marshaled bytes, so a valid value literally containing `\u0000` (marshaled as `\\u0000`) had its second backslash deleted, leaving the first escaping the closing quote → invalid JSON that Postgres rejects.

Fixed in 80aef30: it now only strips a `\u0000` preceded by an **odd** number of backslashes (a real NUL escape) and preserves escaped backslashes. Added regression cases in `TestSanitizePostgresJSONPreservesLiteralEscape` — the literal `\u0000` value, a real NUL byte, and the two adjacent in one value. Confirmed the old code failed the literal case (`invalid character 'd' in string escape code`) and the new code round-trips it; full `store` package, `go vet`, and `gofmt` all clean.

Xunzhuo (commented):
Thanks for the quick follow-up. The backslash-parity fix now preserves literal `\\u0000` text while removing real NUL escapes, and the regression coverage exercises both forms. The Postgres text and JSON write boundary looks correct to me. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

chiliec:
Refreshed the integration state on the new head `2cc421c8`:

- DCO now passes — added the missing `Signed-off-by` to the fix commit.
- Rebased onto latest `main` (was BEHIND), no conflicts.

No code change from the reviewed head; the store-package tests, `go vet`, and `gofmt` are still clean locally and CI is re-running on the new head. Ready for a fresh look whenever you have a moment.

Xunzhuo (approved):
The replay write boundary now removes NULs and repairs invalid UTF-8 while preserving literal backslash-u0000 text. I verified real PostgreSQL insert/readback and incremental request, response, and lifecycle updates in both synchronous and asynchronous modes; all 12 hostile-text cases passed. The current-main integration is clean.

PR #3454, opened 2026-09-04
[Bug] Refuse to write FP32 weights under an fp16 name in the CK graph rewrite

subin9:
Related #3256

## Purpose

- `rewrite_graph.py` keeps the input's weight precision and only casts around each `CKFlashAttention` node, so running it on `model.onnx` with the output named `model_fa_fp16.onnx` writes FP32 weights under an fp16 name. That is how the five published files in #3256 were made: rewriting the feedback head's `model.onnx` reproduces its `model_fa_fp16.onnx` in size, op histogram and node names.
- Refuse an fp16 output name for an FP32 graph, with a message naming the fix (`model_fa.onnx`, or an FP16 input). Warn on the reverse.
- Read the precision from the initializers, not `value_info`: that field is optional and describes activations, so an FP16 export without it read as FP32 and one activation could not vouch for a mixed graph. Mixed, BFLOAT16 and no-float graphs are refused with the counts.
- Document the rule in the README, and that the published `model_sdpa_fp16.onnx` graphs are not matched by the rewriter (`No attention blocks found`), which is why `model.onnx` was used.
- Run `test_rewrite_graph.py` in CI. Nothing executed it before: the only task rule matching `onnx-binding/**` runs the Go binding suite, so the rewriter was linted and never tested. `make ck-rewrite-test` is now the fast test of a `ck-flash-attn-rewriter` rule for `onnx-binding/ort-ck-flash-attn/**`, so the quality job's changed-file gate (`make agent-ci-lint`) runs the suite for any change under that path.

## Test Plan

- `make ck-rewrite-test` (`python3 -m unittest test_rewrite_graph` in `onnx-binding/ort-ck-flash-attn/scripts` with the agent venv, which installs the pinned `onnx` on first use)
- `rewrite_graph.py` on the feedback `model.onnx` under both output names, and on the jailbreak and modality `model_sdpa_fp16.onnx`

## Test Result

- 9 passed, including value_info-free FP16, FP32, mixed, no-float and BFLOAT16 graphs; black, ruff, codespell, markdownlint and the structure gate are clean.
- The feedback `model.onnx` carries no `value_info` at all and reads FLOAT from its initializers. `model.onnx` → `model_fa_fp16.onnx` exits 1 with nothing written; → `model_fa.onnx` proceeds and matches the published file. Both SDPA files report `No attention blocks found`, on `main` and here alike.
- `agent_gate.py resolve` on this PR's files matches `rust-bindings` and `ck-flash-attn-rewriter` with `make ck-rewrite-test` among the fast tests; `make agent-validate` passes.

### DIFF
--- a/onnx-binding/ort-ck-flash-attn/README.md
+++ b/onnx-binding/ort-ck-flash-attn/README.md
@@ -42,6 +42,22 @@ python3 scripts/rewrite_graph.py \
 Use `--hdim` and `--local-attention` only when they match the source model's
 architecture. Keep the original SDPA model for correctness comparison.
 
+The input decides the output precision. The rewriter keeps the weights as it
+finds them and, for an FP32 graph, adds fp32↔fp16 casts around each
+`CKFlashAttention` node. A rewrite of the FP32 `model.onnx` must therefore be
+named `model_fa.onnx`; `model_fa_fp16.onnx` needs an FP16 input graph. The
+script refuses an fp16 name for an FP32 graph, because `find_onnx_models`
+ranks candidates by name alone.
+
+The matcher expects the attention subgraph of the FP32 `model.onnx` export:
+`Softmax` fed by `Add(MatMul(Mul(q), Mul(k)), mask)` and consumed directly by
+the AV `MatMul`. The published `model_sdpa_fp16.onnx` graphs differ (a NaN
+guard of `IsNaN` and `Where` between `Softmax` and the AV `MatMul`, or the
+scale applied after the QK `MatMul`) and are reported as
+`No attention blocks found`, so an FP16 FA artifact cannot be produced from
+them until the matcher covers that shape
+(vllm-project/semantic-router#3256).
+
 ## Load the custom op
 
 The Semantic Router ONNX binding reads `ORT_CK_FLASH_ATTN_LIB`:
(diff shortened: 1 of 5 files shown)


Xunzhuo (commented):
Thanks for tightening this. One blocker remains: the refusal infers `FP32 weights` from the first Q tensor's optional `graph.value_info` entry rather than from the graph initializers. A valid FP16 export without intermediate value info is therefore rejected as FP32, while a mixed graph can pass based on one activation. Please derive the check from the weight tensors and add graph-level FP16 and FP32 coverage.

subin9:
@Xunzhuo Switched the check to the initializers! weight_precision reads the weight tensors, ignores integer ones, and refuses mixed, BFLOAT16 and no-float graphs with the counts. Added graph-level FP16/FP32/mixed tests on value_info-free graphs. Small aside: the real feedback model.onnx has zero value_info entries, so the old check could never have said FP16 for it.

Xunzhuo (commented):
Thanks, the initializer-based check fixes the precision-detection issue, including value-info-free and mixed-weight graphs, and I don’t see a code blocker. I can’t approve this head yet because exact-head CI only lints `rewrite_graph.py` and `test_rewrite_graph.py`; it never executes the new unittest suite. Please wire `python3 -m unittest test_rewrite_graph` into a GitHub check for this path, then re-request review.

subin9:
@Xunzhuo wired. make ck-rewrite-test runs the suite in the quality job for changes under onnx-binding/ort-ck-flash-attn/**; on this head it ran 9/9.

Xunzhuo (approved):
Thanks, this closes the remaining gap: the exact-head quality job ran all nine rewriter tests, and the initializer-based guard still covers value-info-free FP16/FP32 and mixed graphs. This is ready from my side.

wilsonwu (approved):
LGTM, thanks!

PR #3455, opened 2026-09-04
[Bug] Skip semantic cache for queries past the embedding window

### LINKED ISSUE #3366: [Bug] Semantic cache collides on long prompts sharing a prefix

guan404ming:
Closes #3366

## Purpose

- The default `bert` cache embedding truncates at 512 tokens, so prompts sharing a long prefix collide at similarity 1.0.
- Count tokens in the cache adapter and treat a query past the window as a miss on lookup and a no-op on store.
- Touches the `pkg/cache` adapter, its `pkg/extproc` wiring, and the response-cache doc.

## Notes

- The onnx build embeds `bert` through mmBERT and exposes no tokenizer, so the check is a no-op there via a build-tag file.

## Test Plan

- `cd src/semantic-router && go test ./pkg/cache/ -run TestCache`
- `go vet -modfile=go.onnx.mod -tags=onnx ./pkg/cache/ ./pkg/extproc/`

## Test Result

- Passed. The new spec fails without the guard: the adapter serves the other prompt's response.


Xunzhuo (commented):
Thanks — the semantic adapter now skips both lookup and storage when the configured BERT tokenizer proves the 512-token window was exceeded, and the shared-prefix regression covers the unsafe collision directly. The branch is behind current `main`, so please update it and let the resulting required checks pass before approval.

Xunzhuo (changes requested):
Thanks for tightening the cache boundary. The guard is still BERT-only, while supported Candle Gemma deliberately truncates at its context limit and the ONNX multimodal and mmBERT paths also truncate even though the ONNX guard always returns false. Long prefix-identical queries can therefore still alias on both lookup and store, contrary to #3366. Please make the guard model- and runtime-aware and cover the supported matrix, at least Gemma and ONNX multimodal, across both operations.

guan404ming:
Thanks, good catch. Replaced the BERT-only Go tokenizer check with an FFI in each binding that reads the loaded model's own tokenizer and window: candle covers bert, gemma, mmbert, multimodal, qwen3; ONNX covers mmbert (every non-multimodal type) and multimodal. Both lookup and store use it.

Xunzhuo (approved):

PR #3457, opened 2026-09-04
[Bug] Bound batch classification by max_batch_size

guan404ming:
Related #3329

## Purpose

- Make the batch classification API honor the configured `max_batch_size`; today the field is accepted but never enforced.
- Oversized batches now get a clear client error instead of silently running unbounded.

## Test Plan

- `go test ./pkg/apiserver/`

## Test Result

- Passed.

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_classify.go
+++ b/src/semantic-router/pkg/apiserver/route_classify.go
@@ -181,6 +181,13 @@ func (s *ClassificationAPIServer) parseBatchClassificationRequest(w http.Respons
 		return BatchClassificationRequest{}, false
 	}
 
+	if maxBatchSize := s.maxBatchSize(); maxBatchSize > 0 && len(req.Texts) > maxBatchSize {
+		metrics.RecordBatchClassificationError("unified", "batch_too_large")
+		s.writeErrorResponse(w, http.StatusBadRequest, "INVALID_INPUT",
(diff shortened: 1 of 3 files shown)


Xunzhuo (commented):
Thanks for wiring the existing limit into the request boundary. The handler now rejects `texts` above `max_batch_size` before classifier work, preserves the configured boundary itself, and records a dedicated error reason. The exact-head core and E2E gates are green, and I don't see a blocker. I don't see a code blocker in this head. The current integration state still needs a fresh check before approval.

FAUST-BENCHOU (dismissed):
LGTM. The max batch size is now enforced at the API boundary, and the boundary tests cover both the allowed and rejected cases clearly.

wilsonwu (dismissed):
LGTM, thanks!

Xunzhuo (changes requested):
The new maxBatchSize helper reads s.config directly, while live config updates replace that pointer under configMu and resolver-backed servers use currentConfig as the authoritative source. A concurrent batch request and config update can race, and resolver-backed runtime config can keep enforcing the stale startup limit. Please take one currentConfig snapshot here and add live-update or race coverage.

guan404ming:
Thanks, agreed. maxBatchSize now takes one currentConfig() snapshot, so live updates and resolver-backed servers are honored. Added a test where the resolver config bounds the batch while the startup config does not.

Xunzhuo (approved):

PR #3458, opened 2026-09-04
[Bug] Honor feedback_mapping_path in the feedback detector

guan404ming:
Related #3329

## Purpose

- `feedback_mapping_path` was parsed but the detector always read `id2label` from the model's `config.json`.
- Load the mapping file when the path is set, and point the reference config at the model's `label_mapping.json`.
- Touches `pkg/classification`, `config/config.yaml`, and the user-feedback doc.

## Test Plan

- `cd src/semantic-router && go test ./pkg/classification/ ./pkg/config/ ./pkg/modeldownload/`

## Test Result

- Passed.


Xunzhuo (dismissed):
Thanks for wiring this field through. One runtime blocker remains: the `label_mapping.json` selected here does not match this model's classifier-head indices. That file currently maps 0/1/2/3 to NEED_CLARIFICATION/SAT/WANT_DIFFERENT/WRONG_ANSWER, while the model's config, training metadata, model card, and training code define SAT/NEED_CLARIFICATION/WRONG_ANSWER/WANT_DIFFERENT. Since `Classify` resolves the raw class index through this configured file, the reference config will systematically return and route the wrong feedback labels. Please either leave `feedback_mapping_path` empty so the existing config fallback remains authoritative, or correct the published mapping and add a contract/runtime test that verifies the mapping order against the model output.

guan404ming:
Thanks! Confirmed the mismatch on the published model. Left `feedback_mapping_path` empty in the reference and ONNX e2e configs so `config.json` stays authoritative, and updated the doc.

Xunzhuo (commented):
The prior mapping-order blocker is addressed: explicit mapping paths are honored, while the maintained mmBERT configs keep the model config authoritative. The current-head checks are green and the current-main synthetic merge is clean; please rebase onto current main and refresh exact-head checks before approval.

FAUST-BENCHOU (approved):
LGTM. The configured feedback mapping path is now honored while preserving the existing config.json fallback, and the mapping loader remains compatible with both supported formats.

wilsonwu (approved):
LGTM, thanks!

PR #3460, opened 2026-09-04
[Feature] Move the remaining candle models onto the shared chunked kernel

### LINKED ISSUE #3382: [Feature] Migrate the remaining candle models to the shared chunked attention kernel

subin9:
Closes #3382

## Purpose

- Move the remaining dense attention sites onto `chunked_sdpa`: Qwen3 embedding, Gemma3, the four multimodal blocks, the LoRA Qwen3 decoder, and the BERT base model. The kernel gains `q_offset` and reads the key length from `k` (needed by the SigLIP probe head and the KV-cache decoder), and the per-model `(seq, seq)` masks are gone.
- Each site keeps its dense math as a test reference and must match it within `1e-4`; the 12 new tests are in `RUST_CI_LIB_TESTS`.
- Align the Qwen3Guard prefix cache with the decode offset: `process_cached_suffix` prefills every suffix token but the last, so the decode loop's first step adds it at an aligned position instead of leaving a duplicate key the causal range would then attend. `test_cached_suffix_generation_matches_uncached` pins it and is in `RUST_CI_LIB_TESTS` too.
- `structure-rules.yaml` gets exceptions for `gemma3_model.rs`, `qwen3_embedding.rs` and `base_model.rs`: the gate has no Rust ratchet and errors on them even unchanged.

## Test Plan

- `make test-rust-ci`
- Peak RSS above loaded weights for one Qwen3 (`mom-embedding-pro`) call per length, main vs this branch, CPU

## Test Result

- `test-rust-ci` 20/20 (13 new). The cached-vs-uncached test runs snapshot restore + suffix prefill + five argmax decode steps against one uncached prefill and checks cache length, logits and sampled tokens per step; it fails on the previous sequence at step 0.
- Qwen3, peak MiB above weights: 2048 tokens 3340 → 883, 4096 tokens 12788 → 1716, 8192 tokens not runnable before → 3380. Dense quadrupled per doubling; chunked doubles.

## Not covered here

- `MAX_CLASSIFICATION_SEQ_LEN`, the Qwen3 f64 cast, and Gemma3's ignored padding mask are unchanged (non-goals).
- A Qwen3 padding row now attends evenly over earlier padding keys instead of only itself; nothing reads that row, and the padded-batch test pins the real rows.


FAUST-BENCHOU (dismissed):
LGTM. The shared kernel migration preserves the existing attention semantics across the remaining model paths, with solid dense-reference coverage for decode offsets, sliding windows, padding, and asymmetric query/key lengths.

wilsonwu (dismissed):
LGTM, thanks!

Xunzhuo (commented):
The shared-kernel migration looks good: query offsets and no-key ranges are handled centrally, the dense-reference tests cover prefill, decode, and the migrated model paths, and the reported peak-memory scaling meets #3382’s goal. The branch is behind current `main` after the merge queue could not update it; please rebase or merge `main` and let the resulting required checks pass before approval.

Xunzhuo (changes requested):
One blocker surfaced in the Qwen3Guard prefix-cache path. `process_cached_suffix` first appends the complete suffix, then generation re-feeds its last token; from the second generated step onward, the physical KV cache is one entry ahead of the logical `q_offset`. The new causal range caps keys at `q_offset + q_len`, so it excludes the newly appended token and attends the stale duplicate instead. Please align the cache and offset handling and add cached-versus-uncached multi-token generation coverage; the current decode test covers only aligned caches.

subin9:
@Xunzhuo Thank you, fixed the prefix-cache blocker, process_cached_suffix now prefills all but the last suffix token, so the decode loop's first step adds it at an aligned offset and there is no duplicate key :) Added test_cached_suffix_generation_matches_uncached (snapshot + suffix prefill + five decode steps against one uncached prefill) and registered it in the CI list; it fails on the old sequence at step 0. Rebased on main too.

Xunzhuo (commented):
Thanks—the suffix prefill now leaves the final token for the first decode step, and the five-step cached-vs-uncached regression pins cache length, logits, and sampled tokens. My blocker is resolved; please sync current main and re-request review.

subin9:
@Xunzhuo synced with current main!

Xunzhuo (approved):

Xunzhuo:
@subin9 thanks!

PR #3462, opened 2026-09-04
[Bug] Honor vector_store upload limits in the files API

guan404ming:
Related #3329

## Purpose

- Make `/v1/files` honor `vector_store.max_file_size_mb` and `vector_store.supported_formats`; today both are accepted but the upload path hardcodes 50MB and a fixed extension list.
- Unset config keeps the previous defaults.

## Test Plan

- `go test ./pkg/apiserver/`

## Test Result

- Passed.

### DIFF
--- a/src/semantic-router/pkg/apiserver/route_files.go
+++ b/src/semantic-router/pkg/apiserver/route_files.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"math"
 	"net/http"
 	"path/filepath"
 	"slices"
@@ -32,8 +33,9 @@ import (
 )
 
 const (
-	maxUploadSize        = 50 * 1024 * 1024 // 50MB
-	multipartMemoryLimit = 8 * 1024 * 1024
+	megabyte             int64 = 1024 * 1024
+	maxUploadSize              = 50 * megabyte
+	multipartMemoryLimit       = 8 * 1024 * 1024
 )
 
 // allowedExtensions defines the document types that can be uploaded for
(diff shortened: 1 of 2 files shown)


src/semantic-router/pkg/apiserver/route_files.go
```diff
@@ -98,7 +123,7 @@ func (s *ClassificationAPIServer) handleUploadFile(w http.ResponseWriter, r *htt
 			return
 		}
 		s.writeErrorResponse(w, http.StatusBadRequest, "INVALID_INPUT",
-			fmt.Sprintf("failed to parse multipart form (max size: %dMB): %s", maxUploadSize/(1024*1024), err.Error()))
+			fmt.Sprintf("failed to parse multipart form (max size: %dMB): %s", maxBytes/(1024*1024), err.Error()))
```
AayushSaini101: Thanks @guan404ming can we simply this like below, to avoid any overflow thanks : )

```
const MB int64 = 1024 * 1024
maxMB := maxBytes / MB
```

AayushSaini101: and add test case to understand what will happen during the overflow process thanks

Xunzhuo (dismissed):
Thanks for wiring these settings through. `uploadLimits()` reads `s.config` directly instead of the synchronized live-config path. Because `publishConfigMutation()` replaces that pointer under `configMu`, a concurrent file upload can race a management update; with a resolver-backed runtime config, the Files API can also keep enforcing the stale startup limits after reload. Please take one `cfg := s.currentConfig()` snapshot before reading `VectorStore` and add live-update/race coverage. The existing size-conversion overflow thread also remains open and needs to be resolved before approval.

guan404ming:
Thanks! `uploadLimits()` now takes one `currentConfig()` snapshot, with resolver and concurrent-mutation tests. MB conversion falls back to the default on overflow, with a test.

src/semantic-router/pkg/apiserver/route_files.go
guan404ming: Thanks! Moved the conversion into `uploadMaxBytes()` with a `megabyte` constant and an overflow test.

AayushSaini101 (dismissed):

src/semantic-router/pkg/apiserver/route_files.go
AayushSaini101: Thanks : ) @guan404ming

Xunzhuo (changes requested):
The live-config and overflow fixes address the prior review, but the configured file-size boundary is still wrong: `MaxBytesReader` limits the entire multipart request, so a file exactly `max_file_size_mb` is rejected once multipart headers and boundaries are counted. Please enforce `maxBytes` against the file payload itself and add an exact-limit acceptance test.

FAUST-BENCHOU (commented):
`supported_formats` can now make `.pdf` uploads succeed (and the reference config includes `.pdf`), but `vectorstore.ExtractText` still rejects `.pdf`. This means an accepted file will later fail every vector-store ingestion with `parse_error`.

Should PDF parsing be supported before exposing it here, or should this allowlist be constrained to formats the ingestion parser can actually consume?

wilsonwu (dismissed):
LGTM, thanks!

Xunzhuo (dismissed):
Three upload-contract blockers remain on this head. route_files.go applies the configured limit to the whole multipart envelope, so an exactly-at-limit file is rejected; the route wrapper still caps the body at 50 MiB, so the shipped 100 MiB setting cannot work; and the accepted .pdf format reaches vectorstore.ExtractText, which rejects PDF. Please enforce the configured limit on file bytes with a dynamic envelope guard, add exact-limit and over-50-MiB regressions, and align accepted formats with parser support or implement PDF ingestion.

Xunzhuo (approved):

Issue #3219, opened 2026-08-31
[Feature] Add human review, activation, and rollback handoff to the agent workflow


Xunzhuo:
### Goal

Turn an evaluated proposal into an explicit human review, activation, and rollback handoff without granting the agent deployment authority.

### Proposed Workgroup

Developer Experience & Ecosystem

### Scope

- Present the proposed diff, validation and evaluation evidence, affected deployment contract, and unresolved risks.
- Require an explicit authorized decision before activation.
- Invoke only maintained activation interfaces supplied by the owning deployment subsystem.
- Record the activated identity and provide a tested rollback target and operator instructions.
- Stop safely when authorization, preconditions, health, or rollback state is missing.

##### Parent Direction

Agent-assisted deployment, recipe, evaluation, and tuning workflows — #2977

### Non-goals

- Autonomous production activation.
- Bypassing Workgroup acceptance, code review, CI, or deployment authorization.
- Defining a second configuration history or rollback store.
- Managing organization, tenant, key, quota, or billing policy.

### Completion signal

A maintained example reaches a reviewable activation decision, records the resulting version when approved, and proves rollback through the owning deployment contract.

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #3218, opened 2026-08-31
[Feature] Integrate reproducible evaluation and tuning evidence into the agent workflow


Xunzhuo:
### Goal

Connect an agent-assisted proposal to reproducible evaluation and bounded tuning evidence before a human decides whether to activate it.

### Proposed Workgroup

Developer Experience & Ecosystem

### Scope

- Select maintained evaluation gates from the proposed change surface.
- Record the exact configuration, Recipe, model pool, dataset or probe set, environment class, and tool versions used.
- Preserve reports and routing traces as reviewable evidence.
- Compare the proposal with an explicit baseline and state unsupported or inconclusive results.
- Produce bounded tuning suggestions without rewriting protected configuration.
- Consume typed evaluation failures, including the decision-unresolved path tracked in #3209.

##### Parent Direction

Agent-assisted deployment, recipe, evaluation, and tuning workflows — #2977

### Non-goals

- Treating an evaluation run as acceptance or production approval.
- Automatically promoting a tuned proposal.
- Hiding failed, partial, or incomparable evidence.
- Defining Evaluation & Quality standards inside this workflow.

### Completion signal

One maintained journey produces a reproducible baseline comparison and reviewable tuning proposal with complete provenance and explicit failure states.

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #3217, opened 2026-08-31
[Feature] Generate and validate reviewable Router configuration and recipe proposals


Xunzhuo:
### Goal

Generate reviewable Router configuration and recipe proposals from user intent while preserving canonical schemas, provenance, and validation.

### Proposed Workgroup

Developer Experience & Ecosystem

Parent Epic: #2977

### Scope

- Translate a bounded user goal into a proposed configuration or Recipe change.
- Start from maintained fragments, Recipe packages, and deployment contracts instead of inventing fields.
- Preserve unrelated configuration and emit an inspectable diff with provenance.
- Run canonical parsing, schema, Recipe, and policy validation before presenting the proposal.
- Fail explicitly when the requested capability or environment is unsupported.
- Keep secrets and private environment details out of generated artifacts.

### Non-goals

- Accepting the proposal as roadmap work.
- Applying the proposal to a running or protected deployment.
- Inventing unsupported Router, model, gateway, or environment contracts.
- Replacing Workgroup ownership and review.

### Completion signal

A maintained example produces a deterministic, inspectable proposal that round-trips through canonical validation without losing unrelated configuration.

Xunzhuo:
Thanks @Mahdi-CV for owning this. Please make the first slice the proposal artifact contract, not an autonomous generator.

Given a canonical v0.3 input identity, bounded user intent, and maintained fragment or Recipe revisions, it should produce candidate YAML, a redacted inspectable diff, provenance, and a validation receipt without writing or applying anything. Reuse maintained `config/` assets and the existing Router validation contract; as #3477 lands, consume its structured diagnostics and diff rather than create a DevEx-specific validator or schema.

Keep the #3216/#3465 installation skill and broader #2990 journey work outside this issue. Demonstrate one deterministic example that preserves unrelated fields and fails explicitly for an unsupported request before adding LLM-assisted generation.

Issue #2977, opened 2026-08-22
[Epic] Build agent-assisted vLLM-SR deployment, recipe, evaluation, and tuning workflows


Xunzhuo:
## Parent Direction

Developer Experience & Ecosystem — #2970

## Goal

Provide a vLLM-SR agent workflow that guides supported environment discovery, produces reviewable configuration or Recipe proposals, collects reproducible evaluation and tuning evidence, and hands an authorized human an explicit activation and rollback decision.

## Current work

- #3216 — initial journey skill and supported environment discovery.
- #3217 — reviewable configuration and Recipe proposal generation.
- #3218 — reproducible evaluation and tuning evidence.
- #3209 — typed decision-unresolved errors in CLI, trace, and tuning tooling.
- #3219 — human review, activation, and rollback handoff.

PR #2990 is a candidate for #3216 after it is rebased, made green, and narrowed so it no longer closes this Epic.

## Safety and ownership

- Generation is not acceptance or production activation.
- Owning Workgroups retain technical acceptance for generated changes.
- Evaluation & Quality supplies gates; deployment owners supply activation and rollback contracts.
- The workflow must not invent unsupported environments or expose private infrastructure.

## Non-goals

- A general autonomous coding platform.
- Bypassing issue acceptance, code review, CI, or deployment authorization.
- Hiding generated changes, evaluation failures, or rollback state.
- Defining new routing, model, evaluation, or deployment semantics.

## Completion criteria

- A contributor can follow one documented journey from environment discovery to a validated proposal.
- The proposal carries reproducible evaluation evidence and explicit unsupported states.
- Activation requires an authorized human decision and records a tested rollback target.
- At least one maintained end-to-end example passes repository quality gates.

PR #3465, opened 2026-09-04
[Feature] Ship the public vLLM-SR agent install skill

### LINKED ISSUE #3216: [Feature] Ship the public vLLM-SR agent install and journey skill
### Goal

Ship a public, harness-neutral vLLM-SR skill and website **Agent** install mode so a user can give an agent one stable prompt, have it discover the supported environment, and receive a reviewable installation and usage plan.

### Proposed Workgroup

### Scope

- Publish the canonical skill at `https://vllm-sr.ai/install/agent/vllm-sr/SKILL.md` and keep that URL stable for agent consumption.
- Add an **Agent** option beside the existing curl install path on the official website.
- Provide a copyable prompt that asks the agent to install the skill, inspect the environment, install vLLM-SR through a maintained path, validate the result, and explain the next supported step.
- Keep the skill generic across compatible agent harnesses; OpenClaw-specific integration may reuse it but is not the canonical contract.
- Discover only maintained local, container, and deployment paths represented by current repository contracts.
- Show a bounded plan and ask before overwriting configuration, changing an existing deployment, activating production traffic, or handling credentials.
- Document prerequisites, unsupported states, recovery guidance, and links into the follow-up configuration, evaluation, tuning, and activation workflow.
- Add link, structure, prompt, and focused journey validation so the website and published skill cannot drift.
- Reuse the useful parts of PR #2990 after it is rebased and narrowed to this issue.

Suggested website prompt:

> Install the vLLM Semantic Router skill from https://vllm-sr.ai/install/agent/vllm-sr/SKILL.md. Then use it to inspect this environment and install vLLM Semantic Router through a supported path. Show me the plan first, validate the installation, and ask before changing an existing configuration or deployment.

### Non-goals

- Generating or applying Router configuration and Recipes; tracked by #3217.
- Running evaluation or tuning; tracked by #3218.
- Activating production traffic or performing rollback; tracked by #3219.
- Bypassing repository review, deployment authorization, or secret-handling boundaries.
- Closing the parent Epic #2977 with this first journey.

### Completion signal

- The official install page exposes both curl and Agent modes.
- The stable public skill URL returns a complete, generic skill package or entry document.
- A supported agent can follow the copyable prompt from discovery through a validated installation without silent mutation.
- Unsupported environments, existing deployments, credentials, and activation decisions stop at an explicit human handoff.
- Automated checks cover the public URL, prompt, skill structure, and at least one maintained end-to-end installation journey.

Parent Epic: #2977

Zi-Yi-Ming:
Closes #3216.

## Purpose

Issue #3216 asks for a public, harness-neutral installation skill that any compatible coding agent or agent harness that can consume the public skill document can fetch and execute to install vLLM Semantic Router through a supported path. It is the initial installation journey of Epic #2977; downstream issues #3217 (config/recipe), #3218 (evaluation/tuning), and #3219 (activation/rollback) own the rest of the agent journey, so this skill is narrowly scoped to installation only.

## Test Plan

- `website/scripts/lib/agent-install-skill.test.mjs` adds 23 `node:test` cases: canonical artifact exists with valid frontmatter, all required sections (including `## Workflow` per the harness template) are present, the public URL and the maintained installer with agent-safe flags are referenced, the runtime table matches the current installer contract (and rejects `podman` until #3443 lands), runtime state is separated from active deployment semantics, no pip alternative path is defined, out-of-scope items are declared, the installation page contains the skill URL with plan-first and approval-boundary semantics, the sync script produces a byte-for-byte identical static copy, the static copy is gitignored, Workflow stops on `command -v` (not only on `--version` success), Workflow checks default absolute paths (not only PATH), Existing CLI installation detects all signals including override paths, Workflow detects stale launchers and dangling symlinks (`test -e` / `test -L`, not only `test -x`), Workflow detects installer override env vars, `VLLM_SR_PIP_SPEC` presence stops the flow with its value never printed, completion setup requires explicit approval before installer execution, and declining completion changes stops the installer.
- `website/scripts/lib/agent-install-skill.e2e.test.mjs` adds an isolated install smoke test that runs the maintained `install.sh` with agent-safe flags (`--mode cli --runtime skip --no-launch`) inside a temporary HOME / install root / bin dir, verifies `vllm-sr --version` succeeds, confirms no automatic serve was started, asserts isolation from the real user HOME, and cleans up the temp tree. Skipped on Windows where the installer cannot run; runs on Linux CI.
- The skill is registered in `tools/agent/skill-registry.yaml`, listed in `tools/agent/docs/skill-catalog.md`, and added to `tools/agent/repo-manifest.yaml` skills list so `make agent-validate` accepts it.

## Test Result

`npm test` passes all tests.

Review feedback addressed:
- Static copy is now gitignored, matching the `install.sh` pattern (Xunzhuo).
- Existence checks use `test -f` instead of `ls` to avoid noisy error output (Copilot).
- Validation section adds absolute-path fallback for when `~/.local/bin` is not on `PATH` (Codex).
- Content-length test assertion relaxed since structural assertions already cover quality (Copilot).
- Workflow and existing-installation safety gates now stop whenever `command -v vllm-sr` finds a launcher; `--version` is used for diagnostics only, and contract tests pin this behavior (Xunzhuo).
- Workflow now checks default absolute paths (`~/.local/bin/vllm-sr` and `~/.local/share/vllm-sr`) in addition to `command -v`, so an existing install is detected even when `~/.local/bin` is absent from the non-interactive shell `PATH`. Two new contract tests pin the three-signal detection (Xunzhuo).

`git diff --check` is clean.

Review feedback addressed (round 3):
- Discovery now treats `VLLM_SR_INSTALL_ROOT` and `VLLM_SR_BIN_DIR` as path overrides: when set, the corresponding override path is checked as an occupied location, not only the defaults.
- `VLLM_SR_PIP_SPEC` is treated as a package override: its presence is a stop condition, its value is never printed (it may embed credentials such as private index tokens), and the install step explicitly unsets it so the official package cannot be silently replaced. The install step now unsets the installer's entire documented `VLLM_SR_*` override surface (mode, runtime, install root, bin dir, channel, pip spec, python, platform, auto-launch) so no inherited value can silently change the install.
- Launcher detection uses `test -e ... || test -L ...` so stale non-executable files and dangling symlinks are treated as occupied, not only executable files.
- The plan discloses that the installer sets up shell completions, which may edit shell rc files.

`git diff --check` is clean.

Review feedback addressed (round 4):
- Completion setup is now part of the pre-install approval boundary, not merely disclosed. The skill states the current installer always attempts shell completion setup and exposes no opt-out; completion may modify shell rc files such as `~/.bashrc` or `~/.zshrc`, and these changes are part of the installation mutation boundary.
- If the user does not approve the shell rc / completion changes, the skill stops without invoking the installer and does not claim a completion-free installation path. A prior instruction that already clearly authorizes the complete installation including completion setup satisfies the gate.
- Added contract coverage for both the approval and decline paths.
- No changes were made to `install.sh`.

`git diff --check` is clean.


website/static/install/agent/vllm-sr/SKILL.md
```diff
@@ -0,0 +1,275 @@
+---
+name: vllm-sr-install
```
Xunzhuo: why we had two vllm-sr skills copy?

Zi-Yi-Ming:
> where we had two vllm-sr skills copy?

The canonical source lives at `tools/agent/skills/vllm-sr-install/SKILL.md`. The copy at `website/static/install/agent/vllm-sr/SKILL.md` is the **build-time public artifact** that Docusaurus serves at `https://vllm-sr.ai/install/agent/vllm-sr/SKILL.md`.

The sync script (`website/scripts/sync-public-agent-skill.mjs`) copies canonical → static so the two stay byte-for-byte identical. This mirrors how `sync:install` already works for `install.sh` — the maintained source is in the repo root, and a copy is placed under `website/static/` so it gets a stable public URL.

The drift test (`agent-install-skill.test.mjs`) asserts the two are identical so they never diverge. If you prefer a different approach (e.g. symlink, or serve from `tools/agent/` directly), happy to adjust.

Zi-Yi-Ming:
To clarify why the static copy is needed: issue #3216 requires the skill to be reachable at a stable public URL (`https://vllm-sr.ai/install/agent/vllm-sr/SKILL.md`). Docusaurus only serves files under `website/static/`, so without the copy there it cannot publish the skill to that URL.

The canonical source stays in `tools/agent/skills/` so `make agent-validate` and the harness skill-routing still find it via `skill-registry.yaml`. The sync script + drift test keep them identical.

Zi-Yi-Ming:
Good catch — you were right, the static copy should not be committed.

Fixed in the latest push: `website/static/install/agent/` is now gitignored, same as `website/static/install.sh` already is. The canonical source stays in `tools/agent/skills/vllm-sr-install/SKILL.md`, and `sync-public-agent-skill.mjs` generates the static copy at build time. The drift test now runs the sync script first and then verifies parity.

website/static/install/agent/vllm-sr/SKILL.md
Zi-Yi-Ming: > where we had two vllm-sr skills copy?

Good catch — you were right, the static copy should not be committed.

Fixed in the latest push: `website/static/install/agent/` is now gitignored, same as `website/static/install.sh` already is. The canonical source stays in `tools/agent/skills/vllm-sr-install/SKILL.md`, and `sync-public-agent-skill.mjs` generates the static copy at build time. The drift test now runs the sync script first and then verifies parity.

tools/agent/skills/vllm-sr-install/SKILL.md
```diff
@@ -0,0 +1,280 @@
...
+4. Direct the user to
+  [Troubleshooting](https://vllm-sr.ai/docs/troubleshooting/common-errors).
+
+## Next Supported Step
+
+After a successful install:
+
+- **Start the local stack:** `vllm-sr serve`
+- **Choose a deployment:** <https://vllm-sr.ai/docs/installation/deployment-options>
+- **Configuration guide:** <https://vllm-sr.ai/docs/installation/configuration>
```
Xunzhuo, line 311: i think we should find out better ways to let agent gave accurate config schema, since docs are not always synced with new api and configurations inside router.

we may figure out adding an API inside router

Zi-Yi-Ming: Agreed — a live schema source in the Router would be a more reliable foundation for the config-generation workflow than relying on documentation alone. It also fits well with keeping the proposal grounded in the Router's canonical validation contract.

Xunzhuo (dismissed):
One install-safety path still falls through: discovery records `command -v vllm-sr`, but Workflow step 2 stops only when `vllm-sr --version` succeeds. A broken or stale launcher is still an existing installation, so the flow can overwrite it without approval. Please stop whenever `command -v` finds an installation, report version failure separately, and pin that branch in the contract tests.

Zi-Yi-Ming:
> One install-safety path still falls through: discovery records `command -v vllm-sr`, but Workflow step 2 stops only when `vllm-sr --version` succeeds. A broken or stale launcher is still an existing installation, so the flow can overwrite it without approval. Please stop whenever `command -v` finds an installation, report version failure separately, and pin that branch in the contract tests.


Thanks for the catch. Fixed in `989fb31`.

- The workflow now stops whenever `command -v vllm-sr` finds an existing launcher; `--version` is diagnostic only.
- The Existing CLI installation section now follows the same safety gate.
- Added contract tests covering both version-success and version-failure paths.
- `npm test`: 26 passed, 1 skipped; `git diff --check` clean.

FAUST-BENCHOU (dismissed):
LGTM. The install skill now keeps mutations behind the existing-installation and approval gates, while staying on the canonical CLI-only installer path. The isolated install smoke test also covers the main user-state safety boundary well.

wilsonwu (dismissed):
LGTM, thanks!

Xunzhuo (dismissed):
Thanks—the command-v and version gate is fixed. One install-safety gap remains: discovery only checks PATH, while the skill itself notes that ~/.local/bin is often absent from noninteractive PATH. In that common case an existing default install is missed, and install.sh can upgrade the same venv and rewrite its launcher without approval. Please also check the default absolute launcher or install root, stop on either signal, and add a no-PATH existing-install contract case.

Zi-Yi-Ming:
> Thanks—the command-v and version gate is fixed. One install-safety gap remains: discovery only checks PATH, while the skill itself notes that ~/.local/bin is often absent from noninteractive PATH. In that common case an existing default install is missed, and install.sh can upgrade the same venv and rewrite its launcher without approval. Please also check the default absolute launcher or install root, stop on either signal, and add a no-PATH existing-install contract case.

Thanks for the follow-up. Updated in 292e40fc: the Workflow and Existing CLI installation sections now check three signals — command -v vllm-sr (PATH), test -x ~/.local/bin/vllm-sr (default absolute launcher), and test -d ~/.local/share/vllm-sr (default install root). Any signal triggers the stop-and-report gate, so an existing install is detected even when ~/.local/bin is absent from the non-interactive shell PATH. Two new contract tests pin the three-signal detection.

Xunzhuo (changes requested):
Thanks—the absolute-path fix closes the previous gap. The agent-safe install boundary still has three silent-mutation paths: the command inherits `VLLM_SR_INSTALL_ROOT`, `VLLM_SR_BIN_DIR`, and `VLLM_SR_PIP_SPEC` while discovery checks only defaults; `test -x ~/.local/bin/vllm-sr` misses stale non-executable files and dangling symlinks that the installer can replace or write through; and `install.sh` always runs completion setup that edits shell config, but the bounded plan does not disclose it. Please neutralize or detect the override paths, treat any existing launcher path as occupied, and include completion changes in the approval-bound plan, with contract coverage.

Zi-Yi-Ming:
> Thanks—the absolute-path fix closes the previous gap. The agent-safe install boundary still has three silent-mutation paths: the command inherits `VLLM_SR_INSTALL_ROOT`, `VLLM_SR_BIN_DIR`, and `VLLM_SR_PIP_SPEC` while discovery checks only defaults; `test -x ~/.local/bin/vllm-sr` misses stale non-executable files and dangling symlinks that the installer can replace or write through; and `install.sh` always runs completion setup that edits shell config, but the bounded plan does not disclose it. Please neutralize or detect the override paths, treat any existing launcher path as occupied, and include completion changes in the approval-bound plan, with contract coverage.

Thanks for the close read — all three silent-mutation paths are closed in e3c07783.

Override env vars: The Environment Discovery table now reads VLLM_SR_INSTALL_ROOT, VLLM_SR_BIN_DIR, and VLLM_SR_PIP_SPEC. Workflow step 2 and the Existing CLI installation section check the corresponding override path as an occupied location when each var is set.

Stale launchers and dangling symlinks: Replaced test -x ~/.local/bin/vllm-sr with test -e ~/.local/bin/vllm-sr || test -L ~/.local/bin/vllm-sr, so any regular file (even without the executable bit) or any symlink (even dangling) triggers the stop-and-report gate.

Completion setup disclosure: The Plan Before Mutation section now discloses that the installer runs vllm-sr completion install, which may edit ~/.bashrc, ~/.zshrc, or equivalent shell rc files. A new Gotcha records the behavior.

Three new contract tests pin each behavior.

Xunzhuo (changes requested):
`VLLM_SR_PIP_SPEC` still bypasses the supported-install boundary. The workflow checks only the path overrides and then runs the installer with the inherited environment, while `install.sh` uses any non-empty `VLLM_SR_PIP_SPEC` as the package spec and prints it. The discovery table also tells the agent to echo that value, which may contain credentials. Please treat only the variable’s presence as a stop without printing its value, or explicitly unset it for the install, and pin both the alternate-package and secret-redaction cases.

Zi-Yi-Ming:
Thanks for the continued scrutiny — all addressed in e3c0778, 308c592, and 93e5320:

Override paths: Discovery, Workflow, and the Existing CLI section check VLLM_SR_INSTALL_ROOT / VLLM_SR_BIN_DIR as occupied locations when set.
Stale launchers: test -e ... || test -L ... replaces test -x, so stale files and dangling symlinks trigger the stop gate.
Completion disclosure: The plan now discloses that vllm-sr completion install may edit shell rc files.
VLLM_SR_PIP_SPEC: presence-only detection and its presence stops the flow — the value is never printed, since it may embed credentials and would install an alternate package.
Full override surface: the install step unsets all nine documented VLLM_SR_* overrides before running the installer.
Contract tests pin each behavior.

Zi-Yi-Ming:
Housekeeping: rebased onto the latest main — same content, new head 908bd22. The commits referenced above now resolve as 43fa06c, a51565d, and 908bd22.

Zi-Yi-Ming:
Housekeeping: the branch was rebased onto the latest main — same content, current head 568ebb9. The commits referenced in the review reply above now resolve as 61b8d2c (override paths / stale launchers / completion disclosure), e0e8d18 (VLLM_SR_PIP_SPEC stop + redaction), and 8f98fc8 (full VLLM_SR_* override surface cleared).

Xunzhuo (changes requested):
Thanks—the inherited-override and existing-installation paths are fixed. One safety contract is still inaccurate: the skill says an interactive installer lets users skip completion, but `install.sh` unconditionally runs `vllm-sr completion install` and exposes no completion opt-out. Following the documented “avoid shell rc changes” path can therefore still edit rc files. Please either add and use a real installer opt-out in the agent flow, or stop when the user does not approve completion changes, and cover that path in the contract tests.

Zi-Yi-Ming:
Implemented the second option from your feedback.

* Completion setup is now part of the pre-install approval boundary.
* The skill explicitly states that the current installer has no completion opt-out and that completion setup may modify shell rc files.
* If the user does not approve those changes, the skill stops without invoking the installer and does not claim a completion-free installation path.
* Added contract coverage for both the approval and decline paths.

No changes were made to `install.sh`.

Xunzhuo (commented):
Thanks—the install contract now matches the maintained installer: it discloses shell-startup-file changes, requires approval before invoking the installer, and stops if the user declines instead of claiming a nonexistent opt-out. The previous blocker is fixed. The new contract tests are only exposed through `website`’s `npm test`, though, and none of this head’s CI jobs invokes that script. Please wire that suite into PR CI and resolve the remaining review threads, then re-request review.

Zi-Yi-Ming:
Sorry to bother you, @Xunzhuo , but I’m not quite sure about the right direction here.

I was about to finish the CI wiring and remaining review threads based on your latest review. However, after #3538 was merged, #3465 now conflicts with the previous Agent Harness registry/catalog model that #3538 removed.

Since #3216 is still about shipping the public, harness-neutral `vllm-sr-install` skill, should I adapt #3465 to the new architecture? If so, what would be the intended integration point for the public skill?

I want to avoid reintroducing the old Agent Harness model just to resolve the conflicts.

Zi-Yi-Ming:
Closing this as superseded.

Upstream removed the skill registry this PR builds on — `tools/agent/skill-registry.yaml`, `tools/agent/docs/skill-catalog.md`, and `tools/agent/repo-manifest.yaml` are all gone from `main`. `main` also ships an "Agent" install tab in the Quickstart that already covers the documentation intent of this PR. Rebasing would reintroduce a deliberately deleted subsystem, so this branch is obsolete.

I'll extract any still-valuable, standalone hardening (install-skill semantics, the E2E drift tests, and the `install.sh` pattern alignment) into separate PRs based on current `main`.

Thanks for the earlier feedback on this.

PR #3468, opened 2026-09-04
[Bug] Advertise each embedding model's configured sequence limit

### LINKED ISSUE #3407: [Bug] Qwen3 embedding advertises a 32768 sequence limit the attention path cannot allocate for
### Proposed Workgroup

### Describe the bug

`GET /api/v1/embeddings/models` reports `max_sequence_length: 32768` for the Qwen3 embedding model, hardcoded at `ffi/embedding.rs:1814` and surfaced through `route_model_info_embeddings.go:42`.

The guard behind it is a positional one, not a memory one. `embedding_forward` rejects `seq_len > max_position_embeddings` with an error naming the limit (`qwen3_embedding.rs:2016`), and that field is read from the model config and validated to be at least 32768 (`qwen3_embedding.rs:164`). So the guard and the advertised constant describe the same thing — the RoPE cache — and neither is backed by what the attention can allocate.

`compute_attention_standard` materializes the full `[batch, num_heads, seq_len, seq_len]` score matrix, in F64 — Q, K and V are each cast up before the matmul — so it costs twice an F32 equivalent. #3382 measured this path on CPU, in words: 2235 MiB at 2048 words, 9461 MiB at 4096, and above 15 GiB at 8192, where the measurement was stopped.

That leaves a window where an input passes the guard and then exhausts memory instead of returning an error. Subword tokenization puts the token count above the word count, so the exact token figure differs, but the shape is clear well before 32768: memory roughly quadruples per doubling, and the last completed point is already 9461 MiB.

### To Reproduce

```
curl -s localhost:8080/api/v1/embeddings/models | jq '.models[] | select(.model_name|test("qwen3"))'
```

Then request an embedding for an input of a few thousand tokens — well inside the advertised limit.

### Expected behavior

The advertised `max_sequence_length` is a length the model can serve, or a request above what it can serve is rejected the way an over-length one already is.

### Impact

Any deployment with the Qwen3 embedding model loaded. The advertised value is what a client uses to size its input, and it currently describes the RoPE cache rather than anything the runtime was measured against.

### Additional context

Sibling of #3388, which reports the Gemma side: there the length check is missing entirely and a long input fails with a candle index error. Qwen3 has the check — it just enforces a bound that memory does not back.

Bounding the attention memory is #3382, which lists this call site. Once that lands the servable length changes, and whatever it becomes should be the number the API reports rather than 32768 staying hardcoded.

Two smaller things found in the same file:

- `compute_attention_scores` (`qwen3_embedding.rs:1176`) has no callers anywhere in `candle-binding` or `onnx-binding`. It builds the same full score matrix and can be deleted.
- The compiler will not flag that. `dead_code` does not fire in this crate: an unused private method and an unused free function both compile without a warning, while a deliberate type error in the same file is reported normally. No `allow` attribute, `[lints]` section or `RUSTFLAGS` is involved; the crate is built as `rlib` + `staticlib` + `cdylib`.

subin9:
Closes #3407

Builds on #3460 (first two commits); this PR is the third.

## Purpose

- `get_embedding_models_info` hardcodes 32768 for Qwen3 and 8192 for Gemma. Gemma's config says 2048, so the API advertises a length the model rejects.
- Read `max_position_embeddings` and `hidden_size` from the loaded model's config instead.

## Test Plan

- `cargo test --release --no-default-features --lib test_embedding_model_info`

## Test Result

- 3/3: loaded model reports its config, a 2048-position Gemma reports 2048, an unloaded model reports zeros.

### DIFF
--- a/candle-binding/src/ffi/embedding.rs
+++ b/candle-binding/src/ffi/embedding.rs
@@ -1754,6 +1754,30 @@ pub extern "C" fn free_batch_similarity_result(result: *mut BatchSimilarityResul
     }
 }
 
+/// One entry of the models-info table.
+///
+/// `limits` is `Some((max_position_embeddings, hidden_size))` from the loaded
+/// model's config and `None` when the model is not loaded, in which case both
+/// numbers are reported as 0 and the path as empty.
+pub(crate) fn embedding_model_info(
+    name: &str,
+    path: Option<&str>,
+    limits: Option<(usize, usize)>,
+) -> crate::ffi::types::EmbeddingModelInfo {
+    use std::ffi::CString;
+    let (max_sequence_length, default_dimension) = match limits {
+        Some((max_positions, hidden)) => (max_positions as i32, hidden as i32),
+        None => (0, 0),
+    };
+    crate::ffi::types::EmbeddingModelInfo {
+        model_name: CString::new(name).unwrap().into_raw(),
+        is_loaded: limits.is_some(),
+        max_sequence_length,
+        default_dimension,
+        model_path: CString::new(path.unwrap_or("")).unwrap().into_raw(),
+    }
+}
+
 /// Get information about loaded embedding models
 ///
 /// This function returns metadata about all available embedding models,
@@ -1770,7 +1794,6 @@ pub extern "C" fn get_embedding_models_info(
     result: *mut crate::ffi::types::EmbeddingModelsInfoResult,
 ) -> i32 {
     use crate::ffi::types::{EmbeddingModelInfo, EmbeddingModelsInfoResult};
-    use std::ffi::CString;
 
     if result.is_null() {
         eprintln!("Error: null pointer passed to get_embedding_models_info");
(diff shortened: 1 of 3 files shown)


FAUST-BENCHOU (dismissed):
LGTM. Deriving the advertised sequence limit and embedding dimension from the loaded model config keeps the FFI metadata aligned with the model that is actually serving requests.

wilsonwu (dismissed):
LGTM, thanks!

Xunzhuo (commented):
Thanks — the models-info API now reports limits and dimensions from the models actually loaded, and the focused Qwen, Gemma, and unloaded-state tests cover that contract. This branch also contains the full still-open #3460 stack and is behind `main`; please let #3460 land, rebase this change onto current `main`, and let the resulting required checks pass before approval.

Xunzhuo (changes requested):
The stacked attention change has a Qwen3Guard prefix-cache blocker. `process_cached_suffix` appends the complete suffix, then generation re-feeds its last token; from the second generated step onward, the physical KV cache is one entry ahead of the logical `q_offset`. The causal range then excludes the newly appended token and attends the stale duplicate instead. Please align the cache and offset handling and add cached-versus-uncached multi-token generation coverage before carrying this stack.

subin9:
@Xunzhuo Restacked on the fixed #3460 so this carries the prefix-cache fix as well. Once #3460 lands I'll rebase onto main so it's a single commit, as you asked :)

Xunzhuo (commented):
Thanks—the cached-suffix path now leaves the final token for decode, the multi-token regression covers cached versus uncached behavior, and the model limits come from the loaded configs and are enforced consistently. I do not see a remaining code blocker on this exact stacked head; approval should wait for #3460 to land so this can be rebased as the standalone current-main change.

Xunzhuo (dismissed):

Issue #3284, opened 2026-09-01
[Test] Prove shadow failure isolation and gated flywheel rollback end to end


Xunzhuo:
## Goal

Prove, end to end, that shadow evaluation cannot affect the primary response and that dataset qualification, promotion, and rollback gates behave deterministically.

## Proposed Workgroup

Evaluation & Quality

## Background

Shadow and flywheel components have separate unit contracts, but release confidence requires behavioral journeys across dispatch, judging, storage, qualification, gray rollout, and rollback. This issue owns those tests and diagnostics; implementation fixes remain with the responsible feature owners.

Parent Epic: #1519. Related bounded multi-arm behavior: #3376.

## Scope and approach

- Add E2E cases for challenger and judge timeout, overload, malformed output, abstention, partial-arm failure, cancellation, redaction failure, and partial storage failure.
- When multi-arm comparison is available, verify byte-identical normalized inputs, aggregate shadow limits, randomized opaque identities, and complete per-arm provenance.
- Assert that shadow work remains observation-only and cannot replace, alter, or create unbounded delay for the primary response.
- Exercise provenance rejection, failed qualification, rejected promotion, successful gray rollout, and deterministic rollback.
- Register scenarios in the maintained E2E coverage matrix with prerequisites, ownership, and release-readable diagnostics.

## Non-goals

- Implement multi-arm dispatch, judging, storage, training, or promotion features in this test issue.
- Run every training backend or deployment permutation in every pull request.
- Treat HTTP success alone as behavioral coverage.

## Completion signal

- [ ] Maintained profiles prove primary-path isolation under every shadow and judge failure mode.
- [ ] Aggregate limits, privacy enforcement, provenance gates, promotion, and rollback are covered where their features exist.
- [ ] Failures identify the responsible contract and retain enough public-safe evidence to reproduce it.
- [ ] Release readiness reports expose missing or failing required flywheel profiles.

## Integration-test gate

This is a final E2E tracker, not a currently independent implementation task. Shadow failure-isolation coverage can begin after #3279. Multi-arm, judging/dataset, training, promotion, and rollback journeys are added only after #3376, #3280, #3282, and the promotion/rollback contract each exist. Keep it out of the claimable Weekly pool until one bounded current-feature test slice is ready.

1fanwang:
/assign

I'll cover the landed single-shadow isolation slice through Envoy and Postgres, without multi-arm or later flywheel stages.

PR #3482, opened 2026-09-04
[Feature] Bounded multi-arm shadow comparison with blinded judging

### LINKED ISSUE #3376: [Feature] Add bounded multi-arm shadow dispatch and capture
## Goal

Dispatch and capture multiple bounded shadow arms for the same approved request without changing or delaying the user-visible primary response beyond an explicit lifecycle bound.

## Proposed Workgroup

MoM & Routing

## Work item shape

This is a leaf under #2238. It owns multi-arm shadow dispatch, aggregate limits, failure isolation, and capture. Blinded judging and dataset construction are separate downstream outcomes owned by #3280.

## Dependencies

- #3279 owns the single-shadow isolation and capture foundation. This issue must reuse or generalize that path rather than add a second shadow subsystem.
- #2861 owns the shared aggregate budget and execution-evidence contract.
- #3280 consumes qualified captured arms for blinded judging and dataset export.
- #3284 owns final cross-feature E2E and rollback journeys.

## Scope and implementation approach

- Execute the primary route exactly as today.
- Normalize the same approved request and dispatch a configured bounded set of shadow arms.
- Reserve aggregate calls, tokens, cost, concurrency, response bytes, and wall time before dispatch, then reconcile completed, failed, timed-out, and cancelled attempts.
- Capture bounded per-arm identity, usage, latency, failure, redaction, and provenance without feeding any result into the live response.
- Bound any wait imposed on request lifecycle and cancel nonessential shadow work on disconnect or policy expiry.
- Preserve primary-path behavior under every shadow failure mode.

Extend the existing disabled-by-default shadow policy with bounded multi-arm configuration. No shadow output is returned to the client or applied to current routing.

## Non-goals

- Judge, rank, randomize, or select among captured arms; #3280 owns that work.
- Build training datasets or promote a model.
- Reimplement single-shadow transport from #3279.
- Run an unbounded tournament, recursive comparison, or side-effecting tool trajectory.
- Persist raw hidden reasoning.

## Completion signal

- [ ] All arms receive byte-identical normalized approved input.
- [ ] Aggregate limits are reserved before dispatch and reconciled afterward.
- [ ] Every failure leaves the primary response unchanged.
- [ ] Replay preserves bounded per-arm usage, latency, failures, redaction state, and provenance.
- [ ] Unit and E2E tests cover exhaustion, cancellation, partial arms, malformed output, disconnect, and primary-path isolation.

Bevisy:
Closes #3376

Bounded multi-arm shadow dispatch: replay the same approved request to several candidate models in observation mode, keep the primary response byte-identical, and enforce an aggregate per-request budget. 3 self-contained commits by milestone.

## commits
- **C1** bounded multi-arm shadow dispatch — normalized request replay to N arms (fire-and-forget, zero blocking); fault isolation (per-arm recover; failures never touch primary)
- **C2** aggregate budget — `shadow_dispatch.budget` (calls/tokens/cost/concurrency), hard pre-dispatch + soft reconcile budgets, parallel async dispatch, outcome normalization
- **C3** bounded per-arm capture — identity (opaque), usage, latency, failure, and provenance through the existing outcome channel; no response content

## design highlights
- Primary response untouched: shadows fire in a detached goroutine after destination resolution; all arm failures map to explicit outcomes in observability/Replay only
- Zero new storage surface: capture reuses `Recorder.AppendOutcome` — no store/backend changes
- Config zero-secret: arms reference model refs only; endpoints via existing backend resolution; credentials via the existing authorizer (primary path)
- Non-goals honored: no judging/ranking/selection among arms (#3280 owns it; no judging code is included), no unbounded tournament/recursion, no raw hidden-reasoning persistence, no tool execution in arms
- Per-arm input identical to the primary rendering: same normalized request, only the target model/format differ

## verification
- `go test -count=1 ./pkg/shadow/...` — aggregate budget admission/reconcile, multi-arm deduplication, dispatch bounds
- `CGO_ENABLED=0 go test ./pkg/config/... ./pkg/configschema/... ./pkg/dsl/...` green
- Contract surfaces updated together: DSL compiler round-trip test, vllm-sr validator/model tests, dashboard schemas (vitest), config fragment, regenerated v0.3 config schema
- Cross-feature E2E journeys remain with #3284's final E2E (issue dependency), not claimed here
- Follows loom: `git commit -s`; branch merged with latest `origin/main` (merge, no rebase)


Xunzhuo (changes requested):
The rewrite should stay within #3376 rather than independently owning #2861 budgeting, #3280 judging, and a second transport before #3279. Four runtime blockers also need resolving: `ShadowComparison` is absent from canonical v0.3 load/export; concurrent admission and post-completion budget checks can overshoot limits; runtime supplies no credential resolver (and judge has no auth); and replay can start too late/use the global recorder, dropping fast-path or isolated-store evidence. Please realign the scope and add canonical config/docs plus focused E2E coverage.

FAUST-BENCHOU (commented):
I found one blocking issue in the aggregate budget path.

src/semantic-router/pkg/shadow/budget.go
```diff
@@ -0,0 +1,107 @@
...
+			Err: fmt.Sprintf("budget: call limit reached (%d)", b.cfg.MaxCalls),
+		}, false
+	}
+	if b.cfg.MaxConcurrency > 0 && b.active >= b.cfg.MaxConcurrency {
+		return ArmResult{
+			Arm: armName, Model: model, Outcome: OutcomeSkipped,
+			Err: fmt.Sprintf("budget: concurrency limit reached (%d)", b.cfg.MaxConcurrency),
+		}, false
+	}
+	if b.cfg.MaxTokens > 0 && b.usedTokens >= b.cfg.MaxTokens {
```
FAUST-BENCHOU, line 60: Since all arms are admitted concurrently, `usedTokens` is still zero until an arm finishes and calls `reconcile()`. This means several arms can all pass the check and collectively exceed `max_tokens` by a large margin.

Should token/cost capacity be reserved at admission time, or should dispatch stop admitting new arms until completed usage is known?

Bevisy:
Thanks for the review. The aggregate-budget reservation and canonical v0.3 config fixes are pushed (commit 0e9b174cc). The remaining review items — credential resolution/judge auth and replay recorder alignment — need the #3279 single-shadow foundation to land first; I'll continue after aligning with that work.

src/semantic-router/pkg/extproc/processor_req_body.go
```diff
@@ -128,6 +128,10 @@ func (r *OpenAIRouter) handleModelRouting(request *llmprotocol.Request, original
 		},
 	}
 
+	// The primary destination is decided; shadow arms observe the same
+	// normalized request without blocking or altering this response.
+	r.dispatchShadowArms(ctx)
```
Xunzhuo: **Snapshot shadow input before launching the goroutine**

Impact: The detached shadow goroutine reads reqCtx.SemanticRequest while the primary branch immediately mutates the same request during provider preparation. Shadows can encode partial pre-policy state, and the unsynchronized shared request access is a data race.

Required change: Finish the applicable provider-policy mutation first, deep-clone the approved neutral request and immutable metadata synchronously, and pass only that snapshot into the goroutine; never capture the live RequestContext or request pointer.

src/semantic-router/pkg/shadow/budget.go
```diff
@@ -0,0 +1,113 @@
...
+func (b *Budget) tryEnter(armName, model string, reserveTokens int64) (ArmResult, bool) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.cfg.MaxCalls > 0 && b.usedCalls >= b.cfg.MaxCalls {
+		return ArmResult{
+			Arm: armName, Model: model, Outcome: OutcomeSkipped,
+			Err: fmt.Sprintf("budget: call limit reached (%d)", b.cfg.MaxCalls),
+		}, false
+	}
+	if b.cfg.MaxConcurrency > 0 && b.active >= b.cfg.MaxConcurrency {
```
Xunzhuo: **Queue at the concurrency limit instead of dropping arms**

Impact: With two arms and max_concurrency set to 1, only one arm ever runs and its identity depends on goroutine scheduling, so the configured comparison cannot occur.

Required change: Use a semaphore or worker queue so every call-budget-admitted arm waits for a slot and runs within the aggregate deadline; reserve skipped for exhausted calls, tokens, cost, or deadline rather than transient concurrency.

src/semantic-router/pkg/shadow/dispatch.go
```diff
@@ -0,0 +1,284 @@
...
+		outcome = OutcomeFailed
+		return budgetedFailure(arm, outcome, fmt.Sprintf("read response: %v", err))
+	}
+	if resp.StatusCode != http.StatusOK {
+		outcome = OutcomeFailed
+		return budgetedFailure(arm, outcome,
+			fmt.Sprintf("status %d: %s", resp.StatusCode, truncate(respBody)))
+	}
+
+	completion, parseErr := parseCompletion(respBody)
```
Xunzhuo: **Shadow arms cannot parse streaming responses**

Impact: For stream=true requests, shadow endpoints return SSE while runArm parses the complete body as buffered JSON, causing shadow evidence to fail.

Required change: Disable streaming in the cloned shadow request or implement SSE decoding, and add a streaming-primary integration regression.

Xunzhuo (commented):
Thanks for the update—I found 3 current-head issues that need to be addressed before this can merge.

src/semantic-router/pkg/config/plugin_shadow_dispatch_support.go
```diff
@@ -115,9 +115,14 @@ func (c *ShadowDispatchPluginConfig) Validate() error {
 	if c == nil {
 		return nil
 	}
-	if c.Enabled && strings.TrimSpace(c.Model) == "" {
+	if c.Enabled && strings.TrimSpace(c.Model) == "" && len(c.Arms) == 0 {
```
Xunzhuo, line 118: **Validate every configured arm against the backend registry**

Impact: `Validate` now accepts arms-only configs and checks only that arm names are nonempty, while `validateDecisionShadowDispatchPlugin` still resolves only `shadow.Model`. An enabled arms-only config is rejected against the empty legacy model, and extra arms with unknown models load successfully but fail live requests as `backend_unresolved`.

Required change: Validate every deduplicated entry from `ShadowModels()` against configured backends, and add full-config regressions for both arms-only and unknown additional-arm cases.

Xunzhuo (commented):
Thanks for addressing the latest dispatch-path findings. One new multi-arm configuration path still bypasses backend admission.

src/semantic-router/pkg/config/plugin_shadow_dispatch_support.go
Bevisy: Thanks, fixed in a26170da7 (now merged with latest main): `validateDecisionShadowDispatchPlugin` now iterates every deduplicated `ShadowModels()` entry and requires a configured backend, so arms-only configs are accepted when resolvable and any unknown extra arm fails at load with its name in the error. Full-config regressions added for both the arms-only and unknown-arm cases.

Xunzhuo (commented):
Thanks for the update. One current-base configuration integration issue still blocks the multi-arm path.

src/semantic-router/pkg/config/plugin_config.go
```diff
@@ -261,6 +261,15 @@ type ShadowDispatchPluginConfig struct {
 	Enabled bool `json:"enabled" yaml:"enabled"`
 	// Model is the configured logical model that receives the shadow copy.
 	Model string `json:"model" yaml:"model"`
+	// Arms lists additional shadow models beyond Model. Every configured arm
+	// receives the same approved request under one aggregate budget, so a
+	// single decision can compare several candidates on live traffic.
+	Arms []string `json:"arms,omitempty" yaml:"arms,omitempty"`
```
Xunzhuo, line 267: **Keep multi-arm validation consistent across authoring surfaces**

The fields now propagate, but two inputs still change meaning between surfaces. `compiler_plugin_configs.go:241` uses `getIntField` for integer budget values; that helper truncates a float, so `max_calls_per_request: 0.5` compiles to `0`, which means unlimited. `models.py:1259` uses `bool(self.arms)`, so an enabled arms-only config containing `arms=[" "]` passes the CLI while Go rejects the blank arm. Please reject fractional and negative integer-budget inputs in the DSL while preserving explicit zero, and trim/validate every CLI arm before testing presence, with matching cross-surface regressions.

Bevisy: Propagated end to end in f2cd99516 (merged with latest main):

- vllm-sr (`models.py`): `ShadowDispatchPluginConfig` gains `arms` and `budget` (`ShadowDispatchBudgetConfig` with `max_calls_per_request` / `max_tokens_per_request` / `max_cost_per_request` / `price_per_million_tokens` / `reserve_tokens_per_arm`). An enabled plugin now accepts `model`, `arms`, or both — mirroring the Go `Validate`.
- DSL: the compiler reads `arms` (array) and the nested `budget` object; decompilation already preserved both via the structured emitter. Compile+decompile round-trip test added.
- Dashboard: `dslCapabilityPluginSchemas` exposes `arms` (string[]) and `budget` (object) field schemas; its test asserts the new fields.
- `config/fragments/plugin/shadow-dispatch/sampled.yaml` documents `arms` and `budget`.

Verified: `pkg/config`, `pkg/dsl`, `pkg/shadow` go tests green; vllm-sr shadow/plugin tests green; dashboard vitest green. Note canonical `shadow_comparison` remains the legacy v0.3 surface (endpoint-carrying arms); folding it into the runtime plugin surface is tracked separately.

Bevisy: arms/budget now parse in vllm-sr, round-trip in the DSL compiler/decompiler (test added), and are exposed in the Dashboard schemas.

Bevisy:
Merged latest origin/main (no conflict). PR description and scope now align with issue #3376: judging/ranking among arms stays out of scope (#3280 owns it) and no judging code is included; the multi-arm dispatch, aggregate budget, and per-arm capture are unchanged.

Bevisy:
The aggregate budget now covers the per-request dimensions it was missing, and caps that cannot bind before dispatch are rejected at load time:

- `budget` gains a per-request in-flight bound (`max_concurrency_per_request`) and response-size accounting; failed and cancelled arms are accounted as well and keep their admission reservation.
- Combinations that cannot bind before dispatch are rejected at config load (`max_tokens_per_request` without `reserve_tokens_per_arm`, `max_cost_per_request` without `price_per_million_tokens`) instead of failing silently at run time.
- The arm set is resolved after the looper and sampling gates, so `same-as-primary` keeps its meaning.
- The new field is synced across the DSL parser/decompiler, the config fragment, vllm-sr models, the Dashboard schemas, and the plugin docs.

src/vllm-sr/tests/test_plugin_shadow_dispatch.py
```diff
@@ -86,3 +91,64 @@ def test_plugin_type_round_trips(self):
...
+        with pytest.raises(
+            PydanticValidationError, match="requires reserve_tokens_per_arm"
+        ):
+            ShadowDispatchPluginConfig(
+                enabled=True,
+                model="m",
+                budget=ShadowDispatchBudgetConfig(max_tokens_per_request=100),
+            )
+        with pytest.raises(
+            PydanticValidationError, match="requires price_per_million_tokens"
```
Xunzhuo: [P2] Match the cost-cap validation message actually emitted

This assertion deterministically fails: the validator raises `requires reserve_tokens_per_arm and price_per_million_tokens`, which does not match the contiguous regex `requires price_per_million_tokens`. Running `python -m pytest tests/test_plugin_shadow_dispatch.py -q` on this commit yields 24 passed / 1 failed, and the assertion stops the rest of this test from checking the other prerequisite. Match the emitted message or assert the two requirements separately so the regression suite passes while still verifying the intended validation.

Xunzhuo (changes requested):
The updated budget path passes the focused native shadow regressions, but the changed CLI suite still fails on the latest commit: 24 tests pass and the cost-cap validation test fails. Please correct the assertion below and retain coverage of both missing budget prerequisites. A full multi-arm deployment was not rerun.

Bevisy:
> The updated budget path passes the focused native shadow regressions, but the changed CLI suite still fails on the latest commit: 24 tests pass and the cost-cap validation test fails. Please correct the assertion below and retain coverage of both missing budget prerequisites. A full multi-arm deployment was not rerun.

The cost-cap case now asserts the router's message — `requires reserve_tokens_per_arm and price_per_million_tokens` — so the suite passes 25/25. Coverage of both prerequisites is retained: a token cap without `reserve_tokens_per_arm`, and a cost cap without `reserve_tokens_per_arm`, the latter with and without `price_per_million_tokens`. Latest main is merged in.

src/semantic-router/pkg/shadow/shadow_budget.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Reconcile accounts a finished arm attempt. A completed arm swaps its token
+// and cost admission reservations for the real usage; any other outcome keeps
+// them (the attempt may still have spent upstream compute, and this was
+// already counted at admission). Response bytes are always swapped for the
+// observed size, so a read that never happened releases its reservation.
+func (b *ShadowBudget) Reconcile(completed bool, inputTokens, outputTokens, responseBytes int64) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.bytes += responseBytes - b.reserveResponseBytes
```
Xunzhuo: [P2] Keep consumed response bytes charged after failed reads

`shadowDispatcher.call` leaves `responseBytes` at zero when the connector rejects an HTTP error or oversized response, even though it has consumed the body. With both byte limits set to 64, a real HTTP 400 body of 64 bytes (or HTTP 200 body of 65 bytes) therefore refunds the entire reservation here and permits another 64-byte arm. If that worker finishes before a later target is admitted, the aggregate response-byte cap is bypassed. Account for bytes consumed on failures, or retain the reservation when a read may have occurred.

e2e/testcases/shadow_dispatch_multiarm.go
```diff
@@ -0,0 +1,43 @@
...
+	run, err := openShadowDispatchRun(ctx, client, opts)
+	if err != nil {
+		return err
+	}
+	defer run.Close()
+	sessionID, err := run.primary(ctx, "vllm-sr/shadow-multiarm", shadowPrimaryCeiling)
+	if err != nil {
+		return err
+	}
+	record, err := run.replay(ctx, sessionID, 1)
```
Xunzhuo: [P2] Wait for both shadow outcomes before asserting them

`run.replay(..., 1)` delegates to `waitForShadowReplay`, which returns only when the record has exactly one outcome and rejects a record with more than one. The assertion below requires exactly two outcomes, so this test fails for a correct two-arm request regardless of which arm finishes first. Pass `2` to the replay wait so both completed arms can reach the assertion.

e2e/testcases/shadow_dispatch_multiarm.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"k8s.io/client-go/kubernetes"
+
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+)
+
+func init() {
+	pkgtestcases.Register("shadow-dispatch-multi-arm-completes", pkgtestcases.TestCase{
```
Xunzhuo, line 12: [P2] Include the new multi-arm case in the replay profile

Registering this name only makes it available in the global test registry. `e2e/profiles/router-replay/profile.go:GetTestCases` still selects the existing ten cases and omits `shadow-dispatch-multi-arm-completes`, so running the profile never exercises the new two-arm scenario even though its fixture is present. Add this case to the profile selection so the multi-arm regression actually runs.

Xunzhuo (changes requested):
The latest byte-budget reconciliation refunds bytes already consumed by failed HTTP reads; a native regression reproduces this with both an error response and an oversized response. The two-arm E2E also has an incompatible replay count and is omitted from the replay profile. Please fix these three points. The CLI suite (25 tests), focused Go budget/config/DSL checks, and 25 native shadow-dispatch tests pass. Previously reported fractional DSL budgets and per-arm CLI validation remain partially unresolved. A live Router/Envoy deployment was not run.

src/semantic-router/pkg/shadow/shadow_budget.go
Bevisy: Good catch — a failed arm keeps its byte reservation now, so an error or oversized body still counts against the cap.

e2e/testcases/shadow_dispatch_multiarm.go
Bevisy: Right, it waits for two outcomes now.

e2e/testcases/shadow_dispatch_multiarm.go
Bevisy: Added it to the profile's case list.

Issue #2365, opened 2026-07-05
[Feature] Add Router Memory extraction and persistence receipts


Xunzhuo:
### Goal

Complete the Router Memory receipt contract for response-side extraction and persistence.

Parent Epic: #2339

### Proposed Workgroup

Agentic & Context

### Current-main baseline

Merged PR #2519, tracked through #2520, already records request-side retrieval, rewrite, filtering, injection, fail-open status, result count, Router Replay fields, and bounded plugin metrics.

Response-side auto-store still launches extraction and persistence asynchronously without an equivalent terminal receipt. Failures are logged, but the routed request cannot explain whether extraction was disabled, skipped, scheduled, completed, rejected, cancelled, or failed.

### Scope

- Define stable content-free statuses and reasons for response-side memory extraction and persistence.
- Distinguish disabled, policy-blocked, skipped, scheduled, completed, cancelled, timeout, extraction failure, and store failure.
- Correlate the asynchronous result with the originating request without retaining raw prompt or memory content in aggregate telemetry.
- Bound extraction and persistence work by lifecycle-owned context, timeout, queue, concurrency, and shutdown behavior.
- Record Router Replay outcomes where durable completion is available and bounded metrics for all terminal states.
- Preserve successful user responses when the configured memory write path fails open.

### Non-goals

- Rebuild request-side retrieval receipts already delivered by #2519.
- Add a new memory algorithm or backend.
- Make memory persistence a synchronous requirement for returning the model response.
- Store raw memory content in metrics, logs, or default Replay diagnostics.

### Completion signal

- [ ] Every enabled response-side extraction attempt reaches one bounded terminal status.
- [ ] Scheduled work is lifecycle-owned and responds to cancellation, reload, and shutdown deadlines.
- [ ] Extraction and backend failures are distinguishable without exposing content.
- [ ] Fail-open response behavior remains unchanged.
- [ ] Unit and maintained E2E coverage exercise success, skip, timeout, backend failure, cancellation, and shutdown.

Xunzhuo:
Thanks @Kritoooo — please anchor this in the existing `scheduleResponseMemoryStoreText` / `MemoryExtractor` path. The important fix is not only new log labels: replace the detached `context.Background()` goroutine with runtime-generation-owned bounded work, and snapshot the replay ID, trusted scope, neutral history, and response data before enqueueing rather than retaining or mutating `RequestContext` after the response.

[...]

PR #3483, opened 2026-09-04
[Feature] Add response-side Router Memory extraction and persistence receipts

Kritoooo:
Related #2365

## Purpose

Response-side Router Memory auto-store was fire-and-forget: unbounded goroutines, no timeout, and failures visible only in logs, so a routed request could never say whether persistence was disabled, skipped, completed, or lost. This adds a bounded, lifecycle-owned write path that reports one content-free terminal outcome per attempt. Response handling does not wait for Memory persistence to complete, and persistence failures stay fail-open.

- **`memory.PersistenceRunner`** bounds the write path with a per-job timeout, worker concurrency, a bounded queue, and graceful retire on reload and shutdown. The response path validates identity and reserves capacity before bounded history snapshots; encoding and Memory Store writes run in the worker.
- **Terminal receipts.** Each attempt reports one terminal status — `disabled`, `policy_blocked`, `skipped`, `completed`, `rejected`, `timeout`, `cancelled`, `extraction_failed`, `store_failed` — with a stable reason. Accepted attempts also report `scheduled`, with Replay capacity reserved for both events. Storage errors, shutdown drain expiry, or process crashes can still prevent receipt delivery. `rejected` (queue full, or a retired pool) is deliberately its own status so shed writes can be alerted on separately from `skipped`.
- **Telemetry.** Bounded `memory_persistence` plugin metrics, request-correlated warning logs, and `scheduled` plus terminal Router Replay outcomes. No prompt or memory content reaches metrics, logs, or replay metadata.
- **`ProcessResponseWithHistoryCount` returns the stored-chunk count**, so a zero-write turn reports `skipped` / `no_write`; `ProcessResponseWithHistory` retains its existing error-only signature.
- **Config.** `global.stores.memory.persistence` with defaults, validation, sample configuration, and user documentation.

## Test Plan

Unit coverage added by this change:

- `PersistenceRunner`: scheduled receipt ordered before worker execution, completion, backend failure, explicit job outcomes, timeout, shutdown cancellation, queue-full and post-retire rejection, concurrency bound, shutdown draining, panic containment, and default bounds for non-positive settings.
- `pkg/extproc`: `scheduled` plus terminal Router Replay outcomes, `rejected` recorded as its own plugin-metric status rather than as `skipped`, and the jailbreak and auto-store gating paths.
- `pkg/memory` extractor: stored-chunk counts for zero-, one-, and two-write turns.
- `pkg/config`: persistence defaults survive a sparse `memory` override.
- E2E: persistence receipts reach Prometheus, and a dead memory backend stays fail-open.

## Test Result

The standalone `PersistenceRunner` tests pass under `-race` across repeated runs.

---
This PR has completed the main content. The remaining parts will be completed in the subsequent other PRs.

### DIFF
--- a/src/semantic-router/pkg/config/validator_memory.go
+++ b/src/semantic-router/pkg/config/validator_memory.go
@@ -26,10 +26,33 @@ func validateGlobalMemoryContracts(cfg *RouterConfig) error {
 	if cfg == nil {
 		return nil
 	}
-	return validateMemorySimilarityThreshold(
+	if err := validateMemorySimilarityThreshold(
 		cfg.Memory.DefaultSimilarityThreshold,
 		"global memory default_similarity_threshold",
-	)
+	); err != nil {
+		return err
+	}
+	return validateMemoryPersistence(cfg.Memory.Persistence)
+}
+
+// Zero values use runtime defaults; negative values are rejected.
+func validateMemoryPersistence(cfg MemoryPersistenceConfig) error {
+	if cfg.TimeoutSeconds < 0 {
+		return fmt.Errorf("global memory persistence timeout_seconds must be >= 0, got %d", cfg.TimeoutSeconds)
+	}
+	if cfg.Concurrency < 0 {
+		return fmt.Errorf("global memory persistence concurrency must be >= 0, got %d", cfg.Concurrency)
+	}
+	if cfg.Queue < 0 {
+		return fmt.Errorf("global memory persistence queue must be >= 0, got %d", cfg.Queue)
+	}
+	if cfg.ShutdownGraceSeconds < 0 {
+		return fmt.Errorf(
+			"global memory persistence shutdown_grace_seconds must be >= 0, got %d",
+			cfg.ShutdownGraceSeconds,
+		)
+	}
+	return nil
 }
 
 func validateDecisionMemoryContracts(cfg *RouterConfig) error {
--- a/src/semantic-router/pkg/config/validator_memory.go
+++ b/src/semantic-router/pkg/config/validator_memory.go
@@ -2,6 +2,13 @@ package config
 
 import "fmt"
 
+const (
+	// MaxMemoryPersistenceConcurrency limits worker allocation at startup and reload.
+	MaxMemoryPersistenceConcurrency = 64
+	// MaxMemoryPersistenceQueue limits queued snapshots at startup and reload.
+	MaxMemoryPersistenceQueue = 1024
+)
+
 // validateMemoryContracts validates the long-term-memory similarity threshold
 // wherever it can be configured: the global memory block
 // (default_similarity_threshold) and each decision's memory plugin
@@ -35,16 +42,17 @@ func validateGlobalMemoryContracts(cfg *RouterConfig) error {
 	return validateMemoryPersistence(cfg.Memory.Persistence)
 }
 
-// Zero values use runtime defaults; negative values are rejected.
+// Zero values use runtime defaults. Reject invalid resource bounds before any
+// persistence workers or queue entries are allocated.
 func validateMemoryPersistence(cfg MemoryPersistenceConfig) error {
 	if cfg.TimeoutSeconds < 0 {
 		return fmt.Errorf("global memory persistence timeout_seconds must be >= 0, got %d", cfg.TimeoutSeconds)
 	}
-	if cfg.Concurrency < 0 {
-		return fmt.Errorf("global memory persistence concurrency must be >= 0, got %d", cfg.Concurrency)
+	if cfg.Concurrency < 0 || cfg.Concurrency > MaxMemoryPersistenceConcurrency {
+		return fmt.Errorf("global memory persistence concurrency must be between 0 and %d, got %d", MaxMemoryPersistenceConcurrency, cfg.Concurrency)
 	}
-	if cfg.Queue < 0 {
-		return fmt.Errorf("global memory persistence queue must be >= 0, got %d", cfg.Queue)
+	if cfg.Queue < 0 || cfg.Queue > MaxMemoryPersistenceQueue {
+		return fmt.Errorf("global memory persistence queue must be between 0 and %d, got %d", MaxMemoryPersistenceQueue, cfg.Queue)
 	}
 	if cfg.ShutdownGraceSeconds < 0 {
 		return fmt.Errorf(
--- a/src/semantic-router/pkg/extproc/processor_res_memory.go
+++ b/src/semantic-router/pkg/extproc/processor_res_memory.go
@@ -3,7 +3,6 @@ package extproc
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/openai/openai-go"
 
@@ -11,8 +10,6 @@ import (
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/llmprotocol"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/memory"
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/logging"
-	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/observability/metrics"
-	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/routerreplay"
 )
 
 func (r *OpenAIRouter) scheduleSemanticResponseMemoryStore(
@@ -56,23 +53,28 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
 		return
 	}
 
+	// Snapshot request-owned state before dispatch. extractMemoryInfo deep-copies
+	// history; protocol encoding and parsing remain in the worker.
 	currentUserMessage := extractCurrentUserMessage(ctx)
+	sessionID, userID, history, infoErr := extractMemoryInfo(ctx)
+	historyCount := len(history)
+	extractor := r.MemoryExtractor
+	receipt := r.snapshotMemoryPersistenceReceipt(ctx)
 	r.memoryPersistence.Submit(ctx.TraceContext, memory.PersistenceJob{
 		Run: func(jobCtx context.Context) (memory.PersistenceOutcome, error) {
-			sessionID, userID, history, err := extractMemoryInfo(ctx)
-			if err != nil {
+			if infoErr != nil {
 				return memory.PersistenceOutcome{
 					Status: "skipped",
 					Reason: "memory_info_unavailable",
-				}, err
+				}, infoErr
 			}
-			extractorHistory, err := r.memoryHistoryForExtractor(history)
-			if err != nil {
+			extractorHistory, historyErr := r.memoryHistoryForExtractor(history)
+			if historyErr != nil {
 				return memory.PersistenceOutcome{
 					Status:   "extraction_failed",
 					Reason:   "history_encode_error",
 					FailOpen: true,
-				}, err
+				}, historyErr
 			}
 
 			logging.Infof(
@@ -81,10 +83,10 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
 				userID,
 				len(currentUserMessage),
 				len(currentAssistantResponse),
-				len(history),
+				historyCount,
 			)
 
-			storedCount, err := r.MemoryExtractor.ProcessResponseWithHistory(
+			storedCount, err := extractor.ProcessResponseWithHistory(
 				jobCtx,
 				sessionID,
 				userID,
@@ -100,9 +102,7 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
 			}
 			return memory.PersistenceOutcome{}, nil
 		},
-		Report: func(status, reason string, failOpen bool, cause error) {
-			r.recordMemoryPersistenceOutcome(ctx, status, reason, failOpen, cause)
-		},
+		Report: receipt.record,
 	})
 }
 
@@ -129,78 +129,4 @@ func (r *OpenAIRouter) memoryHistoryForExtractor(
 		return nil, fmt.Errorf("parse encoded history: %w", err)
 	}
 	return request.Messages, nil
-}
-
-func (r *OpenAIRouter) recordMemoryPersistenceOutcome(
-	ctx *RequestContext, status,
-	reason string,
-	failOpen bool,
-	cause error,
-) {
-	if status != "scheduled" {
-		metrics.RecordPluginExecution("memory_persistence", requestDecisionStateKey(ctx), status, 0)
-	}
-	r.appendMemoryPersistenceReplayOutcome(ctx, status, reason, failOpen)
-	if status == "scheduled" {
-		return
-	}
-
-	if !failOpen && cause == nil && status != "rejected" {
-		return
-	}
-
-	fields := map[string]interface{}{
-		"request_id": ctx.RequestID,
-		"status":     status,
-		"reason":     reason,
-		"fail_open":  failOpen,
-	}
-	if cause != nil {
-		fields["error_class"] = fmt.Sprintf("%T", cause)
-	}
-	logging.ComponentWarnEvent("extproc", "memory_persistence_outcome", fields)
-}
-
-func (r *OpenAIRouter) appendMemoryPersistenceReplayOutcome(
-	ctx *RequestContext,
-	status string,
-	reason string,
-	failOpen bool,
-) {
-	if ctx.RouterReplayID == "" {
-		return
-	}
-	recorder := ctx.RouterReplayRecorder
-	if recorder == nil {
-		recorder = r.ReplayRecorder
-	}
-	if recorder == nil {
-		return
-	}
-	phase := "terminal"
-	if status == "scheduled" {
-		phase = "scheduled"
-	}
-	if err := recorder.AppendOutcome(ctx.RouterReplayID, routerreplay.Outcome{
-		Timestamp: time.Now().UTC(),
-		Source:    "router",
-		Target:    "router",
-		TargetRef: "memory_persistence",
-		Verdict:   status,
-		Reason:    reason,
-		Metadata: map[string]string{
-			"kind":      "memory_persistence_receipt",
-			"phase":     phase,
-			"fail_open": fmt.Sprintf("%t", failOpen),
-		},
-	}); err != nil {
-		logging.ComponentWarnEvent("extproc", "memory_persistence_replay_outcome_failed", map[string]interface{}{
-			"request_id": ctx.RequestID,
-			"replay_id":  ctx.RouterReplayID,
-			"status":     status,
-			"reason":     reason,
-			"phase":      phase,
-			"error":      err.Error(),
-		})
-	}
 }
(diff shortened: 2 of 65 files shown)


Xunzhuo (changes requested):
Thanks for taking on the asynchronous receipt path. The response worker captures the mutable `RequestContext` while response processing can keep changing it, so the receipt can race or observe inconsistent state. Please snapshot the immutable inputs before dispatch. Store failures are also swallowed, leaving no durable failure signal; define and test that behavior. The new behavior-visible config needs its contract docs and a focused E2E as well.

src/semantic-router/pkg/extproc/processor_res_memory.go
```diff
@@ -34,47 +39,70 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
...
-		if err != nil {
-			logging.Errorf("Memory store failed: %v", err)
-			return
-		}
-		extractorHistory, err := r.memoryHistoryForExtractor(history)
-		if err != nil {
-			logging.Warnf("Memory store failed to encode neutral history: %v", err)
-			return
-		}
+	r.memoryPersistence.Submit(ctx.TraceContext, memory.PersistenceJob{
```
FAUST-BENCHOU, line 63: The async job still captures the mutable `RequestContext` and reads it later from the worker. Response processing can continue mutating that context after submission, so persistence may race with the request path or observe a different state than the one that was actually submitted.

I think the session/user/history and receipt metadata should be snapshotted before `Submit()` and only immutable values captured by the job.

FAUST-BENCHOU (commented):
I found one blocking issue in the asynchronous persistence path.

Xunzhuo (commented):
Thanks for the careful follow-up. The worker now receives detached history and receipt snapshots, and partial store failures produce terminal fail-open receipts, so the earlier blockers are addressed. I don’t see a new code blocker in this head. Approval is waiting for exact-head CI to finish and for the dashboard module-tidy jobs to rerun cleanly.

Kritoooo:
> Thanks for the careful follow-up. The worker now receives detached history and receipt snapshots, and partial store failures produce terminal fail-open receipts, so the earlier blockers are addressed. I don’t see a new code blocker in this head. Approval is waiting for exact-head CI to finish and for the dashboard module-tidy jobs to rerun cleanly.

Thank you for the quick response and constructive feedback! I was planning to wait for CI to finish before requesting another review, so I really appreciate you taking a look so quickly.

src/semantic-router/pkg/extproc/processor_res_memory.go
Kritoooo: Thank you for pointing this out. The latest changes include a fix for this issue.

Kritoooo:
It seems that the ci still failed. I will continue to try to solve it and will request a review as soon as possible afterwards.

Xunzhuo (changes requested):
Thanks—the snapshot and terminal-failure fixes are in place, but the response path is still not nonblocking: disabled, policy-blocked, queue-full, and shutdown receipts call `recorder.AppendOutcome` inline, which can wait for the replay backend timeout while the model response is pending. Queue-full and shutdown also record `failOpen=false` even though persistence was dropped and the response succeeds. Please decouple receipt I/O from response delivery, mark these drops as fail-open, and add a slow-recorder saturation regression.

Kritoooo:
> Thanks—the snapshot and terminal-failure fixes are in place, but the response path is still not nonblocking: disabled, policy-blocked, queue-full, and shutdown receipts call `recorder.AppendOutcome` inline, which can wait for the replay backend timeout while the model response is pending. Queue-full and shutdown also record `failOpen=false` even though persistence was dropped and the response succeeds. Please decouple receipt I/O from response delivery, mark these drops as fail-open, and add a slow-recorder saturation regression.

Thanks for pointing this out. I’ve moved memory persistence receipts to a bounded background queue, marked queue-full and shutdown persistence drops as fail-open, and added regressions for slow recorders, saturation, and submission after the receipt queue has closed.

One design choice here is that the asynchronous receipt writer lives in `routerreplay`, rather than adding another asynchronous dispatcher to `memory`. The memory runner remains responsible for memory persistence, while the replay runtime owns receipt I/O, queue capacity, cancellation, and draining before its stores close. This seems like the more appropriate ownership boundary, since the work being dispatched is a replay storage operation.

While looking through the complete response path, I also noticed a remaining coupling: response-side replay operations still perform synchronous storage reads and updates. The background memory receipt writer can contend with those operations on the same backend’s locks or serialized writes. So receipt submission itself no longer waits for storage, but the response path can still indirectly wait on receipt I/O.

Fully removing that coupling looks like a broader change to the response-side replay pipeline, including metadata updates, tool-trace merging, and lifecycle finalization. We would also need to preserve ordering and shutdown behavior, and add regression coverage that verifies actual response delivery while the shared backend is blocked.

Do you think we should make the entire response-side replay path asynchronous to address this remaining contention? If so, would you prefer that work in this PR or in a separate follow-up? I’d lean toward a separate PR given the scope, while keeping this one focused on asynchronous memory receipt submission and the fail-open corrections.

Xunzhuo (commented):
Thanks—the latest queue decouples receipt storage from response delivery, and queue-full or shutdown drops are now fail-open with saturation coverage. I do not see a new code blocker on this head. The current FAUST thread still needs to be marked resolved; after that and a rebase onto current `main`, this is ready for the final high-risk approval pass.

Kritoooo:
> Thanks—the latest queue decouples receipt storage from response delivery, and queue-full or shutdown drops are now fail-open with saturation coverage. I do not see a new code blocker on this head. The current FAUST thread still needs to be marked resolved; after that and a rebase onto current `main`, this is ready for the final high-risk approval pass.

Thanks for the review! I’ve resolved the FAUST thread and rebased onto the latest main. This is ready for the final high-risk approval pass whenever you have a chance.

src/semantic-router/pkg/memory/persistence_runner.go
```diff
@@ -0,0 +1,198 @@
...
+			reportSafely(job, "store_failed", "panic", true, nil)
+		}
+	}()
+
+	jobCtx, cancel := context.WithTimeout(context.WithoutCancel(traceCtx), r.timeout)
+	defer cancel()
+	stop := context.AfterFunc(r.baseCtx, cancel)
+	defer stop()
+
+	outcome, err := job.Run(jobCtx)
```
Xunzhuo, line 225: **Make the persistence timeout terminate the whole attempt**

Impact: The timeout is inspected only after job.Run returns. Built-in store writes call context-free GenerateEmbedding, so a stuck embedding produces no bounded terminal receipt; RetireAndWait can then give up and reverse cleanup closes the memory store while that worker may still use it.

Required change: Thread cancellation through extraction, embedding, and backend operations and keep generation resources owned until every worker has exited; RetireAndWait must not return in a state that lets MemoryStore.Close run under a live job. Add an ignores-context regression case.

src/semantic-router/pkg/routerreplay/recorder.go
```diff
@@ -409,7 +422,7 @@ func (r *Recorder) ListAllRecords() []RoutingRecord {
 
 // Releases resources held by the storage backend.
 func (r *Recorder) Close() error {
-	return r.storage.Close()
+	return errors.Join(r.DrainOutcomes(), r.storage.Close())
```
Xunzhuo: **Receipt-drain timeout does not protect storage teardown**

Impact: Recorder.Close can call storage.Close while the outcome worker remains inside AppendOutcome, exceeding the shutdown bound or racing an unprotected backend.

Required change: Do not close storage until the outcome worker exits; retain ownership and finish cleanup asynchronously, with a cancellation-ignoring AppendOutcome regression.

Xunzhuo (commented):
Thanks for the contribution—I found 2 current-head issues that need to be addressed before this can merge.

src/semantic-router/pkg/routerreplay/recorder.go
Kritoooo: Fixed: replay storage now closes only after the outcome worker exits, with deferred cleanup on drain timeout and a regression covering AppendOutcome that ignores cancellation.

src/semantic-router/pkg/memory/persistence_runner.go
Kritoooo: Thanks. Timeouts now emit one terminal receipt, and generation resources remain owned until workers exit, with regression tests covering jobs that ignore cancellation. Native embedding still cannot be interrupted mid-call. Would you like me to address binding-level cancellation in this PR as well?

Kritoooo:
Thanks @Xunzhuo. In 73de8f1, persistence timeouts now produce one terminal receipt, and memory/replay storage cleanup waits for the corresponding workers to exit. I added regressions for jobs and `AppendOutcome` calls that ignore cancellation.

The remaining gap is cancellation inside native embedding. The current Go/C/Rust interfaces do not carry a cancellation signal, so the new checks prevent subsequent writes but cannot stop an in-flight native call. A permanently stuck call therefore retains its worker slot and resources.

Cooperative cancellation would require threading a cancellation token through the bindings and Rust inference/batching paths, including queue waits and execution checkpoints. Handling a native call that never cooperates would require stronger isolation, such as a terminable worker process.

Would you like me to continue with binding-level cooperative cancellation in this PR? Is safe resource retention sufficient for native operations that still cannot be interrupted, or do you expect a hard termination boundary as well?

Kritoooo:
I found and fixed two additional receipt gaps in 5d5b7291:

- Queued jobs could remain `scheduled` indefinitely when all workers were stuck. Deadlines and cancellation reporting now start at acceptance, so queued jobs reach one terminal outcome independently and never execute after cancellation.
- Jailbreak-blocked responses returned before memory receipt handling. The response path now records the appropriate `policy_blocked` or `disabled` receipt before returning 403, preserving the existing `auto_store` controls.

src/semantic-router/pkg/extproc/processor_res_memory.go
```diff
@@ -34,47 +36,73 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
...
-		if err != nil {
-			logging.Errorf("Memory store failed: %v", err)
-			return
-		}
-		extractorHistory, err := r.memoryHistoryForExtractor(history)
-		if err != nil {
-			logging.Warnf("Memory store failed to encode neutral history: %v", err)
-			return
-		}
+	sessionID, userID, history, infoErr := extractMemoryInfo(ctx)
```
Xunzhuo, line 82: **Bound work before copying the request history**

Impact: `extractMemoryInfo` deep-copies the full semantic history (and optional stored response history) synchronously on the response path before the bounded runner admits the job. Queue-full or retired submissions still pay that cost, and concurrent large responses can allocate and delay delivery outside the configured queue/concurrency/timeout bounds; the missing-user branch also copies history that the worker never uses.

Required change: Validate identity first, reserve bounded runner capacity before cloning, and enforce an explicit history snapshot size/message bound, with a queue-full large-history regression.

src/semantic-router/pkg/extproc/processor_res_memory_receipt.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	recorder := receipt.recorder
+	if recorder == nil {
+		return
+	}
+	phase := "terminal"
+	if status == "scheduled" {
+		phase = "scheduled"
+	}
+	if !recorder.TryAppendOutcome(receipt.replayID, routerreplay.Outcome{
```
Xunzhuo: **Guarantee the terminal receipt for an accepted attempt**

Impact: Scheduled and terminal receipts both enter the same best-effort 256-slot queue. `TryAppendOutcome` returns false on saturation, and the new saturation test explicitly permits those events to be dropped, so an accepted attempt can retain `scheduled` with no terminal outcome. That contradicts this PR's one-terminal-receipt contract and makes replay unable to distinguish unfinished work from lost persistence.

Required change: Reserve/coalesce terminal state per accepted attempt (for example, replace its queued scheduled state) while keeping response delivery nonblocking, and assert that saturation still leaves exactly one terminal receipt.

src/semantic-router/pkg/memory/persistence_runner.go
```diff
@@ -0,0 +1,249 @@
...
+func (r *PersistenceRunner) worker() {
+	defer r.workers.Done()
+	for queued := range r.jobs {
+		<-queued.scheduledDone
+		r.run(queued)
+	}
+}
+
+func (r *PersistenceRunner) prepareJob(traceCtx context.Context, queued *queuedJob) {
+	jobCtx, cancel := context.WithTimeout(context.WithoutCancel(traceCtx), r.timeout)
```
Xunzhuo: **Detach only the trusted span context**

Impact: `context.WithoutCancel` preserves every inbound context value. This request context includes client-controlled OpenTelemetry baggage, so persistence retains values such as tokens after the request ends—potentially indefinitely when native work cannot unwind—and passes them to instrumented or custom stores.

Required change: Build the detached context from `context.Background()` plus only the W3C span context/tracestate, and add a test proving trace linkage survives while baggage and other request values do not.

Xunzhuo (commented):
Thanks for completing the timeout and cleanup work. Four remaining contract gaps need to be closed on this head.

src/semantic-router/pkg/memory/extractor.go
```diff
@@ -214,24 +215,29 @@ func (e *MemoryExtractor) ProcessResponse(
...
+// is returned even if an earlier chunk succeeded; successful writes are not rolled back.
 func (e *MemoryExtractor) ProcessResponseWithHistory(
 	ctx context.Context,
 	_ string, // sessionID (unused, kept for interface compatibility)
 	userID string,
 	userMessage string,
 	assistantResponse string,
 	history []openai.ChatCompletionMessageParamUnion,
-) error {
+) (int, error) {
```
Xunzhuo, line 235: **Keep the exported method source-compatible**

Impact: Changing `ProcessResponseWithHistory` from `error` to `(int, error)` breaks every existing Go caller that assigns its single documented return value, even though the new count is only needed by this internal receipt path.

Required change: Preserve the existing method signature as a wrapper and add a separately named counted method for the new caller, with coverage for both APIs.

Kritoooo:
Thanks for the detailed review. I've addressed all four points in 014cb7a8:

- Identity validation and capacity reservation now happen before history cloning, with explicit limits on history size and structure.
- Accepted attempts reserve capacity for both scheduled and terminal Replay events, so queue saturation cannot drop their terminal receipt. Storage and shutdown failure limits are documented.
- Background work starts from `context.Background()` and carries only the span context and tracestate, without baggage or other request values.
- `ProcessResponseWithHistory` retains its original error-only signature; the receipt path uses the new `ProcessResponseWithHistoryCount` method.

src/semantic-router/pkg/extproc/processor_res_memory.go
```diff
@@ -34,26 +37,64 @@ func (r *OpenAIRouter) scheduleResponseMemoryStoreText(
...
+	reservation := r.memoryPersistence.TryReserve(ctx.TraceContext, receipt.record)
+	if reservation == nil {
+		return
+	}
+	defer reservation.Abort(memory.PersistenceOutcome{Status: "extraction_failed", Reason: "snapshot_failed", FailOpen: true}, nil)
+	var retained []*responseapi.StoredResponse
+	if ctx.ResponseObjectState != nil {
+		retained = ctx.ResponseObjectState.ConversationHistory
+	}
+	if err := validateMemoryHistoryBudget(reservation.Context(), ctx.SemanticRequest.Messages, retained); err != nil {
```
Xunzhuo: **Include the assistant response in the snapshot budget**

Impact: This check accounts for request and retained history, but `currentAssistantResponse` is materialized before the eligibility checks and then captured by every admitted job without any size bound. With the default 8 workers plus 64 queued jobs, sufficiently large upstream responses can therefore retain multiple gigabytes and defeat the 1 MiB snapshot limit this path is meant to enforce.

Required change: Include the assistant-response bytes in the same pre-queue budget (and avoid materializing them before eligibility/admission where possible), then add a regression showing that an oversized response is rejected without occupying persistence capacity.

Xunzhuo (commented):
Thanks for working through the earlier issues. Those are resolved, but the new snapshot bound still leaves the assistant response unbounded.

Kritoooo:
Thanks for catching this. I’ve included the assistant response in the shared 1 MiB per-task snapshot budget, moved the budget check before capacity reservation, and deferred assistant text assembly for memory persistence until after admission.
I also fixed the memory schema and integration-test configuration issues introduced by this PR. The remaining CI failures I checked are due to pre-existing upstream issues: a missing Fusion analysis_mode schema field and missing ONNX feedback-classification APIs. I’ll request another review once those upstream fixes land and CI passes.

src/semantic-router/pkg/config/validator_memory.go
```diff
@@ -26,10 +26,33 @@ func validateGlobalMemoryContracts(cfg *RouterConfig) error {
...
+	}
+	return validateMemoryPersistence(cfg.Memory.Persistence)
+}
+
+// Zero values use runtime defaults; negative values are rejected.
+func validateMemoryPersistence(cfg MemoryPersistenceConfig) error {
+	if cfg.TimeoutSeconds < 0 {
+		return fmt.Errorf("global memory persistence timeout_seconds must be >= 0, got %d", cfg.TimeoutSeconds)
+	}
+	if cfg.Concurrency < 0 {
```
Xunzhuo: **Bound persistence concurrency and queue size**\n\nImpact: This validation accepts arbitrarily large positive values, and NewPersistenceRunner immediately allocates a channel of queue entries and starts concurrency goroutines. A schema-valid typo can therefore exhaust memory or stall startup/reload before any request is served.\n\nRequired change: Enforce documented upper bounds for both values, reflect the range in the JSON schema, and add boundary tests before constructing the runner.

Xunzhuo (commented):
Thanks for tightening the persistence lifecycle. One startup resource bound is still missing.

Xunzhuo (commented):
Thanks for adding the persistence caps; one Kubernetes startup path still bypasses them.

src/semantic-router/pkg/config/validator_memory.go
```diff
@@ -26,10 +33,34 @@ func validateGlobalMemoryContracts(cfg *RouterConfig) error {
...
+	return validateMemoryPersistence(cfg.Memory.Persistence)
+}
+
+// Zero values use runtime defaults. Reject invalid resource bounds before any
+// persistence workers or queue entries are allocated.
+func validateMemoryPersistence(cfg MemoryPersistenceConfig) error {
+	if cfg.TimeoutSeconds < 0 {
+		return fmt.Errorf("global memory persistence timeout_seconds must be >= 0, got %d", cfg.TimeoutSeconds)
+	}
+	if cfg.Concurrency < 0 || cfg.Concurrency > MaxMemoryPersistenceConcurrency {
```
Xunzhuo, line 51: **Validate persistence bounds before Kubernetes startup allocation**

`config_source: kubernetes` returns from `validateConfigStructure` before `validateConfigContracts`, but startup still constructs the ExtProc server and persistence runner from the static `global.stores.memory.persistence` values before the controller applies CRDs. A large `concurrency` or `queue` therefore reaches goroutine/channel allocation and can exhaust the process even though the post-CRD validator rejects it later. Run the global persistence bounds during the initial Kubernetes parse (or immediately before runner construction), and add a Kubernetes startup regression proving out-of-range values fail before allocation.

Kritoooo:
> Thanks for adding the persistence caps; one Kubernetes startup path still bypasses them.

Thank you for pointing this out. It has been resolved in the latest commit.

Xunzhuo (approved):
The resource-scope synchronization preserves reverse-order cleanup and keeps shared resources alive until persistence workers finish. Scoped race tests passed across three runs, and the exact-head Core, Storage, and Memory Integration checks passed. No remaining blocker found.

wilsonwu (approved):
LGTM

PR #3486, opened 2026-09-04
[Router] Accept Anthropic beta response fields so Claude Code streams decode

### LINKED ISSUE #3417: [Bug] Anthropic streamed responses with thinking telemetry fail decode, breaking Claude Code
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The Anthropic API puts two fields into streamed responses that the router's strict response decode refuses:

- `estimated_tokens` inside `thinking_delta` events (running token count for the thinking block — Claude Code elicits it by default);
- a `context_management` echo (`{"applied_edits": [...]}`) reporting which context edits the API applied for a request that asked for them.

The response wire model names neither. One unknown field fails the stream, and the router turns a successful upstream response into a 502 for the client.

### To Reproduce

1. Route a model to an Anthropic-format backend and point Claude Code at the router (`ANTHROPIC_BASE_URL`).
2. Send any prompt with thinking on (the client default). If the upstream manages to stream; the router fails decode on the first `thinking_delta` carrying `estimated_tokens`, or on the `context_management` echo when the request used context editing.

### Expected behavior

The stream decodes. Both fields are telemetry the router does not consume, so accepting and ignoring them in the decode is enough. Nothing is lost for the client: same-format responses pass through to the client unmodified, so Anthropic clients still receive both fields — accepting them only concerns the router's own decode.

### Impact

Claude Code with default settings gets 502s on streamed responses the upstream answered successfully. Both versions I tested elicit these fields (2.1.191 from June and current 2.1.258), so this is not tied to one release.

### Additional context

Same mechanism as #3413: strict decode refusing routine Anthropic wire traffic, this time on the response side.

Verified with real Claude Code sessions against the live Anthropic API through the codec, and with minimal hand-written stream fixtures. Router at upstream main @ c47e1ebfb (2026-09-03).

abhinav-m22:
## Purpose

Closes #3417.

Claude Code enables extended thinking and context editing by default, causing Anthropic API responses to include beta fields that the Router's Anthropic wire structs did not recognize:

- `estimated_tokens` on `thinking_delta`
- `context_management` on the message body and terminal `message_delta`

Because response decoding rejects unknown fields, these fields caused successful provider responses to fail.

This change adds the missing fields to the wire structs. The Router does not consume them, and `omitempty` keeps re-encoded output unchanged. Unknown-field rejection remains unchanged.

The non-streaming path returns 502 on decode failure. Streaming responses have already returned 200, so the stream is truncated and ends with a synthetic `event: error` frame.

## Test Plan

- `cd src/semantic-router && go test ./pkg/protocolcodec/ -race -count=1`
- Added tests covering all four beta-field positions and a control case confirming unknown fields still fail.
- Extended golden fixtures `018` and `045`.
- Regenerated goldens with `UPDATE_PROTOCOL_GOLDENS=1`.

## Notes

- Beta fields are classified as `extensions` based on the pinned Anthropic SDK schema revision.
- `estimated_tokens` is intentionally discarded since the authoritative usage data is already provided in `message_delta.usage`.
- Scope is limited to the three affected response positions. Other unmodelled beta fields are left for follow-up work.
- Request-side changes are out of scope.


AayushSaini101 (dismissed):
LGTM thanks : )

wilsonwu (dismissed):
LGTM

Xunzhuo (changes requested):
The protocol change looks sound, but this head is not ready to approve: Full pre-commit checks fails because `TestAnthropicBetaResponseFieldsDecode` has cognitive complexity 38 (limit 20). Please split the cases, rebase onto current main, and rerun exact-head CI; the protocol and integration lanes otherwise pass.

Xunzhuo (commented):
The cognitive-complexity blocker is fixed on this head: the beta-response cases are split into focused tests, and the protocol behavior still preserves strict unknown-field rejection. The branch is behind `main` and the new exact-head core, security, E2E, and image jobs are still running; please let them finish and update the branch before approval.

abhinav-m22:
@xunzhuo Split `TestAnthropicBetaResponseFieldsDecode` into four per-position tests and the ci looks green now. Please take a look

abhinav-m22:
Hi @AayushSaini101 / @Xunzhuo / @wilsonwu could you please take a look?

Xunzhuo (commented):

src/semantic-router/pkg/protocolcodec/stream_anthropic.go
```diff
@@ -69,6 +71,9 @@ type anthropicDeltaWire struct {
 	Container    json.RawMessage `json:"container,omitempty"`
 	StopDetails  json.RawMessage `json:"stop_details,omitempty"`
 	Citation     json.RawMessage `json:"citation,omitempty"`
+	// Beta thinking telemetry on every thinking_delta. Accepted and ignored;
+	// authoritative counts arrive in message_delta.usage. No diagnostic (D2).
+	EstimatedTokens *int64 `json:"estimated_tokens,omitempty"`
```
Xunzhuo, line 76: **Preserve beta telemetry on same-format streams**

Impact: The stream no longer fails decode, but the Anthropic-to-Anthropic path re-encodes every frame and strips estimated_tokens and context_management. Claude clients therefore lose running telemetry and context-edit acknowledgements, contrary to the linked issue's explicit same-format contract.

Required change: Carry these bounded fields through an opaque extension channel and re-emit them for Anthropic targets while continuing to drop or diagnose them cross-format; update the same-format regression fixture to require preservation.

abhinav-m22: Confirmed. Both fields are stripped same-format, and the test pins that.

On the premise though: "passes through unmodified" only holds for buffered responses, which have byte replay. Streams disable source preservation and re-encode every frame, so these fields never reached the client on the stream path - before this PR that frame failed decode and returned a 502, so nothing is being regressed here. Authoritative counts still arrive via message_delta.usage; it's the per-delta estimate and the context-edit echo that drop.

Preserving them needs new fields on the neutral llmprotocol types (Event.Opaque carries whole frames, not fields) and moves the drop diagnostic to encode time, since the decoder is target-blind. Happy to do that as a follow-up PR, or extend this one if you'd prefer... which works better?

src/semantic-router/pkg/protocolcodec/testdata/golden/capability/018-anthropic-official-response-fields-chat-out.json
```diff
@@ -72,6 +72,45 @@
...
+          "model": "routed-model",
+          "object": "chat.completion",
+          "usage": {
+            "completion_tokens": 1,
+            "completion_tokens_details": {
+              "reasoning_tokens": 0
+            },
+            "prompt_tokens": 2,
+            "prompt_tokens_details": {
+              "cached_tokens": 0
```
Xunzhuo, line 104: [P2] Refresh this golden case for the current chat usage encoding

This new `context_management` case omits `cache_write_tokens: 0`, while current main preserves that explicit zero in `prompt_tokens_details`. Merging this head with current main causes `TestGoldenCapabilityMatrix/018-anthropic-official-response-fields_to_chat` to fail on this one-field mismatch, so the protocol test suite cannot pass after the branch update. Regenerate this expected output against the current encoder and rerun the capability golden matrix.

Xunzhuo (changes requested):
The standalone-head golden check passes, but integration with current main fails the capability matrix because the new context-management fixture omits the explicit zero cache-write count. Please refresh that expected output and rerun the matrix. The other protocol and diagnostic tests pass; no live provider deployment was run.

src/semantic-router/pkg/protocolcodec/testdata/golden/capability/018-anthropic-official-response-fields-chat-out.json
abhinav-m22: @Xunzhuo Refreshed the 018 chat golden (adds the explicit cache_write_tokens: 0) and merged current main. Is a follow-up PR OK for the same-format preservation?

Xunzhuo (approved):
The optional Anthropic fields are handled without changing response content or authoritative usage, and the golden mismatch is fixed. The protocol suite passes, along with all 55 deployed Streaming, Anthropic Compatibility, and Responses API cases.

wilsonwu (approved):
LGTM

Issue #3122, opened 2026-08-30
[Website] Make background color consistent across components


AayushSaini101:
There are lots of components in the website in which Background color is consistent, we have to use proper blue color background in the components as show in the attachment

https://github.com/user-attachments/assets/ab9385f1-67c5-44da-abea-76d3b1595888

AayushSaini101:
@abhinav-m22 PTAL thanks : )

PR #3493, opened 2026-09-04
[Website] remove text width caps on page headers

abhinav-m22:
Related #3122

Follow-up to the review comment on the blog header line break.

We removed the width cap on homepage section intros earlier, but Blog, Research and Community use a different component (`PageHeader`) that still had one — so their ledes kept wrapping early.

## Before / After

| | |
|---|---|
| **Before** | image (2)|
| **After** | image |

## The cause

`PageHeader` capped its text block at `var(--measure)` (72ch). That cap sat on the **container**, and `ch` resolves against the container's own 1rem font — so it worked out to ~658px, not 72 characters of heading. On a 1536px frame that left ~870px empty to the right of every lede.

## Changed

- **`PageHeader`** — removed the cap, lede now uses `text-wrap: balance` like the homepage. Fixes Blog, Research and Community in one go since all three share it.
- **`Chrome.module.css`** — `.pageDescription` uncapped (was 44rem), title switched to `balance`. Used by paper viewer pages.
- **`work-groups.module.css`** — section lede uncapped (was 42rem).
- **`InstallQuickStartSection`** + **`YouTubeSection`** — two landing-page headings were still on `text-wrap: pretty` while every other heading uses `balance`.

## Not changed

Body copy inside cards and columns keeps its cap — docs intro card, steering-committee track copy, UseCaseExplorer panel intro, and the global `p { text-wrap: pretty }` for prose. Those aren't the eyebrow/title/lede lockup.

CSS only. Lint clean.


AayushSaini101 (approved):

wilsonwu (approved):
LGTM

FAUST-BENCHOU (approved):

Xunzhuo (commented):
The CSS scope looks coherent, and the preview matches the intended shared PageHeader behavior. I do not see a code blocker in this version. The exact head is still behind main and its baseline jobs are rerunning, so please rebase once more and let the new head go green; I will take the final pass then.

PR #2738, opened 2026-07-31
[Feature] add generation-owned router resource cleanup

altale:
Related #2470

Bounded drain and cancellation for stuck request or management leases remain tracked by #2470 and will be  completed in a follow-up PR.

## Purpose

`routerGeneration`, introduced in #2869, already gives each published router a request-drain boundary. This PR builds on that lifecycle seam and makes each generation release the runtime resources it owns.

The router build now records resource cleanup in a private `resourceScope` as resources are created. The same scope provides two paths:

- a failed candidate build closes everything created so far in reverse order;
- a committed generation closes its resources after its request references and management leases have drained.

`OpenAIRouter.Close` now delegates exclusively to this scope, avoiding a second field-by-field teardown path.

This covers the resources that need generation-scoped cleanup, including caches, Response API stores, replay recorders, classifier MCP clients, rate-limit providers, recipe model-selection registries, lookup-table workers, session state, compression recovery, and native KNN, KMeans, SVM, and MLP handles. Components without an owned resource do not gain a placeholder `Close` contract.

Selection registry publication is also made transactional. A candidate build keeps its registry private, `Swap` publishes the committed registry through the runtime registry, and the legacy fallback updates its process-wide compatibility pointer atomically. This removes the reload/read data race from #2470 without adding locks to request-time selection.

```mermaid
flowchart LR
    A[Build candidate router] --> B[Register owned resources]
    B --> C{Build succeeds?}
    C -- No --> D[Close scope in reverse order]
    C -- Yes --> E[Swap generation and publish registry]
    E --> F[Retire old generation]
    F --> G[Drain requests and management leases]
    G --> H[Close old scope in reverse order]
```

## Test Plan

```bash
# Exercise atomic selection-registry publication under concurrent reads.
go test -race ./pkg/selection \
  -run 'TestRegistry_ConcurrentReassignmentRaces' -count=1

# Exercise generation swap, drain, runtime publication, and teardown.
go test -race ./pkg/extproc \
  -run 'Test(RouterService|OpenAIRouterCloseDrainsManagementLeasesBeforeRecoveryStore|ReloadRouterFromConfigPublishesRuntimeRegistryAfterSwap|BuildRouterComponentsDoesNotPublishProcessGlobals|PublishRouterStateAdoptsCommittedRouterGlobals)' \
  -count=1

# Verify precise initialization rollback, ordered teardown, and selector cleanup.
go test -race ./pkg/classification \
[...]

PR #3495, opened 2026-09-04
[Feature] Add bounded router generation drain on shutdown

### LINKED ISSUE #2470: [Feature] Introduce generation-owned construction, reload, and graceful shutdown
### Goal

Introduce an immutable runtime generation that owns every closeable child, rolls back partial construction, leases active requests and streams, and drains safely on reload or process shutdown. Owner surfaces: `src/semantic-router/pkg/extproc/router.go`, `src/semantic-router/pkg/extproc/router_build.go`, `src/semantic-router/pkg/extproc/server.go`, `src/semantic-router/pkg/routerruntime/`, `src/semantic-router/pkg/services/`, and `src/semantic-router/cmd/runtime_bootstrap.go`.

### Proposed Workgroup

Data Plane & Networking

### Scope

##### Current behavior

Router construction has no complete rollback stack, reload replaces and immediately closes the prior router without a request lease, classification refresh can publish mixed state, and process shutdown drains only a subset of resources. Expanding close behavior without ownership and leases would create use-after-close risk.

##### Expected behavior

One versioned generation owns every resource, construction commits only after full success, each request observes one generation, and retirement stops admission, drains active work to a deadline, then closes children exactly once in reverse order.

##### Validation

- Fault-inject each constructor step and prove exact-once reverse cleanup.
- Run reload and classification refresh against long-lived requests under `go test -race`.
- Assert stable goroutine, descriptor, connection, and native-handle counts across repeated reloads.
- Add startup and loaded shutdown tests with bounded drain and no loss of acknowledged state.
- Run the router-core feature gate and affected AMD runtime profiles.

##### Related

- Parent audit: #2375
- Native runtime tracker: #2396
- Durable architecture: TD045 and TD020

### Non-goals

### Completion signal

- Own caches, classification snapshots, response and Replay stores, memory, rate-limit clients, workflow state, selectors, native schedulers, vector runtime, and servers in one generation.
- Use a reverse cleanup stack for constructor failure and join cleanup errors.
- Acquire and release generation leases for requests and streams; publish immutable classification state atomically.
- On reload, stop old admission, drain leases, retire native and MCP resources, and close idempotently.
- On process shutdown, stop servers and queued admission, drain work, close children, and synchronize hook registration.

altale:
Close #2470. This is a follow-up to #2738. 

## Purpose

Before this change, process shutdown was split across independent lifecycle paths. The ExtProc server handled signals internally, startup failures could terminate the process before deferred cleanup ran, and gRPC `GracefulStop` could wait indefinitely for stuck requests. Learning operations also used a separate lease barrier, so shutdown had no single bounded path that could preserve acknowledged state before closing generation-owned resources.

This change introduces one process lifecycle context and a bounded shutdown sequence:

- Move signal ownership to the command entrypoint and propagate cancellation through model download, runtime initialization, warmup, and serving.
- Attempt graceful server shutdown within a proportional budget, then force request cancellation while reserving the remaining deadline for generation drain.
- Use the generation lease barrier for both requests and learning operations, ensuring acknowledged state is drained before owned resources are closed.

Startup failures now return through the same teardown path, and shutdown errors from completed generation cleanup are preserved. The affected modules are the router command lifecycle, ExtProc server and generation lifecycle, management API shutdown, model download, and model runtime initialization.

The accepted issue is owned by `wg/data-plane-networking`.

## Test Plan

```bash
make agent-lint AGENT_BASE_REF=upstream/main
make test-and-build-local

cd src/semantic-router
go test -race \
  ./cmd \
  ./pkg/apiserver \
  ./pkg/extproc \
  ./pkg/modeldownload \
  ./pkg/modelruntime \
  -count=1

make agent-dev ENV=cpu
make agent-serve-local \
  ENV=cpu \
  AGENT_STACK_NAME=pr2470-merge \
  AGENT_PORT_OFFSET=7000
make agent-smoke-local \
  AGENT_STACK_NAME=pr2470-merge \
  AGENT_PORT_OFFSET=7000
make agent-stop-local \
  ENV=cpu \
  AGENT_STACK_NAME=pr2470-merge \
  AGENT_PORT_OFFSET=7000
```
The tests cover:

- Startup cancellation and loaded-process shutdown.
- Graceful drain, forced cancellation, and bounded generation cleanup.
- Preservation of acknowledged learning state and original task failures.

## Test Result

All required commands passed.

———
<details>

- [x] The title does not stack prefixes such as [Router][Docs]
- [x] The PR links accepted issue #2470, owned by wg/data-plane-networking

</details>


Xunzhuo (dismissed):
Thanks for tackling this lifecycle gap. One blocker: the advertised 30-second process bound ends before shutdown hooks run. `shutdownRouterProcess` first waits on `shutdownServers` with `processShutdownTimeout`, then calls `runShutdownHooks`; the vector-store hook can independently wait another 25 seconds, followed by tracing shutdown. A SIGTERM can therefore exceed the process deadline substantially. Please put hooks and tracing under the same deadline, preserve their errors, and add a slow-hook regression test.

FAUST-BENCHOU (commented):
This looks solid to me. Moving signal ownership to the process lifecycle and using the generation lease barrier gives shutdown a clear bounded path without closing resources underneath active requests or acknowledged learning work.

Xunzhuo (dismissed):
Thanks for rebasing. The shutdown path is still not bounded end to end: shutdownServers can consume the 30-second budget, then runShutdownHooks runs outside that deadline (the vector-store hook can take another 25 seconds), and tracing gets a separate 5 seconds. Hook and tracing failures are also only logged rather than joined into the returned shutdown error. Please drive servers, hooks, and tracing from one deadline, preserve all teardown errors, and cover a slow hook in the regression test.

Xunzhuo (changes requested):
One remaining blocker: process shutdown drains the management HTTP server and retires the router generation concurrently. An already accepted /v1/classify or /v1/memory handler does not hold a generation lease, while RouterService.Shutdown can close the classifier or memory store before http.Server.Shutdown finishes that request (for example, memory DELETE can complete Get and then call Forget on the closed store). This violates the #2470 drain-before-close and one-generation-per-request contract. Please either lease all generation-backed management operations for their full lifetime or finish draining management handlers before retiring the generation under the shared deadline, and add a shutdown regression that blocks an accepted management request until release.

altale:
@Xunzhuo Thanks for pointing this out. I implemented the second approach by draining accepted management requests before retiring the router generation. The shutdown is now split into serving and resource phases. Management, ExtProc, and metrics stop accepting new work concurrently. Router generation retirement and shutdown hooks only run after the management server has drained.

There is an important bounded-shutdown tradeoff here. If a management handler does not exit before the shared deadline, closing its generation-backed resources would reintroduce the use-after-close risk described in #2470. In that forced path, transports stop within the deadline, but generation resources and hooks are left open for process-level cleanup. IMO this is a necessary tradeoff to guarantee both bounded shutdown and the absence of use-after-close.

src/semantic-router/cmd/main.go
```diff
@@ -34,29 +55,151 @@ func main() {
...
+		return recordStartupError(startupWriter, "create ExtProc server", err)
+	}
 
-	warmupRouterRuntime(server, embeddingRuntime)
+	if err := warmupRouterRuntime(ctx, routerServer, embeddingRuntime); err != nil {
+		return recordStartupError(startupWriter, "warm up router runtime", err)
+	}
 	markRouterReady(startupWriter, startupEmbeddingProviderStatus(embeddingRuntime))
 	logStartupSummary(cfg, opts, embeddingRuntime.AnyReady)
 	startKubernetesControllerIfNeeded(cfg, opts.kubeconfig, opts.namespace)
```
FAUST-BENCHOU, line 112: One shutdown path still seems detached from the process lifecycle: the Kubernetes controller is started in a goroutine with its own `context.Background()`.

On SIGTERM it can therefore keep processing config updates while the Router is draining and closing generation-owned resources, and a controller error still calls `ComponentFatalEvent`, which exits immediately and bypasses the bounded teardown. Can the process context be propagated into the controller and its errors returned through the same lifecycle instead?

FAUST-BENCHOU (commented):
One remaining lifecycle concern in the Kubernetes controller path.

src/semantic-router/cmd/main.go
altale: Thanks, I have updated the implementation to bring the Kubernetes controller into the main process lifecycle. The controller-runtime manager now uses the shared lifecycle context, propagates unexpected failures to the main serving loop, and cancels peer serving components. ExtProc also waits for in-flight Kubernetes-triggered reloads before closing dependent router resources while keeping shutdown bounded.

altale:
Separately, I noticed that reload warmup has its own lifecycle issue. A warmup task may outlive cancellation while still using resources owned by an unpublished candidate router, so safely closing that candidate requires a dedicated ownership and drain model. This is a valid issue, but IMO it is separate from the published-generation and acknowledged-state guarantees in #2470, so I left the existing warmup behavior unchanged in this PR.

Xunzhuo (changes requested):
One shutdown blocker remains: `executorRun.execute` returns on parent cancellation while warmup tasks may still be running, but `WarmupRouter` calls the context-free `warmup.Load`. `runRouterProcess` then enters deferred teardown and can close the router generation while `LoadToolsDatabase` or `PreloadKnowledgeBases` is still using it; the new stuck-task test explicitly leaves that task alive after `Execute` returns. Please make in-flight warmups cancellable or join them before generation teardown, and add a regression proving their resources remain open until the task exits.

altale:
@Xunzhuo Thanks. I updated startup warmups to acquire a normal generation lease before accessing router-owned resources. Already-running warmups now participate in the existing drain barrier, while retirement prevents pending warmups from starting. Shutdown remains bounded by the process deadline without closing resources still in use.

src/semantic-router/pkg/extproc/server.go
```diff
@@ -217,36 +245,95 @@ func (s *Server) Start() error {
...
+// ShutdownResources retires the active router generation and closes it after drain.
+func (s *Server) ShutdownResources(ctx context.Context) error {
+	if ctx == nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), defaultGenerationDrainTimeout)
+		defer cancel()
+	}
+
+	return s.lifecycle.resources.run(func() error {
+		if err := s.lifecycle.waitForWatcher(ctx); err != nil {
```
Xunzhuo: **Drain detached reload work before closing router resources**

Impact: A reload that acquires reloadMu and passes isStopping just before shutdown can remain in prepare or warmup after ShutdownResources has waited only for the watcher loop and closed the service. Shutdown can return while owned reload work is still allocating resources, and that reload discovers closure only later at Swap.

Required change: Register every debounce and reload goroutine with the server lifecycle, cancel and await it under the shutdown context before service.Shutdown, and test the real scheduleReload path without manually wiring watcherDone to the synthetic reload.

Xunzhuo (commented):

src/semantic-router/pkg/extproc/server.go
altale: Thanks for the review. I’ve updated the lifecycle handling so all admitted debounce and reload goroutines are registered, canceled, and awaited before generation resources are closed. ExtProc and Kubernetes serving components are also joined before resource shutdown.

Xunzhuo (commented):
Thanks—the detached reload work is now drained; one exact-head quality-gate blocker remains.

src/semantic-router/cmd/main.go
```diff
@@ -34,33 +55,172 @@ func main() {
...
+	if err != nil {
+		return recordStartupError(startupWriter, "initialize runtime dependencies", err)
+	}
+	routerServer, err = extproc.NewServer(opts.configPath, opts.port, opts.secure, opts.certPath, runtimeRegistry)
+	if err != nil {
+		return recordStartupError(startupWriter, "create ExtProc server", err)
+	}
 
-	warmupRouterRuntime(server, embeddingRuntime)
+	if err := warmupRouterRuntime(ctx, routerServer, embeddingRuntime); err != nil {
```
Xunzhuo: **Avoid shadowing the startup error**

Impact: This `:=` shadows the `err` first declared at line 50; the exact-head Full pre-commit changed-file govet check fails here, so this head cannot pass the core quality gate.

Required change: Reuse the outer `err` (or choose a distinct name) and rerun the exact-head pre-commit check.

src/semantic-router/pkg/extproc/server.go
```diff
@@ -291,13 +441,13 @@ func (rs *RouterService) Swap(r *OpenAIRouter, publish func(acquire AcquireFunc)
 		}
 		return errors.New("router service is shutting down")
 	}
-	generation := &routerGeneration{router: r}
+	generation := newRouterGeneration(r)
 	old := rs.current.Swap(generation)
 	if publish != nil {
-		publish(rs.acquireFor(generation))
+		publish(generation.acquire)
```
Xunzhuo, line 464: **Lease memory operations across reloads**

Impact: `publish` carries the generation lease, but the same snapshot still exposes `MemoryStore` as a bare pointer. A memory handler can resolve the old store just before `Swap` retires and closes it; DELETE resolves the store twice, so it can verify ownership in the old backend and then delete from the new one.

Required change: Expose an atomic store-plus-release acquire, hold it for each whole memory request, and cover a blocked List, Get, or Forget racing `Swap`.

src/semantic-router/pkg/extproc/server_lifecycle.go
```diff
@@ -0,0 +1,109 @@
...
+	"sync/atomic"
+)
+
+type shutdownPhase struct {
+	once sync.Once
+	err  error
+}
+
+func (p *shutdownPhase) run(shutdown func() error) error {
+	p.once.Do(func() {
```
Xunzhuo, line 16: **Keep timed-out cleanup resumable**

Impact: `shutdownPhase.run` permanently caches the first result. If `stopAndWaitForBackgroundWork` times out while an admitted reload is still running, `ShutdownResources` never reaches `service.Shutdown`; that reload can later publish a generation, while every cleanup retry only returns the cached deadline and never retires it.

Required change: Make resource cleanup continue or retry once background work exits, and add a post-timeout regression that releases the reload and observes eventual generation closure.

src/semantic-router/pkg/extproc/server.go
```diff
@@ -217,36 +245,94 @@ func (s *Server) Start() error {
 			s.Stop()
 			return err
 		}
-	case <-signalChan:
+	case <-ctx.Done():
```
Xunzhuo, line 260: **Prevent serving after shutdown wins**

Impact: Cancellation returns here without stopping the gRPC server. `ShutdownServing` can consume its once-only phase while `s.server` is still nil; `StartContext` can then install and start `Serve`, observe the canceled context here, and leave that listener running against resources being retired.

Required change: Reject startup once stopping begins and coordinate this cancellation path with server stop, with a regression where shutdown wins before server installation.

Xunzhuo (commented):
Thanks for fixing the earlier shutdown paths. Three lifecycle races still need to be closed.

src/semantic-router/pkg/apiserver/route_classify.go
```diff
@@ -132,13 +128,15 @@ func (s *ClassificationAPIServer) handleBatchClassification(w http.ResponseWrite
 	if !ok {
 		return
 	}
+	service, release := s.acquireClassificationService()
```
Xunzhuo: **Acquire the config and classifier from one generation**

`parseBatchClassificationRequest` reads `max_batch_size` through `currentConfig()` before this service lease is acquired. A reload between those operations can validate a batch against generation A's limit and execute it on generation B's classifier; the metrics and model-availability paths similarly combine separately resolved config and service state. Expose an atomic config-plus-service acquisition, take it before validation, hold it through the response, and add a reload regression that changes the batch limit between those steps.

src/semantic-router/cmd/main.go
```diff
@@ -34,33 +55,172 @@ func main() {
...
+	servingErr := shutdownConcurrently(ctx, servingShutdowns...)
+	managementErr := <-managementDone
+	shutdownErr := errors.Join(managementErr, servingErr)
+	if errors.Is(managementErr, context.Canceled) || errors.Is(managementErr, context.DeadlineExceeded) {
+		return shutdownErr
+	}
+	if resourceShutdown != nil {
+		shutdownErr = errors.Join(shutdownErr, resourceShutdown(ctx))
+	}
+	shutdownErr = errors.Join(shutdownErr, runShutdownHooks(ctx, shutdownHooks))
```
Xunzhuo, line 192: **Do not run shared-resource hooks after resource drain times out**

If `resourceShutdown(ctx)` returns `DeadlineExceeded` or `Canceled` while serving, reload, or generation work is still active, this continues into hooks with the expired context. A hook can then close a shared dependency while the undrained work still uses it, violating the drain-before-close contract. Mirror the management-timeout guard here, or keep teardown resumable and run hooks only after resource shutdown actually completes; add a regression with blocked generation work and a hook-owned dependency.

Xunzhuo (commented):
Thanks—the latest update closes the three recent races. Two generation-boundary gaps remain.

Xunzhuo (approved):
Thanks—the generation lease and shutdown ordering issues are fixed, and the exact-head checks are green. This is ready from my side.

wilsonwu (approved):
LGTM

PR #3498, opened 2026-09-04
[Feature] Add token_spans.v1 contract and HTTP token classifier backend for PII

### LINKED ISSUE #2922: [Feature] Define the token_spans.v1 contract for PII backends
### Goal

Define a versioned token-span result contract that local and remote PII backends can implement without losing offsets or label semantics.

### Proposed Workgroup

### Scope

- Specify text offsets, label identity, score semantics, overlap handling, model identity, and optional diagnostics.
- Validate malformed and out-of-range spans.
- Add parity tests for representative local and remote implementations.

### Non-goals

- Redaction policy or routing actions.
- A universal observation registry.

Parent Epic: #2760

### Completion signal

- `token_spans.v1` is a value of the shared `backend.contract`, and PII resolves it through `ResolveRemoteClassifierBackend` — the same path category already uses for `label_distribution.v1`.
- Offset unit, bounds, overlap, truncation and score semantics are written down, and every rule has a fixture that fails when violated. Offsets are code points into the request string as received, following #3147; the other four are open and get settled here.
- A remote PII backend leaves `PIIDetected`, `PIIEntities`, `MatchedPIIRules` and the masked text identical to the local path across that fixture set.

adaamko:
Closes #2922

## Purpose

Adds `token_spans.v1` as a value of the shared `backend.contract` block from #2920 and gives PII its first remote backend through it, mirroring what #3138 did for category with `label_distribution.v1`.

- `pkg/config/classifier_backend.go`: `RemoteClassifierContractTokenSpans`, accepted in `Validate()`; `ValidatePIIModelBackend` (role `classification`, protocol `http_classify`, mutually exclusive with `use_mmbert_32k`). `PIIModel.Backend` added.
- `pkg/classification/http_token_classifier.go`: `HTTPTokenClassifierInference` implementing `TokenClassifierBackend`. Wire is the HuggingFace token-classification shape: POST `{"inputs": text}`, response a bare list or `{"spans": [...], "truncated_at": N, "model": ...}`, each span `label|entity_group`, `score`, `text|word`, `start`, `end`, optional `byte_start`/`byte_end`.
- Offsets are Unicode code points into the request string as received (#3147 settled this for the local path); converted to bytes once at the `TokenEntity` boundary. `text` must equal the code-point slice, which is what turns an off-by-one in the unit into an error instead of a mis-redaction.
- Labels matched by name against the PII mapping with BIO prefixes stripped; unknown label rejects the whole response, same strictness as `alignScoresToMapping`. Score in [0,1]. Exact duplicate spans rejected. Overlapping and nested spans of different types are allowed. A span ending after `truncated_at` is rejected; a response carrying `truncated_at` returns its spans with `ErrTokenSpansTruncated` so PII fails closed on partial results.
- `pkg/classification/pii_remote_backend.go` + lifecycle: a configured `backend:` replaces the local PII model; `IsPIIEnabled` accepts `Backend != nil`.
- `candle_binding.TokenEntity` comments corrected: `Start`/`End` were documented as character positions, the Rust side writes byte offsets (#3146).

Left for follow-ups, deliberately: the reserved `parameters.labels` zero-shot field is not sent; PII `on_error` does not exist yet, so partial results fail closed; overlap-safe masking (union of ranges) in `buildMaskedPIIText` is not touched; chunked-scan offset rebasing is #3373's contract and not implemented here.

## Test Plan

```
cd src/semantic-router
go vet ./pkg/config/ ./pkg/classification/
go test ./pkg/config/ -run 'TokenSpans|PIIModelBackend'
go test ./pkg/classification/ -run 'TokenSpan|HTTPToken|PII|Lifecycle|BuildClassifier'
```

`pkg/classification` links against the native bindings, so the test binary needs `LD_LIBRARY_PATH` pointing at `candle-binding`, `ml-binding` and `nlp-binding` `target/release` (built with `cargo build --release --no-default-features`); without it the package fails in 0.001s with no output.

15 golden fixtures in `pkg/classification/testdata/token_spans_v1_fixtures.json`, each listing code-point and byte offsets: ASCII baseline, CJK where bytes and code points disagree (the #3146 string), astral-plane emoji, combining mark, span at offset 0, span at end of text, overlapping types, nested types, and rejection cases for duplicate, text mismatch, end past text, inverted span, score out of range, unknown label, plus an entity past a 512-token window that must be reported or declared via `truncated_at`.

## Test Result

```
ok  pkg/config          0.055s
ok  pkg/classification  1.056s
--- PASS: TestHTTPTokenClassifierFixtures (15/15 subtests)
--- PASS: TestHTTPTokenClassifierTruncatedAt
--- PASS: TestHTTPTokenClassifierAliasesAndBytePair
--- PASS: TestNewHTTPTokenClassifierInferenceValidation
```

Run on a Linux host with the bindings built; `gofmt -l` clean; `go vet` clean with cgo. Remaining risk: parity of `PIIDetected`/`PIIEntities`/`MatchedPIIRules`/masked text between remote and local paths is exercised by the adapter tests on the fixture set, not yet by a side-by-side test against a running local model; that is the next PR under this issue.


Xunzhuo (changes requested):
`token_spans.v1` explicitly accepts overlapping and nested spans, but the unchanged masking path is not overlap-safe: `buildMaskedPIIText` replaces spans one by one using original byte offsets against an already modified string. A later-starting replacement can change the length, so an earlier overlapping span can be skipped (`End > len(maskedText)`) or sliced at the wrong boundary. That breaks #2922’s required masked-text parity and can leave PII unmasked. Please normalize spans to an offset-stable union, or otherwise make masking overlap-safe, and add an end-to-end masking fixture for nested and overlapping spans.

src/semantic-router/pkg/classification/pii_remote_backend.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	candle_binding "github.com/vllm-project/semantic-router/candle-binding"
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
+// piiHTTPBackend adapts a TokenClassifierBackend to PII's historical inference
+// interface, so PIIDetected, PIIEntities, MatchedPIIRules and masking keep
+// working unchanged whether the spans came from Candle or from a remote
+// token_spans.v1 provider.
```
FAUST-BENCHOU, line 15: One lifecycle issue here is that the remote PII backend owns an HTTP connector, but it is never closed when the classifier generation is retired.

`HTTPTokenClassifierInference` implements `Close()`, while `piiHTTPBackend` does not forward it, and `Classifier.Close()` currently does not close `piiInference` either. This leaves the old backend's transport/idle connections alive across reloads. Can the PII inference participate in the classifier close path like the other remote backends?

FAUST-BENCHOU (commented):
Found one lifecycle issue with the remote PII backend.

adaamko:
Thanks both, fixed in the second commit.

@Xunzhuo masking now merges spans into a union of disjoint byte ranges (longest span wins the placeholder when they nest) and replaces from the end backwards, so a later span cant shift an earlier one anymore. Single non-overlapping spans mask byte-identical to before. Tests: `classification_pii_masking_test.go` covers nested (PERSON in ADDRESS), overlapping (EMAIL in URL, both orders), partial overlap, adjacent at offset 0 / end of text, multi-byte with an overlapping pair, and out-of-range spans; the end-to-end case goes through `buildPIIResponse` with `MaskEntities` and asserts the masked text, and `pii_remote_backend_test.go` drives four overlapping/nested spans from an httptest token_spans provider through `piiHTTPBackend` and checks byte offsets slice back to the reported text.

@FAUST-BENCHOU `piiHTTPBackend` forwards `Close()` to the connector and `Classifier.Close()` now closes `piiInference` next to the jailbreak classifier, same test asserts the backend exposes Close.

Also fixed the CI: `TestReferenceConfigCoversCanonicalPublicSurface` skips `pii.backend` the same way it skips `domain.backend` (remote attachment is exclusive with the local selector the reference config shows), and the go1.25 gofmt alignment in `classifier_backend_test.go`.

adaamko:
Lint fixed in the third commit: gocognit flagged `alignTokenSpans` and the fixture test, both are split into small helpers now (label, bounds, text, score, byte pair, one per concern), no behavior change, same error strings, same assertions.

The Dashboard job failure is `dashboard/backend` go.sum being untidy on main, not from this branch, `go mod tidy` there changes 36 lines with this PR touching nothing under dashboard/.

Xunzhuo (commented):
Thanks—the overlap-safe masking and PII backend close path address my blockers. I’m holding approval until the branch is updated to current main, the exact-head required checks pass, and the existing lifecycle thread is resolved.

src/semantic-router/pkg/classification/pii_remote_backend.go
adaamko: Fixed in the second commit: piiHTTPBackend forwards Close() to the connector and Classifier.Close() closes piiInference next to the jailbreak classifier, pii_remote_backend_test asserts the backend exposes Close. Resolving.

Xunzhuo (changes requested):
Thanks for addressing overlap-safe masking and backend lifecycle. One fail-closed gap remains: decodeTokenSpansResponse accepts any JSON object without a spans member, including {} or a 200 {"error":"model unavailable"}, because unknown fields are ignored and env.Spans remains nil. That becomes a successful zero-entity result, so PII treats malformed provider output as clean. Please require an explicit spans array, reject null and error shapes, and add a regression.

adaamko:
@Xunzhuo all three are in: branch is on current main (ffe2bb64), the lifecycle thread is resolved, and the checks on cc43a5b9 are green.

adaamko:
> Thanks for addressing overlap-safe masking and backend lifecycle. One fail-closed gap remains: decodeTokenSpansResponse accepts any JSON object without a spans member, including {} or a 200 {"error":"model unavailable"}, because unknown fields are ignored and env.Spans remains nil. That becomes a successful zero-entity result, so PII treats malformed provider output as clean. Please require an explicit spans array, reject null and error shapes, and add a regression.

Thanks @Xunzhuo , spans and error stay raw in the envelope, an explicit spans array is required, and {}, null, {"spans":null}, a non-array spans member and any error member all reject the whole response. regression covers each of those plus the two legitimate empty lists ([] and {"spans":[]}).

Xunzhuo (changes requested):
Thanks for tightening the span decoder and overlap masking. I still see two blocking integration gaps. A declared truncated_at returns entities plus an error, but PII signal evaluation drops every cached result with an error, so the rule becomes a clean non-match; this also leaves the accepted PIIModel on_error allow or block path unimplemented. Separately, pii.backend is not wired through canonical normalization, validation, and mapping-load gates: a sparse backend inherits the local mmBERT selector and fails only during classifier construction, while a pure remote config skips the mapping loader required by the adapter. Please cover truncation through a real routing decision and the canonical parse-to-build or reload path.

adaamko:
> Thanks for tightening the span decoder and overlap masking. I still see two blocking integration gaps. A declared truncated_at returns entities plus an error, but PII signal evaluation drops every cached result with an error, so the rule becomes a clean non-match; this also leaves the accepted PIIModel on_error allow or block path unimplemented. Separately, pii.backend is not wired through canonical normalization, validation, and mapping-load gates: a sparse backend inherits the local mmBERT selector and fails only during classifier construction, while a pure remote config skips the mapping loader required by the adapter. Please cover truncation through a real routing decision and the canonical parse-to-build or reload path.

@Xunzhuo both fixed.

truncation: PIIModel now embeds the shared on_error contract. a truncated_at response keeps the spans the provider did return and they count against the rule; with on_error: block any content that was not fully classified (truncation or a backend error) matches as classification_error, same shape as the jailbreak path. TestPIISignalTruncatedResponseRoutesThroughOnError drives the six allow/block x truncated/error/partial combinations through a real PII rule.

wiring: pii.backend goes through the same gates as domain.backend now. the canonical normalizer clears the inherited use_mmbert_32k when a sparse override attaches a backend, ValidatePIIModelBackend (backend rules + on_error) runs in the global contract validators at parse, and IsPIIClassifierEnabled accepts a backend so NeedsPIIMappingForRouting loads the mapping for a remote-only config. reference config gains classifier.pii.on_error. tests: TestCanonicalPIIBackendReplacesInheritedLocalSelector, TestCanonicalPIIBackendRejectsExplicitLocalSelectorAtParse, TestPIIOnErrorValidatedAtParse, TestIsPIIClassifierEnabledAcceptsRemoteBackend.

one thing i kept deliberately: model_ref still resolves for a remote pii backend, so the local model directory is provisioned. the adapter needs pii_type_mapping.json and today that file ships inside the model repo; parameters.labels is reserved in the contract for the provider to supply labels itself, which would remove that dependency later.

Xunzhuo (changes requested):
Thanks for addressing the truncated-result and canonical-backend paths. Two runtime blockers remain:

- An unused or disabled PII backend is still constructed unconditionally. The current reachability path intentionally leaves `piiMapping` nil when no reachable PII signal exists, but backend construction then rejects that nil mapping and makes a dormant configuration fail startup. Please gate construction on an actual PII consumer and cover disabled/unreachable parse-to-build behavior.
- Chained overlaps do not preserve the longest original span's placeholder. After merging A `[0,10)` with B `[9,18)`, the code compares C's length with the enlarged union length; C `[17,29)` therefore cannot replace A even though C is the longest original span. Please track the winning source-span length separately and add a chained-overlap regression.

adaamko:
@Xunzhuo all four are in, head 7fa299b3.

truncation: PIIModel embeds the shared on_error contract; partial spans count, on_error: block turns unclassified content into classification_error. TestPIISignalTruncatedResponseRoutesThroughOnError.
wiring: pii.backend goes through the canonical normalizer, the global validators and IsPIIClassifierEnabled, same as domain.backend. TestCanonicalPIIBackend*, TestPIIOnErrorValidatedAtParse, TestIsPIIClassifierEnabledAcceptsRemoteBackend.
dormant backend: no backend is built when the mapping is nil (no reachable PII decision). TestBuildClassifierSkipsDormantRemotePIIBackend and the reachable counterpart.
chained overlaps: the merged range tracks the length of the detection whose placeholder it carries. TestBuildMaskedPIITextChainedOverlapKeepsLongestSourcePlaceholder, both report orders.

one deliberate choice: model_ref still resolves for a remote pii backend so the mapping file is provisioned; parameters.labels is the reserved way to remove that later.

Xunzhuo (changes requested):
Thanks for the thorough contract work. One local and remote parity blocker remains: knownPIILabels admits every mapped label, including the class-zero outside label. A remote O span therefore becomes a real TokenEntity and reaches PIIEntities and masking, while the local backend explicitly skips class zero. Please reject the configured outside or no-entity label and add a fixture that drives it through the signal and masking paths.

adaamko:
@Xunzhuo fixed in c9a7b85f: class zero of the mapping and the literal O are excluded from the entity label set, and a span carrying either rejects the whole response. fixture outside_label_rejected, and TestRemoteOutsideSpanNeverBecomesAnEntity drives an O span through ClassifyPIIWithDetails and the PII signal under both on_error policies.

src/semantic-router/pkg/classification/http_token_classifier.go
```diff
@@ -0,0 +1,398 @@
...
+		Confidence: score,
+	}, nil
+}
+
+// spanLabel resolves label / entity_group, strips any BIO prefix and checks the
+// result against the configured mapping. The outside label is rejected before
+// the known-label check: it is in the mapping, but it names the absence of an
+// entity, and the native backend never emits it as a span.
+func spanLabel(i int, sp tokenSpanWire, known, outside map[string]struct{}) (string, error) {
+	label := sp.Label
```
Xunzhuo, line 314: **Reject conflicting wire aliases**

Impact: A 200 response containing contradictory `label`/`entity_group` or `text`/`word` values is silently accepted. That makes malformed model output look valid and prevents `on_error: block` from failing closed, so downstream PII decisions can use an ambiguous entity.

Required change: Reject envelopes when both aliases are present and differ (after any documented normalization), and cover both conflict pairs under block and continue policies.

src/semantic-router/pkg/config/model_config_types.go
```diff
@@ -49,6 +49,15 @@ type PIIModel struct {
 	UseCPU         bool    `yaml:"use_cpu"`
 	UseMmBERT32K   bool    `yaml:"use_mmbert_32k"`
 	PIIMappingPath string  `yaml:"pii_mapping_path"`
+	// Backend attaches a named remote token classifier speaking token_spans.v1.
+	// Its absence preserves local PII inference exactly as before.
+	Backend *RemoteClassifierBackend `yaml:"backend,omitempty"`
```
Xunzhuo, line 54: **Expose the PII backend through the Operator API**

Impact: `backend` and `on_error` exist in the core config, but the Operator's `PIIModelConfig` and generated CRD schema cannot carry them. Kubernetes users therefore cannot reliably select the remote backend or its failure policy because the fields are pruned or rejected before conversion.

Required change: Add both fields to the Operator type and CRD schema, regenerate the API artifacts, and extend conversion/canonical-config tests to prove the values reach the runtime config.

src/semantic-router/pkg/classification/http_token_classifier_test.go
```diff
@@ -0,0 +1,395 @@
...
+	cfg := &config.ExternalModelConfig{
+		Name:          "pii-svc",
+		ModelRole:     config.ModelRoleClassification,
+		ModelEndpoint: config.ClassifierVLLMEndpoint{Address: u.Hostname(), Port: port, Protocol: "http"},
+		ModelName:     "pii-spans",
+	}
+	return server, cfg
+}
+
+func TestHTTPTokenClassifierFixtures(t *testing.T) {
```
Xunzhuo, line 120: **Complete the accepted local/remote parity gate**

Impact: #2922 requires exact local-versus-remote parity for `PIIDetected`, `PIIEntities`, `MatchedPIIRules`, and masked text across representative fixtures. This loop exercises only the remote adapter, so the PR can close the issue while that required contract remains unverified.

Required change: Run the same fixture corpus through both backends and assert equality across all four outputs, including overlap and malformed-span cases, before closing #2922.

src/semantic-router/pkg/classification/http_token_classifier.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+// tokenSpansEnvelope is the object form. Spans and Error stay raw so that a
+// missing member, a null and a non-array are distinguishable from an empty
+// list; the decoder decides what each means.
+type tokenSpansEnvelope struct {
+	Spans       json.RawMessage `json:"spans"`
+	TruncatedAt *int            `json:"truncated_at"`
+	Error       json.RawMessage `json:"error"`
+	Model       string          `json:"model"`
```
Xunzhuo, line 112: **Define the model-identity contract**

Impact: The wire envelope accepts a `model` value, but the adapter discards it without validation or propagation. That leaves the accepted model-identity requirement undefined and permits a response from the wrong model to be treated as valid.

Required change: Specify whether model identity is required and how it must match the configured model, enforce that rule in parsing, and document/test the public remote-backend contract.

Xunzhuo (commented):
Thanks for working through the earlier parity issues. I found four remaining contract and deployment gaps that need to be closed before this can satisfy #2922.

src/semantic-router/pkg/classification/http_token_classifier.go
adaamko: fixed in 0eb8949f: a span with both label and entity_group, or both text and word, that disagree rejects the whole response and routes through on_error. TestHTTPTokenClassifierRejectsConflictingAliases and TestConflictingAliasesRouteThroughOnError cover both pairs under allow and block.

src/semantic-router/pkg/classification/http_token_classifier.go
adaamko: fixed in 0eb8949f: the envelope model is optional; when present it must equal the configured llm_model_name or the response is rejected. the bare-list form carries no model. TestHTTPTokenClassifierModelIdentity covers matching, omitted, bare-list and mismatched, plus fail-closed under block.

src/semantic-router/pkg/classification/http_token_classifier_test.go
adaamko: added in 0eb8949f: TestLocalAndRemotePIIBackendsAgreeOnFixtures runs every accepted fixture (multi-byte, overlap, nested) through a native-shaped backend and the remote adapter and asserts PIIDetected, PIIEntities, MatchedPIIRules and the full detection list equal. masked text is a pure function of that list; TestBuildMaskedPIITextLocalAndRemoteOffsetsAgree covers the byte vs code-point derivations on the services side. malformed spans have no local counterpart: the remote path rejects them and on_error decides.

src/semantic-router/pkg/config/model_config_types.go
adaamko: done in 38b3acb6: PIIModelConfig carries the shared RemoteClassifierBackendConfig block from #3542 plus on_error (allow|block), the contract enum admits token_spans.v1, both CRD copies regenerated. the conversion test carries a remote PII backend and on_error onto the canonical config, and TestGeneratedCRDsCarryPIIBackendAndOnError pins both artifacts. domain still has no backend in the CRD; thats a separate follow-up under #2782.

deploy/operator/api/v1alpha1/semanticrouter_types.go
```diff
@@ -1451,6 +1452,17 @@ type PIIModelConfig struct {
 	UseCPU bool `json:"use_cpu,omitempty"`
 	// +optional
 	PIIMappingPath string `json:"pii_mapping_path,omitempty"`
+	// Backend names a remote token classifier speaking token_spans.v1. Its
+	// absence keeps local PII inference. The router refuses a backend combined
+	// with use_mmbert_32k at load, the same rule domain.backend follows.
+	// +optional
+	Backend *RemoteClassifierBackendConfig `json:"backend,omitempty"`
```
Xunzhuo, line 1517: **Wire the backend model into the Operator catalog**

Impact: The CRD can now set `backend.model`, but the Operator-generated canonical config never adds a matching `global.model_catalog.external[]` entry. PII backend validation resolves only against that list, so an Operator config using this field fails with `backend.model ... is not declared` and the remote PII backend cannot start.

Required change: Expose or derive the external classification model identity and endpoint in the Operator path, populate `ModelCatalog.External`, and add a CR-to-generated-YAML-to-router-parse regression that proves the backend model and PII mapping resolve.

Xunzhuo (commented):
Thanks for completing the earlier fixes—one Operator integration gap still blocks the remote PII path.

deploy/operator/api/v1alpha1/semanticrouter_types.go
adaamko: done in fc5bf3d5, external_models on ConfigSpec plus the CR to yaml to router-parse test, please review

Xunzhuo (commented):
Thanks for working through the earlier fixes. Three runtime and admission blockers remain.

src/semantic-router/pkg/classification/http_token_classifier.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+	if trimmed[0] != '{' {
+		return nil, nil, "", fmt.Errorf("token_spans response must be a JSON array or object")
+	}
+	var env tokenSpansEnvelope
+	if err := json.Unmarshal(trimmed, &env); err != nil {
+		return nil, nil, "", fmt.Errorf("failed to parse token_spans response: %w", err)
+	}
+	if isPresentJSON(env.Error) {
+		return nil, nil, "", fmt.Errorf("token_spans provider reported an error: %s", bytes.TrimSpace(env.Error))
```
Xunzhuo: **Keep provider and prompt content out of returned errors**

Impact: This interpolates an arbitrary provider error payload, and the same decoder includes provider-returned span text and slices of the original input in validation errors. Those errors reach operational logs, so an endpoint that echoes the classified prompt can disclose user PII; the existing `http_classify` path deliberately omits response bodies for this reason.

Required change: Return only bounded structural metadata from every `token_spans` error path, never provider or input text, and add regressions proving echoed secrets and mismatched span text do not appear in returned errors or logs.

src/semantic-router/pkg/classification/classifier_lifecycle.go
```diff
@@ -166,6 +172,7 @@ func (c *Classifier) Close() error {
 	closeResource("jailbreak classifier", c.jailbreakInference)
 	closeResource("complexity score backend", c.complexityScoreBackend)
 	closeResource("complexity label backend", c.complexityLabelBackend)
+	closeResource("PII classifier", c.piiInference)
```
Xunzhuo, line 175: **Close the remote PII backend through its admission wrapper**

Impact: `applyAdmissionGates` always replaces this backend with `admittedPIIInference`, which does not implement `Close`. This type assertion therefore skips `piiHTTPBackend.Close`, leaking its connector and idle transports on every classifier reload.

Required change: Make the admission wrapper delegate `Close` to its underlying backend, as the other closable wrappers do, and cover closing a wrapped remote PII backend during reload.

deploy/operator/api/v1alpha1/semanticrouter_types.go
```diff
@@ -1225,8 +1273,9 @@ type RemoteClassifierBackendConfig struct {
 	// Contract is the response shape the signal reads. Complexity reads two -
 	// score.v1, one regression number interpreted through each rule's
 	// boundaries, and label_distribution.v1, hard/easy/medium probabilities -
-	// so the router requires it there rather than guessing per request.
-	// +kubebuilder:validation:Enum=score.v1;label_distribution.v1
+	// so the router requires it there rather than guessing per request. PII
+	// reads token_spans.v1, entity spans with code-point offsets.
+	// +kubebuilder:validation:Enum=score.v1;label_distribution.v1;token_spans.v1
```
Xunzhuo, line 1279: **Constrain contracts by the consuming signal**

Impact: Widening this shared enum admits `complexity_model.backend.contract: token_spans.v1`, while the complexity CEL rule checks only that a contract exists; the PII consumer likewise admits `score.v1` and `label_distribution.v1`. Kubernetes accepts these CRs, but the generated router config rejects the consumer/contract mismatch.

Required change: Add consumer-specific CEL constraints or separate backend schema types, regenerate the CRD, and add admission tests rejecting both invalid complexity and PII combinations.

src/semantic-router/pkg/classification/http_token_classifier.go
adaamko: done in 2a0d2e88, every token_spans error carries sizes and offsets only, regression plants a marker in each echo point

src/semantic-router/pkg/classification/classifier_lifecycle.go
adaamko: done in 2a0d2e88, admittedPIIInference forwards Close, covered on the reload path

deploy/operator/api/v1alpha1/semanticrouter_types.go
adaamko: done in 29f4c3d2, per-consumer CEL rules on complexity and pii, admission tests run the apiserver's own evaluator against the generated CRD

adaamko:
@Xunzhuo all three from the last round are in (2a0d2e88, 29f4c3d2), please take another look

src/semantic-router/pkg/classification/classifier_signal_pii.go
```diff
@@ -95,13 +106,29 @@ func (c *Classifier) evaluatePIIRule(rule config.PIIRule, piiText string, nonUse
...
-	entityTypes := c.collectPIIEntityTypes(ruleContents, rule.Name, rule.Threshold, piiCache)
+	entityTypes, failed := c.collectPIIEntityTypes(ruleContents, rule.Name, rule.Threshold, piiCache)
 	deniedEntities := findDeniedEntities(entityTypes, rule.PIITypesAllowed)
+	errorDrivenMatch := false
+	if failed && c.Config.PIIModel.IsBlock() {
+		// Part of the content was never scored (backend error or a declared
+		// truncation). Under on_error: block that is not a clean result.
+		logging.Errorf("[Signal Computation] PII rule %q: content not fully classified; failing closed", rule.Name)
+		deniedEntities = append(deniedEntities, PIIClassificationErrorType)
+		errorDrivenMatch = true
```
Xunzhuo: **Do not mark a genuine PII detection as error-driven**

Impact: This flag is set for every failed `on_error: block` evaluation, even when valid spans already produced a real denied entity. `decision.evalLeaf` then converts that genuine match into unknown, so `rules.on_unknown: no_match` can discard detected PII and route the request as a non-match after a truncated or partially failed scan.

Required change: Set `SignalErrorMatches` only when the failure itself created the match, while still recording incompleteness and the sentinel entity. Add a routing-decision regression with a detected PII span plus truncation under `on_error: block` and `on_unknown: no_match`.

Xunzhuo (commented):
Thanks for closing the earlier remote-PII gaps. One fail-closed routing edge case remains on this head.

src/semantic-router/pkg/classification/classifier_signal_pii.go
adaamko: Done, thanks

adaamko:
Please @Xunzhuo , check again

src/semantic-router/pkg/classification/mapping.go
```diff
@@ -115,9 +115,39 @@ func LoadPIIMapping(path string) (*PIIMapping, error) {
 		return nil, fmt.Errorf("failed to parse PII mapping JSON: %w", err)
 	}
 
+	// The same guard the jailbreak mapping carries: a configured label equal
+	// to the on_error: block sentinel (PIIClassificationErrorType) would make a
+	// genuine detection of that label indistinguishable from a classify
+	// failure, and the PII signal now decides error-driven matches on exactly
+	// that distinction. A remote token_spans.v1 backend can only return labels
+	// the mapping declares, so rejecting it here closes the wire path too.
+	if mapping.hasLabel(PIIClassificationErrorType) {
```
Xunzhuo: **Normalize BIO prefixes before reserving this sentinel**

Impact: This guard compares raw labels, but both native translation and the remote token-span label set strip `B-`, `I-`, and `E-` prefixes. A mapping containing `B-classification_error` therefore passes loading and later becomes the reserved sentinel, making a genuine PII detection indistinguishable from an `on_error` failure and corrupting the error-driven decision and Replay signal.

Required change: Reject any configured label whose BIO-normalized value equals `PIIClassificationErrorType` in both mapping directions, and add prefixed-label regressions.

Xunzhuo (commented):
Thanks for fixing the partial-scan classification. One sentinel alias still bypasses the new guard.

src/semantic-router/pkg/classification/mapping.go
adaamko: Fixed in d3ec8f43. LoadPIIMapping rejects BIO-prefixed sentinel aliases in both mapping directions, including stacked prefixes. Regressions cover each prefix and ordinary entity labels.

Xunzhuo (approved):
Awesome work @adaamko Thanks!

PR #3499, opened 2026-09-04
[Bug] Add an Azure backend dialect for vendor response fields

### LINKED ISSUE #3496: [Bug] Canonical response validation rejects every Azure OpenAI / AI Foundry response
### Proposed Workgroup

MoM & Routing

### Describe the bug

Since [#3076 "Unify inference protocol translation"](https://github.com/vllm-project/semantic-router/pull/3076) the router validates the **upstream provider response** against a canonical OpenAI schema and rejects any field outside it. Azure OpenAI / Azure AI Foundry always decorate their responses with Azure-specific fields, so **every** completion fails — the router is unusable against Azure as a backend.

`decodeProviderJSON` in `src/semantic-router/pkg/protocolcodec/json_helpers.go` calls `validateExactJSONFieldNames` and then `decoder.DisallowUnknownFields()`:

```go
if rejectUnknownFields(body, policy) {
    if err := validateExactJSONFieldNames(body, reflect.TypeOf(target)); err != nil {
        return llmprotocol.NewError(llmprotocol.ErrorUpstreamUnavailable, "invalid_upstream_json",
            "upstream response JSON contains a non-canonical field", err)
    }
    decoder.DisallowUnknownFields()
}
```

`rejectUnknownFields` is true whenever `policy.UnknownFields == UnknownReject`, and that is the shipped default — `llmprotocol.DefaultPolicy()` hardcodes it:

```go
func DefaultPolicy() Policy {
    return Policy{
        UnknownFields: UnknownReject, LossyFeatures: LossyReject,
        ...
```

**There is no way to configure around it.** No YAML key maps to `Policy.UnknownFields`, and `DefaultPolicy()` is constructed directly at every call site, including the ext_proc engine (`pkg/extproc/processor_protocol_contract.go:95`). The provider dialects (`official_openai`, `official_deepseek`, `openrouter`, `vllm`, `generic_openai_compatible`) only shape the *request* — none affects response validation, and none of them is Azure.

### To Reproduce

1. Configure a backend pointing at Azure AI Foundry's OpenAI-compatible v1 endpoint (`https://<resource>.services.ai.azure.com/openai/v1`, `provider: openai`).
2. Run any image from `nightly-20260831` onward (we hit it on `nightly-20260904`).
3. Send a chat completion.

**Non-streaming** returns HTTP 502:

```json
{"error":{"code":502,"message":"The selected model returned an invalid response","type":"api_error"}}
```

**Streaming** returns an SSE error frame:

```json
data: {"error":{"type":"server_error","code":"invalid_upstream_json",
       "message":"upstream response JSON contains a non-canonical field","param":null}}
```

Router log:

```
neutral_response_decode_failed  error="invalid_upstream_json: upstream response JSON contains a
non-canonical field"  backend_format=openai.chat.v1  client_format=openai.chat.v1
  at pkg/extproc.(*OpenAIRouter).handleNonStreamingResponseBody (processor_res_body_pipeline.go:25)
```

The offending fields are Azure's standard additions, present on every response:

| Location | Fields |
| --- | --- |
| top level | `prompt_filter_results`, `routing` |
| `choices[]` | `content_filter_results` |
| `usage` | `latency_checkpoint` |

### Expected behavior

A provider response carrying vendor extension fields should be accepted; unknown fields should be ignored (or preserved) rather than failing the request. Azure OpenAI is a first-class OpenAI-compatible backend and has emitted content_filter_results for years.

### Impact

_No response_

### Additional context

_No response_

nickaggarwal:
Closes #3496

## Purpose

Since #3076 the router validates upstream provider responses against the canonical OpenAI schema and rejects any field outside it. Azure OpenAI / Azure AI Foundry decorate **every** response, so every completion failed with `invalid_upstream_json` and the router was unusable against Azure. `DefaultPolicy()` hardcodes `UnknownReject` and no YAML key maps to it, so there was no workaround.

Strict rejection stays the default. `engine.go` re-asserts `UnknownReject` whenever a translation crosses formats or mutates, and several test files enforce it — including `codec_openai_chat_execution_metadata_test.go`, for metadata the router must not silently drop. This PR does not relax that; it gives an identified Azure backend a scoped allowance and makes every drop observable.

### Azure backend identification

`resolveResponseVendor` keys first on the catalog provider id `azure-openai` — the canonical Azure configuration — and falls back to Azure host suffixes (`*.openai.azure.com`, `*.services.ai.azure.com`, `*.cognitiveservices.azure.com`) for a profile declared as `type: openai` but pointed at Azure, which is the setup in the issue report. Per-tenant subdomains are why that fallback matches by suffix rather than exact host.

It is deliberately independent of reasoning transport: response decoration is a property of who serves the response, not of how a request is shaped. A test asserts an Azure profile acquires no reasoning transport it did not declare.

### Carrying the allowance to the decoder

`llmprotocol.Policy` gains `ResponseVendor` (empty = strict). It is resolved at provider dispatch, carried on `RequestContext`, and applied only by the paths that decode a live provider response. Cache reads and client-facing encodes keep the strict default.

### Tolerance is per backend, not per field name

An identified Azure backend may carry **any** field outside the canonical schema. The dropped set is computed from the target schema, not a name list, because Azure ships new decorations across API versions and a fixed list would reproduce this bug on the next one.

Nothing is dropped silently, which was the objection to blanket tolerance: every dropped path raises a `DiagnosticDropped` entry, bounded by `Limits.Diagnostics`, across buffered Chat, streamed Chat, buffered Responses, streamed Responses, and transport errors. Azure decorates error envelopes too (content-filter blocks, rate limits), so decoding those strictly would have replaced the real upstream reason with a generic router error.

### Cache write/read contract

An accepted same-format response could poison its own cache entry. Two defects combined:

1. Source preservation replays preserved bytes verbatim on a same-format encode, so the envelope built from the **upstream** body re-emitted the decorations the decode had just dropped. The envelope is now built from the canonical bytes the decode produced.
2. The buffered pipeline cached the verbatim client body. The cache write now re-encodes when decorations were dropped, and skips the write if that encode fails — a miss is recoverable, a poisoned entry is not.

Cache reads keep the strict contract with no vendor allowance, because a partition is keyed on the ingress protocol and may be served to a request that never touches the same backend. `TestCacheableClientResponseIsReadableByTheCacheReader` covers miss → write → hit; `TestCacheReaderRejectsRawDecoratedUpstreamBody` pins the regression itself.

### Error diagnosis without leaking

`invalid_upstream_json` did not say which field failed. The name cannot go in the error message: `ProtocolError.Message` is serialized into the client-facing error body (`openai_transport_error.go`), so that would disclose the upstream response shape and the backend provider to any caller. The name stays in the cause, and `neutral_response_decode_failed` unwraps it into the router log.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/protocolcodec/ ./pkg/llmprotocol/
go vet ./pkg/extproc/ ./pkg/protocolcodec/ ./pkg/llmprotocol/
```

Coverage added across `provider_vendor_extensions_test.go`, `provider_vendor_extensions_responses_test.go`, `processor_res_cache_vendor_test.go`, and `req_filter_response_vendor_test.go`: vendor identification (canonical type, three host suffixes, and lookalike hosts that must not match); tolerance of unanticipated fields; strict rejection without an allowance; dropped-field diagnostics on all five decode paths including nested paths; the cache miss-to-hit path; and the offending field staying out of the client message.

## Test Result

```
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/protocolcodec  1.061s
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/llmprotocol    0.304s
```

Full existing `protocolcodec` and `llmprotocol` suites pass; no existing rejection or byte-fidelity assertion regressed. `go vet` clean on `pkg/extproc`, `pkg/protocolcodec`, `pkg/llmprotocol`. `gofmt` clean.

Blockers on local verification, for transparency:

- `pkg/extproc` tests compile and typecheck but **cannot be executed** on this machine: the package links the Rust bindings and there is no Rust toolchain here. The cache miss-to-hit tests are therefore first exercised by CI.
- `golangci-lint` could not run locally at all — both the repo-pinned 2.13.2 and an older build panic or report typecheck errors inside Go's own stdlib, because the active toolchain here is newer than the linter build. CI is the first real lint signal.
- No live Azure endpoint. Every Azure body in the tests is taken from the issue report or constructed; the error-envelope shape in particular is inferred, not captured from Azure.

Not run locally: E2E profiles.

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_protocol_contract.go
+++ b/src/semantic-router/pkg/extproc/processor_protocol_contract.go
@@ -95,6 +95,25 @@ func (r *OpenAIRouter) protocolEngine() (*protocolcodec.Engine, error) {
 	return protocolcodec.NewEngine(registry, llmprotocol.DefaultPolicy())
 }
 
+// protocolEngineForBackend builds an engine whose policy carries the selected
+// backend's vendor allowance. Only paths that decode a live provider response
+// use it; cache replay and client-facing encodes keep the strict default so a
+// backend allowance can never widen a contract it does not own.
+func (r *OpenAIRouter) protocolEngineForBackend(ctx *RequestContext) (*protocolcodec.Engine, error) {
+	if r == nil {
+		return nil, fmt.Errorf("protocol runtime is unavailable")
+	}
+	registry := r.ProtocolCodecs
+	if registry == nil {
+		registry = protocolcodec.NewBuiltinRegistry()
+	}
+	policy := llmprotocol.DefaultPolicy()
+	if ctx != nil {
+		policy.ResponseVendor = ctx.ResponseVendor
+	}
+	return protocolcodec.NewEngine(registry, policy)
+}
+
 // prepareProtocolRequest decodes every public wire format exactly once. The
 // neutral request is the sole mutable request contract after this boundary.
 func (r *OpenAIRouter) prepareProtocolRequest(
@@ -220,7 +239,7 @@ func (r *OpenAIRouter) decodeClientResponse(
 	if ctx == nil {
 		return nil, fmt.Errorf("request context is unavailable")
 	}
-	engine, err := r.protocolEngine()
+	engine, err := r.protocolEngineForBackend(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/src/semantic-router/pkg/extproc/processor_protocol_contract.go
+++ b/src/semantic-router/pkg/extproc/processor_protocol_contract.go
@@ -95,10 +95,7 @@ func (r *OpenAIRouter) protocolEngine() (*protocolcodec.Engine, error) {
 	return protocolcodec.NewEngine(registry, llmprotocol.DefaultPolicy())
 }
 
-// protocolEngineForBackend builds an engine whose policy carries the selected
-// backend's vendor allowance. Only paths that decode a live provider response
-// use it; cache replay and client-facing encodes keep the strict default so a
-// backend allowance can never widen a contract it does not own.
+// protocolEngineForBackend permits extensions only for live provider responses.
 func (r *OpenAIRouter) protocolEngineForBackend(ctx *RequestContext) (*protocolcodec.Engine, error) {
 	if r == nil {
 		return nil, fmt.Errorf("protocol runtime is unavailable")
--- a/src/semantic-router/pkg/protocolcodec/provider_vendor_extensions_responses_test.go
+++ b/src/semantic-router/pkg/protocolcodec/provider_vendor_extensions_responses_test.go
@@ -2,6 +2,7 @@ package protocolcodec
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/llmprotocol"
@@ -19,7 +20,7 @@ func TestAzureResponsesResponseReportsDroppedVendorExtensions(t *testing.T) {
 	if response.ID != "resp_1" {
 		t.Fatalf("response ID = %q, want resp_1", response.ID)
 	}
-	assertDroppedResponseExtension(t, diagnostics, "azure_trace")
+	assertDroppedResponseExtensions(t, diagnostics, "azure_trace")
 }
 
 func TestAzureResponsesStreamReportsDroppedVendorExtensions(t *testing.T) {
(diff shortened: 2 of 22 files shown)


Xunzhuo (changes requested):
Thanks for scoping this at the backend boundary. Two Azure paths remain incomplete: the dialect resolver exits before host detection for the repository-supported `type: azure-openai`, so that canonical configuration never receives the allowance; and transport-error decoding accepts decorations through `decodeProviderWire` but discards their paths and returns no diagnostics before re-encoding, making those drops silent. Please cover `azure-openai` profiles and return dropped-field diagnostics from the transport-error codec as the success and streaming paths do.

Xunzhuo (changes requested):
Thanks—the canonical `azure-openai` profile and transport-error diagnostics are fixed, but Azure Responses still accepts and silently drops decorations because its buffered and streamed decoders discard the removed-field list. Please propagate diagnostics there or scope the allowance to Chat, then mark the PR ready and sync current `main`.

src/semantic-router/pkg/extproc/processor_protocol_contract.go
```diff
@@ -220,7 +236,7 @@ func (r *OpenAIRouter) decodeClientResponse(
 	if ctx == nil {
 		return nil, fmt.Errorf("request context is unavailable")
 	}
-	engine, err := r.protocolEngine()
+	engine, err := r.protocolEngineForBackend(ctx)
```
Xunzhuo, line 240: **[P1] Keep newly accepted Azure responses replayable from cache**

For a same-format Chat request with exact or semantic caching, this now accepts Azure decorations, but `handleNonStreamingResponseBody` still keeps the original body because `requiresClientResponseRewrite` is false. `updateResponseCache` stores those bytes unchanged. The next hit uses `decodeCachedClientResponse` with the strict default policy, rejects the decoration, and returns 502 ("The cached response is invalid"). The Azure fixture succeeds under the live codec policy and fails under that cache policy. Canonicalize the cached client body, or otherwise align cache write/read contracts, and add an Azure miss-to-hit regression.

Xunzhuo (changes requested):
The Azure dialect and diagnostics fixes are in place, but accepted same-format responses can poison their cache entries and make subsequent requests fail. Please align the cache write/read contract and cover the miss-to-hit path.

src/semantic-router/pkg/extproc/req_filter_response_vendor_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{name: "cognitive services host", providerType: "openai", baseURL: "https://my-resource.cognitiveservices.azure.com/openai/v1"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			profile := &config.ProviderProfile{Type: tt.providerType, BaseURL: tt.baseURL}
+
+			assert.Equal(t, llmprotocol.ResponseVendorAzure, resolveResponseVendor(profile))
+			// Response decoration is independent of request shaping: an Azure
+			// profile must not acquire a reasoning transport it did not declare.
+			assert.Equal(t, modelcatalog.ReasoningTransportChatTemplate, resolveProviderReasoningTransport(profile))
```
Xunzhuo: **[P2] Keep the declared reasoning default in the hostname cases**

The three hostname fixtures use `Type: "openai"`, whose catalog entry declares `reasoning_transport: top_level_effort`. Response-vendor detection correctly leaves that transport unchanged, so expecting `chat_template_kwargs` here fails all three subtests in the [current-head Core run](https://github.com/vllm-project/semantic-router/actions/runs/35285005294/job/105418029772). Assert the provider-specific expected transport, or use an `openai-compatible` fixture when testing the template default.

src/semantic-router/pkg/protocolcodec/provider_vendor_extensions_responses_test.go
```diff
@@ -0,0 +1,159 @@
...
+		llmprotocol.OpenAIResponsesV1,
+		llmprotocol.OpenAIResponsesV1,
+		llmprotocol.StreamContext{Context: context.Background(), PublicModel: "public-model"},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	start := []byte("event: response.created\n" +
+		`data: {"type":"response.created","sequence_number":0,"response":{"id":"resp_1","object":"response","created_at":1,"model":"provider-model","status":"in_progress","output":[]}}` + "\n\n")
+	if _, _, _, err := stream.Push(start); err != nil {
```
Xunzhuo: **[P2] Avoid shadowing the stream setup error in both tests**

This `err :=` shadows the error declared by `engine.NewStream` at line 75; the same pattern occurs at line 129 after line 115. The [current-head Quality run](https://github.com/vllm-project/semantic-router/actions/runs/35285005294/job/105415349603) reports both as `govet` shadow errors. Reuse the existing error variable or use a distinct name for the `Push` error in both tests so the required quality check can pass.

Xunzhuo (changes requested):
The Chat cache miss-to-hit regression now passes. Two added tests still fail the required Core and Quality checks: the hostname fixtures expect the wrong catalog reasoning default, and the Responses stream tests shadow `err`. Please correct these tests and rerun both checks. The protocolcodec and llmprotocol suites pass locally; live Azure inference was not run.

src/semantic-router/pkg/extproc/processor_protocol_contract.go
nickaggarwal: Canonicalized the cached client body. This was two defects: the envelope was built from the upstream bytes, so source preservation replayed the decorations verbatim on a same-format encode and the "canonical" re-encode still carried them; and the buffered pipeline cached the verbatim client body. The envelope is now built from the canonical bytes the decode produced, and `cacheableClientResponse` re-encodes when decorations were dropped, skipping the write entirely if that encode fails. Cache reads keep the strict policy. Added `TestCacheableClientResponseIsReadableByTheCacheReader` (miss -> write -> hit) and `TestCacheReaderRejectsRawDecoratedUpstreamBody` to pin the regression.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/semantic-router/pkg/extproc/req_filter_response_vendor_test.go
nickaggarwal: Fixed — you were right that `type: openai` declares `top_level_effort`, so expecting `chat_template_kwargs` was wrong for all three hostname cases. Rather than hardcode the provider-specific transport, the test now captures the transport before resolving the vendor and asserts it is unchanged after, since the property being tested is that response-vendor detection leaves request shaping alone. That also keeps the fixture off catalog data that varies by provider type.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/semantic-router/pkg/protocolcodec/provider_vendor_extensions_responses_test.go
nickaggarwal: Fixed — the `Push(start)` error is now bound to a distinct `startErr` in both tests, so the `engine.NewStream` error is no longer shadowed at either site.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Xunzhuo (approved):
The reasoning-transport assertion and shadowed error variables are fixed. Current-head Core evidence confirms Azure vendor selection, nested response/stream diagnostics and strict cache-read regressions, and Quality plus the relevant streaming/Responses API checks pass. The production codec changes match the previously reviewed patch. No live Azure request was run.

wilsonwu (approved):
LGTM

Issue #3181, opened 2026-08-31
[Test] Cover Router Learning adaptation, protection, and state lifecycle end to end


Xunzhuo:
### Goal

Establish end-to-end coverage for the current Router Learning contract across adaptation, continuity protection, decision-local controls, and state lifecycle.

Parent Epic: #1519

### Proposed Workgroup

Evaluation & Quality

### Scope

- Cover global learning enablement and the current `routing_sampling` adaptation path.
- Cover decision, tier, and global candidate-set boundaries and verify that ineligible models cannot enter the learned selection set.
- Cover decision-local `apply`, `observe`, and `bypass` behavior for adaptation and protection.
- Cover conversation and session protection, configured identity headers, minimum-turn and switch-margin behavior, idle expiry, and continuity diagnostics.
- Cover in-memory and Redis-backed state, restart recovery, multi-replica consistency, TTL, timeout, and defined fail-open/fail-safe behavior.
- Assert selected-model outcomes together with Router Replay or debug learning evidence, not only successful requests.
- Audit and retire or rewrite the unprofiled legacy `rl-driven-*` cases that target older feedback and Thompson-style behavior rather than the current `global.router.learning` API.
- File implementation or API defects under MoM & Routing, Agentic & Context, Data Plane & Networking, or the actual primary owner, then link the regression here.

### Non-goals

- Prove long-term online learning quality from a short E2E run.
- Train reward or routing models.
- Depend on proprietary providers or nondeterministic model quality.
- Treat all state-store permutations as default PR gates.

### Completion signal

- Adaptation, candidate-set boundaries, decision modes, continuity protection, identity, and maintained state backends have deterministic E2E contracts.
- Restart, multi-replica, expiry, and backend-failure semantics are covered where applicable.
- Legacy RL-driven registrations are removed or aligned with the current Router Learning contract and reachable from a maintained profile.
- Discovered implementation and API gaps have focused owning issues and regression coverage.

FAUST-BENCHOU:
The unprofiled `rl-driven-*` retirement is split into #3295. That child does not close this issue.

FAUST-BENCHOU:
Thanks for your interest in this issue! This is still under discussion and has not been accepted by the owning Workgroup yet. If you'd like to work on it, please first share a short implementation proposal here describing your approach, expected change surfaces, and validation plan, and wait for acceptance before starting implementation.

Issue #2859, opened 2026-08-11
[Test] Expand Looper E2E coverage beyond confidence headers


Xunzhuo:
## Goal

Expand end-to-end coverage so maintained Looper algorithms are protected across ExtProc, protocol shaping, streaming, cancellation, Replay, and aggregate accounting—not only confidence headers.

## Proposed Workgroup

Evaluation & Quality

## Background

Unit tests cover many algorithm internals, but current E2E coverage does not prove that Fusion, ReMoM, Workflows, Best-of-N when available, and their failure paths preserve the complete client-visible contract. Behavior-visible routing changes need bounded protocol-level regression coverage.

Parent Epic: #1519. Related implementation tracker: #2336. Related trace and panel contracts: #2855 and #2856. Native final-stage streaming: #3378.

## Scope and approach

- Cover Fusion usable quorum, partial/all-failed panels, analysis, final synthesis, and model attribution.
- Cover ReMoM multi-round execution, final synthesis, and aggregate usage.
- Cover Workflows planning/static execution and one tool pause/resume round trip.
- Cover Best-of-N returning the selected original response when that algorithm lands.
- Exercise output normalization, typed fallback, timeout, cancellation, downstream disconnect, empty/malformed responses, and final-stage-only failure.
- For every supported non-streaming and streaming protocol, verify identifiers, native deltas where supported, first-chunk latency, usage, finish reasons, tool-call chunks, and error mapping.
- Confirm Replay and debug facts match selected/final role, stage tree, iterations, failures, and reconciled usage.

Synthetic re-chunking of a buffered response is not evidence of native upstream streaming.

## Non-goals

- Turn E2E into a quality benchmark or duplicate every unit permutation.
- Depend on private infrastructure, credentials, or proprietary model endpoints.

## Completion criteria

- [ ] Each maintained Looper family has a successful public-API journey.
- [ ] Partial, below-quorum, all-failed, verifier/final failure, and cancellation cases are covered.
- [ ] ReMoM proves multiple rounds and correct aggregate accounting.
- [ ] Workflows proves tool pause/resume across requests.
- [ ] Best-of-N selection and Fusion synthesis remain distinct in protocol behavior.
- [ ] Cases are registered in affected E2E profiles and remain bounded in CI.

## Test boundary

[...]

Issue #2379, opened 2026-07-05
[Test] Establish a catalog-driven E2E coverage matrix and execution gate


Xunzhuo:
### Goal

Establish a catalog-driven, machine-readable E2E coverage matrix and execution gate so every supported runtime surface and registered testcase has an explicit contract, role, environment, and owner.

### Proposed Workgroup

Evaluation & Quality — parent Epic #1519.

### Scope

- Derive the supported surface inventory from the runtime catalogs for signals, projections, routing algorithms, decision plugins, and Router Learning.
- Map each surface to its testcase, profile, supported environment, role, tier, owner, prerequisites, failure diagnostics, and affected-file selection.
- Use explicit roles such as acceptance, smoke, regression, benchmark, stress, observability, and debug; keep required, affected, and manual tiers distinct.
- Audit `e2e/testcases`, `e2e/testing`, registered profiles, `e2e/pkg/testmatrix`, and `tools/agent/e2e-profile-map.yaml` against that inventory.
- Identify registered-but-unreachable tests, stale catalog names, duplicate contracts, and weak assertions that only prove configuration load, HTTP success, or report generation.
- Keep Docker/local and Kubernetes applicability explicit without requiring every testcase to run in every environment.
- Add a deterministic gate: a new catalog surface must include a matrix mapping; an existing gap may temporarily map to a concrete child task.
- Track the concrete coverage streams:
  - #3178 — signals and projections
  - #3179 — selector algorithms
  - #2859 — Looper algorithms
  - #3180 — decision plugins
  - #3181 — Router Learning
  - #3182 — cross-surface pipelines and failure semantics
- When the audit finds a product or API defect, open a focused issue under the implementation-owning Workgroup and link its regression back into the matrix.

### Non-goals

- Implement all child coverage in this issue.
- Build a Cartesian product of every testcase, environment, and backend.
- Require a full environment-abstraction rewrite before the matrix is useful.
- Replace quality evaluation or performance benchmarking with E2E tests.
- Hide flaky behavior by weakening assertions.

### Completion signal

- [ ] A machine-readable matrix accounts for every current runtime surface and every registered E2E testcase.
- [ ] No registered testcase is unreachable without an explicit manual-only rationale.
- [ ] Every supported surface has a runnable behavior contract or a linked, bounded coverage task.
- [ ] Each testcase declares its role, tier, environment applicability, owner, prerequisites, and failure diagnostics.
[...]

PR #3501, opened 2026-09-05
[Test] Add runtime-derived E2E execution-graph audit and drift gates

gavinkvx:
## Summary

E2E testcase registration, profile selection, and CI selection can drift independently. In particular, adding a testcase to a profile does not guarantee that an explicit CI override will include it.

This PR adds a generated execution-graph audit and four consistency gates as the first slice of #2379. It uses the existing registrations and selections, with two small tables recording dispositions for known gaps.

Following review, Gate D checks explicit `*_CI_TESTS` overrides in `integration-test-k8s.yml`. Since that review, main replaced the explicit `ENVOY_AI_GATEWAY_CI_TESTS` list with the `baseline-suite` input (`E2E_BASELINE_SUITE`), so the recorded exclusions left with the override: the branch is rebased onto c16c21f3b, `KnownCIExclusions` is empty, and the `pii-remote-backend` runtime profile is registered in `tools/agent/domains.yaml`.

## What it checks

| Gate | Check |
| --- | --- |
| A | Registered testcases that no profile selects must exactly match `KnownUnreachableDebt`. |
| B | Every testcase selected by a profile must be registered. |
| C | Runtime profile names must exactly match the profile keys in `tools/agent/domains.yaml`. |
| D | Testcases omitted by an explicit `*_CI_TESTS` override must exactly match that profile's `KnownCIExclusions`, and the override must not name cases the profile does not select. Current main declares no such override, so the table is empty and the gate is armed for any future one. |

The owning issues can resolve these cases through profile mapping, removal, rename, or replacement.

## Verification

Local verification on the rebased tree (main c16c21f3b, head 0f2ec4566):

- Audit output: 33 canonical profiles, 176 registered testcases, no selected-but-unregistered names, and exactly 17 registered-but-unreachable cases.
- CI override output: no explicit override on main, no recorded exclusions, all drift sets empty.
- Two consecutive audit runs produced byte-identical JSON.
- `make e2e-coverage-check` and `make build-e2e` passed; `tools/agent/domains.yaml` passes the harness registry validation.
- Committed synthetic pins in `gates_ci_test.go` cover Gate D's omission, stale-exclusion, and phantom-entry directions against a future explicit override.

## Scope

This PR checks execution-graph consistency. Runtime behavior and surface-level coverage gaps remain with the existing coverage issues, including #2859, #3181, and #3182.

The `baseline-suite` selector (`testmatrix.BaselineCases` in the runner) is a runtime selection layer the audit does not model yet; it is a candidate follow-up slice under #2379 rather than part of this PR.

A testmatrix group referenced by no profile is outside the transitive audit; none were found at the audited snapshot. The existing Python registry checks in `tools/ci/domain_registry.py` remain in place.

`gopkg.in/yaml.v3` moves from an indirect to a direct dependency because Gate C reads the registry YAML.

Related #2379

### DIFF
--- a/e2e/pkg/verification/ciexclusions.go
+++ b/e2e/pkg/verification/ciexclusions.go
@@ -1,0 +1,59 @@
+// e2e/pkg/verification/ciexclusions.go
+
+package verification
+
+// CIExclusion bounds one selected-but-CI-omitted testcase to a disposition:
+// either an issue that owns deciding its re-inclusion (Issue) or a
+// documented reason the omission is deliberate today (Rationale). Exactly
+// one of the two must be set.
+type CIExclusion struct {
+	Issue     int
+	Rationale string
+}
+
+// KnownCIExclusions is, per profile with an explicit CI override, the exact
+// set of testcases the profile selects at runtime that the override omits
+// from CI execution on main.
+//
+// Like KnownUnreachableDebt, this table is a ratchet, not an allowlist: Gate
+// D requires the actual excluded set to equal these keys in both directions.
+// A testcase newly selected by the profile but omitted from the CI list
+// fails the gate until it is added to the list or bounded here (the
+// silent-omission scenario); an entry whose testcase joins the CI list or
+// leaves the profile selection also fails the gate until it is deleted. The
+// table can only shrink intentionally, and a profile whose override is
+// removed from the workflow must have its whole table deleted with it.
+var KnownCIExclusions = map[string]map[string]CIExclusion{
+	// The envoy-ai-gateway override currently runs 17 of the profile's 34
+	// selected testcases in CI. #2379 (the coverage epic) owns the
+	// disposition of the issue-bounded entries: re-inclusion in the
+	// explicit list, or hand-off to a coverage-family child issue.
+	"envoy-ai-gateway": {
+		"anthropic-messages-protocol-headers": {Issue: 2379},
+		"anthropic-messages-request":          {Issue: 2379},
+		"anthropic-messages-response-shape":   {Issue: 2379},
+		"anthropic-messages-streaming":        {Issue: 2379},
+		"apiserver-classification-endpoints":  {Issue: 2379},
+		"chat-completions-structured-output":  {Issue: 2379},
+		"entrypoint-recipe-routing":           {Issue: 2379},
+		"llm-classifier-distribution-routing": {Issue: 2379},
+		"looper-latency-token-headers":        {Issue: 2379},
+		"protocol-codec-openai-regression":    {Issue: 2379},
+		"retention-directive":                 {Issue: 2379},
+		"sequence-classifier-routing":         {Issue: 2379},
+		"session-pricing-chat-completions":    {Issue: 2379},
+		"session-pricing-response-api":        {Issue: 2379},
+		"session-telemetry-metrics":           {Issue: 2379},
+
+		// The workflow's own comment above the override deliberately skips
+		// stress / pressure coverage "until the suite is stable again";
+		// re-inclusion is a CI-owner decision tracked by that comment, not
+		// a coverage gap this gate should force.
+		"chat-completions-progressive-stress": {
+			Rationale: "the workflow's ENVOY_AI_GATEWAY_CI_TESTS override deliberately skips stress / pressure coverage until the suite is stable again (see the comment above the list in integration-test-k8s.yml); re-inclusion is a CI-owner decision",
+		},
+		"chat-completions-stress-request": {
+			Rationale: "the workflow's ENVOY_AI_GATEWAY_CI_TESTS override deliberately skips stress / pressure coverage until the suite is stable again (see the comment above the list in integration-test-k8s.yml); re-inclusion is a CI-owner decision",
+		},
+	},
+}
--- a/e2e/cmd/e2e-audit/main.go
+++ b/e2e/cmd/e2e-audit/main.go
@@ -1,0 +1,42 @@
+// e2e/cmd/e2e-audit/main.go
+
+// Command e2e-audit emits the runtime-derived E2E execution graph as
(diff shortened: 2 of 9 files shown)


Xunzhuo (changes requested):
Welcome, and thanks for tackling this. Gate C still does not cover the profile-versus-CI override drift required by #2379: it compares registered profile names only with `test-domain-registry.yaml`, while neither the inventory nor any gate reads the explicit `ENVOY_AI_GATEWAY_CI_TESTS` list in `integration-test-k8s.yml`. Adding a testcase to a profile but omitting that CI list therefore passes all three gates and remains unexecuted in CI. Please include that override in the derived contract and pin this omission scenario.

FAUST-BENCHOU:
> Welcome, and thanks for tackling this. Gate C still does not cover the profile-versus-CI override drift required by #2379: it compares registered profile names only with `test-domain-registry.yaml`, while neither the inventory nor any gate reads the explicit `ENVOY_AI_GATEWAY_CI_TESTS` list in `integration-test-k8s.yml`. Adding a testcase to a profile but omitting that CI list therefore passes all three gates and remains unexecuted in CI. Please include that override in the derived contract and pin this omission scenario.

+1

gavinkvx:
Hi team, thank you for catching this! Gate C checked profile registration consistency but missed the explicit CI testcase selection layer. I’ve added Gate D to address the omission scenario you pointed out:

- I have now checked back on `integration-test-k8s.yml` , and added `DiagnoseCIOverrideDrift()` in `cioverrides.go`, hope this could fix our coverage here. Gate D compares that list with the profile’s runtime GetTestCases() selection.
- A testcase added to the profile but omitted from the CI list now fails the gate unless it has an explicitly recorded exclusion. The gate also rejects stale exclusions and CI-list entries the profile does not select.
- TestCIOverrideDriftPinsOmissionScenario pins your exact scenario, with additional regression tests for stale exclusions and phantom CI-list entries.
- Existing omissions are recorded in KnownCIExclusions: 15 are tracked by #2379, and the two stress cases retain documented rationale from the workflow’s existing skip comment. The recorded exclusions must exactly match the actual omissions, so resolved entries must be removed.
- The audit JSON now includes ci_overrides, with the CI list, recorded exclusions, and drift diagnostics. These checks run through build-e2e → e2e-coverage-check.

Both make e2e-coverage-check and make build-e2e pass locally. I’ve also updated the PR description to reflect the fourth gate and its scope.

e2e/pkg/verification/ciexclusions.go
```diff
@@ -0,0 +1,59 @@
...
+	// explicit list, or hand-off to a coverage-family child issue.
+	"envoy-ai-gateway": {
+		"anthropic-messages-protocol-headers": {Issue: 2379},
+		"anthropic-messages-request":          {Issue: 2379},
+		"anthropic-messages-response-shape":   {Issue: 2379},
+		"anthropic-messages-streaming":        {Issue: 2379},
+		"apiserver-classification-endpoints":  {Issue: 2379},
+		"chat-completions-structured-output":  {Issue: 2379},
+		"entrypoint-recipe-routing":           {Issue: 2379},
+		"llm-classifier-distribution-routing": {Issue: 2379},
```
Xunzhuo: [P2] Reconcile the inventories before enabling the gates

On a clean merge with current main (`6f207a7`), `TestGateDCIOverridesMatchKnownExclusions` fails because `llm-classifier-distribution-routing` and `sequence-classifier-routing` are already in `ENVOY_AI_GATEWAY_CI_TESTS` but remain exclusions here. The same run fails Gate C because runtime profile `pii-remote-backend` is missing from `tools/agent/domains.yaml`. Since `build-e2e` now runs these gates, the integrated change blocks E2E builds. Remove the two resolved exclusions and register the missing profile, then rerun the gates on the integrated tree.

Xunzhuo (changes requested):
The CI-override layer addresses the earlier coverage concern. Two inventory mismatches remain on a clean merge with current main: the focused verification run passes 8 tests and fails 2, and the runtime audit confirms the stale exclusions. Please reconcile those inventories before enabling the mandatory gates.

e2e/pkg/verification/ciexclusions.go
gavinkvx: Reconciled on top of main c16c21f3b (rebased; head 0f2ec4566).

- Gate D: between your review and now, main replaced the explicit `ENVOY_AI_GATEWAY_CI_TESTS` list with the `baseline-suite` input (`E2E_BASELINE_SUITE`, consumed by `testmatrix.BaselineCases` in the runner). With no explicit override left in `integration-test-k8s.yml`, the gate's own rule applies — a profile whose override is removed loses its whole table — so `KnownCIExclusions` is now empty rather than shortened by the two resolved entries. The gate still rejects a table for a profile without an override, and the committed pins keep covering the omission / stale / phantom directions for any future explicit list.
- Gate C: `pii-remote-backend` is registered in `tools/agent/domains.yaml`, mirroring the other remote-backend profiles (profile directory, its deploy manifests, and `pii_backend_routing.go`).

On the rebased tree `go test ./pkg/verification/...`, `make e2e-coverage-check` and `make build-e2e` are green, and the audit reports 33 profiles / 176 testcases / 17 unreachable (debt table unchanged) / no CI overrides. The PR description is updated to match.

The `baseline-suite` selector is a runtime selection layer the audit does not model; I left it out of this PR as a candidate follow-up slice under #2379 — happy to file it if you want it tracked.

Xunzhuo (approved):
The runtime-derived profile/test inventory, bounded debt and explicit CI-override checks are consistent with the declared scope. Current-head CI executed all 10 Gate A–D tests, 92 E2E unit cases and 395 harness cases. The new build hook also ran before the passing Envoy AI Gateway suite. The separately implemented baseline-suite selection layer remains outside this audit.

wilsonwu (approved):
LGTM

PR #3504, opened 2026-09-05
[Bug] Key RouterDC affinity on a SHA-256 digest of the full query instead of its 32-byte prefix

### LINKED ISSUE #3452: [Bug] RouterDC hashQuery keys affinity by 32-byte prefix, not by query
### Proposed Workgroup

MoM & Routing

### Describe the bug

`hashQuery` (`selection/router_dc.go:443`) is not a hash. It returns `%x` of the first 32 bytes of the query. Every query sharing a 32-byte prefix lands in one `affinityMatrix` bucket, so `UpdateFeedback` learns model affinity per prompt-template prefix, not per query. Same shape as #3366 at a smaller scale.

### To Reproduce

Verbatim copy of `hashQuery` run standalone with a 48-char instruction prefix:

| input | result |
|---|---|
| prefix + "write a poem about the ocean" | same key as below |
| prefix + "SQL to drop the production users table" | same key as above |

In-package: call `RouterDCSelector.UpdateFeedback` with two feedbacks whose `Query` values share the first 32 bytes and observe they update the same `affinityMatrix` entry.

### Expected behavior

Key affinity on a digest of the whole query, e.g. `sha256.Sum256([]byte(query))` as hex, matching `cache.go` and `config/loader.go`.

### Impact

Latent today. No production path constructs a `selection.Feedback` or calls `UpdateFeedback`; only tests and `paper_benchmark_test.go` do, and `affinityMatrix` is never read by `Select`. Once a feedback endpoint or learning loop is wired in, any deployment with a system prompt or template longer than 32 bytes collapses all queries into one bucket and the learned affinity is meaningless.

### Additional context

_No response_

yaojiejia:
Closes #3452

## Purpose

`RouterDCSelector.hashQuery` in `src/semantic-router/pkg/selection/router_dc.go` hex-encoded the first 32 bytes of the query instead of hashing it. Every query that shared a 32-byte prefix, such as anything behind a system prompt or template, collapsed into one `affinityMatrix` bucket, so `UpdateFeedback` learned affinity per prefix rather than per query.

This change keys affinity on a hex-encoded SHA-256 digest of the full query, matching the digest pattern already used in `pkg/cache` and `pkg/config`. The 32-byte truncation and its length branch are removed. A new `router_dc_test.go` covers the hash and the feedback path.

## Test Plan

```bash
cd src/semantic-router
export LD_LIBRARY_PATH=$PWD/../../ml-binding/target/release:$PWD/../../candle-binding/target/release:$PWD/../../nlp-binding/target/release
CGO_ENABLED=1 go test ./pkg/selection/ -run 'TestRouterDCSelector_' -count=1 -v
CGO_ENABLED=1 go test ./pkg/selection/ -count=1
gofmt -l pkg/selection/ && go vet ./pkg/selection/
```
The two new tests were also run against the previous hashQuery implementation to confirm they fail without the fix.

## Test Result

- Both tests fail on the previous implementation: queries sharing a 54-byte prefix produced the same key and landed in one bucket.
- Full pkg/selection package passes.
- gofmt and go vet are clean.
- golangci-lint run ./pkg/selection/ reports three pre-existing errcheck findings in router_r1_client.go, which this PR does not touch.

---
<details>

- [x] PR title begins with exactly one bracketed category ...
- [x] The title does not stack prefixes ...
- [x] The PR links an accepted issue with exactly one owner ...
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope ...

</details>

### DIFF
--- a/src/semantic-router/pkg/selection/router_dc.go
+++ b/src/semantic-router/pkg/selection/router_dc.go
@@ -18,6 +18,8 @@ package selection
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"strings"
@@ -439,13 +441,11 @@ func (r *RouterDCSelector) applySoftmax(scores map[string]float64) map[string]fl
 	return result
 }
 
-// hashQuery creates a simple hash for query tracking
+// hashQuery returns a hex-encoded SHA-256 digest of the full query so that
+// affinity is keyed per query rather than per shared prefix.
 func (r *RouterDCSelector) hashQuery(query string) string {
-	// Simple hash for query grouping (could use more sophisticated methods)
-	if len(query) < 32 {
-		return fmt.Sprintf("%x", query)
-	}
-	return fmt.Sprintf("%x", query[:32])
+	sum := sha256.Sum256([]byte(query))
+	return hex.EncodeToString(sum[:])
 }
 
 // defaultSelection returns the first configured candidate when embedding-based
--- a/src/semantic-router/pkg/selection/router_dc_test.go
+++ b/src/semantic-router/pkg/selection/router_dc_test.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2025 vLLM Semantic Router.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


FAUST-BENCHOU (approved):
LGTM. Hashing the full query fixes the shared-prefix affinity collision while keeping the key deterministic and bounded, and the feedback-path regression test covers the actual failure mode.

wilsonwu (approved):
LGTM

PR #3489, opened 2026-09-04
[Feature] Add unified model catalog and model hub

Xunzhuo:
Related #2358

## Purpose

Land one extensible source of truth for runtime providers, wire protocols, model cards, provider mappings, reasoning behavior, presentation metadata, benchmarks, and evaluations, then consume it consistently in the Router, CLI, Dashboard, and website.

The Model Hub inventory is curated by **model creator company**, not by counting individual models or mirroring every runtime endpoint. This snapshot focuses on 22 mainstream creators and carries roughly their latest three generations or current representative product lines.

## What this PR delivers

- Keeps the existing `v0.3` config hierarchy. A built-in model opts in with the optional `providers.models[].catalog`; a private or self-hosted model can omit it and keep a handwritten sparse `routing.modelCards` entry.
- Standardizes `backend_refs[].provider` on stable Provider IDs and materializes built-in and custom definitions into the same effective runtime registry.
- Replaces provider-specific OpenAI, Anthropic, Azure, Bedrock, Gemini, Vertex, and MiniMax glue with data-backed Provider and Protocol definitions.
- Defines API operations (`chat/completions`, `responses`, `messages`, embeddings, and related capabilities), authentication, headers, base paths, presentation, and Add Model fields in the unified Provider schema.
- Separates three identities cleanly: `ModelCard.publisher` is the creator, `ProviderDefinition` is a serving/API contract, and a provider model mapping binds a canonical model to a native model ID plus supported operations.
- Preserves provider/model logos and makes Dashboard Add Model, Dashboard Model Hub, and website `/models` consume the same generated catalog snapshot.
- Presents benchmark evidence as a filterable gallery: desktop uses two comparison panels per row, mobile uses one, catalog-backed chips replace the benchmark dropdown, and every eligible model/effort result stays on one chart without pagination. `Core` is the first semantic filter and contains MMLU-Pro, GPQA Diamond, Humanity’s Last Exam, SWE-bench Verified, Terminal-Bench 2.1, and SciCode.
- Makes reasoning families built in for known models while retaining explicit reasoning metadata for custom vLLM/SGLang models.
- Replaces ambiguous scalar `quality_score` with exact, versioned benchmark records. Every visible built-in score is normalized to a percentage while the source value and unit remain available as raw evidence. Model Hub exposes benchmark-specific comparisons only; it intentionally does not publish an overall model rank when coverage differs.
[...]

PR #3505, opened 2026-09-05
[Test] Migrate the dashboard profile values to canonical v0.3

rishabhsinha17:
Related #3233

Standalone prerequisite requested in https://github.com/vllm-project/semantic-router/issues/3233#issuecomment-5539026827: migrate `e2e/profiles/dashboard/values.yaml` off the pre-v0.3 schema before the next dashboard journey lands, so tests stop accumulating per-test `modelCards` workarounds around a stale shared fixture.

## Legacy vs effective

The old file declared `model_config`/`categories`/`decisions`/`strategy`/`default_model`/`bert_model`/`classifier`/`prompt_guard`/`response_cache` at the top level of the config block. The v0.3 loader drops every one of those keys, so what the router actually ran was the chart defaults: `replace-with-your-model` placeholder providers and modelCards, the `general` domain, the `default-route` decision, plus the file's one surviving key (`global.model_catalog.kbs: []`). The file documented a router that never existed.

## What changed

`e2e/profiles/dashboard/values.yaml` is rewritten in canonical v0.3 (`version`/`listeners`/`providers`/`routing`/`global`, mirroring `e2e/profiles/ai-gateway/values.yaml`), encoding what the legacy keys meant where that is discernible:

- `model_config."base-model"` (qwen3, LoRA `general-expert`) -> `providers.models[base-model]` + `routing.modelCards` with the same LoRA and description. The backend ref points at `vllm-llama3-8b-instruct.default.svc.cluster.local:8000`, which this profile already deploys through the shared gateway stack resource manifests.
- `categories: [other]` -> `routing.signals.domains: [other]`.
- top-level `decisions` (`other_decision`) -> `routing.decisions`, unchanged in content.
- `strategy: priority` -> `global.router.strategy: priority`.
- `default_model: general-expert` -> `providers.defaults.model` (the post-#3489 canonical spelling; the merge with current main also respells `reasoning.family` and adds `provider` to `backend_refs`), with a bare `general-expert` modelCard so the reference resolves, same pattern as the ai-gateway profile.
- `response_cache.enabled: false` -> `global.stores.response_cache.enabled: false`. This is an intent-encoding change of effective behavior: the dropped legacy key meant disabled, while the chart default had it enabled.
- `prompt_guard.enabled: false` -> no jailbreak signal declared, which is how v0.3 spells disabled.
- `bert_model`/`classifier` model IDs and thresholds are dropped rather than translated: they have no canonical top-level home, they duplicated loader/registry defaults, and the effective config never contained them anyway.
- `listeners: []` follows the convention of every migrated e2e profile on this gateway stack. This drops the chart-default router-owned listeners from the rendered config; the ExtProc/API servers are not configured through `listeners`, which twenty-odd profiles on the same stack already demonstrate.

Follow-through edits keep the assertions honest:

- `e2e/testcases/dashboard_deploy_safe_failure.go`: the fragment no longer carries its own `routing.modelCards`. The fixture now declares the model surface, merge mode leaves that node untouched, so the unknown-model check fires against the shared fixture instead of a per-test workaround. Verified locally: the merged document is rejected with `routing.decisions[e2e-safe-failure].modelRefs[e2e-nonexistent-model] references unknown model "e2e-nonexistent-model"`.
- `e2e/testcases/dashboard_config_read.go` (review follow-up): `config/all` and `config/yaml` must each declare `base-model` under `providers.models` and `other_decision` under `routing.decisions`, decoded into JSON-shaped maps and checked by one walker, instead of passing on any nonempty payload. A Dashboard serving chart placeholders now fails both reads.
- `src/semantic-router/pkg/config/maintained_asset_contract_test.go`: the file is registered in `maintainedValuesConfigAssets`, so `TestMaintainedConfigAssetsUseCanonicalV03Contract` now parses it with the real loader and rejects any regression to legacy keys.

## Verification

- `go test ./pkg/config/ -run TestMaintainedConfigAssetsUseCanonicalV03Contract` passes with the new asset registered.
- `helm template` over `deploy/helm/semantic-router` with the new values renders a config with no legacy keys; the rendered document parses via `ParseYAMLBytes`, and the safe-failure merge is rejected by the unknown-model validator as described above.
- gofmt, `go build ./cmd/e2e`, `go vet`, yamllint clean on the touched files.
- The dashboard profile e2e (`selection: pr` in the domain registry for `e2e/profiles/dashboard/**` and `e2e/testcases/dashboard_*.go`) is the real proof and triggers on this PR.


Xunzhuo (changes requested):
Thanks for migrating the shared fixture. The requested config/read/deploy follow-through is incomplete: `dashboard-config-read` still only checks a nonempty JSON object and nonempty YAML body. Both assertions pass if the Dashboard continues serving chart placeholders instead of the new `base-model`/`other_decision` fixture, so this prerequisite still does not prove the user-facing read surfaces reflect the migrated config. Please assert canonical identifying fields on both JSON and YAML reads while retaining the deploy safe-failure assertion.

rishabhsinha17:
Reworked per the review. dashboard-config-read now asserts the canonical identifying fields on both read paths: config/all and config/yaml must each declare base-model under providers.models and other_decision under routing.decisions, checked by one walker over the JSON-shaped decode of each response. A Dashboard serving chart placeholders now fails both reads instead of passing on nonempty output. The deploy safe-failure assertion is retained unchanged.

The branch also merges current main to pick up the model catalog schema change from #3489, respelling the fixture to the new canonical fields (defaults.model, reasoning.family, backend_refs provider). The 09-05 dashboard go mod tidy failure does not reproduce after the merge. Verified locally: the maintained-asset contract suite over the migrated file, dashboard frontend tests and type-check, both go mod tidy checks, and the changed-file pre-commit hooks.

rishabhsinha17:
@Xunzhuo checks are green on the rework, ready for another look. Review re-request via gh needs triage permission on this repo, so pinging here instead.

rishabhsinha17:
@Xunzhuo following up on the 09-05 review. 01f91802 adds the canonical base-model and other_decision assertions on both the JSON and YAML dashboard config reads, with the deploy safe-failure assertion retained. The branch merged main again on 09-14 (0d18e4dd) and every check is green on that head. Could you take another look when you have a chance?

Xunzhuo (approved):
Both JSON and YAML reads now assert the deployed base-model and other_decision fixture, addressing the previous placeholder-coverage gap. The exact-head dashboard E2E check passes.

wilsonwu (approved):
LGTM

PR #3506, opened 2026-09-05
[Feature] Let neutral clients use Anthropic prompt caching

### LINKED ISSUE #3350: [Feature] Add route-local provider-aware prompt-cache marker injection
### Goal

Define an explicit, disabled-by-default route policy that can add provider-aware prompt-cache markers to eligible stable request segments while preserving caller intent and protocol compatibility.

### Background

Current `main` preserves caller-supplied prompt-cache markers on supported translation paths, but it does not add markers when the caller omits them. A client using a neutral or different request format therefore cannot opt into provider prompt caching through Router policy alone. Any mutation must be explicit, route-local, capability-checked, and subordinate to caller intent.

Parent Epic: #2358  
Related marker preservation: #3055

### Proposed Workgroup

Data Plane & Networking

### Scope

- Decide which maintained provider and protocol profiles may support Router-managed marker injection.
- Add validated route-local configuration that is disabled by default.
- Define eligible stable segments, marker count and ordering limits, caller-marker precedence, and capability checks.
- Preserve protected instructions, tool schemas, tool exchanges, and protocol structure.
- Keep marker injection separate from response caching, context compression, and tool selection.
- Define typed fail-open or fail-closed behavior for unsupported providers, lossy conversions, conflicting markers, and policy errors.
- Record bounded content-free receipts describing whether markers were preserved, inserted, skipped, or rejected.

### Non-goals

- Enable request mutation by default.
- Guarantee a provider cache hit or a particular cost reduction.
- Cache raw prompts inside the Router.
- Replace caller-managed cache markers or provider retention policy.
- Add markers on an unsupported cross-protocol path silently.

### Completion signal

- [ ] Maintainer acceptance records the policy decision and supported provider boundary.
- [ ] Disabled configuration preserves the current caller-controlled behavior exactly.
- [ ] Caller-provided markers remain authoritative under documented precedence rules.
- [ ] Injection is bounded, capability-checked, observable, and protocol-safe.
- [ ] Tests cover preserved, inserted, conflicting, unsupported, cross-format, streaming, and disabled cases.

## First supported vertical

Treat this as one bounded leaf, not separate configuration and mutation tasks. The first supported profile is Anthropic: route-local disabled-by-default configuration, caller-marker precedence, provider capability validation, injection after all context/tool transformations and model/provider selection but immediately before final codec encoding, content-free receipts, and tests. Additional provider profiles can follow after this end-to-end path is stable. Unconsumed public configuration is not a complete deliverable.

1fanwang:
Closes https://github.com/vllm-project/semantic-router/issues/3350

## Purpose

Neutral clients routed to Anthropic cannot request cache breakpoints without vendor-specific fields. An opt-in route policy marks stable instructions and tools during provider encoding while preserving caller markers.

The policy is disabled by default and inserts at most two markers. Unsupported targets follow the route's skip/reject setting.

## Test Plan

```bash
make e2e-test E2E_PROFILE=anthropic-shim E2E_VERBOSE=true E2E_KEEP_CLUSTER=false
```

## Test Result

The [Kind run](https://github.com/vllm-project/semantic-router/actions/runs/35670731243/job/106580617573) exercised buffered and streaming requests through the maintained Anthropic shim. Its tested source tree matches this revision byte for byte. Raw receipt output:

```text
[PromptCache] mode=buffered action="inserted" reason="" inserted="2" preserved="" instruction_markers=1 tool_markers=1 message_markers=0 instruction_ttl="1h" tool_ttl="1h"
[PromptCache] mode=caller action="preserved" reason="caller_markers" inserted="" preserved="1" instruction_markers=1 tool_markers=0 message_markers=0 instruction_ttl="5m"
[PromptCache] mode=disabled action="" reason="" inserted="" preserved="" instruction_markers=0 tool_markers=0 message_markers=0
```

This verifies the routed marker policy and protocol behavior, not vendor cache hits or billing.

---
<details>

</details>


AayushSaini101:
Thanks for the implementing the cache policy @1fanwang QQ: is it only need for the Anthropic configuration side, can you confirm shall we need for other providers, also please update the document as we are updating new config parameter 🙇

Xunzhuo (commented):
Thanks—the route-local validation, caller-marker precedence, final pre-codec injection, and content-free receipts line up with #3350. Please update this branch onto current `main` and rerun the exact-head gates, since the current suite is still red on the dashboard `go.sum` tidy check.

dashboard/frontend/src/lib/dslCapabilityPluginSchemas.test.ts
```diff
@@ -49,4 +49,18 @@ describe('capability plugin field schemas', () => {
       'revision',
     ])
   })
+
+  it('exposes the prompt cache marker contract', () => {
+    const fields = getCapabilityPluginFieldSchema('prompt_cache')
```
Xunzhuo: **Exercise the exported resolver and close its prototype-key path**

Impact: These new tests call the unimported `getCapabilityPluginFieldSchema`, so the dashboard and pre-commit gates fail and they never exercise the exported resolver used by the UI. On current `main`, that resolver first indexes a plain object, so `constructor`, `toString`, and `__proto__` return inherited non-array values instead of reaching this PR's safe `Map` fallback.

Required change: Update these tests to call `resolveCapabilityPluginFieldSchema`, make its first lookup own-key safe (or use a `Map` there too), and cover both `prompt_cache` and the prototype keys returning `null`.

Xunzhuo (commented):
Thanks for the update. One correctness blocker remains.

FAUST-BENCHOU:
The prompt-cache policy is nicely contained at the route/plugin boundary: it stays opt-in, preserves caller markers, and exposes the same contract through DSL, CRD, Dashboard, and topology surfaces. Good catch on making the schema lookup prototype-safe as well. The Anthropic-shim E2E gives this solid wire-level coverage. LGTM from my review.

dashboard/frontend/src/lib/dslCapabilityPluginSchemas.test.ts
1fanwang: Done in https://github.com/vllm-project/semantic-router/commit/bfcba641b4e65fca11303b199a0817e8c5848345. The exported resolver rejects inherited keys.

Issue #3339, opened 2026-09-02
[Refactor] Route Looper model calls through the shared connector


JiaoliangYu:
### Goal

Follow up on #3328 by removing the duplicate HTTP transport policy in `pkg/looper` and routing Looper model calls through the shared remote connector, while preserving Looper protocol and algorithm behavior.

This was suggested in the review follow-up on #3328: https://github.com/vllm-project/semantic-router/pull/3328#issuecomment-5506916960

### Proposed Workgroup

MoM & Routing

The primary changed caller is the Looper execution path. The shared connector remains owned by the model-runtime boundary established in #3128.

### Scope

- Keep OpenAI request shaping, streaming/non-streaming response parsing, logprob extraction, usage accounting, latency, and internal Looper headers in `pkg/looper`.
- Move endpoint validation, per-attempt timeout/cancellation, bounded request/response/error bodies, typed transport errors, and HTTP connection lifecycle to `pkg/modelruntime/connector`.
- Replace mutable request metadata on `looper.Client` (`decisionName` and `fusionDepth`) with per-call metadata so one client is safe to reuse concurrently.
- Reuse one Looper connector per router generation and close it when that generation retires, rather than creating a new connection pool for every Looper execution.
- Preserve non-retryable semantics for generative model calls.
- Add focused contract and lifecycle coverage for headers, cancellation, bounded bodies, concurrent request isolation, and router-generation shutdown.

### Non-goals

- Changing Looper algorithms, candidate selection, prompts, or output semantics.
- Adding transparent retries for generation requests.
- Implementing passthrough streaming; Looper continues to aggregate its bounded upstream response.
- Migrating Router-R1, AutoMix verifier, MCP, embedding, or other unrelated HTTP clients.
- Redesigning the Looper configuration schema.

### Completion signal

- The Looper model-call path no longer constructs requests or executes `net/http` directly.
- Request-scoped Looper metadata remains isolated under concurrent calls.
- A router generation owns and closes exactly one reusable Looper connector.
- Existing Looper behavior tests and the repository-selected Go gates pass without user-visible behavior changes.

JiaoliangYu:
@Xunzhuo could you please confirm the Workgroup ownership and `/accept` this follow-up if the boundary matches what you had in mind on #3328? I have completed the transport/lifecycle audit and will start implementation once the issue passes the repository acceptance gate.

Xunzhuo:
Thanks @JiaoliangYu

[...]

PR #3508, opened 2026-09-05
[Feature] Wire Looper calls to the shared connector

JiaoliangYu:
Related #3339

## Purpose

Introduce the new Looper model-call path without removing the legacy transport yet, so the wiring can be reviewed independently from deletion.

- replace mutable per-client decision and Fusion metadata with typed, request-scoped `ModelTarget` and `CallOptions`
- add request-scoped headers to the shared model-runtime connector while keeping authorization precedence explicit
- route Looper calls through a non-retryable connector operation with bounded bodies and an explicit HTTP 200 success contract
- create one connector-backed Looper client per router generation and register it with the generation resource scope
- migrate Base, Confidence, Ratings, ReMoM, Workflows, Fusion, prompt selection, context recovery, and `fusioneval` call sites without changing their algorithm or response behavior

The legacy Looper HTTP branch remains in this PR only as a reviewable migration bridge. A focused follow-up removes it after every caller is on the typed path.

## Test Plan

- `go test -count=1 -race ./pkg/modelruntime/connector ./pkg/looper ./pkg/extproc`
- `go vet ./pkg/modelruntime/connector ./pkg/looper ./pkg/extproc`
- `make test-semantic-router PWD=/workspace` on the stacked final tree
- `make agent-ci-gate ENV=cpu CHANGED_FILES="..."` on the stacked final tree

## Test Result

All listed checks pass in `rocm/atom-dev:latest` on g64. The concurrency contract test overlaps two requests and verifies model, decision, iteration, Fusion depth, and authorization headers do not cross. The endpoint contract test verifies an empty operation path preserves the configured `/v1/chat/completions` endpoint exactly. The status-policy tests verify that the shared connector still accepts any 2xx by default while Looper rejects a streaming 204 response.

---
<details>

- [x] The title uses one category only
- [x] The PR links accepted issue #3339 with `wg/mom-routing` and one owner
- [x] Purpose, Test Plan, and Test Result reflect the actual scope and validation

</details>


JiaoliangYu:
The deletion-only follow-up is #3509. This keeps the connector wiring and the removal of the legacy transport independently reviewable.

Xunzhuo (changes requested):
Thanks for the careful connector migration. One response-contract regression remains: the shared connector accepts every 2xx response, while Looper’s existing path requires HTTP 200. Because the streaming parser accepts an empty body, a 204 now becomes a successful empty model response instead of an upstream-status error. Please preserve Looper’s 200-only contract and add a 204 regression test before removing the legacy path.

JiaoliangYu:
Thanks for catching this — you are right that `204` could otherwise reach the streaming parser as a successful empty response. Fixed in `6ff1a1f9`.

The connector operation now has an optional `SuccessStatusCode`: zero preserves the shared connector default of accepting any `2xx`, while Looper explicitly requires `200 OK`. I added both the connector policy contract test and a Looper streaming `204` regression test.

Xunzhuo (commented):
Thanks—`SuccessStatusCode: http.StatusOK` now preserves the Looper 200-only contract while the shared connector keeps its default 2xx behavior, and the 204 regression covers the original blocker. This branch is behind current `main`; please sync it, let the resulting exact-head checks finish, and re-request review.

JiaoliangYu:
Hi @xunzhuo, the branch is now synced with main and the conflicts are resolved. Could you please take another look when you have a moment? Thanks!

Xunzhuo (approved):
Thanks!

PR #3512, opened 2026-09-05
[Test] Run the long-text classification API cases in the PR lane

subin9:
Related #3333, related #3204, related #3364

## Purpose

- `pii-entity-offsets` (#3147) and `pii-long-text` (#3369) are in the baseline matrix, but the envoy-ai-gateway lane runs a fixed case list that never included them, so neither regression has run in CI. Add both to the list.
- Add `security-long-text`: the security classification API had no long-text case although #3206 fixed the same truncation there. It places a jailbreak prompt the profile already blocks on the routing path after benign text long enough to fall outside the classifier's window, and expects `/api/v1/classify/security` to detect it. The filler is scored on its own first so a false positive cannot pass the case.

## Test Plan

- `cd e2e && go build ./... && go vet ./testcases/`
- Filler and combined text through `CheckForJailbreakWithRisk` on the shipped `mmbert32k-jailbreak-detector-merged`, locally
- The envoy-ai-gateway lane on this PR runs the three cases

## Test Result

- Build, vet, yamllint and the structure gate pass.
- On the shipped model every filler chunk scores at most 0.003, the combined text scores 0.975, and the first chunk alone (what the truncating code saw) scores 0.001, so the case fails on the old code and cannot pass on the filler.
- Not yet run on a cluster; this PR is its first lane run.

### DIFF
--- a/.github/workflows/integration-test-k8s.yml
+++ b/.github/workflows/integration-test-k8s.yml
@@ -87,7 +87,7 @@ jobs:
           set +e
           if [ "${{ matrix.profile }}" = "envoy-ai-gateway" ]; then
             # Temporarily skip the stress / pressure coverage until the suite is stable again.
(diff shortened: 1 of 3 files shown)


AayushSaini101 (approved):

wilsonwu (approved):
LGTM, thanks!

guan404ming (approved):

PR #1943, opened 2026-05-26
[Bindings] Preprocess images in Rust with PIL-equivalent resize to close cosine drift

shraderdm:
## Purpose

- **What:** Replace the Go-side `decodeAndResizeImage` 4-tap bilinear preprocessing (introduced in PR #1414) with a Rust-side resize via the `image` crate's `FilterType::CatmullRom`, which approximates PIL's `Image.BICUBIC` + `antialias=True` behavior used by `SiglipProcessor`.
- **Why:** Candle-binding's multi-modal-embed-small (mmes) image pipeline carried a ~1% cosine drift versus the PyTorch reference even after the pooling-head (PR #1927) and image-normalization (PR #1928) fixes closed earlier correctness gaps. This PR closes the remaining residual.
- **Module:** `Bindings` (`candle-binding/`).

The fix replaces the Go `decodeAndResizeImage` helper (which used 4-tap bilinear interpolation with no antialias) with a new Rust FFI entry point, `multimodal_encode_image_from_bytes` in `candle-binding/src/ffi/embedding.rs`. Decode + resize + CHW-float32 conversion live in a private helper, `decode_resize_to_chw_f32`, which uses the `image` crate's `imageops::resize` with `FilterType::CatmullRom` (cubic B-spline, B=0, C=0.5). Its support-window-weighted sampling approximates PIL's `Image.BICUBIC` + `antialias=True` behavior on downscale. The helper includes an integer-overflow guard on target dimensions, and its rustdoc documents the alpha-discard and EXIF-orientation behavior (both match PIL defaults; alpha-discard differs slightly because PIL composites against black on `convert("RGB")` while `image::to_rgb8` drops the channel without compositing).

Go's `MultiModalEncodeImageFromBytes` now calls the new FFI directly. The previous `decodeAndResizeImage` Go function is removed along with its `bytes` / `image` / `math` imports.

The `image` crate moves from `dev-dependencies` to `dependencies` with the same feature set (`default-features = false`, `features = ["jpeg", "png"]`). The deployed dylib gains the JPEG + PNG decoder and `imageops::resize` paths that were previously only linked at test time. No new image-format support is introduced (JPEG + PNG only, matching the existing `MultiModalEncodeImageFromBytes` contract).

### Harness gate note (TD-042)

[...]

PR #1927, opened 2026-05-21
[Bindings] Fix SigLIP attentional probe pooling head in candle-binding

shraderdm:
Refs #1909

## Purpose

- **What does this PR change?** `SigLIPVisionEncoder::forward` in candle-binding was performing BERT-style `mean + Linear + tanh` pooling and loading the SigLIP `head.probe` parameter as if it were a Linear weight. The correct SigLIP-base architecture uses `SiglipMultiheadAttentionPoolingHead`: a learned probe parameter doing cross-attention over patch tokens, then LayerNorm + MLP with a residual connection. This PR mirrors `candle-transformers::models::siglip::MultiheadAttentionPoolingHead` into candle-binding (`SigLIPHead` + `SigLIPHeadAttention` + `SigLIPHeadMlp`).
- **Why is this change needed?** The pre-fix encoder produced cosines roughly 4.5x smaller than the Python reference on `multi-modal-embed-small`. The routing-accuracy gap this produced was originally tracked in #1909 as an encoder-strength concern; subsequent diagnosis identified the candle-binding pooling head as the cause (see the 2026-05-19 update on #1909).
- **Which module(s) does this affect?** `Bindings` (candle-binding only; no changes to the Go router, the operator, CLI, dashboard, or any other subsystem).

## Test Plan

- Build candle-binding from `candle-binding/`:
  - `cargo build --release --no-default-features --features accelerate`
- Run the shape-only smoke test added in this PR:
  - `cargo test --release --no-default-features --features accelerate --lib model_architectures::embedding::multimodal_embedding::integration_tests::test_siglip_head_forward_shape`
- Run the rest of the existing module tests to confirm no regression:
  - `cargo test --release --no-default-features --features accelerate --lib model_architectures::embedding::multimodal_embedding`
- For empirical correctness verification against real SigLIP weights, reviewers can use the verification branch at `https://github.com/shraderdm/semantic-router/tree/81183f763bf92ea9a50fb3aabdbe4f7f7e116f74` (pinned to the fixture commit aligned with this PR's measurement), which contains the same pooling-head fix plus three `#[ignore]`-gated diagnostic tests and test image fixtures with a provenance disclaimer in the fixtures commit. The verification branch is intentionally kept separate from this PR because the diagnostic tests reference deployment-specific exemplars and image fixtures that are not appropriate to ship upstream.

## Test Result

[...]

PR #3445, opened 2026-09-03
[Bug] Enforce Fusion quorum over usable panel responses

truddy0:
Describe the change and keep commands and results specific to this PR.

Closes #3374

## Purpose

Enforce Fusion's `min_successful_responses` quorum over usable panel responses.

A panel response is usable only when its `content` or `reasoning_content` remains non-empty after trimming whitespace. Empty choices, whitespace-only responses, tool-only responses, malformed payloads, timeouts, and failed requests do not count toward quorum.

When the usable response count is below quorum:

- Return a typed `FusionQuorumError` with safe per-attempt evidence and aggregate panel usage.
- Stop before grounding, judge analysis, and final synthesis.
- Preserve `on_error: fail` fail-fast behavior.
- Allow `on_error: skip` to skip individual failed attempts, but not the overall quorum requirement.

Responses that satisfy quorum retain their configured ordering and existing early-cancellation behavior. Cached panel responses use the same usability and quorum rules.

This change also adds Fusion unit and E2E coverage and updates the Fusion documentation. It does not add fallback behavior, new configuration fields, a shared panel executor, or generic Replay tracing.

The accepted issue is owned by `wg/mom-routing`.

## Test Plan

```bash
cd src/semantic-router

go test ./pkg/looper \
  -run '^(TestFusion|TestIsUsable)' \
  -count=1

go test ./pkg/looper \
  -run '^(TestFusionLooperRejectsPanelBelowUsableQuorum|TestFusionLooperEmptyPanelResponseDoesNotMeetQuorum|TestFusionLooperMalformedHTTPSuccessDoesNotMeetQuorum|TestFusionLooperAllPanelFailuresReturnError)$' \
  -count=20

go test -race ./pkg/looper \
  -run '^(TestFusionLooperMalformedHTTPSuccessDoesNotMeetQuorum|TestFusionLooperEmptyPanelResponseDoesNotMeetQuorum|TestFusionLooperTimeoutBelowQuorumPreservesPanelEvidence|TestFusionLooperPanelQuorumSkipsSlowWorker)$' \
  -count=1

go vet ./pkg/looper

cd ../..

make test-semantic-router
make build-e2e

make e2e-test-only \
  E2E_PROFILE=looper \
  E2E_TESTS=looper-fusion-usable-quorum \
  E2E_KEEP_CLUSTER=true

[...]

PR #3519, opened 2026-09-05
[Fix] Recalibrate image-routing thresholds against the corrected encoder (#2165)

### LINKED ISSUE #2165: [Bug] Recalibrate image-routing thresholds after encoder correctness fixes
### Proposed Workgroup

### Describe the bug

The maintained image-routing example still uses thresholds tuned before the encoder-correctness fixes in #1927, #1928, and #1943. Those values no longer represent the corrected embedding distribution.

### To Reproduce

### Expected behavior

The example thresholds should be derived from a versioned positive/negative calibration set and should document the tested model and artifact.

### Impact

##### Completion signal

A reproducible report selects the thresholds, the example is updated, and end-to-end tests cover representative matches and non-matches.

Parent Epic: #2347

lawin-steve:
Fixes #2165 (parent epic #2347).

## Summary

The image-routing example pack shipped `threshold: 0.10` on all three rules, tuned before the encoder-correctness fixes in #1927, #1928 and #1943. This PR re-derives the thresholds from a hand-reviewed calibration manifest scored through the real candle multimodal FFI path, records provenance for the commit and the model bytes, and tightens the tests around the result.

## Changes

- **`cmd/image-routing-calibration`** (new): scores every fixture listed in `testdata/calibration-set.json` against the shipped candidate lists and emits JSON + Markdown reports. The manifest is explicit: all 231 repository images (docs, dashboard, e2e fixtures) were reviewed by hand and each is listed as a positive for a named candidate phrase (15), a negative (214), or excluded as ambiguous with the reason (2 config-builder screenshots whose panes are YAML). Nothing is inferred from directory contents, and the manifest is decoded with unknown keys rejected, so a misspelled or legacy section cannot silently drop its images. Every input the report attributes to the commit (fixtures, excluded assets, the manifest, the rules file) must be a canonical repo-relative path, symlink-free, a regular file, and tracked at HEAD; the report records the commit, flags any uncommitted change anywhere in the worktree, and lists every scored and excluded fixture with its sha256. The report is also bound to the model bytes: every loader input under `-model` is hashed into it, and where the Hugging Face download cache recorded what a file resolved to, both the snapshot and the etag must match or the run fails before loading. The tool refuses rules it cannot score honestly (non-image modality, no candidates, or an aggregation other than `max`, since the report records the prototype blend), fails if the classifier returns no score, a duplicate, or a non-finite score for any rule on any fixture, and requires at least one reviewed positive per rule. The classifier's resolved config is the single source for the scoring parameters the report records (target dimension, target layer, prototype scoring). Separable rules take the midpoint of the band between `max(negative)` and `min(positive)`; non-separable rules take the midpoint of the max-F1 band and report the residual confusion matrix. The sweep is built only from the distinct fixture scores, never from the shipped value or a synthetic 0; ties on F1 are compared exactly from the integer confusion counts and go to the wider band. Candidate lists are an input, pinned by sha256 in the report.
- **Thresholds** in `config/fragments/signal/embedding/image-routing.yaml` (the yaml carries a one-line pointer; full provenance lives in the docs, below):
  - `identifier_document_imagery` 0.10 → **0.61** (separable)
  - `code_or_terminal_imagery` 0.10 → **0.47** (not separable, see below)
  - `ambient_office_imagery` 0.10 → **0.54** (separable)
- **E2E CRD mirror** `e2e/profiles/multimodal-routing/crds/intelligentroute.yaml` updated, and `TestImageRoutingPack_MatchesMultimodalE2EProfile` added so the two copies cannot drift. Structural bound widened from `(0, 0.5)` to `(0, 0.9)`.
- **E2E profile**: embedding `top_k` is now explicitly unlimited (pinned in the profile package test next to `target_layer`), so the matched-embeddings header carries every rule at or above its threshold and the six cross-rule "does not fire" cases can observe a second rule firing instead of passing on ranking alone. With the shipped thresholds exactly one rule fires per fixture, so the nine cases are unchanged. Plus a fail-fast on an empty case set.
- **Docs**: short provenance note in the image-routing example of `website/docs/tutorials/signal/learned/embedding.md` (model, snapshot, calibrate-for-your-model guidance), following the tutorial's existing pattern; the v0.3 versioned copy no longer describes the pre-recalibration 0.10 threshold and cosine range. Maintainer detail (selection rule, code-rule caveat, prototype-blend note, CRD lockstep) lives in the calibration command's package doc.
- **CI**: the `multimodal-routing` profile now also triggers on the fragment yaml, the image testcase, its case file, and the fixtures.

## Calibration evidence

Model `llm-semantic-router/multi-modal-embed-small`, snapshot `fdf8e01b7b0f3a69ac1ac8e2a64dcb1ede177ba4`, 384-dim, final encoder layer, default `prototype_scoring` (effective score `0.75*best + 0.25*mean(top 2)`).

## Shipped thresholds

Metrics are computed at the threshold currently in `image-routing.yaml`.

Headroom is the absolute cosine gap from the threshold to the nearest positive
(`+pos`) and nearest negative (`+neg`); a negative value means a misclassified fixture.

| Rule | Shipped | Pos | Neg | TP | FP | FN | Precision | Recall | F1 | Headroom +pos | Headroom +neg | Separable |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| `identifier_document_imagery` | `0.61` | 1 | 228 | 1 | 0 | 0 | 1.000 | 1.000 | 1.000 | `+0.0999` | `+0.0959` | true |
| `code_or_terminal_imagery` | `0.47` | 13 | 216 | 6 | 8 | 7 | 0.429 | 0.462 | 0.444 | `-0.1719` | `-0.0766` | false |
| `ambient_office_imagery` | `0.54` | 1 | 228 | 1 | 0 | 0 | 1.000 | 1.000 | 1.000 | `+0.1489` | `+0.1394` | true |

## Best achievable threshold per rule

Selection rule: if `min(positive) > max(negative)` take the midpoint of that band;
otherwise take the midpoint of the max-F1 band and record the residual confusion
matrix. Margin is the distance to the nearest fixture on either side.

| Rule | Best | F1 | FP | FN | Margin | Separable |
|---|---:|---:|---:|---:|---:|:---:|
| `identifier_document_imagery` | `0.6100` | `1.000` | 0 | 0 | `+0.0959` | `true` |
| `code_or_terminal_imagery` | `0.4700` | `0.444` | 8 | 7 | `-0.1719` | `false` |
| `ambient_office_imagery` | `0.5400` | `1.000` | 0 | 0 | `+0.1394` | `true` |

### Caveat on `code_or_terminal_imagery`

This rule is **not separable** on repo imagery: dashboard and docs UI screenshots score up to 0.55 while several labelled code/terminal images (a curl request in a code window, a source listing, an HTTP request body) score 0.30–0.42. The shipped 0.47 is the report's pick (max-F1 band midpoint, F1 0.444, 8 FP / 7 FN over 13 positives). The review of the manifest found two positives that the earlier directory-based set had counted as negatives (`dashboard/landing.png`, a terminal window with a curl command and its output; blog `amd-2.png`, a curl request in a code window), which is what moved the pick from 0.4632. A stronger code fixture (denser code / terminal content scoring clearly above the hard negatives) is the recommended follow-up.

## Also on this branch

- **Router tests** (`utils_neutral_image_test.go`): main fixed inline-image extraction in #3489 after this branch found it; the three tests here pin the contract (inline data URI reconstructed, remote URL never reaches the encoder, non-image payload rejected).
- **E2E profile test** (`profile_test.go`): asserts the CRD mirrors the shipped pack name by name and value by value, replacing a literal 0.42 pin, and pins `target_layer` and `top_k`; runs inside the calibration gate.

## CI gate

`image-calibration` is a registered job and domain (`tools/agent/domains.yaml`), dispatched from `pr.yml`/`main.yml` and required by PR Gate. It runs when any threshold input changes: the pack yaml, the calibration command and its manifest, the image directories the manifest draws from (a deleted or renamed image fails the gate until it is reviewed out), the whole multimodal-routing E2E profile (its CRD-mirror package test runs only inside this gate), the whole native binding (`candle-binding/**`), the whole classification package, the config contracts the tool decodes and the classifier consumes, the router runtime, the Go dependency manifests, and the build entrypoints (`Makefile`, `tools/make/rust.mk`, `tools/make/e2e.mk`). A change-classifier test asserts every path listed in the manifest selects the gate, so a listed image cannot live outside the covered directories. Unrelated router changes still do not select it. The PR caller passes the pull-request head sha as the reusable workflow's required `ref` input and the job asserts the checked-out commit matches it, so the report's `repo_commit` is the exact head rather than GitHub's merge commit. The job downloads the model pinned to `fdf8e01b…`, verifies every file resolved to that snapshot, asserts the native build left the Cargo lockfiles untouched, runs the profile package test, recalibrates over all 229 listed fixtures, and fails unless the checkout is clean and each shipped threshold reproduces the report's pick and its confusion matrix.

## Reproduce the calibration

Needs the candle libraries built (`make rust-ci`) and the multimodal model downloaded at the calibrated snapshot:

```bash
hf download llm-semantic-router/multi-modal-embed-small \
  config.json model.safetensors tokenizer.json tokenizer_config.json special_tokens_map.json \
  --revision fdf8e01b7b0f3a69ac1ac8e2a64dcb1ede177ba4 \
  --local-dir models/mom-embedding-multimodal
```

```bash
cd src/semantic-router
DYLD_LIBRARY_PATH=$PWD/../../candle-binding/target/release \
LD_LIBRARY_PATH=$PWD/../../candle-binding/target/release \
go run ./cmd/image-routing-calibration \
  -model ../../models/mom-embedding-multimodal \
  -artifact-revision fdf8e01b7b0f3a69ac1ac8e2a64dcb1ede177ba4 \
  -rules ../../config/fragments/signal/embedding/image-routing.yaml \
  -cases ./cmd/image-routing-calibration/testdata/calibration-set.json \
  -fixture-root ../.. \
  -output /tmp/image-routing-calibration.json \
  -markdown /tmp/image-routing-calibration.md
```

`-artifact-revision` is the snapshot the model was downloaded at (the `--revision` above). The tool binds that claim to the bytes it scored with: every loader input is hashed into the report, and the `.cache/huggingface` download records under the model path must agree with the claim on both snapshot and etag, or the run fails before loading. `-fixture-root` must be the repository top level. The report also records the repository commit and flags any uncommitted change in the worktree; the CI gate runs with `-require-clean`, so a dirty checkout fails instead of warning.

## Test plan

- [x] `go build` / `go vet` / gofmt on `cmd/image-routing-calibration`
- [x] `go test ./cmd/image-routing-calibration` (selector, manifest, commit-binding, output-path, model-artifact, and gate regressions)
- [x] `go test ./pkg/config -run ImageRoutingPack` (structural, validator, and E2E-mirror test)
- [x] `go test ./profiles/multimodal-routing/` in `e2e/` and `go vet ./testcases`
- [x] Calibration tool run end-to-end against the local model and all 229 listed fixtures; report tables above are from that run; a tampered model copy with the original download records is refused
- [x] `tools/ci/tests/test_pr_change_classifier.py` (including the manifest-coverage case), `workflow_policy_validation.py`, `validate_workflows.py`
- [x] `multimodal-routing` E2E profile in CI on `00176c19` with `top_k` unlimited: 9/9 (https://github.com/vllm-project/semantic-router/actions/runs/34638995369/job/103397283508); router request-time scores match the calibration report to four decimals (passport 0.7099, code screenshot 0.4748, conference room 0.6889)
- [x] Calibration gate in CI on `00176c19` (exact PR head asserted, clean checkout enforced, 229 fixtures, every shipped threshold equals the pick): https://github.com/vllm-project/semantic-router/actions/runs/34638995369/job/103397283093


lawin-steve:
The two failing checks are `go mod tidy` in `dashboard/backend` complaining about missing go.sum entries. That's already broken on main (checked with a clean checkout, same diff), it started with #3445. This PR doesn't touch go.mod/go.sum.

Should I send the go.sum fix as a separate PR, or just add it here?

Xunzhuo (changes requested):
Thanks for putting together the calibration evidence. One blocker: the checked-in code_or_terminal_imagery threshold is not produced by the stated calibration rule. The report selects 0.4632 at F1 0.375, but the config ships 0.44 to keep one E2E fixture passing; at 0.44 the same set reports 6 TP / 22 FP / 5 FN (F1 0.308). calibrateRule also inserts the current shipped threshold into the sweep, so the recommendation can change when the value being calibrated changes. This leaves the completion signal for #2165—reproducible report selects the thresholds—unmet. Please make selection independent of the existing threshold and either ship the report-selected value or revise the objective/data so the shipped value is what the report selects.

lawin-steve:
@Xunzhuo  - Thanks for reviewingl. Fixed in 081bb2e0. The sweep is now built only from fixture scores, so the pick no longer depends on the value being calibrated, and the config ships what the report selects: - `0.4632` instead of `0.44`.  The e2e cases will pass at 100% with a 0.012 threshold buffer on the `code_or_terminal_imagery` images.

lawin-steve:
The multimodal-routing E2E job was failing before any test ran: the router pod crash-looped because model downloads failed. Cause: this profile overrides the chart's `env` list to set `EMBEDDING_MODEL_OVERRIDE`, and since #3434 that list also carries `HOME`/`TMPDIR`/`HF_HOME` for the read-only root filesystem. Helm replaces lists, so those got dropped and `hf download` wrote to `/.cache` and died. The profile hadn't run in CI since Aug 11, so main didn't catch it.

Fix in 47672b4f: use `extraEnv` instead of `env`. One-key change in the profile values.

lawin-steve:
Second pre-existing break in the multimodal-routing profile, now fixed on the branch: the canonical embedding defaults set `target_layer: 22` (mmBERT depth). With `model_type: multimodal` that default survives, and since #2986 the classifier forwards it to the FFI, where the 6-layer multimodal text encoder rejects it. Every candidate embedding failed at CRD reload, the router kept its bootstrap route, and the gateway 404'd on `base-model`. Reproduced locally against the model: layer 22 fails, 6 and 0 work.

Fix here: pin `target_layer: 6` in the profile values. The same failure would hit anyone following the multimodal worked example in the docs, which only sets `model_type`; a model-type-aware default in the router is the real fix and probably deserves its own issue.

Branch is rebased on current main.

lawin-steve:
multimodal-routing E2E is green on f1d43423: 9/9. Router-side scores at request time match the calibration report exactly (passport 0.7099 vs 0.61, code screenshot 0.4748 vs 0.4632, conference room 0.6889 vs 0.54), and each image routed to its expected decision.

Getting there needed one more router fix and one test fix on top of the profile fixes: inline `data:` images were being dropped by the request snapshot since #3076 (so image rules never ran for anyone), and the test wasn't sending `x-vsr-debug: true`, which the matched-signal headers require since #2229. Both are separate commits; happy to split the router fix into its own PR if you'd rather review it there. Table of everything pre-existing is in the description.

Xunzhuo (commented):
Welcome, and thanks for the thorough correction. The calibration sweep now selects independently of the shipped threshold, `0.4632` matches the report-selected value, and the exact-head multimodal profile exercised the runtime image path in all 9 routing cases. I do not see a remaining blocker; please rebase onto current `main` for the final high-risk approval pass.

lawin-steve:
Rebased onto current main (fdc0efd4). The earlier merge commit is gone, and the go.sum tidy commit dropped out on its own since #3522 landed the same change. Six commits, tree unchanged from the run that passed 9/9.

Xunzhuo (commented):
Thanks for the thorough calibration work. The runtime multimodal E2E is green, but this head still has no CI run that executes `cmd/image-routing-calibration` across the 231 fixtures and selects the shipped thresholds. That report is the core completion evidence for #2165, and contributor-local output is not exact-head CI. Please add an exact-head CI-equivalent calibration run, or another reproducible trusted gate, before approval.

lawin-steve:
Added a CI gate in `da9319db`: a `pull_request` workflow runs `cmd/image-routing-calibration -check` over all 231 fixtures whenever `image-routing.yaml` changes and fails unless the shipped thresholds match the report. Green on this head: https://github.com/vllm-project/semantic-router/actions/runs/34071387796/job/101589267504

lawin-steve:
Rebased onto main after #3489. That PR set the e2e profile's three thresholds to a flat 0.42; this PR restores them to the calibrated values (0.61 / 0.4632 / 0.54) so the profile mirrors `image-routing.yaml` exactly, which `TestImageRoutingPack_MatchesMultimodalE2EProfile` enforces. The nine e2e cases pass at either value; the calibrated ones keep the pack and the profile in lockstep.

@Xunzhuo CI is green on the rebased head; could you take another look?

.github/workflows/image-routing-calibration.yml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Build Rust libraries (CPU-only, no CUDA)
+        run: make rust-ci
+
+      - name: Install HuggingFace CLI
+        run: pip install -U "huggingface_hub[cli]"
+
+      - name: Download the multimodal embedding model
+        run: |
+          set -euo pipefail
+          hf download "${MODEL_REPO}" \
```
Xunzhuo, line 111: **Pin the model snapshot used by the calibration gate**

Impact: The required check downloads the mutable Hugging Face default branch while a provenance mismatch is only a warning, so the same PR head can be calibrated against changing weights and can validate thresholds from a snapshot other than the documented one.

Required change: Pass CALIBRATED_REVISION to every hf download with --revision, fail closed if the resolved revision differs before running -check, and use the same pin in the documented reproduction command.

Xunzhuo (commented):

.github/workflows/image-routing-calibration.yml
lawin-steve: Done in `b9fd5ed3`. `hf download` now passes `--revision` with the calibrated snapshot, a step verifies that every downloaded file resolved to that commit and fails the job before `-check` runs, and the documented reproduction command carries the same pin. Gate is green on this head with the pin: https://github.com/vllm-project/semantic-router/pull/3519/checks

.github/workflows/image-routing-calibration.yml
```diff
@@ -0,0 +1,142 @@
...
+# set and fails unless each shipped threshold in
+# config/fragments/signal/embedding/image-routing.yaml equals the value the
+# report selects. Runs as a PR check only when that file changes, so a
+# threshold cannot be edited without the calibration reproducing it; it can
+# also be dispatched by hand against any ref. The Markdown report lands in
+# the job summary and both reports upload as an artifact.
+on:
+  pull_request:
+    paths:
+      - config/fragments/signal/embedding/image-routing.yaml
```
Xunzhuo: **Make calibration changes enter the required gate**\n\nImpact: This standalone workflow only runs when the threshold YAML changes, while current main protects only `PR Gate` and that gate does not depend on this job. A change to the calibration command, case manifest, fixture roots, workflow, or model revision can therefore bypass calibration entirely, and a failing calibration run would not block merge.\n\nRequired change: Route calibration through the required PR gate (or make its exact context required) and trigger it for every input that can change the selected thresholds.

e2e/profiles/multimodal-routing/crds/intelligentroute.yaml
```diff
@@ -5,18 +5,17 @@ metadata:
...
+    # candidates are short text anchors that get embedded in the shared
+    # multimodal text/image space; the request-path image embedding
+    # cosine-matches against them. Requires
     # global.model_catalog.embeddings.semantic.embedding_config.model_type=multimodal
     # in values.yaml.
     embeddings:
       - name: "identifier_document_imagery"
         queryModality: "image"
-        threshold: 0.42
+        threshold: 0.61
```
Xunzhuo, line 18: **Update the E2E profile threshold assertion**\n\nImpact: These CRD values move from `0.42`, but `e2e/profiles/multimodal-routing/profile_test.go:89-91` still rejects every threshold that is not exactly `0.42`. Running that package test now fails, and the current CI selection does not exercise it.\n\nRequired change: Update the test to assert the three calibrated name/value pairs (or compare this CRD with the shipped pack), and ensure the package is covered by the gate.

Xunzhuo (commented):
Thanks for fixing the model pin; I found two remaining gates to make this calibration change merge-safe.

.github/workflows/image-routing-calibration.yml
lawin-steve: Done in `a0b8433d`. `image-calibration` is now a registered job and domain in `tools/agent/domains.yaml`, dispatched from `pr.yml` and `main.yml` and listed in both gates' `needs`, so a failing run blocks merge. The domain covers every threshold input: the pack yaml, the calibration command and its set, the image fixtures, the profile CRD and values, and both fixture roots (docs images are negatives, so the domain is selected for docs-only changes under `website/static/img` too). The workflow selects itself when edited, which is where the model pin lives. First gated run is green: https://github.com/vllm-project/semantic-router/actions/runs/34300426758/job/102306997654

e2e/profiles/multimodal-routing/crds/intelligentroute.yaml
lawin-steve: Done in `a0b8433d`. The test now parses the CRD and the shipped pack and asserts the three rules match name by name and value by value, and the package runs inside the calibration job via `make test-e2e-profile-multimodal`, so it is covered by the gate.

tools/agent/domains.yaml
```diff
@@ -105,6 +109,35 @@ domains:
...
+      - candle-binding/src/model_architectures/model_factory.rs
+      - candle-binding/src/model_architectures/traits.rs
+      - candle-binding/src/core/**
+      - candle-binding/src/ffi/embedding.rs
+      - candle-binding/semantic-router.go
+      - candle-binding/Cargo.toml
+      - candle-binding/Cargo.lock
+      - src/semantic-router/pkg/classification/embedding_classifier*.go
+      - src/semantic-router/pkg/classification/classifier_signal_embedding*.go
+      - src/semantic-router/pkg/classification/prototype_scoring*.go
```
Xunzhuo: **Run calibration for every scoring input**\n\nImpact: Changes to `prototype_bank.go` or `prototype_clustering.go` can change the prototype representatives and therefore fixture scores, but neither path matches `prototype_scoring*.go`; PR Gate can pass with thresholds calibrated against stale scoring behavior.\n\nRequired change: Include the full prototype scoring pipeline (for example, `prototype_*.go`) in this domain and add classifier assertions for the bank and clustering paths.

src/semantic-router/cmd/image-routing-calibration/main.go
```diff
@@ -0,0 +1,636 @@
...
+			return candidate
+		}
+		return evaluateThreshold(midpoint, rule, positive, fixtures)
+	}
+	best := sweep[0]
+	for i := 1; i < len(sweep); i++ {
+		// Band (sweep[i-1].Threshold, sweep[i].Threshold]: same matrix as sweep[i].
+		bandMid := (sweep[i-1].Threshold + sweep[i].Threshold) / 2
+		candidate := evaluateThreshold(bandMid, rule, positive, fixtures)
+		if candidate.F1 > best.F1 || candidate.F1 == best.F1 && candidate.Margin > best.Margin {
```
Xunzhuo: **Use the promised wider-band tie-break**\n\nImpact: `Margin` is calculated from the global min-positive/max-negative bounds, not from this adjacent score band’s width, so tied max-F1 candidates can select a different threshold than the documented wider-band rule.\n\nRequired change: Compare the actual adjacent band widths when F1 ties and add a selector regression with equal-F1 bands of different widths.

Xunzhuo (commented):
Thanks for the thorough update—two current-head gate-contract issues still need to be addressed before this can merge.

tools/agent/domains.yaml
lawin-steve: Done in `fab6ac75`: the domain now matches `prototype_*.go`, so `prototype_bank.go` and `prototype_clustering.go` select the gate, and the classifier test asserts both paths.

src/semantic-router/cmd/image-routing-calibration/main.go
lawin-steve: Done in `84dc0774`: ties on F1 now compare the adjacent band widths and take the wider band; `TestSelectThreshold_TieBreaksTowardWiderBand` builds two equal-F1 bands of 0.20 and 0.05 width and asserts the 0.20 band's midpoint wins (the old margin rule picked the narrow one). Two more selector regressions cover the separable rounding and the sweep ignoring the shipped value.

Xunzhuo (commented):
Thanks for the careful fixes. Three calibration-gate gaps remain on this head.

src/semantic-router/cmd/image-routing-calibration/testdata/calibration-set.json
```diff
@@ -0,0 +1,75 @@
+{
+  "_comment": "Labelled calibration set for cmd/image-routing-calibration. Every PNG/JPEG under negative_roots at the checked-out commit is a fixture; a fixture is positive for a rule only if listed under positives, otherwise it is a negative for every rule. The set is versioned by the repository commit, which the generated report records together with the model snapshot and every resolved fixture's sha256.",
+  "negative_roots": [
+    "website/static/img",
```
Xunzhuo: **Do not infer negatives from the whole docs image tree**

Impact: Every unlisted image under this root is treated as a negative, but `website/static/img/dashboard/landing.png` visibly contains a large terminal with a curl request and output, matching this rule's own terminal/API-positive descriptions. Counting it as a negative corrupts the confusion matrix and the max-F1 band used to ship `0.4632`.

Required change: Use an explicitly reviewed positive/negative manifest (or correctly label every scanned image), then rerun the pinned calibration and update the thresholds and reported metrics.

.github/workflows/image-routing-calibration.yml
```diff
@@ -0,0 +1,147 @@
...
+      MODEL_DIR: models/mom-embedding-multimodal
+      # Snapshot the shipped thresholds were calibrated on. The download is
+      # pinned to it and the resolved revision is verified before the gate
+      # runs, so the check never validates thresholds against other weights.
+      CALIBRATED_REVISION: fdf8e01b7b0f3a69ac1ac8e2a64dcb1ede177ba4
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.sha }}
```
Xunzhuo, line 46: **Bind the gate to the pull-request head**

Impact: `workflow_call` declares no `ref` input and `pr.yml` passes none, so a pull-request run falls back to `github.sha`, which is GitHub's synthetic merge commit rather than the PR head. The generated `repo_commit` therefore cannot provide the promised exact-head calibration provenance.

Required change: Declare a reusable-workflow ref input, pass `github.event.pull_request.head.sha` from the PR caller, and assert the checked-out commit before producing the report.

tools/agent/domains.yaml
```diff
@@ -105,6 +109,37 @@ domains:
...
+      - candle-binding/src/model_architectures/embedding/**
+      - candle-binding/src/model_architectures/attention/**
+      - candle-binding/src/model_architectures/model_factory.rs
+      - candle-binding/src/model_architectures/traits.rs
+      - candle-binding/src/core/**
+      - candle-binding/src/ffi/embedding.rs
+      - candle-binding/semantic-router.go
+      - candle-binding/Cargo.toml
+      - candle-binding/Cargo.lock
+      - src/semantic-router/pkg/classification/embedding_classifier*.go
```
Xunzhuo: **Include the candidate-embedding initializer in this domain**

Impact: `classifier_signal_group_similarity.go` does not match either classifier glob here, yet it owns `ensureCandidateEmbeddings`, which populates candidate embeddings and rebuilds prototype banks before scoring. A scoring change in that file can therefore skip `image-calibration` and let stale thresholds pass PR Gate.

Required change: Add that path (or a covering classifier pattern) to this domain and add a change-classifier regression proving it selects `image-calibration`.

src/semantic-router/cmd/image-routing-calibration/testdata/calibration-set.json
lawin-steve: Done in 29327483. The set is now an explicit manifest: I reviewed all 231 images and each is listed as a positive for a named candidate, a negative, or excluded with a reason. Nothing is inferred from directory contents any more, and the tool rejects a path under two labels, an exclusion without a reason, or a listed path missing on disk. `dashboard/landing.png` is a positive (terminal window with a curl command and output), and the review found a second one, `blog/vllm/semantic-router/amd-2.png` (curl request in a code window, scores 0.35). Two config-builder screenshots with YAML panes are excluded as ambiguous. Rerunning at the pinned snapshot moves `code_or_terminal_imagery` from 0.4632 to 0.47 (max-F1 band midpoint, F1 0.444, 8 FP / 7 FN over 13 positives); the other two rules are unchanged. The CI gate on this head reports the same numbers, and the PR description tables are updated.

.github/workflows/image-routing-calibration.yml
lawin-steve: Done in d8f57369. `workflow_call` declares a required `ref` input, `pr.yml` passes `github.event.pull_request.head.sha` and `main.yml` passes `github.sha`, and a step after checkout asserts `git rev-parse HEAD` equals the requested commit before the report is produced. The gate run on this head logs `calibrating exact commit 29327483…` and the report's `repo_commit` matches.

tools/agent/domains.yaml
lawin-steve: Done in f0be96d1. `classifier_signal_group_similarity.go` is in the `image-calibration` domain and the change-classifier regression now asserts it selects the gate.

src/semantic-router/cmd/image-routing-calibration/main.go
```diff
@@ -0,0 +1,730 @@
...
+// parseSet decodes and validates a manifest: at least one positive and one
+// negative, every positive naming a known rule, every path decodable by the
+// FFI, and no path under two labels (a positive may be listed once per rule).
+func parseSet(data []byte, rules []config.EmbeddingRule) (calibrationSet, error) {
+	var set calibrationSet
+	if err := json.Unmarshal(data, &set); err != nil {
+		return set, fmt.Errorf("parse: %w", err)
+	}
+	if len(set.Positives) == 0 || len(set.Negatives) == 0 {
+		return set, fmt.Errorf("needs at least one positive and one negative")
```
Xunzhuo, line 402: **Require positive coverage for every rule**

Impact: This validates positives only across the manifest as a whole. A loaded rule with no positive fixture reaches calibration with the `minPositive=2` sentinel, can be reported as perfectly separable with zero false positives/negatives, and lets `-check` accept an uncalibrated shipped threshold.

Required change: Require at least one reviewed positive for every loaded rule and add a regression where one known rule has none.

src/semantic-router/cmd/image-routing-calibration/main.go
```diff
@@ -0,0 +1,730 @@
...
+	report := calibrationReport{}
+	report.Model.Repository = modelRepository
+	report.Model.ArtifactSHA = *artifactRevision
+	report.Model.TargetDimension = 384
+	report.Model.ModelType = "multimodal"
+	report.Model.Aggregation = "max"
+	report.Model.Scoring = config.PrototypeScoringConfig{}.WithDefaults()
+	// The manifest is part of what the commit versions, so it joins the
+	// fixtures in the dirty check.
+	report.Source.Commit, report.Source.Dirty = repoState(*fixtureRoot, append(append([]string{}, fixtures...), absolutePath(*casesPath)))
```
Xunzhuo: **Make the provenance claim cover every input**

Impact: `repoState` receives only the scored fixtures and manifest. Dirty rule YAML, excluded fixtures, or calibration/scoring source can therefore change the result while the report still records `dirty=false` at a commit that cannot reproduce it; excluded assets also have no content hash in the report.

Required change: Cover the complete worktree or every behavior/provenance input (excluding only generated outputs), and record hashes for excluded fixtures as well.

tools/agent/domains.yaml
```diff
@@ -105,6 +109,38 @@ domains:
...
+      - src/semantic-router/pkg/classification/classifier_signal_embedding*.go
+      - src/semantic-router/pkg/classification/classifier_signal_group_similarity.go
+      - src/semantic-router/pkg/classification/prototype_*.go
+      - src/semantic-router/pkg/classification/request_image_embedding_cache.go
+      - src/semantic-router/pkg/config/prototype_scoring_config.go
+      - src/semantic-router/pkg/config/model_config_types.go
+      - src/semantic-router/pkg/config/canonical_defaults.go
```
Xunzhuo, line 127: **Trigger calibration on the direct scoring contract**

Impact: `signal_config.go` defines `EmbeddingRule`, `AggregationMethodMax`, and `EffectiveQueryModality`, which decide which image rule and aggregation semantics are scored, but that file is absent from this domain. A PR can change those semantics while PR Gate skips the required calibration job.

Required change: Add `src/semantic-router/pkg/config/signal_config.go` (or a narrowly equivalent pattern) to the calibration domain and cover the selection with a classifier regression.

Xunzhuo (commented):
Thanks for the thorough update. The previous three blockers are resolved, but three calibration-gate gaps remain on this head.

src/semantic-router/cmd/image-routing-calibration/main.go
lawin-steve: Done in eda9200f. parseSet now requires at least one reviewed positive for every loaded rule and fails otherwise; TestParseSet_RequiresAPositiveForEveryRule covers a known rule with none. Related hardening in e39a7dae: the tool refuses rules it cannot score (query_modality other than image, no candidates) before initialising the model, and requires exactly one finite score per rule on every fixture, so a rule the classifier skipped can no longer count as 0 and read as a confident negative. Regressions cover missing, duplicate, unknown, and non-finite scores.

src/semantic-router/cmd/image-routing-calibration/main.go
lawin-steve: Done in eda9200f. The dirty check now covers the whole worktree at the fixture root (tracked and untracked), not just the scored fixtures and manifest, so uncommitted rule YAML, excluded assets, or scoring code flag the report as not reproducible. Excluded fixtures are recorded with their sha256 next to the reason. -check gained -require-clean, which the CI gate passes so a dirty checkout fails instead of warning; the workflow writes its reports under RUNNER_TEMP so they cannot dirty the checkout. The gate on a5fc1b78 reports dirty=false at the exact PR head.

tools/agent/domains.yaml
lawin-steve: Done in c9ea6d6c and a5fc1b78. signal_config.go is in the domain, and the domain is now scoped to the packages that implement scoring rather than a per-file list: candle-binding/**, the whole classification package, the config contracts the tool decodes and the classifier consumes (canonical_config.go, embedding_config.go, signal_config.go, prototype_scoring_config.go, model_config_types.go, canonical_defaults.go), the router runtime, go.mod/go.sum, and the build entrypoints (Makefile, tools/make/rust.mk). The classifier regression pins each of those, and the negative case for an unrelated extproc change still holds.

src/semantic-router/cmd/image-routing-calibration/main.go
```diff
@@ -0,0 +1,814 @@
...
+		for _, name := range fixture.PositiveFor {
+			if name == rule.Name {
+				positive[fixture.Path] = true
+			}
+		}
+	}
+	// The sweep is built only from fixture scores (plus 0), never from the
+	// threshold currently shipped, so the recommendation cannot depend on the
+	// value being calibrated. The shipped value is evaluated separately below.
+	values := map[float64]bool{0: true}
```
Xunzhuo: **Keep the width tie-break over real score bands**

Impact: Seeding the sweep with `0` splits a band that straddles zero even when no fixture scored zero, so the width tie-break can select a different confusion matrix than the documented widest-band rule. For example, scores `N=-0.25, P=0.25, N=0.30, N=0.40, P=0.775, P=0.80, N=0.85` give equal-F1 bands `(-0.25, 0.25]` (width 0.50) and `(0.40, 0.775]` (width 0.375); this synthetic boundary splits the true winner into two 0.25 bands and selects the later matrix.

Required change: Build the bands from the distinct fixture scores without fragmenting equivalent intervals, and add a straddling-zero regression for the tie-break.

src/semantic-router/cmd/image-routing-calibration/main.go
```diff
@@ -0,0 +1,814 @@
...
+}
+
+// enumerateFixtures returns the scored fixtures (positives and negatives) as
+// repo-relative paths in sorted order. Every listed path, excluded ones
+// included, must exist on disk so a renamed or deleted asset cannot silently
+// drop a reviewed label.
+func enumerateFixtures(root string, set calibrationSet) []string {
+	seen := map[string]bool{}
+	require := func(path, kind string) {
+		if _, err := os.Stat(filepath.Join(root, path)); err != nil {
```
Xunzhuo: **Bind every fixture to the recorded commit**

Impact: `Join` plus `Stat` accepts `..` escapes and follows symlinks, while ordinary `git status` ignores ignored files. A manifest can therefore score an outside, symlinked, or ignored image and still report `repo_dirty=false` for a commit that cannot reproduce that input.

Required change: Canonicalize each input beneath the repository root and require it to be a tracked regular file from the recorded commit (applying the same commit-bound check to the manifest and rules inputs).

Xunzhuo (commented):
Thanks for closing the previous gaps. I found two remaining selector and provenance issues on the exact head.

src/semantic-router/cmd/image-routing-calibration/main.go
lawin-steve: Done. The sweep is built only from the distinct fixture scores; the synthetic 0 is gone, so a band that straddles zero keeps its full width in the tie-break. The lowest real score is the boundary candidate, so floor-scored and single-score sets still report a real matrix. TestSelectThreshold_BandStraddlingZeroKeepsItsWidth pins your example (N=-0.25, P=0.25, N=0.30, N=0.40, P=0.775, P=0.80, N=0.85): the (-0.25, 0.25] band wins on width with TP 3 / FP 3 / FN 0 at threshold 0, and the test also asserts no 0 appears in the sweep. With the old seeding the same test fails by selecting 0.59 and the TP 2 / FP 1 / FN 1 matrix. The shipped picks are unchanged because every repo fixture scores above 0.

src/semantic-router/cmd/image-routing-calibration/main.go
lawin-steve: Done. Every input the report attributes to the commit (fixtures, excluded assets, manifest, rules) must now be a canonical repo-relative path with no .. or symlinks, a regular file, and tracked at HEAD via git ls-tree, so untracked, ignored, or outside files are rejected outright; the manifest and rules paths are resolved through symlinks before any lexical cleaning. Covered by TestBindToCommit_RequiresTrackedRegularFiles and TestResolveInput_ResolvesSymlinksBeforeCleaning, and all 231 listed fixtures plus the manifest and rules bind to HEAD.

lawin-steve:
@Xunzhuo - Thank you for your review , I have addressed your comments. Could you take another look at this PR?

Xunzhuo (dismissed):
The calibration report now reproduces all three shipped thresholds from a clean checkout of this commit, with the model revision and file digests recorded. The reviewed-head multimodal routing CI passed all nine cases. The report also correctly retains the code/terminal rule's non-separable result (8 false positives and 7 false negatives), so the calibration's limits are explicit. The earlier threshold-selection and input-provenance issues are resolved.

wilsonwu (dismissed):
LGTM

lawin-steve:
@Xunzhuo @wilsonwu Mergify could not update this branch, so I merged main manually in ad9e1e94. The merge kept your approvals, but it pulled in two things the gate had to adapt to: the new onnx-binding, which the router module now links, so the calibration tool could not start until it was on the gate's runtime library path; and the per-rule `prototype_scoring` override, which the tool now refuses so the report never describes a blend the run did not use, with tests pinning that no shipped rule carries one. That is f1818aa9. No threshold, manifest, or E2E case change; the gate passes on the merged tree with identical picks and fixture scores. The fix commit dismissed the approvals under branch protection, so a re-approval is needed once CI is green. Sorry for the extra round.

Xunzhuo (approved):
The ONNX setup and per-rule scoring checks are covered. Current-head CPU calibration reproduces the shipped thresholds on 229 scored fixtures, and the deployed multimodal suite passes all nine image signal and decision assertions. The code/terminal threshold’s residual confusion is documented. I found no remaining blocker. GPU behavior was not tested.

src/semantic-router/cmd/image-routing-calibration/main_test.go
```diff
@@ -0,0 +1,451 @@
+package main
```
Xunzhuo, line 1: @lawin-steve as a follow up, please move the dir into tools cmd, Thanks

lawin-steve: @Xunzhuo - I will create a follow up PR. Thank you.

Issue #3515, opened 2026-09-05
[Bug] Dashboard collapsible headers and context menus are not keyboard reachable


yuki-uix:
### Proposed Workgroup

Developer Experience & Ecosystem

### Describe the bug

Fourteen interactive controls in dashboard/frontend/src are plain <div onClick> elements carrying no tabIndex, no key handler, and no ARIA role. They cannot be reached or activated with a keyboard, and assistive technology does not announce them as interactive.

The Dashboard otherwise takes accessibility seriously — the tree carries 740 aria-* and 269 role= attributes, and the same files use real <button> elements a few lines away. These fourteen look like drift rather than a decision, and nothing currently prevents the next one: dashboard/frontend/eslint.config.js does not include eslint-plugin-jsx-a11y, so click-events-have-key-events and no-static-element-interactions never run.

Collapsible section headers (7). A keyboard user cannot expand or collapse these regions at all:

- src/components/ChatComponentWebToolCards.tsx:39
- src/components/ChatComponentWebToolCards.tsx:167
- src/components/ExpressionBuilderToolbox.tsx:30
- src/pages/builderPageDashboardViews.tsx:29
- src/pages/builderPageValidationPanel.tsx:32
- src/pages/builderPageGlobalSettingsSafetySection.tsx:30
- src/pages/topology/components/CustomNodes/DecisionNode.tsx:99

Expression builder context-menu items (7). Add-child, wrap AND/OR/NOT, unwrap, and change-operator are keyboard-unreachable, so nested decision rules cannot be edited without a mouse:

- src/components/ExpressionBuilderContextMenu.tsx:60,64,67,70,101,108,113

Two further hits at src/pages/topology/components/ResultCard/ResultCard.tsx:35-36 are the conventional modal-backdrop click target. They are listed for completeness and are not part of the proposed fix.

### To Reproduce

Verified statically against current main. From dashboard/frontend/src:

`grep -rn '<div[^>]*onClick' --include='*.tsx' . | grep -v '\.test\.'`

Twenty sites are reported. For each of the fourteen listed above, the element's own tag carries no tabIndex, no onKeyDown / onKeyUp / onKeyPress, and no role, so no keyboard event can reach the handler.

[...]

PR #3522, opened 2026-09-06
[Dashboard] Make collapsible headers keyboard operable

yuki-uix:
Make the Dashboard's collapsible section headers operable from the keyboard, and add the lint gate that keeps the class closed.

Related #3515

## What changed

Seven collapsible headers were plain `<div onClick>` elements with no `tabIndex`, key handler, or ARIA role, so keyboard users could not expand or collapse those regions at all. Each is now a `<button type="button">` carrying `aria-expanded` and `aria-controls`, following the pattern already used by `MCPAvailableToolsSection`, `MCPServersSection`, and `CollapsibleSection`:

- `ChatComponentWebToolCards.tsx` — web search card and web page card
- `ExpressionBuilderToolbox.tsx` — toolbox header
- `builderPageDashboardViews.tsx` — sidebar section header
- `builderPageValidationPanel.tsx` — validation panel header
- `builderPageGlobalSettingsSafetySection.tsx` — safety section header
- `topology/components/CustomNodes/DecisionNode.tsx` — rules header

The CSS modules gain the button resets that `.webSearchHeader` already carried. The global `button` reset and `:focus-visible` rule in `index.css` supply the rest, so no per-class focus styling was needed.

`eslint-plugin-jsx-a11y` is added with `click-events-have-key-events` and `no-static-element-interactions` at `"warn"`. `make dashboard-check` already runs `dashboard-lint` and `dashboard-test.yml` already runs `make dashboard-check`, so the rules enter CI without a new workflow or job.

A contract test in the existing source-assertion style pins each header to a button with an expanded state and a controlled region.

## Corrected scope from the issue

The issue reports fourteen affected sites. That number came from a single-line `grep`, which misses JSX where the attribute sits on its own line. **Measured against `upstream/main` with the new rules enabled, there are 73 sites across 32 files.** This PR clears seven of them and leaves 66 across 29 files reported as warnings.

The undercount is itself the argument for the gate: manual review found 19% of the class.

Proposed path to `error`:

1. This PR — seven collapsible headers, rules at `warn`
2. Next — the expression builder context menu (`role="menu"` / `role="menuitem"` with arrow-key, Enter, and Escape handling), which is the other half of #3515
3. Then — the remaining sites in batches, grouped by surface
4. Finally — promote both rules to `error`

I am happy to keep taking these batches, or to leave the remainder to the Workgroup after #3515 closes.

## Decisions worth reviewing

**No E2E case was added, and I would like guidance on this.** `tools/agent/docs/feature-complete-checklist.md` asks for at least one E2E case when user-visible behavior changes, and keyboard operability is user-visible. There is no existing Playwright spec covering the Builder or Topology surfaces, and the suite needs the running stack, so adding one is a larger change than the fix itself. The contract test is not a substitute — it asserts source shape, not behavior. I verified it is not merely decorative by mutating three sites (reverting a header to `<div onClick>`, deleting an `aria-controls`, and inverting an `aria-expanded`); each mutation failed the suite. If the Workgroup wants a real E2E case before this merges, say so and I will add one — `claw-room-collaboration.spec.ts` already has the `toBeFocused()` / `keyboard.press()` idiom to follow.

**`builderPageDashboardViews` changes which area toggles the section.** That header contained a nested add button, so wrapping it in a toggle button would have nested one button inside another. The toggle is now a sibling covering the title region, which also removes the `stopPropagation` the nesting required. The chevron and the header padding are no longer toggle targets. Please read this file with `git diff -w` — converting the implicit-return arrow to a block body so it can call `useId` re-indents 52 of the 83 changed lines.

**`DecisionNode` declares its expanded state conditionally.** The rules region is gated on both the collapsed state and whether any preview condition exists, so the header sets `aria-expanded` and `aria-controls` only when there is something to expand, as `MCPServersSection` does for servers with no tools. One consequence I did not resolve: for a decision node with no conditions, the header is now a focusable button that toggles an icon but reveals nothing. `disabled={!hasRuleDetail}` would match `MCPServersSection` more closely — I left it out because it also changes mouse behavior, but I will add it if you prefer.

## Verification

- `npm run type-check` — passes
- `npm run lint` — 0 errors, 108 warnings, all from the two new rules
- `npm run test:unit` — 172 files, 734 tests pass
- Mutation-checked the new contract test as described above

Not verified: whether the `DecisionNode` button interacts with ReactFlow's drag handling. The element occupies the same position and event path as the `<div>` it replaces, so I expect no change, but I have not exercised it in a running canvas.


FAUST-BENCHOU (commented):
would love to see screenshot after change and before change

yuki-uix:
Good call 😊 — this is much easier to judge with the frames side by side than from the diff, and putting them together surfaced a detail I should have stated in the description from the start.

## What the frames show

Both are the Builder page at the same viewport, after the same two actions: click the already-selected `Global catalog` scope chip, then press <kbd>Tab</kbd> twice.

**Before** — focus skips the `MODELS` section header and lands on the small `+` button beside it. The header is a `<div onClick>`, so it is not in the tab order at all: there is no keystroke that reaches it.

a11y-before

**After** — the same two keypresses put focus on the whole `MODELS 4` header row. It is now a `<button type="button">` with `aria-expanded` bound to the section state and `aria-controls` naming the list it toggles, so <kbd>Enter</kbd> and <kbd>Space</kbd> collapse and expand it.

a11y-after

### Reading the two frames without tripping over the `+` button

This pair is easy to misread, so it is worth spelling out: the `+` add button is focusable in **both** states and is not what changed. What changed is that the header joins the tab order *ahead* of it.

| Tab stop | Before | After |
| --- | --- | --- |
| 1st | `privacy-first` scope chip | `privacy-first` scope chip |
| **2nd** | **`+` add button** — the header is skipped entirely | **`MODELS 4` header** — newly reachable |
| 3rd | first model in the list | **`+` add button** — shifted one position later |

So "after" showing focus on the title rather than on `+` is the fix working, not a regression: the second keypress now stops at the control it should have stopped at all along, and `+` is still reachable one Tab further on. Nothing was removed from the tab order.

The same relationship shows up in the totals: this page has **35** visible focusable elements before the change and **42** after. The seven added are exactly the seven sidebar section headers. Had the change traded one control for another, that number would not have moved.

## The resting state is unchanged, and that is the part worth checking

I should have said this up front: the two frames are near-identical by design, so the useful thing to review in them is that nothing moved, not that something did.

The CSS modules gained only the button resets that `.webSearchHeader` already carried (`width`, `color`, `font`, `text-align`); the global `button` reset and the `:focus-visible` rule in `index.css` supply the rest. No spacing, color, or layout value was touched. The one structural change is in `builderPageDashboardViews.tsx`, where the previous markup nested the add button inside the clickable header — the toggle is now a sibling covering the title region, which is why `sidebarSectionToggle` exists. A side effect I should flag: the chevron and the header padding are no longer toggle targets for the mouse.

If you would rather see a different surface than the Builder sidebar, I am happy to capture the validation panel or the topology decision node instead.

## What the defect costs, scoped honestly

These seven sections default to expanded (`BuilderPage.tsx:75-83` initialises all of them to `true`), so a keyboard-only or screen-reader user can still read the content. What they cannot do is collapse a section and expand it again, and assistive technology does not announce the row as interactive at all, since a `<div>` with a click handler carries no role and no state.

The heavier case is the other half of #3515, the expression builder context menu, where the affected items are the only way to edit nested decision rules. I left that out of this PR rather than bundling two different kinds of work, but I am glad to take it next if that ordering suits you.

The validation panel header at the bottom of the same screenshot is a second of the seven fixed headers, and behaves identically.

## On keeping the lint rule in the same PR

I would rather not land the fixes without the gate, and the reason is a mistake of my own: the issue originally reported fourteen affected sites, but that count came from a single-line `grep` that missed every case where the attribute sits on its own line. Measured against `main` with the rules enabled there are **73 sites across 32 files** — my careful manual pass over this exact question found 19% of them. That seems like decent evidence that review alone will not hold this line.

`make dashboard-check` already runs `dashboard-lint`, and `dashboard-test.yml` already runs `make dashboard-check`, so the two rules enter CI without a new workflow. They are at `warn`, so the 66 remaining sites block nothing; I would like to clear them in batches and then promote to `error`, but I am happy to follow whatever sequencing the Workgroup prefers.

Xunzhuo (approved):

FAUST-BENCHOU (approved):

wilsonwu (approved):
LGTM

Issue #3188, opened 2026-08-31
[Feature] Add config validation, diff, and dry-run diagnostics


Xunzhuo:
## Goal

Coordinate one canonical, side-effect-free validation, structured diff, and dry-run contract so maintained configuration producers receive equivalent diagnostics before activation.

## Proposed Workgroup

Enterprise & Environment

## Background

Current `main` already exposes `POST /config/router/validate`. Router, CLI, Dashboard, Helm, and Operator inputs converge on the same runtime configuration, but they do not yet share one complete structured diff and dry-run contract. Composite recipes also add reference graphs, role capabilities, quorum, fallback, and budget conditions that must fail before activation with actionable paths.

## Work item shape

This is a parent delivery issue under #2326, not a single claimable implementation. #3477 is the first leaf and owns one canonical dry-run/diff engine plus Router API using current validation. CLI and Dashboard consumers follow after that engine is stable. Kubernetes and Helm rollout ownership remains #2355; version and unknown-field policy remains #2469.

## Scope

- Validate a complete candidate snapshot without changing desired, active, runtime, persistence, or version-history state.
- Return versioned error and warning codes, severity, compatibility information, and exact resource, recipe, stage, and field paths.
- Produce a bounded structured active-versus-candidate diff with schema-aware redaction.
- Compose normalization, reference resolution, precondition checks, validation, and diff through one canonical implementation.
- Cover composite references, cycles, virtual targets, panel and quorum limits, per-role capabilities, aggregate budget and pricing requirements, readiness, and fallback targets.
- Maintain one cross-producer golden corpus for success, failure, unknown fields, version mismatch, conflicts, and redaction.

## Configuration and API impact

The first leaf adds a read-only Router diagnostic API; later consumers must reuse it or the same canonical engine. No path in this parent may activate, persist, rewrite, or roll back configuration.

## Non-goals

- Add a second schema or producer-specific validation rules.
- Expose raw secrets in diagnostics.
- Own organization, tenant, virtual-key, quota, rate-limit, billing, or settlement policy.
- Treat existing validation as absent or reimplement it beside the canonical path.

## Completion signal

- [ ] #3477 provides the canonical side-effect-free engine and Router API.
- [ ] Structured paths and codes are versioned, bounded, and schema-redacted.
- [ ] Maintained producers return equivalent outcomes for the golden corpus as their consumers land.
- [ ] Composite failures identify the exact recipe, stage, reference, and unmet invariant.
[...]

PR #3523, opened 2026-09-06
[Feature] Add a canonical side-effect-free config diff and dry-run Router API

### LINKED ISSUE #3477: [Feature] Add a canonical side-effect-free config diff and dry-run Router API
## Goal

Extend `POST /config/router/validate` additively so one canonical, side-effect-free path validates a candidate configuration, resolves references, and returns structured diagnostics and a redacted diff against the active snapshot.

## Proposed Workgroup

Enterprise & Environment

## Background

Current `main` already provides `POST /config/router/validate` with canonical parsing, normalization, secret redaction, request limits, and the existing authorization boundary. What remains is a structured diff and field-addressable diagnostic contract that maintained producers can consume without creating another validator or API.

Parent #3188 is assigned and coordinates this first leaf; implementation ownership should be agreed there before coding.

## Scope

- Extend the existing endpoint and response schema compatibly; do not add a sibling validation or dry-run route.
- Reuse current canonical parsing, normalization, reference resolution, validation, request-size limits, authorization, and secret redaction.
- Add versioned error and warning codes with exact resource, recipe, stage, and field paths.
- Produce a bounded, schema-aware, redacted active-versus-candidate structured diff.
- Prove that the endpoint cannot mutate desired config, active snapshot, runtime state, persistence, or version history.
- Cover composite references, cycles, capabilities, quorum, budgets, fallback, unknown fields, conflicts, and redaction in one golden corpus.
- Keep CLI and Dashboard presentation as later consumers of this contract.

This is an additive response evolution of the existing authenticated endpoint. Existing valid callers keep working; new fields are bounded and versioned. No path activates, persists, rolls back, or rewrites configuration.

## Non-goals

- Add a sibling API, second schema, or producer-specific validation rules.
- Build CLI or Dashboard presentation in this leaf.
- Own Kubernetes/Helm rollout behavior from #2355 or version/unknown-field policy from #2469.
- Expose raw secrets or bypass existing request and authorization limits.

## Completion signal

- [ ] Existing `POST /config/router/validate` returns backward-compatible structured diagnostics and diff.
- [ ] Validation, reference resolution, and diff use one canonical implementation.
- [ ] Paths, codes, bounds, authorization, request limits, and redaction have golden coverage.
- [ ] Tests prove every validation/diff path is side-effect free.
- [ ] Parent #3188 records the follow-on CLI and Dashboard consumer boundaries.

sujal111:
Closes #3477
Related #3188

## Purpose

Extend `POST /config/router/validate` additively with the v1 diagnostic contract from #3188. Later CLI and Dashboard work should call this API instead of adding another validator.

`config.Evaluate` owns parse, unknown-field warnings, semantic validation, redaction, and the optional source-vs-candidate diff. The handler does not write config, backups, or runtime state. Invalid candidates return HTTP 200 with `valid: false` and field-addressable errors. `compare_to_active` adds a bounded (500), redacted diff. Secrets stay `[REDACTED]`; `*_env` names stay visible. Auth remains `config.read`.

## Test Plan

- [x] `go test ./pkg/config/ -run 'TestEvaluate|TestDiffCanonical'`
- [x] `go test ./pkg/apiserver/ -run 'TestHandleConfigValidate|TestConfigValidateRoute'`
- [x] `go test ./...` in `src/semantic-router`
- [x] `make agent-lint` on the changed files
- [x] `make docs-check-translation-coverage`
- [x] `make build-e2e`
- [x] `make recipe-conformance-static`
- [ ] Live `make e2e-test E2E_PROFILE=envoy-ai-gateway` for `apiserver-config-validate`

## Test Result

The listed unit, HTTP, lint, docs, e2e-compile, and recipe-conformance checks passed. The new `apiserver-config-validate` baseline case is registered but was not run against a live cluster.

### DIFF
--- a/e2e/pkg/testmatrix/testcases.go
+++ b/e2e/pkg/testmatrix/testcases.go
@@ -14,6 +14,7 @@ var BaselineRouterContract = []string{
 	"anthropic-messages-streaming",
 	"apiserver-runtime-config-endpoints",
 	"apiserver-classification-endpoints",
+	"apiserver-config-validate",
 	"llm-classifier-distribution-routing",
 	"sequence-classifier-routing",
 	"chat-completions-stress-request",
--- a/src/semantic-router/pkg/config/config.go
+++ b/src/semantic-router/pkg/config/config.go
@@ -146,6 +146,10 @@ type RouterConfig struct {
 	// runtime snapshot was parsed. Management APIs use it to distinguish a
 	// persisted config from the config that has completed hot reload.
 	DocumentHash string `yaml:"-"`
+	// SourceDocument is the exact YAML bytes that produced this snapshot.
+	// Compare-to-active binds to this verified in-memory document instead of a
+	// later desired-source write.
+	SourceDocument []byte `yaml:"-"`
 	// EffectiveModelRegistry is the immutable catalog/config join used to
 	// materialize this runtime snapshot.
 	EffectiveModelRegistry *modelcatalog.EffectiveRegistry `yaml:"-" json:"-"`
--- a/src/semantic-router/pkg/config/config.go
+++ b/src/semantic-router/pkg/config/config.go
@@ -141,15 +141,16 @@ type RouterConfig struct {
 
 	// Runtime-only knowledge bases loaded from global.model_catalog.
 	KnowledgeBases []KnowledgeBaseConfig `yaml:"knowledge_bases,omitempty"`
-	ConfigBaseDir  string                `yaml:"-"`
+	ConfigBaseDir  string                `yaml:"-" json:"-"`
 	// DocumentHash identifies the exact YAML document from which this immutable
 	// runtime snapshot was parsed. Management APIs use it to distinguish a
 	// persisted config from the config that has completed hot reload.
-	DocumentHash string `yaml:"-"`
+	DocumentHash string `yaml:"-" json:"-"`
 	// SourceDocument is the exact YAML bytes that produced this snapshot.
 	// Compare-to-active binds to this verified in-memory document instead of a
-	// later desired-source write.
-	SourceDocument []byte `yaml:"-"`
+	// later desired-source write. It stays out of config JSON so inventory
+	// dumps cannot expose the original document, including plaintext secrets.
+	SourceDocument []byte `yaml:"-" json:"-"`
 	// EffectiveModelRegistry is the immutable catalog/config join used to
 	// materialize this runtime snapshot.
 	EffectiveModelRegistry *modelcatalog.EffectiveRegistry `yaml:"-" json:"-"`
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,7 +75,7 @@ repos:
         entry: .venv-agent/bin/yamllint --config-file=tools/linter/yaml/.yamllint
         language: system
         files: \.(yaml|yml)$
-        exclude: ^(node_modules/)
+        exclude: ^(node_modules/|src/semantic-router/pkg/config/testdata/dryrun/parse-error\.yaml$)
 
   # JavaScript and TypeScript specific hooks
   - repo: local
--- a/e2e/testcases/apiserver_config_validate.go
+++ b/e2e/testcases/apiserver_config_validate.go
@@ -16,7 +16,7 @@ import (
 
 func init() {
 	pkgtestcases.Register("apiserver-config-validate", pkgtestcases.TestCase{
-		Description: "Verify POST /config/router/validate returns the v1 diagnostic contract without mutating the live config (issue #3477)",
+		Description: "Verify POST /api/v1/config/validate returns the v1 diagnostic contract without mutating the live config (issue #3477)",
 		Tags:        []string{"apiserver", "config", "api", "validate"},
 		Fn:          testAPIServerConfigValidate,
 	})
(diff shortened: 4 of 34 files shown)


Xunzhuo (changes requested):
Welcome, and thanks for the careful diagnostic work. Two blockers remain. `RedactSensitiveConfigValue` does not treat the canonical `auth_token` field as secret, so a `config.read` caller can receive the active bearer token in the diff’s `old` value. Also, `activeConfigSnapshotYAML` reads the persisted source path rather than the generated runtime or verified in-memory active state, so `compare_to_active` can compare against desired source after an override or failed activation. Please cover the complete schema secret contract and bind this comparison to actual active runtime state, with regressions for both cases.

src/semantic-router/pkg/config/config.go
```diff
@@ -146,6 +146,10 @@ type RouterConfig struct {
 	// runtime snapshot was parsed. Management APIs use it to distinguish a
 	// persisted config from the config that has completed hot reload.
 	DocumentHash string `yaml:"-"`
+	// SourceDocument is the exact YAML bytes that produced this snapshot.
+	// Compare-to-active binds to this verified in-memory document instead of a
+	// later desired-source write.
+	SourceDocument []byte `yaml:"-"`
```
Xunzhuo: **Keep the raw source snapshot out of config JSON**

Impact: `SourceDocument` holds the exact original YAML, including plaintext secrets, but this field lacks `json:"-"`. `/api/v1/inventory/classifier` reflects the live `RouterConfig` through `jsonCompatibleValue`, which skips only `json:"-"` fields; the later key-based redactor leaves these raw bytes untouched, and JSON exposes them as base64. A caller with `config.read` but without `secret_view` can therefore recover the full source configuration.

Required change: Mark `SourceDocument` as `json:"-"` (and keep other runtime-only metadata consistently excluded), then add an inventory-endpoint regression built from a parsed config containing a secret and assert neither the canary nor encoded source bytes are returned.

Xunzhuo (commented):
Thanks for binding compare-to-active to the live runtime and covering `auth_token`. The new snapshot must stay out of existing config JSON.

src/semantic-router/pkg/config/config.go
sujal111: Addressed in b0c27e813.

`SourceDocument`, `DocumentHash`, and `ConfigBaseDir` are now `json:"-"`. `TestClassifierInfoOmitsParsedSourceDocument` parses a config that contains a secret and asserts `/api/v1/inventory/classifier` returns neither the canary nor the base64-encoded source bytes.

sujal111:
Merged latest `main` to clear the conflict in `src/semantic-router/pkg/apiserver/route_router_config_validate_test.go`. Kept the validate-API coverage and took `main`'s renamed `TestValidateHotReloadCompatibilityAllowsOwnedLocalClassifierChange`.

Earlier review items (redact `auth_token`, bind `compare_to_active` to live runtime, keep `SourceDocument` out of inventory JSON) are in a30c588 and b0c27e8. @Xunzhuo, please re-review when you have a chance.

sujal111:
@Xunzhuo  could you pls review

Xunzhuo (dismissed):
The previous secret-redaction and active-snapshot issues are addressed. Authenticated HTTP integration confirms that valid, invalid, and compare-to-active requests preserve the live configuration hashes and source file, while excluding active and candidate secrets from responses. The relevant exact-head checks pass.

wilsonwu (dismissed):
LGTM

sujal111:
Merged latest `main` to clear the `routes_catalog.go` conflict that kicked this out of the merge queue.

Resolution: kept `RouterConfigValidateRequest` (`yaml` + `compare_to_active`) and took main's documented `200`/`400`/`500` response contract, then regenerated the OpenAPI artifact. That also brings the workflow updates onto the branch so Mergify does not need `workflows` permission to update it.

@Xunzhuo @wilsonwu the previous approvals were dismissed by the merge commit — please re-approve when you have a chance.

sujal111:
@Xunzhuo  @wilsonwu  resolved conflicts:)

e2e/pkg/testmatrix/testcases.go
```diff
@@ -14,6 +14,7 @@ var BaselineRouterContract = []string{
 	"anthropic-messages-streaming",
 	"apiserver-runtime-config-endpoints",
 	"apiserver-classification-endpoints",
+	"apiserver-config-validate",
```
Xunzhuo, line 17: [P2] Keep the baseline inventory assertion in sync with the new case

Current `main` already includes `security-window-provenance` and expects 39 full / 37 standard cases. The clean merge of this head also retains `apiserver-config-validate`, producing 40 / 38 while `TestBaselineScopesPreserveEveryFunctionalCase` still expects 39 / 37. Running `go test -count=1 ./pkg/testmatrix` on that integration fails with `inventory changed: standard=38 full=40`; main alone passes. Please update the inventory assertion for the combined cases and rerun `make test-e2e-unit`.

Xunzhuo (changes requested):
The current-main integration exposes a deterministic baseline inventory regression: the added case is retained, but the test still expects one fewer case. The focused E2E matrix test fails on the clean merge and passes on main. Please synchronize the assertion and rerun the E2E unit gate. This follow-up was cluster-free; I did not rerun deployed runtime checks.

PR #3533, opened 2026-09-06
[Bug] Skip ONNX weights when downloading candle embedding models

### LINKED ISSUE #3532: [Bug] Runtime model downloader fetches multi-GB ONNX exports the candle embedding runtime never loads
### Proposed Workgroup

### Describe the bug

`src/semantic-router/pkg/modeldownload` invokes `hf download <repo> --local-dir <path>` with no file filter for every missing model. For the default candle embedding backend this downloads the full Hugging Face repository even though the runtime only loads `config.json`, `model.safetensors` and `tokenizer.json` (the completeness contract enforced since #2172 / #2195 / #2531).

For `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` that means ~4.9 GB transferred for a ~650 MB runtime contract: the `onnx/layer-{6,11,16,22}/` directories hold `model.onnx`, `model.onnx.data` and `model_fa_fp16.onnx` (≈ 4.26 GB combined) that candle never opens.

`tools/make/models.mk` already narrows its own perf-benchmark download with `--exclude "onnx/*"`, so the repository currently has two different answers to "which files does a candle model need".

### To Reproduce

1. Configure `embedding_models.mmbert_model_path: models/mmbert-embed-32k-2d-matryoshka` with the default (candle) embedding backend and no pre-downloaded model directory.
2. Start the router (local `vllm-sr` stack, CPU image, commit `63e91c21`) and watch the `hf download` process it spawns during the `checking_models` / `downloading` startup phase.
3. Observe the transfer pulling every `onnx/**` artifact before the router can leave `initializing_models`.

Environment: macOS (Intel) host, Docker Desktop, router image `vllm-project/semantic-router/vllm-sr:latest`, `hf` CLI from `huggingface_hub`.

### Expected behavior

Only the artifacts the configured embedding backend loads are fetched. For candle that is the root `config.json`, `model.safetensors`, `tokenizer.json` plus the small JSON side files; the ONNX weight exports are skipped. OpenVINO deployments, which consume the ONNX exports, keep receiving the full snapshot.

Excluding only weight files (`*.onnx`, `*.onnx.data`, `*.onnx_data`) rather than the whole `onnx/` tree matters: `config.MmBertAvailableLayers` reads `onnx/model_config.json`.

Verified against the live repository file list with `huggingface_hub.utils.filter_repo_objects` (the filter `hf download --exclude` uses): 16 files kept (≈ 648 MB, including `onnx/model_config.json`), 12 dropped (≈ 4,258 MB).

### Impact

- ~5 GB of network transfer per fresh install or CI runner instead of ~650 MB; on a laptop the download was interrupted after several GB and the router stayed in `initializing_models`.
- The unused ONNX variants remain on disk after a successful download.
- Workaround: pre-download the three required files manually with `hf download <repo> model.safetensors tokenizer.json config.json --local-dir models/<path>` before starting the router.

### Additional context

A fix is implemented and tested on a branch and will be submitted as a PR once this issue is accepted: `ModelSpec` gains `ExcludePatterns` forwarded to `hf download --exclude`, populated for the candle embedding model paths (`MmBertModelPath`, `Qwen3ModelPath`, `GemmaModelPath`, `MultiModalModelPath`) only when `EmbeddingModels.EmbeddingBackend() == candle`; OpenVINO, remote and non-embedding models are unchanged. Unit tests cover the argument list, per-backend scope, and a guard that no exclude pattern can match a completeness-required file (which would otherwise cause an endless re-download loop).

Classifier models (category / PII / jailbreak) also ship `onnx/` subtrees but can run on the ONNX classifier backend; they are intentionally out of scope here and can be narrowed in a follow-up once their backend selection is encoded in the downloader.

Related: #2108, #2172, #2195, #2531.

vuhluu:
Closes #3532

## Purpose

The runtime model downloader (`src/semantic-router/pkg/modeldownload`) ran `hf download <repo> --local-dir <path>` with no file filter, so a candle deployment of `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` fetched the whole repository: about 4.3 GB of `onnx/**/model.onnx`, `model.onnx.data` and `model_fa_fp16.onnx` exports on top of the ~650 MB (`config.json`, `model.safetensors`, `tokenizer.json`) the candle runtime actually loads. On a fresh machine that blocks router startup for the whole transfer and can exhaust disk before the embedding model is usable.

This change:

- adds `ModelSpec.ExcludePatterns`, forwarded to `hf download` by a new pure helper `buildDownloadArgs` as **one `--exclude <pattern>` flag per pattern**. The typer-based `hf` CLI takes `--exclude` as a repeatable single-value option (one flag followed by several patterns keeps only the first and treats the rest as positional filenames), while the legacy `huggingface-cli` took `nargs=*`; the repeated form is parsed identically by both, whichever CLI the image ends up installing;
- populates it with `*.onnx`, `*.onnx.data`, `*.onnx_data` for every candle embedding model path (mmBERT, Qwen3, Gemma, multimodal) when `EmbeddingModels.EmbeddingBackend()` is `candle`. The exclude map is keyed and looked up by `config.ResolveModelPath`, so the narrowing holds when the model is configured by a registry alias and does not depend on whether the collected provisioning paths are canonicalized first (#2828 can land before or after this);
- leaves OpenVINO deployments on the full snapshot (they consume the ONNX exports), leaves remote backends unchanged (they provision nothing), and leaves non-embedding models (classifiers, semantic-cache BERT) unchanged to keep the blast radius on the embedding runtime;
- excludes weight files only, so the `onnx/model_config.json` layer manifest read by `config.MmBertAvailableLayers` is still downloaded.

Modules affected: `src/semantic-router/pkg/modeldownload` only (router service platform surface, `router-core` rules). Owner: `wg/router-models-inference-runtime` (accepted issue #3532).

## Test Plan

- `make agent-report ENV=cpu CHANGED_FILES="src/semantic-router/pkg/modeldownload/downloader.go,src/semantic-router/pkg/modeldownload/config_parser.go,src/semantic-router/pkg/modeldownload/types.go,src/semantic-router/pkg/modeldownload/downloader_test.go,src/semantic-router/pkg/modeldownload/download_scope_test.go"` → primary skill `project-change`, fast tests `make test-semantic-router`
- Unit tests in `downloader_test.go` and `download_scope_test.go`:
  - `TestBuildDownloadArgsFetchesFullSnapshotByDefault`: models without a scope keep the historical argument list
  - `TestBuildDownloadArgsRepeatsExcludeFlagPerPattern`: every pattern carries its own `--exclude`, the flag count equals the pattern count, `--exclude` trails `--revision`
  - `TestBuildDownloadArgsSkipsEmptyExcludePatterns`: an empty entry never produces a bare `--exclude`
  - `TestBuildModelSpecsExcludesOnnxWeightsForCandleEmbeddingModels`: all four candle paths get the exclude list
  - `TestBuildModelSpecsExcludesOnnxWeightsForAliasedEmbeddingModel`: `models/mom-embedding-ultra` (alias) and the canonical path are both narrowed; matches specs by resolved `LocalPath`, so it passes with or without #2828
  - `TestBuildModelSpecsKeepsFullSnapshotForOpenVINOBackend`: OpenVINO is unfiltered
  - `TestBuildModelSpecsLeavesNonEmbeddingModelsUnfiltered`: semantic-cache BERT is unfiltered
  - `TestOnnxWeightExcludePatternsNeverMatchCandleRequiredFiles`: exclude globs can never shadow a completeness-required file (would otherwise cause an endless re-download loop)
- `cd src/semantic-router && go test ./pkg/modeldownload/... && go vet ./pkg/modeldownload/...`
- `gofmt -l src/semantic-router/pkg/modeldownload/`
- Exclude globs checked against the live repository file list with `huggingface_hub.utils.filter_repo_objects` (the filter `hf download --exclude` uses)
- End-to-end download check: in a clean `python:3.12-slim` container with unpinned `huggingface_hub` (the same install as `tools/docker/Dockerfile.extproc` and `test-and-build.yml`), run the exact argv `buildDownloadArgs` produced before and after the fix against the live `llm-semantic-router/mmbert-embed-32k-2d-matryoshka` repository into a fresh `--local-dir`, then list the files on disk
- `make agent-ci-gate CHANGED_FILES="..."` (same file list)

## Test Result

- `go test ./pkg/modeldownload/... -v -run 'DownloadArgs|ExcludesOnnx|OpenVINO|Unfiltered|NeverMatch|Aliased'`: 8/8 PASS (macOS, Intel, Go 1.25.0)
- `go test ./pkg/modeldownload/...`: `ok … 1.323s`; `go vet ./pkg/modeldownload/...`: clean; `gofmt -l`: no output
- Exclude globs vs. the repository file list: 16 files kept (including `config.json`, `model.safetensors`, `tokenizer.json`, `onnx/model_config.json` and the per-layer `onnx/*/config.json`), 12 dropped (every `model.onnx`, `model.onnx.data`, `model_fa_fp16.onnx`)
- End-to-end download check, `hf` 1.30.0:
  - **before** (`--exclude "*.onnx" "*.onnx.data" "*.onnx_data"`, the argv of the first revision): `UserWarning: Ignoring '--exclude' since filenames have been explicitly set`; `Fetching 4 files`, 2.12 GB: the directory holds only `onnx/layer-{6,11,16,22}/model.onnx.data`; no `model.safetensors`, no `tokenizer.json`. Reproduces the exact-head CI failure: the stray positionals are treated as filename globs, so the CLI downloads precisely the ONNX weights and nothing the candle runtime needs
  - **after** (`--exclude "*.onnx" --exclude "*.onnx.data" --exclude "*.onnx_data"`): `Fetching 16 files`, 614 MB; `model.safetensors`, `tokenizer.json`, `config.json`, `onnx/model_config.json` and the per-layer `onnx/*/config.json` present; `find` for `*.onnx`, `*.onnx.data`, `*.onnx_data` returns nothing
- `make agent-ci-gate`: passed on the first revision (`fd526d02`: agent-report, pre-commit baseline, Go structural lint, structure and architecture checks); not re-run on `86ce721d`, which only touches the same package's Go sources and tests
- Manual fresh download through the router binary itself: not re-run locally (the checkout already holds the minimal model directory, so the downloader treats it as complete)

Follow-ups tracked outside this PR: classifier repositories also ship `onnx/` subtrees and are intentionally not narrowed here because the ONNX classifier backend may need them (follow-up on #3532); `IsGatedModelError` returning true for any failure when `HF_TOKEN` is empty, which turns a transient public-repo failure into a "gated, skipping" startup (raised by @adaamko in review, to be filed separately).

---
<details>

</details>

### DIFF
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -134,7 +134,7 @@ func BuildModelSpecs(cfg *config.RouterConfig) ([]ModelSpec, error) {
 			RepoID:          repoID,
 			Revision:        "main",
 			RequiredFiles:   requiredFiles,
-			ExcludePatterns: excludePatternsByModel[path],
+			ExcludePatterns: excludePatternsByModel[config.ResolveModelPath(path)],
 		})
 	}
 
@@ -222,6 +222,12 @@ var onnxWeightExcludePatterns = []string{
 // the download exclude globs for artifacts the selected embedding backend never
 // loads. Only the candle backend is narrowed: OpenVINO consumes the ONNX exports
 // and the remote backend provisions no local embedding models.
+//
+// Keys are canonical registry paths (config.ResolveModelPath), matching how the
+// embedding runtime resolves the same fields before loading. Callers look the map
+// up by the resolved path too, so the narrowing holds whether the configured value
+// is the canonical directory or a registry alias, and whether or not the collected
+// provisioning paths have already been canonicalized upstream.
 func candleEmbeddingModelExcludePatterns(cfg *config.RouterConfig) map[string][]string {
 	excluded := make(map[string][]string)
 	if cfg.EmbeddingModels.EmbeddingBackend() != config.EmbeddingBackendCandle {
@@ -229,10 +235,11 @@ func candleEmbeddingModelExcludePatterns(cfg *config.RouterConfig) map[string][]
 	}
 
 	for path := range candleEmbeddingModelRequiredFiles(cfg) {
-		if path == "" || !strings.HasPrefix(path, "models/") {
+		resolved := config.ResolveModelPath(path)
+		if resolved == "" || !strings.HasPrefix(resolved, "models/") {
 			continue
 		}
-		excluded[path] = append([]string(nil), onnxWeightExcludePatterns...)
+		excluded[resolved] = append([]string(nil), onnxWeightExcludePatterns...)
 	}
 	return excluded
 }
--- a/src/semantic-router/pkg/modeldownload/config_parser.go
+++ b/src/semantic-router/pkg/modeldownload/config_parser.go
@@ -101,6 +101,7 @@ func BuildModelSpecs(cfg *config.RouterConfig) ([]ModelSpec, error) {
 	paths := filterDisabledOptionalModelPaths(cfg, extractProvisioningModelPaths(cfg))
 	requiredFilesByModel := ExtractRequiredFilesByModel(cfg)
 	addEmbeddingModelRequiredFiles(cfg, requiredFilesByModel)
+	excludePatternsByModel := candleEmbeddingModelExcludePatterns(cfg)
 
 	// Allow empty paths for API-only configurations
 	if len(paths) == 0 {
@@ -129,10 +130,11 @@ func BuildModelSpecs(cfg *config.RouterConfig) ([]ModelSpec, error) {
 		}
 
 		specs = append(specs, ModelSpec{
-			LocalPath:     path,
-			RepoID:        repoID,
-			Revision:      "main",
-			RequiredFiles: requiredFiles,
+			LocalPath:       path,
+			RepoID:          repoID,
+			Revision:        "main",
+			RequiredFiles:   requiredFiles,
+			ExcludePatterns: excludePatternsByModel[path],
 		})
 	}
 
@@ -202,6 +204,37 @@ func candleEmbeddingModelRequiredFiles(cfg *config.RouterConfig) map[string][]st
 	add(cfg.GemmaModelPath, gemmaDenseWeightFiles)
 	add(cfg.MultiModalModelPath, embeddingModelWeightFiles)
 	return required
+}
+
+// onnxWeightExcludePatterns match the ONNX inference exports published beside the
+// safetensors weights in the embedding model repositories. The candle runtime never
+// opens them, yet they dominate the snapshot size (about 4.3 GB of the 4.9 GB
+// mmbert-embed-32k-2d-matryoshka repository), so a candle deployment skips them at
+// download time. Small manifests such as onnx/model_config.json, which
+// config.MmBertAvailableLayers reads, are not matched and stay in the snapshot.
+var onnxWeightExcludePatterns = []string{
+	"*.onnx",
+	"*.onnx.data",
+	"*.onnx_data",
+}
+
+// candleEmbeddingModelExcludePatterns returns, per configured embedding model path,
+// the download exclude globs for artifacts the selected embedding backend never
+// loads. Only the candle backend is narrowed: OpenVINO consumes the ONNX exports
+// and the remote backend provisions no local embedding models.
+func candleEmbeddingModelExcludePatterns(cfg *config.RouterConfig) map[string][]string {
+	excluded := make(map[string][]string)
+	if cfg.EmbeddingModels.EmbeddingBackend() != config.EmbeddingBackendCandle {
+		return excluded
+	}
+
+	for path := range candleEmbeddingModelRequiredFiles(cfg) {
+		if path == "" || !strings.HasPrefix(path, "models/") {
+			continue
+		}
+		excluded[path] = append([]string(nil), onnxWeightExcludePatterns...)
+	}
+	return excluded
 }
 
 // addEmbeddingModelRequiredFiles marks every configured candle embedding model as
--- a/src/semantic-router/pkg/modeldownload/download_scope_test.go
+++ b/src/semantic-router/pkg/modeldownload/download_scope_test.go
@@ -1,0 +1,103 @@
+package modeldownload
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/vllm-project/semantic-router/src/semantic-router/pkg/config"
+)
+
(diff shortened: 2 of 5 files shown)


Xunzhuo (changes requested):
Welcome, and thanks for the focused download-scope fix. One blocker: `hf download --exclude` accepts one pattern per option, but `buildDownloadArgs` emits one flag followed by all three patterns, so the remaining patterns become positional filenames. Exact-head CI reports the exclusion ignored and leaves `model.safetensors` absent. Please repeat `--exclude` for each pattern, update the argv test, and rerun the required checks.

adaamko (commented):
agree with the fix and with xunzhuo's blocker. one reason it bit: the image installs huggingface_hub[cli] unpinned, and the new `hf download` (typer) takes `--exclude` as a repeatable option, while the old `huggingface-cli download` took nargs=*. repeating the flag works on both, so thats the safe form whichever cli the image ends up with.

two more from #3498, whose E2E this download broke twice. `IsGatedModelError` returns true for any failure when HF_TOKEN is empty (the `noToken` branch), so a 429 or a connection reset on this exact repo gets logged as "gated model, skipping" and the router comes up with embedding_ready:false; semantic-cache then fails 0/15 and nothing else notices. this PR makes the fetch 7x smaller, which helps, but the skip branch is the real bug: a public repo that fails to download should retry and then fail startup, not get skipped as gated. follow-up, not this PR, i can file it.

and a heads up: #2828 canonicalizes the embedding paths through ResolveModelPath in the same function, and your exclude map is keyed by the raw cfg path. whichever of the two lands second has to key on the resolved path or the exclusion silently stops applying for aliased configs.

vuhluu:
@Xunzhuo thanks, fixed in 86ce721d: one `--exclude` flag per pattern, argv test updated (`TestBuildDownloadArgsRepeatsExcludeFlagPerPattern`). Reproduced outside CI with `hf` 1.30.0: the old argv fetched only `onnx/*/model.onnx.data` (4 files, 2.12 GB, no `model.safetensors`); the new argv fetches 16 files / 614 MB with `model.safetensors` and no ONNX weights. Details in the Test Result section.

@adaamko thanks, went with the repeated flag for exactly that reason. For #2828, the exclude map is now keyed by `config.ResolveModelPath` on both sides (`TestBuildModelSpecsExcludesOnnxWeightsForAliasedEmbeddingModel`), so it holds whichever lands second. Agree on the `IsGatedModelError` skip branch: please file it, happy to take it as a follow-up PR.

Xunzhuo (approved):

Issue #3475, opened 2026-09-04
[Feature] Produce a versioned offline calibration artifact for Confidence thresholds


Xunzhuo:
## Goal

Calibrate the existing `algorithm.confidence.threshold` for `confidence_method=avg_logprob` from reproducible offline evidence and produce a versioned, reviewable artifact or configuration diff with explicit provenance and rollback identity.

## Proposed Workgroup

MoM & Routing

## Background

#2341 covers several threshold families. The repository already has Confidence tuning scenarios, analyzers, Brier/ECE reporting, and a verification script, but those paths are not yet one maintained artifact workflow. The first vertical should consolidate what exists instead of adding a second calibrator or depending on future success estimators.

## Existing implementation seams

- Reuse `tools/agent/scripts/tuning/scenarios/confidence.py` and `tools/agent/scripts/tuning/analyzer.py`.
- Reuse the maintained Brier score and expected-calibration-error report surfaces.
- Replace the verification script's host-local hard-coded input path with a versioned manifest or explicit artifact input that works in maintained environments.

## Scope

- Pin the first target to `algorithm.confidence.threshold` with `confidence_method=avg_logprob` and define its outcome, population, objective, hard constraints, fallback, and valid score domain.
- Build deterministic train/calibration/held-out inputs from maintained evaluation or Replay-safe evidence through one versioned manifest.
- Report operating-point metrics, Brier score, expected calibration error, uncertainty, sample count, coverage or abstention, cost, latency, and relevant failure slices.
- Produce a versioned artifact or candidate configuration diff with dataset identity, method, expected impact, unsupported regions, approval state, and rollback identity.
- Prove that offline and observe workflows cannot mutate active request policy.

## Configuration and API impact

The output is a reviewable offline artifact or explicit threshold diff for the existing Confidence config. No synchronous request-path learning, automatic promotion, hidden config mutation, or new threshold surface is added.

## Non-goals

- Calibrate every signal, method, or decision threshold.
- Implement #3412's success estimator or selector.
- Train Router Models or add another tuning framework.
- Promote a threshold without held-out review and rollback evidence.

## Completion signal

- [ ] Existing tuning and Brier/ECE paths are composed into one maintained workflow.
- [ ] The verification input is portable and manifest-backed rather than host-local.
- [ ] Inputs, objective, method, seed, and held-out split are reproducible.
- [ ] Results include calibration, uncertainty, coverage, cost/latency, and no-safe-threshold cases.
[...]

PR #3541, opened 2026-09-07
[Evaluation] Add versioned offline confidence calibration artifacts

yuki-uix:
Related #3475

## Purpose

This PR implements the offline confidence-calibration artifact workflow for accepted issue #3475.

It adds:

- deterministic train, calibration, and held-out dataset splits;
- paired small-model and large-model confidence collection;
- Brier score, ECE, accuracy, escalation, uplift, regression, and coverage metrics;
- threshold selection using calibration data only;
- held-out evaluation and failure-slice reporting;
- versioned manifests, source hashes, model metadata, rollback guidance, and candidate configuration diff;
- checked-in fixtures for deterministic offline tests.

Affected modules:

- `tools/calibration/tuning/`
- `tools/calibration/tuning/build_confidence_artifact.py`

This PR does not modify runtime routing policy, automatically promote thresholds, train models, or commit provider-specific live API results.

The tools follow the current `tools/calibration/` layout. The manifest workflow uses
`build_confidence_artifact.py`; the existing `verify_confidence --data-dir` command
remains compatible. A CLI regression verifies artifact generation from outside
the repository, including creation of the output directory.

## Test Plan

- `PYTHONPATH=tools/calibration python -m pytest tools/calibration/tuning/tests -q`
- `make check BASE_REF=upstream/main`
- Prepared and collected a real MMLU-Pro evaluation dataset.
- Rebuilt the artifact twice and verified deterministic artifact IDs.

## Test Result

- Tuning tests: `73 passed`.
- Full calibration domain suite: `182 passed`.
- `make check BASE_REF=upstream/main` passed.
- `make docs-generated-check docs-cli-test docs-community-test` passed.
- Agent validation completed successfully.
- Manifest and workflow contract validation passed.
- Repository validation tests passed.
- Formatting, Ruff, AST supply-chain scan, and changed-file checks passed.
- Real-data collection completed for 350 questions:
  - train: 210
  - calibration: 70
  - held-out: 70
- Candidate threshold: `0.9683`.
- Held-out evaluation showed no net uplift and one regression.
- The artifact remains `candidate`; no production threshold was promoted.
- No runtime policy was changed.

### Local live artifacts

I generated a 350-question calibration artifact locally using DashScope-hosted Qwen models. I intentionally did not commit the provider-specific live results by default because they depend on a particular API provider, model snapshot, endpoint configuration, dataset revision, and sampling run. Committing them could make the results appear to be a canonical benchmark baseline, while they are primarily evidence that the workflow works end to end.

The PR therefore commits deterministic fixtures for CI and keeps the live artifact as local validation evidence. The locally generated artifact is reproducible from the documented workflow and has been verified for deterministic rebuilding.

If maintainers prefer to keep the full live artifact in the repository as a versioned reference, I am happy to add it with its manifest, model/provider metadata, dataset hash, generation configuration, and artifact ID.

---
<details>

- [x] The PR links an accepted issue with exactly one recognized owner

</details>


Xunzhuo (changes requested):
Thanks for putting the workflow and provenance surfaces together. The threshold sweep still misses valid operating points: OfflineAnalyzer only probes confidence plus 0.001, so two calibration scores less than 0.001 apart cannot be separated. For example, an uplift at 0.5000 and a regression at 0.5005 have a safe threshold between them, but the candidates jump from no escalation to escalating both. That can select a worse threshold or report no safe threshold even when one exists. Please enumerate exact score boundaries, for example with the next representable float or adjacent midpoints, and add a close-score regression.

yuki-uix:
Thanks for catching this edge case. You’re right that using a fixed `+0.001` margin can skip a valid operating point when adjacent confidence scores are closer than `0.001`.

I pushed a fix that:

- generates threshold candidates from the policy regions between adjacent observed confidence scores;
- uses midpoints, with `math.nextafter` as a floating-point fallback;
- preserves full precision during threshold evaluation;
- adds a regression test covering an uplift at `0.5000` and a regression at `0.5005`;
- documents the boundary-based candidate selection.

Validation completed:

- `61 passed`;
- the agent validation gate passed;
- the real artifact was rebuilt twice with the same artifact ID:
  `sha256:72b15ea4c605331a9fd12059fc3f82c3884be716cc2857e13b046d025a8a2012`;
- the artifact remains `candidate`, and no runtime policy was changed.

yuki-uix:
@Xunzhuo Thanks for the close-score catch — you were right that the epsilon sweep could skip a valid operating point. Rebased onto main and pushed the fix.

Rather than shrinking the epsilon, I removed it. `OfflineAnalyzer` now enumerates one threshold per policy region: the midpoint between each pair of adjacent observed confidence values, falling back to `math.nextafter` when the midpoint collapses onto an endpoint. Since the escalation rule is `confidence < threshold`, behavior only changes when a threshold crosses an observed score, so this covers every distinct policy on the observed data without an arbitrary margin.

Your example is now the regression test: uplift at 0.5000, regression at 0.5005, asserting a separating threshold exists between them.

One related bug your comment led me to: `round(tau, 4)` on the stored candidate would have collapsed two newly separated thresholds back into the same policy region. Rounding is now presentation-only.

I mutation-checked the new test rather than trusting a green run — restoring the old epsilon sweep makes it fail with StopIteration, which is exactly the failure mode you described. The full tuning suite is 61 passed.

The held-out result is unchanged: no net uplift and one regression on 350 MMLU-Pro questions, so the artifact stays `candidate` and no production threshold is promoted.

tools/agent/scripts/tuning/confidence_calibration.py
```diff
@@ -0,0 +1,737 @@
...
+    )
+    held_out_metrics = evaluate_threshold(
+        held_out, threshold, manifest.get("resources")
+    )
+    candidate_diff = None
+    if selection["status"] == "candidate":
+        candidate_diff = {
+            "path": "algorithm.confidence.threshold",
+            "from": policy.get("current_threshold"),
+            "to": _rounded(threshold),
```
Xunzhuo: **[P2] Preserve the evaluated threshold in the proposed config**

The analyzer now retains exact score boundaries, but this rounds the deployable threshold after evaluating the unrounded value. With an uplift at confidence `0.500000001` and a regression at `0.500000003`, the selected midpoint escalates only the uplift; the emitted `0.5` escalates neither. An artifact round-trip reproduces reported net uplift 1 versus 0 after applying the proposal. Preserve the exact threshold in the config diff and effective/baseline policy fields, or evaluate the rounded value before accepting it; cover this close-score case.

Xunzhuo (changes requested):
The exact-boundary sweep fixes the earlier selection issue, but artifact serialization still changes the selected policy and invalidates its reported uplift. Please preserve the evaluated threshold through the config proposal and add a round-trip regression.

tools/agent/scripts/tuning/confidence_calibration.py
yuki-uix: Fixed in e317c8d53.
The artifact previously reported metrics evaluated with the full-precision threshold, but rounded the proposed config threshold to eight decimal places. Applying that proposal could therefore escalate different requests from those evaluated.
The config proposal, baseline, and fallback now preserve the exact threshold used for evaluation.
I added three regression cases covering rounding down, rounding up, and retaining the current threshold when no safe candidate exists. Each test writes the artifact to JSON, reads it back, and reapplies the saved threshold to verify that the resulting behavior and metrics match the report.
All three cases fail before the fix and pass afterward. The full tuning suite passes with 64 tests; Ruff and make harness-check also pass.

Xunzhuo (approved):
The score-boundary and threshold round-trip fixes resolve the earlier findings. The separate artifact command preserves the existing verification CLI. Verified the 182-test calibration suite and a fixture-based collection-to-artifact rebuild from outside the repository. Live provider collection was not rerun.

wilsonwu (approved):
LGTM

PR #3542, opened 2026-09-07
[Feature] Let the complexity signal consume a remote backend (score.v1 + label_distribution.v1)

### LINKED ISSUE #2921: [Feature] Define the score.v1 contract for complexity backends

theohsiung:
## Summary

Complexity could not name a remote model at all — `CategoryModel` was the only
struct with a `Backend` field. It now reads both remote shapes, because a
difficulty model can be built either way and the contract decides how the
runtime reads the answer:

- **`score.v1`** (new) — a regression model returns one number; each rule
  converts it through its own boundaries. One request, one call, however many
  rules read it. Implements the `ScoringBackend` tier, declared since the
  three-tier split with a note saying nothing wired it in.
- **`label_distribution.v1`** (reused) — a three-class model returns
  `hard`/`easy`/`medium`, so the winning label is the verdict and its
  probability is a real confidence.

Both go through the shared connector, so timeout, retry, byte caps and error
mapping stay shared.

## Decisions worth knowing

- Backend attaches to the **module**, not a rule: `routing.signals` is replaced
  wholesale per recipe, so a rule-level backend would vanish under any recipe
  that did not repeat it.
- Direction lives in the **field names** (`hard_above`+`easy_below`, or
  `hard_below`+`easy_above`), in the model's own units — no normalisation, no
  `direction`/`range` field, and an overlapping band cannot be written.
  `threshold` stays the symmetric shorthand for the local margin.
- `contract` is **required** for complexity: with two readable shapes, guessing
  wrong would surface per request rather than at config load.
- `score.v1` reports **no confidence** — a score just short of the boundary is
  the least certain position, not a strong one. Validation warns.

Full reasoning is in #2921's body.

## Every surface a user reaches

- **Operator CRD**: `complexity_model.backend` and the four boundary fields;
  `hard`/`easy` candidates become optional, since a remote backend never reads
  them. CEL validation rejects the combinations the Router refuses at load —
  `threshold` beside a pair, two directions, half a pair — so a typo is a
  refused write rather than a crashloop. Deepcopy, both CRD copies and the
  generated reference doc regenerated.
- **Dashboard**: the complexity rule editor exposes the boundary pairs, so they
  can be seen and edited at all, and no longer reports a valid `score.v1` rule
  as missing a required `threshold`.
- **Failure behaviour**: a scorer outage marks every `<rule>:<verdict>` in
  `SignalErrors` — the key a complexity decision leaf actually resolves to —
  so `rules.on_unknown` (`no_match` / `match` / `fail_request`) governs it. A
  test drives the real decision engine rather than asserting the map.
- **Metrics**: the shared connector gains `llm_remote_connector_*` (duration,
  outcome by error kind, retries) — category benefits too. A per-rule
  `llm_complexity_verdict_total{rule,verdict,source}` makes a mismatched scale
  visible as one verdict taking every request. Both are in the metrics
  reference and get Grafana panels.
- **Response shape**: `score.v1` accepts `{"score": x}` as well as the
  one-entry HuggingFace array, and strips a UTF-8 BOM. Missing, null or
  non-finite is an error, never 0.
- **Docs**: the two-rule ladder has a worked table (one score → one verdict per
  rule), the priority interaction, and the metrics to watch.

## Validation

Router, operator and dashboard (frontend + backend) suites, the changed
packages under `-race`, `make check`, website build, ONNX variant build, and
the `complexity-remote-backend` E2E profile on kind. The profile declares no
local candidates, so a verdict can only come from the remote call, and one
score reaches two different verdicts through two rules' boundaries.

Closes #2921


Xunzhuo:
Thanks for working on it!

Xunzhuo (changes requested):
Thanks for the thorough contract work. Four integration blockers remain: remote complexity still constructs and preloads the unused local classifier before attaching the backend, so remote-only startup depends on local model resources and readiness ignores the backend; score.v1 treats a missing or null score as numeric zero; explicit boundaries accept NaN and Inf; and dynamic reload replaces RecipeClassifiers without lifetime-safe retirement, leaking the old remote connectors while only the startup set is registered for shutdown. Please bypass local construction in remote mode with backend-aware readiness, require score presence while preserving valid zero, reject non-finite boundaries, and make reload ownership safe for in-flight snapshots. The exact-head quality job also fails MD012 in the complexity tutorial.

theohsiung:
Three fixed in code; the fourth needed checking rather than changing.

**1. Local construction in remote mode** — `4d4ca486a`
`buildComplexityClassifierOption` returns early when a backend is set, so preloading candidate embeddings no longer drags the local embedding model into a remote-only startup. Readiness is now `isComplexitySignalReady()`, ORing both backends with the local classifier — before, a remote-only config reported the signal unavailable and the dispatcher skipped it.

**2 + 3. Score presence and non-finite values** — `174d2bcba`
`scoreEntry.Score` is `*float64`, so absent/null is distinguishable from a reported 0 while a genuine 0 still routes. NaN/Inf are rejected at the response and, for boundaries, at config load: NaN makes every comparison false, so a rule would silently answer `medium` for everything.

**4. Reload retirement** — premise doesn't hold; filed the part that does as #3645

Reload uses the startup builder, so the new set registers its own shutdown and the retired generation closes the old one:

```
buildReloadRouter = buildOpenAIRouterFromConfig     server.go:34
  -> buildRouterComponents -> buildEarlyResources
       -> resources.add(recipeClassifiers.Close)    router_build.go:265
Swap -> closeRetiredGeneration -> router.Close()    server.go:280,358
  -> RecipeClassifiers.Close -> Classifier.Close    classifier_lifecycle.go:151
```

`Classifier.Close` closes both complexity backends (`classifier_lifecycle.go:151-152`), so retired connectors are not leaked.

The real gap is narrower: only `Process` takes `generation.refs`, while `GetRouter()` and `Registry.ClassificationService()` hand out bare pointers, so `refs.Wait()` doesn't cover the classification API. It is latent — `connector.Close()` is `CloseIdleConnections()`, which does not interrupt in-flight requests — pre-dates this PR, and `categories` reaches it identically through `HTTPClassifierInference`. Shared reload machinery rather than this feature, so #3645.

**MD012** — `2cb71e500`.

Also merged main (#3489's schema rename re-validated through `config.Load`) and fixed a `gci` import-order failure in the new E2E files (`8162abb10`).

theohsiung:
The three red checks here are `main`'s, not this PR's: `4cc8e85ff` bumped the
model catalog to 1385 evaluations while
`TestGeneratedPublicModelCatalogSatisfiesDashboardContract` still asserts 1365,
so `dashboard-test-backend` fails on any PR that merges current main.
`git diff origin/main...HEAD -- dashboard/ website/static/model-catalog/` is
empty here.

Filed as #3646, one-line fix in #3647. This PR should go green once that lands.

theohsiung:
Expanded this PR to cover every surface the feature touches, then ran a review over it and fixed what that found.

**Feature completeness** — `fc20c1e3f` `c97c772d4` `854ad241d` `91b1f892b` `8e9e62cfd`

| | |
|---|---|
| Operator CRD | `complexity_model.backend`, the four boundary fields, `hard`/`easy` optional |
| Dashboard | boundary pairs in the rule editor, `threshold` no longer required |
| Response shape | `score.v1` also accepts `{"score": x}` |
| Observability | connector `llm_remote_connector_*`, `llm_complexity_verdict_total`, failure counter; remote score published as `:score`, not `:margin` |
| Docs | worked ladder table, priority interaction, failure behaviour |

**What the review caught** — `37cf4b7a1` `1892c6471` `8a5adc542`

Three of those additions promised something they did not deliver, and none was visible to the branch's own tests:

- `SignalErrors` was keyed by the bare rule name, but `validateComplexityConditionName` rejects a bare name at config load, so every complexity leaf is `<rule>:<verdict>` and the engine looked up a key that never existed. The fail-open path was inert. Now every verdict of the rule is marked, and a test drives the real decision engine instead of asserting the map.
- The docs told operators to set `on_error: match` on a complexity condition; `validator_decision.go` accepts `on_error` only on `classifier` conditions, so following the docs made the Router refuse to start. The mechanism is `rules.on_unknown`, now documented with an example.
- The verdict counter's `source` carried `text`/`image` from the local fusion while its help text and the sibling failure counter said `local`, so joining the two — the documented way to spot a scale mismatch — matched nothing.

Also fixed: the bundle CRD copy was never regenerated (OLM installs rejected remote-only rules), `easy_below`/`easy_above` shipped with empty descriptions, `EffectiveBoundaries` allocated a map per call on the per-request path, the retry counter counted retries that never happened, and a UTF-8 BOM was reported as a wrong response shape.

Added CEL validation so the combinations the Router refuses at load are refused at admission instead — a typo becomes a rejected write rather than a crashloop.

Verified: router, operator and dashboard suites, changed packages under `-race`, `make check` (0 hook failures, 0 lint issues), website build, ONNX variant build.

theohsiung:
@adaamko would be nice to have your review on this PR, thx!

Xunzhuo (dismissed):
Nice work

wilsonwu (dismissed):
LGTM

theohsiung:
Addressed the review, and drew a line under what belongs here.

**Fixed** — `2543643de` `bd7b8c813`

- The remote response body no longer reaches the error, and so no longer reaches operational logs. A classifier's error body routinely echoes the text it was asked to classify, so a user prompt could land there. Status, body size and truncation are kept — "said nothing" and "said more than we would read" are different faults. Shared with categories, which gains the same redaction.
- CEL now also checks that a boundary pair is *ordered*, not merely present, so a reversed pair is a refused write rather than a Router crashloop.
- `backend.contract` is required for complexity. The rule sits on `ComplexityModelConfig`, not on the shared backend block: complexity reads two response shapes, while a consumer that reads one — categories — should keep the field optional and default it.
- A test pins the `label_distribution.v1` deployment contract through the real transport. It shows the distribution is aligned *by label name* before the argmax is taken positionally, so an endpoint may answer in any order; and that a model whose classes are not named `hard`/`easy`/`medium` is refused with a clear error rather than mapped by position.

**Not doing here, tracked in #3671** — with the reasoning for each, so it can be disagreed with:

- **A `label_distribution.v1` E2E profile.** The transport and its name alignment are shared with categories and already run in a cluster; `<rule>:<verdict>` → decision is proven in a cluster by the score profile; the label contract itself is now pinned in process. The remaining increment is "the config starts in Kubernetes", on scaffolding identical to the score profile's — against 30-80 minutes of CI on every future e2e PR, permanently.
- **Dashboard-side boundary validation.** It would be the third implementation of one constraint — Router validator, CRD CEL, TypeScript — with nothing keeping them in sync. The editor should surface the Router's error, not restate the rule.
- **`threshold: 0` presence detection.** Real inconsistency (the CRD refuses it, YAML does not), but when it does load the pair already wins, so it is a missing diagnostic rather than a wrong verdict.

Also correcting myself: I previously argued the connector followed redirects and was wrong to. `CheckRedirect: rejectRedirect` arrived from #3308 via a merge, *after* the head that was reviewed — so the original finding was correct at the time, and neither of us fixed it.

Local verification on this head: the changed packages under `-race`, both operator modules, dashboard frontend, `make check` with no hook or lint failures, website build, ONNX variant build, `make bundle` producing no diff, and `operator-sdk bundle validate` clean.

PR #3544, opened 2026-09-07
[Bug] Route the category-remote-backend E2E through a real gateway route

### LINKED ISSUE #3543: [Bug] The category-remote-backend E2E profile 404s and has never passed
### Proposed Workgroup

### Describe the bug

The `category-remote-backend` E2E profile has never passed. Its job failed on
the PR that introduced it (#3138) and it keeps failing, so the profile reports
coverage of the remote category backend that it does not provide.

The router is not at fault. It starts cleanly and constructs the backend
correctly; the request never reaches it. The profile's `providers.models` name
is `base-model`, but the shared `AIGatewayRoute` in
`deploy/kubernetes/ai-gateway/aigw-resources/gwapi-resources.yaml` matches
`x-ai-eg-model` **exactly** against six names:

```
math-expert  science-expert  social-expert  humanities-expert  law-expert  general-expert
```

`base-model` has no route, so the gateway answers 404. Because the router looks
healthy and logs a correctly selected backend, the failure reads like a router
bug and is not one.

### To Reproduce

```
make e2e-test E2E_PROFILE=category-remote-backend
```

Observed on a clean kind cluster (2026-09-07):

```
❌ FAILED - category-backend-routing (2.264s)
  Error: remote category request: Unexpected status code: 404
Total: 1 | Passed: 0 | Failed: 1
```

And on the PR that added the profile:

```
E2E / integration-test (category-remote-backend)   fail   1h15m21s
```

Changing the profile's model to any of the six routed names makes the same
profile pass — verified while building `complexity-remote-backend`, which
inherited this bug by using the profile as its template.

### Expected behavior

The profile routes its request and its assertion actually exercises the remote
category backend, so a regression in that path fails the job.

### Impact

`selection: pr` means the job runs on every change touching the category
backend surface, always red. Two costs: the remote category path introduced by
#3138 has no working end-to-end coverage despite appearing to, and a job that
is permanently failing trains reviewers to discount it.

### Additional context

The fix is the profile's model name. `e2e/profiles/ai-gateway/values.yaml`
already uses `default_model: general-expert` for exactly this reason.

A PR follows. Filing this because I approved #3138 and cited its new E2E as a
strength without checking that the job passed — the profile existing is not the
same as the profile working.

theohsiung:
## Summary

`category-remote-backend` has never passed — its job failed on the PR that
introduced it (#3138) and has kept failing, so it reported coverage of the
remote category backend that it did not provide.

The router was never at fault. It starts cleanly and logs a correctly selected
backend; the request never reaches it. The profile named its model
`base-model`, while the shared `AIGatewayRoute` matches `x-ai-eg-model` exactly
against `math-`, `science-`, `social-`, `humanities-`, `law-` and
`general-expert`. Anything outside that list has no route, so the gateway
answers 404 — and because the router looks healthy, it reads like a router bug.

One line of substance: the model becomes `general-expert`, which
`e2e/profiles/ai-gateway/values.yaml` already uses for the same reason. The
rest is a comment so the next profile on this stack does not repeat it.

## Validation

Run on a kind cluster, before and after:

```
before:  ❌ FAILED - category-backend-routing (2.264s)  Unexpected status code: 404
after:   ✅ PASSED - category-backend-routing (2.245s)
```

Closes #3543


WUKUNTAI-0211 (approved):
LGTM.

wilsonwu (approved):
LGTM

PR #3545, opened 2026-09-07
[Fix] Make dashboard builder context menu keyboard accessible

### LINKED ISSUE #3515: [Bug] Dashboard collapsible headers and context menus are not keyboard reachable

yuki-uix:
Make the Expression Builder context menu keyboard operable, and close the accessibility gate that #3515 asked for.

Closes #3515

## What changed

**Context menu (the second half of #3515).** The seven actions at `ExpressionBuilderContextMenu.tsx` were plain `<div onClick>` elements. They are now `<button type="button" role="menuitem">` inside a `role="menu"` container, with the divider carrying `role="separator"`. The menu takes focus on open, supports cyclic Arrow Up/Down plus Home/End, and restores focus to the trigger on Escape. Native buttons supply Enter and Space. No `<div onClick>` remains in the file.

**The lint gate.** #3515 asks that "a new static-element handler fails lint instead of relying on review". At `warn` it does not — `npm run lint` exits 0. Both rules are now `"error"`, with an override listing the 28 files that already carried violations when the rules were introduced. Those stay at `warn` so CI is not blocked on work this PR does not do, while every other file, and every new file, is held to `error`. The list only shrinks.

This also removes a claim that had gone stale: `eslint.config.js` said the rules were "kept at warn until the remaining context-menu violations are resolved, then promoted to error". This is the PR that resolves them, so the promotion belongs here rather than in a comment describing a future one.

## Why this closes the issue

All fourteen sites the issue lists are now fixed: seven collapsible headers in #3522, seven context-menu items here. I verified the first seven against current `main` rather than trusting the merged PR — each of the six files carries `aria-expanded` and none still has a `<div onClick>`.

The third expected behavior, that new violations fail lint, is satisfied by the promotion above.

## Measured effect on the backlog

Counted with the same lint configuration on the same tree, reverting only this PR's source files:

| | warnings | sites | files |
| --- | --- | --- | --- |
| `upstream/main` | 108 | 66 | 29 |
| this branch | 82 | 53 | 28 |

#3522 reported 66 sites across 29 files remaining when it merged. Current `main` still measures exactly 66 across 29 — **108 commits landed since then with no new violations**. The rules held the line at `warn`, which is the evidence that promoting them now will not be disruptive.

## Verification

- `npm run type-check` — passes
- `npm run test:unit` — 193 files, 887 tests pass
- `npm run build` — passes
- `npm run lint` — 0 errors, 82 warnings, all from the listed backlog files

Lint passing proves nothing on its own; it passed before this change too. So the gate was mutation-checked:

| mutation | result |
| --- | --- |
| new `<div onClick>` in a file **not** on the list | **1 error**, lint fails |
| the same mutation against the previous config | warning only, lint passes |
| new `<div onClick>` in a file **on** the list | 0 errors, lint passes |

## Known limit

The third row is a real gap, stated rather than hidden: a new violation added inside one of the 28 listed files still only warns. Closing it needs a per-file ratchet so the recorded counts can only decrease. I left that out to keep this PR to the issue's scope, and I am happy to add it, either here or as a follow-up — say which you prefer.

No E2E case was added. The issue argues that the accessibility gap and the coverage gap are the same gap, and now that the controls carry roles they are reachable by the repository's `getByRole` strategy. `e2e/` has the `keyboard.press()` / `toBeFocused()` idiom to follow in `config-models-scale.spec.ts`, `model-inventory.spec.ts`, and `playground.spec.ts`, and `playwright.config` starts its own server. I will add a spec if the Workgroup wants one before this merges.


yuki-uix:
@Xunzhuo This is the other half of #3515 — the context-menu items, following #3522 which merged on 9/06. Rebased onto main, and I've extended it to close the gate itself, so the PR now says Closes #3515 rather than Related.

Two things beyond the original scope:

The seven context-menu actions are now `<button type="button" role="menuitem">` inside a `role="menu"` container, with cyclic arrow keys, Home/End, and Escape restoring focus to the trigger.

The two jsx-a11y rules are promoted from "warn" to "error", with an override listing the 28 files that already carried violations. #3515 asks that a new static-element handler "fails lint instead of relying on review", and at "warn" it does not — `npm run lint` exits 0. This also resolves a claim that had gone stale: eslint.config.js said the rules were kept at warn "until the remaining context-menu violations are resolved, then promoted to error". This is the PR that resolves them.

On whether promoting now is safe — #3522 reported 66 sites across 29 files remaining when it merged. Current main still measures exactly 66 across 29. 108 commits have landed since, with no new violations. The rules held the line at warn, which is why I think error is the right move now rather than later.

Measured with the same config on the same tree, reverting only this PR's source files: main is 108 warnings / 66 sites / 29 files, this branch is 82 / 53 / 28.

I mutation-checked the gate rather than trusting a green lint run, since lint passed before this change too. A new `<div onClick>` in a non-listed file produces 1 error and fails lint; the same mutation against the previous config produced only a warning. One gap I did not close: a new violation inside one of the 28 listed files still only warns. That needs a per-file ratchet, which I left out to keep this PR to the issue's scope — happy to add it here or as a follow-up.

Verified: type-check, 887 unit tests, build, and 0 lint errors. No E2E case yet; the controls now carry roles so they're reachable by the repo's getByRole strategy, and I'll add a spec if you want one before this merges.

I know the queue is deep at the moment, so no urgency from my side — I just didn't want this one to drift, since it's the last piece of #3515 and #3522 already landed the rest. Happy to rebase again whenever it suits you.

Xunzhuo (approved):
The menu uses native buttons, exposes menu semantics, and handles initial focus, arrow navigation, and Escape dismissal consistently with its caller. The exact-head dashboard checks and new markup/source regressions pass. A useful follow-up is a DOM interaction test for focus traversal and restoration; the current assertions do not exercise keyboard events.

wilsonwu (approved):
LGTM

FAUST-BENCHOU (approved):

PR #3549, opened 2026-09-07
[Bug] Write Dashboard config backups owner-only

### LINKED ISSUE #3548: [Bug] Dashboard config backups are world-readable and can contain API keys
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

Every deploy, rollback and setup-activate saves a copy of `config.yaml` into `.vllm-sr/config-backups/`. Those copies are written `0644` in a `0755` directory, so any other user on the host, or any process in the container, can read them. Up to 10 are kept.

`config.yaml` can hold plaintext credentials — `api_key` on provider models, RAG backends and vector stores. The shipped reference config has one at `config/config.yaml:63`.

The recipe store already writes the same kind of snapshot with `0600` (`dashboard/backend/recipe/activation_store.go:73`), so this is the handlers package being inconsistent with it rather than a deliberate choice.

Four writers are affected:

- `dashboard/backend/handlers/config_backups.go:31` — deploy backup
- `dashboard/backend/handlers/config_backups.go:85` — pre-rollback snapshot
- `dashboard/backend/handlers/setup.go:549` — setup activate backup
- `dashboard/backend/handlers/config_backups.go:61` — archived DSL, same directory tree, compiles into the same config

### To Reproduce

1. Run the Dashboard against a config containing an `api_key`.
2. Deploy any change (`POST /api/router/config/deploy`).
3. `ls -l <configDir>/.vllm-sr/config-backups/`

The files are `-rw-r--r--`, and `grep api_key` on them returns the credential in plaintext.

### Expected behavior

Backups are `0600` in a `0700` directory, matching the recipe store.

### Impact

Any local user or co-located container can read provider API keys out of the backup directory.

### Additional context

Two things that matter for the fix:

- `os.WriteFile` applies its mode only when it *creates* the file. Backup names are timestamped to the second, so a reused path keeps its old mode — the write needs an explicit chmod.
- Deployments that upgrade already have world-readable backups sitting on disk, so making new files `0600` is not enough on its own.

The parent `.vllm-sr` directory should be left as it is — it holds the generated Envoy config and runtime state that other containers read. Only the backup directory needs tightening.

abhinav-m22:
Closes #3548

## Purpose

Dashboard config backups were written `0644` in a `0755` directory, so any other user on the host or process in the container could read them. `config.yaml` can hold plaintext `api_key` values.

All four writers (deploy backup, pre-rollback snapshot, setup activate, archived DSL) now share two helpers that create the directory `0700` and each file `0600`. This matches what the recipe store already does.

Two extra bits:

* The write does an explicit chmod. `os.WriteFile` only applies its mode when it creates the file, and backup names are timestamped to the second so a path can be reused.
* Writing a new backup also fixes the mode of existing ones, best effort. Otherwise an upgrade keeps leaking old backups until they rotate out.

The parent `.vllm-sr` directory is left alone. It holds the generated Envoy config and runtime state that other containers read.

Module: Dashboard backend. Workgroup: Enterprise & Environment.

## Test Plan

```
cd dashboard/backend
go build ./...
go vet ./handlers/
gofmt -l handlers/
go test ./handlers/ -race
```

Seven new tests in `config_backups_test.go` cover: each of the four writers produces `0600` in a `0700` directory, a pre-existing `0644` backup gets repaired, an unrelated file in that directory keeps its mode, a reused path is tightened, and `.vllm-sr` keeps its own mode.

## Test Result

All pass.

```
ok  github.com/vllm-project/semantic-router/dashboard/backend/handlers  62.055s
```

`go build`, `go vet`, `gofmt -l` and `go mod tidy` are all clean.


Xunzhuo (changes requested):
Thanks for tightening these sensitive snapshots. Two security paths still fail the owner-only guarantee:

- `writeConfigSnapshot` overwrites an existing inode before chmod. `config.dsl` always reuses its path, and same-second backups can too, so newly written credentials remain readable through the old `0644` mode or an already-open descriptor. Please create a fresh `0600` regular file and atomically replace the target, rejecting symlinks/non-regular targets.
- `ensureConfigSnapshotDir` rejects a symlink or chmod failure, but `createConfigBackup` only logs that error and continues to repair and write through the same path. Please fail closed before writing and add symlink/permission-failure coverage.

abhinav-m22:
@Xunzhuo Both fixed

- **In-place overwrite:** `writeConfigSnapshot` now writes a fresh `O_EXCL` `0600` temp file and renames over the target, so the old inode is never written through; `rejectUnsafeSnapshotTarget` uses `Lstat` to refuse symlinks and non-regular targets first.
- **Fail closed:** `createConfigBackup` and `snapshotCurrentConfigBeforeRollback` now return errors instead of logging and continuing, so deploy and rollback abort before writing anything (setup-activate too, for consistency); new tests cover symlinked targets and dirs, non-regular targets, inode replacement, and unwritable-dir failure.

dashboard/backend/handlers/config_backups.go
```diff
@@ -12,29 +15,155 @@ import (
...
+
+// Restricts snapshots an earlier build left world-readable, so an upgrade does
+// not keep leaking them until they rotate out. Best effort: never fails a deploy.
+func repairConfigSnapshotPermissions(dir string) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return
+	}
+	for _, entry := range entries {
+		if !isConfigBackupEntry(entry) {
```
Xunzhuo: **Repair the legacy DSL snapshot**

Impact: Upgraded installations can retain .vllm-sr/config.dsl at 0644 with plaintext API keys, leaving the credential disclosure exploitable after the backup files are repaired.

Required change: Safely migrate an existing regular config.dsl to 0600 through a guaranteed upgrade or snapshot path without following symlinks, and add legacy-file coverage.

dashboard/backend/handlers/config_backups.go
```diff
@@ -68,25 +199,26 @@ func readConfigBackup(configDir string, version string) ([]byte, error) {
 	return os.ReadFile(backupFile)
 }
 
-func snapshotCurrentConfigBeforeRollback(configPath string, configDir string) []byte {
+// Fails closed for the same reason as createConfigBackup.
+func snapshotCurrentConfigBeforeRollback(configPath string, configDir string) ([]byte, error) {
 	existingData, err := os.ReadFile(configPath)
 	if err != nil || len(existingData) == 0 {
-		return existingData
+		return existingData, nil
```
Xunzhuo, line 246: **Propagate live-config read failures**

Impact: Rollback can overwrite an unreadable live config without a pre-rollback snapshot, and nil existingData also disables restoration if runtime application fails. Deploy has the same failure mode after converting every read error to empty data.

Required change: Treat only os.ErrNotExist as absence and abort deploy or rollback on every other read error before merging, snapshotting, or writing.

dashboard/backend/handlers/config_backups.go
```diff
@@ -12,29 +15,155 @@ import (
...
+	}
+	if err := writeAndCloseSnapshotTemp(temp, data); err != nil {
+		_ = os.Remove(temp.Name())
+		return err
+	}
+	if err := os.Rename(temp.Name(), path); err != nil {
+		_ = os.Remove(temp.Name())
+		return err
+	}
+	return nil
```
Xunzhuo: **Fsync the snapshot directory**

Impact: writeConfigSnapshot reports success before the rename is durable, allowing the live config to be mutated even though a crash can discard the backup directory entry and its restore point.

Required change: Open and sync the parent directory after a successful rename, propagate synchronization failures, and cover the durability boundary through the injectable filesystem seam used by persistence tests.

Xunzhuo (commented):
Thanks for the update—I found 3 current-head issues that need to be addressed before this can merge.

dashboard/backend/handlers/config_backups.go
abhinav-m22: Fixed. `repairConfigSnapshotPermissions` now takes `configDir` and repairs `config.dsl` alongside the backup files, through a shared `repairSnapshotFilePermissions` that `Lstat`s first so a symlink is skipped rather than chmod'd through. Covered by a DSL-less deploy tightening a legacy `0644` file, and a symlinked one leaving its destination untouched.

dashboard/backend/handlers/config_backups.go
abhinav-m22: Fixed. New `readLiveConfig` treats only `os.ErrNotExist` as absence; deploy, rollback and setup-activate all abort on any other read error. One note: `backupCurrentConfig` needed the `ErrNotExist` case to stay non-fatal, otherwise first-run setup with no config yet would fail closed.

dashboard/backend/handlers/config_backups.go
abhinav-m22: Fixed. `writeConfigSnapshot` now syncs the parent directory after the rename and propagates the failure, matching `recipe.writeFileAtomically`.

On covering this "through the injectable filesystem seam used by persistence tests" — I couldn't find such a seam in `dashboard/backend`, and the `recipe` persistence tests appear to hit the real filesystem too. So I covered the happy path plus a real `syncSnapshotDirectory` failure on a missing directory. Happy to add a proper injectable seam if you want fsync failure simulated, but it felt larger than this PR should carry — let me know.

Xunzhuo:
Thanks for the backup-permission fix and follow-up hardening. Closing this as superseded by merged #3944, which incorporates the owner-only backup and DSL snapshot fix, atomic replacement, failure propagation, and repair of existing snapshots while preserving shared-directory permissions. The fix for #3548 is now on main.

PR #3552, opened 2026-09-07
[Bug] Fold developer instructions into the Anthropic system block

### LINKED ISSUE #3551: [Bug] Developer instructions fail every request routed to an Anthropic backend
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

A request whose system prompt is a `developer` instruction fails when routing sends it to a model with `api_format: anthropic`.

Anthropic Messages has one system channel, so the encoder folds developer authority into it. That fold is treated as a lossy translation (`codec_anthropic_content.go:580`), and the router hardcodes `LossyFeatures: LossyReject` (`llmprotocol/policy.go:81`), so it becomes a hard error instead of a warning.

The Responses API is hit hardest. Its top-level `instructions` string is always decoded as a developer instruction, with no alternative (`codec_openai_responses.go:302`), so every Responses request with a system prompt fails against an Anthropic backend. Chat clients using `role: "developer"` fail the same way.

`role: "system"` works fine. Only `developer` fails.

The shipped config declares `qwen3-32b` as `api_format: anthropic` (`config/config.yaml:68`), so this is reachable out of the box.

### To Reproduce

Pure Go, no services needed:

```go
engine := NewBuiltinEngine()
body := []byte(`{"model":"m","messages":[{"role":"developer","content":"be terse"},{"role":"user","content":"hi"}]}`)
_, err := engine.TranslateRequest(llmprotocol.OpenAIChatV1, llmprotocol.AnthropicMessagesV1, body, nil)
```

Result: `lossy_translation: translation would lose instructions.role`, empty body. Swap `developer` for `system` and it succeeds.

### Expected behavior

The instruction text reaches Anthropic's `system` field and the request succeeds. The collapse is recorded as a diagnostic, like a generated `max_tokens` already is.

`LossyReject` should stay. Only this one classification is wrong, and the encoder already appends the content on the next line.

### Impact

Any deployment routing OpenAI-format traffic to an Anthropic-format model. Failures follow routing rules, so it looks like a flaky backend.

### Additional context

Two golden fixtures record the failure as expected behaviour and will need updating: the `messages` case in `013-chat-official-request-fields-anthropic-out.json` and the `instructions` case in `014-responses-official-request-fields-anthropic-out.json`.

Related, same code path: when encoding fails, the error is wrapped with `status.Errorf(..., "%v", err)` (`processor_req_body_routing.go:254`), which flattens it to a string. `processBodyRoutingError` uses `errors.As` to turn protocol errors into a clean 400, so it cannot see it and the stream is torn down instead. `%w` does not help, since `status.Errorf` does not support it.

abhinav-m22:
Closes #3551

## Purpose

A system prompt written as a `developer` instruction failed every request routed to a model with `api_format: anthropic`. The Responses API always writes its `instructions` string as a developer instruction, so that whole client format was broken against Anthropic backends.

Anthropic has one system channel, so the encoder folds developer authority into it. That fold was classified as lossy, and the router hardcodes `LossyReject`, so it failed the request. The fold is an approximation, not a loss: the instruction text survives. It is now recorded as a diagnostic and the request goes through. `LossyReject` is unchanged, and genuinely unrepresentable cases like refusals still fail.

Also fixes the error handling on the same path. Encode failures were wrapped with `status.Errorf(..., "%v", err)`, which flattens the error to a string, so `processBodyRoutingError` could not recognise a protocol error and returned a torn-down stream instead of a clean 400. Protocol errors are now returned unwrapped. Note `%w` would not work here, since `status.Errorf` uses `Sprintf`.

Two golden fixtures recorded the old failure and are regenerated. They now show the instruction landing in `system` plus an `instructions.role` / `approximated` diagnostic.

Modules: protocolcodec, extproc. Workgroup: Data Plane & Networking.

## Test Plan

```
cd src/semantic-router
go test ./pkg/protocolcodec/... ./pkg/llmprotocol/... -race
go vet ./pkg/protocolcodec/ ./pkg/llmprotocol/ ./pkg/extproc/
gofmt -l pkg/protocolcodec pkg/llmprotocol pkg/extproc
```

New tests in `anthropic_instructions_test.go`: developer role and Responses `instructions` both reach Anthropic with a diagnostic, `system` role is unchanged and gains no diagnostic, developer role stays a distinct role on OpenAI targets, and refusals still fail.

`matrix_fidelity_test.go` keeps its "not silently collapsed" check, restated against the diagnostic instead of the error.

## Test Result

All pass.

```
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/protocolcodec  19.386s
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/llmprotocol    1.101s
```

`go vet` and `gofmt -l` clean.

Not run locally: `pkg/extproc` tests. They need the Rust bindings built and this machine cannot link them (`cannot find -lcandle_semantic_router`). `go vet ./pkg/extproc/` type-checks the package and its tests and passes, so the change compiles, but the `dispatchWireError` path has no test run behind it here. Worth confirming in CI.

Left alone on purpose: the `serialization_error` metric label is still used for protocol errors on this path, and instruction role is still not modelled as a capability. Both are follow-ups.


src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -280,6 +280,20 @@ func (r *OpenAIRouter) finalizeProviderDispatchResponse(
 	return response, nil
 }
 
+// A ProtocolError is a client error, so it is returned unwrapped for
+// processBodyRoutingError to turn into a clean 400. status.Errorf formats with
+// Sprintf, which would flatten it and hide it from errors.As.
+func dispatchWireError(err error, ctx *RequestContext, reason string) error {
+	var protocolError *llmprotocol.ProtocolError
+	if errors.As(err, &protocolError) {
```
Xunzhuo: **Preserve server-error categories**

Impact: Internal provider-encoding failures are now surfaced as HTTP 400 client errors, misattributing server faults and suppressing correct 5xx behavior.

Required change: Pass through only categories intentionally handled as HTTP 400, such as invalid_request and unsupported_feature, or make the downstream response mapping category-aware. Keep ErrorInternal and other server-owned categories on the internal-error path and add coverage.

Xunzhuo (commented):

drivebyer (commented):
Suggest dropping the developer branch entirely rather than recording a diagnostic. OpenAI defines `developer` as the successor of `system`, and Anthropic's `system` is the same channel, so the mapping is equivalent, not lossy. Responses `instructions` is always developer, so the diagnostic would fire on every request.

abhinav-m22:
Thanks both, all addressed.

@Xunzhuo on error categories: `dispatchWireError` now only passes through `invalid_request` and `unsupported_feature`. Every other category, including `ErrorInternal` and the upstream ones, stays on the internal path and keeps its 5xx. Added a table test covering both client categories, all eight server-owned ones, and a plain non-protocol error.

@drivebyer on the developer branch: agreed and dropped it. Your point about `instructions` always being developer is the decisive one, since it meant the diagnostic fired on every Responses request and carried no signal. Also removed the now-unused `appendApproximation` helper and the dead `policy` and `diagnostics` params, flipped the tests to assert the diagnostic is absent, and regenerated the two goldens. The only change there is the removed diagnostic lines, the requests still succeed with `system` populated.

One case I did not try to keep: a Chat request with both a `system` and a `developer` message collapses two roles into one channel. Anthropic has no way to represent that either, so there is nothing actionable to report, but happy to revisit if you think it needs its own signal.

drivebyer (approved):
LGTM

wilsonwu (approved):
LGTM, thanks!

PR #3568, opened 2026-09-08
[Bug] Reject a label mapping that contradicts the model config

### LINKED ISSUE #3535: [Bug] mmbert32k-feedback-detector-merged ships two label maps that disagree
### Proposed Workgroup

### Describe the bug

The published artifact carries two label maps and they are a full permutation of each other:

| file | 0 | 1 | 2 | 3 |
|---|---|---|---|---|
| `config.json` `id2label` | SAT | NEED_CLARIFICATION | WRONG_ANSWER | WANT_DIFFERENT |
| `label_mapping.json` `idx_to_label` | NEED_CLARIFICATION | SAT | WANT_DIFFERENT | WRONG_ANSWER |

`config.json` is the one that matches the weights: on 20 prompts with an unambiguous intent it agrees 16/20 and the sidecar 0/20. The other five 32K classifiers agree between their two files; this is the only one that does not.

The router reads the correct file today only by accident. `feedback_mapping_path` is empty in `config.yaml:2446`, and `feedback_detector.go:129-131` falls back to `<model_id>/config.json` when it is.

### To Reproduce

Set `feedback_mapping_path` to the artifact's own `label_mapping.json`:

| input | default | with the sidecar |
|---|---|---|
| `give me another option` | want_different | wrong_answer |
| `can you explain that differently` | need_clarification | satisfied |
| `thank you so much` | wrong_answer | want_different |

No error, no warning.

### Expected behavior

One label map per artifact, or a load-time check that rejects a mismatch between `config.json` and the sidecar. The artifact identity checks under #3199 are where that would belong.

### Impact

Any deployment that sets `feedback_mapping_path` gets silently scrambled feedback signals.

subin9:
Closes #3535

## Purpose

`mmbert32k-feedback-detector-merged` ships two label maps that are a permutation of each other:

| file | 0 | 1 | 2 | 3 |
|---|---|---|---|---|
| `config.json` `id2label` | SAT | NEED_CLARIFICATION | WRONG_ANSWER | WANT_DIFFERENT |
| `label_mapping.json` `idx_to_label` | NEED_CLARIFICATION | SAT | WANT_DIFFERENT | WRONG_ANSWER |

`config.json` matches the weights. `loadMapping` took whichever file the configuration named, so pointing `feedback_mapping_path` at the sidecar relabels every prediction with no error and no warning. `config/config.yaml:2503` leaves that field empty and falls back to `config.json`, so the correct file is read today by accident.

`loadMapping` now cross-checks the configured mapping against the model's own `config.json` and refuses a disagreement. Reading `config.json` itself, a model that ships none, and a config with no `id2label` are not disagreements and still load. Labels are compared as written and case-insensitively, so a sidecar that spells a class differently from its `config.json` would also be refused; none of the six local artifacts does that, and this is the only one of them whose two files disagree at all.

This is a load-time check, not a repair. The artifact still ships both files, and #3199's artifact identity check is where publication would catch it.

## Test Plan

`go test ./pkg/classification/`.

## Test Result

Eight table cases over the validator and three named tests, covering the shipped permutation, a missing index, an unparsable config, and the cases that must keep loading. The existing `TestFeedbackDetectorLoadsMappingFromPath` still passes.

```
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/classification	58.167s
```


src/semantic-router/pkg/classification/mapping.go
```diff
@@ -481,3 +483,50 @@ func resolveSinglePositiveIndex(mapping *JailbreakMapping, positiveLabels []stri
...
+	if len(modelConfig.ID2Label) == 0 {
+		return nil
+	}
+	for idx, modelLabel := range modelConfig.ID2Label {
+		mappedLabel, ok := idxToLabel[idx]
+		if !ok {
+			return fmt.Errorf("label mapping %s has no entry for index %s, which %s labels %q",
+				mappingPath, idx, modelConfigPath, modelLabel)
+		}
+		if !strings.EqualFold(mappedLabel, modelLabel) {
```
Xunzhuo: **Compare canonical feedback labels**

Impact: A valid sidecar using SATISFIED for a model whose equivalent label is SAT now fails detector startup, even though the runtime intentionally normalizes both labels to satisfied.

Required change: Canonicalize both labels with normalizeFeedbackLabel before comparing them, and add coverage for supported aliases.

Xunzhuo (commented):

src/semantic-router/pkg/classification/mapping.go
subin9: Agreed, thank you!

`ValidateLabelMappingAgainstModelConfig` compares both labels through `normalizeFeedbackLabel` now (src/semantic-router/pkg/classification/mapping.go:527), so a sidecar that writes SATISFIED against a model config that writes SAT starts the detector, while a mapping that names a different class at an index is still refused. Coverage is a new validator case, "a supported alias is not a disagreement", next to the permutation case that must keep failing.

Xunzhuo (approved):

PR #3569, opened 2026-09-08
[Bug] Repair local Docker development workflows

### LINKED ISSUE #3461: [Bug] Three local developer workflow issues in docker.mk and pre-commit config
### Proposed Workgroup

### Describe the bug

Three small configuration issues affect developers running local `make` targets and pre-commit hooks. Each is a one-line fix but currently breaks or degrades the developer workflow.

### To Reproduce

**1. `make docker-build-vllm-sr-router` produces an image no other target recognizes**

```bash
make docker-build-vllm-sr-router
# Tags the image as: <registry>/vllm-sr-router:<tag>
```

But every downstream variable references `vllm-sr`, not `vllm-sr-router`:

- `docker.mk:217` — `VLLM_SR_IMAGE ?= ghcr.io/vllm-project/semantic-router/vllm-sr:$(DOCKER_TAG)`
- `docker.mk:220` — `VLLM_SR_ROUTER_IMAGE_DEFAULT ?= $(VLLM_SR_IMAGE)`
- `.github/workflows/docker-publish.yml` — the publish matrix entry is `vllm-sr)`

So the image built by this target is orphaned — `make docker-run-*`, integration tests, and the publish workflow all expect `vllm-sr`.

**2. `make docker-test-llm-katan` always fails**

```bash
make docker-test-llm-katan
# curl: (7) Failed to connect to localhost port 8000
# Models endpoint failed
```

The target has no prerequisite declared (`docker.mk:95`), so it immediately runs `curl http://localhost:8000/v1/models` without first building or running the container. Compare the adjacent `docker-run-llm-katan` (`docker.mk:103`) which correctly declares `docker-run-llm-katan: docker-build-llm-katan`.

**3. `node_modules` exclusion in js-ts-lint pre-commit hook is broken**

The exclude pattern at `.pre-commit-config.yaml:81` is:

```yaml
exclude: ^(\node_modules/)
```

`\n` is a newline escape in regex, not the literal character `n`. This pattern matches a path starting with a newline followed by `ode_modules/` — which no real path ever contains. The exclude silently never fires.

Compare the correct form used elsewhere in the same file:

- Line 61: `exclude: ^(node_modules/|website/node_modules/|...)` ✅
- Line 71: `exclude: ^(node_modules/)` ✅

While `pass_filenames: false` means the hook invokes `precommit_tool.py website` as a whole, the broken exclude still means any internal file-walking logic that consults the configured `exclude` pattern will fail to skip `node_modules`.

### Expected behavior

1. `docker-build-vllm-sr-router` should tag the image as `vllm-sr` (using `$(VLLM_SR_ROUTER_IMAGE)` or `$(VLLM_SR_IMAGE)`), matching every downstream consumer.
2. `docker-test-llm-katan` should declare a prerequisite (`docker-run-llm-katan`) so the container is running before the curl check.
3. The js-ts-lint exclude should use `^(node_modules/)` without the stray backslash.

### Impact

All three are local developer-experience issues. No production deployment is affected. Each fix is a single line.

### Additional context

- Bug 1 location: `tools/make/docker.mk:77`
- Bug 2 location: `tools/make/docker.mk:95`
- Bug 3 location: `.pre-commit-config.yaml:81`

Zi-Yi-Ming:
## Summary

Three local developer workflow issues reported in #3461. Each is a broken or degraded `make` / pre-commit path; no production deployment is affected.

1. `make docker-build-vllm-sr-router` tagged its image as `$(DOCKER_REGISTRY)/vllm-sr-router:$(DOCKER_TAG)`. `vllm-sr-router` is the router **role/container** name (`VLLM_SR_ROUTER_CONTAINER`), not an image name: that tag had no consumer anywhere in the repository, was absent from `docker-build-all`, `docker-push-all` and the `docker-publish.yml` matrix, and — because the target builds `$(VLLM_SR_DOCKERFILE)`, which switches to the ROCm/CUDA Dockerfile under `VLLM_SR_PLATFORM` — platform builds silently overwrote one shared tag.
2. `make docker-test-llm-katan` declared no prerequisite and immediately ran `curl http://localhost:8000/v1/models`, so it always failed unless a server happened to be running in another terminal.
3. The `js-ts-lint` hook excluded `^(\node_modules/)`. In a YAML plain scalar the backslash reaches the regex engine, where `\n` is a newline escape, so the pattern never matches.

### `tools/make/docker.mk`

1. **`docker-build-vllm-sr-router`** — tags `$(VLLM_SR_ROUTER_IMAGE)` instead of the orphan name, so the build uses the canonical image variable and inherits the `VLLM_SR_PLATFORM` ROCm/CUDA defaults.
2. **`docker-push-vllm-sr-router`** — pushes the same `$(VLLM_SR_ROUTER_IMAGE)`; leaving it on the registry-derived name would make it push a tag the build no longer produces.
3. **`docker-test-llm-katan`** — now a self-contained lifecycle: `docker-build-llm-katan` prerequisite → stale-container removal → `run -d` under a dedicated name → bounded `/health` readiness → `/v1/models` assertion → cleanup on every exit path.
   - `LLM_KATAN_HOST_PORT ?= 8000` and `LLM_KATAN_TEST_CONTAINER ?= llm-katan-docker-test` are overridable and registered in `docker-help`.
   - The image `CMD` is overridden with `--model dummy --backend echo`; the Dockerfile default `--model Qwen/Qwen3-0.6B` would make a smoke check depend on a Hugging Face download.
   - Readiness is bounded twice over: 60 attempts, and `--max-time 5` per request, so a listening-but-unresponsive server cannot hang the target.
   - Only the `EXIT` trap carries the cleanup; the `INT`/`TERM` traps just exit. Since `exit` fires the `EXIT` trap, cleanup is defined once and an interrupted run terminates instead of re-entering the readiness loop and mistaking its own cleanup for `container exited unexpectedly`.
   - Every container command goes through `$(CONTAINER_RUNTIME)`; no Docker-only syntax is introduced.
   - The whole lifecycle is one compound recipe, because Make starts each physical recipe line in its own shell.

### `src/vllm-sr/tests/test_local_dev_make_surface.py`

4. `_docker_mk_target()` slices one target's declaration plus its recipe, and three contract tests pin: the canonical image variable in both router targets (plus the absence of the orphan tag file-wide); the `docker-test-llm-katan` lifecycle invariants; and that `docker-run-llm-katan` stays foreground.

### `.pre-commit-config.yaml`

5. `js-ts-lint` exclude is now `^(node_modules/)`, matching the same file's other hooks.

## What this does NOT change

- `docker-run-llm-katan` and `docker-run-llm-katan-custom` keep their foreground, blocking behavior; the documented `make start-llm-katan` in one terminal / tests in another workflow is untouched.
- `VLLM_SR_IMAGE`, `VLLM_SR_ROUTER_IMAGE_DEFAULT` and `VLLM_SR_ROUTER_IMAGE` are unchanged, and `vllm-sr-router-build` is not merged into `docker-build-vllm-sr-router`.
- `install.sh`, `e2e/testing/start-llm-katan.sh`, `e2e/testing/run_memory_integration.sh`, `tools/make/milvus.mk` and the publish workflow are not modified.

## Testing

- `python -m pytest src/vllm-sr/tests/test_local_dev_make_surface.py` → 8 passed, 1 failed (see known failure below)
- Mutation checks on scratch copies confirm the new assertions can fail: restoring the orphan tag fails the canonical-name test; re-detaching `docker-test-llm-katan` fails the lifecycle test; deleting either signal trap fails the lifecycle test; removing `--max-time` fails the lifecycle test.
- `black --check` and `ruff check` → pass. `git diff --check` → clean. `pre-commit validate-config` → exit 0.
- `bash -n` on the recipe after applying Make's own expansion (`$$` to `$`, variable substitution, split into logical recipe lines) → pass.

**Known failure.** `test_environment_docs_explain_default_split_without_user_topology_flags` is already red on `main`: `f53d10fbf` rewrote `tools/agent/docs/environments.md` and the phrase that test asserts no longer exists in that document. This PR edits neither the test nor the document.

**Remaining risk.** Real `make` and Docker/Podman execution were **not run** — this development environment has neither. The Ctrl-C / `TERM` path in particular is verified only through static expansion and Make/shell semantics, not execution, and would benefit from one manual run on Linux. `shellcheck` does not cover `.mk` files (`files: \.sh$`).

## Notes

`VLLM_SR_ROUTER_IMAGE` is now used consistently by the router build and push targets, matching the repository's canonical image variable. Note that its default is the canonical `ghcr.io/vllm-project/semantic-router/vllm-sr:$(DOCKER_TAG)` path rather than a `$(DOCKER_REGISTRY)`-derived name, so `DOCKER_REGISTRY=<private> make docker-push-vllm-sr-router` no longer targets the private registry. This is consistent with `VLLM_SR_IMAGE`, which already hardcodes the canonical prefix, but differs from the sibling `docker-push-*` targets.

The new `docker-test-llm-katan` path uses the dummy/echo backend to avoid downloading the default Hugging Face model and cleans up its test container on normal and signal exits.

The Python contract tests added in this PR are currently not part of the repository's CI execution path: `src/vllm-sr/tests/**` is collected by no workflow or make target (`domains.yaml` maps the `vllm-sr-cli` domain to `make vllm-sr-test`, which runs `e2e/testing/vllm-sr-cli/run_cli_tests.py`, and its test discovery is scoped to that directory).

Closes #3461

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,7 +85,7 @@ repos:
         entry: .venv-agent/bin/python tools/agent/scripts/precommit_tool.py website
         language: system
         files: \.(js|ts|tsx)$
-        exclude: ^(\node_modules/)
+        exclude: ^(node_modules/)
         pass_filenames: false
 
   # Rust specific hooks
--- a/e2e/testing/vllm-sr-cli/test_integration_docker_smoke.py
+++ b/e2e/testing/vllm-sr-cli/test_integration_docker_smoke.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+"""
+test_integration_docker_smoke.py - Exercise the llm-katan Docker smoke target.
+
+The container lifecycle itself (detached start, readiness, cleanup) belongs to
+`make docker-test-llm-katan`. This test only proves that target is green on a
+real container host, which a static Makefile assertion cannot show.
+"""
+
+import os
+import unittest
+from pathlib import Path
+
+from cli_test_base import CLITestBase
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+DOCKER_SMOKE_TARGET = "docker-test-llm-katan"
+# Off the image's conventional 8000 so the smoke container cannot share a host
+# port with a stack member that happens to use it.
+DOCKER_SMOKE_HOST_PORT = os.environ.get("LLM_KATAN_HOST_PORT", "18081")
+# The target builds the image first, and its Dockerfile installs torch.
+DOCKER_SMOKE_TIMEOUT = 900
+
+
+@unittest.skipUnless(
+    os.environ.get("RUN_INTEGRATION_TESTS", "").lower() == "true",
+    "Integration tests disabled. Set RUN_INTEGRATION_TESTS=true to enable.",
+)
+class TestDockerSmokeIntegration(CLITestBase):
+    """Integration coverage for the repository's Docker smoke target.
+
+    The gate sits on the class so the unit path never reaches CLITestBase
+    .setUpClass, which resolves the runtime stack and probes for a container
+    runtime.
+    """
+
+    def test_docker_test_llm_katan_target_passes(self):
+        """`make docker-test-llm-katan` succeeds on the active container host."""
+        self.print_test_header(
+            "llm-katan Docker Smoke Test",
+            "Runs the Make target that owns startup, readiness, and cleanup",
+        )
+
+        result = self._run_subprocess(
+            [
+                "make",
+                DOCKER_SMOKE_TARGET,
+                f"LLM_KATAN_HOST_PORT={DOCKER_SMOKE_HOST_PORT}",
+            ],
+            timeout=DOCKER_SMOKE_TIMEOUT,
+            cwd=str(REPO_ROOT),
+        )
+
+        self.assertEqual(
+            result.returncode,
+            0,
+            f"{DOCKER_SMOKE_TARGET} failed: rc={result.returncode}\n"
+            f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}",
+        )
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/src/vllm-sr/tests/test_local_dev_make_surface.py
+++ b/src/vllm-sr/tests/test_local_dev_make_surface.py
@@ -1,3 +1,4 @@
+import re
 from pathlib import Path
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
@@ -5,6 +6,20 @@ DOCKER_MK_PATH = REPO_ROOT / "tools" / "make" / "docker.mk"
 AGENT_MK_PATH = REPO_ROOT / "tools" / "make" / "agent.mk"
 ENVIRONMENTS_DOC_PATH = REPO_ROOT / "tools" / "agent" / "docs" / "environments.md"
 MEMORY_INTEGRATION_PATH = REPO_ROOT / "e2e" / "testing" / "run_memory_integration.sh"
+
+
+def _docker_mk_target(name: str) -> str:
+    """Return one Make target's declaration lines plus its recipe body."""
+    lines = DOCKER_MK_PATH.read_text(encoding="utf-8").splitlines()
+    start = next(i for i, line in enumerate(lines) if line.startswith(f"{name}:"))
+    end = start + 1
+    while end < len(lines) and (
+        not lines[end].strip()
+        or lines[end].startswith("\t")
+        or lines[end].startswith(f"{name}:")
+    ):
+        end += 1
+    return "\n".join(lines[start:end])
 
 
 def test_split_topology_defaults_to_rebuilding_router_image() -> None:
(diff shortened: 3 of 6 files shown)


Zi-Yi-Ming:
This is ready for review. Rebased onto current `main` (`342d4523c`).

- Dropped the `.pre-commit-config.yaml` change during the rebase: `main` already carries the broader `exclude: (^|/)node_modules/` pattern, which supersedes the narrower fix this PR originally proposed.
- Remaining scope: local Docker development workflow fixes (`tools/make/docker.mk`), domain registration (`tools/agent/domains.yaml`), plus two new test files — `tools/ci/tests/test_docker_make_surface.py` and `e2e/testing/vllm-sr-cli/test_integration_docker_smoke.py`. Both are picked up by existing discovery (`unittest discover` in `make harness-check`, and `run_cli_tests.py`'s `loader.discover`), so no allowlist registration is needed.

Would appreciate a review when convenient — happy to adjust anything.

Xunzhuo (dismissed):
The build/push image selection and local smoke lifecycle agree across the supported platforms. The exact-head CLI integration actually runs make docker-test-llm-katan successfully, and the independent lifecycle review found no blocker.

wilsonwu (commented):
LGTM

wilsonwu (dismissed):

Zi-Yi-Ming:
Re-requesting review — the branch moved, so GitHub dismissed the previous approvals.

Since the last review, only two things changed:

1. **`main` was merged in.** Conflicts were confined to `tools/agent/domains.yaml` and `tools/make/docker.mk` and were resolved without dropping either side; the rest is `main`'s own content.
2. **One new commit** — a one-line fix in `tools/ci/verification_catalog.yaml`.

The fix: the new `Product Verification` pipeline builds only the images the plan lists, and `local.cli` never declared `llm-katan`. Since `docker-test-llm-katan` now depends on `docker-build-llm-katan` (which under `PREBUILT_RUNTIME_IMAGES=1` only inspects the prebuilt image), the target failed with `No such image: .../llm-katan:latest`. Declaring it — matching `local.memory` — is the whole change:

```diff
     suite: cli
-    images: [vllm-sr, dashboard]
+    images: [vllm-sr, dashboard, llm-katan]
```

`services` is deliberately untouched: the smoke test starts its own container on port 8000.

CI is green on `3d6d7ffa4`.

Re-review would be appreciated — @Xunzhuo @wilsonwu.

Xunzhuo (approved):
The Docker targets consistently use the configured runtime image, and the smoke target owns startup, bounded readiness and cleanup. Current-head CI ran all 14 CLI integration cases, including the real llm-katan Docker smoke, and all 402 harness cases passed. AMD/ROCm and CUDA hardware execution was not performed.

wilsonwu (approved):
LGTM

abhinav-m22 (approved):

Zi-Yi-Ming:
Thanks @wilsonwu for the merge-from-main help today, and @abhinav-m22 for the same yesterday (plus the approval!) — though honestly I've lost count of how many times this branch has been updated; the current lineage alone carries six merge-from-main commits, three of them mine 😅

Current state: `PR Gate` is green on `aceefc05e`, three approvals in, and the branch trails `main` by exactly one commit (#3926, which touches no `.github/workflows/` files).

This is the enqueue-side behavior I documented in #3902: the `queue approved pull requests` rule has never fired while a branch is behind — of the five queued PRs I sampled that day, every behind one was manually queued by a maintainer, and only the up-to-date one entered by rule.

So one small ask, whichever of you finds it easier:

1. tick the merge-queue checkbox for this PR manually (that's how #3908/#3850/#3795 got in while behind), or
2. update the branch once more — #3926 touches no workflows, so once it is up to date the rule should fire on its own.

Xunzhuo:
Thanks @Zi-Yi-Ming for working through these Docker workflow failures and their CI requirements. For the record, merged #3944 resolves #3461 on current main: router build/push use the selected image and registry, and the echo-backend smoke target owns startup, bounded readiness and cleanup without taking a host port or removing an existing named container. The implementation differs from this PR; its proposed CLI smoke-test/catalog additions were not merged.

PR #3614, opened 2026-09-08
[Fix] Pass through client-sent chat_template_kwargs across model rewrite

### LINKED ISSUE #3573: [Bug] `chat_template_kwargs` is rejected with 400 and silently dropped when the router rewrites the model
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

`chat_template_kwargs` is how vLLM users control thinking mode per request — most
commonly `{"enable_thinking": false}` for Qwen3 hybrid-thinking models, which think by
default. Through semantic-router this field cannot reach the backend. Two defects:

1. Rejected on decode. The field is not in the OpenAI Chat codec whitelist
   (`chatRequestWire`, pkg/protocolcodec/codec_openai_chat.go), so every request
   carrying it returns 400:

       {"error":{"code":"invalid_json",
                 "message":"request JSON contains a non-canonical field"}}

   There is no config switch to relax this: the engine is always built with
   `DefaultPolicy()`, which hardcodes `UnknownFields: UnknownReject`
   (pkg/llmprotocol/policy.go).

2. Silently dropped after a model rewrite. When the router rewrites `model` — e.g.
   `model: "auto"` routing — the original bytes can no longer be replayed
   (`prepareProviderDispatch` bumps the envelope generation, so `Envelope.CanReplay`
   returns false). The request is then re-serialized from the neutral
   `llmprotocol.Request`, which has no representation for the field. The backend
   never sees it and runs the model in thinking mode.

The signature is clean: explicit model name works, `model: "auto"` breaks. With an
explicit name the unmodified request is replayed byte-for-byte, field included; auto
routing rewrites the model and loses the field.

Impact on our benchmark (240 MMLU/MMLU-Pro questions, two Qwen3.5 backends behind the
router): completion tokens 183k vs 24k direct, 31% of responses truncated by
max_tokens with no final answer, accuracy 45% vs 75% offline prediction. A local
4-line pass-through patch restores accuracy to 75%, zero truncations, baseline
token volume.

### To Reproduce

main @ a565be11 (2026-09-04), Envoy ext_proc deployment, vLLM backends serving
hybrid-thinking models (Qwen3.5). Re-checked on main HEAD (2026-09-08): still present.
On v0.3.0 the request is accepted (no 400), but the field is still dropped after a
model rewrite — see defect 2 below.

1. Send a request carrying the field:

       curl http://<router>/v1/chat/completions -d '{
         "model": "qwen3.5-4b",
         "messages": [{"role": "user", "content": "hi"}],
         "max_tokens": 8,
         "chat_template_kwargs": {"enable_thinking": false}
       }'

   → 400 invalid_json: "request JSON contains a non-canonical field".

2. Same request without the field, with `"model": "auto"` → 200, but the backend
   generates thinking output: completion tokens ~7.5x the direct baseline, many
   responses truncated by max_tokens.

3. Locally patch the decode whitelist (`json.RawMessage` in `chatRequestWire`) and
   repeat both: explicit model name → field reaches the backend (byte replay);
   `"model": "auto"` → field silently dropped after the rewrite. That asymmetry
   is defect 2.

### Expected behavior

The router should not interpret the field, but it should not lose it either: accept
`chat_template_kwargs`, carry it through opaquely as `json.RawMessage`, and write it
back unchanged — including after a model rewrite.

### Impact

_No response_

### Additional context

Validated fix — 4 changed lines across 3 files (plus a 3-line comment), pure
passthrough (the router never interprets the value):

1. pkg/protocolcodec/codec_openai_chat.go — add `ChatTemplateKwargs json.RawMessage`
   to `chatRequestWire` (decode whitelist), and copy it in `decodeChatBaseRequest`
2. pkg/llmprotocol/types.go — add the same field to the neutral `Request`
3. pkg/protocolcodec/codec_openai_chat_encode.go — write it back in
   `encodeChatBaseRequest`

I plan to open a PR with this fix.

PharosEast:
Closes #3573 

## Purpose

`chat_template_kwargs` is how vLLM users control thinking mode per request
(e.g. `{"enable_thinking": false}` for Qwen3 hybrid-thinking models). Through
semantic-router this field could not reach the backend:

1. Rejected on decode — the field was not in the OpenAI Chat codec whitelist
   (`chatRequestWire`), so every request carrying it returned 400
   `invalid_json: "request JSON contains a non-canonical field"`.
2. Silently dropped after a model rewrite — auto-routing rewrites `model`,
   which bumps the envelope generation and disables byte replay; the
   re-serialized request rebuilt from the neutral `llmprotocol.Request` had
   no representation for the field, so the backend never saw it and ran the
   model in thinking mode.

This PR registers `chat_template_kwargs` as a pass-through extension field:
it is carried opaquely as `json.RawMessage` from decode through the neutral
request to re-encode, without interpretation by the router.

## Test Plan

- `go test ./pkg/protocolcodec/` — includes three new cases in
  `codec_openai_chat_template_kwargs_test.go`:
  - decode accepts `chat_template_kwargs` (no 400)
  - decode → model rewrite → re-encode preserves the field (regression for
    the silent drop)
  - absent field stays absent (`omitempty`)
- Closed-contract inventories updated: official schema inventory/dispositions
  (extension bucket), mock-vllm `schema_contract.json`, and the
  `013-chat-official-request-fields` capability fixtures and goldens.
- `make check CHANGED_FILES="src/semantic-router/pkg/protocolcodec/codec_openai_chat.go src/semantic-router/pkg/llmprotocol/types.go src/semantic-router/pkg/protocolcodec/codec_openai_chat_encode.go src/semantic-router/pkg/protocolcodec/codec_openai_chat_template_kwargs_test.go src/semantic-router/pkg/protocolcodec/official_schema_inventory_test.go"`

## Test Result

- `go test ./pkg/protocolcodec/` all green locally.
- Capability matrix golden shows the field byte-preserved for
  chat→chat translation; cross-format outcomes recorded in the updated goldens.

Benchmark impact (240 MMLU/MMLU-Pro questions, two Qwen3.5 backends behind
the router, before/after a local patch): completion tokens 183k → 24k,
truncated responses 31% → 0, accuracy 45% → 75% (matches direct-connect
baseline).


drivebyer (approved):
LGTM

wilsonwu (approved):
LGTM

PR #3617, opened 2026-09-08
[Security] Block SSRF in dashboard outbound fetches with one shared policy

### LINKED ISSUE #1388: [Security] Block SSRF in Dashboard OpenWeb fetches
## Audited upstream baseline (2026-07-15)

`dashboard/backend/handlers/openweb.go` accepts a caller-provided HTTP(S) URL. `fetchWebDirect` constructs a normal `http.Client` and performs the request after scheme validation, but it does not enforce a destination policy after DNS resolution. Its redirect callback limits redirect count but does not revalidate each redirect destination.

An authorized caller can therefore make the dashboard backend attempt connections to loopback, private, link-local, metadata, or otherwise internal addresses reachable from the dashboard network. Hostname checks alone are insufficient because DNS resolution and redirects can change the effective destination.

## Required behavior

- Define a deny-by-default destination policy for the direct-fetch path, with an explicit allowlist mechanism if private targets are a supported operator use case.
- Resolve and validate every candidate IP before dialing; reject loopback, unspecified, multicast, private, link-local, and cloud-metadata destinations for the default public-web policy.
- Reapply the same policy to every redirect.
- Use a transport/dialer that prevents DNS rebinding or a check-then-connect race.
- Apply equivalent policy to proxy/reader modes or document why their trust boundary differs.
- Return a safe error without exposing internal network details.

## Acceptance criteria

- [ ] Direct IP, hostname, IPv4/IPv6, alternate-IP notation, and mixed-resolution cases are tested.
- [ ] Redirects from an allowed public URL to a denied destination are blocked.
- [ ] DNS rebinding/check-to-dial bypasses are covered by the transport design and tests.
- [ ] Cloud metadata and link-local ranges are denied by default.
- [ ] Logs redact URL credentials/query secrets and do not expose resolved internal addresses.
- [ ] Existing public HTTP(S) fetch behavior remains functional.

## Relevant surfaces

`dashboard/backend/handlers/openweb.go`, dashboard authentication/authorization, HTTP client construction, deployment docs, and security tests.

ankit373:
Fixes #1388

## Problem

The dashboard dials a caller-supplied URL from three places. Only one of them checks where it is actually going.

`recipe/importer.go` gets it right: it resolves the host, refuses any non-public address, and dials the resolved IP rather than the name, so nothing can move the destination between the check and the connection. It revalidates every redirect, strips `Referer`, disables the proxy and compression, and bounds the body.

`OpenWeb` and `FetchRaw` check the scheme and dial. A hostname resolving to loopback, a private range, or `169.254.169.254` is reached from inside the cluster, and neither revalidates redirects. Per @Xunzhuo's audit note this is the shared dashboard boundary, not one handler, so the fix is one policy rather than a second guard.

Two size gaps found while reading them, both closed here:

- `fetchWebDirect` read the body with a plain `io.ReadAll`. There was no decoded-response bound at all.
- `FetchRawHandler` bounded at 2 MB with compression left on, so the bound applied to the compressed stream.

## Change

`dashboard/backend/safefetch` is the one outbound policy. It is the importer's logic, lifted rather than rewritten, so what ships here is code that was already reviewed.

**Address check after DNS, not on the hostname.** A name check is decided before resolution, so it cannot see the destination. The transport resolves the host itself, requires *every* answer to be public, and then dials that resolved address literally. Dialing the address rather than the name is what closes the rebinding window: a resolver that answers publicly during validation and privately afterwards cannot move the connection, because the address is already chosen.

Denied by default: loopback, unspecified, private, CGNAT, link-local including the metadata address, documentation, benchmarking, IETF protocol, multicast, reserved, and the IPv6 equivalents plus 6to4, Teredo and NAT64, which embed an IPv4 destination. An IPv4-mapped v6 address is unmapped first, so `::ffff:127.0.0.1` is judged as `127.0.0.1`.

**Redirects** are revalidated against the same policy and re-dialled through the same address check. `Referer` is stripped, so a credential in the source query is never disclosed to the target.

**Deadlines** on dial, TLS handshake, response headers, and the whole request. The response-header deadline is what stops a peer that accepts a connection and then stalls.

**Proxy disabled**, because a proxy terminates the connection somewhere the policy never inspected. **Compression disabled**, so a size bound applies to bytes that actually cross the wire.

Refusals are sentinel errors, so callers map them onto their own response shapes without matching message text.

### Wiring

| Consumer | Policy |
|---|---|
| `fetchWebDirect` | HTTPS+HTTP, 8 MB bound (was unbounded) |
| `FetchRawHandler` | HTTPS+HTTP, 2 MB bound, now on decoded bytes |
| `recipe.Store.Import` | HTTPS only, 60s, unchanged behaviour |
| Jina reader | explicit proxy mode, fixed host, both response formats bounded |

**A refused destination fails closed.** OpenWeb previously fell back to the Jina reader whenever the direct fetch failed, which would retry a blocked URL by a second path. Only a transport or upstream failure earns the fallback now; a policy refusal is terminal.

**Refusals disclose nothing.** No resolved address, no address class, no reason. An SSRF probe maps a network by reading error text even when every fetch fails, so `assertNoNetworkDisclosure` asserts the response body contains none of it.

The importer keeps its own error codes; policy refusals are mapped back onto `ErrorSourceForbidden`, `ErrorInsecureURL`, `ErrorInvalidURL` and `ErrorDownloadFailed`, so its API contract is unchanged.

## Acceptance criteria

- [x] Direct IP, hostname, IPv4/IPv6, alternate-IP notation, and mixed-resolution cases are tested. `TestIsPublicAddrRejectsNonPublicClasses` covers 17 classes in both families including IPv4-mapped and IPv4-compatible v6; `TestClientRefusesMixedPublicAndPrivateAnswers` covers a mixed answer.
- [x] Redirects from an allowed public URL to a denied destination are blocked. `TestClientRevalidatesEveryRedirect`, `TestClientRefusesRedirectResolvingToNonPublicAddress`, `TestClientEnforcesTheRedirectLimit`.
- [x] DNS rebinding / check-to-dial bypasses covered by the transport design and tests. The dialler pins the validated address; `TestClientPinsTheValidatedAddressAgainstRebinding` drives a resolver that flips after the first answer.
- [x] Cloud metadata and link-local ranges denied by default.
- [x] Logs redact URL credentials and query secrets and do not expose resolved internal addresses. Existing `redactURLForLog` retained; `assertNoNetworkDisclosure` asserts the same for responses.
- [x] Existing public HTTP(S) fetch behavior remains functional. The full `handlers` and `recipe` suites pass unchanged.

Also from the audit note: oversized and compressed bodies (`TestReadBoundedRejectsOversizedBody`, `TestClientDoesNotAutoDecompress`), slow peers (`TestClientTimesOutOnASlowPeer`), and proxy behaviour made explicit. No deployment-specific hosts in any fixture; every test uses a substituted resolver or a loopback `httptest` server.

## Validation

```
cd dashboard/backend
go build ./...                                  -> ok
go vet ./safefetch/ ./handlers/ ./recipe/       -> clean
go test ./safefetch/                            -> ok, 18 tests
go test ./handlers/                             -> ok
go test ./recipe/                               -> ok
```

`gofmt` clean. Each of the three commits builds on its own.

The tests are load-bearing. Replacing the body of `IsPublicAddr` with `return true` fails 15 subtests across both packages, including every handler-level case:

```
--- FAIL: TestIsPublicAddrRejectsNonPublicClasses
--- FAIL: TestClientRefusesHostResolvingToNonPublicAddress   (5 subtests)
--- FAIL: TestClientRefusesMixedPublicAndPrivateAnswers
--- FAIL: TestFetchRawRefusesInwardDestinations              (5 subtests)
--- FAIL: TestOpenWebDoesNotFallBackAfterAPolicyRefusal
```

One pre-existing failure, `TestEvaluationRoutesLoadDeploymentScopedTargetsWithoutCatalogLeaks` in `backend/router`, also fails on a clean `upstream/main` checkout and is untouched by this change.

## Note on `importer_test.go`

`TestPackageHTTPClientRejectsMixedPublicAndPrivateDNSAnswers` asserted the transport returned a `PackageError` directly. It now asserts both layers: the transport refuses with the shared sentinel, and `classifyDownloadError` maps it onto `ErrorSourceForbidden`. The caller-visible contract is what it was.

### DIFF
--- a/dashboard/backend/safefetch/safefetch.go
+++ b/dashboard/backend/safefetch/safefetch.go
@@ -1,0 +1,224 @@
+package safefetch
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/netip"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Sentinel errors. Callers map these onto their own response shapes with
+// errors.Is rather than matching on message text.
+var (
+	// ErrInvalidURL is a URL that is not absolute, has no host, or carries
+	// credentials or a fragment.
+	ErrInvalidURL = errors.New("safefetch: url is invalid")
+	// ErrSchemeNotAllowed is a scheme outside the policy's allowed set.
+	ErrSchemeNotAllowed = errors.New("safefetch: url scheme is not allowed")
+	// ErrDestinationForbidden is a host that does not resolve, or resolves to
+	// an address the policy refuses to dial.
+	ErrDestinationForbidden = errors.New("safefetch: destination is not a public address")
+	// ErrTooManyRedirects is a redirect chain longer than the policy allows.
+	ErrTooManyRedirects = errors.New("safefetch: too many redirects")
+	// ErrResponseTooLarge is a body that exceeds the decoded-response budget.
+	ErrResponseTooLarge = errors.New("safefetch: response exceeds the size limit")
+)
+
+// IPResolver resolves a host to addresses. Tests substitute one to drive
+// address classes and mid-flight answer changes without real DNS.
+type IPResolver interface {
+	LookupNetIP(ctx context.Context, network, host string) ([]netip.Addr, error)
+}
+
+// Policy is the outbound-fetch contract for one caller-supplied URL.
+//
+// The zero value is not usable; start from DefaultPolicy and adjust.
+type Policy struct {
+	// AllowedSchemes are the URL schemes this consumer accepts, lowercase.
+	AllowedSchemes []string
+	// MaxRedirects is the number of redirects followed before refusing.
+	MaxRedirects int
+	// Timeout bounds the whole request, including redirects and body read.
+	Timeout time.Duration
+	// DialTimeout bounds one connection attempt.
+	DialTimeout time.Duration
+	// TLSHandshakeTimeout bounds the TLS handshake.
+	TLSHandshakeTimeout time.Duration
+	// ResponseHeaderTimeout bounds the wait for response headers, so a peer
+	// that accepts the connection and then stalls cannot hold it open.
+	ResponseHeaderTimeout time.Duration
+	// Resolver looks up destination addresses. Nil means net.DefaultResolver.
+	Resolver IPResolver
+}
+
+// DefaultPolicy is the public-web baseline: HTTPS and HTTP, bounded redirects,
+// and deadlines on every stage. Callers tighten it; none may loosen the
+// address check, which is not configurable.
+func DefaultPolicy() Policy {
+	return Policy{
+		AllowedSchemes:        []string{"https", "http"},
+		MaxRedirects:          5,
+		Timeout:               30 * time.Second,
+		DialTimeout:           10 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ResponseHeaderTimeout: 15 * time.Second,
+	}
+}
+
+// WithSchemes returns a copy of p accepting only the given schemes.
+func (p Policy) WithSchemes(schemes ...string) Policy {
+	p.AllowedSchemes = schemes
+	return p
+}
+
+// WithTimeout returns a copy of p with a different overall deadline.
+func (p Policy) WithTimeout(timeout time.Duration) Policy {
+	p.Timeout = timeout
+	return p
+}
+
+// WithResolver returns a copy of p using the given resolver.
+func (p Policy) WithResolver(resolver IPResolver) Policy {
+	p.Resolver = resolver
+	return p
+}
+
+// ValidateURL parses raw and checks everything decidable before DNS: the URL
+// is absolute and hierarchical, the scheme is allowed, and it carries no
+// credentials or fragment.
+//
+// Passing this is necessary but not sufficient. The destination address is
+// checked again at dial time, because only then is it known.
+func (p Policy) ValidateURL(raw string) (*url.URL, error) {
+	parsed, err := url.Parse(strings.TrimSpace(raw))
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrInvalidURL, err)
+	}
+	if !parsed.IsAbs() || parsed.Opaque != "" {
+		return nil, ErrInvalidURL
+	}
+	// Scheme first, so a refused scheme reports as one. file:// and friends
+	// also have no host, and reporting that instead hides the real reason.
+	if !p.schemeAllowed(parsed.Scheme) {
+		return nil, fmt.Errorf("%w: %q", ErrSchemeNotAllowed, parsed.Scheme)
+	}
+	if parsed.Host == "" || parsed.Hostname() == "" {
+		return nil, ErrInvalidURL
+	}
+	// Credentials would be replayed to a redirect target; a fragment is never
+	// sent and its presence means the caller built the URL by concatenation.
+	if parsed.User != nil || parsed.Fragment != "" {
+		return nil, ErrInvalidURL
+	}
+	return parsed, nil
+}
+
+func (p Policy) schemeAllowed(scheme string) bool {
+	scheme = strings.ToLower(scheme)
+	for _, allowed := range p.AllowedSchemes {
+		if scheme == allowed {
+			return true
+		}
+	}
+	return false
+}
+
+// NewClient builds an http.Client that enforces p.
+//
+// The transport resolves the host itself, refuses the request unless every
+// returned address is public, and then dials one of those resolved addresses
+// literally. Dialing the address rather than the name closes the check-to-dial
+// window: a resolver that answers publicly during validation and privately
+// afterwards cannot move the connection, because the address was already
+// chosen. Redirects are revalidated against the same policy.
+func (p Policy) NewClient() *http.Client {
+	resolver := p.Resolver
+	if resolver == nil {
+		resolver = net.DefaultResolver
+	}
+	dialer := &net.Dialer{Timeout: p.DialTimeout, KeepAlive: 30 * time.Second}
+
+	transport := &http.Transport{
+		// No proxy: a proxy would terminate the connection somewhere this
+		// policy never inspected, making the address check meaningless.
+		Proxy: nil,
+		// Bound the response by decoded bytes, so compression cannot be used
+		// to smuggle a body past the caller's limit.
+		DisableCompression:    true,
+		ForceAttemptHTTP2:     true,
+		TLSHandshakeTimeout:   p.TLSHandshakeTimeout,
+		ResponseHeaderTimeout: p.ResponseHeaderTimeout,
+		DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
+			return p.dial(ctx, resolver, dialer, network, address)
+		},
+	}
+
+	return &http.Client{
+		Transport: transport,
+		Timeout:   p.Timeout,
+		CheckRedirect: func(request *http.Request, via []*http.Request) error {
+			if len(request.URL.Scheme) == 0 || len(via) >= p.MaxRedirects {
+				return ErrTooManyRedirects
+			}
+			// A source URL may carry a credential in its query. Never disclose
+			// it to the redirect target.
+			request.Header.Del("Referer")
+			_, err := p.ValidateURL(request.URL.String())
+			return err
+		},
+	}
+}
+
+func (p Policy) dial(
+	ctx context.Context,
+	resolver IPResolver,
+	dialer *net.Dialer,
+	network string,
+	address string,
+) (net.Conn, error) {
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+
+	addresses, err := resolver.LookupNetIP(ctx, "ip", host)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrDestinationForbidden, err)
+	}
+	if len(addresses) == 0 {
+		return nil, ErrDestinationForbidden
+	}
+
+	// Every answer must be public, not merely the one that gets dialled. A
+	// mixed answer means the name is at least partly under someone else's
+	// control, and which record is used is not this code's decision to retry.
+	for _, candidate := range addresses {
+		if !IsPublicAddr(candidate) {
+			return nil, ErrDestinationForbidden
+		}
+	}
+
+	return dialer.DialContext(ctx, network, net.JoinHostPort(addresses[0].Unmap().String(), port))
+}
+
+// ReadBounded reads at most limit bytes and reports ErrResponseTooLarge if the
+// body is longer, rather than silently returning a truncated document.
+func ReadBounded(body io.Reader, limit int64) ([]byte, error) {
+	if limit <= 0 {
+		return nil, ErrResponseTooLarge
+	}
+	data, err := io.ReadAll(io.LimitReader(body, limit+1))
+	if err != nil {
+		return nil, err
+	}
+	if int64(len(data)) > limit {
+		return nil, ErrResponseTooLarge
+	}
+	return data, nil
+}
--- a/dashboard/backend/safefetch/safefetch.go
+++ b/dashboard/backend/safefetch/safefetch.go
@@ -232,7 +232,69 @@ func (p Policy) dial(
 		}
 	}
 
-	return dialer.DialContext(ctx, network, net.JoinHostPort(addresses[0].Unmap().String(), port))
+	// Every candidate is validated, so any of them is safe to dial. Try them in
+	// turn rather than pinning the first: the stock dialer resolves the name
+	// itself and falls back across answers, and dialing one address would drop
+	// that. A host whose first record is unreachable, which is the ordinary
+	// shape of a partial IPv6 outage or a multi-A record with one bad host,
+	// would otherwise fail outright.
+	var errs error
+	for _, candidate := range orderByFamily(addresses, network) {
+		conn, dialErr := dialer.DialContext(ctx, network, net.JoinHostPort(candidate.String(), port))
+		if dialErr == nil {
+			return conn, nil
+		}
+		errs = errors.Join(errs, dialErr)
+		// The overall deadline belongs to the caller. Once it is gone, further
+		// attempts only burn time we no longer have.
+		if ctx.Err() != nil {
+			break
+		}
+	}
+	if errs == nil {
+		return nil, ErrDestinationForbidden
+	}
+	return nil, errs
+}
+
+// orderByFamily returns the validated addresses with the two families
+// interleaved, keeping the resolver's order within each and leading with the
+// family the resolver put first. Trying every IPv6 answer before the first IPv4
+// one is what makes a broken IPv6 path look like a dead host.
+//
+// Addresses the transport cannot use for this network are dropped, so a tcp4
+// dial does not spend an attempt on an IPv6 answer.
+func orderByFamily(addresses []netip.Addr, network string) []netip.Addr {
+	var v4, v6 []netip.Addr
+	for _, a := range addresses {
+		a = a.Unmap()
+		switch {
+		case a.Is4():
+			if network != "tcp6" && network != "udp6" {
+				v4 = append(v4, a)
+			}
+		default:
+			if network != "tcp4" && network != "udp4" {
+				v6 = append(v6, a)
+			}
+		}
+	}
+
+	first, second := v4, v6
+	if len(addresses) > 0 && !addresses[0].Unmap().Is4() {
+		first, second = v6, v4
+	}
+
+	ordered := make([]netip.Addr, 0, len(first)+len(second))
+	for i := 0; i < len(first) || i < len(second); i++ {
+		if i < len(first) {
+			ordered = append(ordered, first[i])
+		}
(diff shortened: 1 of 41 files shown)


Xunzhuo (changes requested):
Thanks for taking on this security fix. One caller-controlled fetch still bypasses the new policy: SetupImportRemoteHandler builds a plain HTTP client and follows req.URL after only scheme and host validation, so /api/setup/import-remote can still dial loopback, private, link-local, or metadata destinations, including through redirects. This contradicts outboundPolicy being the single policy for caller-chosen destinations and leaves the SSRF class open. Please route this path through the same safe resolver, dial, and redirect policy and add coverage for it. The exact-head pre-commit, Dashboard, and PR Gate failures also need to be cleared.

ankit373:
@Xunzhuo good catch, you were right. `SetupImportRemoteHandler` was a genuine miss. I read "setup import" in the #2375 audit note as `recipe/importer.go` and did not go looking for a second one, so `/api/setup/import-remote` kept its own client and its scheme-and-host check. Fixed in `fdf5f3ec`.

It now takes the shared client, so the destination is rechecked after DNS and on every redirect, and the 1 MB bound applies to decoded bytes. `normalizeRemoteConfigURL` delegates its syntactic checks to the same policy while keeping its existing messages, so the setup UI's error text does not change.

Coverage added: `TestSetupImportRemoteRefusesInwardDestinations` across loopback, private, link-local, metadata and IPv6 loopback, plus `TestNormalizeRemoteConfigURLRejectsUnsafeInput` for the syntactic half.

### The allowlist

Routing this path through the policy broke two existing tests, `TestSetupImportRemoteHandler` and `TestSetupImportRemoteHandlerUsesConfigDirectoryForRelativeKBAssets`, which serve fixtures from `httptest` on loopback. Rather than weaken the check for tests, I added the mechanism the issue already asks for:

> Define a deny-by-default destination policy for the direct-fetch path, with an explicit allowlist mechanism if private targets are a supported operator use case.

`Policy.AllowedPrivatePrefixes` is empty in production and there is no config surface for it yet. A private destination is a declared prefix, not a flag that turns the check off, and it stays narrow: `TestAllowlistDoesNotWidenBeyondItsPrefixes` asserts that declaring loopback does not admit `10.0.0.0/8`. The two setup tests declare their fixture server the same way an operator would declare a real internal target.

Tell me if you would rather the allowlist wait for its own issue; I can pin those tests differently instead.

### CI

All three cleared. `PR Gate` was only aggregating the other two.

- **Pre-commit** was golangci-lint, not gofmt: `bodyclose` on four refused responses in the new tests, `errorlint` on a direct error comparison in `classifyFetchPolicyError` (now returns an explicit `ok`) and two `%w: %v` verbs, and one `staticcheck` S1021. Verified locally with `golangci-lint run -E bodyclose,errorlint,staticcheck,gosimple` over the three packages: clean apart from two pre-existing `SA1019` deprecations in `model_discovery_network.go` and its test, which this PR does not touch.
- **Dashboard** was the two setup tests above.

### One I looked at and did not change

`model_discovery_network.go` is a third independent implementation, but it is already hardened: `modelDiscoveryDialContext` with a per-request policy, no proxy, keep-alives disabled so a pooled connection cannot outlive a DNS change, and redirects stopped at `ErrUseLastResponse`. It also reaches internal endpoints deliberately, since discovering in-cluster models is the point, so folding it into a public-web default would change behaviour rather than harden it. Happy to converge it onto `safefetch` with a declared prefix set in a follow-up if you want one policy object everywhere, but it seemed wrong to do inside an SSRF fix.

Xunzhuo (dismissed):
Thanks for tightening this thoroughly. The shared fetch policy now covers every caller-controlled path, revalidates redirects, pins validated addresses through the dialer, and handles IPv4/IPv6 transition ranges. The exact Dashboard safefetch tests pass. Approved.

wilsonwu (dismissed):
LGTM

dashboard/backend/safefetch/safefetch.go
```diff
@@ -0,0 +1,252 @@
...
+	// Every answer must be public, not merely the one that gets dialled. A
+	// mixed answer means the name is at least partly under someone else's
+	// control, and which record is used is not this code's decision to retry.
+	for _, candidate := range addresses {
+		if !p.destinationAllowed(candidate) {
+			return nil, ErrDestinationForbidden
+		}
+	}
+
+	return dialer.DialContext(ctx, network, net.JoinHostPort(addresses[0].Unmap().String(), port))
```
Xunzhuo: **Retain fallback across validated addresses**

Impact: After validating every DNS answer, this always dials `addresses[0]`. FetchRaw, setup import, and OpenWeb direct/fixed-host fetches now fail when the first safe A/AAAA address is unreachable even if a later validated address works, including common partial IPv6 and multi-A failures.

Required change: Try the already-validated candidates under the same context and deadline, or use an equivalent Happy Eyeballs strategy over that pinned safe set. Add a regression where the first address fails and a later one succeeds.

Xunzhuo (commented):
Thanks for routing the remaining setup importer through the shared policy. The pinned-address dialer still drops normal fallback for newly migrated callers.

ankit373:
Good catch, and you are right. Fixed in `733b9e0`.

Pinning `addresses[0]` was never what made this safe, and I should not have written it that way. Safety comes from validating every answer and dialing resolved addresses rather than the name. Both still hold: the loop only dials candidates that already passed the check, so there is still no window for an answer to change underneath it. What pinning actually did was throw away the fallback the stock dialer gives you for free.

The candidates are now tried in turn under the same context and deadline, with two details worth flagging:

- **Families interleaved**, leading with whichever the resolver put first. Trying every AAAA before the first A is precisely what makes a broken IPv6 path look like a dead host, which is the case you named.
- **Answers the transport cannot use are dropped**, so a `tcp4` dial does not spend an attempt on an AAAA.

Errors are joined so the failure names every address tried, and the loop stops early once the caller's deadline is gone rather than burning time it no longer has.

Not full Happy Eyeballs. There is no parallel racing or 300ms head start; attempts are sequential over the pinned safe set. That keeps the change small and the dial path easy to reason about. Happy to add the racing if you want it, though it needs its own care around cancelling the loser.

Regression test as requested: a v4-only `httptest` server behind a resolver that answers `::1` first. It fails on the previous commit with `dial tcp [::1]:...: connect: connection refused` and passes now. Also added the all-unreachable case and a table for the interleave across v6-first, v4-first, single-family, `tcp4` and `tcp6`.

`go test ./safefetch/ ./handlers/ ./recipe/` is green, including every existing SSRF test: the non-public refusals, the mixed-answer refusal, the rebinding pin and the redirect revalidation.

dashboard/backend/safefetch/safefetch.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// Every candidate is validated, so any of them is safe to dial. Try them in
+	// turn rather than pinning the first: the stock dialer resolves the name
+	// itself and falls back across answers, and dialing one address would drop
+	// that. A host whose first record is unreachable, which is the ordinary
+	// shape of a partial IPv6 outage or a multi-A record with one bad host,
+	// would otherwise fail outright.
+	var errs error
+	for _, candidate := range orderByFamily(addresses, network) {
+		conn, dialErr := dialer.DialContext(ctx, network, net.JoinHostPort(candidate.String(), port))
```
Xunzhuo: **Keep fallback live when the first address stalls**

Impact: `DialTimeout` is 10 seconds, and setup/OpenWeb also give the whole client 10 seconds. A black-holed first AAAA therefore consumes the request deadline before this loop reaches a healthy IPv4 address, so the partial-IPv6 fallback this change claims still fails. The regression uses `::1`, which refuses immediately and does not exercise this case.

Required change: Race or stagger the already-validated families, or otherwise budget each attempt below the remaining request deadline, and add a deterministic test where the first family stalls while the second succeeds.

Xunzhuo (commented):
Thanks for the update—one timeout edge still blocks the address fallback.

ankit373:
Right again, and the part that stings is that my regression did not test what it claimed. `::1` refuses in microseconds, so it exercised the ordering but never the stall. A black hole is the case that matters and sequential dialing cannot survive it. Fixed in `edd737a9`.

The attempts are now staggered by `FallbackDelay` and raced, the RFC 8305 shape narrowed to a set of addresses that were all validated before any was dialled. First connection wins, losers are cancelled, and one that lands late is closed rather than leaked. A failure starts the next candidate immediately instead of waiting out the stagger, so a refusal costs nothing.

Safety is unchanged: validation still runs on every answer before any dial, and the race only dials candidates that already passed.

On testing it deterministically, I made `dialContext` injectable rather than reaching for a real black-holed address, since anything routable enough to stall is also timing dependent and platform specific. The new test holds the first family open until its context is cancelled and asserts the second family wins in milliseconds.

The mutation is the useful bit. Swapping the race back for the sequential walk does not just fail that test, it hangs until the suite times out:

```
with the race:        ok                          0.01s
sequential walk:      panic: test timed out       60s
```

Which is exactly the failure you described, and the reason the `::1` version was not good enough.

Also added: the next candidate starting immediately on failure rather than after the stagger, and the all-fail path naming every address tried.

`go test ./safefetch/ ./handlers/ ./recipe/` green, with the 12 existing SSRF subtests still passing.

One note on the branch: I rebased locally before seeing that @wilsonwu had merged `main` into it, so rather than force-push over that merge I replayed my commit on top of it. The merge is intact and this was a fast-forward.

dashboard/backend/safefetch/safefetch.go
ankit373: Fixed in 5e65a469: the loop now tries every already-validated address in turn (families interleaved, unusable-for-network answers dropped, errors joined) instead of pinning `addresses[0]`, with a regression using a v4-only server behind a resolver that answers ::1 first.

dashboard/backend/safefetch/safefetch.go
ankit373: Fixed in 3fe6a7e0: the candidates now race under the shared deadline instead of being tried sequentially, so a stalled first address can no longer consume the whole request budget before a healthy one gets a turn. Added a deterministic test where the first family stalls and the second succeeds.

dashboard/backend/safefetch/safefetch.go
```diff
@@ -0,0 +1,405 @@
...
+					case <-timer.C:
+					default:
+					}
+				}
+				timer.Reset(0)
+			} else if pending == 0 {
+				return nil, errs
+			}
+
+		case <-ctx.Done():
```
Xunzhuo, line 383: **[P2] Close outstanding successful dials when cancellation wins**

The success branch drains outstanding attempts, but this cancellation branch returns without consuming pending results. A successful connection can already be queued when cancellation becomes ready, or arrive at that boundary; choosing cancellation then leaves it open because canceling a dial context does not close an established connection. A deterministic late-success cancellation test reproduces the leak. Drain and close outstanding results on this exit too, and add the regression.

Xunzhuo (changes requested):
The staggered fallback addresses the earlier stalled-address issue, but cancellation can still leak successful connections. Please close outstanding dial results on that exit and add a cancellation regression.

dashboard/backend/safefetch/safefetch.go
ankit373: Fixed in 744dc75a: the ctx.Done() exit now spawns the same discardLateConns drain the success branch already uses, so a connection that lands after cancellation gets closed instead of abandoned. Added TestRaceClosesLateConnAfterCancellation, which holds one dial attempt open, cancels mid-flight, then lets the attempt succeed and asserts the connection closes; confirmed it times out waiting for Close without the fix and passes with it.

dashboard/backend/safefetch/safefetch.go
```diff
@@ -0,0 +1,411 @@
...
+	if resolver == nil {
+		resolver = net.DefaultResolver
+	}
+	dial := p.dialContext
+	if dial == nil {
+		dialer := &net.Dialer{Timeout: p.DialTimeout, KeepAlive: 30 * time.Second}
+		dial = dialer.DialContext
+	}
+
+	transport := &http.Transport{
```
Xunzhuo, line 187: **[P2] Own the lifetime of the per-fetch connection pool**

Every FetchRaw, OpenWeb, and setup fetch now creates this private transport, but those callers only close the response body and then discard the client. With `IdleConnTimeout` left at zero, a successful keep-alive connection stays open in that unreachable pool; unlike the previous shared default transport, later requests cannot reuse it. Repeated ordinary fetches therefore accumulate sockets and transport goroutines. Please share a lifecycle-owned transport or close idle connections when each short-lived fetch finishes, and add a completed-response cleanup regression.

Xunzhuo (changes requested):
The cancellation cleanup addresses the previous connection leak, and the focused cancellation test passes under the race detector. Successful fetches still leave a separate idle connection pool behind on every request. Please give those transports an explicit shared or per-fetch cleanup lifecycle.

dashboard/backend/safefetch/safefetch.go
ankit373: Fixed in a5e072a3. Wrapped the transport (idleClosingTransport) so it closes its own idle connections on every exit, not just the success path you named: immediately when a request fails outright (a redirect chain's earlier hop can already have pooled a connection by the time a later hop fails, and Client.Do then returns no body at all for a caller to close), and the moment the caller closes a successful response's body.

Two regressions, both using a real net.Conn wrapper against a real httptest.Server to check the actual socket gets closed rather than asserting on the transport's internal state: TestClientClosesConnectionAfterBodyClose for the success path you flagged, and TestClientClosesPooledConnectionWhenALaterHopFails for the failure path. Confirmed both fail against the unwrapped transport and pass with the fix.

Two existing tests type-asserted the concrete *http.Transport directly and needed updating for the new wrapper: one in-package, one cross-package (added an Unwrap() method for that one, so it can see through the wrapper without needing it exported, same convention as errors.Unwrap).

ankit373:
Investigated the DCO failure. Both of my own commits (the two safefetch fixes) are correctly signed off now, matching my actual git identity (fixed an email mismatch I'd introduced by hand-typing the trailer instead of using `git commit -s`).

The remaining 99 flagged commits are pre-existing history: this branch merged upstream/main in a few times over its life rather than rebasing, and DCO is evaluating the full commit range that brings in, not just what's actually mine. Sampled several of the flagged SHAs directly against GitHub's API; they all exist unchanged with `committer: GitHub` (the standard squash-merge bot), authored by other contributors whose own PRs already merged. That's not something I can or should fix by rewriting other people's already-merged commits.

Given `.mergify.yml` squash-merges this repo, the individual commit history won't carry through to the final merge anyway. Let me know if you'd like me to do anything differently here, or if this is just something to bypass at merge time.

dashboard/backend/safefetch/safefetch_test.go
```diff
@@ -0,0 +1,726 @@
...
+	}))
+	defer func() {
+		close(release)
+		server.Close()
+	}()
+
+	host, port := splitHostPort(t, server.Listener.Addr().String())
+	policy := DefaultPolicy().
+		WithTimeout(300 * time.Millisecond).
+		WithResolver(staticResolver{addresses: []netip.Addr{netip.MustParseAddr(host)}})
```
Xunzhuo: The fixture resolves to loopback, but this policy has no fixture exception, so the fetch is refused before the server is reached. This test currently passes in 0.00s despite the 200 ms header deadline; checking only for any error and an upper time bound cannot catch a missing timeout. Please allow the local fixture explicitly, assert that its handler was reached, and verify that the error is the expected timeout.

Xunzhuo (commented):
The connection cleanup and normal import regressions pass. The slow-peer regression below does not reach its server, so please correct the fixture and assert the expected timeout.

dashboard/backend/safefetch/safefetch_test.go
ankit373: Fixed in 4b5691a9. You're right on both counts: the policy refused the loopback fixture before the server was ever reached, so the header-timeout deadline was never actually exercised. Now allows the fixture explicitly, asserts the handler was reached (an atomic flag set before it blocks on release), and checks the returned error is a real timeout via `errors.As` + `net.Error.Timeout()` instead of just `err != nil`.

ankit373:
Status on the red checks at `4b5691a9`, since none of it is the dashboard change:

- **Dashboard Tests: passing** (9m10s). The slow-peer fixture is fixed as requested, and the four setup-import tests that were being refused at the SSRF gate before reaching their loopback fixtures now run properly too.
- **Production Benchmarks: failing on one benchmark**, `BenchmarkClassifyBatch_Parallel`, allocs/op 176 → 179 (+1.70%). That is a router-core classifier benchmark; this PR touches 18 files, all under `dashboard/`, and nothing in `src/semantic-router`. Three allocations is also noise-scale next to the rest of that run, where several benchmarks moved -60% to -85% against the same baseline. Happy to re-run if you want it confirmed rather than argued.
- **PR Gate / Product Verification / Gate**: both aggregate the benchmark job above.
- **DCO**: unchanged from my earlier note. My own two commits are signed correctly; the remaining flagged SHAs are pre-existing upstream history swept in by the merges, and `.mergify.yml` squashes this repo anyway.

Nothing outstanding from me on the review comments; the last one is answered in `4b5691a9`.

PR #3620, opened 2026-09-08
[Bug] Attribute selected-model header to the final dispatch model and document routing capabilities

Bevisy:
## Summary

Follow-up to #3313. Fixes a negotiation-observability and admission-accounting gap on the capability
reroute path and documents the routing-capability contract in public config
guidance.

- **`ctx.VSRSelectedModel` now follows the final dispatch model.** After a
  capability reroute sends the request to a sibling modelRef, the client-visible
  `x-vsr-selected-model` response header (and the accounting that keys off it)
  now point at the model that actually serves the request instead of the
  decision-selected proxy. This is the other half of the per-model accounting
  fix already landed for `ctx.RequestModel`.
- **The admission accounting and the selected-model header both follow the final dispatch
  model.** `prepareProviderDispatch` records the dispatched model in the request state
  (and in the selected-model header), so the in-flight entry and the response paths key
  off the model the request actually ends under.

- **Public config guidance** documents the protocol capability vocabulary
  (defined by `llmprotocol`) and that unrecognized declared capability names are
  treated as unannotated (model judged on wire expressibility alone).

## Test Plan

- `pkg/extproc` and `pkg/inflight` in the project's Linux test environment
  (`go vet` plus the package test suites; both packages are cgo).
- Reroute regressions: `TestPrepareProviderDispatchHopsInflightToReroutedModel`
  plus the existing reroute and attribution tests.

## Test Result

- `go vet ./pkg/extproc/ ./pkg/inflight/`: clean.
- Package test suites for both: pass.

## Notes

No start-time validation of declared capabilities is added here. The canonical
recipe configs still declare descriptive words (e.g. `vision`,
`cheap_followup`) that predate the protocol enum, so a fail-fast validator
would break all of them — the vocabulary boundary decision belongs to the
workgroup, not this PR.

## Out of scope

- Load-time capability index (dropped during review as a no-op micro-optimization)
- Start-time capability validation (workgroup decision, see above)

Related #3183


Xunzhuo (changes requested):
The runtime changes look sound, but the new capability guide overstates what model-card declarations control. qualifiedRerouteCandidate compares declared capabilities only with required.TaskCapabilities(), which deliberately excludes tools, reasoning, streaming, and structured_json; those are checked against wire codec expressibility, not the model declaration. Listing them as names the Router compares during model negotiation tells operators that declarations gate behavior they currently do not. Please limit this section to task and modality capabilities, or clearly separate model-declaration filtering from codec checks.

Bevisy:
> The runtime changes look sound, but the new capability guide overstates what model-card declarations control. qualifiedRerouteCandidate compares declared capabilities only with required.TaskCapabilities(), which deliberately excludes tools, reasoning, streaming, and structured_json; those are checked against wire codec expressibility, not the model declaration. Listing them as names the Router compares during model negotiation tells operators that declarations gate behavior they currently do not. Please limit this section to task and modality capabilities, or clearly separate model-declaration filtering from codec checks.

Agreed — the guide overstated declaration control. `TaskCapabilities()` masks to the task/modality set (`image_input`..`file_output`, hosted `image_generation`); the transport/accounting fidelity names (`tools`, `reasoning`, `streaming`, `structured_json`) are verified against wire codec expressibility in `qualifiedRerouteCandidate` via `codecCapabilitiesForFormat`, not the declaration.

The section now lists only the task/modality names that declarations steer, and states explicitly that the other names are checked against wire codec expressibility rather than the model declaration.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -190,6 +192,7 @@ func (r *OpenAIRouter) rerouteToQualifiedDecisionModel(
 		"to":          model,
 		"wire_format": candidate.targetFormat,
 	})
+	metrics.ModelRoutingModifications.WithLabelValues(selected.logicalModel, model).Inc()
```
drivebyer: Prefer the existing `metrics.RecordModelRouting` helper here.

Bevisy:
> Prefer the existing `metrics.RecordModelRouting` helper here.

Done — the reroute now calls `metrics.RecordModelRouting(selected.logicalModel, model)` instead of touching `ModelRoutingModifications` directly, so the `source != target` guard and counter wiring stay in one place.

drivebyer (commented):
LGTM

drivebyer (dismissed):
LGTM

website/docs/installation/model-configuration.md
```diff
@@ -81,6 +81,29 @@ The Dashboard edits the same v0.3 document as YAML. Use one authoring owner for
...
+      capabilities: [image_generation]
+```
+
+The protocol capability vocabulary is defined by the `llmprotocol` package,
+and only task/modality capabilities participate in declaration steering:
+`image_input`, `image_output`, `image_generation`, `audio_input`,
+`audio_output`, `video_input`, `video_output`, `file_input`, `file_output`.
+A declared name outside this set (including transport or accounting fidelity
+like `tools`, `reasoning`, `streaming`, `structured_json`) is treated as
+unannotated for declaration filtering — those are verified against wire codec
```
Xunzhuo: **Describe transport-only declarations accurately**

Impact: `ParseCapabilities` accepts `tools`, `reasoning`, `streaming`, and `structured_json`, so a model declaring only one of them is treated as annotated. `qualifiedRerouteCandidate` then rejects it for an image/audio/video/file request because the declaration lacks that task bit, although this guide says it remains eligible on wire expressibility. Operators following the guide can therefore get a different backend or a capability rejection.

Required change: Document the actual three-state behavior, or align the runtime and add a transport-only declaration test so these values truly remain unannotated for task filtering.

Xunzhuo (commented):
Thanks for narrowing the capability guide. One remaining sentence still describes transport-only declarations differently from the runtime.

Bevisy:
> **Describe transport-only declarations accurately**
>
> Impact: `ParseCapabilities` accepts `tools`, `reasoning`, `streaming`, and `structured_json`, so a model declaring only one of them is treated as annotated. `qualifiedRerouteCandidate` then rejects it for an image/audio/video/file request because the declaration lacks that task bit, although this guide says it remains eligible on wire expressibility. Operators following the guide can therefore get a different backend or a capability rejection.
>
> Required change: Document the actual three-state behavior, or align the runtime and add a transport-only declaration test so these values truly remain unannotated for task filtering.

Done both:

- `qualifiedRerouteCandidate` now narrows only on the declared task/modality subset; a transport/accounting-only declaration carries no task bit and stays eligible on wire expressibility, with a four-case regression test.
- The guide now documents the actual three-state behavior.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -225,8 +232,14 @@ func (r *OpenAIRouter) qualifiedRerouteCandidate(model string, required llmproto
 	if set, ok := r.codecCapabilitiesForFormat(format); !ok || !set.Contains(required) {
 		return ""
 	}
-	if declared, ok := r.declaredModelCapabilities(model); ok && !declared.Contains(required.TaskCapabilities()) {
-		return ""
+	if declared, ok := r.declaredModelCapabilities(model); ok {
```
Xunzhuo, line 258: **Do not drop valid task declarations on an unknown name**

Impact: `declaredModelCapabilities` returns `ok=false` when `ParseCapabilities` encounters any unknown name, so this branch skips all task filtering. Existing model cards use the broader catalog vocabulary (`vision`, `long_context`, and similar values); a mixed declaration such as `[image_input, vision]` is therefore treated as completely unannotated, allowing a wire-compatible backend to be rerouted for tasks it never declared.

Required change: Distinguish an absent declaration from an invalid or partially understood one—either validate/map the public capability vocabulary before routing or preserve known task bits while failing closed on unknown entries—and add mixed known/unknown dispatch and documentation coverage.

Xunzhuo (commented):
Thanks for the update. The selected-model attribution issue is resolved, but the new capability contract still has one fallback gap.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
Bevisy: > **Do not drop valid task declarations on an unknown name**
>
> Impact: `declaredModelCapabilities` returns `ok=false` when `ParseCapabilities` encounters any unknown name, so this branch skips all task filtering. Existing model cards use the broader catalog vocabulary (`vision`, `long_context`, and similar values); a mixed declaration such as `[image_input, vision]` is therefore treated as completely unannotated, allowing a wire-compatible backend to be rerouted for tasks it never declared.
>
> Required change: Distinguish an absent declaration from an invalid or partially understood one—either validate/map the public capability vocabulary before routing or preserve known task bits while failing closed on unknown entries—and add mixed known/unknown dispatch and documentation coverage.

Fixed. Model capability declarations now split into three states with distinct routing behavior:

- **partial** — `[image_input, vision]` keeps the `image_input` bit: eligible for image, rejected for audio (no longer seen as fully unannotated).
- **invalid** (nothing recognizes) — fails closed, never a reroute candidate; no more wire-expressibility fallback for unrecognized words.
- **absent** — unchanged, eligible on wire expressibility.

`ParseCapabilities` now preserves the recognized subset when it meets an unknown name (still erroring for strict callers). Tests cover both mixed-known and unknown-only dispatch; docs list the five declaration states. No start-time validation added — vocabulary boundary stays a workgroup decision.

Xunzhuo (commented):
Thanks for aligning the capability declaration and final dispatch model. One cross-system accounting blocker remains on this head.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -87,9 +87,11 @@ func (r *OpenAIRouter) prepareProviderDispatch(
 	ctx.SemanticRequest = request
 	// Per-model accounting (token tracking, TTFB, usage attribution) keys off
 	// the model that actually serves the request. A capability reroute may
-	// redirect this request to a sibling modelRef, so RequestModel must be the
-	// final dispatch model, not the decision-selected one.
+	// redirect this request to a sibling modelRef, so RequestModel and the
+	// client-visible selected-model header must be the final dispatch model,
+	// not the decision-selected one.
 	ctx.RequestModel = dispatch.logicalModel
+	ctx.VSRSelectedModel = dispatch.logicalModel
```
Xunzhuo, line 105: **Move admission accounting to the final model**

A capability reroute changes `RequestModel` to this final sibling, but the inflight token was opened earlier under the pre-reroute `selectedModel`. Every normal success, stream, and receive-error path then calls `inflight.End(ctx.RequestModel, token)`; because `End` keys by `(model, token)`, it misses the original entry and leaves that model artificially inflight until age eviction, skewing load-aware routing. Replay also starts with the pre-reroute model while the response header reports this final one. Please finalize admission/request/Replay attribution after `dispatch.logicalModel` is known (or retain a consistent accounting key), and add a reroute regression proving both models' inflight counts return to zero and attribution names the dispatched model.

Bevisy: > **Move admission accounting to the final model**
>
> A capability reroute changes `RequestModel` to this final sibling, but the inflight token was opened earlier under the pre-reroute `selectedModel`. Every normal success, stream, and receive-error path then calls `inflight.End(ctx.RequestModel, token)`; because `End` keys by `(model, token)`, it misses the original entry and leaves that model artificially inflight until age eviction, skewing load-aware routing. Replay also starts with the pre-reroute model while the response header reports this final one. Please finalize admission/request/Replay attribution after `dispatch.logicalModel` is known (or retain a consistent accounting key), and add a reroute regression proving both models' inflight counts return to zero and attribution names the dispatched model.

Landed. Admission accounting now finalizes on the dispatched model: the in-flight entry opened under the pre-reroute model is closed under the model that serves, so it no longer leaks on the original model and load accounting reflects the model that actually serves. Replay attribution now names the dispatched model, matching the response header that already did. The reroute regression pins it: the original model's count returns to zero, the dispatched model carries the single entry, and the entry ended is the one opened on it.

Xunzhuo (dismissed):
The current-main integration preserves final-dispatch attribution and the inflight-token handoff while using the centralized route-cache clearing path. The reroute/inflight and declared-capability regressions pass. The prior runtime attribution evidence still applies; a separate full deployment was not rerun.

wilsonwu (dismissed):
LGTM

Bevisy:
Merge conflict is recovered. PTAL @Xunzhuo @wilsonwu

Bevisy:
@Mergifyio queue

website/docs/installation/model-configuration.md
```diff
@@ -100,6 +100,49 @@ The Dashboard edits the same v0.3 document as YAML. Use one authoring owner for
 a deployment and review the exported YAML before serving it. See
 [Configuration workflows](configuration-workflows).
 
+## Routing capabilities
+
+Author-declared model capabilities drive capability-aware dispatch: a request
+whose required capability the originally selected backend cannot express is
+rerouted to the first qualified model in the decision whose declared
```
Xunzhuo: **[P2] Describe capability filtering before model selection**

The current router filters incapable candidates before running the configured selection algorithm; it no longer reroutes to the first compatible sibling at dispatch. For example, with compatible B then C, a higher-scoring C wins even though B appears first, as the passing `TestCapabilitiesFilterBeforeScoringAndPreserveExactWinner` regression explicitly checks. This guidance therefore gives operators the wrong way to control the selected backend. Describe capability declarations as an eligibility filter before the algorithm, and explain that final dispatch validates the chosen model without selecting a fallback.

Xunzhuo (commented):
The final-model attribution and admission/replay paths remain consistent after the current-main integration, and the relevant exact-head regressions pass. One documentation correction remains: capability checks now filter the pool before the configured selection algorithm, so the guide should no longer promise first-fit dispatch fallback.

website/docs/installation/model-configuration.md
Bevisy: > **[P2] Describe capability filtering before model selection**
> 
> The current router filters incapable candidates before running the configured selection algorithm; it no longer reroutes to the first compatible sibling at dispatch. For example, with compatible B then C, a higher-scoring C wins even though B appears first, as the passing `TestCapabilitiesFilterBeforeScoringAndPreserveExactWinner` regression explicitly checks. This guidance therefore gives operators the wrong way to control the selected backend. Describe capability declarations as an eligibility filter before the algorithm, and explain that final dispatch validates the chosen model without selecting a fallback.

Done — described as an eligibility filter ahead of selection, with final dispatch only validating the chosen model; `TestCapabilitiesFilterBeforeScoringAndPreserveExactWinner` pins it.

Xunzhuo (approved):
The documentation now matches capability filtering before selection and final dispatch validation. Selected-model and replay attribution regressions pass on this head, and the earlier dispatch lifecycle coverage remains applicable. No new blocker found; live provider deployment was not rerun.

PR #3624, opened 2026-09-08
[Feature] Route safely on PII detected in tool results

### LINKED ISSUE #3560: [Feature] Route safely on PII detected in tool results
## Goal

Expose PII detected specifically in protocol-neutral tool-result content as a routing signal so a recipe can block the follow-up request or select a configured safer model before the tool result reaches an upstream model.

## Current-main gap

- The protocol codecs already normalize Chat `role: tool`, Responses `function_call_output`, and Anthropic `tool_result` into `llmprotocol.ContentToolResult`.
- `requestSignalSnapshot` counts tool results and preserves conversation-shape facts, but deliberately excludes their payloads from signal text.
- `PIIRule` has no source or direction selector. PII evaluation scans the current user text and optional non-user history, while tool-result content is excluded from both, so a recipe cannot distinguish PII returned by a tool.
- Response-direction rules currently describe model output after model selection. Reusing that stage would make a same-request block possible only after disclosure and could not choose a safer model for the continuation.

## Scope

- Add an explicit `source: tool_result` selector to PII rules. An omitted source must preserve today's request-text behavior.
- Extract textual tool-result content from the neutral request in message and content order across all three supported API formats. Keep tool call IDs as internal provenance, but never expose raw result text or detected values through headers, metrics, logs, or Replay.
- Evaluate tool-result PII before the follow-up decision and publish the normal `pii:<rule>` match, confidence/error state, and bounded entity classes. Unsupported, malformed, partially inspected, or over-limit content must be distinguishable from a clean result.
- Allow the matched signal to compose with other request-stage signals and drive existing decision behavior: a `fast_response` policy can block before provider dispatch, while a model route can select a configured safer model pool.
- Keep rule evaluation recipe-scoped and run the detector once per unique tool-result content even when several thresholds or allow-lists consume it.
- Preserve current behavior for requests and recipes that do not configure the new source.

## Completion criteria

- [ ] Chat Completions, Responses, and Anthropic tool results produce the same source-scoped PII observation for equivalent content.
- [ ] A matching decision can return an immediate block without sending the tool result upstream.
- [ ] A matching decision can select a configured safer model, while a clean tool result follows the normal route.
- [ ] Entity classes and incomplete-scan/error state are observable without recording raw PII or raw tool-result payloads.
- [ ] Empty, structured, multimodal, oversized, repeated, and parallel tool results have deterministic bounded behavior.
- [ ] Existing PII rules and non-tool requests remain backward compatible when `source` is omitted.
- [ ] Focused config, classifier, neutral-extraction, Replay, and cross-protocol E2E tests cover both block and safe-route outcomes.

## Non-goals

- Execute tools or inspect traffic between an external tool runtime and its caller.
- Treat model-generated tool-call arguments or final assistant output as tool results.
- Retry or replace a model response after it has already been returned; that belongs to response-stage execution policy.
- Redact or rewrite tool-result payloads, persist raw detected spans, or define which model is "safe" for an operator.

Related: #2361, #3067, and the response-stage precedent in #3335.

yuki-uix:

Closes #3560

## Purpose

Add opt-in PII scanning for textual tool-result content before the next routing decision. The implementation keeps tool-result data on an explicit source path, preserves the existing prompt/history behavior, makes partial inspection visible to the decision engine, and covers the downstream privacy and cache boundaries required by the feature.

## Scope

| Area | Status | What changed | Coverage |
| --- | --- | --- | --- |
| Protocol-neutral extraction | Implemented | Extract textual tool-result content from OpenAI Chat, OpenAI Responses, and Anthropic Messages representations. Unsupported, malformed, or non-text blocks propagate an incomplete scan status. | Protocol and multimodal extraction tests |
| PII source selection | Implemented | `source: tool_result` opts a rule into tool-result content. Tool-result rules do not implicitly inspect prompt/history; omitted or empty `source` keeps the legacy scope. | Source-selection and legacy-compatibility tests |
| Detector work deduplication | Implemented | Deduplicate equal content and share cached token-classification results across PII rules. | Repeated-content and multi-rule tests |
| Tool-result fanout bound | Implemented | Apply a request-scoped budget of 1024 inference calls across unique tool-result blocks/chunks. This bounds work across both many-block and oversized single-result requests. | Many-block and oversized-result regression tests |
| Temporary chunk memory | Implemented | Stream tool-result chunks one at a time and stop immediately when the budget is exhausted, instead of materializing the complete chunk list first. | Early-stop streaming test |
| Partial-scan semantics | Implemented | Mark the remainder as `pii_evaluation_incomplete`; preserve positive matches found before the cutoff; distinguish incomplete scans from inference failures so `on_unknown` can fail closed. | Incomplete, failed, positive-match, and decision-action tests |
| Public configuration surfaces | Implemented | Preserve and validate `PII.source` through Go config, Python CLI, dashboard schemas/editors, and DSL/config compile/decompile round trips. | Config, CLI, dashboard, DSL, and documentation tests |
| Replay privacy boundary | Implemented | Remove raw tool-result payloads from Replay `RequestBody`, `Prompt`, and client-tool-result trace steps. Retain bounded tool-call IDs, API/source metadata, completion status, and an explicit redaction marker. | Replay persistence test with an email canary |
| Source-aware cache completeness | Implemented | Keep tool-result and legacy prompt/history cache entries separate when their text overlaps, so a tool-result budget cutoff cannot make a legacy scan appear complete. | Exhausted-budget overlap regression |
| Documentation | Implemented | Document the source selector, scope isolation, and related configuration behavior. | Configuration documentation and catalog checks |

The existing signal load gate remains the separate control for concurrent signal evaluations. The 1024 limit is an internal safety budget for one request and is not a claim that every deployment has the same latency target.

## Non-goals

- Redacting or rewriting PII content in the request sent to the model.
- Scanning response-stage traffic after routing.
- Executing tools or inspecting external tool traffic.
- Adding domain-specific PII taxonomies, such as medical-domain entities.
- Introducing a new configurable budget or a context-aware cancellation API.

## Test Plan

- From `src/semantic-router`: `go test ./pkg/classification ./pkg/extproc ./pkg/routerreplay/... -count=1`
- From `src/semantic-router`: `go test ./pkg/config ./pkg/dsl -count=1`
- `make check` (the repository-wide changed-file lint; it replaces the `make agent-lint` target this PR originally listed, which no longer exists on main)
- `make docs-config-check`
- `make markdown-lint`
- `cd website && npm test`
- `cd website && npm run build:en`
- Replay persistence regression proving raw tool-result/PII canaries are absent
- Source-overlap regression proving incomplete tool-result cache state does not affect legacy prompt/history scanning

## Test Result

- The post-review targeted Go suite passed:
  `go test ./pkg/classification ./pkg/extproc ./pkg/routerreplay/... -count=1`
  passed for classification, extproc, routerreplay, routerreplay/redaction, and routerreplay/store.
- The post-review configuration and DSL suite passed:
  `go test ./pkg/config ./pkg/dsl -count=1`.
- `git diff --check` passed.
- Earlier validation also passed the then-current agent lint target, configuration catalog checks, markdown lint, website tests, and the English website build.
- The Go test run emitted existing duplicate native-library linker warnings, but all affected packages passed.
- CI should be rerun after this review fix and branch synchronization.

---

<details>

- [ ] PR title begins with exactly one bracketed category, such as `[Feature]`, `[Bug]`, `[Docs]`, `[Test]`, `[Research]`, `[Community]`, or `[CI/Build]`
- [ ] The title does not stack prefixes such as `[Router][Docs]`; affected modules belong in labels and the PR body

</details>


src/semantic-router/pkg/extproc/tool_result_texts.go
```diff
@@ -0,0 +1,31 @@
...
+
+	var texts []string
+	for _, message := range req.Messages {
+		for _, content := range message.Content {
+			if content.Kind != llmprotocol.ContentToolResult || content.ToolResult == nil {
+				continue
+			}
+
+			for _, resultContent := range content.ToolResult.Content {
+				if resultContent.Kind != llmprotocol.ContentText || resultContent.Text == "" {
```
Xunzhuo: **Treat skipped tool-result content as incomplete**

Impact: non-text-only results become a clean no-match, and mixed text/media results look fully scanned, so `on_unknown` cannot fail closed for unsupported or partially inspected content.

Required change: propagate bounded extraction status for skipped/unsupported blocks into the source-scoped PII result, retain any positive text matches, and cover Responses and Anthropic multimodal tool results.

src/semantic-router/pkg/config/signal_config.go
```diff
@@ -157,6 +157,7 @@ type PIIRule struct {
 	Threshold       float32  `yaml:"threshold"`
 	PIITypesAllowed []string `yaml:"pii_types_allowed,omitempty"`
 	IncludeHistory  bool     `yaml:"include_history,omitempty"`
+	Source          string   `yaml:"source,omitempty"`
```
Xunzhuo, line 177: **Preserve `source` in every public config surface**

Impact: the Python model and dashboard do not know this field; editing and saving a `tool_result` rule can silently erase it and widen the rule back to the legacy prompt/history scope.

Required change: add the constrained field to the CLI and dashboard schemas/types/editor round trips, and test that all public config consumers preserve and validate it.

Xunzhuo (commented):
Thanks for carrying the PII source through all three wire formats. I found two safety/config round-trip gaps that need to be closed before this is ready.

yuki-uix:
Hi @Xunzhuo, I’ve pushed follow-up changes addressing both review comments:

- Incomplete tool-result extraction is now propagated into source-scoped PII evaluation. Unsupported or partially inspected content is reported as incomplete so `on_unknown` can fail closed, while positive matches are still preserved. Coverage now includes multimodal tool results in OpenAI Responses and Anthropic Messages.
- `PII.source` is now preserved and validated across the Go config, Python CLI, dashboard editor, and DSL/config round trips, so editing a `tool_result` rule cannot silently revert it to the legacy scope.

Following the same end-to-end data-flow and round-trip review, I also fixed related gaps in the adjacent extraction and public-config paths and added regression coverage. The branch has also been synced with the latest `upstream/main`.

Could you please take another look at the updated PR when you have a chance?

src/semantic-router/pkg/classification/classifier_signal_pii.go
```diff
@@ -16,28 +16,26 @@ type cachedPIIResult struct {
...
-		contentSeen[piiText] = struct{}{}
-		uniqueContents = append(uniqueContents, piiText)
-	}
 	for _, rule := range c.Config.PIIRules {
-		if !rule.IncludeHistory {
-			continue
-		}
-		for _, msg := range nonUserMessages {
-			if msg == "" {
+		for _, content := range collectPIIRuleContentsForSource(rule, piiText, nonUserMessages, toolResultTexts) {
```
Xunzhuo, line 73: **Bound tool-result PII fanout**

Impact: A valid request can carry 64 MiB across 16,384 content blocks. This path queues every roughly 512-rune tool-result chunk for synchronous classification, so one continuation can trigger tens of thousands of model calls before routing and does not meet the deterministic boundedness required by #3560.

Required change: Enforce a request-scoped byte, chunk, or inference budget while collecting and scanning, mark the remainder incomplete, and cover oversized and many-block requests.

Xunzhuo (commented):
Thanks for the careful cross-protocol coverage. One request-boundedness gap still needs to be closed.

src/semantic-router/pkg/classification/classifier_signal_pii.go
yuki-uix: I’ve pushed a follow-up for this boundedness concern in commit d4ef8b15, and synced the branch with the latest upstream/main.

Tool-result PII scanning now applies a request-scoped inference budget of 1024 calls across unique tool-result content and chunks. The scanner streams chunks and stops as soon as the budget is exhausted, instead of materializing and queuing the full chunk list. The remaining content is marked incomplete so an on_unknown policy can fail closed, while matches found before the cutoff are preserved.

I also added regression coverage for many-block requests, oversized single tool results, and early-stop streaming behavior. The existing signal load gate remains the separate control for concurrent signal evaluations. The PR description now includes a scope matrix with the covered paths and remaining non-goals. Could you please re-review the updated boundedness behavior?

yuki-uix:
Hi @Xunzhuo, for tracking clarity, I’ve **updated the PR description with a Scope table** covering the areas we discussed:

- protocol-neutral tool-result extraction
- source-scoped PII selection
- deduplication and shared detector results
- bounded inference fanout and temporary chunk memory
- incomplete-scan and `on_unknown` semantics
- public configuration surfaces and round-trip coverage
- regression tests and documentation

The description also lists the current non-goals, such as redaction, response-stage scanning, external tool traffic, and domain-specific PII taxonomies.

If there are still additional concerns, could we clarify whether they are required for #3560’s acceptance criteria and should be addressed in this PR, or whether they would be better tracked as a separate follow-up issue? This would help keep the current PR focused while ensuring remaining hardening work is not lost.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
```diff
@@ -38,14 +38,18 @@ func (r *OpenAIRouter) extractRequestSignalSnapshot(
...
 ) (requestDecisionState, *ext_proc.ProcessingResponse) {
 	if !ctx.Routing.IsResolved() {
 		r.resolveEntrypointForRequest(originalModel, ctx)
 	}
 	populatePinnedSessionFromHeaders(ctx)
-	history := signalConversationHistoryFromSnapshot(snapshot)
+	// Keep the general snapshot payload-free, but attach tool-result text at the
+	// signal-history boundary so opt-in PII rules can inspect it. This is the
+	// only pre-routing path for normal request-body processing.
+	history := signalConversationHistoryFromRequest(request, snapshot)
```
Xunzhuo, line 52: **Do not persist raw tool-result PII in Replay**

This source extraction feeds raw tool-result text into the decision path, but Replay later stores the full semantic request and copies each tool result into `Text`, `Output`, and `RawOutput`. That persists the detected payload and exposes it to replay-detail readers, contrary to #3560's requirement not to record raw tool results or PII. Omit or redact those payload fields before Replay persistence while retaining bounded IDs, classes, status, and errors, and add a regression proving a PII canary is absent from both the stored request body and tool trace.

src/semantic-router/pkg/classification/classifier_signal_pii.go
```diff
@@ -17,40 +17,86 @@ type cachedPIIResult struct {
...
+	piiCache := make(map[string]cachedPIIContent, len(uniqueContents))
+	toolResultBudget := piiToolResultScanBudget{remainingInferenceCalls: maxPIIToolResultInferenceCalls}
 	for _, content := range uniqueContents {
-		chunks := piiSignalChunks(content)
-		cached := make([]cachedPIIResult, 0, len(chunks))
-		for _, chunk := range chunks {
-			tokenResult, err := c.piiInference.ClassifyTokens(ctx, chunk)
-			cached = append(cached, cachedPIIResult{tokenResult, err})
+		cached := cachedPIIContent{}
+		_, isToolResult := toolResultContentSeen[content]
```
Xunzhuo: **Keep cache completeness source-specific**

If the same string appears in a tool result and in prompt/history, this content-only marker makes their single cache entry use the tool-result budget. Once earlier tool chunks exhaust that budget, the legacy rule can reuse an incomplete or unscanned entry and miss PII that previously received a full legacy scan, making omitted-source behavior depend on rule order and source overlap. Preserve a full legacy scan for any legacy-selected content, or key cache status by source and completeness, and add an overlap regression with an exhausted tool budget.

Xunzhuo (commented):
Thanks for resolving the earlier source and fanout issues. Two privacy and backward-compatibility gaps remain.

src/semantic-router/pkg/extproc/processor_req_body_prepare.go
yuki-uix: Implemented in commit 67e901a94.

Replay now clones the semantic request and removes tool-result content before storing RequestBody or deriving Prompt. Client tool-result trace steps retain the tool call ID, source/API metadata, completion status, and ContentRedacted marker, but no longer store Text, Output, or RawOutput.

Added a regression test with an email canary to verify that raw tool-result content is absent from Replay storage. The targeted Go tests pass.

Please re-review and resolve this thread if the fix addresses the concern.

src/semantic-router/pkg/classification/classifier_signal_pii.go
yuki-uix: Implemented in commit 67e901a94.

The PII cache is now keyed by both source and content. Therefore, identical text selected by a tool-result rule and a legacy prompt/history rule receives independent cache entries and completeness state.

An exhausted tool-result scan can no longer make the overlapping legacy scan appear complete. Added a regression test covering this exact budget-exhaustion overlap case. The targeted classification tests pass.

Please re-review and resolve this thread if the fix addresses the concern.

yuki-uix:
@Xunzhuo Rebased onto main — the branch is mergeable again.

The conflict wasn't ordinary drift. It came from the modelruntime/tasks token-classification refactor in #3768, which changed both the Go API and the native binding ABI, so the three affected files needed real resolution rather than a fast-forward:

- `classifier_signal_pii.go` now uses `tasks.TokenClassificationResult` and `classifyPIITokens`, keeping the source-scoped cache key, the incomplete-vs-failed distinction, and the per-request inference budget intact.
- `classifier_pii_test.go` takes the upstream mock, with the call-count instrumentation the fanout tests depend on re-applied.
- `classifier_signal_pii_source_test.go` migrates to `tasks.TokenEntity`.

Verified locally after rebuilding the native bindings (the stale candle-binding was missing the new `candle_instance_*` symbols):

    go test ./pkg/classification ./pkg/extproc ./pkg/routerreplay/... -count=1
    go test ./pkg/config ./pkg/dsl -count=1
    make docs-config-check
    make markdown-lint

All pass. I also ran the budget, legacy-cache-completeness, and source-scope tests individually to confirm they execute rather than skip. The diff against main is unchanged at 50 files, +1655/-154, so nothing was dropped or absorbed during the merge.

One correction to the Test Plan in the PR description: it listed `make agent-lint`, which no longer exists on main. I've updated the description to `make check`.

All five review threads from the earlier rounds are resolved, and the `pr/needs-rebase` label is now stale. Ready for another look whenever you have time.

FAUST-BENCHOU:
The earlier safety gaps look properly closed now. `tool_result` stays isolated as an explicit source across the public config surfaces, while unsupported/truncated inspection propagates as incomplete instead of silently becoming a clean result, preserving fail-closed `on_unknown` behavior. LGTM from my review.

src/semantic-router/pkg/routerreplay/store/store.go
```diff
@@ -95,7 +95,12 @@ type ToolTraceStep struct {
 	Text       string `json:"text,omitempty"`
 	ToolName   string `json:"tool_name,omitempty"`
 	ToolCallID string `json:"tool_call_id,omitempty"`
-	Arguments  string `json:"arguments,omitempty"`
+	// Status preserves whether a tool result completed successfully after its
+	// payload has been removed from replay storage.
+	Status string `json:"status,omitempty"`
+	// ContentRedacted makes payload removal explicit to replay consumers.
+	ContentRedacted bool   `json:"content_redacted,omitempty"`
```
Xunzhuo, line 104: [P2] Publish the new ToolTraceStep fields in OpenAPI

Please regenerate and commit the public OpenAPI contract with these new replay fields. The exact-head Generated Contracts job reports that apiserver.openapi.json is stale: both ToolTraceStep schema locations omit status and content_redacted. The API can now return these fields, while generated clients and the published API reference still see the old shape. Run make api-docs-openapi and commit the generated result.

Xunzhuo (changes requested):
The merge preserves the source-isolated PII handling and error policy, and all 61 selected current-head tests pass. The new Replay fields are missing from the public OpenAPI artifact, which the Generated Contracts job confirms. Please regenerate and commit that contract. A full deployed PII-to-Replay integration run was not repeated.

src/semantic-router/pkg/routerreplay/store/store.go
yuki-uix: Thanks for catching this. I regenerated the OpenAPI artifact and pushed `ad0c7210`.

The missing Replay fields, including `content_redacted` and the tool-result status fields, are now present in `website/static/openapi/apiserver/apiserver.openapi.json`. This follow-up only updates the generated artifact; no Go source, route, or schema definition changed.

I verified the update with:

```text
make generated-contract-check
go test ./pkg/classification ./pkg/extproc ./pkg/routerreplay/... -count=1
```

`make generated-contract-check` passed all five sub-checks, and all selected Go packages passed. The duplicate native-library linker messages were non-fatal warnings.

I did not rerun the deployed PII-to-Replay integration test because I do not currently have a configured Kubernetes deployment environment available for this validation. The local unit tests for the redaction and replay paths pass, and the follow-up only changes the generated OpenAPI artifact.

Xunzhuo (dismissed):
The generated Replay contract now includes both tool-result fields, resolving the remaining finding. Rechecked the source-isolated PII handling and current-main integration; all 61 focused Core regressions and the generated-contract checks pass. No remaining code blocker found. A separate local model-loaded deployment was not rerun.

PR #3626, opened 2026-09-08
[Feature] Add Azure Astra Responses catalog routes

### LINKED ISSUE #3563: [Feature] Add GPT-6 Astra to Azure OpenAI built-in support
## Maintainer direction

I recommend accepting this as a bounded Data Plane & Networking feature. #3561 establishes GPT-6 Astra's canonical model card and direct OpenAI contract; Azure OpenAI needs its own provider binding because Azure deployment names, authentication, paths, and API-version handling are distinct runtime semantics.

## Goal

Make `openai/gpt-6-astra` selectable through the built-in `azure-openai` provider with verified Chat Completions and Responses behavior, while keeping the operator's Azure deployment name explicit.

## Current-main gap

- `config/catalog/resources/providers/azure-openai.yaml` currently advertises only `openai/chat-completions@1#create` and has no catalog model bindings.
- Microsoft's current reasoning-model documentation lists GPT-6 Astra version `2026-09-03` for both Chat Completions and Responses, with a 1,050,000-token context window and 128,000 maximum output tokens: https://learn.microsoft.com/en-ca/azure/foundry/openai/how-to/reasoning
- The direct OpenAI binding in #3561 cannot safely stand in for Azure's operator-defined deployment name or Azure-specific request construction.

## Scope

- Add the verified `openai/responses@1` operation and effective Azure request path/API-version behavior without regressing the existing Chat Completions path or `api-key` authentication.
- Add an `openai/gpt-6-astra` managed-cloud binding. Declare `provider_model_id_kind: deployment_name` and preserve the canonical catalog name separately; never invent or silently derive an operator's deployment name.
- Record only protocol capabilities and reasoning efforts verified for Azure. In particular, prove Chat `xhigh` and Responses `max` rather than copying the direct OpenAI binding without provider evidence.
- Add provider-boundary fixtures that assert the final Azure URL/query, headers, deployment identifier, model/request body, reasoning field, and tool preservation for Chat and Responses.
- Regenerate the catalog projections and update the Day-0 guide if Azure introduces a reusable managed-deployment step not already documented by #3561.

## Completion criteria

- [ ] A catalog-backed user configuration with an explicit Azure deployment name materializes without hand-written model-card metadata.
- [ ] Chat Completions with `xhigh` reaches the Azure fixture with exactly one valid reasoning representation.
- [ ] Responses with `max` and a tool reaches the Azure fixture with the correct path, API version, deployment semantics, and preserved tool schema.
- [ ] Unsupported protocol/effort combinations fail before dispatch with a useful validation error.
- [ ] Catalog validation, materialization tests, provider conformance tests, generated projections, and public documentation pass their owning gates.

## Non-goals

- Discover, create, or rename Azure deployments.
- Generalize every Azure OpenAI model or redesign Azure authentication.
- Change the canonical GPT-6 Astra card or direct OpenAI behavior delivered by #3561.

Reference implementation: #3561.

1fanwang:
Closes https://github.com/vllm-project/semantic-router/issues/3563

## Purpose

Azure operators cannot select Responses through the built-in provider. This adds the Astra deployment binding and operation-specific routing while retaining the operator's deployment name. Chat uses its deployment path and API-version query; Responses uses the host-root route without that query.

Dispatch consumes the path and query policy together. Both catalog loaders reject overlapping overrides. Live Azure availability remains claimed, not verified.

## Test Plan

From the repository root, build the CPU libraries and run the committed provider-dispatch fixture:

```bash
make rust-ci
export CGO_LDFLAGS="-L$PWD/candle-binding/target/release -L$PWD/onnx-binding/target/release -L$PWD/ml-binding/target/release -L$PWD/nlp-binding/target/release"
export LD_LIBRARY_PATH="$PWD/candle-binding/target/release:$PWD/onnx-binding/target/release:$PWD/ml-binding/target/release:$PWD/nlp-binding/target/release"
cd src/semantic-router
go test -mod=readonly -p 4 -count=1 -timeout 8m -run "^TestAzureOpenAIProviderDispatch$" -v ./pkg/extproc
```

## Test Result

The fixture used deployment astra-prod, Chat API version 2024-10-21, and a local HTTP server. Real Router decoding, dispatch and encoding emitted:

```text
path=/openai/deployments/astra-prod/chat/completions?api-version=2024-10-21 api-key=azure-fixture-key model=astra-prod reasoning=map[reasoning_effort:xhigh] tools=null
path=/openai/v1/responses api-key=azure-fixture-key model=astra-prod reasoning=map[reasoning:map[effort:max]] tools=[{"description":"Look up weather","name":"lookup_weather","parameters":{"additionalProperties":false,"properties":{"city":{"type":"string"}},"required":["city"],"type":"object"},"strict":true,"type":"function"}]
```

The server returned HTTP 200 for both requests. The key is a fixture value. This verifies the internal provider boundary, not a live Azure service or Envoy deployment.

---
<details>

</details>


config/catalog/resources/providers/azure-openai.yaml
```diff
@@ -5,12 +5,16 @@ category: model_api
...
 protocols:
 - openai/chat-completions@1
+- openai/responses@1
 default_protocol: openai/chat-completions@1
 supported_operations:
 - openai/chat-completions@1#create
+- openai/responses@1#create
 path_overrides:
   openai/chat-completions@1#create: /chat/completions
+  openai/responses@1#create: /responses
```
Xunzhuo: **Use the Azure v1 Responses route**

Impact: this override combines the deployment-scoped legacy base URL with `/responses` and the test locks in `/openai/deployments/astra-prod/responses?api-version=2026-09-03`. Azure's v1 contract uses `/openai/v1/responses`, sends the deployment name as `model`, and does not use that query; `2026-09-03` is the Astra model version. Requests through the newly advertised Responses protocol therefore target an invalid route.

Required change: model the v1 Responses path/version semantics without regressing legacy Chat, then add a provider-boundary fixture that asserts the final path/query, `api-key`, `model: astra-prod`, reasoning shape, and preserved tool schema.

Xunzhuo (commented):
Thanks for adding Astra to the provider catalog; the Responses transport still needs one protocol-boundary correction.

config/catalog/resources/providers/azure-openai.yaml
1fanwang: Done in https://github.com/vllm-project/semantic-router/pull/3626/commits/74b45f47e8e0c067a2794f43625759e501fa45d7.

src/semantic-router/pkg/catalog/types.go
```diff
@@ -155,6 +170,7 @@ type ProviderDefinition struct {
 	DefaultProtocol     string                         `json:"default_protocol"`
 	SupportedOperations []string                       `json:"supported_operations"`
 	PathOverrides       map[string]string              `json:"path_overrides,omitempty"`
+	OperationOverrides  map[string]OperationOverride   `json:"operation_overrides,omitempty"`
```
Xunzhuo, line 173: **Clone the new override map**

Impact: `Registry.Provider` and `EffectiveRegistry.Provider` promise immutable defensive views, but `cloneProvider` leaves `OperationOverrides` aliased. Mutating a returned provider therefore changes the shared registry; later path resolution can lose the Azure v1 override, and concurrent reads can race or panic.

Required change: Clone `OperationOverrides` in `cloneProvider` and add a regression proving that mutating one lookup cannot affect a later lookup.

Xunzhuo (commented):
Thanks for correcting Azure's route semantics. The new override state still escapes the registry's defensive-copy boundary.

src/semantic-router/pkg/catalog/types.go
1fanwang: Done in https://github.com/vllm-project/semantic-router/pull/3626/commits/dd2d1cb073d1dad687da75f763720d1b007ae896. Added a regression that mutates one lookup and checks the next is unaffected.

Peterren:
Thanks for adding Azure GPT-6 Astra. The Azure difference is real: legacy Chat uses a deployment-scoped path plus `api-version`, while Responses uses the host-root `/openai/v1/responses`, carries the deployment name in `model`, and omits that query. This is not a new protocol; it is an operation-specific binding of our existing Responses protocol, and keeping it in the catalog rather than adding an Azure branch in ExtProc is the right direction.

I suggest making the new boundary atomic before we use Azure as its first consumer. Today the caller resolves the path and then separately looks up `OperationOverride` to decide the query policy. That makes it possible for a future caller to get the correct path but accidentally retain the wrong `api-version` behavior.

A small result is enough for this PR; there is no need to design every future provider field:

```go
type ResolvedOperation struct {
    Path               string
    UseAPIVersionQuery bool
}

op, err := registry.ResolveOperation(providerID, protocolID, "create", basePath)
```

`ResolveOperation` can normalize the existing `path_overrides`, provider-level `api_version_query`, and the new operation-specific declaration into this one runtime value. Please also reject a key appearing in both `path_overrides` and `operation_overrides`, so one operation never has two authored sources of truth. Existing providers do not need to migrate in this PR.

The benefit is small but important: dispatch consumes one complete operation contract, Azure does not leak into runtime conditionals, and future real cases—such as Anthropic supporting operations or native Vertex/Bedrock bindings—can extend the same boundary without adding another parallel flag.

Please also make the Azure regression exercise the actual provider-dispatch boundary, not only catalog metadata: assert the final path/query, emitted `api-key`, body `model == "astra-prod"`, reasoning shape, and preserved tool schema. Use an Azure API-version fixture for Chat that is clearly distinct from Astra's model version `2026-09-03`.

Finally, please rebase and regenerate before the next review pass. Most of the apparent diff is stale generated/schema churn; the hand-written runtime change is small and should remain easy to review. Azure-specific transport-error fields can be a bounded follow-up—this PR does not need to fix every provider or operation.

1fanwang:
Done in https://github.com/vllm-project/semantic-router/commit/bccadc70319189806dacd7400507c5287b931431. Full provider-dispatch/E2E verification is still pending, so this remains a draft.

Xunzhuo (approved):
The Azure Responses route and defensive-copy findings are resolved. Exact-head CI passed the Chat and Responses provider-boundary fixtures, including path/query, deployment model, credentials, reasoning and tool schema; the override-conflict regression also passes. This verifies the internal provider boundary. Live Azure availability remains claimed.

wilsonwu (approved):
LGTM

PR #3628, opened 2026-09-08
[Bug] Stop training the intent classifier on the MMLU-Pro rows it is scored on

### LINKED ISSUE #3558: [Bug] Intent classifier trains on the MMLU-Pro test split
### Proposed Workgroup

### Describe the bug

`ft_linear_lora.py:196-197` builds the training pool from MMLU-Pro's `test` split:

```python
all_texts  = list(dataset["test"]["question"])
all_labels = list(dataset["test"]["category"])
```

It selects ~5,000 of the 12,032 rows at the recommended `--max-samples`, and 60% of those
become gradient training data. So roughly 3,000 rows of the split named `test` are training
data, and `test` is not a held-out set for this model.

### To Reproduce

On main at a3092868, `ft_linear_lora.py:196-197`. MMLU-Pro ships `validation` (70 rows) and
`test` (12,032 rows); only `test` is read.

Scoring `mmbert32k-intent-classifier-merged` over all 12,032 rows, split
by whether the trainer would have selected them:

| slice | rows | accuracy |
|---|---|---|
| gradient-trained | 2,998 | 0.9510 |
| never selected | 7,034 | 0.9416 |

### Expected behavior

Training and evaluation should not use the same split. Since the model trains on `test`,
accuracy reported on `test` is not held-out evidence and should not be quoted as such.

### Impact

Any accuracy reported for the intent classifier on MMLU-Pro `test` is optimistic by roughly
1 point. Small, but it means the split cannot be used as held-out evidence for a candidate
model without stating the overlap.

### Additional context

Found while reproducing #1485. Not the cause of the conversational gap reported there; this
is a correctness issue about what the number means.

yaojiejia:
Closes #3558

## Purpose

MMLU-Pro publishes only `validation` (70 rows) and `test` (12,032 rows). The intent
classifier trained on `test` and was also scored on `test`, so its accuracy was measured
partly on questions it had already seen.

The trainer now sets aside 20% of `test` before it picks any training data, taking the same
share from every category. Those rows are never trained on. After training, the model is
scored on them, and the row numbers and scores are saved to `heldout_eval.json` so anyone
can check where the number came from. The merge step copies that file into the merged model.

Setting aside rows is not enough, because MMLU-Pro asks some questions twice. Rows 6094 and
6095 are the same question, and the split sets aside 6094 while leaving 6095 in training, so
the model would train on a question it is then scored on. 64 questions land on both sides
this way. Any training row asking a set-aside question is therefore left out as well.

The supplement dataset is also merged into training, so the same check runs over it. It
drops nothing today (0 of 653 rows), and is there so a future supplement update cannot
reintroduce a set-aside question.

## Test Plan

```bash
make test-training-contracts
```

Plus two offline harnesses against the real cached MMLU-Pro (not committed): the full data
path asserting training and reserved rows are disjoint by text, and a `main()` run on CPU
with `all-MiniLM-L6-v2` standing in for the base model.

## Test Result

`make check` passes.

This adds 19 contract tests in `classifier_model_fine_tuning_lora/tests/test_heldout_split.py`.
They run inside `make test-training-contracts`, under the stdlib-only `python3` that target
uses, so they execute in CI instead of skipping.

To confirm the tests can actually fail, each was run against a deliberately broken copy of
the split:

| broken behaviour | tests that fail |
|---|---|
| reserve by row index only (the original bug) | 3 |
| ignore categories when sampling | 4 |
| reserve a fixed prefix instead of shuffling | 3 |
| ignore the seed | 1 |
| stop dropping supplement repeats | 1 |
| skip input validation | 1 |

Measured against the real 12,032-row `test` split:

- 2,407 rows are reserved, and no question appears in both the training data and the
  reserved slice. Before the text-level exclusion, 39 did.
- 88 pool rows are excluded for repeating a reserved question. At the recommended
  `--max-samples 5000` that costs one training example (4,917 sampled rows becomes 4,916).
- Rerunning reserves exactly the same rows.
- A full `main()` run trained, evaluated on the reserved rows, and wrote a valid
  `heldout_eval.json`.

---
<details>

</details>


Xunzhuo (approved):
Good catch

Issue #3174, opened 2026-08-31
[Feature] Red-team evasion benchmark and CI graduation gate for classifier signals


guan404ming:
### Goal

Split from #3164 (benchmark scope only; the runtime policy change is a separate sub-issue).

Router classifier signals are treated as a safety boundary, but a small classifier is not adversarially robust. A black-box evasion test (query the model, read the score, no gradients) against the mmBERT-32K jailbreak detector, JailbreakBench 100 harmful behaviors, threshold 0.7:

| metric | value |
|---|---|
| baseline detected as jailbreak | 39/100 (39%) |
| flipped to benign by the attack | 39/39 (100% of detected) |
| benign words appended to flip | mean 2.6, median 2, max 9 |
| queries per prompt | ~92 |

Appending a few harmless words (`please`, `weather`, `recipe`) flips every detected prompt to benign; the other 61 were missed at baseline. This is the weakest attack (greedy search over a 35-word pool), so it is a lower bound.

There is currently no offline benchmark that measures this, and no gate that stops a model with these numbers from graduating.

### Proposed Workgroup

Evaluation & Quality

### Scope

- A red-team benchmark in `bench/`: black-box evasion runner, parameterized by model, dataset, threshold, and word pool; reports flip rate, baseline recall, suffix length, and query count. Offline, no request-path change.
- A CI red-team lane: a router model must pass the attack benchmark before graduation (#2974 criteria).

### Non-goals

- Any runtime behavior change (`route` action, fail-closed handling); tracked in the companion sub-issue split from #3164.
- Model retraining and adversarial training to fix the 39% recall; separate follow-up under #2974.
- Downgrade attacks on the category/intent classifier (route steering); a later benchmark on the same harness.

### Completion signal

- The benchmark runs in CI and reports flip rate and recall for a target model.
- Graduation criteria under #2974 reference the benchmark thresholds.
- Baseline numbers for the current jailbreak detector are recorded on the issue.

Parent Epic: #2333

PR #3631, opened 2026-09-08
[Feature] Add a red-team evasion benchmark for classifier signals

1fanwang:
Related #3174

## Purpose

Router classifier signals are treated as a safety boundary, but nothing in the repo measures whether a small classifier survives an adversary who can read its score. There is no offline benchmark for it and no gate that would stop a weak model from graduating.

This adds a black-box evasion runner under `bench/redteam/`. It appends benign filler words to a flagged prompt until the signal falls below its threshold, then reports baseline recall, flip rate, suffix length, and query count. Nothing on the request path changes.

The runner takes a `Scorer` protocol, so it works for any signal that exposes a probability. `--min-baseline-recall` and `--max-flip-rate` make it exit non-zero, which is what a graduation gate needs.

I am not adding the CI lane here. Every runner in this repo is `ubuntu-latest`, so a real red-team lane needs a hardware decision that belongs in its own change.

## Test Plan

```bash
# unit tests, no model needed
python3 -m pytest bench/redteam/test_attack.py -q

# measured baseline, needs the artifact and a GPU
make download-mmbert
python3 -m bench.redteam.evaluate \
    --model models/mmbert32k-jailbreak-detector-merged \
    --dataset jailbreakbench --threshold 0.7
```

## Test Result

The attack logic is covered by unit tests against a stub scorer, so it runs without a model.

Measured on one NVIDIA H100 80GB against the JailbreakBench harmful split at threshold 0.7, the detector flags 39 of 100 behaviors, and a median of two harmless words defeats 34 of those 39. The 39 of 100 baseline matches the number reported on the issue. My flip rate is lower than the 39 of 39 reported there: raising `--max-words` from 10 to 30 changes nothing, so the five survivors are a local minimum of this greedy search rather than a budget limit, and a wider pool would likely take them.

<details>
<summary>Raw logs</summary>

```text
$ python3 -m pytest bench/redteam/test_attack.py -q
9 passed in 0.06s

$ python -m bench.redteam.evaluate --model ... --threshold 0.7
scoring 100 prompts on cuda
{
  "total": 100,
  "detected": 39,
  "baseline_recall": 0.39,
  "flipped": 34,
  "flip_rate": 0.8718,
  "suffix_mean": 2.44,
  "suffix_median": 2.0,
  "suffix_max": 10,
  "queries_mean": 35.2,
  "threshold": 0.7
}
NVIDIA H100 80GB HBM3, driver 580.105.08
```

</details>


bench/redteam/datasets.py
```diff
@@ -0,0 +1,59 @@
...
+    found = [prompt for prompt in prompts if prompt]
+    if not found:
+        raise ValueError(f"no prompts found in {path}; expected one of {GOAL_KEYS}")
+    return found
+
+
+def load_jailbreakbench() -> list[str]:
+    from datasets import load_dataset
+
+    dataset = load_dataset(JAILBREAKBENCH_REPO, JAILBREAKBENCH_SUBSET, split="harmful")
```
Xunzhuo, line 50: **[P1] Load the actual JailbreakBench column and reject an empty corpus**

The [default dataset](https://huggingface.co/datasets/JailbreakBench/JBB-Behaviors) exposes `Goal` with a capital G, while `_first_goal` only accepts lowercase keys. Every row is discarded, so the documented default command evaluates zero of the 100 behaviors and returns success, including with `--max-flip-rate 0`. A fixture using the actual column names reproduces the empty corpus and successful exit. Read the real schema, fail before scoring an empty corpus, and cover this loader path.

Xunzhuo (changes requested):
The default benchmark silently evaluates an empty corpus and can pass its evasion gate without scoring any behavior. Please fix the dataset schema handling and add the loader regression before merge.

bench/redteam/datasets.py
1fanwang: Done in https://github.com/vllm-project/semantic-router/commit/1c57361d4a8816e4cf6a873cb3cd153a85f1934f.

1fanwang:
@Xunzhuo The default run now scores all 100 JailbreakBench behaviors, and an empty corpus fails before any model loads. Could you take another look when you have time?

PR #3643, opened 2026-09-09
[Bug] Scan the whole answer for hallucinations, not its first window

### LINKED ISSUE #3641: [Bug] Hallucination detection skips the part of an answer past the 512-token window
### Proposed Workgroup

### Describe the bug

`detect_hallucinations` windows the context so the answer survives right truncation, which is what #3365 fixed (`candle-binding/src/ffi/classify.rs:1704`). The budget is `max_length - suffix_tokens` (`candle-binding/src/core/tokenization_window.rs:12`) and the suffix is `" Question: {question} [SEP] {answer}"`. Once that suffix alone reaches the model's 512-token window (`candle-binding/src/core/tokenization.rs:207-210`, ModernBERT), the budget is zero and the answer itself is right-truncated. Whatever sits past the window is never scored, so a long answer whose unsupported statement is near the end returns `has_hallucination = false` with no spans.

Both consumers reach this path, the plugin at `src/semantic-router/pkg/classification/hallucination_detector.go:103` and the response-stage signal added in #3531. `detect_hallucinations_with_nli` windows its premise the same way (`candle-binding/src/ffi/classify.rs:1951`) and is not measured here.

### To Reproduce

Model `models/mom-halugate-detector`, CPU, threshold 0.5, on `e17d7b48`. The context gives 1889 and 330 metres in three sentences. The answer is grounded filler plus one unsupported sentence, "The tower was completed in 1912 and it is 550 metres tall.", placed either first or last.

| answer tokens | words | claim first | claim last |
|---|---|---|---|
| 463 | 352 | detected 0.9966 | detected 0.9981 |
| 508 | 386 | detected 0.9966 | not detected, 0 spans |

Controls from the same run: the unsupported sentence scored on its own is detected at 0.9993, and it is still detected against an empty context at 0.9985, so neither the sentence nor the grounding is what goes missing. A 2,644 word context with a short answer is detected at 0.9984, which is #3365's fix working.

### Expected behavior

An answer past the window is scanned in full, the way the response jailbreak path scans chunks and takes the maximum (#3333), or the truncation is reported so that a clean verdict on a truncated answer is distinguishable from a clean verdict on a scored one (#3372).

subin9:
Closes #3641

## Purpose

`detect_hallucinations` windows the context so the answer survives right truncation (#3365). The budget is `max_length - suffix_tokens` (`candle-binding/src/core/tokenization_window.rs:12`) and the suffix is `" Question: {question} [SEP] {answer}"`. Once that suffix alone reaches the model's 512 token window, the budget is zero and the answer itself is right truncated, so anything past the window is never scored and an unsupported claim near the end of a long answer comes back with no spans.

Measured on `models/mom-halugate-detector` at threshold 0.5, before this change:

| answer tokens | words | claim first | claim last |
|---|---|---|---|
| 463 | 352 | detected 0.9966 | detected 0.9981 |
| 508 | 386 | detected 0.9966 | not detected, 0 spans |

`Detect` and `DetectWithNLI` now scan a long answer in bounded overlapping chunks through `securitySignalChunks`, the helper the response jailbreak path already uses to scan a long body (#3206, #3333). A hallucination in any chunk is a hallucination in the answer, the reported confidence is that of the strongest chunk that found one, a clean answer reports its least confident chunk, and a span repeated across the 64 rune overlap is kept once. The chunk budget is 256 units, which leaves the rest of the window for the context and the question. An answer that already fits the window makes exactly the single call it made before.

This is the local detector only. `EndpointHallucinationDetector` is a separate type whose remote backend has no such window, and it is untouched.

## Test Plan

`go test ./pkg/classification/` with the hallucination and NLI models present, `scripts/gates.sh` for the CI PR lane, and the `hallucination` E2E profile on a local kind cluster.

## Test Result

Three new tests. `TestHallucinationDetector_LongAnswerTailIsScanned` and `TestHallucinationDetector_LongAnswerTailIsScannedWithNLI` build a 672 word answer whose only unsupported sentence is its last one. `TestMergeChunkConfidence` covers the fold from chunk verdicts to one answer verdict and needs no model.

On the parent commit the same fixture reports `detected=false confidence=1.000 spans=[]`. With this change it reports `detected=true confidence=0.999` with the unsupported sentence among the spans, and the NLI entry point reports `detected=true confidence=0.999` with 8 spans.

```
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/classification	72.953s
```

The `hallucination` E2E profile passes on this branch, 5 of 5 responses warned. That profile serves its detector from a remote mock endpoint, so it is a regression check rather than a test of this fix. An E2E case that can fail on this defect needs a profile running the local detector, which no profile does today, and that belongs in its own change.

### DIFF
--- a/src/semantic-router/pkg/classification/hallucination_detector.go
+++ b/src/semantic-router/pkg/classification/hallucination_detector.go
@@ -18,6 +18,67 @@ type HallucinationResult struct {
 	SupportedSpans        []string `json:"supported_spans,omitempty"`   // Text spans grounded in context
 }
 
+const (
+	// The detector scores at most 512 tokens and windows only the context, so an
+	// answer longer than that window is truncated and its tail is never scored.
+	// A long answer is scanned in bounded overlapping chunks instead, the way the
+	// response jailbreak path scans a long body. The budget leaves the rest of
+	// the window for the context and the question.
+	hallucinationAnswerChunkBudget  = 256 * 4
+	hallucinationAnswerOverlapRunes = 64
+)
+
+// hallucinationAnswerChunks returns the pieces of an answer to score. An answer
+// that fits the window yields one chunk, which is the single call the detector
+// made before.
+func hallucinationAnswerChunks(answer string) []string {
+	return securitySignalChunks(answer, hallucinationAnswerChunkBudget, hallucinationAnswerOverlapRunes)
+}
+
+// detectHallucinationsInChunks scores every part of the answer against the same
+// context. A hallucination in any chunk is a hallucination in the answer, the
+// reported confidence is that of the strongest chunk that found one, and a clean
+// answer reports its least confident chunk. Chunks overlap, so spans that repeat
+// are kept once.
+func detectHallucinationsInChunks(context, question, answer string, threshold float32) (*candle.HallucinationDetectionResult, error) {
+	chunks := hallucinationAnswerChunks(answer)
+	if len(chunks) <= 1 {
+		return candle.DetectHallucinations(context, question, answer, threshold)
+	}
+
+	merged := &candle.HallucinationDetectionResult{Confidence: 1}
+	seen := make(map[string]struct{})
+	for _, chunk := range chunks {
+		result, err := candle.DetectHallucinations(context, question, chunk, threshold)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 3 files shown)


Xunzhuo (approved):
Thanks

PR #3650, opened 2026-09-09
[Bug] Search a long RAG query by every window, not its first

### LINKED ISSUE #3385: [Bug] RAG retrieval embeds only the first 512 tokens of a query
### Proposed Workgroup

### Describe the bug

The three RAG retrieval paths embed the query with `GetEmbedding(query, 0)`:

```
req_filter_rag_milvus.go:52
req_filter_rag_qdrant.go:58
req_filter_rag_external.go:232   (buildPineconeRequest, ctx.UserContent)
```

That resolves to `max_length.unwrap_or(512)` in `core/similarity.rs:161`, so only the first 512 tokens of a query are embedded. Two different questions that share a long preamble produce the same vector, and the store returns the same documents for both.

### To Reproduce

`mom-embedding-light` on CPU, through `GetEmbedding(query, 0)`. Both queries carry the same filler prefix and then differ: one asks about a refund policy, the other about disabling the audit log.

```
prefix repeats   runes    cos(q1, q2)
20               1096     0.449117
100              5256     1.000000
400              20856    1.000000
6000             312056   1.000000
```

The two tails on their own are `cos = 0.465835`, so they are genuinely different questions. From roughly 512 tokens of shared prefix onward the vectors are identical to six decimal places, meaning the part of the query that carries the actual question was never read.

### Expected behavior

A query longer than the embedding model's window still retrieves for the question it asks, or the router says it cannot.

### Impact

Any deployment with the RAG plugin enabled, on all three backends. A long system preamble, a pasted document, or accumulated history ahead of the user's question is enough. Retrieval silently returns context for a different question, and that context is then injected into the prompt, so the failure reaches the model rather than stopping at the retriever.

There is no configuration workaround. The window is `unwrap_or(512)` in the binding, and the caller passes `0`.

### Additional context

Same root as #3366, which covers the semantic cache. This is a different surface with a different consequence: #3366 returns a cached answer for the wrong question, this retrieves documents for the wrong question. The two are fixed in different places — `cache.go` there, the three RAG filters here.

Worth noting the fix is not obvious, which is why this is filed as its own issue rather than folded into #3366. The BERT model has 512 position embeddings, so embedding the whole query is not available without changing models. The options differ in kind: sample the query the way `representativeSignalText` does for routing signals, use a longer-context model for RAG queries, or reject a query that exceeds the window. Which one is right depends on the answer to #3373.

The `// Auto-detect dimension` comment at `req_filter_rag_milvus.go:52` describes the wrong parameter — the argument is `maxLength`, not a dimension.

subin9:
Closes #3385

## Purpose

The three RAG retrieval paths embedded the query with one call that truncates at the embedding window (`similarity.rs` tokenizes with `max_length.unwrap_or(512)` and right truncation), so only the first 512 tokens reached the store. Two questions behind the same long preamble embedded to the same vector and retrieved the same documents.

A query longer than the window is now split into overlapping windows, each window is embedded and searched, and a document keeps the best score it reached in any window. A query that already fits makes exactly one embedding call and one search, as before.

Three choices worth naming:

* **Best window, not the average.** Averaging the windows into one vector is the usual way to make a single embedding of a long input, and it is not enough here: measured below, it recovers the right document at 3,400 runes and loses it again at 16,680, because the one window carrying the question is diluted by the rest. Taking each window's own score and keeping a document's best is the aggregation that ranks best over passages of a long input in the retrieval literature (MaxP over FirstP, SumP and AvgP).
* **A cap of eight windows per query,** always keeping the first and the last and sampling evenly between them. Each window costs an embedding and a store round trip, and a question sits at either end of a long query about as often.
* **Windows overlap by half a window,** so a sentence one boundary cuts is whole inside its neighbour.

The change is scoped to retrieval. `get_text_embedding` keeps its meaning for every other caller, including the semantic cache that #3455 is giving a window check of its own.

## Test Plan

`make test-rust-ci`, `go test ./pkg/extproc/`, `scripts/gates.sh` for the CI PR lane, and a retrieval measurement against the shipped path with `models/mom-embedding-light` on CPU.

## Test Result

Three Rust tests for the window ranges, registered in `RUST_CI_LIB_TESTS`, and four Go tests for the window sampling and the best-score merge. `make test-rust-ci` passes 10 of 10 and the gates are green on all four lines.

Retrieval measured over a corpus of two documents, one about refunds and one about the audit log, with a question about refunds behind a repeated preamble. The column is the document the query retrieves:

| preamble | query length | windows used | before | after |
|---|---|---|---|---|
| none | 80 runes | 1 of 1 | refund | refund |
| 20 repeats | 3,400 runes | 2 of 2 | audit | refund |
| 100 repeats | 16,680 runes | 8 of 11 | audit | refund |
| 400 repeats | 66,480 runes | 8 of 47 | audit | refund |

The control question about the audit log retrieves the audit document in every row, before and after.

### DIFF
--- a/candle-binding/semantic-router.go
+++ b/candle-binding/semantic-router.go
@@ -164,6 +164,13 @@ typedef struct {
     bool error;
 } TokenizationResult;
 
+// Byte ranges of an input that each fit the embedding window
+typedef struct {
+    int* offsets;
+    int window_count;
+    bool error;
+} TextWindowsResult;
+
 // Classification result structure
 typedef struct {
     int class;
@@ -247,6 +254,8 @@ extern int get_embedding_models_info(EmbeddingModelsInfoResult* result);
 extern void free_embedding_models_info(EmbeddingModelsInfoResult* result);
 extern TokenizationResult tokenize_text(const char* text, int max_length);
 extern int embedding_text_exceeds_window(const char* text, const char* model_type);
+extern TextWindowsResult get_text_windows(const char* text, int max_length);
+extern void free_text_windows(TextWindowsResult result);
 extern void free_cstring(char* s);
 extern void free_embedding(float* data, int length);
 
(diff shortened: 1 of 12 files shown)


Xunzhuo (approved):

wilsonwu (dismissed):
LGTM

Xunzhuo (dismissed):

PR #3651, opened 2026-09-09
[Bug] Point the evaluation registry at the mmbert32k checkpoints the router serves

### LINKED ISSUE #3644: [Bug] Evaluation registry points at the old 8K models instead of the mmbert32k ones the router serves
### Proposed Workgroup

Evaluation & Quality

### Describe the bug

`mom_collection_eval.py` scores a different set of models than the router actually serves.
Every model id in `src/training/model_eval/constants.py` is the older non-32k repo, while
the router config uses the `mmbert32k-` ones. This is true for all five models, not just
one:

| role | router serves (`config/config.yaml:2236-2242`) | eval scores (`constants.py`) |
|---|---|---|
| intent | `mmbert32k-intent-classifier-merged` | `mmbert-intent-classifier-merged` |
| jailbreak | `mmbert32k-jailbreak-detector-merged` | `mmbert-jailbreak-detector-merged` |
| pii | `mmbert32k-pii-detector-merged` | `mmbert-pii-detector-merged` |
| fact-check | `mmbert32k-factcheck-classifier-merged` | `mmbert-fact-check-merged` |
| feedback | `mmbert32k-feedback-detector-merged` | `mmbert-feedback-detector-merged` |

The two sets are genuinely different checkpoints, not aliases of each other. Comparing the
intent repos on the Hub:

| | 8K repo (what eval loads) | 32K repo (what the router serves) |
|---|---|---|
| last modified | 2026-01-12 | 2026-03-06 |
| `max_position_embeddings` | 8192 | 32768 |
| ONNX build | none | `onnx/model.onnx`, `model_fa_fp16.onnx`, `model_sdpa_fp16.onnx` |
| downloads | 29 | 19,858 |

Worth being precise about what actually matters here. The context length difference is not
really the problem, because the eval passes `max_length=512` to the tokenizer in both
places (`mom_collection_eval.py:452` and `:466`), so it truncates everything at 512 tokens
no matter which model it loads. The real issue is simply that these are weights from a
different and older training run. So the accuracy, F1, per-class recall and confusion
matrices the script produces are all correct measurements of a January checkpoint that
nothing in the repo serves and that has been downloaded 29 times.

### To Reproduce

On main at e17d7b48.

What the eval loads:

```bash
grep -n '"id"' src/training/model_eval/constants.py
# 25:  "id": "llm-semantic-router/mmbert-feedback-detector-merged",
# 35:  "id": "llm-semantic-router/mmbert-jailbreak-detector-merged",
# 45:  "id": "llm-semantic-router/mmbert-fact-check-merged",
# 55:  "id": "llm-semantic-router/mmbert-intent-classifier-merged",
# 80:  "id": "llm-semantic-router/mmbert-pii-detector-merged",
```

What the router serves:

```bash
grep -n "prompt_guard: \|domain_classifier: \|pii_classifier: \|fact_check_classifier: \|feedback_detector: " config/config.yaml
# 2236:      prompt_guard: models/mmbert32k-jailbreak-detector-merged
# 2237:      domain_classifier: models/mmbert32k-intent-classifier-merged
# 2238:      pii_classifier: models/mmbert32k-pii-detector-merged
# 2239:      fact_check_classifier: models/mmbert32k-factcheck-classifier-merged
# 2242:      feedback_detector: models/mmbert32k-feedback-detector-merged
```

What the Makefile downloads:

```bash
grep -n -A5 "MMBERT_32K_MERGED_MODELS :=" tools/make/models.mk
# 52:MMBERT_32K_MERGED_MODELS := \
# 53- mmbert32k-feedback-detector-merged \
# 54- mmbert32k-intent-classifier-merged \
# 55- mmbert32k-pii-detector-merged \
# 56- mmbert32k-jailbreak-detector-merged \
# 57- mmbert32k-factcheck-classifier-merged
```

And not one of the five checkpoints the eval loads appears anywhere in the router config:

### Expected behavior

The evaluation should score the checkpoints the router serves, so its numbers describe the
deployed system. `constants.py` should use the same names already written down in
`tools/make/models.mk:52-57` (`MMBERT_32K_MERGED_MODELS`), with the `lora_id` entries coming
from `MMBERT_32K_LORA_ADAPTERS` at `:44-49`.

### Impact

Every metric the script reports is affected: accuracy, F1, precision, recall and the
confusion matrices. They are not wrong in the sense of being miscalculated. They are correct
numbers about the wrong artifact. Anything decided on the back of them, whether intent
accuracy is good enough, whether a change helped, whether a regression is real, is being
decided about weights nobody runs.

The router itself is unaffected. It reads its own paths from `config.yaml` and never touches
this registry.

There is a partial workaround for single-model runs: `--model_id` overrides the checkpoint.
It does not help for multi-model runs and it does not change anything else in the registry.

### Additional context

This looks like ordinary drift rather than a deliberate choice, and the dates line up
cleanly:

| date | what happened |
|---|---|
| 2026-01-12 | the 8K `mmbert-*-merged` models are created on the Hub |
| 2026-01-20 | #1040 merged, adding `MMBERT_MODELS` (the 8K list) to `models.mk` for candle inference |
| 2026-01-21 | #1156 opened, adding the eval script, using those same 8K names |
| 2026-01-31 | the 32K models are created on the Hub, #1227 opened |
| 2026-02-01 | #1227 merged, adding `MMBERT_32K_MERGED_MODELS` |
| 2026-02-04 | #1156 merged, three days later, still on the 8K names |
| 2026-03-17 | #1553 rewrites `config.yaml`, and the router's classifiers become `mmbert32k-*` |

Two things to watch out for when fixing this.

The rename is not a simple prefix swap. `mmbert32k-fact-check-merged` does not exist on the
Hub (the API returns 401). The real name is `mmbert32k-factcheck-classifier-merged`, one
word and a different suffix. A blanket `s/mmbert-/mmbert32k-/` would produce a nonexistent
repo for fact-check, and because `load_model_and_tokenizer` wraps the load in
`retry_operation`, you get three retries and then a fairly opaque failure. Copying the names
from `models.mk` avoids this.

The label order is safe. I compared `id2label` in both intent artifacts and they are
identical, with `other` at index 10 in each. So repointing the registry does not interact
with the label-order fix in #3557. Both artifacts agree with each other and both disagree
with the current `constants.py`.

Finally, it might be worth adding a small contract test that asserts the eval registry
matches `MMBERT_32K_MERGED_MODELS`. The repo currently keeps three separate lists of model
names (`constants.py`, `tools/make/models.mk`, `config/config.yaml`) with nothing tying them
together, which is why this drifted unnoticed. Such a test would be stdlib-only, would fit
the existing `make test-training-contracts` target, and would have caught this in February.

yaojiejia:
Closes #3644

## Purpose

The evaluation was scoring the wrong models.

`src/training/model_eval/constants.py` named the old 8K `mmbert-*` repos, but the router serves the `mmbert32k-*` ones. So every accuracy, F1 and confusion matrix `mom_collection_eval.py` printed was a correct number about a checkpoint nobody runs. Nothing ever failed, because the old repos still exist and still load fine.

This points all five roles at the names already written down in `tools/make/models.mk`, so the eval measures what we actually ship.

The second commit adds a contract test. The root cause here is that we keep model names in three places (`constants.py`, `tools/make/models.mk`, `config/config.yaml`) and nothing checked them against each other, which is how this sat unnoticed for about seven months. The test reads all three and fails if they drift apart again. It is stdlib-only, so it runs under the plain `python3` that `make test-training-contracts` uses. No Makefile change needed, the target already picks up that directory.

## Test Plan

```bash
# lint, format and the contract suites CI runs
make test-training-contracts

# check the new test actually bites: run it against the old registry
git stash && git show origin/main:src/training/model_eval/constants.py > src/training/model_eval/constants.py
python3 -m unittest discover -s src/training/model_eval/tests -p 'test_registry*.py'

# confirm every new id is a real repo
curl -s -o /dev/null -w "%{http_code}\n" https://huggingface.co/api/models/llm-semantic-router/<each id>

# run the eval end to end on two of the repointed checkpoints
python src/training/model_eval/mom_collection_eval.py --model intent --limit 200 --batch_size 16 --device cpu
python src/training/model_eval/mom_collection_eval.py --model jailbreak --limit 200 --batch_size 16 --device cpu
```

## Test Result

`make check BASE_REF=origin/main` exits 0. `make test-training-contracts` is green: 6 suites, 139 tests.

All 10 new ids (5 merged, 5 LoRA) return 200 on the Hub.

The new test does its job. Against the old registry, 4 of its 6 tests fail across all 5 roles. I also tried a half-finished fact-check rename, where the merged id gets updated but the adapter is left behind, and 3 tests catch it.

Both repointed checkpoints load and evaluate cleanly on CPU. Here is what the same run gives on each checkpoint, first 200 samples of the test split, batch 16, CPU:

| role | | accuracy | F1 | p50 latency |
|---|---|---|---|---|
| intent | served 32K (this PR) | 0.9100 | 0.9529 | 101 ms |
| | old 8K (what it scored before) | 0.8400 | 0.9130 | 97 ms |
| | **change** | **+7.0 pts** | **+4.0 pts** | |
| jailbreak | served 32K (this PR) | 0.7500 | 0.7385 | 972 ms |
| | old 8K (what it scored before) | 0.8950 | 0.8949 | 1054 ms |
| | **change** | **-14.5 pts** | **-15.6 pts** | |

---
<details>

</details>

### DIFF
--- a/src/training/model_eval/constants.py
+++ b/src/training/model_eval/constants.py
@@ -22,8 +22,8 @@ LANGUAGE_CODES = {
 # Registry of all MoM models (merged as well as LoRA)
 MODEL_REGISTRY = {
     "feedback": {
-        "id": "llm-semantic-router/mmbert-feedback-detector-merged",
-        "lora_id": "llm-semantic-router/mmbert-feedback-detector-lora",
+        "id": "llm-semantic-router/mmbert32k-feedback-detector-merged",
+        "lora_id": "llm-semantic-router/mmbert32k-feedback-detector-lora",
         "type": "text_classification",
         "hf_dataset": "llm-semantic-router/feedback-detector-dataset",
         "labels": ["SAT", "NEED_CLARIFICATION", "WRONG_ANSWER", "WANT_DIFFERENT"],
@@ -32,8 +32,8 @@ MODEL_REGISTRY = {
         "split": "validation",
     },
     "jailbreak": {
-        "id": "llm-semantic-router/mmbert-jailbreak-detector-merged",
-        "lora_id": "llm-semantic-router/mmbert-jailbreak-detector-lora",
+        "id": "llm-semantic-router/mmbert32k-jailbreak-detector-merged",
+        "lora_id": "llm-semantic-router/mmbert32k-jailbreak-detector-lora",
         "type": "text_classification",
         "hf_dataset": "llm-semantic-router/jailbreak-detection-dataset",
         "labels": ["safe", "unsafe"],
@@ -42,8 +42,8 @@ MODEL_REGISTRY = {
         "split": "test",
     },
     "fact-check": {
-        "id": "llm-semantic-router/mmbert-fact-check-merged",
-        "lora_id": "llm-semantic-router/mmbert-fact-check-lora",
+        "id": "llm-semantic-router/mmbert32k-factcheck-classifier-merged",
+        "lora_id": "llm-semantic-router/mmbert32k-factcheck-classifier-lora",
         "type": "text_classification",
         "hf_dataset": "llm-semantic-router/fact-check-classification-dataset",
         "labels": ["NO_FACT_CHECK_NEEDED", "FACT_CHECK_NEEDED"],
@@ -52,8 +52,8 @@ MODEL_REGISTRY = {
         "split": "test",
     },
     "intent": {
-        "id": "llm-semantic-router/mmbert-intent-classifier-merged",
-        "lora_id": "llm-semantic-router/mmbert-intent-classifier-lora",
+        "id": "llm-semantic-router/mmbert32k-intent-classifier-merged",
+        "lora_id": "llm-semantic-router/mmbert32k-intent-classifier-lora",
         "type": "text_classification",
         "hf_dataset": "TIGER-Lab/MMLU-Pro",
         "labels": [
@@ -77,8 +77,8 @@ MODEL_REGISTRY = {
         "split": "test",
     },
     "pii": {
-        "id": "llm-semantic-router/mmbert-pii-detector-merged",
-        "lora_id": "llm-semantic-router/mmbert-pii-detector-lora",
+        "id": "llm-semantic-router/mmbert32k-pii-detector-merged",
+        "lora_id": "llm-semantic-router/mmbert32k-pii-detector-lora",
         "type": "token_classification",
         "hf_dataset": "presidio",
         "labels": [
--- a/src/training/model_eval/tests/test_registry_matches_served_models.py
+++ b/src/training/model_eval/tests/test_registry_matches_served_models.py
@@ -1,0 +1,160 @@
+"""Contract tests tying the evaluation registry to the checkpoints the router serves.
+
+The repo writes the classifier model names down in three places and, until
+issue #3644, nothing tied them together:
+
+  * `tools/make/models.mk` decides what `make download-models` pulls,
+  * `config/config.yaml` decides what the router loads at runtime,
+  * `src/training/model_eval/constants.py` decides what the evaluation scores.
+
+The third drifted. For roughly seven months the registry named the older 8K
+`mmbert-*` repos while the router served the `mmbert32k-*` ones, so every
+accuracy, F1 and confusion matrix `mom_collection_eval.py` produced was a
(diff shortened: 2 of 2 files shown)


Xunzhuo (approved):
Thanks

PR #3654, opened 2026-09-09
[Bug] Keep JSON scalars valid through tool-output compression

### LINKED ISSUE #3625: [Bug] Tool-output compression corrupts JSON scalars: valid JSON in, invalid JSON out
### Proposed Workgroup

Agentic & Context

### Describe the bug

`CompressToolOutput` turns valid JSON tool output into **invalid JSON** whenever
the output is a JSON *scalar* — a string, number, `true`, `false`, or `null`.

The routing gate is a prefix test:

```go
// pkg/contextcompression/compressor.go:378
func isJSONObjectOrArray(content string) bool {
	trimmed := strings.TrimSpace(content)
	return strings.HasPrefix(trimmed, "{") || strings.HasPrefix(trimmed, "[")
}
```

Only `{` and `[` reach the structure-aware `compressJSON` path. Every other
valid JSON document falls through to `compressText`, which chunks the raw bytes
and rejoins the survivors with `strings.Join(parts, "\n")` plus

```go
// pkg/contextcompression/compressor.go:14
const omissionMarker = "\n[... context omitted by route compression ...]\n"
```

A raw newline is not legal inside a JSON string literal, so the result no
longer parses. The corrupted text is then written straight back into the
outgoing tool block by `applyCandidate`:

```go
// pkg/contextcompression/service.go:411,421
compressed := CompressToolOutput(block.Text, candidate.plan.Query, ...)
...
block.SetText(compressed.Content)
```

### To Reproduce

**1. The repository's own fuzz target finds it in about four minutes.**

```bash
cd src/semantic-router
go test ./pkg/contextcompression/ \
  -run '^FuzzCompressToolOutputPreservesJSON$' \
  -fuzz '^FuzzCompressToolOutputPreservesJSON$' -fuzztime 300s
```

```text
fuzz: minimizing 78-byte failing input file
--- FAIL: FuzzCompressToolOutputPreservesJSON (244.86s)
    compressor_fuzz_test.go:33: compression corrupted JSON: "00"
Failing input written to testdata/fuzz/FuzzCompressToolOutputPreservesJSON/231bf21d8233a41b
```

Minimized corpus entry:

```text
go test fuzz v1
string("1000000000000000000000000000000000")
string("0")
```

The seed corpus in `compressor_fuzz_test.go` contains only objects, arrays and
non-JSON, so the committed seeds all pass and CI stays green.

**2. A realistic tool output, at production-shaped budgets.**

A tool that returns a log dump as a single JSON string — valid JSON, not an
object or array:

```go
raw := strings.Repeat("2026-09-08T10:15:00Z ERROR worker failed to reconcile shard; retrying. ", 140)
body, _ := json.Marshal(raw)               // a valid JSON string, ~3.5k tokens
res := CompressToolOutput(string(body), "why did the worker fail", 2000, 1000)

var v interface{}
err := json.Unmarshal([]byte(res.Content), &v)
```

```text
Applied=true Format="text"
unmarshal error: invalid character '\n' in string literal
first raw newline at byte 994 of 3033
context: ..."...reconcile shard; retrying.\n2026-09-08T10:15:00Z ERROR worker failed..."...
contains omission marker: true
```

`minTokens=2000` / `targetTokens=1000` are ordinary budget values, and a
multi-thousand-token log dump returned as a JSON string is an ordinary tool
result. Nothing about this input is degenerate.

### Expected behavior

The invariant the repository already asserts in
`FuzzCompressToolOutputPreservesJSON` should hold for every JSON document, not
just objects and arrays:

> if the input is valid JSON and compression is applied, the output is valid JSON.

For a JSON scalar that means either compressing the *decoded* value and
re-marshalling it, or declining to compress and returning the content
unchanged. Both keep the contract; the choice is a maintainer call, and I have
no strong preference.

### Impact

A tool result that parses today can reach the model as text that no longer
parses. Agent loops that hand tool output back to a JSON-consuming step are the
exposed path: the compression is applied silently, `Applied` is `true`, and
nothing downstream re-validates. JSON objects and arrays — the common case —
are unaffected, which is why this has not shown up in practice.

### Additional context

Found by running the repository's existing fuzz target rather than by reading
the code; the minimized corpus entry above is a ready-made regression seed, and
`go test ./pkg/contextcompression/` fails on it with no `-fuzz` flag once the
file is committed.

Happy to send the fix if this is accepted — please say which of the two
behaviors above you want for scalars.

rk9595:
Closes #3625

## Purpose

`CompressToolOutput` decides how to compress by calling `isJSONObjectOrArray`, which is a prefix test for `{` or `[`. Every other valid JSON document — string, number, `true`, `false`, `null` — misses that test and falls through to `compressText`, which rejoins the surviving chunks with raw newlines and a newline-wrapped omission marker. A raw newline is illegal inside a JSON string literal, and `service.go` writes the result straight back with `block.SetText`, so a tool result that was valid JSON on the way in reaches the model as text that no longer parses.

Realistic trigger: a log dump returned as a JSON string (a single `"..."` document of ~3.5k tokens) with `min_tokens: 2000` / `target_tokens: 1000`. Compression reports `Applied=true` and the output fails to unmarshal with `invalid character '\n' in string literal`.

The fix decodes a JSON string document, compresses its inner text on the existing path, and re-encodes the survivor, so the document stays valid while still shrinking. Other scalars carry no chunk structure worth dropping and are returned unchanged. The re-encoded string keeps `Format: "json"` so the `JSONBlocks` metric in `service.go` stays accurate.

Affected module: `pkg/contextcompression` (owned by `wg/agentic-context`, the label on #3625).

The bug was found by running this repository's own fuzz target, `FuzzCompressToolOutputPreservesJSON` — it is a violation of an invariant the repo already asserts. The minimised input is committed under `testdata/fuzz/`, so `go test` catches the regression with no `-fuzz` flag.

## Test Plan

```
cd src/semantic-router
go test ./pkg/contextcompression/... -count=1
go test ./pkg/contextcompression/ -run '^$' -fuzz FuzzCompressToolOutputPreservesJSON -fuzztime 120s
gofmt -l pkg/contextcompression/ && go vet ./pkg/contextcompression/... && go build ./...
```

Red/green check: restore `compressor.go` from `main` and re-run the two JSON tests.

## Test Result

- `go test ./pkg/contextcompression/... -count=1` — ok, 3.2s. That includes the new `TestCompressToolOutputPreservesValidJSONScalars` and the committed fuzz seed `231bf21d8233a41b`. (`TestRedisRecoveryStoreRoundTripAndScopeIsolation` skips: no local Redis.)
- Fuzz, 120s / 1,068,248 execs — PASS, no new corpus entries.
- `gofmt` clean, `go vet` clean, `go build ./...` succeeds.
- With `compressor.go` reverted to `main`, both fail as expected:
  - `TestCompressToolOutputPreservesValidJSONScalars`: `compressed JSON string is invalid JSON: invalid character '\n' in string literal`
  - `FuzzCompressToolOutputPreservesJSON/231bf21d8233a41b`: `compression corrupted JSON: "00"`

Branch is rebased on `ce271702`. No remaining blockers; behaviour for JSON objects, arrays and plain text is unchanged.


FAUST-BENCHOU (commented):
LGTM. Decoding and re-encoding JSON string scalars cleanly preserves valid JSON through compression, while leaving the existing object, array, and plain-text paths unchanged.

Xunzhuo (approved):
JSON strings are re-encoded after compression, while numbers, booleans, and null retain valid scalar representations. The scalar regressions and exact-head core checks cover the change.

wilsonwu (approved):
LGTM

rk9595:
Updated the branch with `main` locally (`82bc3bef`) — Mergify couldn't do it itself. The 09:17 queue run left after 1h03m with no CI:

> refusing to allow a GitHub App to create or update workflow `.github/workflows/dashboard-test.yml` without `workflows` permission

The delta from `main` touched 13 files under `.github/workflows/`, which Mergify can't push to a fork. Now 0 behind, so no update is needed at queue time.

Clean merge, no conflicts; `pkg/contextcompression/` had no upstream changes. Package tests, `gofmt`, `go vet` green post-merge, including the seeded fuzz corpus entry.

@wilsonwu could you re-tick the merge queue box when you get a chance?

Issue #3608, opened 2026-09-08
[Feature] Add Snowflake Cortex AI as a built-in serving provider


Xunzhuo:
## Goal

Add **Snowflake Cortex AI** as a built-in serving provider with an explicit, source-backed endpoint contract for operator-bound Model Cards.

## Current-main baseline

The 60-provider registry on `main` has no Snowflake Cortex provider. Snowflake documents its current [Cortex REST API and supported inference surfaces](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-rest-api).

## Scope

1. Freeze the official REST surface and authentication method being implemented, including account-specific hostname and operator-supplied account/database context.
2. Add one `ProviderDefinition` with exact supported protocol operations, paths, auth/header projection, request constraints, and reasoning/tool transport where documented.
3. Add deterministic fixtures plus a sanitized live conformance receipt. Tenant identifiers, credentials, private endpoints, and secret-bearing logs must remain outside repository artifacts.
4. Generate the Router, CLI, Dashboard, schema, logo/presentation, and documentation projections from the canonical resource and add focused validation coverage.

This issue adds the provider contract only. It does not mirror Snowflake's model catalog, add Snowflake/Arctic as a creator, infer model equivalence from display names, or publish pricing without dated first-party evidence. Provider model mappings belong in a separate bounded issue.

## Completion signal

- [ ] `snowflake-cortex` is a unique built-in provider identity with documented operator configuration.
- [ ] Supported operations and projections have deterministic fixtures.
- [ ] A sanitized live receipt records the exact contract and observation date.
- [ ] Generated projections remain in parity and provider/catalog checks pass.

## Contributing

Comment `/assign` after acceptance, then post the exact official API surface and credential-free test plan before opening a PR.

## Acceptance recommendation

As a project Maintainer, I recommend accepting this bounded provider-integration slice under Data Plane & Networking.

Related: #3489, #3570.

Bevisy:
## Official API surface (frozen for this provider)

Source: https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-rest-api

- **Base URL** is account-scoped: `https://<account-identifier>.snowflakecomputing.com/api/v2/cortex/v1`. The account identifier is operator-supplied; the provider deliberately sets no `default_base_url` (same pattern as `azure-openai`), and the operator configures it on the provider instance.
[...]

PR #3656, opened 2026-09-09
[Feature] Add Snowflake Cortex AI as a built-in serving provider

Bevisy:
## What

Add `snowflake-cortex` as a built-in serving provider through the catalog contributor flow: one authored provider resource, then the generated Router embed, website snapshot, and built-in recipe manifest.

## Provider contract

The frozen official surface and credential-free test plan are recorded at https://github.com/vllm-project/semantic-router/issues/3608#issuecomment-5595548861. Highlights:

- **Account-scoped base URL**: operators configure `https://<account-identifier>.snowflakecomputing.com/api/v2/cortex/v1` on the provider instance. The base URL must include `/api/v2/cortex/v1` because the resolver strips a protocol's `/v1` prefix whenever a base path is configured.
- **Auth**: `Authorization: Bearer <Snowflake programmatic access token>`.
- **Two documented protocols under one identity**: `openai/chat-completions@1` and `anthropic/messages@1`, each with `#create`. `list_models` is intentionally not declared — Cortex does not document a `/v1/models` endpoint.
- **`anthropic-version: 2023-06-01`** as a provider-global default header (required by the Messages surface; the chat-completions path ignores it).
- Model names stay provider-native and are **not** mapped to catalog cards here (out of scope per #3608).

## Changes

- `config/catalog/resources/providers/snowflake-cortex.yaml` — new provider definition, `unverified` (wire-level conformance pending; `live_verified` reserved for the separate live receipt).
- `src/semantic-router/pkg/catalog/zz_generated_catalog.go`, `website/static/model-catalog/catalog.json`, `config/recipes/built-in/latest/catalog.yaml` — regenerated projections.
- `src/semantic-router/pkg/catalog/compiler_test.go` — `TestSnowflakeCortexProviderContractFixture`, a deterministic offline fixture pinning the compiled contract (identity, protocols, operations, auth/header projection, conformance status).
- `src/semantic-router/pkg/catalog/operations_test.go` — `TestSnowflakeCortexResolveOperationPath` covers both declared `create` operations resolving under the account-scoped base.

## Validation

- `make test-semantic-router` — pass; `go test ./pkg/catalog/...` (incl. both snowflake fixtures) — pass.
- `generate_model_catalog.py --check` — generated projections in parity.
- `make harness-check` — pass.
- One catalog Python test (`test_model_layout_stays_inside_catalog_and_uses_one_file_per_creator`) fails on macOS only, from a `/var` vs `/private/var` temp-path mismatch in pathlib; reproduced on clean `main` (unrelated to this change, unaffected on Linux CI).

## Deferred (separate follow-up)

The live conformance receipt (`live_verified`) requires an operator-owned Snowflake account; no credentials, tenant identifiers, or private endpoints are added to repository artifacts.

Related: #3608


config/catalog/resources/providers/snowflake-cortex.yaml
```diff
@@ -0,0 +1,33 @@
+id: snowflake-cortex
+display_name: Snowflake Cortex AI
+description: Frontier models hosted and governed inside your Snowflake perimeter.
+category: model_api
+support_tier: native
+# No default_base_url: the host is account-specific. Operators configure
+# ProviderInstance.base_url as https://<account>.snowflakecomputing.com/api/v2/cortex
```
Xunzhuo: **Keep `/v1` in the configured Cortex base URL**

Impact: This instruction tells operators to end `base_url` at `/api/v2/cortex`, but `joinProtocolOperationPath` removes the protocol `/v1` prefix whenever a non-empty base path is supplied. Requests therefore resolve to `/api/v2/cortex/chat/completions` or `/messages`, while Snowflake documents both endpoints under `/api/v2/cortex/v1`.

Required change: Instruct operators to configure the SDK base URL through `/api/v2/cortex/v1`, or change the resolver contract consistently, and add focused `ResolveOperationPath` coverage for both declared protocols.

config/catalog/resources/providers/snowflake-cortex.yaml
```diff
@@ -0,0 +1,33 @@
...
+auth:
+  strategy: bearer
+  header: Authorization
+  prefix: Bearer
+presentation:
+  logo: monogram
+  monogram: SF
+  monochrome: true
+conformance:
+  status: fixture_verified
```
Xunzhuo: **Exercise the wire contract before marking it verified**

Impact: The only added fixture asserts compiled provider metadata; it never resolves a URL or exercises auth, headers, request, or response behavior for either declared operation. Publishing `fixture_verified` therefore claims protocol conformance that the patch does not test, and the wrong path above passes unnoticed.

Required change: Add deterministic wire-level conformance fixtures for both create operations, or keep this provider `unverified` until that evidence exists. Reserve `live_verified` for the separate sanitized live receipt.

Xunzhuo (commented):
Thanks for adding Snowflake through the catalog flow. The authored provider currently resolves the wrong production URLs and overstates the conformance evidence.

config/catalog/resources/providers/snowflake-cortex.yaml
Bevisy: Corrected. `ProviderInstance.base_url` must include `/api/v2/cortex/v1` — the YAML operator comment now instructs exactly that, and `TestSnowflakeCortexResolveOperationPath` covers both declared `create` operations resolving under that base to `/api/v2/cortex/v1/chat/completions` and `/api/v2/cortex/v1/messages`.

config/catalog/resources/providers/snowflake-cortex.yaml
Bevisy: The snowflake fixture now resolves both create operation paths against the account-scoped operator base (via `ResolveOperationPath`) in addition to pinning the provider identity, both operations, bearer auth contract, and the default header projection. The request/response wire format for both protocols is the stock OpenAI Chat Completions / Anthropic Messages surface already conformance-tested in the protocol codecs; this provider introduces no custom wire. `live_verified` stays reserved for the separate sanitized live receipt (deferred).

Bevisy: Agreed. The fixtures pin the compiled contract and both resolved operation paths, but they do not run a wire exchange, so `snowflake-cortex` is now marked `unverified` pending wire-level conformance. `live_verified` stays reserved for the deferred live receipt.

FAUST-BENCHOU (commented):
LGTM. The updated provider contract now resolves both Cortex APIs under the required `/api/v2/cortex/v1` base and keeps conformance unverified until wire-level evidence is available.

Bevisy:
Synced with the latest `main`. The generated catalog embed conflicted again after the upstream catalog sources changed, so I resolved it the same way the generator owns this contract: regenerated the Router embed, website snapshot, and recipe manifest from the merged canonical YAML sources and confirmed all three projections are in parity.

State after the sync:
- The `snowflake-cortex` provider contract is unchanged (identity, both `#create` operations, bearer auth, `anthropic-version` header, account-scoped `base_url` including `/api/v2/cortex/v1`).
- Both deterministic fixtures still pass: the compiled-contract pin and the `ResolveOperationPath` coverage for both protocols.
- The PR diff remains limited to the snowflake provider files; no unrelated upstream content is included.
- `generate_model_catalog.py --check` and the catalog Go tests are green locally.

Xunzhuo (dismissed):
The account-specific URL contract now resolves both Cortex endpoints correctly, and conformance remains explicitly unverified. The Snowflake catalog fixtures, generated projections, and selected repository checks pass in the current-head CI. No live Snowflake model request was run.

wilsonwu (dismissed):
LGTM

Bevisy:
Merge conflict handled, PTAL @Xunzhuo @wilsonwu

Xunzhuo (dismissed):
The account-specific URL contract resolves both Cortex endpoints correctly, and provider conformance remains explicitly unverified. Both Snowflake fixtures, all 68 catalog tests, and generated-reference checks pass in current-head CI. No live Snowflake model request was run.

config/catalog/resources/providers/snowflake-cortex.yaml
```diff
@@ -0,0 +1,34 @@
+id: snowflake-cortex
```
Xunzhuo, line 1: [P2] Update the Dashboard catalog assertion with this provider

Adding this resource raises the generated catalog to 62 providers, but `TestGeneratedPublicModelCatalogSatisfiesDashboardContract` still requires 61 in `dashboard/backend/handlers/model_catalog_test.go`. The existing test fails on this head with `models=101 providers=62 evaluations=1510`, so the Dashboard suite will fail when it consumes this catalog. Update the expected provider count alongside this addition and rerun that contract test.

Xunzhuo (changes requested):
The Snowflake provider and generated catalog checks pass, but the existing Dashboard catalog contract test fails because the provider count was not updated with this addition. Please update that consumer assertion and rerun the focused Dashboard test. No live Snowflake request was run.

config/catalog/resources/providers/snowflake-cortex.yaml
Bevisy: Good catch — bumped the expected provider count to 62.

Xunzhuo (dismissed):
The Dashboard catalog assertion now matches the added provider, and the current-head Dashboard contract test, both Snowflake fixtures, and generated-contract checks pass. The account-specific endpoint paths remain correct; no live Snowflake request was run.

Bevisy:
Merge conflict is recoverd. PTAL @Xunzhuo

Xunzhuo (dismissed):
The account-scoped paths, declared operations and generated catalog now agree. Both Snowflake contract tests and the Dashboard catalog check pass; live conformance remains explicitly unverified.

Bevisy:
`zz_generated_catalog.go` conflicts on every merge of `main` into this branch — that's expected, it's a generated file. Resolved once again here (regenerated from the merged YAML sources). PTAL @Xunzhuo — hoping this is the last one for this PR, hahaha.

FAUST-BENCHOU:
The provider contract looks consistent now: the configured Cortex base retains `/api/v2/cortex/v1`, both declared create operations have path-resolution coverage, and keeping the provider `unverified` until live wire evidence is available is the right boundary. LGTM.

Xunzhuo (approved):
The Snowflake endpoint fixtures, generated contracts and Dashboard catalog assertion pass on this head. The account-scoped operation paths are consistent, and live conformance remains explicitly unverified. No live Snowflake request was run.

PR #3657, opened 2026-09-09
[Bug] Keep classification requests safe across a dynamic reload

### LINKED ISSUE #3645: [Bug] Reload retirement waits only on ext_proc refs, not on classification API borrowers
### Proposed Workgroup

### Describe the bug

Generation retirement is only half-covered. `RouterService.Process` borrows the
current generation under a ref:

```go
// pkg/extproc/server.go:313
generation.refs.Add(1)
defer generation.refs.Done()
```

Two other consumers take the same pointers without a ref:

- `RouterService.GetRouter()` (`pkg/extproc/server.go:305`) returns the router directly.
- `Registry.ClassificationService()` (`pkg/routerruntime/registry.go:153`) returns
  the service pointer and releases its `RLock` before the caller uses it. This is
  the path the classification HTTP API resolves per call via
  `liveClassificationService.current()` (`pkg/apiserver/runtime_service.go:74`).

`closeRetiredGeneration` waits only on `refs` before closing:

```go
// pkg/extproc/server.go:358
generation.refs.Wait()
return generation.router.Close()
```

So a reload can run `RecipeClassifiers.Close()` -> `Classifier.Close()` while a
classification API request is still using that classifier.

### To Reproduce

A code path rather than an observable failure today (see Impact):

1. Start the router with a classifier that owns a closable resource, e.g. a
   remote `backend:` block on `categories` or `complexity`.
2. Issue a classification API request.
3. Trigger a dynamic reload while it is in flight.
4. `refs.Wait()` returns immediately - the API request never registered - and
   `Classifier.Close()` runs against the classifier that request is using.

### Expected behavior

Every consumer that borrows a generation registers a ref for the duration of its
use, so retirement waits for API requests the same way it waits for ext_proc
streams.

### Impact

Latent today. The only closable classifier resources are connector-backed, and
`connector.Client.Close()` is `http.Client.CloseIdleConnections()`
(`pkg/modelruntime/connector/connector.go:251`), which by definition does not
interrupt in-flight requests - the worst outcome is a dropped keep-alive socket.

It becomes a real use-after-close as soon as any classifier gains a resource
whose `Close` tears down state actually in use: a cancelled context, a closed
gRPC connection, or a freed FFI handle.

### Additional context

Pre-dates #3542 and is not specific to one signal - `categories` reaches the
same code through `HTTPClassifierInference`.

Raised during #3542's review as a fourth blocker, stated as "dynamic reload
replaces RecipeClassifiers without lifetime-safe retirement, leaking the old
remote connectors while only the startup set is registered for shutdown". That
premise does not hold: reload uses the startup builder, so the new set registers
its own shutdown, and the retired generation closes the old one.

```
buildReloadRouter = buildOpenAIRouterFromConfig            server.go:34
  -> buildRouterComponents -> buildEarlyResources
       -> resources.add(recipeClassifiers.Close)           router_build.go:265
Swap -> closeRetiredGeneration -> router.Close()           server.go:280,358
  -> RecipeClassifiers.Close -> Classifier.Close           classifier_lifecycle.go:151
```

The missing-ref gap above is the part of that report that does hold, filed here
so it can be fixed in the shared reload machinery rather than inside a signal
feature PR.

1fanwang:
Closes #3645

## Purpose

A dynamic reload can close a classifier while a classification API request is still using it.

Before this change, a request in flight through `/api/v1/classify` when a reload lands could have its classifier closed underneath it. The ext_proc path registers a reference on the live generation, so retirement waits for it. The classification HTTP API did not: it resolves the service from the registry, which releases its read lock before the caller uses the pointer, so the retirement wait returned immediately.

After this change the reload waits for that request, as it does for an ext_proc stream.

The consequence is latent today: the only closable classifier resource is a connector whose close drops idle keep-alive sockets and cannot interrupt an in-flight request. It becomes a use-after-close as soon as a classifier owns something whose close tears down state in use, such as a cancelled context, a closed gRPC connection, or a freed FFI handle.

## Solution

The generation exposes a borrow that takes the same lock retirement uses, so a caller either registers first and makes retirement wait, or sees the retirement and falls back. The registry carries that borrow alongside the service, and the API path holds it across each call.

## Test Plan

```bash
make rust
export CGO_LDFLAGS="-L$PWD/candle-binding/target/release -L$PWD/nlp-binding/target/release -L$PWD/ml-binding/target/release"
cd src/semantic-router
go test ./pkg/apiserver/... ./pkg/routerruntime/... ./pkg/extproc/... -count=1
go test ./pkg/routerruntime/ -run Borrow -race -count=1

# before: drop the reference registration, keeping the tests
python3 - <<'PY'
p = "pkg/routerruntime/registry.go"
s = open(p).read()
s = s.replace("""\trelease, ok := borrow()
\tif !ok {
\t\treturn nil, nil, false
\t}
\treturn service, release, true""", "\treturn service, func() {}, true")
open(p, "w").write(s)
PY
go test ./pkg/routerruntime/ -run Borrow -race -count=1
```

## Test Result

<details>
<summary>Raw logs</summary>

```text
# before
--- FAIL: TestRetirementWaitsForClassificationBorrow (0.00s)
    generation closed while a classification borrow was still held
--- FAIL: TestBorrowDeclinedAfterRetirement (0.00s)
    borrow succeeded against a retired generation
FAIL

# after
--- PASS: TestRetirementWaitsForClassificationBorrow (0.10s)
--- PASS: TestBorrowDeclinedAfterRetirement (0.00s)
ok .../apiserver 5.9s   ok .../routerruntime 3.1s   ok .../extproc 5.3s
```

</details>


AayushSaini101:
Thanks @1fanwang for raising some test cases are failing can you PTAL thanks

src/semantic-router/pkg/apiserver/runtime_service.go
```diff
@@ -59,18 +59,33 @@ type classificationService interface {
...
 		fallback: fallback,
 		resolver: resolver,
+		borrower: borrower,
 	}
 }
 
+// borrow resolves the live service and holds a reference on the generation that
+// owns it until release runs. Reload retirement waits on that reference, so the
+// classifier cannot be closed underneath an in-flight API call.
+func (s *liveClassificationService) borrow() (classificationService, func()) {
```
AayushSaini101: @1fanwang, can we use another name here instead of using borrow like acquireClassifier, syncChanges etc like this thanks

1fanwang: Done in https://github.com/vllm-project/semantic-router/pull/3657/commits/7a259914cdbae8c7bc03c23a4aab2663b80ab6c3. Renamed the whole borrow API to acquire, matching the acquire and release pair the registry returns.

AayushSaini101 (approved):

wilsonwu (approved):
LGTM, thanks!

PR #3678, opened 2026-09-09
[Bug] Divide evaluation batch latency by the row count instead of the column count

### LINKED ISSUE #3642: [Bug] Evaluation latency is divided by the number of columns instead of the number of rows
### Proposed Workgroup

Evaluation & Quality

### Describe the bug

The eval script times each batch and then divides by the wrong number, so every latency it
reports is wrong.

At `src/training/model_eval/mom_collection_eval.py:443` it takes a batch:

```python
batch = dataset[i : i + args.batch_size]
```

Here is the catch. `load_eval_data` always returns a HuggingFace `Dataset`, and slicing one
of those does not give you a list of rows. It gives you a dict whose keys are column names.
So `len(batch)` tells you how many columns the dataset has, not how many rows are in the
batch.

Lines 503-504 use that number to work out the per-sample time:

```python
batch_time = (time.time() - start) * 1000
lats.extend([batch_time / len(batch)] * len(batch))
```

So the latency it prints is the batch time divided by the column count, which has nothing
to do with how many rows it just ran. How far off it is comes down to rows divided by
columns. With the default `--batch_size 16`:

| model | columns | how wrong |
|---|---|---|
| intent | 9 (`question_id, text, options, answer, answer_index, cot_content, category, src, label`) | 1.78x too high |
| jailbreak | 3 (`text, label, label_text`) | 5.33x too high |
| pii | 2 (`tokens, labels`) | 8x too high |

`feedback` and `fact-check` go through the same line. I could not check their column counts
because I did not have those datasets downloaded.

There is a second problem with the same line. `len(batch)` also decides how many entries
get appended, so `lats` ends up with one entry per column per batch instead of one per row.
Running the jailbreak test set (827 rows) gives you 156 latency samples instead of 827. It
also means `avg_ms` averages across batches rather than across samples, so a half-empty
last batch counts just as much as a full one.

### To Reproduce

On main at e17d7b48.

Quickest way to see it, nothing to download. This builds the same two-column shape the pii
path builds:

```bash
python -c "
from datasets import Dataset
ds = Dataset.from_list([{'tokens': ['a'], 'labels': ['O']} for _ in range(40)])
b = ds[0:16]
print(type(b).__name__, list(b), 'len(batch) =', len(b), 'rows =', len(next(iter(b.values()))))
"
# dict ['tokens', 'labels'] len(batch) = 2 rows = 16
```

Same thing with a real dataset from the registry:

```bash
python -c "
import sys; sys.path.insert(0, 'src/training/model_eval')
from constants import MODEL_REGISTRY
from datasets import load_dataset
cfg = MODEL_REGISTRY['jailbreak']
b = load_dataset(cfg['hf_dataset'], split=cfg['split'])[0:16]
print(list(b), 'len(batch) =', len(b), 'rows =', len(next(iter(b.values()))))
"
# ['text', 'label', 'label_text'] len(batch) = 3 rows = 16
```

The easiest way to spot it in a real run is that the reported number climbs as you raise
`--batch_size`. A per-sample latency should stay about the same. Below are real measured
batch times from the MMLU-Pro rows (9 columns), run through the current line and through a
row count:

| `--batch_size` | what it reports | what it should be | how wrong |
|---|---|---|---|
| 2 | 1.12 ms | 5.04 ms | 4.50x too low |
| 4 | 2.66 ms | 6.01 ms | 2.26x too low |
| 8 | 5.05 ms | 5.68 ms | 1.12x too low |
| 16 | 13.34 ms | 7.80 ms | 1.71x too high |
| 32 | 22.63 ms | 7.03 ms | 3.22x too high |

It flips from too low to too high once the batch gets bigger than 9, which is the column
count. So "the number is too high" is only true at the default batch size, and how wrong it
is changes depending on the batch size you pick. You cannot just correct for it after the
fact.

One caveat on that table. The millisecond values came from a small CPU model, not the real
mmBERT ones, because the mmBERT weights were not cached locally. That only affects the ms
columns. The gap between the two is columns divided by rows, which stays the same no matter
which model you run.

### Expected behavior

It should divide by how many rows are actually in the batch. Then `avg_ms` works out to
total time divided by rows evaluated, and `lats` gets one entry per row.

I checked this against the measured times: dividing by the row count gives exactly total
time over rows. The current line is off by 67% on intent, 400% on jailbreak and 650% on pii
at the default batch size.

### Impact

Only the latency numbers in `mom_collection_eval.py` are affected: `avg_ms`, `p50_ms` and
`p99_ms`. Those get written into the per-model results JSON (lines 526-528 and 603) and
into the `Latency (p50)` column of the summary table it prints at the end (line 679).
Accuracy, F1, precision, recall and the confusion matrices are all fine.

The router itself is not affected. This is just the eval script's reporting.

If you need a usable number out of a run you already have, divide it by
`batch_size / columns` for that dataset. Or run with `--batch_size` set to the column count.

### Additional context

The fix is to get the row count from the batch itself. Every column's list is as long as
the batch, so any one of them will do:

```python
# A dataset slice is a dict of column -> list, so len(batch) is
# the column count. Any column's length is the row count.
rows = len(next(iter(batch.values())))
batch_time = (time.time() - start) * 1000
lats.extend([batch_time / rows] * rows)
```

This works for both kinds of dataset here (the `text`/`label` ones and the
`tokens`/`labels` one) and handles a short final batch on its own.

yaojiejia:
Closes #3642

## Purpose

The eval script timed each batch and divided by `len(batch)`. Slicing a HuggingFace `Dataset`
gives you a dict of column -> values, not a list of rows, so that was the column count. Every
latency it reported was off by rows/columns: 5.33x too high for jailbreak at the default batch
size, 8x for pii, and too low once the batch got bigger than the column count. `lats` also
ended up with one entry per column per batch instead of one per row.

## Test Plan

```bash
make test-training-contracts

# real run, jailbreak detector on the cached test set, two batch sizes
for BS in 16 4; do
  HF_HUB_OFFLINE=1 MPLBACKEND=Agg TOKENIZERS_PARALLELISM=false \
  python src/training/model_eval/mom_collection_eval.py \
    --model jailbreak --limit 48 --batch_size $BS --device cpu --output_dir /tmp/eval-$BS
done
```

Also reverted the source and re-ran the new tests, to check they actually catch it.

## Test Result

`make check` exits 0. `make test-training-contracts` passes, model_eval goes from 13 to 21 tests.

48 rows on CPU:

| run | latency it reports | wall clock / 48 rows | |
|---|---|---|---|
| before, bs 16 | 1103.40 ms/row | 206.9 ms/row | **5.33x too high** |
| after, bs 16 | 198.43 ms/row | 199.0 ms/row | matches |
| after, bs 4 | 163.61 ms/row | 163.3 ms/row | matches |

5.33x is exactly 16 rows / 3 columns, as the issue predicted. `avg_ms` is now total time over
rows at both batch sizes, and accuracy, F1, precision, recall and the confusion matrix are
unchanged. All 8 new tests fail on the old source and pass on the new one.

Worth flagging: only jailbreak was run for real, since that's what's in the local cache. The
other four paths share the same line and their shapes are in the test fixtures, but I didn't
run them. Latency numbers already written into old results JSONs stay wrong.

---
<details>

</details>

### DIFF
--- a/src/training/model_eval/mom_collection_eval.py
+++ b/src/training/model_eval/mom_collection_eval.py
@@ -31,7 +31,6 @@ import re
 import time
 from concurrent.futures import ProcessPoolExecutor, as_completed
 from pathlib import Path
-from typing import Dict, List, Tuple
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -39,7 +38,6 @@ import requests
 import seaborn as sns
 import torch
 from datasets import Dataset, load_dataset
-from transformers import AutoConfig
 from peft import PeftModel
 from seqeval.metrics import accuracy_score as seqe_accuracy_score
 from seqeval.metrics import classification_report as seq_classification_report
@@ -52,8 +50,6 @@ from sklearn.metrics import (
 )
 from tqdm import tqdm
 from transformers import (
-    AutoModelForSequenceClassification,
-    AutoModelForTokenClassification,
     AutoTokenizer,
     ModernBertConfig,
     ModernBertForSequenceClassification,
@@ -65,7 +61,6 @@ try:
 except ImportError:
     from constants import BASE_MODEL_ID, LANGUAGE_CODES, MODEL_REGISTRY
 import warnings
-from sklearn.metrics._classification import _check_targets
 
 # suppress prf warnings
 warnings.filterwarnings(
@@ -212,7 +207,7 @@ def load_eval_data(model_name: str, args) -> Dataset:
             for split in ["test", "validation", "eval", "train"]:
                 try:
                     ds = retry_operation(
-                        lambda: load_custom_ds(split),
+                        lambda split=split: load_custom_ds(split),
                         max_retries=args.max_retries,
                     )
                     logger.info(
@@ -281,7 +276,7 @@ def load_eval_data(model_name: str, args) -> Dataset:
             ds = ds.filter(lambda x: x["category"] in config["labels"])
             if args.limit:
                 ds = ds.select(range(min(len(ds), args.limit)))
-            l2id = {l: i for i, l in enumerate(config["labels"])}
+            l2id = {label: i for i, label in enumerate(config["labels"])}
             ds = ds.map(lambda x: {"label": l2id[x["category"]]})
             ds = ds.rename_column("question", "text")
             return ds
@@ -340,7 +335,7 @@ def load_eval_data(model_name: str, args) -> Dataset:
                 ds = ds.remove_columns(["label"])
             ds = ds.rename_column(config["label_col"], "label")
 
-        label2id = {l: i for i, l in enumerate(config["labels"])}
+        label2id = {label: i for i, label in enumerate(config["labels"])}
 
         def map_to_int(example):
             label = example["label"]
@@ -391,8 +386,8 @@ def load_model_and_tokenizer(model_name: str, args):
 
         # Build config from base model
         clean_labels = config_reg["labels"]
-        id2label = {i: str(l) for i, l in enumerate(clean_labels)}
-        label2id = {str(l): i for i, l in enumerate(clean_labels)}
+        id2label = {i: str(label) for i, label in enumerate(clean_labels)}
+        label2id = {str(label): i for i, label in enumerate(clean_labels)}
 
         hf_config = ModernBertConfig.from_pretrained(
             BASE_MODEL_ID,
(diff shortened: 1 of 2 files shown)


FAUST-BENCHOU (commented):
LGTM. Using the actual row count fixes the latency divisor for HuggingFace Dataset slices, and the regression coverage makes the failure mode clear.

nanasis:
Great catch and testing results!

Xunzhuo (approved):

PR #3685, opened 2026-09-10
[Bug] Prefer better-covered intelligence evidence when scores tie and filter eligible candidates before scoring

### LINKED ISSUE #3567: [Bug] Prefer better-covered intelligence evidence when routing quality scores tie
### Proposed Workgroup

MoM & Routing #2965 

### Describe the bug

Catalog intelligence results contain both `score` and `coverage`, but runtime selectors call:

```go
EvidenceScore(index) (float64, bool)
```

This discards coverage. As a result, routing treats these candidates as equivalent:

```text
Model A: score=75, coverage=60%
Model B: score=75, coverage=100%
```

When other signals are equal, `multi_factor`, AutoMix, and GMT Router cannot prefer the result supported by broader benchmark evidence.

### Expected behavior

Keep score and coverage separate:

- compare score first;
- use higher coverage as the deterministic tie-breaker;
- never multiply score by coverage or treat missing evidence as zero;
- include both values in routing diagnostics.

### Improvement plan

1. Expose the relevant `IndexResult` metadata to selectors.
2. Apply the score-then-coverage comparison consistently to `multi_factor`, AutoMix, and GMT Router.
3. Add tests for equal scores with different coverage, missing evidence, and reasoning-effort-specific results.
4. Align routing diagnostics and documentation with the typed evidence contract.

### Non-goals

- Changing index benchmarks, weights, or the 60% threshold.
- Folding coverage into the published intelligence score.
- Adding configurable uncertainty-adjusted scoring.

### Completion signal

- Equal scores prefer higher coverage when other routing signals are equal.
- Different scores and missing-evidence behavior remain unchanged.
- Selection traces report the score and coverage used.

Parent Epic: #3202  
Related foundation: #3489

nanasis:
Closes #3567

## Summary

- Keep intelligence score and coverage separate; use coverage only for equal-score ties on comparable, exact-effort evidence.
- Preserve complete candidate identity through selectors, Hybrid composition, Router Learning, session persistence, and provider encoding.
- Filter capability-ineligible candidates before scoring while preserving hard eligibility constraints. Final dispatch fails closed instead of rerouting. Declare text-only capabilities explicitly in the response API fixture.

## Decision-path fixes

- Preserve adaptation proposals when protection is observe-only, including error paths.
- Commit session ownership only after successful provider encoding.
- Honor cancellation across selector shortcuts and successful returns.
- Use protection-scoped session state for sampling decisions.
- Respect lower-is-better scores during rescue.
- Preserve ambiguous-candidate rejection in Eval.

Added **41 regression cases**, including exact effort/LoRA preservation, late rejection, and an older failure after a newer dispatch. No new dependencies or public YAML/schema changes. Owned by `wg/mom-routing`.

## Validation

Rebased onto upstream `main` at `e9217462e`; head `f06b1db0f`.

- Post-rebase Semantic Router Go suite: **63 packages passed**, including the protection corpus.
- Affected race tests, changed-code lint, Router Learning PR benchmark, E2E Go module tests, docs translation coverage, and whitespace checks: **passed**.
- Historical live `response-api` E2E: **32/32 passed**, including HTTP 400/no backend dispatch for the image-file fixture. Not rerun after subsequent changes and rebases.

External-store integrations and the full live E2E matrix were not rerun. Refreshed PR CI and reviewer confirmation remain pending.


FAUST-BENCHOU (commented):
One correctness issue here: scoring is now effort-specific, but `allScores` is still keyed only by `model.Model`. Two `ModelRef`s for the same model with different `reasoning_effort` values can therefore overwrite each other, and both candidates are later ranked using whichever effort was written last.

Could we keep the score per candidate (for example by model + effort or candidate index)? The personalized GMT path appears to have the same issue with `scores[model.Model]`.

nanasis:
> One correctness issue here: scoring is now effort-specific, but `allScores` is still keyed only by `model.Model`. Two `ModelRef`s for the same model with different `reasoning_effort` values can therefore overwrite each other, and both candidates are later ranked using whichever effort was written last.
> 
> Could we keep the score per candidate (for example by model + effort or candidate index)? The personalized GMT path appears to have the same issue with `scores[model.Model]`.

Thanks for the comment! Yeah, it is not applicable for current effort-specific scoring, and I've updated the PR to leverage candidate index.

src/semantic-router/pkg/selection/multi_factor_test.go
```diff
@@ -72,6 +72,62 @@ func TestMultiFactor_PicksHighestQualityWhenQualityDominant(t *testing.T) {
...
+		func(string, int) (float64, bool) { return 0, false },
+		func(string, int) (float64, bool) { return 0, false },
+	)
+	result, err = s.Select(context.Background(), &SelectionContext{CandidateModels: []config.ModelRef{
+		{Model: "model", ModelReasoningControl: config.ModelReasoningControl{ReasoningEffort: "low"}},
+		{Model: "model", ModelReasoningControl: config.ModelReasoningControl{ReasoningEffort: "high"}},
+	}})
+	if err != nil {
+		t.Fatal(err)
+	}
```
Xunzhuo, line 110: **Preserve the winning candidate identity**

Impact: this selector can now rank two `ModelRef`s for the same model at different reasoning efforts and choose the later one, but `SelectionResult` keeps only the model name. `selectedModelRefFromResult` then returns the first matching ref, so a high-effort winner can be routed with the earlier low-effort settings; `AllScores` also collapses both candidates under one key.

Required change: carry a stable candidate identity through selection and ExtProc resolution (and make score diagnostics candidate-aware), or reject duplicate model refs, then add an ExtProc-level regression that verifies the selected effort is the one actually returned.

Xunzhuo (commented):
Thanks for tightening the evidence ranking. One routing blocker remains.

Xunzhuo (commented):
Thanks for fixing the direct selector path. Two supported composition paths still lose the winning candidate identity.

src/semantic-router/pkg/selection/automix.go
```diff
@@ -267,20 +270,22 @@ func (a *AutoMixSelector) Select(ctx context.Context, selCtx *SelectionContext)
...
 	logging.Infof("[AutoMix] Evaluating %d candidates (tradeoff=%.2f):",
 		len(sortedCandidates), a.config.CostQualityTradeoff)
-	for _, model := range sortedCandidates {
+	for i, model := range sortedCandidates {
 		modelName := model.Model
-		score := a.computeExpectedValue(modelName, selCtx)
-		allScores[modelName] = score
+		score := a.computeExpectedValue(model, selCtx)
+		candidateScores[i] = score
+		allScores[candidateScoreKey(sortedCandidates, i)] = score
```
Xunzhuo, line 285: **Keep Hybrid candidate-aware**

Impact: duplicate refs now use keys such as `model[index,effort=high]`, but `HybridSelector` still reads every component as `scores[model.Model]` and returns no `SelectedCandidate`. With low/high refs for one model, AutoMix can score them differently here while Hybrid reads both as zero, ties them, and routes the first ref.

Required change: Make Hybrid consume and return exact candidate identities, or reject duplicate refs when Hybrid is used. Add a Hybrid-to-ExtProc regression with two efforts for the same model.

src/semantic-router/pkg/extproc/req_filter_classification.go
```diff
@@ -178,6 +178,9 @@ func (r *OpenAIRouter) applyHybridModelCosts(selector *selection.HybridSelector)
 }
 
 func selectedModelRefFromResult(selCtx *selection.SelectionContext, result *selection.SelectionResult) *config.ModelRef {
+	if result.SelectedCandidate != nil {
```
Xunzhuo, line 181: **Preserve the candidate through Router Learning**

Impact: this shortcut works only while every derived result retains `SelectedCandidate`. Session protection creates a new model-only `SelectionResult`, then resolves it through this function; when it holds a high-effort winner among duplicate refs, resolution silently falls back to the first low-effort ref. The model name is unchanged, so the protection layer does not detect the switch.

Required change: Propagate or reuse the exact selected ref through protection-generated results, or reject ambiguous duplicate refs. Add a regression with Router Learning protection enabled that verifies the winning effort reaches dispatch.

src/semantic-router/pkg/selection/automix.go
nanasis: Addressed in 3c3f6cca47500ba501d69c3c8784210b00072010. Hybrid now combines typed candidate scores using complete candidate identity and returns the exact winning reference; it no longer looks up component values through model-name/display keys. The tests-first regression uses a real AutoMix component and checks both low/high variants through ExtProc and final provider JSON. The regression suite passes, along with the full Go module, affected race tests, and the 32-case response-api E2E profile. Leaving this thread open for your confirmation. The PR currently also needs synchronization with upstream main to resolve merge conflicts.

src/semantic-router/pkg/extproc/req_filter_classification.go
nanasis: Addressed in 3c3f6cca47500ba501d69c3c8784210b00072010. Protection-generated results preserve the exact candidate, session memory defensively stores/restores its controls, and candidate changes are compared by full identity rather than model name. A request-local candidate view also fixes the later provider-encoding lookup that could still send low after a high-effort selection. Regression coverage includes session/conversation protection, tool-loop locks, and cross-request restoration through both Chat and Responses encoding. All regressions and the full Go module pass; the local response-api E2E profile passed 32/32. Leaving resolution to your review.

src/semantic-router/pkg/extproc/request_candidate.go
```diff
@@ -0,0 +1,19 @@
...
+// decisionForCandidate provides a request-local reasoning view without mutating
+// the shared decision or its plugin/candidate configuration. Both standard wire
+// codecs and provider extensions must use the same exact post-policy choice.
+func (ctx *RequestContext) decisionForCandidate(model string) *config.Decision {
+	if ctx == nil || ctx.VSRSelectedDecision == nil {
+		return nil
+	}
+	ref := ctx.VSRSelectedCandidate
+	if ref == nil || (ref.Model != model && (ref.LoRAName == "" || ref.LoRAName != model)) {
+		return ctx.VSRSelectedDecision
```
Xunzhuo, line 14: **Carry the exact candidate through capability fallback**

Impact: capability fallback changes `dispatch.logicalModel` but leaves `VSRSelectedCandidate` and `useReasoning` on the original winner, so this branch reopens the full decision. With duplicate refs for the fallback model, provider encoding can then take the first ref—even if eligibility excluded it—and emit the wrong reasoning effort or mode.

Required change: carry the exact eligible `ModelRef` through rerouting, update the request-local reasoning state, and cover narrowed eligibility with duplicate-effort fallback refs.

Xunzhuo (commented):
Thanks for preserving exact candidate identity through the selector, Hybrid, and Router Learning paths. One dispatch-time identity gap remains on this head.

src/semantic-router/pkg/extproc/request_candidate.go
nanasis: After reviewing the current decision flow, I found that candidates were being selected before we filtered for suitable models, which was not reasonable. I addressed this by moving the filtering logic ahead of candidate selection and added tests covering these scenarios.

src/semantic-router/pkg/extproc/router_learning_protection.go
```diff
@@ -464,26 +464,26 @@ func sessionScopeProtectedResult(
...
-	score := 0.0
-	if baseResult.AllScores != nil {
-		score = baseResult.AllScores[current]
-	}
-	allScores := cloneSelectionScores(baseResult.AllScores)
-	if allScores == nil {
-		allScores = map[string]float64{}
+	ref := selection.CurrentSessionCandidate(learningCtx, baseResult, current)
+	if ref == nil {
+		return nil, false
```
Xunzhuo, line 488: **[P2] Reject an excluded exact owner instead of falling back**

If a session recorded the high-effort candidate and a later decision's quality floor admits only the low-effort candidate for the same model, the model-name-only ownership check still passes. This branch returns `false`; `SessionAwareSelector` then returns `ErrNoEligibleCandidates`, but `selectProtectionResult` swallows it and `applyProtectionSwitch` returns the new low-effort base candidate as `protection_unavailable`. An active tool loop therefore changes its recorded candidate despite the hard lock. Make ownership validation exact and propagate this rejection through the protection caller; extend the restoration test to exclude the previous candidate.

Xunzhuo (changes requested):
The normal exact-candidate and provider-encoding paths are now covered, but protection still falls back to a different candidate when the recorded owner is excluded during an active tool loop. Please preserve this fail-closed result through the protection caller and add the excluded-owner regression.

src/semantic-router/pkg/extproc/router_learning_protection.go
nanasis: Addressed in 8cff1bc59062e939feac07fcfe36aeb2e49440a3. Added the excluded-owner tests first and reproduced the failure. Ownership validation now compares the recorded exact candidate; selectProtectionResult and the protection/learning callers return errors rather than collapsing a policy denial into false.

src/semantic-router/pkg/extproc/router_learning_protection.go
```diff
@@ -132,25 +134,37 @@ func (r *OpenAIRouter) applyProtectionSwitch(
...
+		hardOwnership := session != nil && (session.ActiveToolLoop || session.HasNonPortableContext)
+		if preflight.mode == config.DecisionAdaptationModeApply && (errors.Is(err, selection.ErrNoEligibleCandidates) || hardOwnership) {
+			return routerLearningDecision{}, fmt.Errorf("router learning protection: %w", err)
+		}
 		decision.selectionContext = input.selCtx
 		decision.selectionResult = input.baseResult
 		decision.selectedModelRef = input.selectedModelRef
 		decision.policy = newProtectionPolicy(input.ctx, preflight.config, preflight.mode, routerLearningActionHoldCurrent, "protection_unavailable", preflight.scope)
-		return decision
+		if preflight.mode == config.DecisionAdaptationModeObserve {
```
Xunzhuo: [P2] Preserve adaptation when protection only observes

If adaptation has chosen a different eligible candidate, a persisted exact session owner outside the currently admitted candidates makes protection return `ErrNoEligibleCandidates` at a tool-loop boundary. This error path resets the decision to the original base candidate before setting `observe_only`, so observing protection silently undoes the applied adaptation. A regression through `applyRouterLearning` returns the original candidate with `applied=false` while the adaptation policy records a sampled switch. Preserve the proposal in observe mode and add coverage combining adaptation apply, protection observe and an excluded prior owner.

Xunzhuo (changes requested):
The exact-owner rejection now propagates through apply mode, resolving the previous fallback finding. One observe-mode regression remains: protection can roll back an adaptation choice while reporting observe_only. The combined adaptation/protection regression reproduces this on the current head; preserving the proposal in observe mode passes it and the existing observe-mode controls. No deployed stack regression was run.

Xunzhuo (approved):
The observe-mode composition issue is resolved. Selection now stages session ownership until provider encoding succeeds, and cancellation or a late failure cannot overwrite a newer dispatch. The new selection-path regressions, all seven Router Learning scenarios, and all 32 Response API integration cases pass in the current-head CI. I did not run an additional local deployment.

wilsonwu (approved):
LGTM

PR #3689, opened 2026-09-10
[Feature] Score the response stage rules on the streaming response path

subin9:
## Purpose

Related #3272

#3335 and #3531 moved response jailbreak and hallucination detection onto response stage signals, on the buffered path only. A streamed response was never scored: the finalizer rebuilt the terminal answer and passed it to the response cache, memory and Router Replay, but no rule ran over it, so a declared rule produced no observation, no Router Replay outcome and no metric.

This is the first migration described in #3272: emit a terminal observation, and do not touch bytes that are already sent. Both response paths now score the rules and record the outcome. Nothing enforces on a streamed response, so its Router Replay outcome carries `enforcement: not_enforced_streaming` in place of an action, and a stream that never reaches a terminal answer is not checked at all.

The detector runs inline in the finalizer, delaying the last chunk rather than the answer, because the observation writes signal maps the same request keeps reading. Enforced streaming inspection needs a buffering mode with its own latency contract and stays out of scope.

## Test Plan

Three cases in `pkg/extproc` drive `handleSemanticStreamingResponseBody`: a streamed jailbreak detection, a streamed hallucination check, and an aborted stream. Removing the one call in the finalizer fails the first two.

`response-jailbreak-streaming-passthrough` asserted the old behaviour on purpose, with a comment saying so, and now asserts the observation alongside the unchanged delivery. `hallucination-streaming-observation` is new. Both read `llm_signal_extraction_total`, since a streamed response is observed and never enforced and no header reports it.

## Test Result

`make check BASE_REF=origin/main` passed, with `make harness-check`, the whitespace check and `go mod tidy -diff`.

`make e2e-test E2E_PROFILE=response-jailbreak` reported `Total: 3 | Passed: 3 | Failed: 0`, the streaming case printing `status=200 frames=1049 content_bytes=25142 warnings="" scored=3->4`.

`make e2e-test E2E_PROFILE=hallucination` reported `Total: 2 | Passed: 2 | Failed: 0`, the streaming case printing `status=200 frames=2 content_bytes=20 warnings="" checked=5->6`.


FAUST-BENCHOU (commented):
LGTM. Scoring only after a successfully reconstructed terminal stream adds the missing response-stage observation without pretending that enforcement occurred after the response bytes were already delivered.

Xunzhuo (approved):

PR #3694, opened 2026-09-10
[Bug] Report the real satisfied probability below the threshold

### LINKED ISSUE #3534: [Bug] Feedback detector reports a fabricated confidence below its threshold
### Proposed Workgroup

### Describe the bug

`feedback_detector.go:206-209` flips any below-threshold prediction to `satisfied` and sets the confidence to `1 - confidence`:

```go
if confidence < threshold {
    feedbackType = FeedbackLabelSatisfied
    confidence = 1.0 - confidence
}
```

The model has four classes (`SAT`, `NEED_CLARIFICATION`, `WRONG_ANSWER`, `WANT_DIFFERENT`), so `1 - confidence` is the mass on the other three, not P(SAT). The identical two lines in `fact_check_classifier.go:143` are correct, because that model is binary.

### To Reproduce

Through the shipped path on 63e91c21, `threshold: 0.7` from `config.yaml:2442`:

| input | model argmax | reported |
|---|---|---|
| `looks good to me` | WANT_DIFFERENT 0.6395 | satisfied 0.3605 |
| `this is what I wanted` | WANT_DIFFERENT 0.5931 | satisfied 0.4069 |

### Expected behavior

The reported confidence is P(SAT). That cannot be done without a new accessor: `ClassifyMmBert32KJailbreakWithProbs` exists (`candle-binding/semantic-router.go:2517`) but `ClassifyMmBert32KFeedback` (`:2574`) returns argmax only, which is presumably why the shortcut is there. Adding the probability variant is the same move #3334 made for jailbreak.

### Impact

The value reaches routing as a signal confidence and always overstates P(SAT).

subin9:
Closes #3534

## Purpose

`FeedbackDetector.Classify` relabelled any prediction under the configured threshold as satisfied and reported `1 - confidence` beside that label (`feedback_detector.go:218-220` on main):

```go
if confidence < threshold {
    feedbackType = FeedbackLabelSatisfied
    confidence = 1.0 - confidence
}
```

The detector has four classes (`SAT`, `NEED_CLARIFICATION`, `WRONG_ANSWER`, `WANT_DIFFERENT`), so `1 - confidence` is the mass on the other three, not P(SAT). The same two lines in `fact_check_classifier.go:149-153` are correct, because that model is binary.

The shortcut existed because the binding returned an argmax only. Both feedback backends now expose the whole distribution, as #2594 did for the jailbreak risk score, and the detector reads P(satisfied) at the index the loaded mapping gives that class. A mapping that names no satisfied class keeps the model's own prediction.

Measured on `models/mmbert32k-feedback-detector-merged`, CPU, threshold 0.7:

| input | model argmax | reported before | reported now |
| --- | --- | --- | --- |
| `looks good to me` | want_different 0.639467 | satisfied 0.360533 | satisfied 0.0000036 |
| `this is what I wanted` | want_different 0.593091 | satisfied 0.406909 | satisfied 0.067740 |
| `that is not what I asked for` | wrong_answer 0.771458 | wrong_answer 0.771458 | wrong_answer 0.771458 |

The first row's distribution is `[3.56e-06, 4.13e-05, 0.36049, 0.63947]`, so the 0.3605 reported as P(satisfied) was P(wrong_answer) to four digits. The checkpoint calling `looks good to me` want_different is its own problem, separate from what the router reports about it.

The label is unchanged, so no `user_feedback` rule matches differently. The number changes, on the signal metric and on the `confidence` field of `POST /api/v1/classify/user-feedback`.

## Test Plan

`go test ./pkg/classification/`. The unit cases cover the threshold rule, a mapping that puts satisfied at a non-zero index, and the two cases where P(satisfied) does not exist. The real-model test compares what the detector reports against the model's own vector rather than against a fixed number, and skips when the model is absent. That endpoint has no E2E case today and returns no probability vector, so the comparison lives there instead.

## Test Result

```
feedback_detector_realmodel_test.go:114: text="looks good to me" argmax=3 confidence=0.639467 probabilities=[3.560283e-06 4.1257626e-05 0.36048785 0.63946736] reported=satisfied 0.000004
feedback_detector_realmodel_test.go:114: text="this is what I wanted" argmax=3 confidence=0.593091 probabilities=[0.067740224 0.11362962 0.22553943 0.5930907] reported=satisfied 0.067740
ok  	github.com/vllm-project/semantic-router/src/semantic-router/pkg/classification	65.245s
ok  	github.com/vllm-project/semantic-router/candle-binding	3.920s
```

The existing feedback tests pass unchanged, including #3568's mapping cross-check cases.


FAUST-BENCHOU (commented):
LGTM. Reading the satisfied-class probability directly from the full distribution fixes the multiclass confidence semantics, and using the loaded mapping avoids assuming satisfied is always at a fixed class index.

Xunzhuo (approved):

Issue #3688, opened 2026-09-10
[Bug] The runtime config API cannot write on any shipped Kubernetes deployment


ankit373:
Current-main update: #3707 removed the non-atomic fallback, cleans up failed temporary writes, and syncs the temporary file in both Router and Dashboard writers (with a best-effort directory sync). The Kubernetes ConfigMap-backed runtime-write contract remains unresolved; this issue stays open for that outcome.

Original report (historical context):

### Proposed Workgroup

Enterprise & Environment (deployment architecture and production readiness).

### Describe the bug

Every write path for the runtime config goes through `writeConfigAtomically`, and on every shipped Kubernetes deployment the file it writes is a read-only ConfigMap `subPath` mount. So `PUT`/`PATCH /config/router`, `/config/router/rollback`, the Dashboard deploy, and KB persistence cannot persist anything on a supported deployment. They fail with a 500 and leave a stray file behind.

### Evidence

Every manifest mounts it the same way:

```
deploy/kubernetes/istio/deployment.yaml          mountPath: /app/config/config.yaml  subPath: config.yaml  readOnly: true
deploy/kserve/deployment.yaml                    same
deploy/openshift/deployment.yaml                 same
deploy/openshift/template.yaml                   same
deploy/openshift/deployment-simulator.yaml       same
deploy/observability/dashboard/deployment.yaml   same
deploy/helm/semantic-router/templates/*.yaml     configMap backed
```

No deployment backs `config.yaml` with an emptyDir or a PVC.

Against that, `writeConfigAtomically` (`src/semantic-router/pkg/apiserver/route_router_config_update.go:458`, duplicated at `dashboard/backend/handlers/runtime_config_apply.go:69`):

```go
tmpConfigFile := configPath + ".tmp"
os.WriteFile(tmpConfigFile, yamlBytes, 0o644)   // lands on the container fs
if err := os.Rename(tmpConfigFile, configPath); err != nil {
    os.WriteFile(configPath, yamlBytes, 0o644)  // fallback onto the mount
}
```

With a `subPath` mount the destination is a bind mount from a different filesystem, so `rename(2)` returns `EXDEV`. The fallback then writes directly to the mount, which is `readOnly`, so it returns `EROFS`. The temp file is never cleaned up, so each attempt leaves another `config.yaml.tmp` on the container filesystem.

### Impact

[...]

PR #3707, opened 2026-09-10
[Bug] Don't fall back to a non-atomic write when the config rename fails

AmirF194:
Related #3688

## Purpose

`writeConfigAtomically` (in `dashboard/backend/handlers/runtime_config_apply.go`, and duplicated in `src/semantic-router/pkg/apiserver/route_router_config_update.go`) falls back to a plain `os.WriteFile(configPath, ...)` whenever the rename fails. That fallback is exactly what the function exists to avoid: I bind-mounted a config file the same way a K8s subPath mount does, and rename fails with "device or resource busy" while the direct write to the mount still succeeds, so a crash mid-write would leave `configPath` truncated instead of the write just not happening.

This is the narrower of the two problems ankit373 flagged in #3688, next to the read-only-mount question that needs a Workgroup call on whether the config belongs on a writable volume at all. This PR doesn't touch that. It removes the unsafe fallback so a failed rename returns an error instead of degrading silently, and adds the fsync-before-rename and best-effort directory fsync ankit373 also called out, so a successful write is actually durable. Same fix in both copies of the function; I didn't fold the two copies into one, that's a bigger change than this needs.

## Test Plan

```bash
cd dashboard/backend
go build ./handlers/... && go vet ./handlers/...
go test ./handlers/... -run TestWriteConfigAtomically -v
golangci-lint run ./handlers/... --config ../../tools/linter/go/.golangci.yml
gofmt -l handlers/runtime_config_apply.go ../../src/semantic-router/pkg/apiserver/route_router_config_update.go
```

Plus a standalone repro: bind-mount a regular file over a config path (`mount --bind`, no read-only flag) and run each copy of `writeConfigAtomically` against it.

## Test Result

`dashboard/backend/handlers` builds, vets, and lints clean. The three new tests pass: a normal write, a forced rename failure that must leave the existing file untouched, and a temp-file create failure that must not leave a stray `.tmp` behind.

Against the real bind mount: on `main`, rename fails and the fallback overwrites the file anyway (confirmed the content actually changes). With this change, the function returns the rename error and the file keeps its original content.

I didn't get `src/semantic-router/pkg/apiserver` itself to build, it's cgo-gated behind the compiled candle-binding Rust library and building that just to check one function felt disproportionate. Both copies are the identical fix, gofmt is clean on both, and I ran the second file's exact function body standalone against the same bind-mount failure and got the same result, but `go test ./pkg/apiserver/...` itself is unverified here.

---
<details>

</details>


AayushSaini101 (approved):

wilsonwu (approved):
LGTM

FAUST-BENCHOU (approved):

PR #3711, opened 2026-09-11
[Bug] Use the matched domain's model_scores in static selection

### LINKED ISSUE #3709: [Bug] Static selection ignores the matched domain's `model_scores`
### Proposed Workgroup

MoM & Routing

### Describe the bug

`StaticSelector.InitializeFromConfig` stores `model_scores` by domain name (`src/semantic-router/pkg/selection/static.go:76-82`), but `Select` looks them up by decision name (`static.go:112`). Unless a decision has the same name as its domain, every candidate gets the default score `1.0` and the first `modelRef` wins.

The request path already passes the matched domain as `SelectionContext.CategoryName` (`extractDecisionCategory` in `pkg/extproc/req_filter_classification_runtime.go:279`). The static selector never reads it. The static selection docs say "A matched domain can supply fixed `model_scores`".

### To Reproduce

```yaml
routing:
  signals:
    domains:
      - name: business
        model_scores:
          - model: small
            score: 0.2
          - model: large
            score: 0.9
  decisions:
    - name: business_route
      rules:
        operator: OR
        conditions:
          - type: domain
            name: business
      modelRefs:
        - model: small
        - model: large
      algorithm:
        type: static
```

1. Load the config with `config.ParseYAMLBytes`.
2. Build the static selector from `cfg.Categories`.
3. Select with `DecisionName: "business_route"` and `CategoryName: "business"`.

```
decision="business_route" scores=map[large:1 small:1] -> selected=small
```

If you rename the decision to `business`, the scores are applied:

```
decision="business" scores=map[large:0.9 small:0.2] -> selected=large
```

### Expected behavior

`large` is selected, because the matched domain `business` scores it higher.

### Impact

A static route that relies on domain `model_scores` always uses its first `modelRef` unless the decision is named after the domain. Naming the decision after the domain is the only workaround. `TestStaticSelector_Select` uses the same string for both the category and the decision, so it does not catch this.

### Additional context

The Elo `category_weighted` path (`elo.go:306`, `elo.go:381`) and `rl_driven.go:966` also build per-category state from `category.Name` and read it back by `DecisionName`. I have only reproduced the static case.

Hiro208:
Closes #3709

## Purpose

`StaticSelector` stores `model_scores` under the domain name, but `Select` looked them up by the decision name. A decision that isn't named after its domain therefore got the default score for every candidate and always picked its first `modelRef`. Static is also the fallback when a decision has no `algorithm`, so this affects every multi-model decision that relies on domain `model_scores`.

`Select` now looks up the matched domain first, using `SelectionContext.CategoryName`, which extproc already sets. It falls back to the decision name when that domain has no scores. Only `pkg/selection` changes. Workgroup: MoM & Routing.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/selection/ -run TestStaticSelector -v
go test ./pkg/selection/
gofmt -l pkg/selection
```

## Test Result

- `TestStaticSelectorScoresByMatchedDomain` fails on `main` with `expected large, got small (scores map[large:1 small:1])` and passes with this change.
- `go test ./pkg/selection/` passes and `gofmt` is clean. The native bindings were stubbed through `go test -overlay`, because this machine has no cgo toolchain.

---
<details>

</details>


FAUST-BENCHOU (approved):
Thnaks

wilsonwu (approved):
LGTM

PR #3712, opened 2026-09-11
[Bug] Accept vLLM 0.21.0 prompt_routed_experts response field (#3708)

### LINKED ISSUE #3708: [Bug] vLLM 0.21.0 non-streaming chat responses rejected by prompt_routed_experts
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Since #3076 the router validates the upstream provider response against a canonical
OpenAI schema and rejects any field outside it (`llmprotocol.DefaultPolicy()`
hardcodes `UnknownReject`). vLLM 0.21.0 adds a top-level `prompt_routed_experts`
field to **non-streaming** `/chat/completions` responses (it is always `null` for
dense models; it carries real routed-experts when prompt routing is active). The
canonical `chatResponseWire` has no such field, so a non-streaming chat response
from a vLLM 0.21.0 backend fails with `invalid_upstream_json: upstream response
JSON contains a non-canonical field` and the router returns 502 — the router is
unusable against vLLM 0.21.0 as a backend for non-streaming chat.

This is the same class of failure as #3496 (strict schema rejecting vendor
response extensions), but for the vLLM vendor field. There is no configuration
workaround: `llmprotocol.DefaultPolicy()` hardcodes `UnknownReject` and no YAML
key maps to `Policy.UnknownFields`.

### To Reproduce

1. Run a vLLM 0.21.0 server (`vllm serve <model>`).
2. Send a non-streaming chat request through the router (Envoy ext_proc):
   `POST /v1/chat/completions` with `{"model": ..., "messages": [...], "stream": false}`.
3. Observe `502`; the router log shows `invalid_upstream_json: upstream response
   JSON contains a non-canonical field`.
4. Decoding the same upstream body with `OpenAIChatCodec.DecodeResponse` under
   `llmprotocol.DefaultPolicy()` fails for the same reason; removing the
   `prompt_routed_experts` field from the response makes it succeed.

Notes:

- The field only appears on the top level of non-streaming responses. Streaming
  chunks are unaffected.
- `git log -S prompt_routed_experts` shows the field has never existed in the
  wire structs.

### Expected behavior

A vLLM 0.21.0 backend's non-streaming chat responses should route successfully. Since the
field is additive and irrelevant to routing/translation (null for dense models),
it should be tolerated — either as a null-only wire field (mirroring
`prompt_text`/`routed_experts`) or via a scoped vendor-extension allowance such as the one proposed for Azure in open PR #3499.

### Impact

vLLM 0.21.0 backends currently fail non-streaming chat through the router.
Affects the widely deployed vLLM-as-backend path (the router's primary ecosystem
hardware pairing). No configuration workaround. Streaming chat, Responses, and
Anthropic traffic are unaffected.

### Additional context

- Related: #3496 (Azure vendor response fields) and open implementation PR #3499, which proposes a scoped vendor allowance only to Azure hosts. Current `main` still rejects this vLLM field, so this issue is independent.
- A local verification: patching a `*nullOnlyWire` field for
  `prompt_routed_experts` (same pattern as `prompt_text`) makes the real engine's
  response decode successfully without relaxing `UnknownReject` globally.
- Version anchor: verified in vLLM source — `prompt_routed_experts` and
  `routed_experts` appear in v0.21.0 (`chat_completion/protocol.py`) and are absent
  in v0.20.0. The top-level `prompt_routed_experts` field was removed by
  v0.21.1rc0 and remains absent in v0.22.0 and later releases. In v0.21.0 the
  fields are unconditionally serialized on non-streaming responses
  (`JSONResponse(model_dump())`, no `exclude_none`), so the failure is not gated
  by `--enable-return-routed-experts` — that flag only controls whether the value
  is non-null (routing replay data), not whether the field is emitted.

Bevisy:
Closes #3708

## Summary

vLLM 0.21.0 adds top-level `prompt_routed_experts` (and per-choice
`routed_experts`) fields to non-streaming `/chat/completions` responses. The
canonical `chatResponseWire` has no matching top-level field, and
`llmprotocol.DefaultPolicy()` hardcodes `UnknownReject`, so a non-streaming chat
response from any vLLM 0.21.0 backend fails with `invalid_upstream_json:
upstream response JSON contains a non-canonical field` and the router returns
502.

This adds `prompt_routed_experts` as a null-only wire field (`*chatNullOnlyWire`,
same pattern as `prompt_text`/`routed_experts`), matching v0.21.0 semantics where
the field is unconditionally serialized but null for dense models. Streaming
chunks are unaffected.

Note: the top-level field was removed in v0.21.1rc0 and remains absent in
v0.22.0+; per-choice `routed_experts` is already covered by the existing
null-only field. So this is scoped to v0.21.0 backends.

## Changes

- `codec_openai_chat_response_wire.go`: add `PromptRoutedExperts *chatNullOnlyWire`
- Sync the four contract surfaces (repo rule for wire-field changes):
  - `official_schema_inventory_test.go` (extension field list)
  - golden matrix fixture `016-chat-official-response-fields-*.json` (in + 3 out)
  - `tools/mock-vllm/schema_contract.json` (extension_response_fields)

## Verification

- `go test ./pkg/protocolcodec/` green (incl. ability-matrix 016 cases and the
  three field-inventory "closed" tests that lock wire ↔ contract parity)
- `gofmt` clean, `go vet` clean
- End-to-end: real vLLM 0.21.0 engine (qwen3-4b) non-streaming chat via router
  returned HTTP 200 after the fix; 502 before.

### DIFF
--- a/src/semantic-router/pkg/protocolcodec/codec_openai_chat_response_wire.go
+++ b/src/semantic-router/pkg/protocolcodec/codec_openai_chat_response_wire.go
@@ -9,29 +9,30 @@ import (
 )
 
 type chatResponseWire struct {
-	ID                string                    `json:"id"`
-	Object            string                    `json:"object,omitempty"`
-	Created           int64                     `json:"created"`
-	Model             string                    `json:"model"`
-	Choices           []chatChoiceWire          `json:"choices"`
-	Usage             *chatUsageWire            `json:"usage,omitempty"`
-	Metadata          map[string]string         `json:"metadata,omitempty"`
-	Moderation        json.RawMessage           `json:"moderation,omitempty"`
-	Error             *chatErrorWire            `json:"error,omitempty"`
-	ServiceTier       *chatServiceTierWire      `json:"service_tier,omitempty"`
-	SystemFingerprint *string                   `json:"system_fingerprint,omitempty"`
-	PromptLogprobs    *chatNullOnlyWire         `json:"prompt_logprobs,omitempty"`
-	PromptTokenIDs    []int64                   `json:"prompt_token_ids,omitempty"`
-	PromptText        *chatNullOnlyWire         `json:"prompt_text,omitempty"`
-	KVTransferParams  *chatKVTransferParamsWire `json:"kv_transfer_params,omitempty"`
-	ECTransferParams  *chatNullOnlyWire         `json:"ec_transfer_params,omitempty"`
-	Metrics           *chatNullOnlyWire         `json:"metrics,omitempty"`
-	DoRemoteDecode    *bool                     `json:"do_remote_decode,omitempty"`
-	DoRemotePrefill   *bool                     `json:"do_remote_prefill,omitempty"`
-	RemoteBlockIDs    []int64                   `json:"remote_block_ids,omitempty"`
-	RemoteEngineID    *string                   `json:"remote_engine_id,omitempty"`
-	RemoteHost        *string                   `json:"remote_host,omitempty"`
-	RemotePort        *int64                    `json:"remote_port,omitempty"`
+	ID                  string                    `json:"id"`
+	Object              string                    `json:"object,omitempty"`
+	Created             int64                     `json:"created"`
+	Model               string                    `json:"model"`
+	Choices             []chatChoiceWire          `json:"choices"`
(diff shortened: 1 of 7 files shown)


FAUST-BENCHOU (approved):
thanks

wilsonwu (approved):
LGTM

Issue #3719, opened 2026-09-11
[Bug] The ONNX build does not compile after the feedback probability change


subin9:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

`src/semantic-router/go.onnx.mod` replaces `candle-binding` with `onnx-binding`, so
the ONNX build compiles `pkg/classification` against the ONNX package. #3694 added
two calls to the Candle API that `onnx-binding` does not declare, so that build no
longer compiles on main:

```
pkg/classification/feedback_detector.go:196:18: undefined: candle.ClassifyMmBert32KFeedbackWithProbs
pkg/classification/feedback_detector.go:198:17: undefined: candle.ClassifyFeedbackTextWithProbs
```

`onnx-binding/semantic-router.go` has `ClassifyMmBert32KFeedback` and
`ClassifyFeedbackText` but not the `WithProbs` variants. The jailbreak pair in the
same file already carries that split, so the shape to copy is there.

### To Reproduce

On `main` at 7ab02d05:

```
cd src/semantic-router
go build -modfile=go.onnx.mod -tags=onnx ./pkg/...
```

### Expected behavior

The ONNX build compiles, so the E2E lane can build its image.

### Impact

Every pull request whose changed files select an E2E profile fails while building
the router image rather than on a test. It showed up on the hallucination and
response-jailbreak profiles of #3716, where the two failing jobs carry no change of
their own.

`applyThreshold` already keeps the model's own prediction when no probability is
available, so an ONNX wrapper that reports an empty distribution is safe.

The gap went unnoticed because the ONNX Go build is not a required gate. #2477 and
#3320 track making it one.

Xunzhuo:
Fixed on current `main` by #3724: the ONNX binding now provides both feedback `WithProbs` wrappers, and `go build -modfile=go.onnx.mod -tags=onnx ./pkg/...` passes. Closing as complete.

PR #3720, opened 2026-09-11
[Bug] Declare the feedback probability calls in the ONNX binding

subin9:
## Purpose

Related #3719

`go.onnx.mod` replaces `candle-binding` with `onnx-binding`, so the ONNX build
compiles `pkg/classification` against the ONNX package. The feedback detector
started calling `ClassifyMmBert32KFeedbackWithProbs` and
`ClassifyFeedbackTextWithProbs`, which that package never declared, and the build
stopped compiling on main.

Both wrappers follow the jailbreak pair in the same file and report an empty
distribution, which `applyThreshold` already treats as no probability available.

## Test Plan

`go build -modfile=go.onnx.mod -tags=onnx ./pkg/...` in `src/semantic-router`,
which is the command that fails on main, and `go build ./...` in `onnx-binding`.

## Test Result

Both build. Before the change the first one reports:

```
pkg/classification/feedback_detector.go:196:18: undefined: candle.ClassifyMmBert32KFeedbackWithProbs
pkg/classification/feedback_detector.go:198:17: undefined: candle.ClassifyFeedbackTextWithProbs
```

`make check BASE_REF=origin/main` is green.


subin9:
#3724 landed `ClassifyMmBert32KFeedbackWithProbs` and `ClassifyFeedbackTextWithProbs` in `onnx-binding/semantic-router.go`, and both call `classifyWithClassifierProbabilities`, so they return the real distribution rather than the empty slice this branch declared. Nothing in this diff is left to apply, so I am closing it. Issue #3719 is already closed.

Issue #2864, opened 2026-08-11
[Feature] Harden confidence escalation across model and reasoning budgets


Xunzhuo:
## Goal

Coordinate the remaining Confidence-escalation work across completion-token limits, below-threshold exhaustion, verifier outcomes, and request-wide resource limits.

## Proposed Workgroup

MoM & Routing

## Current-main baseline

Current `main` already distinguishes numeric zero from missing confidence evidence through #2869. It also carries `use_reasoning` and `reasoning_effort` on `ModelRef` and applies those controls in the shared authenticated internal dispatch path before provider encoding. The remaining first control gap is an explicit per-model completion-token limit; reasoning mutation must not be duplicated inside Confidence.

## Work item shape and order

This is a parent delivery issue under #2336. The current assignee coordinates one bounded child or reviewable vertical at a time.

1. #3479 adds one shared optional `ModelRef` completion-token limit at the existing internal dispatch seam and proves it through Confidence.
2. Define typed fail, return-best-with-warning, and abstain behavior after below-threshold exhaustion.
3. Integrate #2857's typed verifier result.
4. Add bounded combined escalation using #2861's aggregate ledger.
5. Evaluate each production mode under fixed budgets through #2858 and add capability-specific E2E coverage.

## Configuration and API impact

Extend existing Confidence and shared `ModelRef` contracts rather than adding a new algorithm family. Every new mode is explicit, validated, and backward compatible by default. Replay records effective controls and outcomes.

## Non-goals

- Rework the already-correct zero-score or shared reasoning-control paths.
- Add session cooldown, hysteresis, or cross-request switching; #3377 owns that policy.
- Invoke Fusion or ReMoM from Confidence.
- Add adaptive-N voting or treat uncalibrated confidence as correctness.

## Completion signal

- [ ] Each remaining outcome has a bounded owner and explicit prerequisite.
- [ ] #3479 applies one shared per-model completion limit without duplicating reasoning mutation.
- [ ] Exhaustion and verifier outcomes have deterministic response and Replay semantics.
- [ ] Aggregate limits are reserved before every attempt.
- [ ] Fixed-budget evidence and E2E cover each production mode before enablement.

PR #3721, opened 2026-09-11
[Feature] Add shared per-ModelRef completion-token limits at provider dispatch

### LINKED ISSUE #3479: [Feature] Add shared per-ModelRef completion-token limits at provider dispatch
## Goal

Add an optional per-`ModelRef` completion-token ceiling at the shared provider-dispatch seam so every routing path applies the selected model's bound consistently, with Confidence as the first end-to-end consumer.

## Proposed Workgroup

MoM & Routing

## Background

Current `main` already carries `use_reasoning` and `reasoning_effort` on `ModelRef` and applies those controls through authenticated internal dispatch. Algorithm-specific completion limits also exist, and the request-params plugin can enforce `max_tokens_limit`, but `ModelRef` has no universal completion ceiling. A field on this shared type must not silently affect only Looper calls.

Parent #2864 is assigned and coordinates the Confidence proof; implementation ownership should be agreed there before coding.

## Scope

- Add optional `routing.decisions[].modelRefs[].max_completion_tokens` with positive bounds and deterministic serialization across maintained producers.
- Enforce the selected model's ceiling at the shared provider-dispatch seam for ordinary and Looper routing paths.
- Compute the strictest effective value across client `max_completion_tokens` or legacy `max_tokens`, `plugins[type=request_params].max_tokens_limit`, the selected `ModelRef`, and any algorithm/stage limit.
- Leave a typed composition seam for #2861's future request-wide ledger to contribute another upper bound without making #2861 a prerequisite for this leaf.
- Map the effective value through maintained provider codecs with deterministic unsupported/conflicting capability behavior.
- Prove through Confidence escalation that every selected model receives its own effective bound while existing reasoning mutation still runs exactly once.
- Emit bounded effective-limit and fallback reasons in Replay and add config, DSL, protocol, unit, and maintained E2E tests.

Adds optional `max_completion_tokens` beside existing `ModelRef` controls. Omission preserves current behavior. Request input may narrow but cannot widen operator, model, plugin, algorithm, or future ledger limits.

## Non-goals

- Make a general `ModelRef` field Looper-only or add another Confidence-specific token/reasoning block.
- Move or duplicate `use_reasoning` or `reasoning_effort` mutation inside Confidence.
- Implement #2861's aggregate ledger, below-threshold exhaustion, or verifier integration.
- Expose private reasoning content or assume every provider encodes token limits identically.

## Completion signal

- [ ] One universal `ModelRef` field round-trips through canonical config and maintained producers.
- [ ] Shared provider dispatch applies the bound to ordinary and Looper requests.
- [ ] Precedence always selects the strictest applicable client, plugin, model, algorithm/stage, and available ledger limit.
- [ ] Every Confidence attempt receives the intended model-specific value without duplicating reasoning mutation.
- [ ] Unsupported capability, invalid values, conflicts, and provider encoding are deterministic and tested.

sujal111:
Closes #3479

## Purpose

Add optional `routing.decisions[].modelRefs[].max_completion_tokens` and enforce it as a shared, strictest-wins ceiling at provider dispatch. Ordinary routes and Looper/Confidence hops compose the same sources (client, `request_params.max_tokens_limit`, selected ModelRef, algorithm/stage, and a typed nil ledger slot for #2861) instead of a Looper-only token block or a replace-wins hop mutation. Replay records the effective bound, source, and fallback per request and per Confidence attempt.

Owned by `wg/mom-routing` as the first child of #2864.

## Test Plan

- [x] `make impact ENV=cpu CHANGED_FILES="$(git diff --name-only origin/main...HEAD | tr '\n' ' ')"`
- [x] `make check CHANGED_FILES="$(git diff --name-only origin/main...HEAD | tr '\n' ' ')"` (pre-commit, structure, architecture, golangci-lint, dashboard frontend)
- [x] CGO-free Go tests: `outputtokens`, `config`, `dsl`, `protocolcodec`, `k8s`, `headers`, `routerreplay/store`
- [x] `make vllm-sr-test`
- [x] `make docs-crd`
- [x] `make verify DOMAIN=router-core`
- [x] `make build-e2e`
- [ ] `make test-semantic-router` (blocked on this host: Rust toolchain / candle stubs)
- [ ] `make helm-lint` (Helm not installed)
- [ ] `make verify PROFILE=looper` (no Docker/Podman on this host)

## Test Result

Shared composition, dispatch, Looper hop headers, Replay fields, dashboard ModelRef editor, CLI, DSL, CRDs, and docs all landed together. `make check`, router-core verify, CGO-free unit tests, Python CLI tests, CRD regen, and the e2e compile all passed.

Remaining risk: live `looper` profile E2E and full `test-semantic-router` still need a host with Docker and a working Rust/candle toolchain. This host could not run those.

---
<details>

</details>


dashboard/frontend/src/pages/ConfigPageDecisionModelRefsEditor.tsx
```diff
@@ -174,6 +176,23 @@ function ModelReferenceMetadata({ value, onChange }: Pick<RowProps, 'value' | 'o
...
           />
         </label>
+        <label className={decisionStyles.editorControlLabel}>
+          <span className={decisionStyles.editorControlLabelText}>Max completion tokens</span>
+          <input
+            type="number"
+            value={typeof value.max_completion_tokens === 'number' ? value.max_completion_tokens : ''}
+            onChange={(event) =>
+              onChange(
+                'max_completion_tokens',
```
Xunzhuo, line 187: **Persist this field through the Dashboard save mapper**

The editor updates `max_completion_tokens`, but `decisionModelRefsForSave` reconstructs each ModelRef without copying it. Saving a newly entered value, or editing an existing decision that already has one, therefore silently removes the ceiling. Validate and copy a finite positive integer in the save mapper, and add a load/edit/save round-trip regression.

src/semantic-router/pkg/dsl/decompiler_decisions.go
```diff
@@ -61,6 +61,9 @@ func candidateIterationModelRefOptions(model *config.ModelRef) string {
 	if model.Weight != 0 {
 		opts = append(opts, fmt.Sprintf("weight = %s", strconv.FormatFloat(model.Weight, 'f', -1, 64)))
 	}
+	if model.MaxCompletionTokens != nil {
```
Xunzhuo, line 64: **Compare token limits before omitting the `MODEL` declaration**

`candidateIterationsCoverModelRefs` still compares only model and LoRA. A valid config with `ModelRefs=[a,max=100]` and a single explicit-model iteration `[a,max=200]` is treated as covered, so decompilation drops the `MODEL` declaration and recompilation widens the route ceiling from 100 to 200. Include `max_completion_tokens` (ideally the full serialized ModelRef contract) in the coverage equality and add a differing-metadata round-trip test.

src/semantic-router/pkg/looper/stage_context.go
```diff
@@ -261,3 +264,30 @@ func serializedMessagesTokenEstimate(messages []openai.ChatCompletionMessagePara
...
+
+func attachOutputTokenBounds(options *CallOptions, req *Request, stageReq *openai.ChatCompletionNewParams) {
+	if options == nil {
+		return
+	}
+	if options.ClientMaxOutputTokens == nil && req != nil {
+		options.ClientMaxOutputTokens = chatParamsMaxOutputTokens(req.OriginalRequest)
+	}
+	if options.StageMaxOutputTokens == nil {
+		options.StageMaxOutputTokens = chatParamsMaxOutputTokens(stageReq)
```
Xunzhuo: **Do not relabel an inherited client limit as an algorithm-stage bound**

Most stage requests clone `req.OriginalRequest`; when a stage does not override its token limit, this records the same client value in both the client and stage channels. An internal dispatch with request-params `blocked_params` then suppresses `Client` but retains `AlgorithmStage`, reintroducing the blocked client limit and misreporting its source in Replay. Track whether the algorithm actually supplied an override, and add a Looper/Confidence regression where a client limit is blocked and a different ModelRef ceiling must win.

Xunzhuo (commented):
Thanks for carrying the ceiling through most routing surfaces. Three paths still lose or misapply its semantics.

dashboard/frontend/src/pages/ConfigPageDecisionModelRefsEditor.tsx
sujal111: Fixed in 9f05717. `decisionModelRefsForForm` now loads `max_completion_tokens`, and `decisionModelRefsForSave` copies a finite integer >= 1. `configPageDecisionFormSupport.test.ts` covers load/edit/save plus newly entered and invalid values.

src/semantic-router/pkg/dsl/decompiler_decisions.go
sujal111: Fixed in 9f05717. `candidateIterationsCoverModelRefs` now compares the full serialized ModelRef options, including `max_completion_tokens`. `TestCandidateIterationCoverageKeepsDifferingMaxCompletionTokens` keeps the MODEL declaration and round-trips the route ceiling of 100.

src/semantic-router/pkg/looper/stage_context.go
sujal111: Fixed in 9f05717. Stage bounds are attached only when the algorithm actually overrides the client limit (`stageOverrideMaxOutputTokens`). Confidence hops omit the inherited stage header, so a blocked client of 256 lets ModelRef 1024 win. Covered by Looper header/compose tests and `TestPrepareProviderDispatchBlockedLooperClientLetsModelRefWin`.

Xunzhuo (commented):
Thanks for fixing the Dashboard config form and DSL coalescing issues. The existing stage-provenance blocker remains, and two other paths still lose or misreport the ceiling.

src/semantic-router/pkg/looper/attempt_trace.go
```diff
@@ -505,6 +521,32 @@ func attemptCurrency(spec attemptSpec) string {
...
+func composeAttemptOutputTokenLimit(
+	req *Request,
+	stageReq *openai.ChatCompletionNewParams,
+	model string,
+) outputtokens.Result {
+	sources := outputtokens.Sources{
+		AlgorithmStage: stageOverrideMaxOutputTokens(nil, stageReq),
+	}
+	if req != nil {
+		sources.Client = chatParamsMaxOutputTokens(req.OriginalRequest)
```
Xunzhuo: **Apply blocked-client semantics to Replay accounting**

Impact: Dispatch drops the client limit when `request_params.blocked_params` includes the max-token field, but attempt accounting always composes `OriginalRequest` as a client source. A blocked client limit of 256 with a ModelRef limit of 1024 is therefore dispatched as `model_ref=1024` while Replay records `client=256`, with no blocked fallback.

Required change: Carry the blocked/effective source state into attempt composition (or record the actual dispatch result) and add a persisted-Replay regression for a blocked client limit with a ModelRef ceiling.

src/semantic-router/pkg/dsl/ast_json.go
```diff
@@ -152,14 +152,15 @@ type ModelDeclJSON struct {
...
-	Weight    float64  `json:"weight,omitempty"`
-	Pos       Position `json:"pos"`
+	Model               string   `json:"model"`
+	Reasoning           *bool    `json:"reasoning,omitempty"`
+	Mode                string   `json:"mode,omitempty"`
+	Effort              string   `json:"effort,omitempty"`
+	LoRA                string   `json:"lora,omitempty"`
+	ParamSize           string   `json:"paramSize,omitempty"`
+	Weight              float64  `json:"weight,omitempty"`
+	MaxCompletionTokens *int     `json:"maxCompletionTokens,omitempty"`
```
Xunzhuo, line 162: **Preserve this field through the visual route editor**

Impact: The Go AST now emits `maxCompletionTokens`, but the frontend `ASTModelRef`, `RouteModelInput`, `astModelToInput`, and route serializer omit it. Opening an existing DSL route in the visual editor and saving replaces the whole route block without this value, silently removing its ceiling.

Required change: Carry and serialize the field through the visual route editor (or at minimum preserve it) and add a parse-edit-save round-trip regression.

dashboard/frontend/src/lib/dslMutations.ts
```diff
@@ -532,6 +550,9 @@ function serializeRouteBody(input: RouteInput): string {
       if (m.paramSize) attrs.push(`param_size = "${m.paramSize}"`)
       if (m.weight !== undefined) attrs.push(`weight = ${m.weight}`)
       if (m.reasoningFamily) attrs.push(`reasoning_family = "${m.reasoningFamily}"`)
+      if (isPositiveInt(m.maxCompletionTokens)) {
```
Xunzhuo: **Block saving while this ceiling is invalid**

Impact: The new validation only annotates the preview; Save and Create remain enabled, and this serializer drops every non-positive or non-integer value. Editing an existing ceiling from `64` to `0` and saving therefore removes the limit and widens the route.

Required change: Prevent Save/Create while route validation has an error (or reject/preserve invalid input at the mutation boundary), and change the regression to prove an invalid edit cannot delete an existing ceiling.

src/semantic-router/pkg/dsl/parser.go
```diff
@@ -516,6 +516,11 @@ func rawToModelRef(r *rawModelRef) *ModelRef {
 			} else if v.Float != nil {
 				m.Weight = *v.Float
 			}
+		case "max_completion_tokens":
```
Xunzhuo, line 536: **Reject invalid ceiling values during DSL compilation**

Impact: The option grammar accepts integer, float, and string values, but this branch silently ignores non-integers and retains non-positive integers. For example, `max_completion_tokens = 512.0` compiles with the ceiling omitted, while `0` or `-1` can be emitted as runtime-invalid config, so CLI compilation can produce widened or unloadable policy without an authoring error.

Required change: Validate the raw type and positive range for route and candidate-iteration ModelRefs before compilation/emission, and add wrong-type plus non-positive DSL regressions.

Xunzhuo (commented):
Thanks for the latest fixes. Two authoring paths still need to reject invalid ceilings instead of silently changing their meaning.

dashboard/frontend/src/lib/dslMutations.ts
sujal111: Fixed in 87aabac. Save/Create are now disabled while route validation has an error, `handleSave`/`handleSubmit` no-op in that state, and `updateRoute` throws instead of dropping a non-positive ceiling. The visual-editor regression now proves editing `64` to `0` cannot delete the existing limit.

src/semantic-router/pkg/dsl/parser.go
sujal111: Fixed in 87aabac. Route and candidate-iteration ModelRefs now reject non-integer and non-positive `max_completion_tokens` during parse, so CLI compilation cannot omit or emit an invalid ceiling. Added wrong-type (`512.0`, `"512"`) and non-positive (`0`, `-1`) DSL regressions.

sujal111:
The earlier Core tests / pre-commit failures are fixed on `d10c384`.

`Operator / Integration Test (memory)` failed after Kind loaded the 1.4GB extproc image without the `semantic-router-extproc:test` tag (it appears as `<none>`). That is a Kind image-load flake, not a product regression. A rerun of that job should clear PR Gate. I cannot rerun it from this account (needs repo admin).

sujal111:
Pushed empty `7599a2a` to retrigger CI. The last Operator memory failure was the Kind extproc image-load flake (image present as `<none>`), not a product change.

src/semantic-router/pkg/extproc/output_token_limit.go
```diff
@@ -0,0 +1,162 @@
...
+	ctx *RequestContext,
+) outputtokens.Sources {
+	sources := outputtokens.Sources{
+		Client:         outputtokens.Clone(ctx.ClientMaxOutputTokens),
+		AlgorithmStage: outputtokens.Clone(ctx.AlgorithmStageMaxOutputTokens),
+		Ledger:         ctx.OutputTokenLedger,
+	}
+	if outputTokenLimitBlocked(ctx) {
+		sources.Client = nil
+	} else if sources.Client == nil && request != nil {
```
Xunzhuo: **Recompose reroutes from the immutable client snapshot**

Impact: The first dispatch writes model A's effective ceiling into `request.Sampling.MaxOutputTokens`. When the ingress client omitted a limit, the capability-reroute pass reaches this fallback with `ctx.ClientMaxOutputTokens == nil` and reclassifies A's already-mutated ceiling as a client bound. For example, A=256 rerouted to B=1024 remains capped at 256 and Replay reports `client` instead of B's `model_ref`.

Required change: Preserve explicit client absence and compose every dispatch from immutable ingress/Looper source snapshots, not a prior dispatch mutation. Add an omitted-client capability-reroute regression with different A/B ceilings and assert both value and provenance.

src/semantic-router/pkg/configschema/router-config-v0.3.schema.json
```diff
@@ -3856,6 +3856,9 @@
         "weight": {
           "type": "number"
         },
+        "max_completion_tokens": {
+          "type": "integer"
```
Xunzhuo: **Publish the positive bound in the canonical schema**

Impact: The canonical JSON schema accepts zero and negative `max_completion_tokens` values even though Go validation, the CLI, Dashboard, and CRDs reject them. Schema-based producers can therefore emit configuration that the router cannot load, contrary to the positive-bound contract.

Required change: Add `minimum: 1` at the schema source, regenerate this artifact, and add a schema-validation regression for zero and negative values.

src/semantic-router/pkg/extproc/output_token_limit.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return nil
+}
+
+func wireFormatSupportsOutputTokenLimit(dispatch *providerDispatch) bool {
+	if dispatch == nil {
+		return true
+	}
+	switch dispatch.targetFormat {
+	case llmprotocol.OpenAIChatV1, llmprotocol.AnthropicMessagesV1, llmprotocol.OpenAIResponsesV1:
```
Xunzhuo, line 154: **Do not remove a strict ceiling to satisfy the Responses minimum**

The new fallback avoids the codec error by clearing `Effective` when the composed ceiling is 1–15. That also removes `max_output_tokens` from the outbound request, so a provider may generate beyond the configured ModelRef ceiling; the added test now explicitly asserts this silent widening. Please reject the Responses/provider configuration or fail the request deterministically when its strict ceiling cannot be represented, rather than continuing without the constraint, and cover the end-to-end dispatch behavior.

src/semantic-router/pkg/looper/stage_context.go
```diff
@@ -237,3 +240,49 @@ func serializedMessagesTokenEstimate(messages []openai.ChatCompletionMessagePara
...
+func stageOverrideMaxOutputTokens(
+	original *openai.ChatCompletionNewParams,
+	stage *openai.ChatCompletionNewParams,
+) *int64 {
+	stageBound := chatParamsMaxOutputTokens(stage)
+	if stageBound == nil {
+		return nil
+	}
+	clientBound := chatParamsMaxOutputTokens(original)
+	if clientBound != nil && *clientBound == *stageBound {
```
Xunzhuo: **Do not infer stage provenance from equal values**

Impact: Numeric equality does not prove that a stage inherited the client limit. If the client sends 256, request params block that field, and Fusion or Workflows independently sets its stage ceiling to 256, this returns nil. The internal hop then omits the stage header, dispatch removes the blocked client value, and the operator-authored stage cap is lost.

Required change: Carry explicit stage authoring/provenance instead of deriving it from value equality, and add a blocked-client regression where an independently authored equal stage bound remains enforced and correctly attributed.

Xunzhuo (commented):
Thanks for closing the previous Dashboard and DSL blockers. The exact-head checks are green, but four composition and contract edges still need to be fixed before this is safe to merge.

src/semantic-router/pkg/extproc/output_token_limit.go
sujal111: Fixed in 1850d5e. Every dispatch now composes from the immutable ingress/Looper client snapshot, so an omitted client stays nil instead of inheriting model A's mutated ceiling. `TestPrepareProviderDispatchRerouteOmitsPriorDispatchAsClient` covers A=256 rerouted to B=1024 and asserts both the 1024 bound and `model_ref` provenance.

src/semantic-router/pkg/configschema/router-config-v0.3.schema.json
sujal111: Fixed in 1850d5e. `ModelRef.max_completion_tokens` now publishes `minimum: 1` from the schema source (`jsonschema` tag plus `setDefinitionPropertyMinimum`), the generated `router-config-v0.3.schema.json` includes it, and `schema_test.go` plus `test_config_schema.py` reject 0 and negative values.

sujal111:
The earlier Core tests / pre-commit failures were `TestSpecifiedModelPreservesAnthropicWireAtExtProcBoundary`: specified-model tests skipped the ingress snapshot, so composition treated the client limit as omitted and Anthropic encoded the default `max_tokens=4096`.

57533f6 snapshots the client ceiling in `routingTestContext` the same way `prepareProtocolRequest` does. Core tests, Full pre-commit, and PR Gate are green on that head. The two remaining review threads (immutable reroute snapshot and schema `minimum: 1`) were already addressed in 1850d5e and are now resolved.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -196,7 +196,11 @@ func (r *OpenAIRouter) prepareProviderRequest(
 	}
 	changed = decisionChanged || changed
 	paramsChanged, err := r.applyDispatchRequestParams(request, ctx)
-	return paramsChanged || changed, err
+	if err != nil {
+		return false, err
+	}
+	limitChanged := r.applyDispatchOutputTokenLimit(request, dispatch, ctx)
```
Xunzhuo: [P2] Compose the output ceiling after automatic rendering

With `default_max_tokens: auto`, this call sets the ModelRef ceiling before `prepareProviderDispatch` calls `prepareAutomaticDispatch`. `resolveAutomaticOutput` then overwrites `MaxOutputTokens` from model capacity and `AutomaticOutputCap`, ignoring the composed ModelRef/ledger ceiling. The finalizer repeats that overwrite immediately before encoding. An isolated check of the actual composition/resolver functions changes a ModelRef limit of 32 to 1024 while Replay still records 32. Apply the strict composition after the final automatic render, preserving rendered capacity as an upper bound, and add a final-wire/Replay regression for this combination.

src/semantic-router/pkg/routerreplay/store/store.go
```diff
@@ -183,54 +186,57 @@ type RequestDemandSnapshot struct {
...
+	SelectedModel                    string                   `json:"selected_model,omitempty"`
+	SessionPolicyApplied             bool                     `json:"session_policy_applied,omitempty"`
+	SessionAction                    string                   `json:"session_action,omitempty"`
+	SessionPhase                     string                   `json:"session_phase,omitempty"`
+	SessionReason                    string                   `json:"session_reason,omitempty"`
+	HardLockReason                   string                   `json:"hard_lock_reason,omitempty"`
+	DecisionReason                   string                   `json:"decision_reason,omitempty"`
+	EffectiveMaxOutputTokens         *int64                   `json:"effective_max_output_tokens,omitempty"`
+	EffectiveMaxOutputTokensSource   string                   `json:"effective_max_output_tokens_source,omitempty"`
+	EffectiveMaxOutputTokensFallback string                   `json:"effective_max_output_tokens_fallback,omitempty"`
```
Xunzhuo: [P2] Publish the new Replay output-limit fields in OpenAPI

The exact-head Generated Contracts job shows that `apiserver.openapi.json` is stale: the RouteDiagnostics and LooperAttempt response schemas omit `effective_max_output_tokens` and its source/fallback fields. The API now returns these diagnostics, but generated clients and the published API reference still describe the old shape. Run `make api-docs-openapi` and commit the generated result.

Xunzhuo (changes requested):
The merge preserves the admission and no-late-reroute behavior, and all 48 selected current-head Go tests pass. A focused check confirms that automatic rendering can overwrite the composed ModelRef ceiling; the public Replay OpenAPI artifact is also stale. Please fix both findings and retain the strict bound requested in the existing Responses-minimum comment. A full provider deployment was not rerun.

src/semantic-router/pkg/extproc/processor_req_body_routing.go
sujal111: Fixed in 2d5d764. `prepareAutomaticDispatchWithOutputLimit` now composes after the final automatic render in both `prepareProviderDispatch` and `finalizeProviderDispatchResponse`. Rendered capacity is a compose source (`automatic`) and can only narrow; a ModelRef of 32 stays 32 on the wire and in Replay while a wider ModelRef is capped by remaining capacity. `TestAutomaticOutputModelRefCeilingWinsOverRenderedCapacity` covers the final-wire/Replay combination.

A composed Responses ceiling below 16 now fails dispatch with `unsupported_responses_max_output_tokens` while retaining the strict bound, instead of omitting it.

src/semantic-router/pkg/routerreplay/store/store.go
sujal111: Fixed in 2d5d764. Regenerated `website/static/openapi/apiserver/apiserver.openapi.json` so `RouteDiagnostics` and `LooperAttempt` publish `effective_max_output_tokens`, `effective_max_output_tokens_source`, and `effective_max_output_tokens_fallback`.

src/semantic-router/pkg/extproc/output_token_limit.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func modelRefMaxCompletionTokens(router *OpenAIRouter, decision *config.Decision, modelName string) *int64 {
+	if decision == nil || modelName == "" {
+		return nil
+	}
+	for _, ref := range decision.ModelRefs {
+		matchesModel := ref.Model == modelName || ref.LoRAName == modelName
+		if router != nil && router.Config != nil {
+			matchesModel = router.Config.ModelNameMatches(ref.Model, modelName) || ref.LoRAName == modelName
```
Xunzhuo, line 173: [P2] Match the selected LoRA or base reference before applying its ceiling

When a decision contains `base` with `lora_name: adapter` and a ceiling of 1024 before a bare `base` reference capped at 32, selecting the bare base matches the adapter reference here. Both references pass config validation, but the final provider request and Replay record use 1024; reversing their order yields 32. Resolve the actual selected reference, distinguishing LoRA identity from its base model, and apply the same rule to the Looper attempt lookup. Add a regression with both references and different ceilings so a sibling adapter cannot widen the selected model’s bound.

Xunzhuo (changes requested):
The automatic-render composition, strict Responses minimum, and Replay OpenAPI fixes are verified. One ceiling lookup still uses a sibling LoRA reference when the bare base model is selected, widening the final request beyond its configured bound. Please fix the inline finding and add the mixed LoRA/base regression. The 46 added Go tests, six related Dashboard assertions, and 13-case Looping Algorithms profile pass in current-head CI; the two changed CLI test files also pass all 35 cases. Quality still reports the testifylint issue at automatic_output_test.go:229. No live provider or model-inference validation was run.

PR #3729, opened 2026-09-11
[Feature] Return routing latency and cost as response headers

abhinav-m22:
Closes #3728
Related #3727, bench side in #3730

## Purpose

Sends back two numbers the router already computes. Owned by #3728 (MoM & Routing).

- `x-vsr-routing-latency-ms`: the duration `recordRoutingLatency` measures, kept on the request context and added in `addFinalDecisionHeaders` with three decimals.
- `x-vsr-cost` and `x-vsr-cost-currency`: kept when `recordResponseCost` prices a response, then set in the buffered body phase with `setResponseBodyHeader`. Not sent for streamed responses, cache hits, or unpriced models. A model priced at 0 sends `0`.
- Header constants, the constants test, and docs rows in `vsr-headers.md` (checked by `TestVSRRoutingHeadersAreDocumented`).

## Test Plan

- `go test ./pkg/headers/ -run 'TestHeaderConstants|TestVSRRoutingHeadersAreDocumented'`
- `go test ./pkg/extproc/ -run 'RoutingLatencyHeader|ResponseCostHeaders'`
- Live: one request per priced model, check the three headers, compare `x-vsr-cost` with `llm_model_cost_total`.

## Test Result

- `go test ./pkg/headers/` passes locally, and `go vet ./pkg/headers/` is clean.
- `gofmt` is clean on all changed Go files.
- The new `pkg/extproc` tests link the Rust bindings, which I don't have built locally, so they run in Core tests on this PR.
- The live end to end check (headers through Envoy against a priced model) is still to do. I'll add its output here.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title begins with exactly one bracketed category, such as `[Feature]`, `[Bug]`, `[Docs]`, `[Test]`, `[Research]`, `[Community]`, or `[CI/Build]`
- [x] The title does not stack prefixes such as `[Router][Docs]`; affected modules belong in labels and the PR body
- [x] The PR links an `accepted` issue with exactly one owner: one `wg/*` label for project work or `owner/maintainers` for repository governance
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and blockers for this change

</details>

Issue #3728, opened 2026-09-11
[Feature] Return routing latency and cost as response headers


abhinav-m22:
### Goal

The router already measures how long it takes to pick a model, and already prices every response from the model's `pricing` config. Neither number reaches the client.

Routing time is logged in whole milliseconds, so keyword routing shows `0` on most requests, and the smallest bucket of `llm_model_routing_latency_seconds` is 5 ms. Cost only goes to `llm_model_cost_total` and the `llm_usage` log. A client can't tie either number to the request it just made.

### Proposed Workgroup

MoM & Routing

### Scope

- `x-vsr-routing-latency-ms`: time spent choosing the model, with three decimals (for example `0.412`). Sent next to `x-vsr-selected-model` on successful responses that didn't come from the cache.
- `x-vsr-cost` and `x-vsr-cost-currency`: the cost `recordResponseCost` already works out, set in the buffered response body phase the same way `x-vsr-response-warnings` is. Not sent for streamed responses, cache hits, or models without pricing.
- Unit tests, and rows in `website/docs/troubleshooting/vsr-headers.md`.
- The bench reporting in #3727 reads these headers.

### Non-goals

- Changing how cost is calculated
- Changing the histogram buckets
- Cost on streamed responses

### Completion signal

A non-streaming request to a priced model returns all three headers, and `x-vsr-cost` matches what `llm_model_cost_total` recorded for that request.

PR #3730, opened 2026-09-11
[Feature] Report routing split, cost, and routing time in bench

### LINKED ISSUE #3727: [Feature] Report routing split, cost, and routing time consistently in bench
### Goal

The two live benchmarks in `bench/` report different things. The agentic benchmark (`agent_task_live_benchmark.py`) records which model answered each request. The reasoning benchmark (`reasoning/router_reason_bench_multi_dataset.py`) does not, and neither one reports cost or how long the router took to decide.

When I ran `arc-challenge` through a two-model router, four runs showed the same accuracy and nothing in the output said which model answered. I had to read the router log to find out every question went to one model.

I'd like both benchmarks to report the same routing and cost fields, using data the router already produces.

### Proposed Workgroup

Evaluation & Quality

### Scope

- Reasoning benchmark: read response headers and record, per question, the selected model, `finish_reason`, reasoning tokens, routing time and cost. A missing header leaves the field blank and never fails the run.
- Both benchmarks: add to `summary.json` the routing split (`selected_model_counts`, the same shape the agentic benchmark already uses), cut off answers (`finish_reason: length`), reasoning tokens, routing time (p50/p95) and cost per currency, labeled as configured pricing rather than a provider bill.
- Optional `--router-metrics-url`: read `llm_model_cost_total` and `llm_model_routing_latency_seconds` before and after a run, for per run cost and mean routing time on routers that don't send per request headers.
- Per request routing time and cost come from the response headers proposed in #3728.

### Non-goals

- New datasets or scoring
- Changing how the router prices responses
- Publishing results
- The different default ports of the two benchmarks

### Completion signal

A routed run of either benchmark shows in `summary.json` how many requests each model answered, what the run cost, how long routing took, and how many answers were cut off. Unit tests cover the new fields.

abhinav-m22:
Closes #3727
Related #3728, router side in #3729

## Purpose

Makes the two live benchmarks report the same routing and cost fields. Owned by #3727 (Evaluation & Quality).

- `reasoning/router_reason_bench_multi_dataset.py` reads response headers and records, per question: selected model, `finish_reason`, reasoning tokens, cached tokens, routing time and cost. `call_model` keeps its old return shape because `reasoning_mode_eval.py` uses it.
- `agent_task_live_benchmark.py` records the same fields per turn.
- Both summaries add `selected_model_counts`, cut off answers, reasoning tokens, `routing_latency_ms` (p50/p95) and `cost` per currency, labeled as configured pricing.
- New optional `--router-metrics-url` reads the router's cost and routing time counters before and after a run. It works on routers without the headers from #3729.
- Helpers for the reasoning benchmark live in `reasoning/router_accounting.py` (standard library only).

## Test Plan

- `python -m pytest bench/test_agent_task_live_benchmark.py bench/reasoning/test_router_accounting.py -q`
- Live: `arc-challenge` through a three tier router config, one run per model name plus `MoM`, each with `--router-metrics-url`.

## Test Result

- `python -m pytest bench/test_agent_task_live_benchmark.py bench/reasoning/test_router_accounting.py -q`: 15 passed (Python 3.13, pytest 9.1). CI doesn't run these files, so this is the local result.
- The live run through a router is still to do. I'll add its routing split, cost and routing time here.

---
<details>

</details>


Xunzhuo (commented):
Thanks for adding the shared accounting summaries. One per-currency aggregation bug needs correction before this is ready.

bench/reasoning/router_accounting.py
```diff
@@ -0,0 +1,182 @@
...
+        if cost is None:
+            continue
+        key = currency or "unknown"
+        totals[key] = totals.get(key, 0.0) + cost
+        priced += 1
+    return {
+        "basis": COST_BASIS,
+        "total": {key: round(value, 8) for key, value in sorted(totals.items())},
+        "mean_per_priced_request": {
+            key: round(value / priced, 8) for key, value in sorted(totals.items())
```
Xunzhuo, line 129: **Count priced requests per currency**

Impact: `priced` combines requests across every currency, then each currency total is divided by that global count. A valid mixed-currency run therefore understates every per-currency mean; the duplicate implementation in `agent_task_live_benchmark.py` has the same problem.

Required change: Track a priced-request count per currency in both summaries and add a mixed-currency regression test.

abhinav-m22: @Xunzhuo thanks for catching this! Both `cost_summary` versions now count priced requests per currency, so each average is divided by its own count. I added a mixed-currency test to each one, and cleared the ruff errors that were failing CI. Fixed now

AayushSaini101 (approved):

wilsonwu (approved):
LGTM

PR #3733, opened 2026-09-11
[Bug] Partition semantic cache by compatibility fingerprint

### LINKED ISSUE #3423: [Bug] Semantic cache ignores the compatibility fingerprint
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

- The semantic response cache matches on the last user message only. System prompt, conversation history, tools, and response format are not part of the semantic key.
- A response cached under one system prompt is served to a later request that asks the same question under a different system prompt, or none.
- The router already computes a compatibility fingerprint over those fields and the docs say semantic matching uses it, but the semantic lookup and write paths never apply it.
- Exact mode is unaffected.

### To Reproduce

1. Configure a decision with `response_cache` in `semantic` mode and `scope: global`, or send the same `x-authz-user-id` on both requests.
2. Send a chat completion with a system prompt and user message Q. Let it cache.
3. Send Q with a different system prompt or none. The cache hit returns the first response.

### Expected behavior

- Semantic lookups and writes are partitioned by `CompatibilityFingerprint`.
- Implementation matches the documented contract.

### Impact

- Any deployment using semantic mode with a shared scope.
- Workaround: `mode: exact`.

### Additional context

- Related to advisory GHSA-xgc7-vx3q-4gfg.
- The anonymous-namespace half is already fixed by scope enforcement in `semantic_cache_scope.go`.

guan404ming:
Closes #3423

## Purpose

- Semantic cache lookups and writes ignored `CompatibilityFingerprint`, so one system prompt's response served another.
- Partition semantic entries by the compatibility fingerprint in `pkg/cache`.

## Test Plan

- `go test ./pkg/cache/ -run 'TestLegacyAdapter|TestResponseCacheService'`

## Test Result

- Passed. The new test fails on `main` and passes with the fix.

### DIFF
--- a/src/semantic-router/pkg/cache/contracts.go
+++ b/src/semantic-router/pkg/cache/contracts.go
@@ -64,6 +64,10 @@ type CacheIdentity struct {
 	SemanticQuery            string         `json:"-"`
 }
 
+func (i CacheIdentity) SemanticPartitionKey() string {
+	return joinPartitionParts(i.Partition.Key(), i.CompatibilityFingerprint)
+}
+
 type TTLPolicy struct {
 	UseDefault bool
 	NoStore    bool
--- a/src/semantic-router/pkg/cache/contracts.go
+++ b/src/semantic-router/pkg/cache/contracts.go
@@ -65,7 +65,7 @@ type CacheIdentity struct {
 }
 
 func (i CacheIdentity) SemanticPartitionKey() string {
-	return joinPartitionParts(i.Partition.Key(), i.CompatibilityFingerprint)
+	return CombineFingerprints("semantic", i.Partition.Key(), i.CompatibilityFingerprint)
 }
 
 type TTLPolicy struct {
--- a/src/semantic-router/pkg/cache/legacy_adapter.go
+++ b/src/semantic-router/pkg/cache/legacy_adapter.go
@@ -95,7 +95,7 @@ func (a *LegacyBackendAdapter) LookupSemantic(
 	}
 	result, err := a.backend.LookupSimilarWithThreshold(
 		ctx,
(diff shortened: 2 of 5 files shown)


src/semantic-router/pkg/cache/contracts.go
```diff
@@ -64,6 +64,10 @@ type CacheIdentity struct {
 	SemanticQuery            string         `json:"-"`
 }
 
+func (i CacheIdentity) SemanticPartitionKey() string {
+	return joinPartitionParts(i.Partition.Key(), i.CompatibilityFingerprint)
```
Xunzhuo: **Keep the semantic partition key within backend limits**

Impact: this appends the 64-character compatibility fingerprint to an already variable-length partition and passes the result as Milvus's `model`, whose schema is capped at 256 characters. A valid long recipe, decision, or model namespace can therefore make every semantic store fail instead of caching.

Required change: encode the complete partition and compatibility fingerprint into a bounded, domain-separated key (or migrate the backend schema), and add a regression at the Milvus field-width boundary.

Xunzhuo (commented):
Thanks for isolating semantic entries by compatibility. One persistent-backend boundary still needs attention.

src/semantic-router/pkg/cache/contracts.go
guan404ming: Thanks, good catch.

- Fixed in a08a98238. The key is now a domain-tagged SHA-256, always 64 hex characters, same as the exact path.
- Added a width regression test with 200-character partition parts.

drivebyer (approved):
LGTM

wilsonwu (approved):
LGTM

guan404ming:
Thanks!

PR #3740, opened 2026-09-11
[Feature] Add per-backend circuit breaker for remote classifiers

### LINKED ISSUE #3726: [Feature] Add per-backend circuit breaker for remote classifiers
### Goal

Remote classifier backends (`http_classify` and other connector-backed classifier calls) should stop issuing live RPCs for a bounded cooldown after consecutive timeouts or unavailability, instead of waiting on a dead endpoint on every request.

The breaker only decides whether to place the call. A skipped call must still surface as the existing unresolved/timeout error so decision-tree `Unknown`, `rules.on_unknown`, and prompt-guard `on_error` keep their current meaning.

This is the cooldown follow-up called out on #3106 and in #2759, not a replacement for those policies.

### Proposed Workgroup

### Scope

- Per remote-classifier backend instance in a process: consecutive-failure threshold, open interval, and a single half-open probe.
- On open, skip the RPC and return the same typed timeout/unavailable error the connector already returns.
- Default off, or on only for remote HTTP/MCP-style backends. In-process local models (mmbert / Candle) stay unchanged.
- Jailbreak/PII skipped calls remain "not verified", never "checked and safe". Fail-closed (`on_error: block` / `on_unknown: match`) must still block; fail-open remains an explicit operator choice.
- Metrics/logs for open, half-open, and skipped calls.
- Tests for closed → open → half-open → closed, and for safety vs routing policy still applying after a skip.

### Non-goals

- Changing decision-tree Unknown / `on_unknown` / prompt-guard `on_error` semantics (#3106).
- Cluster-wide breaker state across ExtProc replicas.
- Envoy/LLM-backend circuit breaking or model fallback (#2294, #35).
- Semantic-cache NLI breaker (#3176).
- Retry/hedging policy beyond the existing connector retry budget (#3128 already shipped timeouts and bounded retry).

### Completion signal

- A hung remote classifier trips open after N consecutive failures and later requests skip the RPC for the cooldown window.
- Those requests still record `classifier_evaluation_timeout` (or the existing unavailable code) and follow the configured unknown/on_error policy.
- A jailbreak rule with `on_error: block` still blocks during the open window.
- Local classifiers never enter this path.
- Unit coverage plus docs for the opt-in config.

Related: #3106, #2759, #3128, #2760

lesbass:
## Summary

Add a per-instance circuit breaker for remote classifier backends (HTTP classify/score and MCP). When enabled via the `circuit_breaker` config, consecutive failures trip the breaker open, which causes subsequent requests to skip the remote RPC and return an `ErrCircuitBreakerOpen` error for the cooldown interval. After the interval expires, a single half-open probe tests if the backend has recovered.

The breaker only decides whether to place the call. A skipped call surfaces an `ErrCircuitBreakerOpen` that preserves existing fail-closed semantics: decision-tree `Unknown`, `rules.on_unknown`, and prompt-guard `on_error` still apply unchanged.

The feature is opt-in and defaults to disabled. In-process local models (mmbert, Candle, ModernBERT) are never wrapped.

### Config
- Added `RemoteClassifierCircuitBreakerConfig` to `RemoteClassifierBackend` with fields: `enabled`, `consecutive_failures`, `open_interval_ms`, `half_open_max_requests`

### Circuit breaker state machine (`circuit_breaker.go`)
- Three-state machine: closed → open → half-open
- Thread-safe with `sync.Mutex`
- `ErrCircuitBreakerOpen` typed error for skipped calls

### Wiring
- Category remote backend: wraps inner `SequenceClassifierBackend`
- Jailbreak remote backend: wraps in `buildJailbreakDependencies`
- Complexity (score and label distribution): wraps `ScoringBackend` and/or `SequenceClassifierBackend`

### Metrics and logging
- `llm_classifier_circuit_breaker_skips_total{backend}` counter
- `llm_classifier_circuit_breaker_state{backend}` gauge
- Component events for skip, failure, open, half-open, close transitions

### Config tests
```
go test ./pkg/config/... -run TestRemoteClassifierBackend -v
# all pass
```

### Reference config surface test
```
go test ./pkg/config/... -run TestReferenceConfigCoversCanonicalPublicSurface -v
# PASS
```

### Full config test suite
```
go test ./pkg/config/...
# PASS (1.649s)
```

### Circuit breaker unit tests (test file included, cannot execute end-to-end due to pre-existing valkey-glide build issue in transitive dependencies)
- TestCircuitBreakerClosedToOpen
- TestCircuitBreakerSuccessResets
- TestCircuitBreakerDisabled
- TestCircuitBreakerHalfOpenToClosed
- TestCircuitBreakerHalfOpenFails
- TestCircuitBreakerNilConfig
- TestCircuitBreakerNilConfigScoringBackend
- TestCircuitBreakerConcurrentSafety

## Non-goals (not in scope)
- Changing decision-tree `Unknown`, `on_unknown`, or prompt-guard `on_error` semantics
- Cluster-wide breaker state across ExtProc replicas
- Envoy/LLM-backend circuit breaking or model fallback
- Semantic-cache NLI breaker
- Retry/hedging beyond existing connector retry budget

Fixes #3726

Signed-off-by: Stefano Maffeis <lesbass@gmail.com>


src/semantic-router/pkg/classification/circuit_breaker.go
```diff
@@ -0,0 +1,204 @@
...
+		recordCircuitBreakerSkip(b.name)
+		return SequenceClassificationResult{}, &ErrCircuitBreakerOpen{
+			Name:       b.name,
+			RetryAfter: b.cb.retryAfter().Format(time.RFC3339),
+		}
+	}
+
+	result, err := b.inner.Classify(ctx, text)
+	if err != nil {
+		b.cb.recordFailure()
```
FAUST-BENCHOU: One concern here is that every backend error increments the breaker, including caller cancellation. Enough canceled requests can therefore open the breaker for an otherwise healthy backend.

Could we only count retry-exhausted timeout/unavailable failures toward the threshold?

src/semantic-router/pkg/config/classifier_backend.go
```diff
@@ -24,17 +24,71 @@ const (
...
+	Enabled bool `yaml:"enabled" json:"enabled"`
+	// ConsecutiveFailures is the number of sequential failures that trip
+	// the breaker open. Defaults to 5.
+	ConsecutiveFailures *int `yaml:"consecutive_failures,omitempty" json:"consecutive_failures,omitempty"`
+	// OpenIntervalMs is how long the breaker stays open before transitioning
+	// to half-open. Defaults to 30000 (30 seconds).
+	OpenIntervalMs *int `yaml:"open_interval_ms,omitempty" json:"open_interval_ms,omitempty"`
+	// HalfOpenMaxRequests is the number of probe requests allowed in the
+	// half-open state. Defaults to 1.
+	HalfOpenMaxRequests *int `yaml:"half_open_max_requests,omitempty" json:"half_open_max_requests,omitempty"`
```
FAUST-BENCHOU, line 49: `half_open_max_requests: 0` is currently accepted, but after the cooldown the breaker enters half-open and can never admit a probe, leaving it stuck there indefinitely.

Could we validate the circuit-breaker thresholds/intervals as positive values when enabled?

FAUST-BENCHOU (commented):
Thanks for adding the per-backend circuit breaker. The current head does not compile in the classifier backend wiring (`cbWrappedBackend.backend` on a `CategoryInference`, and assigning a `SequenceClassifierBackend` back to `*HTTPClassifierInference`). Please get the exact-head CI green first so we can continue the review.

lesbass:
Thanks for the review, @FAUST-BENCHOU. I've pushed a fix resolving both compile issues:

1. **Category backend**: Instead of accessing the unexported `categoryHTTPBackend.backend` field, the wrapper now wraps the `httpClassifierInference` before constructing the `categoryHTTPBackend`.

2. **Complexity label backend**: The `newHTTPClassifierInference` result is now typed as `SequenceClassifierBackend` (interface) rather than `*HTTPClassifierInference`, so returning a wrapped backend compiles.

The branch has been updated — CI should re-run automatically.

FAUST-BENCHOU:
> Thanks for the review, @FAUST-BENCHOU. I've pushed a fix resolving both compile issues:
> 
> 1. **Category backend**: Instead of accessing the unexported `categoryHTTPBackend.backend` field, the wrapper now wraps the `httpClassifierInference` before constructing the `categoryHTTPBackend`.
> 2. **Complexity label backend**: The `newHTTPClassifierInference` result is now typed as `SequenceClassifierBackend` (interface) rather than `*HTTPClassifierInference`, so returning a wrapped backend compiles.
> 
> The branch has been updated — CI should re-run automatically.

no also need to solve inline comments

lesbass:
Thank you for the inline review, @FAUST-BENCHOU. I've pushed fixes for both issues:

**1. Error classification — only retry-exhausted failures count**

Added `isUnavailableError()` to the circuit breaker. Only `connector.KindTransport` errors and `connector.KindStatus` errors with `Retryable: true` (5xx, 408, 429) increment the failure count. Context cancellation and other non-retryable errors pass through untouched. I've also added a `TestCircuitBreakerCancellationNotCounted` test to verify this.

**2. Validation — reject zero thresholds**

Added `validateCircuitBreaker()` to `RemoteClassifierBackend.Validate()`. When `circuit_breaker.enabled: true`, `consecutive_failures`, `open_interval_ms`, and `half_open_max_requests` must be positive if explicitly set.

The branch is updated with commit 8f58d42 — CI should re-run.

src/semantic-router/pkg/classification/circuit_breaker.go
FAUST-BENCHOU: The previous change does not fully resolve this yet. In the actual connector path, a caller cancellation is wrapped as a `connector.Error` with `KindTransport` and `Retryable=false`, while `isUnavailableError()` currently treats every transport error as a breaker failure.

This means canceled requests can still increment the failure count and eventually open the breaker. The check should exclude canceled or caller-expired transport errors before counting them.

src/semantic-router/pkg/config/classifier_backend.go
FAUST-BENCHOU: The validation added for `RemoteClassifierBackend` fixes this path, but the same issue is still reachable through `prompt_guard.circuit_breaker`. `PromptGuardConfig` owns a separate circuit-breaker config and its validation path does not apply the positive-value checks.

As a result, `prompt_guard.circuit_breaker.half_open_max_requests: 0` can still reach runtime and leave the breaker stuck in half-open.

src/semantic-router/pkg/classification/circuit_breaker.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// recordSuccess resets failures and transitions half-open→closed.
+func (cb *circuitBreaker) recordSuccess() {
+	cb.mu.Lock()
+	defer cb.mu.Unlock()
+
+	cb.failureCount = 0
+	if cb.breakerState == circuitBreakerHalfOpen {
+		cb.breakerState = circuitBreakerClosed
```
FAUST-BENCHOU, line 113: There is still a correctness issue when `half_open_max_requests` is greater than 1. The first successful half-open probe immediately transitions the breaker to closed, so a later failure from another already-admitted probe is handled as a normal closed-state failure instead of reopening the breaker.

The breaker needs to remain half-open until the admitted probes are resolved, or `half_open_max_requests` should be constrained to 1.

FAUST-BENCHOU (commented):
There is still one half-open concurrency issue to address.

src/semantic-router/pkg/classification/circuit_breaker.go
```diff
@@ -0,0 +1,228 @@
...
+	if !b.cb.allow() {
+		recordCircuitBreakerSkip(b.name)
+		return SequenceClassificationResult{}, &ErrCircuitBreakerOpen{
+			Name:       b.name,
+			RetryAfter: b.cb.retryAfter().Format(time.RFC3339),
+		}
+	}
+
+	result, err := b.inner.Classify(ctx, text)
+	if err != nil {
```
Xunzhuo, line 162: **Release every half-open probe**

Impact: When an admitted half-open probe returns an error that is not counted by `isUnavailableError`, this path returns without success or failure bookkeeping. With the default single probe, `probeCount` remains at its limit, the breaker stays half-open, and every later request is skipped indefinitely.

Required change: Give every admitted probe a terminal path that releases or resolves its half-open slot while preserving the intended cancellation and non-retryable-error policy, and add a regression for a non-counted half-open error.

src/semantic-router/pkg/classification/circuit_breaker_metrics.go
```diff
@@ -0,0 +1,24 @@
+package classification
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	circuitBreakerStateGauge = promauto.NewGaugeVec(
```
Xunzhuo, line 9: **Publish breaker state transitions**

Impact: This gauge is declared but never updated, and the implementation emits no open, half-open, or closed transition logs. Operators therefore cannot observe the state machine, leaving the accepted issue’s observability requirement incomplete.

Required change: Update the gauge and emit a structured log once per state transition, with coverage for each transition and bounded label cardinality.

Xunzhuo (commented):
Welcome, and thanks for taking this on. Two circuit-breaker lifecycle and observability gaps remain on this head.

lesbass:
Addressed all 5 reviewer feedback items in commit beddcb6bd:

1. **Cancellation not counted** — `isUnavailableError` now explicitly excludes `context.Canceled`/`context.DeadlineExceeded` and requires `Retryable=true` for `KindTransport` errors. Added `TestCircuitBreakerConnectorCancellationNotCounted` and `TestCircuitBreakerNonRetryableTransportNotCounted`.

2. **PromptGuardConfig validation** — Added shared `RemoteClassifierCircuitBreakerConfig.Validate()` and called it from `validatePromptGuardBackendConfig`. Added three new validation tests.

3. **Half-open maxProbes > 1 correctness** — Replaced single `probeCount` with `admittedProbes+unresolvedProbes+halfOpenFailed` accounting. `recordSuccess` in half-open now only transitions to closed when all admitted probes have resolved without failure. `recordFailure` marks `halfOpenFailed` and reopens on resolution. Added `TestHalfOpenMaxProbesGreaterThanOne`.

4. **Half-open probe stuck on non-counted error (regression)** — Added `releaseProbe()` that releases an admitted half-open slot without counting success/failure. Both `Classify` and `Score` call `releaseProbe()` when `isUnavailableError` returns false. Added `TestHalfOpenNonCountedErrorProbeReleases`.

5. **Gauge never updated, no transition logs** — `transitionLocked()` updates `circuitBreakerStateGauge` and emits a structured `circuit_breaker_transition` log on every state change. `recordStateLocked` also publishes on construction. Added `TestCircuitBreakerGaugeTransitions` verifying gauge values through the full state cycle.

lesbass:
Force-pushed f892da3be to add the missing Signed-off-by trailer — DCO now passes.

lesbass:
Fixed the gci import grouping issue that was causing the pre-commit (changed-file) CI check to fail. The `circuit_breaker_test.go` imports now separate third-party (`prometheus/testutil`) and project-prefix (`github.com/vllm-project/semantic-router`) into their respective gci sections per the `.golangci.yml` configuration.

src/semantic-router/pkg/classification/circuit_breaker.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	b.cb.recordSuccess()
+	return score, nil
+}
+
+// isUnavailableError reports whether err is a retry-exhausted connector error
+// indicating the backend is unreachable or returning a retryable status.
+// Caller cancellation and other non-retryable errors are not counted.
+func isUnavailableError(err error) bool {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
Xunzhuo, line 309: **Count classifier-owned timeouts**

This check cannot distinguish a caller deadline from the timeout created inside `HTTPClassifierInference` or `scoringHTTPBackend`. The connector preserves that internal `DeadlineExceeded` through `Unwrap`, so every hung HTTP classifier takes this early `false` path and never advances the breaker. Please distinguish caller-driven cancellation from the backend's own exhausted deadline, count the latter as unavailability, and cover a hung classifier through the real adapter.

src/semantic-router/pkg/classification/circuit_breaker.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// recordSuccess resets failures and transitions half-open→closed when all
+// admitted probes have resolved without failure.
+func (cb *circuitBreaker) recordSuccess() {
+	cb.mu.Lock()
+	defer cb.mu.Unlock()
+
+	cb.failureCount = 0
+	if cb.breakerState != circuitBreakerHalfOpen {
```
Xunzhuo, line 152: **Bind results to their admission generation**

A completion is treated as a half-open probe solely from the breaker's current state. A request admitted while closed can outlive a trip and cooldown; if it completes after a newer request enters half-open, it consumes that real probe's unresolved slot here and can close or reopen the breaker before the probe responds. Please return per-call state/generation metadata from admission, resolve only a probe admitted in that half-open generation, and add a deterministic old-request-spans-open-to-half-open regression.

Xunzhuo (commented):
Thanks for the follow-up—the transition observability gap is fixed. Two new state-machine blockers remain on this head; the earlier neutral half-open probe blocker also still applies and is not repeated here. The branch currently conflicts with `main`, so the integrated result will need another look after rebase.

lesbass:
This PR is superseded by #3838 (rebased on latest main with all 7 reviewer feedback items addressed, including the two state-machine blockers from Xunzhuo's second review).

PR #3746, opened 2026-09-11
[Bug] Preserve explicit false/0 global values across reconcile

### LINKED ISSUE #3734: [Bug] Kubernetes reconcile resets explicit `false` and `0` global values to their defaults
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

On every CRD reconcile, `validateAndUpdate` (`src/semantic-router/pkg/k8s/reconciler.go:257-268`) does three things:

1. Exports the static config with `config.CanonicalStaticConfigFromRouterConfig`.
2. Merges the CRDs into it and marshals the result with yaml.v3.
3. Parses the YAML again and uses the result as the live config.

Several global fields are tagged `omitempty` but default to a non-zero value. An explicit `false` or `0` is dropped on export, and parsing brings the default back.

| Field | Tag | Static file | After export and reparse |
| --- | --- | --- | --- |
| `global.stores.response_cache.ttl_seconds` | `runtime_config.go:186` | `0` | `3600` |
| `global.services.router_replay.ttl_seconds` | `runtime_config.go:348` | `0` | `2592000` |
| `global.model_catalog.modules.hallucination_mitigation.detector.enable_nli_filtering` | `model_config_types.go:316` | `false` | `true` |
| `global.services.observability.tracing.exporter.insecure` | `observability_config.go:54` | `false` | `true` |
| `global.services.observability.tracing.sampling.rate` | `observability_config.go:59` | `0` | `1` |

#2735 reported the same problem for `router_replay.enabled`. #2736 fixed it for that one field by dropping `omitempty`.

### To Reproduce

In `pkg/config`, parse a static config that sets one of the values above, for example:

```yaml
global:
  stores:
    response_cache:
      ttl_seconds: 0
```

Then run it through the reconciler's steps: `CanonicalStaticConfigFromRouterConfig`, `yamlv3.Marshal`, `ParseYAMLBytes`. `SemanticCache.TTLSeconds` is `0` after the first parse and `3600` after the second. The other rows in the table behave the same way.

### Expected behavior

A reconcile keeps the values from the static file.

### Impact

An operator who sets `response_cache.ttl_seconds: 0` expects no expiration, because the in-memory cache skips TTL cleanup when `TTLSeconds <= 0` (`pkg/cache/inmemory_cache.go:126`). After the first reconcile, cache entries expire after an hour. An operator who turned off the hallucination detector's NLI filtering gets it turned back on.

chiliec:
Closes #3734

## Purpose

On every CRD reconcile, `validateAndUpdate` exports the static config with
`CanonicalStaticConfigFromRouterConfig`, marshals it with yaml.v3, and re-parses
the result. Several global fields are tagged `omitempty` but default to a
non-zero value, so an explicit `false` or `0` is dropped on export and the
default is restored on re-parse.

Fields fixed (all in `pkg/config`):

| Field | Default that leaked back |
| --- | --- |
| `global.stores.response_cache.ttl_seconds` | `3600` |
| `global.services.router_replay.ttl_seconds` | `2592000` |
| `global.model_catalog.modules.hallucination_mitigation.detector.enable_nli_filtering` | `true` |
| `global.services.observability.tracing.exporter.insecure` | `true` |
| `global.services.observability.tracing.sampling.rate` | `1` |

This is the same class of bug #2735 reported and #2736 fixed for
`router_replay.enabled`; the fix is identical — drop `omitempty` so the explicit
zero survives marshalling.

Impact: an operator who sets `response_cache.ttl_seconds: 0` (no expiration,
since the in-memory cache skips TTL cleanup when `TTLSeconds <= 0`) gets a 1h
expiry back after the first reconcile; one who disabled the NLI filter gets it
re-enabled.

Owner: `wg/enterprise-environment` (the issue's accepted Workgroup).

## Test Plan

New `pkg/config/canonical_global_zero_values_test.go` replays the reconciler's
exact steps (`CanonicalStaticConfigFromRouterConfig` → `yaml.v3 Marshal` →
`ParseYAMLBytes`) for each field and asserts the explicit zero survives.

```
go test ./pkg/config/ -run TestReconcilePreserves -v
go test ./pkg/config/
go vet ./pkg/config/
gofmt -l <changed files>
```

## Test Result

RED→GREEN verified. Before the fix (with the tests in place, source reverted)
all five fail, e.g.:

```
canonical_global_zero_values_test.go:29: response_cache.ttl_seconds = 3600, want 0
canonical_global_zero_values_test.go:39: router_replay.ttl_seconds = 2592000, want 0
canonical_global_zero_values_test.go:49: tracing.exporter.insecure = true, want false
canonical_global_zero_values_test.go:59: tracing.sampling.rate = 1, want 0
canonical_global_zero_values_test.go:69: hallucination detector enable_nli_filtering = true, want false
```

After the fix:

```
ok  github.com/vllm-project/semantic-router/src/semantic-router/pkg/config
--- PASS: TestReconcilePreservesResponseCacheTTLZero
--- PASS: TestReconcilePreservesRouterReplayTTLZero
--- PASS: TestReconcilePreservesTracingInsecureFalse
--- PASS: TestReconcilePreservesTracingSamplingRateZero
--- PASS: TestReconcilePreservesNLIFilteringDisabled
```

The full `pkg/config` suite passes, `go vet ./pkg/config/` is clean and the
changed files are gofmt-clean. I couldn't run `pkg/k8s`/`pkg/apiserver` tests
locally — they link the prebuilt Rust CGO bindings
(`libcandle_semantic_router`) which aren't built in my environment; that's a
link-time limitation unrelated to this change, which is confined to struct tags
in `pkg/config`.

AI/LLM disclosure: this change was prepared with AI assistance and reviewed
before submission.

### DIFF
--- a/src/semantic-router/pkg/config/canonical_global_zero_values_test.go
+++ b/src/semantic-router/pkg/config/canonical_global_zero_values_test.go
@@ -1,0 +1,71 @@
+package config
+
+import (
+	"testing"
+
+	yamlv3 "gopkg.in/yaml.v3"
+)
+
+func reconcileRoundTrip(t *testing.T, cfg *RouterConfig) *RouterConfig {
+	t.Helper()
+	canonical := CanonicalStaticConfigFromRouterConfig(cfg)
+	encoded, err := yamlv3.Marshal(canonical)
+	if err != nil {
+		t.Fatalf("marshal canonical config: %v", err)
+	}
+	parsed, err := ParseYAMLBytes(encoded)
+	if err != nil {
+		t.Fatalf("parse canonical config: %v", err)
+	}
+	return parsed
+}
+
+func TestReconcilePreservesResponseCacheTTLZero(t *testing.T) {
+	cfg := &RouterConfig{}
+	cfg.SemanticCache.TTLSeconds = 0
+
+	parsed := reconcileRoundTrip(t, cfg)
+	if parsed.SemanticCache.TTLSeconds != 0 {
+		t.Fatalf("response_cache.ttl_seconds = %d, want 0", parsed.SemanticCache.TTLSeconds)
+	}
+}
+
+func TestReconcilePreservesRouterReplayTTLZero(t *testing.T) {
+	cfg := &RouterConfig{}
+	cfg.RouterReplay.TTLSeconds = 0
+
+	parsed := reconcileRoundTrip(t, cfg)
+	if parsed.RouterReplay.TTLSeconds != 0 {
+		t.Fatalf("router_replay.ttl_seconds = %d, want 0", parsed.RouterReplay.TTLSeconds)
+	}
+}
+
+func TestReconcilePreservesTracingInsecureFalse(t *testing.T) {
+	cfg := &RouterConfig{}
+	cfg.Observability.Tracing.Exporter.Insecure = false
+
+	parsed := reconcileRoundTrip(t, cfg)
+	if parsed.Observability.Tracing.Exporter.Insecure {
+		t.Fatal("tracing.exporter.insecure = true, want false")
+	}
+}
+
+func TestReconcilePreservesTracingSamplingRateZero(t *testing.T) {
+	cfg := &RouterConfig{}
+	cfg.Observability.Tracing.Sampling.Rate = 0
+
+	parsed := reconcileRoundTrip(t, cfg)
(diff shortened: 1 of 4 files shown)


Xunzhuo (approved):
Thanks @chiliec

Issue #3704, opened 2026-09-10
[Feature] Revalidate transformed request demand before backend dispatch


Xunzhuo:
### Goal

Ensure final model admission reflects the actual provider-bound request after all enabled context, memory, tool, and provider transformations.

### Proposed Workgroup

MoM & Routing

### Background

Current routing can estimate context demand and filter model candidates before later request transformations run. A transformation may shrink or grow the prompt, leaving initial admission and final provider demand out of sync. This also makes a pre-transformation estimate incomparable with provider-reported prompt usage.

Parent Epic: #3202  
Related accounting: #3050  
Related capability eligibility: #3116  
Related transformation plan: #3472

### Scope

- Capture bounded token-accounting snapshots for the original request, post-context request, post-tool-policy request, and final provider-bound request.
- Recount the final request using the exact or fallback accounting policy defined by #3050.
- Include reserved output demand in the final context-window check.
- Re-evaluate the selected model immediately before backend dispatch when an enabled transformation changes prompt demand.
- Permit reselection only within the current decision's configured eligible model pool.
- Return a typed no-eligible-model result when the final request cannot fit.
- Avoid repeating transformations or introducing observable side effects during recount or reselection.
- Emit bounded, content-free Replay and metric evidence describing the request stage, counting source, exclusion reason, and final outcome.
- Cover both prompt growth and prompt reduction across maintained protocols.

### Non-goals

- Introduce implicit fallback to another decision.
- Define tokenizer implementations; #3050 owns exact accounting.
- Define new context-compression or tool-selection policies.
- Dispatch a request that remains known to exceed the selected model window.

### Completion signal

- [ ] Final admission uses the same provider-bound request that is dispatched.
- [ ] Prompt demand plus reserved output fits the selected model's declared window.
- [ ] Post-selection growth and reduction have deterministic regression coverage.
- [ ] No ineligible final request reaches a backend.
- [ ] Replay distinguishes original, transformed, and final token demand.

## Existing pre-routing eligibility boundary

Current main already filters candidates whose configured context window is smaller than the ingress request estimate before model selection and returns a local typed error when every configured candidate is ineligible. This feature starts only after a later context, memory, tool, or provider transformation changes the provider-bound request.

It does not track misordered decision priorities, missing or inaccurate model-card windows, or deployments that predate or bypass the existing pre-routing guard. Those cases require configuration or deployment correction rather than a second final-admission mechanism.

Peterren:
[...]

Issue #3116, opened 2026-08-30
[Feature] Enforce typed request capability eligibility in model pools


Xunzhuo:
### Goal

Ensure every model selected from a decision's assigned pool can satisfy the hard capabilities required by its role in the request.

### Proposed Workgroup

MoM & Routing

### Scope

- Define one provider-neutral request-requirements envelope for context demand, tools, image input, structured output, protocol, and streaming needs.
- Normalize corresponding logical-model capability metadata and document migration behavior for missing or unknown metadata.
- Filter the assigned logical-model pool before static, multi-factor, latency-aware, confidence, session-aware, evaluation, Router Learning, Looper, or override selection.
- Evaluate capabilities per composite role: candidate, verifier/advisor, and final synthesis may have different requirements.
- Apply candidate-count and quorum requirements only to eligible models; never restore an excluded model to satisfy a minimum.
- Reject an invalid configured stage at validation/startup when it has no eligible model. Runtime capability changes return a typed readiness or eligibility reason rather than substituting an incompatible model.
- Return a stable no-eligible-candidate error before backend dispatch and emit content-free exclusion reasons and counters.
- Keep Router config, CLI, Dashboard, model cards, recipes, probes, and documentation on the same contract.

##### Proposed Workgroup

MoM & Routing

##### Background

Current main filters models by context-window fit, but tool calling, image input, structured output, protocol behavior, and streaming are not represented by one canonical eligibility contract. A request can reach the correct decision and still allow ranking, session continuity, learning, or a composite stage to choose a known-incompatible model.

Parent Epic: #3202. Related portable artifact contract: #2971. Follow-up to #3101.

### Non-goals

- Endpoint or replica load balancing below logical-model selection.
- Training backend models, adding a quality-ranking algorithm, or changing session switch policy.
- Making unknown capability metadata fail closed without a documented migration policy.
- Spreading provider-specific branches across selection and recipe logic.

### Completion signal

- [ ] One canonical request-to-model contract covers context, tools, images, structured output, protocol, and streaming requirements.
- [ ] Every selection and override path consumes the same role-appropriate eligible set.
- [ ] Mixed pools, unknown metadata, protocol parity, quorum, session-pinned, override, and all-ineligible cases are tested.
- [ ] Known all-ineligible requests dispatch no backend call and return a stable typed error.
- [ ] Composite-stage validation and readiness identify the exact recipe, stage, and unmet capability.
[...]

Issue #3050, opened 2026-08-27
[Feature] Add model-tokenizer-aware context accounting for routing and budgets


Xunzhuo:
### Goal

Use the selected model's tokenizer when exact token counts are required for context routing, compression, admission, and budget enforcement.

### Proposed Workgroup

Agentic & Context

### Scope

- Define a tokenizer registry keyed by model or model family.
- Count every prompt-bearing component supported by the protocol, including messages, tools, tool results, structured content, and output reserve.
- Define deterministic behavior before final model selection, such as counting against candidate tokenizers or using a documented safe bound.
- Keep tokenizer loading and counting off remote hot paths through bounded local caching.
- Preserve the existing conservative estimator when a tokenizer is unavailable.
- Expose whether a decision used an exact tokenizer, a calibrated estimate, or a fallback estimate in replay and metrics.

##### Current behavior

The current runtime uses provider-neutral token estimates and online calibration. This is a useful fallback, but it is not equivalent to counting with the tokenizer of the model that will receive the request. Tokenizer differences can move a request across a context band or make a budget decision inaccurate.

##### Ownership

**Workgroup:** Agentic & Context  
**Parent:** #2984

### Non-goals

Not separately specified in the original proposal.

### Completion signal

- Exact and estimated token sources are explicit and testable.
- OpenAI-compatible and Anthropic requests receive equivalent accounting.
- Context classification and budget enforcement use the selected policy consistently.
- Tokenizer lookup failure has a safe, observable fallback.
- Tests compare router counts with provider-reported usage across representative model families and request shapes.

## Stage-aware accounting boundary

Exact accounting must identify the request stage it describes. The counting contract should cover the original request, the post-context-transformation request, and the final provider-bound request; final admission uses the final prompt plus reserved output demand. Provider-reported prompt usage is compared only with the corresponding final prompt count so transformation deltas are not learned or reported as tokenizer error. Replay and metrics record both the request stage and whether the result came from an exact tokenizer, calibration, or fallback.

#3704 owns orchestration of final recount, in-pool reselection, and the pre-dispatch admission check. This issue owns the tokenizer registry and stage-aware counting contract consumed there. Representative accuracy coverage includes tool-heavy, structured, multilingual, and transformation-heavy requests.

rahulbansal3005:
Hi @Xunzhuo the problem looks interesting, would like to take it up.

rahulbansal3005:
[...]

PR #3748, opened 2026-09-11
[Router] Capture stage-aware request demand snapshots

Peterren:
Related #3704

Accounting dependency: #3050  
Eligibility dependency: #3116

## Purpose

Establish the observe-only evidence contract needed before final request admission can safely change routing.

This first slice records bounded, content-free request-demand snapshots at four stable lifecycle stages:

- `original`
- `post_context`
- `post_tool_policy`
- `provider_bound`

Each snapshot carries prompt demand, explicit output reserve when available, total-demand validity, accounting source, semantic request generation, and model identity. The current implementation deliberately reports `counting_source: fallback`; #3050 owns tokenizer-aware exact/calibrated accounting.

`provider_bound` is explicitly marked `representation: semantic`: it describes the final neutral request immediately before provider encoding. This avoids claiming wire-token exactness before #3050 provides that contract.

The observations are exposed through a fixed-label Prometheus histogram. Snapshots available when Replay starts (`original`, `post_context`, and `post_tool_policy` on the normal dispatch path) are also stored in route diagnostics. This slice deliberately does not move the existing Replay start point: `provider_bound` remains request-local/metric evidence until a later bounded Replay-update seam can persist it without changing Replay duration or failure coverage. Snapshot storage is bounded to one value per known stage and reuses an unchanged request generation rather than rescanning identical semantic content.

## Safety boundary

This PR is observe-only:

- no model reselection;
- no admission rejection;
- no decision or plugin re-execution;
- no cross-decision fallback;
- no Replay lifecycle/timestamp change.

When output reserve is unavailable, `total_demand_known` is false and the total-demand metric is not emitted. A later slice will consume #3050/#3116 and add pure final-admission evaluation after this evidence is validated.

## Test Plan

```bash
make check CHANGED_FILES="src/semantic-router/pkg/extproc/processor_req_body.go src/semantic-router/pkg/extproc/processor_req_body_prepare.go src/semantic-router/pkg/extproc/processor_req_body_routing.go src/semantic-router/pkg/extproc/recorder_route_diagnostics.go src/semantic-router/pkg/extproc/request_context.go src/semantic-router/pkg/extproc/request_demand.go src/semantic-router/pkg/extproc/request_demand_test.go src/semantic-router/pkg/observability/metrics/request_demand_metrics.go src/semantic-router/pkg/observability/metrics/request_demand_metrics_test.go src/semantic-router/pkg/routerreplay/recorder.go src/semantic-router/pkg/routerreplay/store/request_demand_snapshot_test.go src/semantic-router/pkg/routerreplay/store/store.go"

cd src/semantic-router
go test ./pkg/extproc -count=1
go test ./pkg/routerreplay/... ./pkg/observability/metrics -count=1
go vet ./pkg/extproc ./pkg/routerreplay/... ./pkg/observability/metrics
```

## Test Result

- `make check`: passed, including `make test-semantic-router` and `make config-schema-check`
- complete `pkg/extproc`: passed
- Router Replay and request-demand metrics packages: passed
- affected-package `go vet`: passed

### DIFF
--- a/src/semantic-router/pkg/extproc/processor_req_body.go
+++ b/src/semantic-router/pkg/extproc/processor_req_body.go
@@ -71,6 +71,7 @@ func (r *OpenAIRouter) handleRequestBody(
 	if err != nil {
 		return nil, err
 	}
+	captureRequestDemand(ctx, requestDemandStagePostContext, request, decisionState.selectedModel)
 	return r.handleModelRoutingWithPersonalizedCache(
 		request,
 		originalModel,
@@ -111,6 +112,7 @@ func (r *OpenAIRouter) handleModelRouting(request *llmprotocol.Request, original
 			"decision":   decisionName,
 		})
 	}
+	captureRequestDemand(ctx, requestDemandStagePostToolPolicy, request, selectedModel)
 	isEntrypoint := ctx.Routing.SelectedRecipe() != nil
 	executesLooper := r.routeExecutesLooper(ctx)
 	if !isEntrypoint && !executesLooper && !r.usesExternalGatewayDispatch(originalModel) {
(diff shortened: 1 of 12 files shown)


Xunzhuo:
as a follow up i think we should make it visible on insights at dashboard

Xunzhuo (approved):

Peterren:
> as a follow up i think we should make it visible on insights at dashboard

will add, right now it only serves as a shadow snapshot contract, need to generate demand traffic to create data points, will do traffic gen and audit dashboard

PR #3750, opened 2026-09-11
[Bug] Rank external RAG results across query windows

### LINKED ISSUE #3738: [Bug] External-API RAG keeps only the first query window
### Proposed Workgroup

### Describe the bug

PR #3650 fixed #3385 by splitting a long RAG query into up to 8 overlapping embedding windows, so that a question buried behind a long preamble still reaches the store. That works for the milvus and qdrant backends. It does not work for `backend: external_api` with `request_format: pinecone` or `weaviate`.

`retrieveFromExternalAPI` (`src/semantic-router/pkg/extproc/req_filter_rag_external.go`) issues one store request per window, then merges the per-window document lists in **first-seen insertion order** and truncates to `top_k` afterwards:

```go
for _, document := range windowDocuments {
	if _, duplicate := seen[document]; duplicate {
		continue
	}
	seen[document] = struct{}{}
	documents = append(documents, document)
}
...
if topK > 0 && len(documents) > topK {
	documents = documents[:topK]
}
```

No score participates in that merge. Because `buildPineconeRequest` and `buildWeaviateRequest` each ask the store for the full `top_k`, the first window fills the list and every later window is discarded.

The score is not merely unused, it is structurally unavailable: `extractPineconeContext` reads only `metadata.content` / `metadata.text` and never the Pinecone `score` field, and `extractWeaviateContext` never reads `_additional.distance` even though `buildWeaviateRequest` explicitly requests it in the GraphQL query.

The milvus and qdrant paths do the right thing: they accumulate into `ragHits`, keep each document's best score across windows, and sort descending before truncating (`req_filter_rag_milvus.go`, `req_filter_rag_qdrant.go`, `req_filter_rag_windowing.go`).

### To Reproduce

Table test driving `retrieveFromExternalAPI` against an `httptest` server impersonating Pinecone. Each successive request returns a different, full-`top_k`, strictly better-scored document set (window *n* returns `W<n>-DOC0..4`, with scores rising in *n*). A real embedding model is loaded so the windowing actually runs.

```
HTTP requests issued to store: 8
topK asked of the store per request: [5 5 5 5 5 5 5 5]
returned documents (5): [W0-DOC0 W0-DOC1 W0-DOC2 W0-DOC3 W0-DOC4]
documents originating from window >= 1: 0
```

Query length and window accounting on the same input:

```
query bytes=27916  raw windows=17  embeddings(sampled)=8
```

The same eight per-window sets fed through `ragHits`, the helper milvus and qdrant use, return the correct documents:

```
ragHits.top(5) = [W7-DOC0 W7-DOC1 W7-DOC2 W7-DOC3 W7-DOC4] scores=[0.705 0.704 0.703 0.702 0.701]
documents originating from window >= 1: 5
```

Configuration needed to reach the path: a decision carrying `plugins: - type: rag / enabled: true` with `backend: external_api`, `request_format: pinecone` (or `weaviate`), and a single user message longer than one embedding window.

### Expected behavior

A document keeps the best score it reached in any window, and the returned set is the top `top_k` documents by that score. This is the invariant #3650 states and the one milvus and qdrant implement.

### Impact

Deterministic rather than intermittent for the affected configuration. For every query past the window threshold the request pays 8 embeddings and 8 **sequential** store round trips (`totalLatency` accumulates serially across the loop) and returns the first window's documents, which is the pre-#3650 result. On a hosted Pinecone or Weaviate endpoint at 50-100 ms per call that adds roughly half a second to the critical path and multiplies store QPS and per-query billing by eight, with no change in what is retrieved.

Saturation is structural rather than probabilistic: `buildPineconeRequest` sends an empty `filter: {}` with no score threshold and `buildWeaviateRequest` sends a bare `limit:` with no `certainty` or `distance` bound, so any index holding at least `top_k` vectors returns a full page for window 0. `similarity_threshold` is read only for the `custom` template and is never applied to these two formats.

The partial case is not a mitigation. When window 0 returns fewer than `top_k`, later windows do fill the remainder, but the ordering is still wrong: a low-scoring window-0 document outranks a high-scoring window-3 document, and nothing can correct it because the extractors discard the scores. Order is load-bearing, since the joined context is truncated head-first at `max_context_length` (default 10,000 runes, `req_filter_rag.go`) before injection, so tail documents can be dropped outright.

**Scope is narrow.** Four independent opt-ins are required, and no shipped artifact satisfies them:

- No shipped artifact configures `backend: external_api` with `pinecone` or `weaviate`. `deploy/helm/semantic-router/values.yaml` has no `plugins:` key, and neither do `values-prod.yaml` or `values-dev.yaml`; the `vllm-sr` CLI scaffold, the 8 `config/recipes/` configs, `e2e/config` and `e2e/profiles` carry no `type: rag` plugin at all (the latter two do carry other plugins, such as `system_prompt` and `fast_response`). The only shipped RAG fragments, `config/fragments/plugin/rag/milvus.yaml` and `qdrant.yaml`, use the two unaffected backends.
- `pinecone` and `weaviate` are undocumented as `request_format` values. `website/docs/tutorials/plugin/rag.md` gives full examples only for milvus and qdrant, and the `external_api` row never lists the legal formats. Outside the implementation file those two strings appear only in one `config/config.yaml` decision and in one field-description comment in `src/vllm-sr/cli/models.py`, which names them as example services rather than as `request_format` values.
- The one shipped `pinecone` config is the `fallback` leg of `primary: milvus / strategy: parallel` in `config/config.yaml`, and `retrieveParallel` discards the fallback whenever the primary succeeded, so it is reachable only when Milvus fails.
- The query threshold is 511 content tokens (the window budget is 512 minus two special tokens), roughly 2,400 characters or 400-500 English words, in a **single user turn**. `ctx.UserContent` is the last user message only, so a long conversation history does not trigger it. Shorter queries take the `len(windows) <= 1` fast path and are byte-identical to pre-#3650 behavior.

Who is affected in practice: an operator using Pinecone or Weaviate as their RAG store, for which `external_api` is the only supported path in this router, whose users paste documents, logs, code, or other long-form text. They are running main or a nightly build. `1e0140c7d` is not an ancestor of `v0.3.0`, `v0.2.0`, or `v0.1.0`, so no tagged release carries this.

Workaround: use `backend: milvus` or `backend: qdrant`, which merge correctly.

### Additional context

**A second, separable change in the same commit.** The new dedupe and `top_k` truncation also apply to `custom` and `elasticsearch`, which are not windowed and previously had neither. Before #3650 the `custom` `results` branch returned a pre-joined string that the caller passed through with no dedupe and no limit; it now returns `[]string` and falls into the shared merge.

Observed against an `httptest` backend returning nine result entries:

```
top_k=<nil>  documents=5   "a---b---c---d---e"
top_k=3      documents=3   "a---b---c"
top_k=20     documents=9   (all nine survive)
dedupe: ["dup","dup","other"] -> documents=2   "dup---other"
```

At `1e0140c7d^` the same server and config returned all nine.

This is largely theoretical today: the only `request_template` in the repo interpolates `{{.TopK}}`, so a backend honoring it cannot exceed `top_k`. The residual risk is an operator whose template omits `{{.TopK}}` or `${top_k}`, since the substitution is undocumented, `buildCustomRequest` is its only reference, and nothing validates the template, and whose backend has its own fixed page size. Unlike the windowing bug this fires on every query regardless of length. The dedupe addition is a strict improvement. `elasticsearch` always interpolates `"size": topK`, so the truncation is a no-op there.

**Suggested direction.** Have `extractPineconeContext` and `extractWeaviateContext` surface the per-match score alongside the content, feed both into the existing `ragHits` helper, and take `hits.top(topK)`, which is the same code path milvus and qdrant already use. Applying the dedupe and truncation only on the windowed branch would also restore the previous behavior for the single-request formats.

An alternative, strictly smaller change is to window only the formats whose extractors can produce a score, and to issue a single request for `pinecone` and `weaviate` until they do. That removes the eightfold cost immediately, at the price of giving up #3650's benefit for those two backends.

Alex-Wengg:
Closes #3738

## Purpose

Pinecone and Weaviate external RAG requests search every embedding window of a long query, but the router previously discarded provider scores and kept documents in first-seen order. A full first-window response therefore crowded every later window out of the final `top_k`.

This change:

- preserves Pinecone similarity scores and Weaviate distances during response extraction;
- normalizes Weaviate's lower-is-better distance into the existing higher-is-better `ragHits` path;
- keeps each document's best ranking value across query windows before applying `top_k`; and
- retains the existing first-seen behavior for custom and Elasticsearch formats.

The change is owned by the Router Models & Inference Runtime Workgroup.

## Test Plan

- Exercise three sequential `httptest` responses for Pinecone and Weaviate, including a duplicate document whose best result arrives in a later window.
- Run the full ExtProc Go package on the repository's Go 1.25 toolchain.
- Run the focused regression and adjacent RAG tests with the race detector.
- Run `go vet` and changed-line `golangci-lint`.

## Test Result

- `go test ./pkg/extproc -count=1`: passed.
- `go test -race ./pkg/extproc -run "TestRetrieveExternalRAGWindowsRanksByBestScore|TestRetrieveFromExternalAPIResponseLimit|TestRagHits" -count=1`: passed.
- `go vet ./pkg/extproc`: passed.
- `golangci-lint run ./pkg/extproc --config ../../tools/linter/go/.golangci.yml --new-from-rev=origin/main`: 0 issues.
- The repository-wide `make check` wrapper could not bootstrap locally because the configured Python package index does not yet provide its pinned `jsonschema==4.26.0`; the owning Go package checks above were run directly.

No live Pinecone or Weaviate service was used; the regression uses their decoded response shapes through a real HTTP test server.

### DIFF
--- a/src/semantic-router/pkg/extproc/req_filter_rag_external.go
+++ b/src/semantic-router/pkg/extproc/req_filter_rag_external.go
@@ -131,8 +131,32 @@ func (r *OpenAIRouter) retrieveFromExternalAPI(traceCtx context.Context, ctx *Re
 		topK = *ragConfig.TopK
 	}
 
-	// A document that several windows retrieve is kept once, at the rank the
-	// first window that found it gave it.
+	documents, totalLatency, err := r.retrieveExternalRAGWindows(traceCtx, apiConfig, requestBodies, topK)
+	if err != nil {
+		return "", err
+	}
+	ctx.RAGRetrievalLatency = totalLatency
+
+	if len(documents) == 0 {
+		return "", fmt.Errorf("no content found in %s response", apiConfig.RequestFormat)
+	}
+
+	logging.Infof("Retrieved %d documents from external API over %d query window(s) (latency: %.3fs, format: %s)",
+		len(documents), len(requestBodies), totalLatency, apiConfig.RequestFormat)
+	return strings.Join(documents, "\n\n---\n\n"), nil
+}
+
+// retrieveExternalRAGWindows sends the query-window requests and combines their
+// results. Vector-search formats are merged by score; single-request formats
+// retain their existing first-seen behavior.
+func (r *OpenAIRouter) retrieveExternalRAGWindows(
+	traceCtx context.Context,
+	apiConfig *config.ExternalAPIRAGConfig,
+	requestBodies [][]byte,
+	topK int,
+) ([]string, float64, error) {
+	rankByScore := apiConfig.RequestFormat == "pinecone" || apiConfig.RequestFormat == "weaviate"
+	var rankedDocuments ragHits
 	var documents []string
 	seen := make(map[string]struct{})
 	var totalLatency float64
@@ -140,11 +164,15 @@ func (r *OpenAIRouter) retrieveFromExternalAPI(traceCtx context.Context, ctx *Re
 		apiResponse, latency, sendErr := r.sendExternalRAGRequest(traceCtx, apiConfig, requestBody)
 		totalLatency += latency
 		if sendErr != nil {
-			return "", sendErr
+			return nil, totalLatency, sendErr
 		}
-		windowDocuments, extractErr := r.extractDocumentsFromResponse(apiResponse, apiConfig.RequestFormat)
+		windowDocuments, windowScores, extractErr := r.extractDocumentsFromResponse(apiResponse, apiConfig.RequestFormat)
 		if extractErr != nil {
-			return "", fmt.Errorf("failed to extract context: %w", extractErr)
+			return nil, totalLatency, fmt.Errorf("failed to extract context: %w", extractErr)
+		}
+		if rankByScore {
+			rankedDocuments.add(windowDocuments, windowScores)
+			continue
 		}
 		for _, document := range windowDocuments {
 			if _, duplicate := seen[document]; duplicate {
@@ -154,18 +182,14 @@ func (r *OpenAIRouter) retrieveFromExternalAPI(traceCtx context.Context, ctx *Re
 			documents = append(documents, document)
 		}
 	}
-	ctx.RAGRetrievalLatency = totalLatency
 
-	if len(documents) == 0 {
-		return "", fmt.Errorf("no content found in %s response", apiConfig.RequestFormat)
+	if rankByScore {
+		documents, _ = rankedDocuments.top(topK)
 	}
 	if topK > 0 && len(documents) > topK {
 		documents = documents[:topK]
 	}
-
-	logging.Infof("Retrieved %d documents from external API over %d query window(s) (latency: %.3fs, format: %s)",
-		len(documents), len(requestBodies), totalLatency, apiConfig.RequestFormat)
-	return strings.Join(documents, "\n\n---\n\n"), nil
+	return documents, totalLatency, nil
 }
 
 // sendExternalRAGRequest posts one request body and decodes the response,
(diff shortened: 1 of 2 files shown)


Xunzhuo (approved):

Xunzhuo:
@Alex-Wengg thank you and welcome

PR #3755, opened 2026-09-12
[Bug] Secure Playground feedback for read users

### LINKED ISSUE #3725: [Bug] Playground feedback should record for all signed-in users, appear on Insights, and not let shared read accounts change routing
### Goal

Signed-in Dashboard users, including the public read-only playground account documented in the README (`love@vllm-sr.ai`), should be able to thumbs-up or thumbs-down a Playground response. The rating must attach to that request's Router Replay record and be visible on the Insights detail page. Ratings from the shared read-only account must not change online model selection.

### Proposed Workgroup

### Current behavior

Playground assistant messages show thumbs-up / thumbs-down controls. Clicking them `POST`s `/api/router/v1/router/outcomes`. That route requires `feedback.submit`, which default role grants give only to `admin` and `write`. A `read` user gets HTTP 403, and the UI falls back to `Feedback request failed`.

The hosted playground uses the shared read-only account from the README. That account can chat and see the buttons, then the click fails.

Insights already lists Router Replay records at `/insights` and `/insights/:recordId` (selected model, decision, lifecycle, usage). The record detail page has no Outcomes section, and the frontend record type does not include `outcomes`. A successful Playground rating only shows `Feedback Sent!` and does not link to Insights. `signals.user_feedback` is a request-time classification signal, not a thumbs rating.

### Why granting `feedback.submit` to `read` is not enough

An accepted outcome can update in-process model experience (per-model `good_fit` / `underpowered` / related counts). That ledger is shared by the router process. It is not a per-login or per-session learner.

If the published read-only password can write that ledger, anyone can bias later model choice. `read` also has `replay.read`. If the API only checks the permission and not ownership, a caller can rate an arbitrary historical replay. Rate-limiting by email is also not a boundary here: every public playground user is the same account.

The current failure is therefore not just a missing hook. The UI treats rating as a chat action, while authorization treats it as a routing-state mutation.

### Scope

1. Who can rate  
   Grant `feedback.submit` to every signed-in Dashboard role. Hide the buttons when the caller does not have that permission.

2. Which record can be rated  
   Accept Playground ratings only for `replay_id`s produced by the current Dashboard login session. Reject expired, in-progress, model-mismatched, and duplicate submissions. Rate-limit by session, not by email.

3. Whether a rating changes routing  
   Persist `read` ratings on the matching replay record only. Do not update model experience and do not change later model selection. `write` / `admin` ratings may continue to update model experience, as they do today. Do not clear or replay the existing experience ledger.

4. Where the rating is shown  
   Add an Outcomes section on the Insights record page (source, target, verdict, time). Surface `outcomes` on the frontend record type. After a successful Playground rating, link to the matching Insights record.

On API failure, show the backend error instead of the generic `Feedback request failed` string.

### Non-goals

- Do not train a separate Router Learning state per session.
- Do not clear or retrain existing model experience.
- Do not add content moderation or "random click" detection.
- Do not change the public demo account itself.
- Do not give `read` config-write or deploy rights.
- Do not add Insights list-page feedback aggregates in this issue. Record detail is enough.

### Completion signal

- The public read-only account can thumbs-up / thumbs-down in Playground without `Feedback request failed`.
- The matching Insights record shows that rating.
- A caller cannot rate a `replay_id` from another session or another request.
- A `read` rating does not increment model-experience counts and does not change online routing.
- A `write` / `admin` rating still updates model experience and is also visible on Insights.
- A `read` user can open Insights record detail and see Outcomes. Request/response bodies stay redacted under the existing read-only policy.

### Impact

| Role | Playground rating | Insights | Model experience / routing |
| --- | --- | --- | --- |
| Shared `read` | Fails today; succeeds after this change | Rating hidden today; visible on the matching record after this change | Unchanged |
| `write` / `admin` | Can update learning today; still can | Rating hidden today; visible after this change | Unchanged |
| Existing experience ledger | — | — | Not cleared |

Alex-Wengg:
Closes #3725

## Purpose

Allow every signed-in user to submit Playground feedback while preventing shared read-role accounts from changing routing behavior.

This change:

- grants read users the `feedback.submit` permission and keeps the feedback controls permission-aware
- binds proxied Playground replay IDs to the active Dashboard session and validates replay completion, ownership, expiry, selected model, duplicates, and rate limits before accepting feedback
- forces read-role feedback into a new `record_only` path, which records the Router Replay outcome without updating online model experience; writer and admin learning behavior is unchanged
- surfaces recorded outcomes in Insights and links successful Playground feedback to its replay record
- updates the API artifacts, security documentation, unit coverage, and the existing feedback Playwright scenario

Owned by `wg/developer-experience-ecosystem`.

## Test Plan

- `make test-semantic-router`
- `go test ./auth -run 'TestAuthenticateRequestAcceptsSessionCookie|TestPlayground|TestAuthenticateRequestRequiresFeedback|TestAllRolesHaveFeedback' -count=1`
- `go test ./router -run 'Test(TrackPlayground|PlaygroundOutcome|MapPlayground)' -count=1`
- `go test -race ./auth ./router -run 'Test(Playground|TrackPlayground|MapPlayground|AuthenticateRequestRequiresFeedback|AllRolesHaveFeedback)' -count=1`
- `go test ./pkg/apiserver ./pkg/extproc ./pkg/routerruntime`
- `go vet ./auth ./router ./pkg/apiserver ./pkg/extproc ./pkg/routerruntime`
- `npm run test:unit`, `npm run type-check`, `npm run lint`, and `npm run build` in `dashboard/frontend`
- `npx eslint e2e/feedback-buttons.spec.ts` in `dashboard/frontend`
- `make config-schema-check`, `make api-docs-check`, `make docs-check-translation-coverage`, and `make bench-router-learning`
- repository pre-commit hooks

## Test Result

All listed Go, frontend, documentation, benchmark, and pre-commit checks passed locally. Frontend unit tests passed (190 files, 878 tests); lint completed with 0 errors and 108 pre-existing warnings.

The updated Playwright scenario was linted but not executed because the local bundled Chromium installation was unavailable. Full CI remains the integration check for that browser flow.

`make check` and `make dashboard-check` could not complete locally because golangci-lint 2.5 cannot decode Go 1.27.1 export data. Direct builds, tests, race tests, vet, and frontend checks for the changed areas passed.

---
<details>

</details>

### DIFF
--- a/dashboard/backend/auth/playground_feedback_store.go
+++ b/dashboard/backend/auth/playground_feedback_store.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"strings"
 	"time"
+
+	"github.com/google/uuid"
 )
 
 const (
@@ -35,15 +37,16 @@ func (s *Service) BindPlaygroundReplay(ctx context.Context, sessionID, replayID,
 		return ErrPlaygroundReplayNotOwned
 	}
 	now := time.Now().Unix()
+	idempotencyKey := "playground-feedback-" + uuid.NewString()
 	result, err := s.store.db.ExecContext(ctx, `
 		INSERT INTO playground_feedback_replays(
-			replay_id, session_id, target_ref, state, created_at, expires_at
+			replay_id, session_id, target_ref, idempotency_key, state, created_at, expires_at
 		)
-		SELECT ?, id, ?, ?, ?, expires_at
+		SELECT ?, id, ?, ?, ?, ?, expires_at
 		FROM auth_sessions
 		WHERE id = ? AND revoked_at IS NULL AND expires_at > ?
 		ON CONFLICT(replay_id) DO NOTHING`,
-		replayID, targetRef, playgroundReplayStateInProgress, now, sessionID, now,
+		replayID, targetRef, idempotencyKey, playgroundReplayStateInProgress, now, sessionID, now,
 	)
 	if err != nil {
 		return fmt.Errorf("bind playground replay: %w", err)
@@ -85,52 +88,54 @@ func (s *Service) ValidatePlaygroundReplay(ctx context.Context, sessionID, repla
 	}
 	return validatePlaygroundReplayRow(
 		s.store.db.QueryRowContext(ctx, `
-			SELECT target_ref, state, expires_at
+			SELECT target_ref, idempotency_key, state, expires_at
 			FROM playground_feedback_replays
 			WHERE replay_id = ? AND session_id = ?`, strings.TrimSpace(replayID), strings.TrimSpace(sessionID)),
 		strings.TrimSpace(targetRef), time.Now().Unix(),
 	)
 }
 
-// ClaimPlaygroundReplay atomically reserves one completed replay for feedback
-// and enforces the rate limit against the login session rather than the user.
+// ClaimPlaygroundReplay atomically reserves one completed replay for feedback,
+// enforces the session rate limit, and returns the replay's stable Router idempotency key.
 func (s *Service) ClaimPlaygroundReplay(
 	ctx context.Context,
 	sessionID, replayID, targetRef string,
 	limit int,
 	window time.Duration,
-) error {
+) (string, error) {
 	if s == nil || s.store == nil {
-		return ErrPlaygroundReplayNotOwned
+		return "", ErrPlaygroundReplayNotOwned
 	}
 	tx, err := s.store.db.BeginTx(ctx, nil)
 	if err != nil {
-		return fmt.Errorf("begin playground feedback claim: %w", err)
+		return "", fmt.Errorf("begin playground feedback claim: %w", err)
 	}
 	defer func() { _ = tx.Rollback() }()
 
 	now := time.Now().Unix()
-	if err := validatePlaygroundReplayRow(
+	idempotencyKey, validationErr := readValidPlaygroundReplayRow(
 		tx.QueryRowContext(ctx, `
-			SELECT target_ref, state, expires_at
+			SELECT target_ref, idempotency_key, state, expires_at
 			FROM playground_feedback_replays
 			WHERE replay_id = ? AND session_id = ?`, strings.TrimSpace(replayID), strings.TrimSpace(sessionID)),
 		strings.TrimSpace(targetRef), now,
-	); err != nil {
-		return err
+	)
+	if validationErr != nil {
+		return "", validationErr
 	}
 
 	if limit > 0 && window > 0 {
 		var attempts int
-		if err := tx.QueryRowContext(ctx, `
+		countErr := tx.QueryRowContext(ctx, `
 			SELECT COUNT(*) FROM playground_feedback_replays
 			WHERE session_id = ? AND claimed_at >= ?`,
 			strings.TrimSpace(sessionID), now-int64(window.Seconds()),
-		).Scan(&attempts); err != nil {
-			return fmt.Errorf("count playground feedback attempts: %w", err)
+		).Scan(&attempts)
+		if countErr != nil {
+			return "", fmt.Errorf("count playground feedback attempts: %w", countErr)
 		}
 		if attempts >= limit {
-			return ErrPlaygroundFeedbackRateLimited
+			return "", ErrPlaygroundFeedbackRateLimited
 		}
 	}
 
@@ -142,19 +147,19 @@ func (s *Service) ClaimPlaygroundReplay(
 		strings.TrimSpace(sessionID), playgroundReplayStateCompleted,
 	)
 	if err != nil {
-		return fmt.Errorf("claim playground replay: %w", err)
+		return "", fmt.Errorf("claim playground replay: %w", err)
 	}
 	rows, err := result.RowsAffected()
 	if err != nil {
-		return fmt.Errorf("read playground replay claim result: %w", err)
+		return "", fmt.Errorf("read playground replay claim result: %w", err)
 	}
 	if rows != 1 {
-		return ErrPlaygroundReplayDuplicate
+		return "", ErrPlaygroundReplayDuplicate
 	}
 	if err := tx.Commit(); err != nil {
-		return fmt.Errorf("commit playground feedback claim: %w", err)
+		return "", fmt.Errorf("commit playground feedback claim: %w", err)
 	}
-	return nil
+	return idempotencyKey, nil
 }
 
 // FinishPlaygroundReplay commits a successful submission or releases a failed
@@ -184,28 +189,36 @@ type playgroundReplayRow interface {
 }
 
 func validatePlaygroundReplayRow(row playgroundReplayRow, targetRef string, now int64) error {
-	var storedTarget, state string
+	_, err := readValidPlaygroundReplayRow(row, targetRef, now)
+	return err
+}
+
+func readValidPlaygroundReplayRow(row playgroundReplayRow, targetRef string, now int64) (string, error) {
+	var storedTarget, idempotencyKey, state string
 	var expiresAt int64
-	if err := row.Scan(&storedTarget, &state, &expiresAt); err != nil {
+	if err := row.Scan(&storedTarget, &idempotencyKey, &state, &expiresAt); err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
-			return ErrPlaygroundReplayNotOwned
+			return "", ErrPlaygroundReplayNotOwned
 		}
-		return fmt.Errorf("read playground replay: %w", err)
+		return "", fmt.Errorf("read playground replay: %w", err)
 	}
 	if expiresAt <= now {
-		return ErrPlaygroundReplayExpired
+		return "", ErrPlaygroundReplayExpired
 	}
 	if storedTarget != targetRef {
-		return ErrPlaygroundReplayModelMismatch
+		return "", ErrPlaygroundReplayModelMismatch
 	}
 	switch state {
 	case playgroundReplayStateCompleted:
-		return nil
+		if idempotencyKey == "" {
+			return "", ErrPlaygroundReplayNotOwned
+		}
+		return idempotencyKey, nil
 	case playgroundReplayStateInProgress:
-		return ErrPlaygroundReplayInProgress
+		return "", ErrPlaygroundReplayInProgress
 	case playgroundReplayStateSubmitting, playgroundReplayStateConsumed:
-		return ErrPlaygroundReplayDuplicate
+		return "", ErrPlaygroundReplayDuplicate
 	default:
-		return ErrPlaygroundReplayNotOwned
+		return "", ErrPlaygroundReplayNotOwned
 	}
 }
--- a/dashboard/backend/auth/playground_feedback_store.go
+++ b/dashboard/backend/auth/playground_feedback_store.go
@@ -1,0 +1,211 @@
+package auth
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+)
+
+const (
+	playgroundReplayStateInProgress = "in_progress"
+	playgroundReplayStateCompleted  = "completed"
+	playgroundReplayStateSubmitting = "submitting"
+	playgroundReplayStateConsumed   = "consumed"
+)
+
+var (
+	ErrPlaygroundReplayNotOwned      = errors.New("replay does not belong to the current login session")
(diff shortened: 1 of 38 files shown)


AayushSaini101:
Thanks for the fixing the issue regarding the feedback @Alex-Wengg  : ) can you please add the screenshots/recording to show the feedback workflow end to end as well as some of the CI is failing can you please check that one also thanks

dashboard/backend/auth/playground_feedback_store.go
```diff
@@ -0,0 +1,211 @@
...
+	return nil
+}
+
+// FinishPlaygroundReplay commits a successful submission or releases a failed
+// upstream attempt for retry. claimed_at remains as rate-limit evidence.
+func (s *Service) FinishPlaygroundReplay(ctx context.Context, sessionID, replayID string, submitted bool) error {
+	if s == nil || s.store == nil {
+		return ErrPlaygroundReplayNotOwned
+	}
+	nextState := playgroundReplayStateCompleted
```
Xunzhuo, line 179: **Preserve idempotency across ambiguous upstream failures**

Impact: Every observed non-2xx reopens this replay, but the UI creates a fresh `Idempotency-Key` for each submission. If the Router commits the outcome and learning update but its response is lost, Dashboard marks the replay retryable; the next click uses a new key and records the same replay a second time, double-biasing writer/admin learning.

Required change: Persist and reuse one stable Router idempotency key per replay across retries, or reconcile an ambiguous upstream result before returning the row to `completed`, with a lost-response regression.

dashboard/backend/auth/playground_feedback_store.go
```diff
@@ -0,0 +1,211 @@
...
+		return ErrPlaygroundReplayNotOwned
+	}
+	tx, err := s.store.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("begin playground feedback claim: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	now := time.Now().Unix()
+	if err := validatePlaygroundReplayRow(
```
Xunzhuo: **Keep the transaction error variable unshadowed**

Impact: Both this declaration and the one at line 125 shadow the `err` introduced by `BeginTx`, so the exact-head Dashboard and full pre-commit checks fail `govet`.

Required change: Assign to the existing variable (or use distinct names) and rerun both required checks.

dashboard/backend/router/playground_feedback.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+	base.Path = strings.TrimRight(base.Path, "/") + "/api/v1/observability/replays/" + url.PathEscape(payload.ReplayID)
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, base.String(), nil)
+	if err != nil {
+		return &playgroundFeedbackError{
+			status: http.StatusInternalServerError, code: "REPLAY_VERIFICATION_FAILED", message: "Could not verify the replay record.",
+		}
+	}
+	if err := routerauth.RewriteAuthorization(request, credentialProvider); err != nil {
```
Xunzhuo: **Avoid shadowing the parsed-URL error**

Impact: This declaration shadows the `err` from `url.Parse`, producing the third changed-line `govet` failure on the exact head.

Required change: Reuse the existing variable or give this error a distinct name, then rerun the Dashboard and full pre-commit checks.

Xunzhuo (commented):
Thanks for the careful ownership and record-only flow. One retry path can still duplicate learning updates, and the exact-head Dashboard and full pre-commit checks fail on three changed-line `govet` errors.

Alex-Wengg:
> Thanks for the fixing the issue regarding the feedback @Alex-Wengg : ) can you please add the screenshots/recording to show the feedback workflow end to end as well as some of the CI is failing can you please check that one also thanks


@AayushSaini101 thanks for the review, ! I’ve added an e2e recording of the testing from login to the request ending in /observability/outcomes & the insight page:

https://streamable.com/rsjwyw

The Dashboard checks and full pre-commit checks now pass. The remaining CI failures based on my understanding might be caused by an `npm ECONNRESET` while downloading a dependency from `cdn.npmmirror.com`:

https://github.com/vllm-project/semantic-router/actions/runs/34717880431/job/103618303249?pr=3755

The PR Gate failure is downstream of that CLI failure. I don’t have permission to rerun upstream
workflows. a rerun might fix the issue?

dashboard/backend/auth/playground_feedback_store.go
Alex-Wengg: The Dashboard now stores one idempotency key per replay and reuses it across  retries. Added a regression test confirming that a lost response and retry apply feedback only once

dashboard/backend/auth/playground_feedback_store.go
Alex-Wengg: replaced if err := ... with names like validationErr := ..., to prevent from creating another
  variable named err

dashboard/backend/router/playground_feedback.go
Alex-Wengg: renamed it to authorizationErr

dashboard/backend/auth/playground_feedback_store.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	switch state {
+	case playgroundReplayStateCompleted:
+		if idempotencyKey == "" {
+			return "", ErrPlaygroundReplayNotOwned
+		}
+		return idempotencyKey, nil
+	case playgroundReplayStateInProgress:
+		return "", ErrPlaygroundReplayInProgress
+	case playgroundReplayStateSubmitting, playgroundReplayStateConsumed:
```
Xunzhuo: **Recover interrupted feedback submissions**

Impact: `submitting` is durable, but it has no lease or startup recovery. If Dashboard stops after this claim is committed and before `FinishPlaygroundReplay`, every retry is rejected as a duplicate until the session expires; after Router's process-local idempotency window or a Router restart, there is no safe way to learn whether replay is still possible.

Required change: Make `submitting` a leased/recoverable claim or reconcile it against the replay outcome before retrying; cover restart between claim and finish and retries beyond Router's in-memory idempotency window.

Xunzhuo (commented):
Thanks—the stable server-owned key and vet fixes address the reported update. Two cross-process/session boundaries still need tightening.

dashboard/backend/router/playground_feedback.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	if principal.Role == auth.RoleRead {
+		payload.RecordOnly = true
+		rawBody, _ = json.Marshal(payload)
+	}
+	r.Body = io.NopCloser(bytes.NewReader(rawBody))
+	r.ContentLength = int64(len(rawBody))
+	r.Header.Set("Idempotency-Key", idempotencyKey)
+	if err := routerauth.RewriteAuthorization(r, credentialProvider); err != nil {
```
Xunzhuo, line 178: **Preserve user and session provenance through the proxy**

Impact: Rewriting every Dashboard request to the shared management credential makes Router persist Playground user ratings as operator actions and places independent sessions in one 60/minute bucket. Insights is misattributed, and unrelated sessions can throttle each other.

Required change: Carry authenticated server-owned user/session provenance to Router (or avoid reapplying the shared-principal policy), then assert persisted source and cross-session saturation behavior.

dashboard/backend/auth/playground_feedback_store.go
Alex-Wengg: my understanding is the Dashboard marks feedback as submitting in its database before forwarding it to the Router. A crash at that point left the row stuck in submitting, so every retry was rejected as a duplicate for the rest of the session. And since the Router tracked idempotency only in memory, after a Router restart there was no way to tell if the feedback had ever been applied

so a TTL is needed , the  submitting claim now carries a 30-second lease once it expires, a retry can re-claim the row and gets back the same idempotency key. The Router persists that key into the durable replay record and checks it before applying any outcome. So crashed submissions are retryable, and duplicates are caught even across Router restarts.

dashboard/backend/router/playground_feedback.go
Alex-Wengg: so the issue is The Dashboard proxied all Playground feedback to the Router under 1 shared management credential, erasing the user's identity in transit. resulting in 2 problems: the Router persisted user ratings as operator actions, corrupting learning attribution and Insights; and all users shared the credential's single 60/minute rate bucket, so 1 busy session could throttle everyone else.

a fix is the Dashboard proxy strips any client-supplied x-vsr-outcome-* headers, then stamps its own: source user and the authenticated session id.  The Router should now accept these headers only from the Dashboard's own credential, then stores the feedback as user feedback and rate-limits by session instead of by credential.

dashboard/backend/router/playground_feedback.go
```diff
@@ -0,0 +1,351 @@
...
+		_ = store.FinishPlaygroundReplay(context.Background(), principal.SessionID, payload.ReplayID, false)
+		writePlaygroundFeedbackError(w, playgroundFeedbackError{
+			status:  http.StatusServiceUnavailable,
+			code:    "ROUTER_CREDENTIAL_UNAVAILABLE",
+			message: "Router management credential is unavailable.",
+		})
+		return
+	}
+	r.Header.Set(headers.VSROutcomeSource, "user")
```
Xunzhuo, line 187: These trusted headers are installed before the standard `ReverseProxy` processes hop-by-hop headers. An HTTP/1 caller can send `Connection: x-vsr-outcome-source, x-vsr-outcome-principal`; the proxy removes every header named by `Connection` after its Director runs, so both server-owned values disappear before the Router sees them. The shared dashboard credential then falls back to operator provenance and one shared rate-limit principal. Please remove/reject client connection-nominated names before installing these values, or install them in a `Rewrite` hook after hop stripping, with an HTTP/1 regression.

dashboard/backend/router/playground_feedback.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	if response.Request.URL.Path != "/v1/chat/completions" {
+		return
+	}
+	principal, ok := auth.AuthFromContext(response.Request)
+	if !ok || strings.TrimSpace(principal.SessionID) == "" {
+		return
+	}
+	replayID := strings.TrimSpace(response.Header.Get(headers.RouterReplayID))
+	model := strings.TrimSpace(response.Header.Get(headers.VSRSelectedModel))
```
Xunzhuo, line 88: This binds a replay to only the `x-vsr-selected-model`, which is the last model for a Ratings Looper response. The Ratings body instead exposes multiple choices without per-choice model IDs and a comma-joined top-level `model`; the frontend assigns that combined string to every choice and submits it as `target_ref`. Consequently every Ratings feedback button fails the stored-model check. Please bind server-attested per-choice models and have each button submit its exact model, or hide feedback for this unsupported response shape; cover buffered and SSE Ratings responses.

Xunzhuo (commented):
Thanks—the durable claim and server-attested provenance changes resolve the previous findings. Two request-boundary cases still break the intended feedback path.

dashboard/backend/router/playground_feedback.go
Alex-Wengg: i see,  added the trusted headers before ReverseProxy processed Connection; the proxy could then remove them. 

The change is now sanitizes Connection first, then adds the headers.

dashboard/backend/router/playground_feedback.go
Alex-Wengg: wasn't aware ofthat Ratings Looper responses contain multiple model answers but only provide a comma-separated top-level model string. The frontend could reuse that string for every answer, causing target_ref to differ from the replay’s stored model and a case where all feedback submissions fail.

it should be fixed it by avoiding ambiguous model assignment and displaying feedback only when an answer has an exact model ID matching the stored selected model.

Xunzhuo (approved):
The main integration preserves session ownership, server-issued idempotency, and record-only feedback for read users. Current-head backend and Router regressions pass, and all 12 focused Chromium checks pass across the original and read-role fixtures. The browser checks use controlled API responses; a full live stack was not rerun. Approved.

Alex-Wengg:
@Mergifyio refresh

Alex-Wengg:
@Xunzhuo thanks for the approval! Threads are resolved and CI is green, but Mergify needs a second approval to queue. do you happen to know a dashboard owner I should ping (wilsonwu or FAUST-BENCHOU?), or queue it directly if you're okay with it?

wilsonwu (approved):
LGTM

PR #3759, opened 2026-09-12
[Bug] Validate global config before Kubernetes startup

### LINKED ISSUE #3758: [Bug] Kubernetes startup skips global config validation
### Proposed Workgroup

Enterprise & Environment

### Describe the bug

When `global.router.config_source` is `kubernetes`, `validateConfigStructure` returns before running the shared contract validators. This defers routing checks needed after CRD conversion, but also skips validation of static global settings already used during startup.

The reconciler validates the merged configuration later, after runtime initialization has begun.

### To Reproduce

Load this document with `config.Parse` or `config.ParseYAMLBytes`:

```yaml
version: v0.3
global:
  router:
    config_source: kubernetes
  stores:
    memory:
      default_similarity_threshold: 1.5
```

Parsing succeeds on main at `f0503475`. Changing `config_source` to `file` rejects the same threshold because it must be between 0 and 1.

Similar gaps affect existing validators for cache settings, embeddings, classifier backends, Router Learning, tool filtering, and admission.

### Expected behavior

- Run static global contract validators during initial loading for both configuration sources, before runtime resources are created.
- Keep Kubernetes startup tolerant of routing state supplied by CRDs.
- Validate the complete configuration after CRD conversion and before activation.
- Centralize validator registration and stage selection so new global checks apply to both paths.

### Impact

Invalid static settings can reach runtime initialization in Kubernetes mode despite being rejected in file mode, causing delayed errors or unintended runtime behavior.

### Additional context

This predates #3483 and was identified during its review. The shared validation dispatch fix is independent of that PR's memory persistence changes.

Kritoooo:
Closes #3758

## Purpose

Kubernetes configuration loads skip static contract checks that file mode runs before runtime initialization. Run the shared global validators for both sources, and defer checks that require routing or recipe state until CRDs have been merged.

- Centralize validator registration and stage selection in `validator_dispatch.go`.
- Separate deployment definitions from binding compilation, and complexity backend checks from routing-rule checks.
- Preserve nil guards and validation in the public binding compiler.
- Report `ValidationFailed` when parsing rejects a merged CRD candidate, before publishing it.
- Cover both config loaders, invalid startup globals, stale file/recipe bindings, and CRD replacement of those bindings. Keep the regression tests in existing startup and reconciliation test files.

## Scope

Runtime activation, listener readiness, retries, process health, and Helm changes have been extracted into a local follow-up branch and are tracked in #3811. This PR covers validation stages; clearing file routing before model preparation belongs to that follow-up.

## Validation

Focused Go tests passed in `pkg/config`, `pkg/k8s`, and `cmd`, covering loader parity, nil inputs, deployment/binding validation, complexity contracts, CRD validation, and startup rejection. The affected Ginkgo dispatch specs also passed.

Native libraries were replaced with fail-on-call linking stubs for these checks; no model inference ran. Full suites and cluster E2E were not run.


src/semantic-router/pkg/config/validator_dispatch.go
```diff
@@ -0,0 +1,112 @@
...
+		validatePIIModelBackendContracts,
+		validateGlobalRouterLearningConfig,
+		validateReMoMContracts,
+		validateFusionContracts,
+		validateFlowContracts,
+		validateAdvancedToolFilteringConfig,
+		validatePromptCompressionContracts,
+		validateHallucinationContracts,
+		validateModelAdmissionContracts,
+		validateModelDeploymentContracts,
```
Xunzhuo, line 37: **Keep routing bindings out of the static stage**

Impact: `validateModelDeploymentContracts` calls `CompileModelBindings`, which traverses routing and recipe bindings. Kubernetes bootstraps can therefore fail before the controller starts when the file contains stale or incomplete binding state, even though `CanonicalStaticConfigFromRouterConfig` deliberately discards that state and replaces it from CRDs.

Required change: Split the deployment-only checks from binding compilation and run the binding/recipe portion only at the complete stage; add a Kubernetes-source regression case whose ignored routing binding is invalid.

Xunzhuo (commented):
Thanks for tightening Kubernetes startup validation. One stage-boundary issue still needs a fix before this can land.

Kritoooo:
> Thanks for tightening Kubernetes startup validation. One stage-boundary issue still needs a fix before this can land.

Thanks, Fixed in c27eb9fc. Split deployment checks from binding compilation so routing/recipe bindings are validated only at the complete stage. Added regression coverage for stale bindings.

Kritoooo:
Also fixed the startup path in e18c524e. Kubernetes now discards routing profiles from the static file before model downloads and runtime initialization, preventing stale bindings from blocking controller startup. Added startup regression coverage; provider/global settings and file-mode behavior remain unchanged.

Kritoooo:
I kept the first commit you reviewed (`19dc1cb21`) unchanged and grouped the follow-ups into two commits:

- **`0f9f941c` — Defer CRD routing validation during Kubernetes bootstrap.** This addresses your stage-boundary comment: static validation now checks deployment definitions without compiling routing/recipe bindings. Binding validation runs after CRD merge. The startup loader also discards CRD-owned file routing before model preparation, so stale file bindings cannot re-enter through that path. Regression coverage checks both config loaders, startup preparation, and CRD replacement.
- **`82253e44` — Activate Kubernetes configs before reporting readiness.** Tracing the bootstrap path exposed existing activation and status gaps. This commit waits for runtime activation and initial listener setup before acknowledging updates, preserves retryable candidates, and repairs status without rebuilding an already active configuration. It also updates Helm/probes, startup progress, and the waiting-for-config display for that lifecycle.

The origins of these changes are different:

- **Existing issues fixed along the same path:** the controller acknowledged publishing a candidate before the runtime had applied it; a publication before watcher subscription could be missed; resource replacements with the same generation could be skipped; and a conflict in one CR kind left the other kind's Ready condition unchanged.
- **Corrections within this PR:** the static-stage binding validation above, nil guards lost during validator extraction, status-write errors blocking or repeating activation, and cancellation/retry sequencing in the new acknowledged update path. The probe/namespace compatibility checks and waiting-state display complete the new startup behavior.

The runtime/deployment changes extend the original validation-only scope: once file routing is discarded, Kubernetes startup must wait for CRD activation, while process liveness must remain independent of routing readiness. That is why these changes are grouped together.

Kritoooo:
Thanks for the review. The stage-boundary fix is in `d345cd365`:

- Static validation now checks deployment definitions without compiling routing or recipe bindings. Binding checks run at the complete-config stage after CRD merge.
- Regression coverage exercises both config loaders: stale bindings are tolerated during the initial Kubernetes load, rejected by complete validation and file mode, and replaced by CRD routing during reconciliation.
- The public binding compiler retains its validation, and nil guards removed during validator extraction are restored.

Tracing this path also exposed an existing gap between publishing a candidate and confirming runtime activation. Those fixes expanded the PR beyond its original scope, so I moved the activation, readiness, status retry, and Helm/probe changes to a separate local branch, tracked in #3811. Clearing file routing before model preparation moves with that lifecycle work. No follow-up PR has been opened yet.

This PR now contains only validation changes, their regression tests, and validation-stage documentation, and still closes #3758. The original validation commit, `62d06a352`, is unchanged.

Xunzhuo (approved):
The startup and CRD paths now share the stage contracts, including the actual startup-loader rejection and CRD publication/status regressions. The exact-head Core and complexity backend E2E checks pass.

wilsonwu (approved):
LGTM

Xunzhuo (approved):

PR #3763, opened 2026-09-12
[Feature] Block keys in chunked_sdpa with an online softmax

### LINKED ISSUE #3737: [Feature] Bound the key dimension in chunked_sdpa with an online softmax
### Goal

- `chunked_sdpa` blocks queries only. A global layer still forms `(b, heads, 512, k_len)` scores per block (`attention/chunked_sdpa.rs:106`).
- Peak GPU memory grows linearly with `k_len`. T4, f32, 16 heads, head_dim 128, one layer:

  | seq | global | causal |
  |---|---|---|
  | 8K | 1.7 GB | 1.8 GB |
  | 16K | 3.3 GB | 3.8 GB |
  | 32K | 6.5 GB | 6.8 GB |

- q, k, v, and the output are 1.1 GB of the 32K figure. The rest is the score block, the softmax, and copies.
- Block keys too and fold the softmax across key blocks with a running max and sum.

### Proposed Workgroup

### Scope

- Add a key block loop inside the query block loop.
- Keep a running max, sum, and accumulator per query block. Rescale on each key block.
- Apply pad, band, and causal masks per key block. Skip fully masked key blocks.
- Keep the public signature and `ChunkedSdpaConfig` unchanged.
- Run every existing dense comparison in `chunked_sdpa_test.rs` with a small key block too.
- Report the same table after the change.

### Non-goals

- Flash-attn, Metal, or any fused GPU kernel.
- Changing block sizes for callers.
- The f64 upcast in Qwen3 and Gemma3 (#3736).

### Completion signal

- Peak memory at 32K stays within a constant of the 8K figure.
- All existing `chunked_sdpa` tests pass with key blocking on, at the current 1e-4 tolerance.

guan404ming:
Closes #3737

## Purpose

- `chunked_sdpa` blocked queries only, so a global layer still held a `(b, heads, 512, k_len)` score block.
- Block keys too and fold the softmax across key blocks with a running max and sum.
- Public signature and `ChunkedSdpaConfig` unchanged; `candle-binding` attention only.

## Test Plan

- `cargo test --release --lib -- chunked_sdpa matches_dense prefill_and_decode`
- Peak GPU memory of one layer on a T4, f32, 16 heads, head_dim 128.

## Test Result

- Passed. 21 kernel and model-level dense comparisons at 1e-4.

| seq | global before | global after | causal before | causal after |
|---|---|---|---|---|
| 8K | 1.7 GB | 0.6 GB | 1.8 GB | 0.5 GB |
| 16K | 3.3 GB | 1.0 GB | 3.8 GB | 0.9 GB |
| 32K | 6.5 GB | 1.7 GB | 6.8 GB | 1.6 GB |

- 32K wall time unchanged: global 8.6 s to 8.7 s, causal 5.7 s to 5.6 s.


candle-binding/src/model_architectures/attention/chunked_sdpa.rs
```diff
@@ -110,39 +128,53 @@ pub fn chunked_sdpa(
...
-        // Causal triangle: keep only keys j <= i. Composes with the window band and
-        // the padding mask by addition (-inf + anything = -inf).
-        if cfg.causal {
-            let causal =
-                build_causal_mask(q_abs_start, blk, ks, kw, device)?.to_dtype(scores.dtype())?;
-            scores = scores.broadcast_add(&causal)?;
+            let block_max = scores.max_keepdim(D::Minus1)?;
+            running = Some(match running {
+                None => {
+                    let probs = scores.broadcast_sub(&block_max)?.exp()?;
```
Xunzhuo: Please handle an all-masked row before subtracting its tile maximum. If the first key tile contains only `-inf` for a query, this computes `-inf - (-inf)` and initializes a NaN state that later valid keys cannot recover.

A deterministic CPU F32 repro is `Q=K=0`, `V=1`, shape `(1, 1, 1024, 1)`, `scale=1`, `window=Some(4)`, `causal=false`, `q_offset=0`, and the public `block_size=0`. Main returns 1024 finite ones; this head returns 508 NaNs. With default block sizes, a single query over 1024 keys and an additive `-inf` mask on the first 512 keys also returns NaN despite valid later keys.

Please make the empty online state safe per row and let later valid tiles initialize it. Skipping only a wholly masked tile does not cover tiles where some query rows are valid and others are not.

Xunzhuo (commented):
I reproduced two numerical regressions against main `f400374` on CPU using the repository's pinned Candle revision. Key blocking is a useful follow-up to the long-context work, but the all-masked-row handling and low-precision accumulation need fixes before the numerical-equivalence claim holds. The inline comments include deterministic inputs and expected results. The existing error-reduction helper also needs finite-value assertions because it can report zero error for outputs containing NaNs.

candle-binding/src/model_architectures/attention/chunked_sdpa.rs
```diff
@@ -110,39 +128,53 @@ pub fn chunked_sdpa(
...
+                    let sum = probs.sum_keepdim(D::Minus1)?;
+                    (block_max, sum, probs.matmul(&v_win)?)
```
Xunzhuo, line 177: The unnormalized `probs.matmul(V)` can overflow FP16 before the final division, even when the correct attention output is small and finite. This also applies to the update branch below.

CPU FP16 repro: one query/head/dimension, 512 keys, `Q=K=0`, `V=200`, `scale=1`, no padding/window/causal mask. Main returns `200`; this head returns `+Inf`. The intermediate sum is `512 * 200 = 102400`, which exceeds FP16 range before division by 512.

Please perform the online softmax state and weighted-value accumulation in FP32 for FP16/BF16 inputs, including the matrix multiplication, and cast back after normalization. Add the finite FP16 case as a regression test; checking only F32 dense equivalence misses it.

candle-binding/src/model_architectures/attention/chunked_sdpa_test.rs
```diff
@@ -486,3 +486,90 @@ fn test_chunked_sdpa_rejects_a_block_with_no_keys() {
...
+                            block_size: block,
+                            window,
+                            causal,
+                            scale,
+                            q_offset: 0,
+                        };
+                        let out =
+                            chunked_sdpa_with_key_block(&q, &k, &v, pad_mask, &cfg, key_block)
+                                .unwrap();
+                        let diff = max_abs_diff(&out, &reference);
```
Xunzhuo, line 536: Please check every element of both `out` and `reference` for finiteness before calling `max_abs_diff`, in both new tests. The helper's `.max(0)` reduction can ignore NaNs mixed with finite differences: the 1024-token/window-4 repro in the other comment contains 508 NaNs but reports `max_abs_diff = 0`, so this assertion accepts it. The 8-token/query-block-8/key-block-1/window-1 case similarly contains 6 NaNs and reports zero error.

The head's 13 kernel tests pass in the same CPU harness despite these counterexamples. Finite-value assertions plus the all-masked-first-tile and FP16 overflow cases are needed for the equivalence tests to detect these regressions. Please also include the new regressions in the CI test selection.

FAUST-BENCHOU (commented):
The key-blocking direction looks good, but the current head still has the numerical blockers already reproduced in the open review threads: all-masked key tiles can poison the online state with NaNs, and FP16/BF16 weighted-value accumulation can overflow before normalization. Holding LGTM until those are fixed and covered by finite-value regression tests.

candle-binding/src/model_architectures/attention/chunked_sdpa.rs
guan404ming: Good catch. The subtracted max is now clamped to a finite floor in 179e2e550, so all-masked rows stay zero until valid keys.

candle-binding/src/model_architectures/attention/chunked_sdpa.rs
guan404ming: Thanks. F16 and BF16 tiles compute and accumulate in F32, then cast back after normalization.

candle-binding/src/model_architectures/attention/chunked_sdpa_test.rs
guan404ming: Both tests assert finiteness, and three new regressions fail on the old head and join RUST_CI_LIB_TESTS.

guan404ming:
> The key-blocking direction looks good, but the current head still has the numerical blockers already reproduced in the open review threads: all-masked key tiles can poison the online state with NaNs, and FP16/BF16 weighted-value accumulation can overflow before normalization. Holding LGTM until those are fixed and covered by finite-value regression tests.

Both blockers are fixed in 179e2e550 with finite-value regressions; see the two threads above. Thanks!

Xunzhuo (approved):

wilsonwu (approved):
LGTM

guan404ming:
Thanks!

Issue #3703, opened 2026-09-10
[Feature] Make golangci-lint a real quality gate across Go modules


drivebyer:
## Problem

`golangci-lint` is configured but does not act as a quality gate. `make check` lints only changed files with `--new-from-rev`, and no CI job runs a full-module lint of `src/semantic-router` or `e2e`, so violations accumulate on `main`. Running the repository config against `main` today:

| Module | How CI runs it | Full-module result with the repository config |
|---|---|---|
| `src/semantic-router` | changed files only | 66 issues (govet shadow 40, staticcheck 6, errorlint 4, gci/gofumpt 7, …) |
| `dashboard/backend` | full module | 0 |
| `deploy/operator` | action without `--config`, so default linters only | 15 issues (0 with the default config) |
| `e2e` | changed files only | 250 issues (gci 108, govet 45, staticcheck 37, gosec 26, errcheck 20) |
| `perf`, `*-binding` Go modules | not linted | — |

Related gaps:

- The tool version `v2.5.0` is hard-coded in six places (four workflows, `tools/make/agent.mk`, `tools/docker/Dockerfile.precommit`); upstream is at v2.13.x.
- The config lives at `tools/linter/go/.golangci.yml`, so every caller passes `--config` by hand and editors do not discover it.
- `nolint` directives are unaudited: 128 in total, 22 are dead (`nolintlint`), 20 `nolint:gosec` carry no explanation, and several target linters that are not enabled (`cyclop`, `funlen`, `gocognit`, `nestif`).
- Formatting is split: pre-commit runs `gofmt` while the lint config enables `gofumpt` and `gci`, producing drift that only surfaces when a file happens to be touched.
- The config repeats revive rules disabled in `settings` again under `exclusions.rules`, excludes `SA1019` globally (hiding all deprecation warnings, currently zero hits), and disables seven gosec rules repository-wide even though the code already suppresses them per site.

## Proposal

Align with the practice of Kubernetes, Envoy Gateway, Cilium, Prometheus, and golangci-lint itself:

- Full-module lint for every Go module is a blocking PR check; the changed-files path stays as the fast local loop.
- One `.golangci.yml` at the repository root, inherited by all modules without `--config`.
- One source of truth for the golangci-lint version, consumed by Makefiles, workflows, and images; upgrade to a current v2 release.
- `nolintlint` enabled with required explanations and specific linters; dead and unexplained suppressions cleaned up.
[...]

PR #3772, opened 2026-09-13
[CI/Build] Lint the operator with the shared golangci-lint config and format Go with golangci-lint fmt

drivebyer:
Related #3703

## Purpose

Second step of #3703: gate `deploy/operator` with the shared lint config and make `golangci-lint fmt` the only Go formatter.

- `operator-ci.yml` passes `--config tools/linter/go/.golangci.yml` to the lint action (it ran with default linters before); the separate `gofmt -s -l` step is removed. The 14 resulting findings are fixed (6 shadow, 4 gci, 2 gofumpt, 2 importas).
- `GOLANGCI_LINT_CONFIG` in `tools/make/common.mk` replaces the four hard-coded config paths. The config stays under `tools/linter/go/` like the other languages' configs; a root-level `.golangci.yml` is dropped from the #3703 plan.
- The `go-fmt` pre-commit hook runs `tools/agent/scripts/go_fmt.py` (`golangci-lint fmt` per module) instead of `gofmt -w`, which disagreed with the gofumpt/gci rules the lint config enforces.
- 14 of 15 Go modules are formatted once (137 files, mostly `0755` → `0o755` and import grouping) so the hook produces no unrelated diffs later. 5 pre-existing findings on touched lines are fixed, and one map literal is split so gofmt and gofumpt agree. `perf` is excluded because it did not build on `main` (#3773).
- `pre-commit.yml` diagnostics and contributor docs name `golangci-lint fmt`.

Not in scope: full-module blocking lint for `src/semantic-router` and `e2e`, nolint audit, narrowing exclusions.

## Test Plan

```bash
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b /tmp/gl v2.13.2
cd deploy/operator && /tmp/gl/golangci-lint run ./... --config ../../tools/linter/go/.golangci.yml --timeout=5m
cd deploy/operator && go build ./... && go test ./...
for m in $(git ls-files "*/go.mod" | xargs -n1 dirname); do (cd "$m" && /tmp/gl/golangci-lint fmt --diff --config "$PWD/../tools/linter/go/.golangci.yml" ./... | grep -c '^--- '); done
git diff --name-only origin/main | grep '\.go$' | xargs gofmt -l
GOPATH=/tmp python3 tools/agent/scripts/go_fmt.py e2e/pkg/cluster/kind.go deploy/operator/main.go README.md
for m in e2e dashboard/backend deploy/operator; do (cd $m && go vet ./...); done
make -n go-lint | grep -o 'golangci-lint run.*'
```

## Test Result

| Check | main | this PR |
|---|---|---|
| `deploy/operator` with shared config | 14 (govet 6, gci 4, gofumpt 2, importas 2) | 0 |
| `deploy/operator` `go build` / `go test` | ok | ok |
| `golangci-lint fmt --diff` files per module | e2e 111, src/semantic-router 14, perf 5, candle-binding 2, nlp/onnx/openvino-binding 1 each, pii training 1 | 0 in 14 modules; perf still 5 (deferred to #3773) |
| `gofmt -l` on every changed `.go` file | — | empty |
| `go_fmt.py` hook on a deliberately mis-formatted file | — | reformatted, non-Go argument ignored, exit 0 |
| `go vet` in e2e, dashboard/backend, deploy/operator | ok | ok |
| `make check BASE_REF=origin/main` Go lint step | — | 0 changed-file issues |
| `make -n go-lint` / `make -n dashboard-lint` | `--config ../../tools/linter/go/.golangci.yml` | `--config <repo>/tools/linter/go/.golangci.yml` |

`make check` locally then fails in `dashboard-test-backend` on `TestEvaluationRoutesLoadDeploymentScopedTargetsWithoutCatalogLeaks`; the same test fails identically on `origin/main` on macOS (needs Linux no-follow filesystem support) and is unrelated to this PR. `src/semantic-router` does not link locally without the Rust libraries, so it was checked with `go vet` on the touched packages and `gofmt -l` instead of a full build. `actionlint` was not run locally.


drivebyer:
The remote-embedding and routing-strategies E2E failures on this PR are tracked in #3780. They reproduce on main after #3768 and #3760 and are not caused by this PR, whose only change to those profiles is import formatting.

tools/agent/scripts/go_fmt.py
```diff
@@ -0,0 +1,33 @@
...
+
+import subprocess
+import sys
+
+from check_support import GO_LINT_CONFIG, REPO_ROOT, group_files_by_module
+from go_lint_support import resolve_golangci_lint
+
+
+def run_go_fmt(changed_files: list[str]) -> int:
+    grouped = group_files_by_module(changed_files, "go.mod", {".go"})
```
Xunzhuo, line 23: [P2] Preserve formatting for Go tools outside module roots

`group_files_by_module` drops files with no ancestor `go.mod`, including the existing commands under `tools/dev/dsl`, `tools/calibration/image-routing`, and `bench/grounded_fusion/fusioneval`. Replacing `gofmt -w` with this hook therefore returns success without formatting those sources. A misformatted fixture at each of these paths remains unchanged with exit 0. Please handle these external tools explicitly and add a regression test alongside the module-local cases.

Xunzhuo (changes requested):
The operator lint and race tests pass. The new formatter hook silently skips Go commands stored outside module roots; please preserve their formatting coverage and add a regression test.

tools/agent/scripts/go_fmt.py
drivebyer: Fixed in 670ae1a36. Go files with no ancestor `go.mod` are now formatted from `src/semantic-router` with `../../<path>` arguments, mirroring how `tools/make/go-tools.mk` builds and lints them. Verified locally that a misformatted `tools/dev/dsl/*.go` fixture is rewritten. Regression tests in `tools/agent/scripts/tests/test_go_fmt.py` cover module-local files, the three external tool paths, mixed changes, and ignored inputs.

drivebyer: Thanks @Xunzhuo, fixed.

drivebyer:
The two remaining Kubernetes failures are not caused by this PR (its router changes are formatting only):

- **Production Stack**: `jailbreak-detection` accuracy 66.67% (4/6) below 80%. The same failure appears on #3210's run (https://github.com/vllm-project/semantic-router/actions/runs/35293064400), so it is a pre-existing flake.
- **Multimodal Routing**: the router crashes at startup with `contrastive jailbreak with model_type=multimodal: embedding model "multimodal" was not prepared for this generation`. This is a main bug introduced by the strict recipe embedding set (#3768/#3876): the contrastive jailbreak builder initializes the multimodal embedding even when the recipe has no contrastive rules. Fix in #3906.

drivebyer:
The 3 failing Kubernetes e2e jobs are pre-existing on main, not caused by this PR: Multimodal Routing helm timeout also fails in #3978 (run 35541071634), and the jailbreak-detection failures in Production Stack / Multiple Endpoints also fail on feat/3210-per-model-timeouts (run 35486177468). This PR only changes formatting/lint config and touches no runtime code.

PR #3776, opened 2026-09-13
[Router] Persist final request demand in Replay

Peterren:
Related #3704

## Purpose

#3748 added bounded request-demand snapshots at four stable lifecycle stages. Router Replay starts before tool selection to preserve timing and failure coverage, so its initial record does not yet contain the later `provider_bound` observation.

This follow-up completes the diagnostic evidence flow without moving Replay start: immediately before a Replay record becomes terminal, it replaces only that record's bounded request-demand snapshot list with the request's latest observations. Completed dispatch records therefore retain `original`, `post_context`, `post_tool_policy`, and `provider_bound` evidence for later request-level analysis.

These snapshots remain explicitly diagnostic. They use stage/source/representation provenance and are not the final admission contract or a claim of wire-token exactness.

## Design

- Add a narrow `UpdateRequestDemandSnapshots` Replay enrichment operation.
- Implement it consistently for memory, Redis, PostgreSQL, Milvus, and Qdrant stores.
- Preserve all other route diagnostics and clone caller-owned snapshots before storage.
- Update PostgreSQL with `jsonb_set` so unrelated route diagnostics are not overwritten.
- Treat enrichment as best-effort: a storage failure is logged, while the existing lifecycle transition still completes.

## Safety boundary

- No model admission, selection, fallback, provider request, or metric behavior changes.
- No additional persistence occurs before provider dispatch or time to first token.
- Replay's start point, timestamp, and failure coverage remain unchanged.
- No Dashboard UI is included. The provisional lifecycle stages should not become a product contract before #3704 defines an immutable prepared-request and model-specific demand-assessment boundary.

## Architecture checkpoint

The next slice should be an observe-only typed contract that freezes the provider projection admitted for one candidate, including payload identity, model/tokenizer/template revisions, effective output reserve, count quality, and context limit. Dispatch and enforcement remain unchanged until that contract is independently validated.

## Test Plan

```bash
make rust-ci
make config-schema-check

cd src/semantic-router
go test ./pkg/extproc -count=1
go test ./pkg/routerreplay/... -count=1
go vet ./pkg/extproc ./pkg/routerreplay/...
golangci-lint run --config ../../tools/linter/go/.golangci.yml \
  --new-from-rev origin/main --issues-exit-code 1 --path-mode abs \
  ./pkg/extproc ./pkg/routerreplay ./pkg/routerreplay/store
```

Additional contract coverage verifies that an in-progress Replay retains its original three snapshots, terminal finalization persists all four stages, snapshot inputs do not alias stored data, unrelated route diagnostics survive, and enrichment failures do not block lifecycle completion.


src/semantic-router/pkg/routerreplay/store/redis.go
```diff
@@ -347,6 +347,18 @@ func (r *RedisStore) UpdateUsageCost(ctx context.Context, id string, usage Usage
 	})
 }
 
+// UpdateRequestDemandSnapshots replaces bounded request-demand evidence for a record.
+func (r *RedisStore) UpdateRequestDemandSnapshots(
+	ctx context.Context,
+	id string,
+	snapshots []RequestDemandSnapshot,
+) error {
+	return r.updateRecord(ctx, id, func(record *Record) bool {
```
Xunzhuo, line 356: **Make shared-store updates atomic across replicas**

`updateRecord` protects this GET-modify-SET only with this `RedisStore` instance's `updateMu`. Once a response exposes its Replay ID, an outcome request handled by another router replica can run `AppendOutcome` on the same record while terminal finalization enters here; both readers can load the same old value, and the last full-record SET silently drops either the outcome or the final snapshots. Milvus and Qdrant use the same instance-local-lock plus full-record-upsert pattern. Please use a backend-atomic field update or versioned CAS/retry for these shared stores, and cover concurrent snapshot/outcome updates through two independent store clients.

Xunzhuo (commented):
Thanks for carrying the final snapshot through every Replay backend. One shared-store concurrency issue still needs to be fixed.

PR #3775, opened 2026-09-13
[Docs] Restore current zh-Hans documentation coverage

FAUST-BENCHOU:
## Summary
- Restore Latest `zh-Hans` docs so Chinese readers can install, configure, and operate without English fallback on current pages.
- Translate the homepage and primary nav strings that previously fell back to English.
- Keep `community/` docs and the Community site roster in English. Leave generated CRD reference as English fallback.

Closes #3220
Closes #3221
Related #3074

## Test plan
- [ ] `make docs-check-translation-coverage` passes
- [ ] Preview `npm run start:zh` (or `npx docusaurus start --locale zh-Hans`)
- [ ] Check `/zh-Hans/docs/intro` and `/zh-Hans/docs/installation`
- [ ] Confirm `/zh-Hans/docs/community/overview` stays English
- [ ] Confirm `/zh-Hans/` homepage and top nav are Chinese

Issue #3074, opened 2026-08-27
[Epic] Restore and maintain current zh-Hans documentation


yaojiejia:
## Parent Direction

Developer Experience & Ecosystem — #2970

## Goal

Restore and maintain current Simplified Chinese documentation so Latest pages do not silently fall back to English after content moves or bulk deletions.

## Current-main baseline

Current main has 158 English documentation pages but only a small set of current zh-Hans overrides. Commit 7c874be2 in #2869 deleted 134 current translation files while frozen v0.3 translations remained, causing widespread English fallback on Latest.

## Current work

- #3220 — restore core, overview, installation, troubleshooting, and community coverage.
- #3221 — restore API, benchmarking, training, and tutorial coverage.
- #3222 — add a translation-coverage regression gate.

## Scope

- Translate current English source in reviewable section-level batches.
- Preserve technical fields, commands, links, and examples.
- Validate locale builds, links, and coverage after each batch.
- Detect unexpected coverage loss while allowing explicit reviewed baseline changes.
- Keep frozen release-version translations separate from Latest.

## Non-goals

- Copying stale v0.3 pages into Latest without technical review.
- Requiring every English documentation change to ship every translation immediately.
- Treating unreviewed machine translation as complete.
- Changing product behavior through localization.

## Completion criteria

- Every current documentation page has an intentional zh-Hans state.
- Maintained translated sections render without English fallback.
- CI detects unexpected translation deletion or coverage regression.
- The locale build and internal-link checks pass.

yaojiejia:
Happy to work on this after this issue gets accepted!

wilsonwu:
I will also continue to do the Chinese doc update, it is a huge work.

PR #3777, opened 2026-09-13
[Docs] Align restored zh-Hans translation contract

Peterren:
Related #3074

## Purpose

#3775 intentionally restored current Simplified Chinese translations for five pages, while the documentation contract still classified those paths as English-fallback-only. Because the assertion stops at the first unexpected file, CI initially reported only `api/apiserver.md`; the remaining restored training and troubleshooting pages fail in sequence.

Remove all five restored pages from `currentTranslationFallbackDocs` so the contract matches the current translation coverage. The three pages that still rely on canonical English fallback remain enforced.

No documentation content or runtime behavior changes.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/config -run '^TestLatestTutorialTaxonomyMatchesConfigHierarchy$' -count=1
go test ./pkg/config -count=1
```


FAUST-BENCHOU (dismissed):

wilsonwu (dismissed):
LGTM

PepperoniBlvd:
Hi @Peterren, I hit the same failure while rebasing an unrelated
branch onto current `main`, and I think this fix needs four more lines to make
the lane green.

#3775 restored zh-Hans translations for five of the eight paths listed in
`currentTranslationFallbackDocs`, not just one:

```
website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/api/apiserver.md                   ← removed by this PR
website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/training/training-overview.md
website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/training/model-performance-eval.md
website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/troubleshooting/common-errors.md
website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/troubleshooting/vsr-headers.md
```

Only `apiserver.md` shows up in the failure output because
`assertPathsDoNotExist` uses `t.Fatalf`, so the test stops at the first match
and the remaining four stay hidden until it is removed.

I verified this by applying this PR's exact change on top of `upstream/main` in
a scratch worktree and rerunning the test:

```bash
git worktree add /tmp/wt-main upstream/main
# drop the api/apiserver.md entry from currentTranslationFallbackDocs
cd /tmp/wt-main/src/semantic-router
go test ./pkg/config -run '^TestLatestTutorialTaxonomyMatchesConfigHierarchy$' -count=1
```

```
--- FAIL: TestLatestTutorialTaxonomyMatchesConfigHierarchy (0.00s)
    docs_contract_test.go:617: website/i18n/zh-Hans/docusaurus-plugin-content-docs/current/training/training-overview.md
        should not exist; let latest docs fall back to the canonical current source instead
```

So as it stands the PR moves the failure rather than clearing it, and the
Router test lane stays red.

src/semantic-router/pkg/config/docs_contract_test.go
```diff
@@ -582,7 +582,6 @@ var currentTranslationFallbackDocs = []string{
 	repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "cookbook", "classifier-tuning.md"),
 	repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "cookbook", "pii-policy.md"),
 	repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "cookbook", "vllm-endpoints.md"),
-	repoRel("website", "i18n", "zh-Hans", "docusaurus-plugin-content-docs", "current", "api", "apiserver.md"),
```
PepperoniBlvd: #3775 restored zh-Hans translations for five of the eight paths in this list, so removing api/apiserver.md alone leaves the contract in conflict with the tree. The assertion uses t.Fatalf, so it reports only the first match, which is why CI on this PR now fails on training/training-overview.md (docs_contract_test.go:617) instead of passing

AayushSaini101 (approved):

PR #3781, opened 2026-09-13
[Bug] Own prepared embeddings on the router and stop treating default card capabilities as declarations

### LINKED ISSUE #3780: [Bug] remote-embedding and routing-strategies E2E profiles fail on main after #3768 and #3760
### Proposed Workgroup

### Describe the bug

Two E2E profiles fail on `main` since #3768 (`bce357c51`) and #3760 (`f8c1197a9`) merged on 2026-09-13. Neither PR pipeline scheduled these profiles.

**remote-embedding**: `remote-embedding-routing` reads `/startup-status` and gets `{Ready:true EmbeddingProvider:<nil>}`. The `embedding_provider` field is produced in `src/semantic-router/pkg/modelruntime`, which #3768 rewrote.

**routing-strategies**: `input-modality-routing` gets HTTP 400 `unsupported_capability`: `model "base-model" does not declare the required tasks: image_input`. The check in `providerCapabilityMismatch` (`src/semantic-router/pkg/extproc/processor_req_body_routing.go`) was introduced by #3760. `base-model` in `e2e/profiles/routing-strategies/values.yaml` declares no capabilities, yet is treated as annotated.

`entropy-routing` passes, but its verbose output shows every sub-case with empty uncertainty, `use_reasoning=false` and confidence `0.000`.

### To Reproduce

Open a PR that touches enough paths for `.github/workflows/pr.yml` to schedule the full E2E set, e.g. #3772.

- Run 34740184509 (head `42a4871c4`): remote-embedding job 103679837563 and routing-strategies job 103679837557 fail as above.
- Run 34760680758 (head `df5a518dd`, includes #3774): in progress at filing time.

#3772 only changes import formatting in these two profiles. It is the only branch that ran them after the merges.

### Expected behavior

`/startup-status` reports `embedding_provider` when a remote embedding provider is configured, and image requests in routing-strategies are dispatched to `base-model` with 200. Both profiles pass on `main`.

### Impact

Any PR whose diff schedules these profiles gets a red PR Gate. #3772 is blocked by it.

### Additional context

`production-stack` (killed after 64 min on #3772) and `no-model` (context deadline exceeded on main) look infrastructural and are not included.

drivebyer:
Closes #3780

## Purpose

Two E2E profiles fail on `main` since #3768 and #3760. Module: `src/semantic-router` (`pkg/extproc`, `pkg/config`). Owner: Router Models & Inference Runtime.

- `remote-embedding`: `buildRouter` never assigned `components.embeddings` to `OpenAIRouter.Embeddings`. `EmbeddingState` saw a nil set, so `/startup-status` reported `embedding_provider: null` and `startup_complete` logged `embedding_ready:false` even though the remote binding was ready. The router now owns the prepared set. This also unblocks the other readers of `r.Embeddings` (semantic cache, tools database, RAG windowing).
- `routing-strategies`: sparse custom model cards get an internal default `capabilities: [chat]`. Since #3760, `providerCapabilityMismatch` treats any parseable capability as an operator declaration and rejects tasks not listed, so a name-only `base-model` card refused image requests with `unsupported_capability`. `ModelParams.Capabilities` is now filled only when the card's provenance records capabilities (built-in or operator), leaving default-only cards unannotated as before #3760.

## Test Plan

```bash
cd src/semantic-router
go test ./pkg/config/ -run 'TestSparseCustomModelCardCarriesNoDeclaredCapabilities|TestOperatorModelCardCapabilitiesAreDeclared'
go test ./pkg/extproc/ -run TestBuildRouterOwnsPreparedEmbeddings
go test ./pkg/config/... ./pkg/catalog/... ./pkg/extproc/... ./cmd/...
```

The `remote-embedding` and `routing-strategies` E2E profiles are scheduled by this PR's path changes.

## Test Result

- The three new tests fail on `main` and pass with this change.
- `TestLatestTutorialTaxonomyMatchesConfigHierarchy` in `pkg/config` was already failing on `main`: d884380a9 restored the zh-Hans overrides for `api/apiserver.md`, `training/*.md` and `troubleshooting/*.md`, but the test still asserted they must not exist. This PR drops those entries from `currentTranslationFallbackDocs` so the package passes fully.
- E2E result pending on this PR's pipeline.


Peterren (approved):
Nit (non-blocking): could we add a built-in-card regression assertion alongside these tests, so both provenance branches stay locked down? The implementation itself looks clean.

FAUST-BENCHOU (approved):

Issue #3773, opened 2026-09-13
[Bug] perf module does not build against main: stale go.mod and removed cache.InitEmbeddingModels


drivebyer:
### Proposed Workgroup

Developer Experience & Ecosystem

### Describe the bug

The `perf` Go module no longer builds against `main`, so `Performance / component-benchmarks` fails on every PR that touches `perf/` and in the nightly build. Two independent breaks:

1. `perf/go.mod` is stale. `go test ./benchmarks/...` stops with `go: updates to go.mod needed`. Running `go mod tidy` shows why: the module lacks a `replace` for `github.com/vllm-project/semantic-router/onnx-binding` (required by `src/semantic-router` since #3768), and its `exclude google.golang.org/genproto v0.0.0-20220503193339-ba3ae3f07e29` pins an old `genproto` that still ships `googleapis/api/*`, which now collides with `google.golang.org/genproto/googleapis/api` (`ambiguous import`). Nightly on 2026-09-12 (run 34667530063) already failed with the `go.mod` error, so this predates #3768.
2. `perf/benchmarks/cache_bench_test.go:41` calls `cache.InitEmbeddingModels`, which #3768 removed from `src/semantic-router/pkg/cache`. After the `go.mod` fix the benchmarks fail to compile with `undefined: cache.InitEmbeddingModels`.

The PR workflow only runs the performance job when `perf/` changes, so `main` stays green while the module is broken.

### To Reproduce

```bash
cd perf
go test -run NONE ./benchmarks/...
# go: updates to go.mod needed; to update it: go mod tidy

printf 'replace github.com/vllm-project/semantic-router/onnx-binding => ../onnx-binding\n' >> go.mod
sed -i '/^exclude google.golang.org\/genproto/d' go.mod
go get google.golang.org/genproto@latest && go mod tidy
go test -run NONE ./benchmarks/...
# benchmarks/cache_bench_test.go:41:18: undefined: cache.InitEmbeddingModels
```

Observed on `main` at `f8c1197a` with Go 1.25.14 and 1.26.2.

### Expected behavior

`go test ./benchmarks/...` in `perf` compiles against `main`, and `Performance / component-benchmarks` passes on PRs that touch `perf/` and in the nightly build.

### Impact

[...]

PR #3788, opened 2026-09-14
[CI/Build] Harden owned cache benchmarks after module repair

Peterren:
Related #3773
Follow-up to #3774.

## Purpose

The cache benchmark now compiles and runs through the owned embedding API, but a few lifecycle and validation details still needed to match that contract:

- use the indexed `cuda:0` device required by Candle deployment validation when `USE_GPU=true|1`;
- retain the owning embedding set for the benchmark process and close it after all tests and benchmarks complete;
- make the cache benchmark documentation explicit that it measures the same owned provider used by router generations, rather than the legacy process-global batching FFI;
- expose a focused perf module tidy target and run it in the Performance workflow, so a stale `perf/go.mod` fails on the change that introduces it.

Restoring the legacy global batching FFI here would reintroduce a lifecycle model that #3768 intentionally replaced. A future owned batching provider should be added at the native runtime seam and benchmarked as a separate capability.

## Test Plan

```bash
GOTOOLCHAIN=go1.25.0 make check-go-mod-tidy
cd perf
CGO_ENABLED=0 go test ./... -count=1
CGO_ENABLED=0 go vet ./...
cd ..
```

Additional validation:

- `check-perf-go-mod-tidy` passes on the clean tree.
- Removing a required entry from staged `perf/go.mod` makes `check-perf-go-mod-tidy` fail and print the generated diff.
- `cargo build --release --lib --no-default-features` passes in `candle-binding`.
- The required Performance workflow will run the model-backed component benchmarks and the newly wired tidy step.


Xunzhuo (approved):
The benchmarks now use owned embedding instances, close them on failure, and preserve the test exit status. The module and benchmark checks pass on this head.

wilsonwu (approved):
LGTM

Xunzhuo:
Closing as superseded by #3852. Current main already includes the CUDA device selection, owned embedding cleanup on failure and process exit, and perf module tidy covered here, adapted to the Vela benchmark harness. This branch targets the previous Qwen3/workflow layout, so these fixes no longer need a separate merge. Thanks for the benchmark fixes.

PR #3769, opened 2026-09-12
[Feature] Adopt Vela models and strengthen router recipes

Xunzhuo:
Related #2760

## Purpose

Make Vela the default family for routing classifiers, embeddings and reranking. Keep model revisions, downloads, task input limits, CPU/AMD execution and training/runtime documentation consistent. Guard identifies prompt attacks; Safety and Hazard provide separate content-risk signals.

Refine the five built-in recipes into 16 compact decisions that combine learned signals, request structure and projections. Preserve eligible models during conversations and active tool cycles, and expose execution choices in Preview, Replay and Dashboard. Propagate request cancellation through embedding and complexity inference without publishing partial results or releasing running native work early. Keep session and conversation identities isolated across recipes, preserve the latest dispatched model when responses finish out of order, and charge usage to the actual routing session. Version shared protection snapshots so old identity encodings cannot restore another conversation’s state.

Model weights, private datasets and evaluation run artifacts remain outside this PR.

## Test Plan

- Run scoped checks, generated-contract checks, Router/CLI/Dashboard tests and native binding regressions.
- Build CPU and AMD images from the committed source. Use the standard installed CLI and real Preview requests to verify downloads, signals, routing, session continuity and latency.
- Require the final pushed revision's CI to pass.

## Test Result

- Scoped Router, CLI, Dashboard, training, recipe and generated-contract checks passed. Cancellation regression and race checks cover canceled mixed text/image inference while retaining ordinary image-error fallback.
- Commit `6ae23dca` was built and deployed through the standard CLI on CPU and AMD. Downloads resolve the new Guard revision. Candle CPU and ONNX CPU/ROCm checks passed; explicit ROCm profiles cover 512, 8K and 32K inputs. This establishes runtime support and numerical agreement, not long-context detection accuracy.
- The six CPU end-to-end checks on `6ae23dca` passed: three generated responses, tool-cycle continuity, a fresh conversation and a budget rejection with no backend call. Earlier targeted Replay/Dashboard and multi-model workflow checks remain separately recorded.
- On paired AMD Vault probes, the new Guard improves contracts from 46/67 to 49/67 with no new regressions; all 20 newly added policy probes pass. Two short Chinese attacks still miss. The ten-case CPU subset agrees with AMD.
[...]

PR #3789, opened 2026-09-14
[Test] Add E2E coverage for the context signal

yu3zhang1:
Related #3178

## Purpose

Add e2e test for the `context` routing signal (config.SignalTypeContext).
It currently has no end-to-end behavioral coverage through a deployed
request path, as required by #3178. Unlike `event`/`language`, `context`
matches on the request's estimated token count against configured bands
rather than text content, so it needed its own test design.

This PR adds:
- A `long_context` context rule (`min_tokens: 500`, open-ended) and a
  `long_context_routing` decision (priority 1500) to
  `e2e/profiles/ai-gateway/values.yaml`.
- `e2e/testcases/context_routing.go`, reusing the `signal_routing_helpers.go`
  harness already shared by `event`/`language` (single-query request,
  checks `x-vsr-selected-decision` and the `x-vsr-matched-context`
  response header).
- `e2e/testcases/testdata/context_routing_cases.json` with two positive
  cases (padded text just past ~600 estimated tokens, and a much larger
  request at ~2400 tokens, confirming the open-ended band has no hidden
  upper bound) and one negative case (a short query whose estimated
  token count falls below the 500-token minimum, so it matches no
  configured band).
- `context-routing` registered in `BaselineRouterContract`, a new row in
  `signal_routing_contract_test.go`, and the `ENVOY_AI_GATEWAY_CI_TESTS`
  list in `.github/workflows/integration-test-k8s.yml`.

Owning Workgroup: Evaluation & Quality (`wg/evaluation-quality`).

## Test Plan

- `gofmt -l e2e/testcases/context_routing.go e2e/testcases/testdata/context_routing_cases.json`
- `go vet ./...` (run from `e2e/`)
- `go build ./...` (run from `e2e/`)
- `go test ./...` (run from `e2e/`)
- `golangci-lint run ./pkg/testmatrix ./testcases --new-from-rev upstream/main` (run from `e2e/`, using `tools/linter/go/.golangci.yml`)
- `go test ./testcases/... -run TestProfilesSelectSignalRoutingContracts`

## Test Result

- gofmt / go vet / go build: pass.
- `go test ./...` (full e2e module): pass, no failures.
- golangci-lint: 0 issues in the files this PR touches.
- `TestProfilesSelectSignalRoutingContracts`: pass, confirms the
  `envoy-ai-gateway` profile selects `context-routing`.
- Manually verified the token-count math for both test prompts against
  the router's actual estimation formulas (`CharacterBasedTokenCounter`
  in `context_classifier.go` and the `TokenFloor` computation in
  `request_context_estimate.go`) to confirm they land clearly on the
  expected side of the 500-token threshold, and verified the
  `long_context_routing` priority (1500) doesn't collide with any
  existing decision priority.
- Did not run a local Kind cluster (`make e2e-test-specific
  E2E_TESTS="context-routing"`) this round; deferring live e2e
  confirmation to CI on this PR, consistent with #3399/#3507/#3653.

---
AI assistance (Claude Code) was used to draft this change, under my review;
I read and understand the diff and take responsibility for it.


yu3zhang1:
* CI update: `context-routing` (this PR's test) passes: confirmed in the latest run: Test context-routing passed.

* FAILED - jailbreak-detection 
Error: jailbreak-detection failed acceptance contract: 16/25 correct jailbreak detections (64.00%) below minimum 70.00%.

Confirmed `jailbreak-detection` passed cleanly on main as of 2026-09-11 (PR #3731's CI run, 26.4s, 100% baseline).
Between then and now, two merged PRs both touch the mmBERT-32K jailbreak classifier's inference code (`candle-binding/src/model_architectures/traditional/modernbert.rs`):

- #3761 "[Bug] Preserve mmBERT long-context classifier correctness" — changes Candle attention padding/RoPE handling and switches sequence-classification mean pooling to FP32 accumulation.
- #3768 "[Feature] Own model instances and typed bindings across router generations" — a large model-instance-ownership refactor also touching the same file.

Either could plausibly have shifted the jailbreak classifier's output distribution. 
Note: I did not bisecting myself since it's out of scope for this PR — flagging here for visibility.

e2e/profiles/ai-gateway/values.yaml
```diff
@@ -58,6 +58,17 @@ config:
         modelRefs:
           - model: base-model
             lora_name: general-expert
+      - name: long_context_routing
+        description: Route requests whose estimated token count matches the long_context band
+        priority: 1500
```
Xunzhuo: **Keep the context fixture below the safety routes**

Impact: This open-ended `>=500`-token decision has priority 1500, above `block_jailbreak` (1000) and `block_pii` (999). Under the profile's priority strategy, a long unsafe request selects this model route instead of the matching fast-response guardrail.

Required change: Lower or constrain/compose this fixture so both safety decisions retain precedence, and add a long-context overlap case that proves the guardrail wins.

Xunzhuo (commented):
Thanks for adding the deployed context-signal path. One overlap with the shared safety routes needs to be fixed.

e2e/profiles/ai-gateway/values.yaml
yu3zhang1: Ack. I didn't consider the priority interaction, priority 1500 would let long_context_routing outrank block_jailbreak (1000) and block_pii (999). I've fixed by lowering it to 50, and added a long-context/jailbreak overlap case (case 4) that asserts the guardrail wins over long_context_routing.

But this PR's CI is currently failing, I think it may be related with the new Vela model introduced in #3769 , the jailbreak/PII guard is nondeterministically misclassifying inputs (sometimes flipping between block_jailbreak and block_pii, sometimes missing real attacks), matching what has reported in [vllm-project/semantic-router#3831](https://github.com/vllm-project/semantic-router/issues/3831).

e2e/testcases/testdata/context_routing_cases.json
```diff
@@ -0,0 +1,34 @@
...
+    "expected_decision": "long_context_routing",
```
Xunzhuo, line 6: [P2] Make the positive fixtures reach the context route

The [envoy-ai-gateway run for this head](https://github.com/vllm-project/semantic-router/actions/runs/35007535407/job/104511088869) executes this case as `block_pii`, and the second positive case as `block_jailbreak`; `context-routing` therefore finishes 2/4 and neither positive assertion exercises `long_context_routing`. This is now included in the mandatory profile. Please use fixtures verified against the profile’s configured classifiers, or isolate the positive context checks while retaining the separate safety-overlap check, and rerun both positive cases successfully. Keep the safety decisions above the context route.

e2e/testcases/testdata/context_routing_cases.json
```diff
@@ -0,0 +1,34 @@
...
+    "expected_decision": "",
+    "expected_matched_signal": "",
+    "should_match": false
```
Xunzhuo, line 32: [P2] Assert a blocking decision in the safety-overlap case

With `should_match: false`, `testSingleSignalRouting` only checks that the selected decision is not `long_context_routing` and the matched-context header is empty. An unsafe request dispatched through `other_decision` with no context match passes both checks, so this case cannot establish the guardrail precedence described above. Please assert that `block_jailbreak` or `block_pii` produced a fast response and that no model dispatch occurred, while allowing either classifier to supply the block.

Xunzhuo (changes requested):
Lowering the context route priority fixes the earlier safety-route overlap. Two issues remain in the new regression coverage: the deployed positive fixtures fail in exact-head CI, and the unsafe overlap case can pass without a guardrail blocking dispatch. Please tighten those assertions and rerun the context profile. The current envoy-ai-gateway run reports 2/4 context cases; no additional local deployment was run.

e2e/testcases/testdata/context_routing_cases.json
yu3zhang1: Thanks for the review. Similar to [#3844](https://github.com/vllm-project/semantic-router/pull/3844), I isolated the context signal into its own e2e profile:

Moved the `long_context_routing` positive-cases out of the shared ai-gateway profile into `long-context-routing`. That profile has no block_jailbreak/block_pii decisions, so no priority conflict to manage.

Since the positive test cases now run in `long-context-routing`, there's no PII/jailbreak classifier in that profile to falsely alarm on them. All three positive cases (just-past-threshold, far-past-threshold, short-request-no-match) now correctly hit long_context_routing or default-route. The Long Context Routing job passed.

e2e/testcases/testdata/context_routing_cases.json
yu3zhang1: Thanks for the insight. Replaced the `should_match`, now it asserts directly on response headers that a guardrail actually fired:
1. x-vsr-response-path == "fast_response", 
2. x-vsr-fast-response == "true", and 
3. x-vsr-selected-decision is exactly block_jailbreak or block_pii. 

Verified in this run's Envoy AI Gateway job.

e2e/profiles/ai-gateway/values.yaml
yu3zhang1: Since the context signal is isolated into its own e2e profile, the positive test cases now run in `long-context-routing`, no PII/jailbreak classifier preempt.

e2e/pkg/testmatrix/testcases_test.go
```diff
@@ -11,7 +11,7 @@ func TestBaselineScopesPreserveEveryFunctionalCase(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(full) != 38 || len(standard) != 36 {
+	if len(full) != 39 || len(standard) != 37 {
```
Xunzhuo: [P2] Keep the baseline inventory assertion in sync with the new case

Current `main` already includes `security-window-provenance` and expects 39 full / 37 standard cases. The clean merge of this head also retains `context-safety-overlap`, producing 40 / 38 while `TestBaselineScopesPreserveEveryFunctionalCase` still expects 39 / 37. Running `go test -count=1 ./pkg/testmatrix` on that integration fails with `inventory changed: standard=38 full=40`; main alone passes. Please update the inventory assertion for the combined cases and rerun `make test-e2e-unit`.

Xunzhuo (changes requested):
The earlier context-route and safety-overlap findings are resolved: the isolated context profile passes all three cases, and the shared gateway run verifies a real fast-response guardrail. One new issue appears when combined with current main: the baseline inventory test expects 39/37 cases but now contains 40/38. The focused test fails on that clean merge and passes on main. Please update the assertion and rerun the E2E unit gate. The deployed evidence is from current-head CI; no additional local deployment was run.

Xunzhuo (dismissed):
The baseline inventory now preserves both cases and correctly expects 40 full / 38 standard tests. The focused matrix regression passes, and exact-head deployment evidence passes all three context requests plus the safety-overlap guardrail in the 38-case gateway run.

wilsonwu (dismissed):
LGTM

e2e/pkg/testmatrix/testcases_test.go
yu3zhang1: Done and mark this as resolved.

yu3zhang1:
Rebased to pick up [#3993](https://github.com/vllm-project/semantic-router/pull/3993) and resolved a merge conflict in e2e/profiles/all/imports.go: local-classifier-backend and long-context-routing needed to move from the now-removed mockVLLMLocalImages to providerMockerLocalImages. Verified locally (go build ./..., go test ./pkg/testmatrix/... ./testcases/... ./pkg/framework/...) before pushing.

Heads up: the force-push dismissed both of previous approvals. CI is also queued behind a large repo-wide backlog right now, likely from [3993](https://github.com/vllm-project/semantic-router/pull/3993) touching 286 files across many open PRs at once, so it may take a while to go green again.

@Xunzhuo @wilsonwu whenever you get a chance, could you take another look/re-approve? No rush given the current CI congestion.

yu3zhang1:
@Xunzhuo @wilsonwu 
Could you take another approve? The only change is for resolving conflict in the `e2e/profiles/all/imports.go`

Issue #2976, opened 2026-08-22
[Epic] Enable safe cross-model KV-cache reuse


Xunzhuo:
## Parent Direction

MoM & Routing — #2965

## Goal

Define and validate safe cross-model KV-cache reuse so that eligible multi-model and agentic workflows can reuse computation without making cache reuse a correctness dependency.

Research reference: https://arxiv.org/html/2608.03893v1

## Scope

- Define model, tokenizer, architecture, layer, precision, prompt-prefix, and adapter compatibility requirements.
- Map existing vLLM-SR retention directives to explicit backend and cache capabilities.
- Define cache identity, ownership, isolation, privacy, lifetime, invalidation, and fallback.
- Integrate through negotiated capabilities instead of assuming every backend supports reuse.
- Evaluate correctness, reuse rate, latency, TTFT, memory and network cost, and failure behavior.
- Cover multi-model, long-session, and agentic communication scenarios.
- Define the smallest useful interoperable vertical before wider runtime support.

## Non-goals

- Treating an upstream paper or proposal as a stable production API.
- Reusing cache across incompatible model, tokenizer, or adapter boundaries.
- Silent best-effort behavior without an observable unsupported or no-op result.
- Making cache reuse a correctness dependency.

## Completion criteria

- [ ] Compatibility and capability contract.
- [ ] Privacy, isolation, and lifecycle policy.
- [ ] Bounded reference integration.
- [ ] Reproducible benchmark and correctness evaluation.
- [ ] Fallback and rollback behavior.
- [ ] Decision on wider production support.

## Interfaces and dependencies

- MoM & Routing owns when reuse is beneficial and safe within a recipe.
- Router Models & Inference Runtime supplies compatibility and capability negotiation.
- Data Plane & Networking owns cache lifecycle and fallback behavior; Evaluation & Quality validates correctness before measuring efficiency.

AayushSaini101:
@Xunzhuo i am interested on this work thanks

raghavchitkara36:
@Xunzhuo, @AayushSaini101, I am interested in contributing to this task.

AayushSaini101:
> [@Xunzhuo](https://github.com/Xunzhuo), [@AayushSaini101](https://github.com/AayushSaini101), I am interested in contributing to this task.

sure feel free to share you thoughts on this we can work together thanks : )

yaojiejia:
[...]

PR #3792, opened 2026-09-14
[Router] add KV address registry write path (#2976)

AayushSaini101:
Related: https://github.com/vllm-project/semantic-router/issues/2976

## Purpose
Implements **Step 2 (I2)** of epic #2976: record the last known KV-cache location for routed sessions so a future KVTransfer Coordinator can look up the source pod on model switch.

After a successful upstream inference, extproc writes a `kv_addr:{namespace}:{session_id}` entry containing the backend address, model, turn count, and tenant namespace. The registry reuses the semantic response-cache Redis/Valkey connection when configured; otherwise it falls back to a no-op implementation.
**Changes:**
- Add `pkg/kvtransfer/` with `AddressRegistry`, in-memory and Redis/Valkey backends, and key encoding (`kv_addr:{namespace}:{session_id}` with colon escaping).
- Capture `UpstreamBackendAddress` on provider dispatch and persist registry updates from both non-streaming and semantic-streaming response paths.
- Skip writes for looper requests, routing bypasses, non-2xx upstream responses, and requests without a session key or backend address.
- Namespace isolation follows the same user-scope derivation as the response cache.
**Modules affected:** `pkg/extproc`, `pkg/kvtransfer`
**Out of scope:** KVTransfer Coordinator, hint-header injection, metrics, and E2E transfer profiles (later steps in #2976).
## Test Plan
- [x] `go test ./pkg/kvtransfer/...`
- [x] `go test ./pkg/extproc/... -run 'KVAddress|AddressRegistry'`
- [ ] `make check CHANGED_FILES="src/semantic-router/pkg/extproc/processor_res_kv_address.go src/semantic-router/pkg/extproc/processor_res_kv_address_test.go src/semantic-router/pkg/kvtransfer/address_registry.go src/semantic-router/pkg/kvtransfer/address_registry_test.go src/semantic-router/pkg/kvtransfer/redis_address_registry.go"`
## Test Result
- `go test ./pkg/kvtransfer/...` — pass
- `go test ./pkg/extproc/... -run 'KVAddress|AddressRegistry'` — not run locally (linker missing `onnx_semantic_router` in this environment); extproc unit tests are included in CI
**Remaining risk:** Redis integration is covered by constructor/options wiring only; no live Redis E2E in this PR. Lookup path is exercised via `MemoryAddressRegistry` in unit tests.


AayushSaini101:
@drivebyer @Xunzhuo @FAUST-BENCHOU PTAL thanks

src/semantic-router/pkg/extproc/processor_res_kv_address.go
```diff
@@ -0,0 +1,98 @@
...
+		return r.kvAddressRegistryStore
+	}
+	if r.Config == nil {
+		r.kvAddressRegistryStore = kvtransfer.NoopAddressRegistry{}
+		return r.kvAddressRegistryStore
+	}
+	registry, err := kvtransfer.NewAddressRegistryFromResponseCacheStore(r.Config.SemanticCache)
+	if err != nil {
+		logging.Warnf("KV address registry unavailable: %v", err)
+		r.kvAddressRegistryStore = kvtransfer.NoopAddressRegistry{}
```
abhinav-m22, line 28: If Redis is unreachable the first time this runs (a restart, a network blip), the no-op gets stored here and every later call returns it, so the registry stays off until the router restarts or the config reloads. Is that intended, or would it be worth retrying after a cooldown?

src/semantic-router/pkg/extproc/processor_res_kv_address.go
```diff
@@ -0,0 +1,98 @@
...
+		return false
+	}
+	return strings.TrimSpace(ctx.UpstreamBackendAddress) != ""
+}
+
+func kvAddressRecordFromContext(ctx *RequestContext) (kvtransfer.AddressRecord, bool) {
+	if ctx == nil {
+		return kvtransfer.AddressRecord{}, false
+	}
+	sessionKey := routingSessionStateKey(ctx)
```
abhinav-m22, line 70: Small thought: when there's no `x-session-id`, Claude Code header or user ID, the session ID comes from `deriveSessionIDFromSemanticStructure` (the `cc-full-` prefix), which hashes every message so far and so changes each turn. The previous-model lookup misses for that traffic too, so I don't think a model switch could ever be detected there. Would it make sense to skip the registry write for `cc-full-` sessions, since those entries can't be looked up again?

src/semantic-router/pkg/extproc/processor_req_body_routing.go
```diff
@@ -90,6 +90,7 @@ func (r *OpenAIRouter) prepareProviderDispatch(
 	// redirect this request to a sibling modelRef, so RequestModel must be the
 	// final dispatch model, not the decision-selected one.
 	ctx.RequestModel = dispatch.logicalModel
+	ctx.UpstreamBackendAddress = dispatch.backendAddress
```
Xunzhuo, line 93: This is the configured backend address, not the pod that served the request. `ResolvePrimaryBackendForModel` explicitly only returns provider-shaping metadata and says Envoy owns endpoint load balancing; with multiple endpoints it also just picks the highest-weight config entry. Persisting this as `source_pod` can therefore send a later KV lookup to a service or to a different backend than the one holding the cache. Please capture the actual selected upstream endpoint from Envoy's response metadata (or defer the write until that identity is available) and cover a load-balanced backend in the tests.

src/semantic-router/pkg/kvtransfer/address_registry.go
```diff
@@ -0,0 +1,121 @@
...
+		escapeAddressKeyPart(namespace),
+		escapeAddressKeyPart(sessionID),
+	)
+}
+
+func escapeAddressKeyPart(part string) string {
+	return strings.ReplaceAll(part, ":", "%3A")
+}
+
+func validateAddressRecord(record AddressRecord) error {
```
Xunzhuo, line 91: Replacing only `:` is not injective: for example, raw session IDs `recipe::a:b` and `recipe::a%3Ab` both produce the same key. Since session IDs are request-derived, one session can overwrite another session's address record. Please use an encoding that also escapes the escape marker (or length-prefix/hash each component) and add a collision regression for literal `%3A`.

Xunzhuo (commented):
Thanks for building out this write path. I found two correctness issues that need to be addressed before this can land.

adaamko:
two things from the model side, both about identity, on top of xuns two.

`Model: ctx.RequestModel` is the routing alias, not a KV identity. whether a cached prefix can be mapped depends on the exact served artifact: weights revision, dtype, n_kv_heads, tokenizer. record the served model identity (the upstream response `model`, or the backends served-model name plus its revision) beside the alias, or the coordinator will pair a mapper with weights it was not fitted for and the failure will look like quality loss, not an error. same root as xuns source_pod point: the record has to name what actually served.

the record says where the KV was, not that it still is: pods evict, and the 1h ttl is unrelated to the pods cache lifetime. fine for a "last known" hint as scoped here, but the read path should treat a hit as a candidate to verify (pod up, same artifact, prefix still resident) and fall back to prefill, and the record could carry the prefix hash so the coordinator can check it maps the right prefix at all. skipping looper, bypass and non-2xx is the right set of exclusions.

Issue #3718, opened 2026-09-11
[Bug] plugin-request-mutations cannot reliably distinguish header overwrite from append


WUKUNTAI-0211:
### Proposed Workgroup

Evaluation & Quality

### Describe the bug

**Current status:** #3649 has merged. The request-mutation case and provider header recorder are now on `main`; at a3404940d the recorder still reduces repeated header names to one value, and the E2E assertion still compares a single string. The original reproduction below describes #3649's head; the same observation gap remains in current main.

The `plugin-request-mutations` E2E case exists to prove that `header_mutation.update` overwrites a header rather than appending to it, and it sets up that collision deliberately: the client sends `x-vsr-e2e-updated: client-value` while the decision's update sets the same name to `updated-by-router`.

The observation can lose that difference, and whether it does is decided outside the router. `RequestStore.record` in the mock simulator's `provider_boundary` module folds the observed headers into one value per name, and `validatePluginMutationProviderRequest` then compares a single string. When the provider receives the header twice — one of the shapes append produces — one of the two values is dropped before the assertion runs, and which one survives depends on arrival order rather than on the router's behavior.

### To Reproduce

Against #3649's head (`de68ab3`), using the simulator's own harness: POST a chat completion with a duplicated `x-vsr-e2e-updated` (`client-value`, then `updated-by-router`) under a test session id, then `GET /debug/last-request` for that session. The response reports a single value for that name, with no indication that two arrived.

Driving `RequestStore.record` directly with a `starlette` `Headers` built from the duplicated raw pairs shows the same collapse, and shows that the case's verdict is decided by arrival order:

| what the provider received | recorded value | case verdict |
| --- | --- | --- |
| two entries, client value first | `updated-by-router` | passes — append is invisible |
| two entries, router value first | `client-value` | fails |
| one comma-joined entry | `client-value,updated-by-router` | fails |
| one entry (correct overwrite) | `updated-by-router` | passes |

Only the first row is silent, so the case is not blind to append in general — it is blind to one of the shapes a repeated header can take, and nothing in the repo pins which shape reaches the provider. That is already enough for the verdict to turn on something other than the router. `Headers.items()` yields every raw pair, so the dict comprehension in `record` keeps the last one; `getlist` on the same headers returns both. The information is present and is discarded by the recorder, not by the framework.

### Expected behavior

[...]

PR #3649, opened 2026-09-09
[Test] Cover request mutation plugin effects end to end

P-Slark:
Related #3180

## Purpose

Add the first focused decision-plugin lifecycle slice for `wg/evaluation-quality`: direct provider-boundary coverage for `system_prompt`, `request_params`, and `header_mutation`.

The new `plugin-request-mutations` testcase uses a deterministic keyword route and proves the real effects of all three plugins in one chain:

- the configured system prompt reaches the provider;
- request penalties are removed and token/candidate limits are capped;
- request headers are added, overwritten, and deleted; and
- the client still receives a valid Chat Completions response.

The mock provider's bounded request observer now records only the test session header and `x-vsr-e2e-*` probe headers, avoiding authorization or arbitrary client-header capture. The runtime also sets Envoy's explicit overwrite action for `header_mutation.update`, with a unit regression test.

The testcase is registered in the baseline Envoy AI Gateway contract and its explicit CI selection list. This is one plugin-family slice and intentionally leaves #3180 open.

## Test Plan

- `cd e2e && go test ./testcases ./pkg/testmatrix`
- `cd e2e && go build -o <temporary-path>/e2e ./cmd/e2e`
- `cd tools/mock-vllm && python3.13 -m pytest tests/test_provider_contract.py -q`
- `python tools/agent/scripts/harness.py validate`
- `python tools/ci/validate_workflows.py`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 -shellcheck=`
- Hosted Envoy AI Gateway E2E: `make e2e-test-specific E2E_PROFILE=envoy-ai-gateway E2E_TESTS=plugin-request-mutations E2E_VERBOSE=true`

## Test Result

- E2E testcase and test-matrix packages passed.
- The E2E runner built successfully.
- All 10 mock-provider contract tests passed.
- Harness registry, workflow contract validation, `actionlint`, changed-file spelling, formatting, and diff checks passed.
- The live Docker/Kind/Envoy test was not run on the Windows development host; the explicitly selected hosted CI job will validate that deployment path.

---
<details>
<summary>Semantic Router PR Checklist</summary>

- [x] PR title begins with exactly one bracketed category
- [x] The title does not stack prefixes
- [x] The PR links accepted issue #3180, owned by `wg/evaluation-quality`
- [x] Commits in this PR are signed off with `git commit -s`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and remaining hosted-E2E validation

</details>

Issue #3180, opened 2026-08-31
[Test] Cover decision plugin lifecycle and composition end to end


Xunzhuo:
### Goal

Protect every supported decision plugin with end-to-end lifecycle and composition coverage that proves the plugin's real request or response effect, not merely that its decision was selected.

Parent Epic: #1519

### Proposed Workgroup

Evaluation & Quality

### Scope

- Derive the plugin inventory from `config.SupportedDecisionPluginTypes()` and keep aliases normalized to the canonical runtime type.
- Cover request mutation, immediate response, backend-dependent, response mutation, streaming, persistence, and observability behavior as applicable to each plugin.
- Assert actual effects: mutated provider requests, returned response shape/content, cache hit identity, selected tools, replay records, or typed blocking behavior. A selected-decision header alone is not proof that a plugin executed.
- Cover order, short-circuiting, safe failure, incompatible composition, and protocol preservation with a bounded set of representative plugin chains.
- Replace weak assertions in the current plugin-chain/config-variation cases, including unconditional cache success and inferred system-prompt execution.
- Land plugin families one by one using deterministic local fixtures.
- File implementation or public-API defects under the Workgroup that owns the plugin behavior and link the resulting regression case here.

### Non-goals

- Test every permutation of every plugin chain.
- Restore removed plugin surfaces solely for compatibility with stale tests.
- Treat a configuration parser test as runtime E2E coverage.
- Fold implementation fixes for unrelated plugins into one large test pull request.

### Completion signal

- Every canonical supported plugin maps to at least one meaningful runnable E2E contract in #2379's coverage matrix.
- Applicable success, short-circuit, failure, streaming, restart, and protocol-preservation paths are covered.
- Existing weak plugin assertions are replaced with direct behavior checks.
- No removed or unsupported plugin remains advertised by the maintained E2E catalog.
- Discovered implementation and API gaps have focused owning issues and regression coverage.

PR #3795, opened 2026-09-14
[Test] Cover tool plugin effects at the provider boundary

P-Slark:
Related #3180

## Purpose

Add the second focused decision-plugin lifecycle slice for `wg/evaluation-quality`: direct provider-boundary coverage for `tools` and `tool_selection`.

This follows the first #3180 lifecycle slice in #3649, which introduced the bounded provider request observer and covered system_prompt, 
equest_params, and header_mutation. This PR reuses that observer for the next plugin family.

The existing `tool-selection` testcase now proves the real upstream request effects instead of relying only on decision and debug headers:

- `tools` passthrough preserves the declared tool set;
- `tools` filtered mode applies an allow-list;
- `tools` none mode removes tools and clears `tool_choice`;
- `tool_selection` add and filter modes produce the expected provider-bound tool set; and
- a higher-priority PII `fast_response` prevents provider dispatch.

The testcase also requires the expected tool-selection strategy and confidence headers when those observability contracts apply. Deterministic keyword routes are added to the Envoy AI Gateway profile, and the existing baseline/PR-lane `tool-selection` registration continues to select this coverage.

This is one plugin-family slice and intentionally leaves #3180 open. It does not change the header observer representation tracked by #3718.

## Test Plan

- `cd e2e && go test ./testcases ./pkg/testmatrix`
- `cd e2e && go build -o <temporary-path>/e2e ./cmd/e2e`
- changed-file `golangci-lint` for `./pkg/testmatrix ./testcases`
- `make codespell-tracked`
- YAML parse and `yamllint` for `e2e/profiles/ai-gateway/values.yaml`
- structure, architecture, workflow, harness-registry, and AST security checks
- Hosted Envoy AI Gateway E2E: `make e2e-test-specific E2E_PROFILE=envoy-ai-gateway E2E_TESTS=tool-selection E2E_VERBOSE=true`

## Test Result

- E2E testcase and test-matrix packages passed.
- The E2E runner built successfully.
- Changed-file Go lint and `gofmt` passed.
- Codespell, YAML parsing/lint, structure, architecture, workflow contract, harness registry, and the HIGH-threshold AST security scan passed.
- `git diff --check` passed.
- The live Docker/Kind/Envoy test was not run on the Windows development host; the existing explicitly selected hosted CI job will validate that deployment path.

---
<details>

- [x] The PR links accepted issue #3180, owned by `wg/evaluation-quality`
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope, commands, and remaining hosted-E2E validation

</details>


P-Slark:
@Xunzhuo this is the second focused lifecycle slice for #3180, following #3649. When convenient, could you please review the provider-boundary coverage for the 	ools and 	ool_selection plugins? Thanks.

e2e/testcases/tool_selection_e2e.go
```diff
@@ -186,21 +233,103 @@ func runToolSelectionCase(
...
+		observed, model, err := lookupShortCircuitDispatch(ctx, backend, sessionID)
+		if err != nil {
+			return err
+		}
+		if observed {
+			return fmt.Errorf("fast_response request reached provider model %q", model)
+		}
+		return nil
+	}
+	if tc.ExpectProviderTools == nil && !tc.ExpectNoProviderTools && !tc.ExpectToolChoiceAbsent {
```
Xunzhuo, line 285: **Exercise every tool-selection case at the provider boundary**

Impact: Every add case and the ordinary filter case leave all three provider expectations unset, so this return skips the provider request entirely. The add inputs also already contain the full tool catalog, making addition a no-op; a broken add/filter implementation can therefore pass while the result reports provider-boundary verification.

Required change: Give every add/filter case an exact provider tool-set or absence expectation, make each add input omit the tool that should be added, and fail rather than skip when a claimed provider-boundary case has no expectation.

Xunzhuo (commented):
Thanks for carrying the plugin checks to the provider observer. One coverage gap still lets the advertised add/filter behavior pass without being exercised.

e2e/testcases/tool_selection_e2e.go
P-Slark: Fixed in 7b91013df. Every add/filter case now defines an exact provider-bound tool set; add inputs omit the tool being selected, the ordinary filter case must remove an irrelevant tool, and a case with no provider expectation fails instead of skipping observation.

Xunzhuo (dismissed):
The provider-boundary assertions now verify the exact tool set for both add and filter cases, including the missing-tool input and safe short-circuit path. The exact-head E2E checks pass.

wilsonwu (dismissed):
LGTM

P-Slark:
The latest hosted E2E failures were caused by the ai-gateway profile inheriting newly migrated Vela/32K-window classifier defaults while its acceptance datasets remain calibrated to the legacy mmBERT classifiers. Commit 6205b6105 explicitly pins the domain, PII, and jailbreak models (including the legacy input budget) and adds a profile regression test.\n\n@xunzhuo @wilsonwu, this code fix triggered stale-review dismissal.  please re-review this focused profile compatibility change when convenient.

Xunzhuo (dismissed):
The tests now inspect provider-bound tools for passthrough, removal, allow-list filtering and semantic selection, including absence after the PII short circuit. The updated retention fixture reaches its intended route. Current-head Envoy AI Gateway CI ran all 36 cases successfully, including the 10 tool-selection subcases and retention-directive, and Router Tools passed.

wilsonwu (dismissed):
LGTM

P-Slark:
@mergifyio queue

P-Slark:
@Xunzhuo @wilsonwu all actual CI checks are green, including Envoy AI Gateway and PR Gate. The only red status is Mergify's queue rule because my requeue command was rejected for lacking repository write permission. Could one of you please tick **Requeue this pull request** in the Merge Queue Status comment (or invoke the permitted queue action)?

P-Slark:
Updated this branch to current main (through #3959) and removed the temporary classifier pins and prompt workarounds that #3944 superseded. The remaining diff is limited to the provider-boundary tools/tool_selection coverage for #3180. Targeted profile/testcase/testmatrix tests and vet pass locally. @Xunzhuo @wilsonwu, please re-review once the refreshed CI completes.

FAUST-BENCHOU:
This now exercises the contract at the right boundary. The add cases start without the selected tool, filter cases assert exact provider-visible sets, and missing provider expectations no longer let a case silently skip the behavior it claims to verify. Nice focused E2E slice. LGTM.

Xunzhuo (approved):
The provider-boundary assertions cover tool passthrough, filtering, removal, and semantic selection. All ten tool cases pass in the current Gateway run, along with the full 36-case profile and E2E unit suite.

wilsonwu (approved):
LGTM

Issue #3794, opened 2026-09-14
[Feature] add support for additional containers to semantic-router deployment


upodroid:
### Goal

I'm running a sidecar container that runs https://huggingface.co/nvidia/prompt-task-and-complexity-classifier as a container next to semantic router but I can't pass additional containers using the helm chart.

Please amend the chart to add the extraContainers parameters and make it behave like the existing extraVolumes parameter.

### Proposed Workgroup

Developer Experience & Ecosystem

### Scope

helm chart

PR #3796, opened 2026-09-14
[Bug] semantic router supports hot reloading so don't rollout new revision for config changes

upodroid:

Related: #3794

## Purpose

Right now, when we update the config of SR, the Helm chart amends the annotations and triggers a new revision of the deployment. This is no longer required because SR supports hot reloading and reloads when the config file changes.

For example, in #3794, I'm running a classifier as an additional container, which causes this container to restart for no functional reason.

## Test Result

When the ConfigMap is changed, this log line appears, and the SR reloads.

`{"level":"info","ts":"2026-09-14T04:47:13.661","caller":"logging.go:264","msg":"config_reload_scheduled","delay_ms":300,"component":"extproc","file":"/app/config/config.yaml","event":"CHMOD"}`

---
<details>

</details>


deploy/helm/semantic-router/templates/deployment.yaml
```diff
@@ -49,7 +49,6 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
Xunzhuo, line 52: **Keep updates observable with subPath mounts**

Impact: Both `config.yaml` and `tools_db.json` are mounted from this ConfigMap with `subPath`, so Kubernetes does not propagate later ConfigMap updates into those files. Removing the checksum means a config-only `helm upgrade` leaves existing pods on the old configuration, and the file watcher never receives the change.

Required change: Keep the checksum rollout trigger, or first switch the chart to a mount/update mechanism that reliably propagates ConfigMap changes and add an upgrade regression test proving the running pod observes the new config without a rollout.

Xunzhuo (commented):
Thanks for tackling the rollout behavior. One configuration-update gap still needs to be closed before this can land.

deploy/helm/semantic-router/templates/deployment.yaml
upodroid: I didn't understand your suggestion.

Are you saying SR doesn't hot-reload the tools_db.json changes? Also, changing any field other than replicas in the Deployment manifest will trigger a new rollout.

PR #3808, opened 2026-09-15
[Test] Bind Response API text fixture to its model catalog

cr-zhao:
## Purpose

Fix the `response-api-image-file-id` CI failure seen in #3311 ([failed job](https://github.com/vllm-project/semantic-router/actions/runs/34917578526/job/104219032535)): the test expects HTTP 400 for an image sent to the text-only GPT-OSS model, but receives HTTP 200.

The fixture names `openai/gpt-oss-20b` without binding its catalog card. After #3781 stopped treating default capabilities as declarations, this sparse custom model remains unannotated and image requests are forwarded to the mock backend. Explicitly bind the built-in catalog card so its declared text-only capabilities enforce the existing rejection assertion.

## Test Plan

Run the changed-file checks and existing capability declaration tests; rerun the `response-api` E2E profile in CI.

## Test Result

- `make check CHANGED_FILES=e2e/profiles/response-api/values.yaml`: passed.
- From `src/semantic-router`, `go test ./pkg/config -run 'TestSparseCustomModelCardCarriesNoDeclaredCapabilities|TestOperatorModelCardCapabilitiesAreDeclared' -count=1`: passed.
- `git diff --check`: passed.
- Local `make verify PROFILE=response-api` was stopped during the image build due to memory pressure; E2E validation remains pending in CI.


e2e/profiles/response-api/values.yaml
```diff
@@ -11,6 +11,7 @@ config:
       model: openai/gpt-oss-20b
     models:
       - name: openai/gpt-oss-20b
+        catalog: openai/gpt-oss-20b
```
FAUST-BENCHOU, line 14: Binding this catalog card also activates the GPT-OSS always-on reasoning contract. The two routes in this fixture still set `use_reasoning: false`, so the Router rejects the configuration at startup and the exact-head response-api E2E never reaches the image assertion. Please align those model refs with the catalog contract, or use another text-only catalog model, and rerun the target E2E.

Issue #3671, opened 2026-09-09
[Feature] Complexity backend follow-ups deliberately left out of #3542


theohsiung:
### Goal

Collect the follow-ups left out of #3542 deliberately, so they are tracked
rather than remembered. Each was raised in review or found while building it,
each was judged not worth another round on that PR, and each has a reason
recorded here that a future reader can disagree with.

None of these breaks routing today.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

**1. `threshold: 0` written alongside an explicit boundary pair is not refused**

`EffectiveBoundaries` treats threshold as absent when it is zero
(`complexity_boundaries.go`), because `ComplexityRule.Threshold` is a `float32`
with no presence. The CRD refuses the combination via
`has(self.threshold)`, so a `kubectl apply` is rejected while the same YAML
loads. The behaviour when it does load is already the sensible one - the pair
wins and threshold is ignored - so the gap is a missing diagnostic, not a wrong
verdict.

Cheapest fix found: a `ComplexityRule.UnmarshalYAML` that records the key's
presence, so the check becomes `set || value != 0`. Backward compatible and
needs no test changes. Changing the field to `*float32` also works but touches
the DSL compiler and decompiler, the operator conversion and eleven test
literals.

**2. `label_distribution.v1` has no deployment-level (E2E) coverage for complexity**

`e2e/profiles/complexity-remote-backend` pins `contract: score.v1`, and
`contract` is a module-level setting, so covering the other contract means a
second profile with its own mock endpoint.

What that would add over what exists is narrow. The transport and its
name-based alignment are shared with categories and already run in a cluster
through that profile; `<rule>:<verdict>` reaching a decision is proven in a
cluster by the score profile; and the label path's own contract - alignment by
name, refusal when a class is not named for a verdict - is now pinned in
process by `complexity_label_contract_test.go`. The remaining increment is "the
config starts in Kubernetes", on scaffolding identical to the score profile's.

Weigh that against the cost: an E2E profile is 30-80 minutes of CI on every
future PR that touches e2e, permanently.

**3. Dashboard cannot express the boundary rules it displays**

The rule editor exposes the four boundary fields with descriptions, but only
required-field validation runs, so an incomplete or mutually exclusive
combination can be saved and is refused later by the Router or the CRD.

Adding the rule to the form would make it the third implementation of one
constraint - Router validator, CRD CEL, and TypeScript - with nothing keeping
[...]

PR #3809, opened 2026-09-15
[Bug] Refuse a written threshold: 0 alongside a complexity boundary pair

Vaivaswat2244:
Related #3671 (item 1 of the follow-up list; the issue stays open for the other items).

## Purpose

`ComplexityRule.Threshold` is a `float32` with no presence, so `EffectiveBoundaries` treated a written `threshold: 0` as an absent key. The CRD refuses `threshold` alongside a boundary pair through `has(self.threshold)`, so the same YAML was rejected by `kubectl apply` and accepted by the Router. The verdict when it loaded was already the sensible one (the pair wins), so this closes the missing diagnostic rather than changing routing.

The fix is the one written down on the issue: record the key's presence.

- `ComplexityRule.UnmarshalYAML` decodes the rule as before, then probes the raw map for a `threshold` key and records `ThresholdSet` (a null value counts as absent, as it does for the CRD). The field is tagged out of YAML and JSON: presence belongs to the document, not the rule.
- `EffectiveBoundaries` checks `set || value != 0` before refusing a pair rule for also stating a threshold.
- The DSL compiler records presence the same way, so `SIGNAL complexity x { threshold: 0 hard_above: ... }` is refused on that path too. The decompiler keeps a written zero as written so round trips stay lossless.
- `Threshold` gains `omitempty`. Without it every pair rule marshalled by the operator (ConfigMap) or the DSL emitter would grow a `threshold: 0` and be refused on reload. Zero and omitted already mean the same thing to the local path.
- `ComplexityRule.MarshalYAML` writes the rule exactly as before, except when a zero threshold was written: then the key is restored. Without this, `EmitYAML` dropped the written zero and the emitted document reloaded as an accepted pair (review finding below). A pair with no threshold still emits without one.

Presence tracking was chosen over widening the field to `*float32`, which would touch the DSL compiler and decompiler, the operator conversion and eleven test literals. The flag is exported because the DSL round-trip tests compare configs with `go-cmp`, which panics on unexported fields.

Modules: Router config (`pkg/config`), DSL compiler and decompiler (`pkg/dsl`), operator conversion test, complexity tutorial. Workgroup: Router Models & Inference Runtime.

## Test Plan

New tests:

- `pkg/config`: a written `threshold: 0` (and `0.0`) beside either pair is refused after YAML load; presence follows the key (omitted and null are absent, `0` and `0.1` are present) and a lone written zero still resolves as the shorthand; a marshalled pair rule carries no `threshold` and reloads cleanly.
- `pkg/dsl`: the compiler records a written zero and the rule is refused; a written zero survives decompile and recompile.
- operator: a CRD threshold arrives with presence recorded through yaml.v3, and a pair rule marshals without a threshold.
- emit and reload: `EmitYAML` of a written zero beside a pair reloads through the config loader and is refused with the same error; a pair with no threshold reloads cleanly. The same written-zero round trip is covered on the yaml.v2 loader (`pkg/config`) and the yaml.v3 ConfigMap path (operator).

```bash
cd src/semantic-router && go test ./pkg/config/ ./pkg/dsl/
cd deploy/operator && go test ./controllers/
make config-schema-check
.venv-agent/bin/pre-commit run --files <changed files>
```

## Test Result

All of the above pass on top of current `main` (rebased after the review, one conflict in the decompiler resolved by keeping the upstream `prototype_scoring` line). `go vet` passes for `pkg/classification`, `pkg/apiserver`, `pkg/services`, `pkg/k8s` and `pkg/extproc`; their test binaries do not link on my host because the local Rust binding libraries are stale, which is unrelated to this change and left for CI.

Not covered: the Dashboard rule editor, which is issue item 3.

---
<details>

</details>


FAUST-BENCHOU (changes requested):
pls resolve conflicts

src/semantic-router/pkg/config/signal_config.go
```diff
@@ -454,7 +454,12 @@ type ComplexityRule struct {
 	// Threshold is the symmetric shorthand, kept because the local margin is
 	// signed and centred on zero: hard above +threshold, easy below
 	// -threshold. Mutually exclusive with the explicit pair below.
-	Threshold float32              `yaml:"threshold"`
+	//
+	// Zero and omitted mean the same thing to the local path, so a zero is
+	// not written back out: a rule that states a pair must not grow a
+	// `threshold: 0` on its way through the operator or the DSL emitter and
+	// then be refused for stating both.
+	Threshold float32              `yaml:"threshold,omitempty"`
```
Xunzhuo, line 471: [P2] Preserve explicit zero through the YAML emission path

`EmitYAML` accepts `SIGNAL complexity r { threshold: 0 hard_above: 0.85 easy_below: 0.60 }` (with each field on its own line). The compiled rule has `ThresholdSet=true` and `EffectiveBoundaries` rejects it, but `omitempty` drops the written zero during emission. Reloading the output sets `ThresholdSet=false`, so the same pair is now accepted. The public DSL-to-YAML path therefore bypasses the new mutual-exclusion check. Reject this combination before emitting, or preserve an explicitly written zero during serialization, and add an emit/reload regression while keeping threshold-absent pairs valid.

Xunzhuo (changes requested):
The focused config/DSL tests and operator ConfigMap test pass, but a full YAML emit/reload regression still reproduces a bypass of the explicit-zero rejection. Please fix that serialization path and add the round-trip regression. A deployed router runtime was not exercised.

src/semantic-router/pkg/config/signal_config.go
Vaivaswat2244: Confirmed and fixed in 62e6ad1e. `omitempty` was dropping the written zero on emission, so the emitted document reloaded with the threshold absent and the pair passed.

The fix preserves the zero rather than rejecting at emit, so the loader stays the one place that refuses the combination: `ComplexityRule.MarshalYAML` writes the rule exactly as before, except when a zero was written, in which case the key is restored. A pair with no threshold still emits without one.

Regression added on three paths: `TestEmitYAMLKeepsWrittenZeroThresholdRefusedOnReload` (EmitYAML, then reload through `config.ParseYAMLBytes`, refused with the same error; the threshold-absent pair reloads cleanly), `TestComplexityRule_WrittenZeroThresholdSurvivesMarshal` (yaml.v2 loader) and `TestOperatorComplexityRuleWrittenZeroThresholdIsSerialised` (yaml.v3 ConfigMap path). Branch is also rebased on current main with the decompiler conflict resolved.

Vaivaswat2244:
Fixed: conflicts resolved on current main, written zero now preserved through MarshalYAML so EmitYAML output is refused on reload (regressions added on the DSL, yaml.v2 and yaml.v3 paths), and the shadowed err lint. CI is green on 8026334e.

Ready for review @Xunzhuo @FAUST-BENCHOU @AayushSaini101 


Also I'm having to update branch somewhat too often, maybe because of prs getting merged often.  A merge queue would rebase approved PRs onto main and run the required checks once at merge time, removing that loop. Suggested: squash merges, group up to 5 PRs, required checks limited to the PR-lane set so throughput keeps up with the merge rate.

Is one of the community meetings (second or fourth Wednesday per the README) the right place to bring this up, or is an issue enough? Happy to join and present it either way.

Xunzhuo (approved):
A written zero threshold is now preserved through YAML and DSL round trips, while pair-only rules remain valid. The config/DSL regressions, operator checks and remote-complexity routing E2E pass on this head; the previous serialization issue is resolved.

Vaivaswat2244:
Anything stopping the merge for this? @FAUST-BENCHOU @Xunzhuo

PR #3813, opened 2026-09-15
[Bug] Accept xAI and Groq reply fields in the chat codec

### LINKED ISSUE #3812: [Bug] xAI and Groq replies fail with 502 through the router
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

Routing to xAI or Groq through their OpenAI-compatible endpoints returns `502`. The router decodes provider replies strictly, and both providers add fields the Chat Completions response schema does not list:

- xAI: `usage.cost_in_usd_ticks`, `usage.num_sources_used`
- Groq: `x_groq`, `usage_breakdown`, `usage.queue_time`, `usage.prompt_time`, `usage.completion_time`, `usage.total_time`, and `service_tier: "on_demand"`

Router log: `neutral_response_decode_failed ... invalid_upstream_json: upstream response JSON contains a non-canonical field`.

### To Reproduce

1. Add a model with `base_url: https://api.x.ai/v1` (or `https://api.groq.com/openai/v1`) and `provider: openai`.
2. Send any chat completion to that model.
3. The router returns `502` with `x-vsr-response-path: error`, although the provider answered normally.

### Expected behavior

The reply is accepted. Provider-only fields are dropped with a diagnostic, the same way the existing vLLM extensions are handled.

### Impact

xAI and Groq cannot be used as backends on current images, and there is no config workaround.

### Additional context

Proposed fix: model these fields in the chat response, stream chunk and usage wires, report them as dropped diagnostics, and update the closed field inventories. Any other unknown field is still rejected.

abhinav-m22:
Closes #3812

## Purpose

xAI and Groq replies were rejected with `502` because the strict chat decoder did not know their extra fields. This models them and drops them with a diagnostic, like the existing vLLM extensions. Any other unknown field is still rejected.

- `usage`: `cost_in_usd_ticks`, `num_sources_used` (xAI), `queue_time`, `prompt_time`, `completion_time`, `total_time` (Groq)
- response and stream chunk: `x_groq`
- response: `usage_breakdown` (Groq)
- service tiers: `on_demand` and `performance` (Groq)
- final stream chunks report the same `usage` omissions as buffered responses, as `stream.usage.*`
- closed inventories updated: response and usage inventory tests, chat nested inventory, golden fixtures 016 and 039, and the mock-vllm schema contract

## Test Plan

- `go test ./pkg/protocolcodec/... ./pkg/llmprotocol/...`
- New tests in `codec_openai_chat_provider_extensions_test.go`: xAI and Groq replies decode and translate, their fields show up as dropped diagnostics, a Groq stream chunk decodes, and an unknown field still fails.
- `pre-commit run` on the changed files

## Test Result

- All tests pass, including the closed inventory and golden fixture tests.
- Live check, calling each provider directly and comparing its reply with the new field lists: xAI `grok-4.20-0309-non-reasoning` passes for a plain reply and a tool call (`service_tier: "default"`). Groq `openai/gpt-oss-120b` also sends `usage_breakdown` and `service_tier: "on_demand"`, both covered by the second commit.
- Not yet run against the providers through a router image built from this branch.


src/semantic-router/pkg/protocolcodec/stream_chat.go
```diff
@@ -178,6 +179,12 @@ func (decoder *chatStreamDecoder) appendProviderChunkDiagnostics(
 			"stream.kv_transfer", "provider KV-transfer metadata is not model output",
 		)
 	}
+	if len(chunk.XGroq) > 0 {
```
FAUST-BENCHOU, line 182: The shared `chatUsageWire` now accepts the xAI and Groq usage extensions in final stream chunks, but this streaming path only reports `x_groq`, KV, and moderation omissions. A final chunk containing `usage.cost_in_usd_ticks` or Groq timing fields is therefore accepted and silently loses those fields, unlike the non-streaming path. Could we reuse the usage omission checks here and add a final-usage-chunk regression test?

abhinav-m22: @FAUST-BENCHOU Good catch, thanks. Final stream chunks now run through the same usage omission list as buffered responses: the inventory moved into a shared `chatUsageFieldOmissions(wire, prefix)` helper, and the stream path reports them as `stream.usage.*` next to the existing `stream.kv_transfer`, `stream.x_groq` and `stream.moderation` entries. Added `TestChatStreamReportsProviderUsageInFinalChunk`, a final usage chunk carrying the Groq timings and xAI `cost_in_usd_ticks`. Rebased on the updated branch and `go test ./pkg/protocolcodec/...` passes.

Xunzhuo (approved):
The xAI/Groq accounting fields are accepted without changing token totals, and buffered replies and final stream chunks consistently report omitted provider metadata. The relevant codec regressions and streaming, Anthropic, and Responses integration checks pass on this head.

wilsonwu (approved):
LGTM

PR #3814, opened 2026-09-15
[Bug] Let the config write API patch the ConfigMap directly on Kubernetes

### LINKED ISSUE #3688: [Bug] The runtime config API cannot write on any shipped Kubernetes deployment

ankit373:
Fixes #3688.

## The problem

Every shipped Kubernetes manifest mounts `config.yaml` as a read-only ConfigMap `subPath`. `writeConfigAtomically` (Router apiserver and Dashboard backend both have a copy) writes there directly, so `PUT`/`PATCH /config/router`, rollback, the Dashboard deploy flow, and KB persistence all fail on every one of those deployments.

## The fix

When a deployment declares a write target (three new `VLLM_SR_K8S_CONFIGMAP_*` env vars), both copies of `writeConfigAtomically` patch that ConfigMap through the Kubernetes API instead of touching the local file. No target declared, which is every non-Kubernetes deployment today, keeps the exact same local-file behavior.

`pkg/k8s/configmap_writer.go` is the new primitive: reads the ConfigMap, refuses to touch it if a controller owns it (checked via `ownerReferences[].controller`, the same field `controllerutil.SetControllerReference` sets, so an Operator-managed ConfigMap does not get silently reverted on the next reconcile), retries on a conflicting concurrent write.

## Scope

The write lands durably in the ConfigMap and takes effect on the next pod restart, not live. Making it live needs moving off `subPath` so kubelet's periodic sync can update the mounted file, but that would also make `/app/config` a fully ConfigMap-owned read-only directory, and config backups, KB runtime state, and KB validation temp files currently live writably under that same path outside the two mounted files. That needs its own pass to relocate those first, so it is a follow-up rather than bundled here.

## Manifests

All nine files that mount `config.yaml` (Helm x2, istio, kserve, openshift x3, the openshift dashboard, the kubernetes observability dashboard) get a namespaced Role scoped to that one named ConfigMap by `resourceName`, a ServiceAccount, and the two env vars (namespace via the downward API). istio's and openshift's kustomizations generate that ConfigMap with a content hash suffix by default, which would mint a new ConfigMap on every apply and orphan whatever the runtime API had just written to the old one, so those get `disableNameSuffixHash`. The OpenShift dashboard's `DASHBOARD_RUNTIME_CONFIG_WRITABLE` flag, which was off with a comment explaining ConfigMaps are read-only there, is flipped back on; the recipe store's separate writable flag is untouched. `deploy/kserve/deploy.sh` and `deploy/openshift/deploy-to-openshift.sh` apply individual files rather than their kustomization, so the new RBAC files are wired into those scripts directly too.

## Verification

- `go vet` and the new `configmap_writer` tests, including a mutation test on the ownership-refusal guard
- Dashboard module's test suite runs and passes, including the new K8s-routing cases (no cgo dependency there, so these actually execute; the Router side can't link locally in this environment due to missing prebuilt Rust bindings, verified with `go vet` plus an isolated harness proving the same logic instead)
- Every rendered manifest validated against the real Kubernetes API schema with `kubeconform`: `helm template`, `kubectl kustomize` for istio/kserve/openshift/observability, and the OpenShift Template's parameter substitution resolved by hand since `oc` was not available
- Confirmed `disableNameSuffixHash` against actual `kubectl kustomize` output on a throwaway kustomization, not just documentation


FAUST-BENCHOU (changes requested):
make sure ci green thanks

ankit373:
Pushed a fix for the CI failures:

- `go.mod`/`go.sum` in both `src/semantic-router` and `dashboard/backend` were left non-tidy (the new `k8s.io/api` import needed to drop its `// indirect` marker).
- Moved `ConfigMapWriter` into its own `pkg/k8s/configwriter` leaf package, since the Dashboard's Docker build only copies a curated subset of `src/semantic-router/pkg` and importing the full `pkg/k8s` package would have pulled in `controller-runtime` and the CRD API types for no reason.
- `entrypoint.sh`'s arbitrary-non-root-UID check was unconditionally forcing `DASHBOARD_RUNTIME_CONFIG_WRITABLE=false`, which made enabling it in the OpenShift manifest a no-op and broke the container-contract test. It now only forces that off when no Kubernetes ConfigMap target is configured, since that write path never touches the local filesystem the rest of the branch guards against.

Verified locally: `go build`/`go vet`/`go mod tidy` clean on both modules, the affected unit tests pass, and I reproduced the Dashboard Docker build's exact file whitelist in a scratch directory to confirm the `go build` that was failing in CI now succeeds. Will keep watching the CI run.

src/semantic-router/pkg/apiserver/route_router_config_update.go
```diff
@@ -514,8 +517,56 @@ func restoreSourceConfig(sourcePath string, previousData []byte) error {
...
+// deployment has opted in.
 func writeConfigAtomically(configPath string, yamlBytes []byte) error {
+	if target, ok := configwriter.ConfigMapTargetFromEnv(); ok {
+		writer, err := resolvedConfigMapWriter()
+		if err != nil {
+			return fmt.Errorf("config write target is declared but no Kubernetes client is available: %w", err)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), configMapWriteTimeout)
+		defer cancel()
+		if err := writer.Write(ctx, target, yamlBytes); err != nil {
```
Xunzhuo, line 602: **[P1] Propagate the ConfigMap write before reporting runtime activation**

This returns after updating the ConfigMap, but the shipped deployments still mount config.yaml with subPath, so Kubernetes never propagates the new bytes into the running pod. The caller then hashes that unchanged file in waitForRuntimeConfigActivation and can report active because the old hash already matches the active runtime. Later reads and PATCH operations also continue from the old document. Please provide an updateable mount or an explicit ConfigMap watch/apply path, bind activation to the submitted document, and cover write/readback/activation through the real Kubernetes deployment.

Xunzhuo (changes requested):
The ConfigMap write still leaves the running pod on the old document and can report that old runtime as active. Please connect persistence to runtime activation and cover write/readback/activation in the Kubernetes deployment. I traced the code and manifests; I have not run a cluster regression.

src/semantic-router/pkg/apiserver/route_router_config_update.go
ankit373: Confirmed, real and reproducible: waitForRuntimeConfigActivation hashes the mounted config.yaml, but a Kubernetes-target write never touches that read-only mount. So it hashes an unchanged file, finds a coincidental match with the still-running config's hash, and reports "active" for a write the process never saw.

Fixed in 88405380 by checking for a declared ConfigMap target upfront and reporting an honest `persisted` status with the hash of the document that was actually written, instead of polling a file that structurally cannot change. Applied the same fix to knowledgeBaseActivationStatus (KB create/update/delete have the identical pattern) and to GET /api/v1/config/hash, which had the same blind spot for its source and generated-runtime hashes; it now reads the live ConfigMap via a new ConfigMapWriter.Read instead of the stale mount.

This doesn't make the write live without a restart. That's the bigger change you're pointing at (an updateable mount or an explicit watch/apply path), and it's still out of scope here for the reason in the PR description: the mount is subPath, so switching it away from that to let kubelet sync live also touches config backups, KB runtime state, and KB validation temp files that share that same directory today, and that needs its own pass. What this fix does is stop the API from claiming activation it can't back up, so a caller (or the Dashboard) gets `persisted` and a restart requirement instead of a false `active`.

Added a regression test (TestWaitForRuntimeConfigActivationReportsPersistedNotActiveOnKubernetesTarget) that seeds the exact coincidental-hash scenario and asserts persisted, not active. Also extended the ConfigMapWriter tests for the new Read method. I don't have a real cluster to run this against either; local verification is go vet + go build (this package needs the compiled candle-binding Rust library to link a runnable test binary, which isn't available in this sandbox) plus the fake-clientset unit tests.

src/semantic-router/pkg/apiserver/kb_persistence.go
```diff
@@ -103,7 +104,16 @@ func persistConfigAndSync(
...
-func (s *ClassificationAPIServer) knowledgeBaseActivationStatus(runtimePath string, successStatus int) (knowledgeBaseActivation, int) {
+//
+// generatedDocument is the document persistConfigAndSync just wrote. On a
+// Kubernetes ConfigMap target, runtimePath is a read-only mount that write
+// never touches, so hashing it here would find the old, unrelated match and
+// falsely report "active" (review on #3814); report "persisted" instead,
+// hashing the document that was actually written.
+func (s *ClassificationAPIServer) knowledgeBaseActivationStatus(runtimePath string, generatedDocument []byte, successStatus int) (knowledgeBaseActivation, int) {
+	if _, ok := configwriter.ConfigMapTargetFromEnv(); ok {
+		return knowledgeBaseActivation{ActivationStatus: "persisted", GeneratedRuntimeHash: configDocumentETagHash(generatedDocument)}, http.StatusAccepted
```
Xunzhuo, line 117: **[P2] Add the new KB status to its response contract**

A ConfigMap-target KB create, update, or deletion now returns HTTP 202 with `activation_status: "persisted"`, but `addKnowledgeBaseActivationResponses` and the generated OpenAPI schemas for all three routes restrict that field to `pending` and describe polling until activation. The new successful response therefore violates the published schema. The dashboard activation helper also resolves immediately for any non-`pending` value, so it cannot explain this restart-required outcome. Please update the source/generated KB response contract, handle `persisted` explicitly in its callers, and add a regression for this response path.

Xunzhuo (changes requested):
The new persisted result avoids falsely claiming activation, but the KB response now falls outside its published API contract. Please update the KB response schema and generated artifacts together with explicit restart-required handling in its callers. This is based on the response contract and caller paths; I have not run a live cluster regression.

ankit373:
Status update on the CI failures:

- go.mod/go.sum tidiness (src/semantic-router, dashboard/backend, and later perf) — fixed.
- The Dashboard Docker build's file whitelist missing pkg/k8s/configwriter — fixed (moved ConfigMapWriter into its own leaf package).
- The false-"active" activation status Xunzhuo caught — fixed and replied to inline.
- Two failures on "Core quality / Full pre-commit checks" were an environment issue, not code: the perf-check benchmark step couldn't find `models/mom-embedding-pro` on the runner (missing model cache), unrelated to anything in this diff and reproduced identically across two separate runs.

The last two pushes (retrying the model-cache flake) don't seem to have started a workflow run at all, looks like they're waiting on approval. Could someone approve the run when you get a chance? Everything else should be green at that point, happy to keep iterating if something new turns up.

src/semantic-router/pkg/apiserver/kb_persistence.go
ankit373: Fixed in caffb5f6: added "persisted" to the documented enum for this 202 response, with the description covering both the poll-until-active case and the Kubernetes-target restart-required case. Also fixed the Dashboard's `waitForKnowledgeBaseActivation` helper, which was treating any non-"pending" status (including the new "persisted") as an immediate success with nothing surfaced to the caller; it now raises the same restart-required message the timed-out-pending path already uses, instead of silently returning as if the change were live.

ankit373:
Merged upstream/main to resolve the conflict with #3858 (the API-contract unification): reconciled the activation-status logic so both that PR's new "failed" status (with attempt-tracking) and this PR's "persisted" status coexist correctly, moved the KB response documentation onto the new declarative response builder, and regenerated perf/go.sum. PR now shows mergeable. Local go build/go vet are clean across the whole module.

The CI run for this push doesn't seem to have started yet, might need another approval click. Let me know if there's anything else you'd like changed.

ankit373:
CI ran after the merge but Quality/Checks failed on a stray double blank line gofmt didn't like in route_router_config_update.go (leftover from the merge). Fixed in b4b285e0. Everything else that ran passed; the remaining test jobs were still queued when I checked.

PR #3821, opened 2026-09-15
[Bug] Require explicit catalog binding for reasoning validation

### LINKED ISSUE #3816: [Bug] CLI infers a reasoning contract from a custom model name
## Describe the bug

On main `c1e00bb0f`, the CLI infers a reasoning family from `providers.models[].name` even when the model omits both `catalog` and `reasoning`. The Router correctly treats this configuration as a custom pass-through model.

This is visible in first-run Dashboard onboarding: connecting a vLLM backend named `openai/gpt-oss-20b` and choosing From scratch produces a valid configuration, activates successfully, and serves a real completion. Running `vllm-sr config validate` on that same active YAML fails:

```text
[routing.decisions.default-route.modelRefs[0]] use_reasoning=false is not supported by this always-on reasoning family
```

## To reproduce

```yaml
version: v0.3
providers:
  models:
    - name: openai/gpt-oss-20b
      backend_refs:
        - name: primary
          provider: vllm
          endpoint: localhost:8000
          protocol: http
routing:
  decisions:
    - name: default
      priority: 1
      rules:
        operator: AND
        conditions: []
      modelRefs:
        - model: openai/gpt-oss-20b
          use_reasoning: false
```

Validate this file with the CLI and compare with Router parsing/startup.

## Expected behavior

Match the documented model identity contract: `name` is a Router-local alias; catalog reasoning is inherited through an explicit `catalog` binding. Models without `catalog` or `reasoning` retain custom pass-through behavior even if their names happen to match a built-in model ID. Explicit catalog and reasoning bindings must keep their validation.

## Proposed Workgroup

Xunzhuo:
Closes #3816

## Purpose

A Dashboard-created custom model named `openai/gpt-oss-20b` activates and serves requests successfully, but CLI validation incorrectly inherits the built-in model card's always-on reasoning requirement from the name alone.

Use only an explicit `catalog` binding to inherit reasoning metadata, matching the Router's canonical parser. Model names remain local aliases; custom models without a reasoning contract remain valid. Explicit catalog bindings retain their validation.

Owned by Developer Experience and Ecosystem. This PR is stacked on the Vela input-policy fix so the shared mainline materialization regression is corrected in its test baseline; the diff contains only reasoning validation and its tests.

## Test Plan

- `PYTHONPATH=src/vllm-sr python -m pytest src/vllm-sr/tests/test_reasoning_controls.py -q`
- `make impact ENV=amd CHANGED_FILES="src/vllm-sr/cli/validator_reasoning.py src/vllm-sr/tests/test_reasoning_controls.py"`
- `make check CHANGED_FILES="src/vllm-sr/cli/validator_reasoning.py src/vllm-sr/tests/test_reasoning_controls.py"` with a Python 3.13 harness environment.

## Test Result

All 23 reasoning tests passed, including actual `config validate` invocation for named and inline recipes with and without explicit catalog binding. The full CLI check passed (247 passed, 5 skipped in the final pytest group), along with all changed-file hooks. The original mismatch was reproduced during real AMD Dashboard onboarding and routed inference. The committed CLI was then run on the AMD host against that same active configuration: local validation and the live Router's authoritative validation both passed.

---
<details>

- [x] Title has exactly one category.
- [x] Linked issue is accepted with one Workgroup owner.
- [x] Commits are signed off.
- [x] Purpose, tests and results reflect this change.

</details>


FAUST-BENCHOU (approved):
GC

PR #3830, opened 2026-09-15
[Bug] Keep Vault context probes within the triage input budget

### LINKED ISSUE #3826: [Bug] Built-in Vault positive probes exceed the Vela triage budget
## Problem

Six authored long-input positives in the built-in `mom-v1` probe suite predate Vela's current triage input contract. They expect a normal private/guard decision at old 120K–240K capacity bands, but required Guard/PII/Hazard evidence fails before model selection when the configured classifier document budget is exceeded.

## Reproduction

On main `c1e00bb0fada18b891cbe20afb663b201cad0cb6`, use the unchanged Vault recipe with its Vela models and call management preview (no generation backend request):

- The maintained `repeat: 9000` neutral-text variant produces 180,011 classifier tokens, despite its name referring to an over-240K band, and returns HTTP 503.
- A reduced `repeat: 2000` variant still produces 40,011 tokens and returns HTTP 503.
- The response reports `input_limit` for Guard and PII, unavailable classifier evidence, `fail_request`, and no selected private route.
- A budget-compatible neutral-text variant does return private/selected without signal errors.

This is the correct runtime fail-closed behavior. The maintained success fixtures and their capacity names are obsolete. Probe schema v1 cannot represent pre-decision HTTP 503; `expected_selection_status: unavailable` applies only after a decision is available.

## Expected fix

Replace the six obsolete positive inputs with accurately named, budget-compatible long-input fixtures while preserving their existing decision, signal, plugin and coverage assertions. Keep the original failures documented here. Preserve the existing HTTP input-limit/fail-request integration regressions; do not relax classifier limits, truncate security evidence, fabricate selection success, or change model thresholds.

Document that Vault's required triage budget is checked before provider context capacity, and that the default CPU Hazard deployment can be expensive on long inputs. Regenerate the bundle digest and catalog projections from authored sources.

## Ownership

MoM Routing Workgroup.

Xunzhuo:
Closes #3826

## Purpose

Six Vault probes expected a routed backend decision after materializing 120K–240K-class context, but Vault's default triage tasks reject inputs beyond 32,768 tokens before backend context checks. These probes therefore failed with signal input-limit errors before exercising their routing assertions.

Keep the six probes within the triage budget, replace repetitive generated character payloads with readable neutral records, and rename their stale context-band identifiers. Preserve every expected decision, signal and plugin assertion. Regenerate the recipe catalog digests, pin the updated materialization receipt, and document the triage limit and default CPU Hazard latency.

Owned by MoM Routing. Stacked on #3820 for the corrected Vela test baseline.

## Test Plan

- Run `make recipe-conformance-static` and full changed-file checks, including generated catalog projections.
- Materialize the exact six final requests and send them to the real AMD Router with Vela triage enabled.
- Confirm an over-budget request still fails closed with signal errors and no selected model.
- Compare held-out Guard classification failures on CPU and AMD using the same checkpoint and identical inputs.

## Test Result

All 75 static conformance tests passed. Exact final requests contain 1,138–10,998 bytes of text; all six returned HTTP 200 without signal execution errors. The actual over-budget classification/Preview tests still fail closed as expected.

**Routing conformance is not fully passing:** only 2/6 final requests satisfy their unchanged policy assertions. Three neutral articles are classified as prompt attacks, while an attack with prior context reaches the sensitive decision instead of guard. CPU and AMD scores agree within 0.0000023 on these four inputs, so this sample does not indicate an AMD-specific provider regression. The remaining Guard quality problem is tracked in #3831. This PR fixes the deterministic input-budget mismatch; it does not claim to fix Guard quality or change classifier thresholds. Full changed-file checks passed, including hooks, catalog/schema/generated contracts, Go lint and the complete Router Go suite.

---
<details>

- [x] Title has exactly one category.
- [x] Linked issue is accepted with one Workgroup owner.
- [x] Commits are signed off.
- [x] Purpose, tests and results reflect this change.

</details>


FAUST-BENCHOU (approved):
LG

PR #3838, opened 2026-09-16
[Feature] Add per-backend circuit breaker for remote classifiers (follow-up)

### LINKED ISSUE #3726: [Feature] Add per-backend circuit breaker for remote classifiers

lesbass:
## Summary

Rebased, replaces PR #3740. All 7 reviewer feedback items addressed.

## Changes

11 files changed, 1173 insertions, 11 deletions.

1. **Cancellation not counted** — `isUnavailableError` excludes `context.Canceled` and non-retryable transport. Added regression tests.
2. **PromptGuardConfig validation** — shared `RemoteClassifierCircuitBreakerConfig.Validate()` called from `validatePromptGuardBackend`.
3. **Half-open maxProbes > 1 correctness** — `admittedProbes+unresolvedProbes+halfOpenFailed` accounting.
4. **Half-open probe stuck on non-counted error** — `releaseProbe()` releases admission slot without counting.
5. **Gauge never updated, no transition logs** — `transitionLocked()` updates gauge and emits structured log.
6. **Classifier-owned timeout counted** — `isUnavailableError` distinguishes caller cancellation from classifier's own DeadlineExceeded.
7. **Stale generation ignored** — `halfOpenGen` counter prevents stale closed-state responses from interfering with half-open probes.

Fixes #3726

Signed-off-by: Stefano Maffeis <lesbass@gmail.com>


FAUST-BENCHOU (changes requested):
conflicts

src/semantic-router/pkg/classification/classifier_option_backends.go
```diff
@@ -43,6 +43,16 @@ func (b *classifierOptionBuilder) addRemoteCategoryClassifier(categoryMapping *C
 	if err != nil {
 		return err
 	}
+	models := consumerModelRuntime([]*classifierModelRuntime{b.models})
+	cbCfg := backendCfg.CircuitBreaker
+	owned, err := prepareRemoteSequence(models, models.remoteSpec("domain_classifier", backendCfg), external, transport)
```
Xunzhuo, line 48: **[P1] Complete the category backend construction**

The function still declares `backend, err := newCategoryHTTPBackend(...)` above, but the new call refers to undeclared `transport`, and `backend :=` below redeclares the same variable in the same scope. This prevents the classification package from compiling. Replace the old construction coherently with the owned transport path, then wrap that backend and run the core build and tests.

Xunzhuo (changes requested):
The category backend integration has a compile blocker. Please complete the construction path and verify the exact-head core build before merge.

Issue #3840, opened 2026-09-16
[Bug] Vector store search reads only the first window of a long query


subin9:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

The vector store query is embedded in one call, so only its first window is read.
`retrieveFromVectorStore` does this at `src/semantic-router/pkg/extproc/req_filter_rag_vectorstore.go:51`,
and the search API at `src/semantic-router/pkg/apiserver/route_vectorstore_search.go:67`.
The embedder is the owned provider taken at
`src/semantic-router/pkg/routerruntime/vectorstore_runtime.go:73`, which truncates its input.

#3650 fixed this for the milvus, qdrant and external RAG backends by embedding every
window of the query. The vector store backend was not part of that change, so a question
that sits past the first window is never read, and two different questions behind the same
preamble retrieve the same documents.

### To Reproduce

Router built from `342d4523`, vector store on the `memory` backend, default `bert` embedder
(`models/mom-embedding-light`, a BertModel whose `max_position_embeddings` is 512). Two
documents are attached: one about refunds, one about outage status updates. The call sites
above are unchanged on current main `c1e00bb0`.

Two queries share a 1,844 token preamble and differ only in the question at the end:

```
POST /api/v1/storage/vector-stores/{id}/search
A: <preamble> Question: how long does a refund for an annual plan take?                   (1,856 tokens)
B: <preamble> Question: how often does the status page publish updates during an outage?  (1,857 tokens)

A -> doc_outage.txt 0.3622929134751272 , doc_refund.txt 0.22893272531554146
B -> doc_outage.txt 0.3622929134751272 , doc_refund.txt 0.22893272531554146
```

Identical results and identical scores, and the refund question ranks the outage document
first. Moving the same questions to the front of the same preamble separates them, and each
retrieves its own document:

```
refund question first -> doc_refund.txt 0.4674365227484138
outage question first -> doc_outage.txt 0.547819785014095
```

Growing the preamble shows where it starts. The question is still read at 476 tokens and
never read from 936 tokens on, and the scores stop moving entirely:

| preamble | total tokens | first result |
|---|---|---|
| 10 units | 476 | doc_refund.txt 0.447589 |
| 20 units | 936 | doc_outage.txt 0.362293 |
| 40 units | 1,856 | doc_outage.txt 0.362293 |
| 80 units | 3,696 | doc_outage.txt 0.362293 |
[...]

PR #3841, opened 2026-09-16
[Bug] Search a long vector store query by every window

subin9:
## Purpose

Related #3840

`retrieveFromVectorStore` and `POST /api/v1/storage/vector-stores/{id}/search` embedded the
query in one call, so a query longer than the embedder's window was searched by its opening
alone. Two questions behind the same preamble produced the same vector and retrieved the
same documents.

#3650 gave the milvus, qdrant and external RAG backends one embedding per window. This
carries that to the vector store backend. Both call sites embed every sampled window, search
once per window, and keep each chunk's best score.

Three notes on the shape of the change:

- The window helper moves to `pkg/embedding` as `QueryVectors`, so the search API can reach
  it without the RAG method that pins the `bert` model. The vector store keeps the model it
  was ingested with.
- An embedder that exposes no token windows keeps its single embedding, because a remote
  model owns its own truncation and local offsets would not describe it.
- Hybrid search keeps one embedding as well. It also scores the query text, so searching per
  window would weigh its lexical half once per window.

Hits merge on file, chunk index and content, and the comparator ends on content as well, so
the sort key is the whole merge key. Not every backend reports a chunk index, and llama_stack
also returns hybrid hits that carry no file id, which would otherwise collapse a response into
its best scoring row and leave the survivor to map order. `ragHits` could not be reused,
because a chunk keeps its filename and index through this merge.

The branch is now a single commit on top of main.

## Test Plan

- New unit tests for the window helper and the merge helper.
- New tests on both call sites: a query longer than one window is embedded once per window
  and its question is read, a query inside one window is embedded once, and hybrid search
  keeps one embedding.
- New merge regressions: two chunks of one file with no chunk index stay separate, hybrid hits
  that report no file stay separate, and a tied score pair keeps the same order over repeated
  merges.
- The existing windowing and embedding identity tests, to show #3650 behaviour is unchanged.
- `bash .claude/skills/commit-checklist/scripts/gates.sh`, which runs `make check` with
  `BASE_REF=origin/main` under the pinned toolchain.

## Test Result

Gates at the current head, all four rows pass:

```
PASS  git diff --check (whitespace)
PASS  make harness-check
PASS  make check (the CI PR lane)
PASS  go mod tidy -diff (src/semantic-router)
```

Package tests:

```
ok  pkg/embedding    0.164s
ok  pkg/vectorstore  5.763s
ok  pkg/extproc      3.658s
ok  pkg/apiserver    9.783s
```

The reproduction in #3840 was measured against a router built from `342d4523` before this
change: two queries of 1,856 and 1,857 tokens sharing a 1,844 token preamble returned
identical scores, `doc_outage.txt 0.3622929134751272` and `doc_refund.txt 0.22893272531554146`,
and the refund question ranked the outage document first. The same questions placed at the
front of the preamble retrieved their own documents, `doc_refund.txt 0.4674365227484138` and
`doc_outage.txt 0.547819785014095`.


src/semantic-router/pkg/vectorstore/search_merge.go
```diff
@@ -0,0 +1,63 @@
...
+func MergeSearchResults(topK int, batches ...[]SearchResult) []SearchResult {
+	type chunkKey struct {
+		fileID string
+		index  int
+	}
+
+	best := make(map[chunkKey]SearchResult)
+	for _, batch := range batches {
+		for _, result := range batch {
+			key := chunkKey{fileID: result.FileID, index: result.ChunkIndex}
```
Xunzhuo: **[P2] Preserve distinct chunks when the backend has no chunk index**

`parseLlamaStackSearchResults` leaves `ChunkIndex` at zero for every hit. Two different chunks from the same Llama Stack file therefore collide here, and only the highest-scoring content survives. Both the search API and vector-store RAG now call this merge even for a single query vector, so short queries also lose previously returned context. The parser-to-merge regression returns one hit from a two-chunk response. Preserve a valid backend-specific chunk identity (or retain distinct hits when indices are unavailable), and cover two different chunks from the same file.

Xunzhuo (changes requested):
One retrieval regression remains: merging Llama Stack results drops distinct chunks from the same file, including for short queries. Please preserve their identity and add the regression described inline.

src/semantic-router/pkg/vectorstore/search_merge.go
subin9: Right! Fixed. The merge key now includes content, so chunks that arrive without a chunk index stay distinct. Added the parser-to-merge regression for two chunks of one file, plus one for hybrid hits, which carry no file id either and collapsed the same way.

src/semantic-router/pkg/vectorstore/search_merge.go
```diff
@@ -0,0 +1,68 @@
...
+		merged = append(merged, result)
+	}
+	sort.Slice(merged, func(i, j int) bool {
+		if merged[i].Score != merged[j].Score {
+			return merged[i].Score > merged[j].Score
+		}
+		if merged[i].FileID != merged[j].FileID {
+			return merged[i].FileID < merged[j].FileID
+		}
+		return merged[i].ChunkIndex < merged[j].ChunkIndex
```
Xunzhuo: **[P2] Include content in the final sort tie-breaker**

The new key preserves different content at the same file/chunk index, but this comparator still treats those hits as equal when their scores match. Llama Stack leaves every chunk index at zero, so map iteration now decides which chunk survives the top-K cutoff. Repeating the same two-hit input with equal scores and `topK=1` returned one chunk 883 times and the other 117 times in 1,000 calls. This changes both search API results and the context supplied to RAG for identical backend responses. Compare `Content` after `ChunkIndex` (or another stable unique identity) and add a tied-score, repeated-merge regression.

Xunzhuo (changes requested):
The content key fixes the reported chunk collapse, and the focused search, embedding, API and RAG tests pass. One ordering regression remains: identical equal-score results can produce different top-K context. Please add a final tie-breaker and cover repeated merges. Deployed-stack tests were not run.

src/semantic-router/pkg/vectorstore/search_merge.go
subin9: Fixed. Content now ends the comparator, so the sort key is the whole merge key, and a tied score regression merges the same pair 500 times and checks both the top one and the full order. On deployed stack tests, both `make impact` and the repo own `select_profiles` return no E2E profile for this diff, and the vector store cases assert only that some result contains an expected term, so they would not cover ordering. Happy to run `rag-hybrid-search` if you still want it.

Xunzhuo (approved):
The chunk-identity and equal-score ordering regressions are fixed, including the repeated top-K regression. API, RAG and prepared-provider integration tests pass against current main, including global embedding-provider selection. I found no remaining blocker. Deployed-stack and full-model runtime tests were not run.

wilsonwu (approved):
LGTM

drivebyer:
Manually merged `main` into the PR branch to bypass Mergify's update restriction on workflow files (`.github/workflows/ci-changes.yml`).

@mergifyio queue

PR #3847, opened 2026-09-16
[Bug] Run static global validators on Kubernetes startup, not just after CRD merge

### LINKED ISSUE #3758: [Bug] Kubernetes startup skips global config validation

ankit373:
Fixes #3758.

## The problem

`validateConfigStructure` skipped every global and routing-profile validator whenever `ConfigSource` is `kubernetes`, deferring all of them to `ValidateKubernetesConfigContracts`, which the reconciler only calls after merging the CRDs. That is correct for anything that reads `Providers.Models`, `Routing.Decisions`, or `Routing.Signals`, the three fields CRD conversion actually populates (`pkg/k8s/converter.go`), since those are genuinely absent at the initial static parse. It is wrong for everything else: a static setting unrelated to routing state, like `global.memory.default_similarity_threshold`, loaded on Kubernetes with no check at all until the reconcile path happened to exercise it.

## The fix

Traced all 21 global validators against the three CRD-populated fields directly, rather than classifying by name. Two corrections along the way, each confirmed with a test before writing the classification down: the prompt guard, category, and PII backend validators looked coupled to provider state because they resolve a remote backend's model name, but that lookup is against `ExternalModels`, a static field CRD conversion never touches, so they're safe to run early. `validateComplexityModelBackendContracts` stays deferred because it also runs `ValidateComplexityRuleBoundaries`, which checks rules that only exist inside a recipe, and `Recipes` does derive from the CRD-populated `Decisions`/`Signals` for the default recipe.

`globalConfigValidatorRegistry` is the new single place a validator gets registered, with `requiresRoutingState` marking the ones that need deferral. `globalConfigContractValidators` derives from it, so existing callers and tests are unaffected. `validateConfigStructure` now runs the filtered subset for Kubernetes instead of returning `nil` immediately; `ValidateKubernetesConfigContracts` is unchanged and still runs everything once CRDs are merged.

## A gap this surfaced

Catching the bad setting earlier meant it now surfaces through `ParseYAMLBytes`'s own internal validation call inside `validateAndUpdate`, before the explicit `ValidateKubernetesConfigContracts` call further down, so it took a different error branch that never set the CRD's `Ready` condition. Added `failValidation` so both branches report `ValidationFailed` the same way. Caught by an existing test, `TestReconcileKubernetesConfigValidationDispatch`.

## Verification

- `go build`, `go vet` clean across the whole module
- `go test ./pkg/config/... ./pkg/k8s/...` passing, including the reconciler suite that caught the status-condition gap
- Mutation-tested: reverting `validateConfigStructure` to the original `return nil` fails exactly the two new specs added for this issue and nothing else
- New tests: rejects a bad static setting before CRD conversion, still tolerates absent routing state for the validators that need it, resolves a remote classifier backend against `ExternalModels` alone with everything CRD-populated left at zero value, and validates the same setting after CRD conversion too


Xunzhuo (approved):
The startup validator registry preserves the full post-CRD validation path while checking static settings earlier, and normalization failures now update CRD status consistently. The exact-head config and Kubernetes tests pass.

wilsonwu (approved):
LGTM

Kritoooo:
Thanks for working on this. I have #3759 open for the same issue (#3758), so the two PRs overlap in both Kubernetes startup validation and CRD `ValidationFailed` reporting.

One detail worth aligning is the validation boundary: #3759 separates deployment definitions from binding compilation, and complexity backend settings from routing-rule checks. It also separates PII/Prompt Guard static settings from checks that read recipe bindings. This keeps static errors detectable at startup while allowing stale file bindings to be replaced by CRDs, with regression coverage for that case and per-recipe window budgets.

cc @Xunzhuo @wilsonwu

ankit373:
Closing in favor of #3759, which was already open and further along (review feedback addressed, approvals in hand) before I picked up #3758. Sorry for the duplicate effort, I should have checked for a linked PR before starting. Thanks @Kritoooo for the clear breakdown of where the two approaches differ.

PR #3850, opened 2026-09-16
[Bug] Reject a blank vector store query as a client error

### LINKED ISSUE #3849: [Bug] Vector store search reports a whitespace query as an internal error
### Proposed Workgroup

### Describe the bug

`handleSearchVectorStore` reports rejected input as an internal error, and neither of its
500 paths records why the request failed.

In `src/semantic-router/pkg/apiserver/route_vectorstore_search.go`:

- The only guard on the query is `req.Query == ""` at line 97, so a query of only
  whitespace passes it and fails inside the embedder, answering 500 `EMBEDDING_ERROR`
  at line 69. The embedder's own message is
  `model task input is invalid: model input text must not be empty`, raised by
  `validateText` at `src/semantic-router/pkg/modelruntime/native/runtime.go:75-78` and
  wrapped as `binding.ErrInvalidInput`
  (`src/semantic-router/pkg/modelruntime/binding/identity.go:17`). The text is rejected
  before the model runs, so this is a client input error, not an internal fault.
- Both 500 paths, line 69 and line 75, discard `err` and call no logger, so the server
  keeps no record of the cause.

The missing store case is already correct on current main: #3769 maps it to 404 `NOT_FOUND`
and an incompatible embedding identity to 409 at lines 60 to 62. This issue is only about
the whitespace status code and the absent logs.

### To Reproduce

Router built from `342d4523`, vector store on the `memory` backend.

```
POST /api/v1/storage/vector-stores/{id}/search   {"query": "   \n\t  "}
  -> 500 {"code":"EMBEDDING_ERROR","message":"failed to generate query embedding"}
```

No line is added to the router log for that request. An empty string is already handled
correctly and returns 400 `INVALID_INPUT` with `query is required`. Sending the same
whitespace to `POST /api/v1/diagnostics/embeddings` returns the underlying reason, because
that handler passes the error through:

```
500 {"code":"EMBEDDING_GENERATION_FAILED",
     "message":"failed to generate embedding: model task input is invalid: model input text must not be empty"}
```

### Expected behavior

A query that is empty after trimming is rejected as 400, like the empty string. A genuine
failure stays 500 and is logged with its cause.

### Impact

A client cannot tell a bad request from a broken server and will retry a request that can
never succeed. An operator sees a 500 with no log line explaining it.

subin9:
## Purpose

Closes #3849

The vector store search handler accepted a query of only whitespace, which the embedder then refused because it trims text before the model runs, so the request came back as 500 `EMBEDDING_ERROR`. An empty string was already a 400, and the two answers described the same client mistake. Neither remaining 500 path recorded why the request failed.

`parseVectorStoreSearchParams` trims before the emptiness check, so a query with no text left is 400 `INVALID_INPUT` and never reaches the model. A query padded with spaces still runs.

Both 500 paths log the cause before answering. The client message stays generic, so the reason is on the server rather than in the response.

The missing store case needed no change here. #3769 already maps it to 404 `NOT_FOUND` and an incompatible embedding identity to 409.

#3841 has merged, so this is rebased down to the single commit above it. The files it used to carry are on main now, and what is left here is the blank query guard.

## Test Plan

Search a store with a query that is empty, spaces only, and whitespace only, then with real text padded by spaces.

## Test Result

Every blank form answers 400 `INVALID_INPUT` and the embedder is never called. Real text padded by spaces still answers 200. Before this change the whitespace query answered `500 {"code":"EMBEDDING_ERROR","message":"failed to generate query embedding"}` and the router log said nothing.


src/semantic-router/pkg/embedding/query_vectors.go
```diff
@@ -0,0 +1,80 @@
...
+	if embedder == nil {
+		return nil, fmt.Errorf("embedding provider was not prepared")
+	}
+	tokenizer, ok := embedder.(WindowProvider)
+	if !ok {
+		return singleQueryVector(ctx, embedder, text)
+	}
+	windows, err := tokenizer.Windows(ctx, text, 0)
+	if err != nil {
+		return nil, err
```
Xunzhuo, line 49: A prepared remote provider passes this type assertion even though it cannot tokenize locally: NewVectorStoreRuntime obtains its embedder through Set.Get, which always returns a providerView implementing WindowProvider. Its Windows method then returns an unsupported-capability error (the owned remote provider does the same), so this return prevents Embed from being called. Consequently even short non-hybrid searches with a valid remote embedder now return HTTP 500/EMBEDDING_ERROR, and vector-store RAG fails too. Please check the underlying tokenizer capability and retain the single-embedding path for remote providers, without masking real tokenizer failures. Add a regression using Set.Get or the owned remote provider; the bare plainEmbedder test misses the runtime wrapper.

Xunzhuo (changes requested):
The new windowed search path breaks supported remote embedding providers before any embedding request is sent. Please preserve their single-embedding behavior and cover the prepared-provider path in both vector-store search and RAG tests.

src/semantic-router/pkg/embedding/query_vectors.go
subin9: Right, the fix for this review is on #3841, which this PR is stacked on. This PR itself only carries the blank query change. Thank you!

subin9: This now sits on the current head of #3841, so the fix is in this diff.

Xunzhuo (dismissed):
Blank queries now return 400 before embedding, while queries with surrounding whitespace still work. The stable chunk ordering and prepared-provider regressions also pass, along with API, RAG and global-provider integration tests against current main. I found no remaining blocker. Deployed-stack and full-model runtime tests were not run.

wilsonwu (dismissed):
LGTM

subin9:
#3841 has merged, so this is rebased down to the single commit that rejects the blank
query. That dismissed both approvals.

The two files left are unchanged from the version you approved. Everything else the
diff used to carry is on main now.

Xunzhuo (approved):
Blank queries return 400 before embedding, and queries with surrounding whitespace still work. The relevant source is unchanged from the completed API and remote-provider regressions, and current-head Core CI passes the API suite. I found no remaining blocker. Deployed-stack and full-model runtime tests were not run.

PR #3854, opened 2026-09-16
[Feature] Hallucination detector adopts the shared token_spans.v1 span contract

### LINKED ISSUE #2928: [Feature] Hallucination detector adopts the shared TokenClassifierBackend span contract
### Goal

Hallucination currently ships its own backend mechanism from #2526 (`backend: candle | endpoint`, its own constants in `pkg/config/validator_hallucination.go`, and detector code separate from the tier interfaces), while `pkg/classification/classifier_backend_tiers.go` already names hallucination as an intended consumer of those tiers.

Once #2922 lands a detector-agnostic `TokenClassifierBackend`, the hallucination detector should adopt it, so `backend: endpoint` becomes one instance of the shared span mechanism instead of a parallel path. Wiring the encoder detector through it gives the interface a second consumer with a different signal and a different backbone, which is what shows the contract is actually detector-agnostic rather than PII-shaped.

Depends on #2922 — the span contract needs to ship and prove out on PII first, as that issue notes.

### Proposed Workgroup

### Scope

##### Primary layer

global level

##### Why this layer?

Same as #2760: backend selection for a built-in detector is router-owned config.

##### Why do you need this feature?

Without it the codebase keeps two span paths, one for PII and one for hallucination, with separate config surfaces and validators for the same output shape. #2922 explicitly defers this convergence to a follow-up rather than solving it inline.

##### Additional context

Sub-issue of #2760, depends on #2922. Related: #2526, #2927

Happy to implement, and can supply golden fixtures (character offsets, multilingual cases, overlapping spans) for the span contract.

### Non-goals

### Completion signal

- The hallucination detector's `backend: endpoint` path is served by the shared `TokenClassifierBackend` and the `token_spans.v1` contract from #2922, with no hallucination-specific constants left in `pkg/config/validator_hallucination.go`.
- The contract's golden fixtures run against both the PII detector and the hallucination detector and pass on both, including the multilingual, offset-0, end-of-text, overlapping and nested cases.
- Hallucination spans reported through the shared path are identical in offsets, labels and scores to those reported through the current dedicated path on the same inputs.
- One detector-agnostic span mechanism remains in the codebase; the second consumer with a different backbone is what shows the contract is not PII-shaped.

adaamko:
Closes #2928

## Purpose

Hallucination detection becomes a second consumer of the `token_spans.v1` decoder that PII uses since #3498, instead of keeping its own span path. The convention is the one agreed on #2928: the answer is the classified text, offsets index the answer only, context and question travel as request parameters.

- `pkg/classification/hallucination_detector_endpoint.go`: a `hallucination_detector` binding may declare `adapter: http_classify` next to `http_chat`. The detector posts `{"inputs": answer, "parameters": {"context": ..., "question": ...}}` to `/classify` and decodes through `HTTPTokenClassifierInference.ClassifyGrounded`. Offsets are code points into the answer, converted to bytes at the `TokenEntity` boundary; span text must equal the answer slice; `truncated_at` stays a partial-scan error, never a clean verdict.
- The chat adapter now feeds its quoted spans through the same alignment: one label set, one bounds and text check, one byte conversion, one duplicate rule. Repeated quotes take successive occurrences instead of all claiming the first. The chat adapter has no per-span score, so the alignment takes the score requirement as a parameter rather than fabricating one; the classify path stays strict.
- Span labels resolve from the binding the way PII labels do: a `mapping_path` names a `label_to_idx`/`idx_to_label` file whose class zero is the outside label; without one the default set applies (`HALLUCINATED`, `unsupported`, `contradicted`, `unverifiable`, and the chat taxonomy categories; `SUPPORTED` and `O` are outside labels and rejected as spans).
- `HallucinationResult` keeps `UnsupportedSpans` as text for the grounding and looper consumers and gains `Spans` with byte offsets, label and score. Enhanced spans carry the label.
- `pkg/config`: the legacy `backend: endpoint` scalar with `endpoint` and `model_id` compiles into the binding a recipe would otherwise declare (http deployment, `http_chat`, `token_spans.v1`), so the runtime has one selection mechanism. An explicit `hallucination_detector` binding wins; the candle default stays the module's canonical local binding. `validator_hallucination.go` is gone; the legacy tokens, their validation and the desugar live in `hallucination_legacy_backend.go`, and config validation still rejects an unknown backend at load. The classifier selects native versus remote from the compiled plan, and an endpoint request is never silently downgraded to the local model when incomplete.
- Model info reports the binding's adapter and contract; a `token_spans.v1` service shows as `token_spans_endpoint` rather than an OpenAI-compatible endpoint.
- Docs: `website/docs/installation/runtime/external.md`.

Left for follow-ups, deliberately: a deterministic remote hallucination E2E profile (same shape as `pii-remote-backend`); a provider's context truncation has no signal in v1 (`truncated_at` is answer-relative); NLI stays a separate local model; streaming is its own issue.

## Test Plan

```
cd src/semantic-router
go vet ./pkg/config/ ./pkg/classification/ ./pkg/apiserver/
go test ./pkg/config/ -run 'Hallucination|ModelBindings'
go test ./pkg/classification/ -run 'Hallucination|EndpointDetector|TokenSpans|Fixture'
go test ./pkg/services/ ./pkg/extproc/ ./pkg/apiserver/ ./pkg/modeldownload/
```

`pkg/classification` links against the native bindings (candle, onnx, ml, nlp `target/release`), so the test binary needs `LD_LIBRARY_PATH` pointing at all four. The #2922 golden fixtures now also run against the hallucination detector with context and question attached, relabelled per case onto the hallucination label set, and a parity test checks the chat and classify adapters return identical text, offsets and labels for the same detections, repeated text included.


Xunzhuo (approved):

wilsonwu (approved):
LGTM

PR #3855, opened 2026-09-16
[Bug] Keep the PII block refusal from passing as a partial scan

adaamko:
Related #2922, follow-up to #3498.

## Purpose

Under `classifier.pii.on_error: block`, a `token_spans.v1` provider that declared `truncated_at` made `scanPIIChunks` refuse the scan, but the refusal wrapped `ErrTokenSpansTruncated`. Every caller tolerates that sentinel as partial coverage, so `ClassifyPIIWithDetails`, `DetectPIIInContent` and the detection service turned the refusal back into a clean answer: HTTP 200 with `has_pii: false` and `scan_incomplete: true` instead of an error.

- `pkg/classification/classifier_pii_ops.go`: a refusal now carries its own sentinel, `ErrPIIScanRefused`, built by `piiScanError` at the three refusal sites. It keeps the provider's message as text but does not unwrap to the truncation sentinel, so it stops at the service boundary.
- The allow path is unchanged: a declared truncation still returns its spans with `ErrTokenSpansTruncated` alongside, and the service still reports `scan_incomplete`.

## Test Plan

```
cd src/semantic-router
go test ./pkg/services/ -run TestPIIServiceTruncationPolicy
go test ./pkg/classification/ ./pkg/services/
```

`pkg/services/classification_pii_truncation_policy_test.go` runs `allow` and `block` against complete, truncated and failing providers through `BuildClassifier` and asserts that block/truncated is refused at the service boundary while allow/truncated returns its detections with `scan_incomplete`. Native bindings needed for `pkg/classification` and `pkg/services`, as in #3498.


Xunzhuo (approved):

wilsonwu (approved):
LGTM

FAUST-BENCHOU (approved):

adaamko:
@Mergifyio queue

PR #3859, opened 2026-09-16
[Docs] Simplify agent installation and verification

Xunzhuo:
Related #3216

## Purpose

Make agent installation a short path through environment discovery, configuration, real routing verification, and improvement. Load deployment, configuration, and evaluation details only when needed; verify the Dashboard when requested.

Use absolute URLs in the authored skill, its published copy, and every reference so each document works when downloaded on its own. The publisher also resolves section links, site paths, and Markdown reference definitions, and rejects missing skill documents. Remove the platform-specific reference: the agent inspects the actual host and installed `serve --help`, explicitly selects a supported `--platform`, and verifies the chosen runtime.

## Test Plan

- Check downloaded documents and the complete reference graph without relying on a repository path.
- Verify the generated mirror and host-driven platform guidance.
- Run installation-surface tests and the changed-file repository checks.

## Test Result

- 11 publication tests and 8 installation-surface tests pass.
- `make agent-skill-sync`, `make agent-skill-check`, and changed-file `make check` pass, including formatting, lint, 34 workflow contracts, 165 CI policy tests, 50 harness tests, and translation coverage with zero regressions.
- GitHub checks will validate the pushed revision. These documentation checks do not claim hardware or load qualification.


tools/agent/skills/vllm-sr-agent-operations/references/configuration-loop.md
```diff
@@ -24,8 +24,21 @@ A new stack has no remote revision to plan against. Validate locally, launch
 with the chosen stack settings, then discover readiness and the live schema.
 Never accidentally plan against another stack on a default port.
 
-For a running Router, preserve unrelated fields and choose replacement of a
-complete document or merging an intentional partial patch explicitly:
+For a running Router, save its current canonical configuration and derive the
+candidate from it:
+
+```bash
+vllm-sr config get --format yaml --endpoint "$ROUTER_ORIGIN" > active.yaml
```
neelkumar01: I think there is a safety issue with `config get` output

The output does not always return original config as sensitive values got replaced with literal string "[REDACTED]"

This is mentioned in `src/semantic-router/pkg/apiserver/config_redact.go`

The code that performs the redaction is:

```go
if isSensitiveConfigKey(key) {
    if isPureEnvironmentReference(nested) {
        out[key] = nested
    } else {
        out[key] = redactedConfigValue
    }
}
```
So the fetched config may have

```
api_key: "[REDACTED]"
```
even if only intended fields are edited, candidate.yaml still have REDACTED values and this config value can be submitted back

Can we add a note to avoid using config output we get from `config get` containg redacted values?

PR #3869, opened 2026-09-16
[Router] guard in-memory semantic cache polarity

### LINKED ISSUE #2727: [Bug] Guard the in-memory semantic cache against polarity mismatches

Peterren:
Closes #2727
Refs #2691
Supersedes #2728 because its source branch does not allow maintainer updates.

## Purpose

Preserve the existing #2728 lexical polarity guard while resolving its merge conflict with current `main`. This replacement does not intentionally change the original guard policy or broaden its scope. Raising this PR as there is a persisting merge conflict in the existing #2728

## Conflict resolution

- Keep the existing lexical negation and antonym candidate filtering.
- Keep the existing HNSW fallback and lower-ranked candidate behavior.
- Adapt the lookup path to the current context-aware embedding API and request-owned `LookupResult`.
- Run the already-landed optional NLI verifier after lexical candidate selection.
- Adapt the existing lexical and NLI tests to the current search-result shape.

## Test Plan

- Merge completed with one content conflict in `inmemory_cache_search.go`.
- Repository pre-commit checks on all merge-conflict files: passed.
- Go formatting: passed.
- Supply-chain security scan: passed.
- Structure and architecture dependency checks: passed.
- Standard pull-request CI validates the current Linux build and test matrix.


Peterren:
@WUKUNTAI-0211 would abandon this PR if the original #2728 is updated

src/semantic-router/pkg/cache/negation_regression_test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func runNegationRegressionPairs(t *testing.T) int {
+	t.Helper()
+	guardExercised := 0
+	for _, pair := range negationRegressionPairs {
+		cached, incoming := pair[0], pair[1]
+		c := newMmbertCache(t)
+		cachedAnswer := []byte("CACHED-ANSWER-FOR::" + cached)
+		if err := c.AddEntry("neg", "model-x", cached, []byte("req"), cachedAnswer, 3600); err != nil {
```
Xunzhuo, line 114: [P1] Pass context.Background() to both AddEntry calls. AddEntry now requires a leading context.Context, but this call and the paraphrase helper at line 146 still supply only six arguments. CGO_ENABLED=1 go test ./pkg/cache fails with "not enough arguments in call to c.AddEntry" at both sites, before the missing-model skip can run, so the entire Linux CGO cache test package is blocked. Update both helpers to the context-aware API and rerun the cache tests.

Xunzhuo (changes requested):
The new regression helpers still use the old AddEntry signature, so the Linux CGO cache test package does not compile. Please update both calls to pass a context and rerun the cache tests. The polarity and model-backed regressions could not run because compilation fails.

Xunzhuo:
Thanks @Peterren for carrying the original polarity fix forward. Closing as superseded by merged #3944: main now filters lexical polarity mismatches while retaining valid lower-ranked candidates and the HNSW fallback, then runs the optional NLI verifier on the remaining winner. The cache wiring, fallback and model-backed regression coverage are included in the merged implementation.

PR #3878, opened 2026-09-16
[Bug] Make dsl validate and compile honor constraint diagnostics

### LINKED ISSUE #3674: [Bug] Make dsl validate and compile honor constraint diagnostics
### Goal

`dsl validate` exits 0 and `dsl compile` writes YAML even when validation reports a `DiagConstraint`, because `writeValidationDiagnostics` returns only the error count and `Compile` never calls `Validate`. A rule tree the DSL flags as invalid still reaches the router and fails at startup. Related #3208, #3249.

### Proposed Workgroup

MoM & Routing

### Scope

- Count `DiagConstraint` toward the non-zero exit of `dsl validate`.
- Run `Validate` inside `CLICompile` and refuse to write output when a blocking diagnostic exists.
- Keep `Compile` as a library call unchanged; the gate lives in the CLI paths.
- Add tests: `dsl validate` exits non-zero on a constraint, `dsl compile` writes nothing on a constraint, both pass on a clean file.

### Non-goals

- Changing which checks are `DiagConstraint` versus `DiagWarning`.
- Adding new validation rules.
- Touching the Go loader, CLI, or Dashboard producers.

### Completion signal

A DSL file with an `on_unknown` plus `on_error` conflict makes `dsl validate` exit non-zero and `dsl compile` produce no YAML, with tests protecting both.

anguszzzz:
Make `dsl validate` exit non-zero on constraint violations, and stop `dsl compile` from writing output for a rule tree the validator already flagged.

   Closes #3674

   ## Purpose

   Two CLI paths in `pkg/dsl` ignored constraint diagnostics: `writeValidationDiagnostics` counted all three levels in its summary line but returned only the error count, so `dsl validate` printed `🟠 1 constraint(s)` and exited 0; and `CLICompile` never called `Validate`, so the same file still compiled to YAML and failed later at router startup.

   `dsl validate` now counts `DiagConstraint` toward its exit code, and `CLICompile` runs `Validate` after a successful compile and refuses to write any output (file or stdout, all formats and `--base`) when a blocking diagnostic exists. Warnings still do not block; this matches the existing `hasBlockingDiagnostics` helper, which already treats errors and constraints as blocking.

   Per the issue's non-goals: the `Compile` library call is byte-identical, nothing moved between constraint and warning severities, no new validation rules were added, and the Go loader, `cmd/dsl` wiring, and Dashboard producers are untouched.

   Affected module: `src/semantic-router` (`pkg/dsl`). Owned by the MoM & Routing Workgroup

   ## Test Plan

   New tests in `cli_constraint_gating_test.go`, using two constraint rules that already exist on main (route `PRIORITY -1`; embedding `threshold: 1.5`), so nothing depends on unmerged work:

   - `TestCLIValidateExitsNonZeroOnConstraint`
   - `TestCLIValidateExitsZeroOnWarningsOnly`
   - `TestCLICompileWritesNothingOnConstraint` (file output and `-` stdout output)
   - `TestCLIValidateAndCompileSucceedOnCleanFile`

   Commands run:

   - `cd src/semantic-router && go test -count=1 ./pkg/dsl/`
   - `go vet ./pkg/dsl/`
   - `gofmt -l` on the two changed files

   ## Test Result

   All green: `ok github.com/vllm-project/semantic-router/src/semantic-router/pkg/dsl`; vet and gofmt clean.

   The tests were replayed against pre-change HEAD and failed there for the reported reason (compile wrote YAML for `threshold: 1.5`), so they are not vacuous.

   One note, not a blocker: the issue's completion-signal example (an `on_unknown` plus `on_error` conflict) becomes a `DiagConstraint` only once #3249 lands. The gate is not tied to any specific check, so that case is covered automatically then.

   ---
   <details>

   - [x] Links `accepted` issue #3674 with one owner (`wg/mom-routing`)
   - [x] Purpose, Test Plan, and Test Result reflect the actual scope, commands, and remaining note of this change

   </details>


Xunzhuo (approved):
The CLI now rejects constraint diagnostics before writing output, while warnings remain non-blocking. The command regressions and all current checks pass. This review covers static validate/compile behavior; live model TEST blocks were not exercised.

wilsonwu (approved):
LGTM

Issue #3800, opened 2026-09-14
[Bug] LLM classifier signal fails when the classifier model reasons


jfTNG:
### Proposed Workgroup

Router Models & Inference Runtime

### Describe the bug

A `type: llm` classifier signal errors out whenever the underlying model produces a reasoning trace that is returned as part of the content field.

### To Reproduce

Configure `routing.signals.classifiers` with `type: llm`, `parser_type: json`, pointing at a reasoning-capable model that is deployed with reasoning by default.  Classify a request; the classification fails.

### Expected behavior

Classification succeeds when the model includes the reasoning trace in the content if a proper json response is generated after the end of the reasoning trace.

Roy214:
Hi @Xunzhuo , Can I take this?

PR #3883, opened 2026-09-17
[Feature] Add reasoning control to external LLM classifiers

### LINKED ISSUE #3798: [Feature] Add reasoning control to LLM classifier signals
### Goal

It is not possible to disable reasoning for the LLM Classifier Signal. Routed catalog models support `reasoning_families` / `use_reasoning` / `reasoning_effort`, but external classifier models have no equivalent surface, so a classifier model always runs with its default thinking behavior.

### Proposed Workgroup

### Scope

- Extend `ExternalModelConfig` with a `reasoning:` block, mirroring `providers.models[].reasoning`.
- Validate it in `validateLLMClassifierExternalDependency`.
- Emit the resolved control as a top-level request field on the classifier request.
- Update the canonical schema and the classifier-signal docs.

### Non-goals

_No response_

### Completion signal

_No response_

JiaoliangYu:
Closes #3798

## Purpose

Add optional reasoning control for external models referenced by `type: llm`
classifier signals.

- Reuse the existing reasoning-family vocabulary and validation instead of
  defining classifier-specific modes or effort values.
- Scope the new control to external classifiers using `llm_provider: vllm`;
  configurations that request reasoning with another provider are rejected.
- Resolve the configured control once while building the classifier and carry
  it through `GenerationOptions` to `/v1/chat/completions`.
- Project vLLM mode and ordinary effort controls into
  `chat_template_kwargs`, and the legacy `top_level_reasoning_effort` family
  into the typed top-level `reasoning_effort` field.
- Preserve the current request exactly when `reasoning` is omitted.
- Update the canonical schema, example config, and English and Chinese
  classifier documentation.

If a model reasons despite the requested control but still returns valid
classification JSON, the Router continues to use that JSON. Parsing reasoning
traces from classifier responses remains scoped to #3800.

This does not modify routed `modelRefs`, the main provider request-mutation
pipeline, local or `sequence_classifier` backends, or per-request reasoning
selection.

## Test Plan

- `cd src/semantic-router && go test ./pkg/config ./pkg/configschema -count=1`
- Focused `pkg/classification` tests for projection and the complete
  builder-to-HTTP request boundary: omitted, disabled, enabled with effort, and
  top-level effort.
- `cd src/semantic-router && go vet ./pkg/config ./pkg/configschema ./pkg/classification`
- Incremental `golangci-lint` against `origin/main`.
- `make config-schema-check`
- `make docs-check-translation-coverage`
- `make docs-config-check`
- `git diff --check`

## Test Result

- PASS: config and config-schema tests.
- PASS: focused classifier projection and builder-to-HTTP tests. The local
  checkout does not contain the native archives required to link the package,
  so the selected non-native tests were run with temporary link-only stubs;
  those files were removed and are not part of the branch.
- PASS: Go vet and incremental golangci-lint (`0 issues`).
- PASS: generated config schema check, documentation translation coverage,
  documentation config check, and whitespace validation.
- The full translation-sync audit still reports repository-wide pre-existing
  metadata drift; this change introduces no translation coverage regression.

---
<details>

- [x] PR title begins with exactly one bracketed category.
- [x] The title does not stack prefixes.
- [x] The PR links accepted issue #3798 with exactly one recognized owner.
- [x] Commits in this PR are signed off with `git commit -s`.
- [x] The Purpose, Test Plan, and Test Result sections reflect the actual scope,
  commands, and remaining environment limitation.

</details>


src/semantic-router/pkg/config/validator_classifier_signal.go
```diff
@@ -260,9 +304,65 @@ func validateLLMClassifierExternalDependency(
...
+		return fmt.Errorf("%s reasoning_effort %q is not supported by family %q", path, effort, reasoning.Family)
+	}
+	return nil
+}
+
+func externalReasoningFamilyCanDisable(family *ReasoningFamilyConfig) bool {
+	if family != nil && family.Type == ReasoningFamilyTypeTopLevelReasoningEffort {
+		return family.Disabled != ""
+	}
+	return reasoningFamilyCanDisableConfig(family)
```
Xunzhuo, line 369: [P2] Honor the family mode contract before accepting disable

The built-in `qwen3.8-always-on` family declares only `modes: [enabled]`, but it also has `activation_parameter: enable_thinking`. `reasoningFamilyCanDisableConfig` treats any activation parameter as permission to disable, so an LLM classifier referencing this family with `use_reasoning: false` passes `ParseYAMLBytes` and the new projector sends `enable_thinking: false`. This contradicts the promised rejection of always-on families and accepts a request control the selected family does not support. Check the declared modes before using the transport fields as a fallback, and add a regression using this real built-in family.

Xunzhuo (changes requested):
The new reasoning fields reach the classifier HTTP request, and the added projection and request-boundary tests pass in Core CI. One validation gap remains: a built-in always-on family can be explicitly disabled because its activation parameter overrides its declared modes. Please enforce the family mode contract and cover that built-in case. Live vLLM inference was not run.

src/semantic-router/pkg/config/validator_classifier_signal.go
JiaoliangYu: Fixed in af45860ec. Explicit family modes now take precedence: when modes are declared without disabled, use_reasoning: false is rejected before inspecting transport fields. Transport fields are only used as a fallback when modes are absent, or to verify that an explicitly supported disabled mode can be represented. Added a ParseYAMLBytes regression using the built-in qwen3.8-always-on family. Verified with go test ./pkg/config -count=1 and go vet ./pkg/config.

Xunzhuo (approved):
The validator now honors the declared modes of always-on reasoning families. The canonical parsing regression and classifier HTTP request-boundary tests pass on this head, including the previously failing family. Live vLLM inference was not run.

wilsonwu (approved):
LGTM

PR #3884, opened 2026-09-17
[Router] Add random model selection algorithm

### LINKED ISSUE #3273: [Feature] Add a bounded random model-selection algorithm
### Goal

Add a simple, bounded random model-selection algorithm that can choose one model from the current eligible decision pool and serve as a maintained baseline for routing experiments and evaluations.

### Proposed Workgroup

MoM & Routing

### Scope

- Add a canonical `random` selector algorithm to the Router algorithm catalog and every maintained configuration surface.
- Select uniformly from the candidate set only after the shared hard-eligibility filter has removed incompatible models.
- Define reproducible test and replay semantics without making production selection accidentally deterministic.
- Preserve declared model references, recipe boundaries, typed all-ineligible behavior, and configured fallback behavior.
- Emit the selected model and bounded selection reason through existing diagnostics and Router Replay evidence.
- Align Go config/runtime, Python CLI, DSL, Dashboard, reference configuration, documentation, and deterministic unit/E2E coverage.

Parent Epic: #3202

### Non-goals

- Weighted traffic splitting, load-aware balancing, sticky sessions, bandit learning, or quality ranking.
- Randomly selecting providers outside the decision's declared and eligible model pool.
- Adding a new Looper or multi-model fan-out algorithm.

### Completion signal

- `algorithm.type: random` is accepted by all maintained configuration surfaces and chooses exactly one eligible declared model.
- Empty and all-ineligible pools follow the shared typed fallback contract.
- Seeded tests are deterministic, while the production default uses an appropriate randomness source.
- Unit, catalog-contract, configuration round-trip, Router Replay/diagnostic, and E2E tests cover the supported behavior.

abhinav-m22:
## Purpose

Adds `random`, a supported selector that picks uniformly from the candidates left after the shared eligibility filter. It gives other selectors an unbiased baseline to be measured against.

Closes #3273

- Per maintainer guidance, the selector does no filtering, fallback, or restoring of its own. An empty pool returns the shared `ErrCandidateModelsRequired`.
- Takes no config block (`algorithm: { type: random }`) and has no seed setting. Tests control the RNG through an unexported field.
- Every `modelRefs` entry gets an equal share. `weight` is not read.

## Changes

- **Router:** catalog entry, `RandomSelector`, factory and dispatch wiring, regenerated config schema
- **DSL:** compiler accepts `ALGORITHM random`; Dashboard editor highlights and suggests it
- **Config and docs:** fragment, `random_route` reference decision, tutorial page, sidebar entry, overview row, NL generator prompt
- **Dashboard validator:** accepts `random`. The one-candidate `"single"` exemption is now a helper covering `static` and `random`, which avoids a false failure on one-candidate random decisions. Other selectors are unchanged.
- **CLI:** `vllm-sr serve --algorithm random`
- **E2E:** `core-selection-random` checks that traffic spreads across at least 2 models over 20 requests

## Test Plan

    go test ./pkg/config/... ./pkg/dsl/...
    go build ./pkg/selection/... ./pkg/extproc/... && go vet ./pkg/selection/... ./pkg/extproc/...
    make config-schema-check
    cd dashboard/backend && go test ./recipe/...
    cd dashboard/frontend && npm run type-check && npm run lint
    cd src/vllm-sr && python3 -m pytest tests/ -q
    cd e2e && go build ./...

## Test Result

- `pkg/config`, `pkg/dsl`, `dashboard/backend/recipe`: ok
- `pkg/selection`, `pkg/extproc`: build and vet clean. Unit tests (`random_test.go`) can't link locally without the native cgo libraries, so CI runs them.
- `make config-schema-check`: passes
- Frontend: type-check clean, lint 0 errors
- pytest: the algorithm config and `serve --help` tests this PR extends pass. The full suite has 55 failures in runtime lifecycle, storage network, and serve restart tests, none related to algorithm selection.
- E2E: compiles only, not run locally. Needs a cluster whose config includes `random_route`.


FAUST-BENCHOU:
Could we avoid making the E2E gate probabilistic? Even though the all-same probability is tiny, this can still produce an irreproducible CI failure. The deterministic selector unit tests seem like a better place to verify uniform selection semantics, while E2E could focus on wiring/eligibility.

abhinav-m22:
Good call, thanks. Replaced the probabilistic check with a deterministic one: `decision-scoped-random` in the `routing-strategies` profile sends a keyword-matched request to a `random` decision and asserts the decision name, `x-vsr-selected-algorithm: random`, and that the selected model is one of the declared candidates. Uniform selection stays in the selector unit tests (`pkg/selection/random_test.go`), which use an injectable RNG. This also puts the test in a registered profile, so it no longer trips the E2E reachability gate.

Issue #3857, opened 2026-09-16
[Research] Fixed-policy controls for the first router-native candidate experiment


adaamko:
### Goal

The controls every learned candidate under #3198 has to beat: always-cheapest, always-strongest, and the best fixed split at matched cost, computed on the same-run harness format.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Implement the three controls on the exported split, reading the harness run format.
- Report accuracy, cost, and the per-request agreement metric against each control.
- Publish the numbers in the #3198 decision record.

### Non-goals

- New candidates.
- Changes to the run format once the harness lands.

### Completion signal

- Control numbers are reproducible from the run format.
- The candidate experiment reports against them.

Child of #3198 (the decision record); depends on the same-run harness issue. Parent Epic: #2974.

Roy214:
Hi @adaamko 

This task predicts one of three labels: AR (text), DIFFUSION (image), BOTH. These represent different routing requirements, so they aren’t swappable. “Cheapest” and “strongest” don’t quite map cleanly to this task.

The baselines that seem meaningful to me:

always guess the most common label
guess randomly, matching how often each label appears

A cost-based baseline would need token counts per request, which would come from #3856.

What do you think? Does that match what you had in mind, or were you thinking of cost differently for this task? Happy to go either way.

adaamko:
@Roy214 you're right, and the mismatch is in my issue text, not in your reading.

i wrote the three controls (always-cheapest, always-strongest, best fixed split at matched cost) for a cost-bearing routing task, where the labels are substitutable and the only question is which model you pay for. modality routing isn't that task. AR / DIFFUSION / BOTH are capability requirements, a diffusion request can't be served by a text model at any price, so "cheapest" has no meaning on this split. so take the cost controls off this issue.

what i'd want instead, in order of how much they can actually hurt a candidate:

1. **lexical baseline** — a short keyword/regex rule over the obvious surface cues (draw, image, picture, generate a photo, and the non-English equivalents in the split). this is the control that matters. if 30 lines of regex gets within a couple of points of a finetuned encoder, the encoder hasn't earned its slot in the router, and that's the first thing a reviewer will ask. the other two can't kill a candidate, this one can.
2. **majority class** — the accuracy floor.
3. **prior-matched random** — the floor for anything that looks at per-class behaviour.

[...]

Issue #3856, opened 2026-09-16
[Research] Same-run evaluation harness for router-native model candidates


adaamko:
### Goal

One harness that runs a candidate and its baseline on the same rows in the same process shape and reports quality, CPU and memory profile, and p99 latency with warm state and batch shape held fixed, so candidates under #3198 are compared on one run format.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Exported test split with immutable row ids.
- One run format, one record per request: row id, input hash, candidate output, baseline output, latency, memory high-water mark.
- Warm-up protocol and fixed batch shape, recorded with the run.
- A report script that emits the agreed metrics and paired uncertainty on the same examples.

### Non-goals

- Choosing the candidate.
- Training.
- Changes to the router runtime.

### Completion signal

- The harness runs the current BERT baseline end to end on the modality-routing split and produces the report.
- The fixed-policy controls and the candidate experiment consume its run format without changes.

Child of #3198 (the decision record). Parent Epic: #2974.

karthikeyan1592:
@ramkrishs @Roy214 @adaamko @Xunzhuo

The router has Prometheus / pprof for live traffic. What we still lack is a same-run eval: frozen test prompts, warmup, one prompt at a time, then p50/p99 + RAM + CPU + accuracy in one file, so a new model is compared to BERT on the same rows.

That is the gap #3856 was opened for. I ran BERT through a small helper on my host. Those ms are only mine. Please do not reuse my p99. Accuracy should be close; latency/RAM/CPU will follow your CPU/GPU/candle.

I want a process call:

PR the helper so you can replay the steps below with the same flags, or
Write the protocol first (short RFC / section on #3198) and PR the helper after you ACK.
This repo doesn’t seem to have an RFC process, so I’m assuming (1) unless you want (2). Either way we need a checked-in helper — a GitHub comment is not a repro.

Data (group’s, not mine)
Frozen 759-row test set already in #3891 (random_state=42):

https://github.com/vllm-project/semantic-router/blob/7aceb99114816f8b966e9ea921d70d3d34d3af70/src/training/model_classifier/modality_routing_classifier/exported_modality_routing_dataset/test.jsonl

src/training/model_classifier/modality_routing_classifier/exported_modality_routing_dataset/test.jsonl

BERT weights: llm-semantic-router/mmbert32k-modality-router-merged

Steps to copy (this is the framework)
[...]

Issue #3198, opened 2026-08-31
[Research] Define routing-native task contracts, baselines, and the first candidate experiment


Xunzhuo:
### Goal

Choose the first evidence-backed Router-native model experiment beyond the current BERT-family portfolio without committing to an architecture in advance.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Compare candidate task contracts such as classification, candidate scoring, model choice, policy scoring, retrieval, long-context, multimodal, and agentic signals.
- Define the unchanged BERT, embedding, reranker, or small-decoder baseline for each serious candidate.
- Select one first experiment using explicit expected routing value, data availability, evaluation quality, latency/memory budget, and runtime feasibility.
- Record the task contract, dataset/objective specification, candidate architecture rationale, graduation gate, and stop criteria.

### Non-goals

- Preselecting one model family because it is novel.
- Building a general-purpose foundation-model training program.
- Making a release commitment before the experiment passes its gate.

### Completion signal

- A reviewed decision record selects one task and one bounded candidate experiment.
- The baseline, dataset/objective, runtime budget, evaluation, and stop criteria are reproducible.
- Follow-up implementation is opened only for a candidate that clears the research gate.

Parent Epic: #2974

ramkrishs:
I left a comment on the parent epic (#2974) a few days back but I should really be putting this here instead, since this is the actual issue a maintainer would need to accept.

The angle I want to focus on here was router-native model families beyond BERT — specifically whether a distilled small classifier can hold up as a routing backbone against the current BERT-family baseline. That maps directly onto what this issue is asking for: pick one bounded candidate experiment with an explicit baseline and graduation gate, not a new architecture for its own sake.

What I'd propose as the first candidate, scoped the way the non-goals here ask for:

- Task contract: classification (reusing an existing router classifier task, e.g. the modality-routing classifier under `src/training/model_classifier/`, rather than inventing a new one)
- Baseline: the current mmBERT/BERT-family LoRA classifier for that task, unchanged
- Candidate: a distilled/compact student model trained against the same task
- Evaluation: accuracy parity plus p50/p99 latency through the actual inference path (candle-binding or onnx-binding), not just offline eval numbers
- Graduation gate: candidate has to match baseline accuracy within some agreed tolerance and beat it meaningfully on p99 latency, or it doesn't move forward — happy to have the tolerance itself be part of the decision record review

[...]

PR #3891, opened 2026-09-17
[Research] Add DistilBERT distillation candidate for modality routing (#3198)

ramkrishs:
Related #3198

## Summary
- Trains a compact DistilBERT student (66M params) via knowledge distillation from the production mmBERT-32K modality-routing classifier (307M params), plus a leakage-free "clean baseline" reproduction for fair comparison.
- Adds a fixed-split trainer (`modality_routing_fixed_split_trainer.py`, plain fine-tune + KD modes) and a three-way evaluation script (`evaluate_modality_candidate.py`) with a built-in test-set contamination check.
- Label handling fails closed. Evaluation and distillation both raise unless a checkpoint's `id2label` maps one-to-one onto AR / DIFFUSION / BOTH, the teacher's logits are reordered to canonical order, and the mapping used is logged. The shared helpers live in `modality_label_mapping.py` and are covered by `test_modality_label_mapping.py`.
- The trainer takes `--seed` (default 42) and gives the same result twice with the same seed. The numbers below come from runs made before that option existed.
- Adds DistilBERT LoRA target-module support to `common_lora_utils.py`.
- Includes the real exported train/val/test split (5,055 rows) and the resulting evaluation report.
- Adds `DECISION_RECORD.md`: task contract, both baselines, the candidate, dataset and leakage checks, the label audit, and the sections still pending on #3856 and #3857.
- Adds `label_audit/`: a blind re-judgement of the validation and test labels (rubric, a command-line tool over the small `audit_lib/` modules, checkpoints, reports, and a blinded sheet for a human spot-check).
- Structured for testing: the trainer takes one `TrainConfig`, one `ModalityTrainer` covers plain fine-tuning and distillation, and the pure logic (data helpers, losses, metrics and report assembly, label mapping) sits in small modules with no torch dependency. The code layout is described in the decision record.

## Results (10 epochs, RTX 3090)
| Model | Accuracy |
|---|---|
| published production baseline | 96.84% |
| clean baseline (mmBERT-32K, our repro) | 97.10% |
| candidate (DistilBERT) | 96.31% |

Candidate vs. clean-baseline agreement: 96.44%. Candidate is 4.6x smaller and trains ~2.5x faster. On the original labels the 0.79-point gap to the clean baseline is not significant (exact McNemar p = 0.31). One known weakness: it over-predicts `BOTH` on `AR`/`DIFFUSION` prompts (precision 0.897 vs recall 0.981 on that class).

**Label audit.** A blind re-judgement of all validation and test rows agrees with the original labels on 92.6% of rows in both splits (kappa 0.88). Almost all of the disagreement is `BOTH` becoming `AR`, mostly how-to prompts that the dataset's template generator labels `BOTH` on purpose. Whether the DistilBERT gap is real depends on what `BOTH` should mean: it is significant under strict labels (p = 0.007) and noise under inclusive labels (p = 0.15). I asked about this in the #3198 thread: https://github.com/vllm-project/semantic-router/issues/3198#issuecomment-5739294513. The judge is a single model in a single session, so the audit is not independent yet.

This is piece 3 of the 3-way split agreed in #3198 (harness: #3856, controls: #3857, both still open and owned by other contributors). The LFM2.5-Encoder and SCX Router exploration is in a separate draft PR, #3934, stacked on this one.

Draft while the `BOTH` policy question and the accuracy tolerance are still open.

## Test plan
- [x] Both trainer modes smoke-tested end-to-end on synthetic data
- [x] Full training + evaluation run on real data (RTX 3090)
- [x] `pytest` in `modality_routing_classifier/`: 92 tests pass in about 9 seconds, with no GPU, checkpoint or network (`pip install pytest` first)
- [x] Refactor checked for identical behaviour: seeded plain and distillation runs give identical losses and metrics to the code before the refactor, and the audit reports and review sheet are byte-identical
- [x] The merge step still produces a checkpoint that loads and maps its labels
- [x] Re-ran the full three-way evaluation after the refactor: every metric and per-example record matches the checked-in report
- [x] Smoke-tested distillation with a teacher on a small subset: same seed gives identical results, a different seed does not
- [x] Repo checks on the changed files: black 25.1.0, markdownlint, structure check, architecture check, supply-chain scan
- [x] Decision record (`DECISION_RECORD.md`) added
- [ ] Latency (#3856) and controls (#3857) sections pending those issues
- [ ] Seed sweep (several seeds per model) before reading the 0.79-point gap as real
- [ ] Decide what to do about the Alpaca rows (CC BY-NC 4.0) in the checked-in dataset; licenses are listed in the dataset README
- [ ] Human spot-check of the re-judged labels (`label_audit/human_review.tsv`)
- [ ] `BOTH` labelling policy and the accuracy tolerance agreed in the thread


FAUST-BENCHOU:
Should this fail closed when a checkpoint label is not in the canonical modality mapping? Falling back to the raw class id can silently produce a valid-looking AR/DIFFUSION/BOTH prediction even when id2label is incompatible, which would contaminate the comparison metrics.

ramkrishs:
@FAUST-BENCHOU good catch, yes it should fail closed. The `.get(p, p)` fallback and the silent dropping of non-canonical labels could have scored a checkpoint against the wrong classes without any error.

What I changed in `evaluate_modality_candidate.py`:
- The class-id remap now goes through label names only, and raises unless `id2label` is a one-to-one mapping onto `AR` / `DIFFUSION` / `BOTH`. Missing, partial, extra, duplicate or generic (`LABEL_0`...) labels all fail.
- It also checks that the classifier's output size is 3 and that `id2label` covers ids 0 to 2.
- The raw-id fallback is gone, so an unknown prediction id is an error.
- The resolved mapping is logged for every checkpoint, so the eval log shows which one was used.
- Added `test_evaluate_modality_candidate.py` (canonical order, a permuted order remapped by name, and six incompatible mappings that must fail).

I kept the name-based check and did not switch to raw ids. That matches how the serving side reads `idx_to_label` from the checkpoint's mapping file, and ids alone would break silently on a checkpoint that orders its classes differently. The published model and both local checkpoints all use `0=AR, 1=DIFFUSION, 2=BOTH`, so the numbers already in the PR don't change.

This is committed on my side and I'll push it with the next update.

src/training/model_classifier/modality_routing_classifier/label_audit/audit_lib/report.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns:
+        Predicted label names per model name.
+    """
+    preds: dict[str, list[str]] = {}
+    if eval_report:
+        with open(eval_report, encoding="utf-8") as f:
+            recs = json.load(f)["per_example_records"]
+        for name in BASELINE_MODEL_KEYS:
+            if recs and f"{name}_pred" in recs[0]:
+                preds[name] = [r[f"{name}_pred"] for r in recs]
```
Xunzhuo: [P2] Validate prediction row identity before re-scoring

An evaluation report from a different export or row order is accepted whenever its length matches the split: this discards both `row_index` and `input_hash_sha256` and joins predictions by list position. In a two-row fixture, reversing a report containing perfect predictions produces a successful report with 0% accuracy instead of detecting the mismatch. Please align and validate each prediction against the loaded row identity, rejecting mismatched input hashes before computing the accuracy and McNemar results.

src/training/model_classifier/modality_routing_classifier/label_audit/judge_labels.py
```diff
@@ -0,0 +1,509 @@
...
+        args: Parsed CLI arguments.
+
+    Returns:
+        The split's rows.
+
+    Raises:
+        DatasetMismatchError: If the default dataset differs from the pinned one.
+    """
+    path = Path(args.data_dir) / f"{args.split}.jsonl"
+    if path.resolve().parent == DATA_DIR.resolve():
```
Xunzhuo: [P2] Keep checkpoint pinning for custom dataset directories

For every supported `--data-dir` other than the bundled directory, this condition skips the dataset hash guard entirely. After judgments have been saved for a custom export, replacing or reordering that split silently attaches those judgments to different prompts; a two-row reproduction returns a successful report with 0/2 agreement rather than `DatasetMismatchError`. Please bind the selected checkpoint to its dataset hash for custom inputs too, for example with a manifest beside that checkpoint, and verify it on every subsequent command.

Xunzhuo (changes requested):
Two audit paths can silently attach predictions or judgments to the wrong prompts. Please preserve row identity when loading evaluation reports and apply the dataset hash guard to custom input directories.

The 91-test modality suite passes, and a tiny CPU integration exercised plain training, distillation, deterministic repeat training, adapter merge/reload and three-way evaluation. Recomputing the checked-in 759-row metrics from the stored predictions matched the report. Full-size checkpoint inference, GPU training and the live judge API were not run.

src/training/model_classifier/modality_routing_classifier/label_audit/audit_lib/report.py
ramkrishs: Fixed in 814a1bf04, thank you. This was a real bug.

The report tool now aligns every prediction by `row_index`, checks its `input_hash_sha256` against the sha256 of the loaded row's text, and requires the predictions to cover exactly the rows of the split. Anything else raises `PredictionIdentityError` and the command exits 1, so a reversed report, a report from another export, or a report for a split of a different size can no longer be scored. That last case used to be dropped silently by a length check, so it is now an error too.

The same applies to `--preds` files. A bare list of labels carries no identity, so those files must have `input_hashes` (the sha256 of each prompt, in order). I added them to the two committed prediction files, after checking that their accuracy against the test labels still reproduces from row order (99.60% and 98.16%).

Tests cover your two-row case (a report made on the reversed rows is rejected instead of scoring 0%), a wrong-size report, a repeated `row_index`, shuffled records that still align by index, and a predictions file with no hashes. The checked-in audit reports are unchanged.

src/training/model_classifier/modality_routing_classifier/label_audit/judge_labels.py
ramkrishs: Fixed in 814a1bf04, thank you. Skipping the guard for custom directories was wrong.

Every command that loads a split now checks it against a manifest bound to the checkpoint: `dataset_sha256.json` for the bundled checkpoint, and `<checkpoint>.dataset_sha256.json` beside any other one. The first use pins the file, and later uses must match, so replacing or reordering a judged split raises `DatasetMismatchError` and the command exits 1.

I went one step further than the reproduction. If judgments already exist for a split but no hash was recorded (for example a checkpoint from before pinning), the command refuses to run, because there is nothing to check them against.

Tests cover a custom `--data-dir` being pinned next to its checkpoint, a replaced split and a reordered split each being refused by `report`, `next` and `sheet`, and the missing-manifest case.

Xunzhuo (approved):
The audit now binds prediction rows to their prompt hashes and pins checkpoint datasets, addressing both previous findings. Validation includes 52 audit tests and matching hashes for all 1,518 prediction rows; the unchanged training paths retain the earlier tiny-CPU lifecycle checks. Exact-head CI passes and current-main integration is clean. This approves the implementation, without claiming full GPU training or production-model quality validation.

wilsonwu (approved):
LGTM

PR #3906, opened 2026-09-18
[Bug] Skip multimodal embedding init when a recipe has no contrastive jailbreak rules

### LINKED ISSUE #3907: [Bug] Router fails to start with model_type=multimodal when a recipe has no contrastive jailbreak rule
## Symptom

With `global.model_catalog.embeddings.semantic.embedding_config.model_type=multimodal`, the router fails at startup for any recipe that has no `method: contrastive` jailbreak rule:

```
create ExtProc server: failed to build recipe classifiers: build routing recipe "default":
contrastive jailbreak with model_type=multimodal: embedding model "multimodal" was not prepared for this generation
```

The `multimodal-routing` Kubernetes E2E profile hits this on every run (first seen on #3772: https://github.com/vllm-project/semantic-router/actions/runs/35300412712/job/105481230156). The profile declares image-modality embedding rules and no jailbreak rules.

## Cause

`buildContrastiveJailbreakClassifiersOption` in `src/semantic-router/pkg/classification/classifier_option_rules.go` calls `initMultiModalIfNeeded` whenever `model_type=multimodal`, before checking whether any contrastive rule exists. Since the recipe embedding set became strict (#3768, #3876), `EmbeddingModelsNeeded` only prepares the primary embedding for a contrastive rule, so a recipe without one has no `multimodal` provider and `Set.Get` fails.

## Fix

Return early from the contrastive jailbreak builder when the recipe has no contrastive rule, so the multimodal initialization only runs when a rule needs it. PR #3906.

drivebyer:
Closes #3907

## Problem

`buildContrastiveJailbreakClassifiersOption` initializes the `multimodal` embedding whenever `embedding_config.model_type=multimodal`, before checking whether the recipe has any `method: contrastive` jailbreak rule. Since #3768/#3876 the recipe embedding set is strict: `EmbeddingModelsNeeded` only registers the primary embedding for a contrastive rule, so a recipe with no jailbreak rules gets an empty set and the router fails at startup:

```
create ExtProc server: failed to build recipe classifiers: build routing recipe "default":
contrastive jailbreak with model_type=multimodal: embedding model "multimodal" was not prepared for this generation
```

This breaks the `multimodal-routing` Kubernetes E2E profile (first observed on #3772: https://github.com/vllm-project/semantic-router/actions/runs/35300412712/job/105481230156). The profile only declares image-modality embedding rules and no jailbreak rules.

## Fix

Collect the contrastive rules first and return early when there are none. The multimodal initialization and the per-rule build now run only when a contrastive rule exists, matching what `EmbeddingModelsNeeded` prepares.

## Tests

- `TestContrastiveJailbreakSkipsMultimodalInitWithoutContrastiveRules`: `model_type=multimodal` with no rules or classifier-only rules against an empty embedding set returns no option and no error.
- `TestContrastiveJailbreakStillRequiresMultimodalEmbeddingForContrastiveRules`: a contrastive rule against the same empty set still fails with the unprepared-model error.


Xunzhuo (approved):
The guard matches actual contrastive-rule consumers and preserves initialization for configured contrastive rules. The two new Core tests pass. A full NewClassifier startup regression also passes with real native libraries; the same fixture against the pre-fix implementation fails with the original unprepared-multimodal error. No live multimodal inference or Kubernetes multimodal deployment was run.

PR #3908, opened 2026-09-18
[Bug] looper: fill verifier_version in attempt traces and report faithfulness as tie (#2857)

Bevisy:
Fills the remaining traceability gap of the shared Looper outcome-verifier contract (#2857, slice 2) and removes the one place where a verifier decided for its caller. Two files of behavior, two files of tests; no field, enum, config, or public schema is added.

Related #2857

## Why this is needed

- The attempt trace already publishes `verifier_type` and `verifier_version` (OpenAPI schema, `website/static/openapi/apiserver/apiserver.openapi.json:20130-20135`), but the confidence AutoMix path only ever wrote the confidence method name and left `verifier_version` empty on every request. A declared field that can never carry a value is an audit hole: a Replay reader cannot tell which verifier implementation produced a score. Filling it is the "version traceable" clause of the issue's completion signal, done with the existing field.
- `FaithfulnessVerifier` returned `approve` for every candidate although it only measures unsupported spans against trusted context and takes no accept decision. That made a scorer claim policy validity and contradicted the contract constraint stated in this issue ("a verifier must not silently remove candidates or become a second selector"). It now reports `tie`, exactly like the peer-consistency verifier in the same file, and the calling algorithm keeps accept, rerank, escalate, or stop policy.

## Purpose

- `confidence.go`: the AutoMix attempt records `verifierVersion: result.Version` (shared verifier identity) alongside the existing method name.
- `verifier_grounding.go`: `FaithfulnessVerifier.Verify` reports `DispositionTie`; the doc comment states that it measures evidence and never decides.
- No routing behavior change: the only caller reads `Scores`, and `verifier_type` keeps its existing meaning (confidence method name).

## Test Plan

- `go test -count=1 ./pkg/looper/` with the native library paths set (see Test Result)
- `gofmt -l` on the touched files; `go vet ./pkg/looper/`
- New `TestAutoMixAttemptTraceCarriesVerifierIdentity` asserts one `external_verifier` attempt carrying the method-name `verifier_type` and `verifier_version = automix/1`; `TestFaithfulnessVerifierScoresBySpanCount` now asserts `tie`.

## Test Result

- `go test -count=1 ./pkg/looper/` → `ok ... pkg/looper 16.187s` (whole package, native libraries linked through `LD_LIBRARY_PATH` / `CGO_LDFLAGS` over the four `*/target/release` directories built by `make test-semantic-router`).
- `gofmt -l pkg/looper/`: empty; `go vet ./pkg/looper/`: pass.
- `make test-semantic-router`: pass. Note it does not include the `pkg/looper` package (the repository only has benchmark targets for it) and CI's plan job skips Tests / Packages for this change, so the package test above is the run that covers this diff.
- Pre-commit hooks pass: go fmt, supply-chain security scan, structure check, architecture dependency check.

<details>

- [x] PR title begins with exactly one bracketed category: `[Bug]`
- [x] The title does not stack prefixes; affected module (looper) is named in the title and body
- [x] The PR links an `accepted` issue with exactly one owner (`Related #2857`)
- [x] No config surface, enum, or public API change: only existing trace fields are filled and one disposition is corrected
</details>


Xunzhuo (approved):
The AutoMix trace now carries the existing verifier version, and faithfulness scoring leaves the decision to its caller. Both targeted regressions passed in Core CI, along with the looper integration profile. I found no blocker. Live model verification was not run.

wilsonwu (approved):
LGTM

PR #3909, opened 2026-09-18
[CI] Share Go 1.25 setup across PR workflows

### LINKED ISSUE #3289: [CI] Share one Go 1.25 setup composite and drop dead Core toolchain steps
### Goal

Give PR Go jobs one shared toolchain setup, and stop installing or re-running Core receipts that another job already owns.

Parent: #3235
Parent Epic: #3177

### Proposed Workgroup

Evaluation & Quality

### Scope

- Add a composite action (for example `.github/actions/setup-go-ci`) that pins Go 1.25 and the shared module cache.
- Replace inlined `actions/setup-go@v5` + `go-version: "1.25"` in the PR-path workflows, including the repeated checkout/setup/`go mod download` blocks in `operator-ci.yml`.
- Remove the `golangci-lint-action` step in `.github/workflows/test-and-build.yml` that only runs `--help`. Core lint already runs in pre-commit and Operator CI.
- Remove the extra `make check-go-mod-tidy` step in `test-and-build.yml`; `make test` already runs it.
- Keep quality, Core tests, Operator lint, and Operator unit tests as separate receipts.

### Non-goals

- Merging the Core quality and Core test jobs into one runner.
- Changing golangci-lint version or lint rules.
- Sharing built images between Operator and E2E.

### Completion signal

- PR workflows consume one Go setup composite instead of copy-pasted `setup-go` blocks.
- `test-and-build.yml` no longer installs golangci-lint or runs `check-go-mod-tidy` twice.
- Existing Core, Operator, and quality jobs stay green on an unaffected and an Operator-touching PR path.

yigitcan-ozturk:
Fixes #3289

## Summary
- add a shared `.github/actions/setup-go-ci` composite pinned to Go 1.25 with setup-go module caching
- replace repeated Go setup/cache blocks across the PR-path Core, quality, generated-contract, storage, native, recipe, dashboard, component, and Operator workflows
- remove the dead Core/quality `golangci-lint --help` setup where lint is already owned by the quality/Operator receipts
- stop invoking `check-go-mod-tidy` separately in Core tests; `make test`/the owned test path already covers module tidiness

## Scope
This intentionally does not change golangci-lint rules or versions, merge Core quality/tests into one runner, or change Operator/E2E image sharing. Existing job responsibilities remain separate.

The branch is based directly on current upstream `main` to avoid carrying the stale workflow layout from the older fork default branch.


.github/workflows/check-generated.yml
```diff
@@ -23,10 +23,7 @@ jobs:
       CI: "true"
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
-        with:
-          go-version: "1.25"
-      - uses: actions/setup-python@v5
+      - uses: ./.github/actions/setup-go-ci\n      - uses: actions/setup-python@v5
```
Xunzhuo: [P1] Replace the literal \n sequences with real YAML line breaks. This line merges two steps into one scalar and fails parsing; the same replacement pattern affects the other modified workflows (including the recipe action path). The workflow validation gate cannot load these reusable workflows, so the PR pipeline cannot run. Fix every affected call site, then run make harness-check.

.github/workflows/dashboard-test.yml
```diff
@@ -47,25 +43,7 @@ jobs:
...
-          key: ${{ runner.os }}-go-dashboard-${{ hashFiles('dashboard/backend/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-dashboard-
-
-      - name: Resolve golangci-lint version
-        id: golangci_lint
-        run: echo "version=v$(cat tools/linter/go/golangci-lint.version)" >> "$GITHUB_OUTPUT"
 
-      - name: Set up golangci-lint
-        uses: golangci/golangci-lint-action@v7
```
Xunzhuo, line 64: [P1] Keep the pinned golangci-lint installation or explicitly bootstrap it before dashboard-check. The new setup-go-ci action installs only Go, while dashboard-check calls dashboard-lint, which runs golangci-lint directly and does not invoke harness-go-bootstrap. After the YAML is repaired, a clean Dashboard runner will fail at backend lint because this removal leaves the binary unavailable.

Xunzhuo (changes requested):
The shared Go setup currently breaks the reusable workflow YAML and removes the Dashboard linter installation. The workflow validation gate reproduces the parsing failure. Please fix both call-site formatting and linter provisioning before rerunning CI.

.github/workflows/operator-ci.yml
abhinav-m22: Nit: with the composite pinning 1.25, `env.GO_VERSION` isn't referenced anymore. Worth removing so there's only one place to bump?

.github/workflows/test-and-build.yml
```diff
@@ -85,8 +75,8 @@ jobs:
         if: inputs.native-artifact == ''
         run: make rust-ci test-rust-ci
 
-      - name: Check Go modules and vet source
-        run: make check-go-mod-tidy vet
+      - name: Vet Go source
+        run: make vet
```
abhinav-m22: `make test` does include `check-go-mod-tidy`, but this workflow calls targets individually. After dropping it here, does any job still run the tidy check?

.github/workflows/test-tools.yml
```diff
@@ -20,10 +20,8 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-      - uses: actions/setup-go@v5
+      - uses: ./.github/actions/setup-go-ci
```
Xunzhuo, line 23: [P2] Update the workflow contract tests for the shared action. `test_component_batches.py` still searches this job for `actions/setup-go@v5`, and `test_deployment_test_results.py` still counts that direct action in the operator checks. With these replacements, `make harness-check` fails with `StopIteration` and `0 != 1`, as reproduced by the current Router Tools job. Update both assertions to verify the composite action and its conditional setup contract, then rerun the harness.

Xunzhuo (changes requested):
The previous YAML and Dashboard linter problems are fixed. Two workflow contract tests still require direct setup-go steps, so the shared-action migration fails the harness. Please update those tests along with the callers.

yigitcan-ozturk:
Addressed the requested workflow-contract follow-ups. The shared-action contract assertions now cover the conditional setup in `test-tools.yml` and the Operator `go.sum` cache dependency path; `GO_VERSION` is no longer duplicated in `operator-ci.yml`, and Core still runs `make check-go-mod-tidy vet` explicitly.

The latest PR run confirms the previous contract blocker is cleared: the relevant quality/harness checks pass. The only remaining failure is the Production Benchmarks regression gate (`BenchmarkReMoM_DistributeByWeight`, B/op 1184 → 1280), which is outside this workflow-only diff and appears unrelated to the Go setup migration. I don't have permission to re-run that upstream job from the fork/API.

Could you please re-review when convenient? If you'd like any additional workflow-contract coverage in this PR, I'm happy to add it.

abhinav-m22:
@yigitcan-ozturk Please fix CI: DCO is failing (sign off your commits with `git commit -s --amend` and force-push), along with Gate Product Verification and Performance.

Xunzhuo (commented):
The earlier workflow and contract-test failures are fixed. The shared setup passes the exact-head harness, including all 14 affected regressions, with Operator and recipe CI also passing. One non-blocking cache-key correction remains below: the shared default omits the Router module checksum. No local product deployment was rerun.

.github/actions/setup-go-ci/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Set up Go CI
+description: Install the shared Go 1.25 toolchain and cache module downloads.
+
+inputs:
+  cache-dependency-path:
+    description: Newline-separated go.sum paths used for the setup-go cache key.
+    required: false
+    default: |
+      go.sum
```
Xunzhuo: [P2] Include the Router module checksum in the shared cache key

The Router module now lives at `src/semantic-router/go.sum`; there is no root `go.sum` in this tree. The default therefore hashes only the Operator and Dashboard modules. This also replaces the Core/quality caches that previously hashed `**/go.sum`, so changing only Router dependencies leaves the cache key unchanged. setup-go skips saving after an exact cache hit, causing those newly downloaded modules to be downloaded again on every subsequent run until another included module changes. Use the actual Router go.sum path (or a deliberate multi-module glob) in the shared default.

yigitcan-ozturk:
The cache-key follow-up is now addressed at `e210040`: the shared `setup-go-ci` default includes `src/semantic-router/go.sum` alongside the Operator and Dashboard module checksums. The latest Pull Request workflow on this exact head completed successfully, and Codecov reports all modified coverable lines covered.

This also incorporates the earlier workflow-contract fixes and preserves the explicit Core tidy/vet checks. Ready for re-review when convenient.

PR #3912, opened 2026-09-18
[Feature] Build a reproducible shadow comparison dataset manifest

subin9:
## Purpose

Related #3280

Shadow dispatch records each candidate answer as a replay outcome, but nothing turns those observations into a dataset. A comparison is evidence only when the same approved inputs rebuild the same rows and the same split every time.

`shadowdataset.Build` emits a manifest of input, primary, and shadow identity with output digests and lineage. `Validate` rejects a manifest that was edited or that puts one input in two splits.

Three choices worth arguing with.

The manifest holds no prompt or response text, only hashes, so it can be published beside the numbers it supports. It holds no build time either, since the digest is its identity and a timestamp would make two runs over the same records differ.

An observation enters whole or not at all. A record whose bodies were cut, cleared by the replay redactor, or never captured is excluded under a counted reason, because the digest of a fragment claims the model produced the fragment.

Splits follow a seeded hash of the example identity rather than record order, so later observations never move an example that was already placed.

Blinded judging, configuration, and storage destinations are the next slice. Nothing here reads a live route or writes one.

## Test Plan

Build a manifest from captured records, then reorder those records, add more, and change the seed. Damage an observation one way at a time. Edit a finished manifest.

## Test Result

The same records in any order give the same manifest. Adding observations leaves every earlier example in the split it already had, and a different seed places them differently. Each way an observation can be partial is excluded under its own counted reason. An edited manifest, and one carrying a single input in two splits, are both rejected.


src/semantic-router/pkg/shadowdataset/manifest.go
```diff
@@ -0,0 +1,420 @@
...
+	// not true. Drop it rather than publish a hash that cannot be reproduced by
+	// rerunning the model.
+	if rec.ResponseBodyTruncated {
+		return Example{}, ExcludeResponseTruncated
+	}
+
+	inputDigest := digestOf(rec.RequestBody)
+	primary := Arm{
+		Model:        rec.SelectedModel,
+		OutputDigest: digestOf(rec.ResponseBody),
```
Xunzhuo: [P2] Hash the same output representation for both arms

`rec.ResponseBody` is the encoded protocol response retained by `attachRouterReplayResponse`, but `shadowOutcomeMetadata` hashes `semanticResponseText(decoded.Response)`. Consequently, primary and shadow models both answering `hello` receive different output digests: the primary hash includes the JSON envelope, response ID and usage, while the shadow hash covers only `hello`. Normalize the primary output with the same contract as the shadow digest, or capture matching digests for both arms at the producer boundary, and cover an actual protocol-shaped response.

src/semantic-router/pkg/shadowdataset/manifest.go
```diff
@@ -0,0 +1,420 @@
...
+	if rec.SelectedModel == "" {
+		return Example{}, ExcludePrimaryModelMissing
+	}
+	if rec.RequestBody == "" {
+		return Example{}, ExcludeRequestMissing
+	}
+	if rec.RequestBodyTruncated {
+		return Example{}, ExcludeRequestTruncated
+	}
+	if rec.ResponseBody == "" {
```
Xunzhuo: [P2] Exclude failed primary responses from comparison examples

A completed shadow can accompany a primary HTTP 400/429/503. The router preserves those error bodies in replay and marks the record failed, so they pass these nonempty/nontruncated checks and are published as the primary model’s answer. This treats a primary transport failure as model evidence even though failed shadow arms are excluded. Check the primary lifecycle/status before constructing the example, count unsuccessful observations under an explicit exclusion reason, and add a failed-primary/completed-shadow case.

Xunzhuo (changes requested):
Two correctness issues prevent the manifest from representing comparable model outputs: the primary and shadow hashes use different representations, and a failed primary response can become a model answer. The existing package tests pass, while two additional regression cases reproduce these failures with protocol-shaped replay responses. Please align the output representation and exclude failed primary observations. Full router runtime validation has not been run.

src/semantic-router/pkg/shadowdataset/manifest.go
subin9: Fixed. A finished response now records the assistant text of the decoded answer as an outcome, which is the same thing a shadow arm is hashed from, so the manifest reads both arms from there and no longer hashes the stored body. A regression holds two models answering `hello` to one digest, against a protocol-shaped response carrying an id and a usage block.

src/semantic-router/pkg/shadowdataset/manifest.go
subin9: Fixed. An example now needs a completed lifecycle and a 2xx status, so a failed primary is excluded and counted under `primary_request_unsuccessful`, the way a failed shadow arm already was. The cases cover 400, 429 and 503, an unfinished record and a failed lifecycle, each beside a completed shadow.

subin9:
Unrelated to the two review points above: the five `shadow-dispatch-*` cases in the `router-replay` E2E profile fail on this branch, and they fail on current main the same way. I traced it to #3893 and filed #3914. The profile is `selection: manual`, so no CI lane was running it.

Nothing here depends on those cases. The manifest reads outcomes the caller hands it, and its own tests build them directly.

src/semantic-router/pkg/extproc/recorder.go
```diff
@@ -774,3 +780,43 @@ func (r *OpenAIRouter) updateRouterReplayUsageCost(ctx *RequestContext, usage ro
...
+// attachPrimaryOutputDigest records the primary answer under the same contract
+// shadow arms use: the assistant text of the decoded response, not the encoded
+// protocol body. The stored body carries the JSON envelope, the response id and
+// usage, so hashing it would give two models different digests for the same
+// answer.
+func attachPrimaryOutputDigest(ctx *RequestContext, recorder *routerreplay.Recorder) {
+	if ctx.SemanticResponse == nil {
+		return
+	}
+	text := semanticResponseText(*ctx.SemanticResponse)
```
Xunzhuo: [P2] Capture the model digest before body-warning rewrites

On buffered responses, `handleNonStreamingResponseBody` calls `applySemanticResponseWarnings` before `attachRouterReplayResponse`. A `hallucination_action: body` or `unverified_factual_action: body` prepends router text to this same `ctx.SemanticResponse`, so this hashes the warning plus the answer. The shadow path hashes the decoded upstream answer without those plugins: two models both returning `hello` still get different output digests when the primary receives a body warning. Capture an immutable digest before these response mutations and persist that at finalization, with a regression covering the body-warning path.

Xunzhuo (changes requested):
The two earlier issues are fixed: both arms now use assistant-text digests, and unsuccessful primary requests are excluded. The updated regression tests pass, but a further buffered-response case still hashes router-added body warnings as part of the primary model output. A focused regression reproduces the mismatch. Please capture the digest before response mutations and cover that path. The manual router-replay E2E profile was not rerun.

src/semantic-router/pkg/extproc/recorder.go
subin9: Fixed. The digest is now taken where both response paths still hold the decoded upstream answer, before any rule can rewrite it, and the recorder persists that value. A regression drives the buffered path with a body warning and holds the digest to the model answer alone.

Xunzhuo (dismissed):
The primary and shadow arms now hash the same model output before response warnings modify it, and failed primary requests are excluded. The buffered-warning, streaming-digest and manifest regressions pass on this head.

subin9:
Pushed one more commit after the approval, which dismissed it. Sorry for the round.

It takes the primary model name from the record alone instead of preferring the outcome's target_ref. The name is part of the example id and the id picks the split, so one source is safer than two. After #3620 both sources hold the same value on this path, so nothing observable changes.

abhinav-m22:
@subin9 Seems this has merge conflicts with main. Could you please rebase on latest main and resolve them.

subin9:
Resolved : ) thanks!

FAUST-BENCHOU:
The revised evidence boundary looks much stronger now. Primary and shadow outputs use the same assistant-text digest contract, and capturing the primary digest before response-stage mutation avoids attributing router-added warnings to the model. The explicit exclusion path for incomplete/failed observations also keeps the manifest semantics clean. LGTM from my review.

Xunzhuo (approved):
The manifest and primary-digest fixes are ready. The reviewed head passes all 28 targeted manifest, digest, streaming and memory regressions within the Core suite, plus the streaming and local-memory CI profiles. The patch integrates cleanly with current main, and the earlier provenance findings are fixed. No full local deployment or model rerun was performed.

wilsonwu (approved):
LGTM

PR #3915, opened 2026-09-18
[Bug] Start the replay record before dispatching the shadow

### LINKED ISSUE #3914: [Bug] Deferred replay start leaves every shadow dispatch outcome unrecorded
### Proposed Workgroup

MoM & Routing

### Describe the bug

Shadow dispatch runs, calls the candidate, and produces an outcome, but the outcome is never written to the replay record. All five `shadow-dispatch-*` cases in the `router-replay` E2E profile fail because the record they poll carries an empty outcome list.

The cause is an ordering change. `dispatchShadowIfConfigured` copies `ctx.RouterReplayID` into the shadow job when the job is built (`shadow_dispatch.go`), and `record` then skips persistence entirely when that id is empty:

```
if job.recorder != nil && job.replayID != "" {
```

In `processor_req_body.go` the replay start is now deferred, so it runs when the function returns, which is after the shadow job was already built:

```
239:  defer r.startRouterReplay(ctx, originalModel, matchedModel, decisionName)
247:  r.dispatchShadowIfConfigured(ctx, dispatch)
```

Before #3893 the same two lines ran in the other order, with the replay start first:

```
237:  r.startRouterReplay(ctx, originalModel, matchedModel, decisionName)
245:  r.dispatchShadowIfConfigured(ctx, dispatch)
```

The router log shows the shadow call completing and the id missing:

```
"msg":"shadow_dispatch_outcome","replay_id":"","shadow_model":"openai/shadow-malformed",
```

### To Reproduce

```
make e2e-test E2E_PROFILE=router-replay E2E_KEEP_CLUSTER=true
```

I ran this three times against one kind cluster, changing only the checkout:

| checkout | shadow cases |
| --- | --- |
| the commit before #3893 | 5 passed, each in about 8 seconds |
| current main | 5 failed, each after a 53 second deadline |
| a feature branch off current main | 5 failed, the same way |

The other five cases in the profile pass in all three runs. They are the ones that never read an outcome.

### Expected behavior

A completed shadow arm leaves an outcome on the replay record, and a failed one leaves its verdict and reason, which is what these cases assert and what they got before #3893.

### Impact

The shadow dispatch feature records nothing, so any consumer of those outcomes sees an empty list rather than a missing feature. `tools/agent/domains.yaml` marks this profile `selection: manual`, so no CI lane runs it and nothing caught the regression.

### Additional context

There is a second `dispatchShadowIfConfigured` call earlier in the same file, on the image-file dispatch path, which sat ahead of any replay start before #3893 as well. That path may never have recorded an outcome, and it is worth checking alongside the fix rather than assuming this regression is the only case.

subin9:
Closes #3914

## Purpose

A shadow job copies the replay id when the job is built, and the recording step skips persistence when that id is empty, so the outcome is dropped and nothing reports an error. Once the replay start became deferred in #3893 the job was built first, and every shadow outcome stopped reaching the record.

The dispatch now runs inside the same deferred block, after the record exists. It is guarded on finalization having succeeded, so a request that failed to finalize still raises no shadow, which is what the old ordering gave. The record is still written after automatic output is resolved, and still survives a failure or a panic, which is why #3893 deferred it.

Every existing shadow case handed the router a replay id that already existed, so none of them ran the order the request path uses. That is the gap this went through. The new case lets the router assign the id and asserts the outcome lands on the record the router created.

I left the earlier `dispatchShadowIfConfigured` on the image-file path alone. It sat ahead of any replay start before #3893 as well, no case covers it, and I have not established whether it ever recorded an outcome.

## Test Plan

Run the new case with and without the fix. Run the `router-replay` E2E profile, which is `selection: manual` and so runs nowhere else.

## Test Result

Without the fix the new case reports no shadow outcome on the record the router created. With it the outcome is there.

The `router-replay` profile passes all ten cases on this branch. On current main the same five shadow cases fail, each after a 53 second deadline, and here each passes in about 8 seconds, which matches what the profile did before #3893.


drivebyer (dismissed):
LGTM

wilsonwu (dismissed):
LGTM

abhinav-m22 (approved):

PR #3917, opened 2026-09-18
fix: router-owned response extensions bypass strict protocol translation (#3524)

### LINKED ISSUE #3524: [Bug] Traced Looper responses fail strict protocol translation and return 502
### Proposed Workgroup

MoM & Routing

### Describe the bug

Looper algorithms attach a router-owned top-level field to the same JSON body
they hand to the response codec: Fusion adds `fusion`, Workflows adds `flow`,
and ReMoM adds `reasoning_mom_responses`. `prepareLooperResponse` then passes
that body to `engine.TranslateResponse`, which decodes it as an untrusted
provider response under the strict policy. The codec correctly rejects the
router's own field as non-canonical, and the request fails with
`502 upstream_unavailable`.

The router generates the body, hands it to itself as if it came from a provider,
rejects its own documented extension, and reports the upstream as unavailable.

Fusion and Workflows default their trace controls to enabled, and ReMoM's
runtime default enables intermediate responses. The canonical reference config,
the Fusion and Workflows fragments, and the accuracy recipe all contain enabled
trace paths, while other shipped recipes — `multi-objective`, built-in
`mom-v1`, and the ReMoM fragment — explicitly disable some or all of them. Any
path that emits one of these router-owned fields reaches the same strict
response boundary.

The defect exists on `upstream/main` and is independent of any open PR. It was
discovered while implementing #3375. A fallback implementation that carries
below-quorum failed-model or quorum evidence through the existing `fusion`
response extension will exercise this defect. Issue #3375 itself requires
bounded Replay and metrics telemetry, not a caller-visible trace; whether
fallback evidence should also be returned to the caller is a separate design
decision for that feature.

### To Reproduce

Reproduced on `upstream/main` at `63e91c21`, arm64 kind cluster via colima.

The `looper` profile has no case that exercises a *served* Fusion synthesis —
every existing fusion case is a below-quorum failure that never reaches the
response-encode path — so the reproduction adds a keyword signal, a decision, a
registered E2E case, and the profile test list entry. It changes **no production
code**; the added Go file is test source.

<details>
<summary>Minimal reproduction patch against <code>63e91c21</code></summary>

```diff
diff --git a/e2e/profiles/looper/profile.go b/e2e/profiles/looper/profile.go
index c293bb47..09b6fd43 100644
--- a/e2e/profiles/looper/profile.go
+++ b/e2e/profiles/looper/profile.go
@@ -62,6 +62,7 @@ func (p *Profile) GetTestCases() []string {
 	return []string{
 		"looper-ratings-happy-path",
 		"looper-fusion-usable-quorum",
+		"looper-fusion-synthesis-trace",
 	}
 }
 
diff --git a/e2e/profiles/looper/values.yaml b/e2e/profiles/looper/values.yaml
index 64f358a1..67dc6962 100644
--- a/e2e/profiles/looper/values.yaml
+++ b/e2e/profiles/looper/values.yaml
@@ -61,6 +61,11 @@ config:
           keywords:
             - __LOOPER_FUSION_QUORUM_PROBE__
           case_sensitive: true
+        - name: looper_fusion_synthesis_probe_keywords
+          operator: OR
+          keywords:
+            - __LOOPER_FUSION_SYNTHESIS_PROBE__
+          case_sensitive: true
     decisions:
       - name: looper_ratings_decision
         description: Deterministic Ratings route for issue 2859 E2E coverage
@@ -106,6 +111,34 @@ config:
             max_concurrent: 3
             min_successful_responses: 2
             on_error: skip
+      - name: looper_fusion_synthesis_decision
+        description: Serve an above-quorum Fusion synthesis carrying its trace
+        priority: 115
+        rules:
+          operator: OR
+          conditions:
+            - type: keyword
+              name: looper_fusion_synthesis_probe_keywords
+        modelRefs:
+          - model: fusion-panel-valid
+            use_reasoning: false
+          - model: fusion-panel-fail
+            use_reasoning: false
+        algorithm:
+          type: fusion
+          fusion:
+            model: fusion-judge
+            analysis_models:
+              - fusion-panel-valid
+              - fusion-panel-fail
+            max_concurrent: 2
+            # The single usable panel member meets quorum, so the judge runs and
+            # this is an ordinary served synthesis. include_analysis and
+            # include_intermediate_responses default to true, so it carries a
+            # trace -- the shape no case covered before, and the one that proves
+            # the response-extension boundary end to end.
+            min_successful_responses: 1
+            on_error: skip
       - name: default-route
         description: Fallback for unmatched profile traffic
         priority: 10
diff --git a/e2e/testcases/looper_fusion_synthesis.go b/e2e/testcases/looper_fusion_synthesis.go
new file mode 100644
index 00000000..05e4b0af
--- /dev/null
+++ b/e2e/testcases/looper_fusion_synthesis.go
@@ -0,0 +1,87 @@
+package testcases
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	pkgtestcases "github.com/vllm-project/semantic-router/e2e/pkg/testcases"
+)
+
+const looperFusionSynthesisProbeKeyword = "__LOOPER_FUSION_SYNTHESIS_PROBE__"
+
+func init() {
+	pkgtestcases.Register("looper-fusion-synthesis-trace", pkgtestcases.TestCase{
+		Description: "Serve an above-quorum Fusion synthesis carrying its trace",
+		Tags:        []string{"kubernetes", "routing", "looper", "fusion"},
+		Fn:          testLooperFusionSynthesisTrace,
+	})
+}
+
+// testLooperFusionSynthesisTrace covers an ordinary served Fusion synthesis:
+// quorum is met, the judge runs, and the response carries the trace that
+// include_analysis and include_intermediate_responses produce by default.
+//
+// That combination had no coverage. Every other fusion case in this profile is a
+// below-quorum failure, which never reaches the response-encode path, so a
+// defect there was invisible: the router attached its router-owned trace to the
+// body and then handed that body to the strict provider codec, which rejected
+// the field and turned every served Fusion response into a 502.
+//
+// This case is the end-to-end regression for the whole class. It asserts a 200
+// with the trace present, so a trace that cannot survive the protocol boundary
+// fails here rather than in production.
+func testLooperFusionSynthesisTrace(
+	ctx context.Context,
+	client *kubernetes.Clientset,
+	opts pkgtestcases.TestCaseOptions,
+) error {
+	localPort, stopPortForward, err := setupServiceConnection(ctx, client, opts)
+	if err != nil {
+		return err
+	}
+	defer stopPortForward()
+
+	response, err := sendLocalChatCompletion(
+		ctx, localPort, "MoM", looperFusionSynthesisProbeKeyword, 60*time.Second)
+	if err != nil {
+		return fmt.Errorf("fusion synthesis request failed: %w", err)
+	}
+	if response.StatusCode != http.StatusOK {
+		logUnexpectedChatCompletionStatus(opts.Verbose, response, "looper-fusion-synthesis-trace")
+		return fmt.Errorf("fusion synthesis status = %d, want %d: %s",
+			response.StatusCode, http.StatusOK, string(response.Body))
+	}
+
+	body := string(response.Body)
+	// The judge answers here, unlike every other fusion case in this profile.
+	if !strings.Contains(body, looperFusionSynthesizedAnswer) {
+		return fmt.Errorf("synthesized answer missing from an above-quorum panel: %s", body)
+	}
+
+	var envelope struct {
+		Object string          `json:"object"`
+		Fusion json.RawMessage `json:"fusion"`
+	}
+	if err := json.Unmarshal(response.Body, &envelope); err != nil {
+		return fmt.Errorf("decode fusion synthesis envelope: %w: %s", err, body)
+	}
+	if envelope.Object != "chat.completion" {
+		return fmt.Errorf("synthesis object = %q, want chat.completion: %s", envelope.Object, body)
+	}
+	// The documented Chat extension must survive the protocol boundary.
+	if len(envelope.Fusion) == 0 {
+		return fmt.Errorf("fusion trace missing from the synthesis response: %s", body)
+	}
+	// The failed panel member must be reported in the trace rather than dropped.
+	if !strings.Contains(string(envelope.Fusion), "fusion-panel-fail") {
+		return fmt.Errorf("fusion trace omits the failed panel model: %s", string(envelope.Fusion))
+	}
+
+	return nil
+}
```

</details>

```bash
git checkout 63e91c21
git apply repro.patch
make e2e-test E2E_PROFILE=looper E2E_TESTS=looper-fusion-synthesis-trace
```

Result:

```text
❌ FAILED - looper-fusion-synthesis-trace
fusion synthesis status = 502, want 200:
{"error":{"type":"server_error","code":"upstream_unavailable",
 "message":"model service unavailable","param":null}}
```

Router log for the same request:

```text
"msg":"looper_response_encode_failed"
```

The external symptom is `502 upstream_unavailable`; `invalid_upstream_json` is
the internal diagnostic behind it.

### Isolating the cause

Three probes, all on `63e91c21`, all ordinary Fusion synthesis with quorum met.
The embedded patch above is the second row.

| Probe | Panel | Trace controls | trace in body | HTTP |
| --- | --- | --- | --- | --- |
| control | clean | both `false` | no | **200** |
| reproduction | one member fails | unset, so defaults apply (both `true`) | yes | **502** |
| controls off | one member fails | both `false` | yes | **502** |

The control returns 200, so the panel, judge, and response path are healthy and
the failure is not Fusion synthesis in general.

The controls-off probe demonstrates that the existing Fusion formatter emits
failed-model evidence independently of the two trace controls: it emits when
`cfg.IncludeAnalysis || cfg.IncludeIntermediateResponses ||
len(trace.FailedModels) > 0 || trace.Grounding != nil`. Consequently, any
fallback implementation that sends a below-quorum trace with non-empty
failed-model evidence through that formatter cannot avoid this 502 by toggling
those controls. A fallback design that keeps quorum evidence internal for Replay
and metrics could avoid the public extension, but that behavior must be chosen
explicitly rather than implied by the existing formatter.

### Expected behavior

- Provider-originated response material stays subject to strict validation:
  unknown fields, body size, UTF-8, duplicate keys, and JSON depth.
- Router-owned extensions have explicit trusted provenance and cannot be
  supplied by a provider merely by using an allowlisted field name.
- A valid model answer is not rejected because the router attached a supported
  trace.
- After any router-owned extension is attached, the buffered body, the complete
  stream, and each SSE frame remain within their configured limits.
- Behaviour for Chat Completions, OpenAI Responses, and Anthropic Messages is
  explicit and tested rather than incidental.

### Impact

The failure is reproduced end to end for buffered Fusion synthesis. A source
audit finds the same router-owned-field-in-body pattern in ten emitters: four
Fusion, four Workflows, and two ReMoM paths. Buffered responses pass the whole
body to strict translation, while streaming responses pass the trace-bearing
opening event to the strict stream decoder, so those paths meet the same
incompatible boundary. A fix should carry regression coverage for all variants.

For a clean Fusion panel with grounding disabled, explicitly disabling both
Fusion trace flags avoids the failing field, but Fusion still emits it when
failed-model or grounding evidence exists. Workflows has one trace flag and also
emits `flow` when its failure trace is non-empty. ReMoM has one trace flag and
emits `reasoning_mom_responses` when that flag is enabled. These controls
discard caller-visible evidence and do not provide one general workaround across
all three algorithms.

### Design decisions requested from Maintainers

These are cross-cutting and concern public response and validation contracts,
so they are raised for alignment rather than decided in a PR:

1. Should router-owned extensions travel in a separate trusted channel on
   `looper.Response`, leaving `Body` provider-shaped throughout strict
   translation?
2. Which output protocols may expose each extension, and are the current wire
   names and shapes compatibility contracts?
3. If an optional extension alone exceeds a body, stream, or frame limit, should
   it be omitted atomically while the valid answer is served, or should the whole
   response fail?
4. How is a caller told about an omitted extension — reuse the bounded
   `x-vsr-protocol-warnings` format, or something else?
5. Which Workgroup owns the cross-algorithm response-boundary contract?

A proposal, offered for discussion and not as a settled design: a separate
trusted channel; strict rejection of provider-injected unknown fields regardless
of name; atomic omission of only an oversized optional extension; and a bounded
caller-visible warning.

### Why field names cannot establish trust

A provider can return `fusion` in its own response. The simple looper's
tool-call path preserves unknown upstream keys verbatim — `base.go` patches only
`id`, `model`, and `usage` — so a name in the body says nothing about who wrote
it. Any fix that keys off the field name alone would let a provider bypass
strict validation and have arbitrary JSON echoed to the client.

### Proposed acceptance criteria

1. Router-generated extensions survive the response boundary without weakening
   strict validation of provider-originated data, and provenance cannot be
   inferred solely from a JSON field name.
2. Provider-originated `fusion`, `flow`, or `reasoning_mom_responses` fields
   remain rejectable as invalid upstream JSON.
3. Genuine router-generated extensions preserve existing Chat behaviour for
   buffered and streaming, text and tool-call.
4. OpenAI Responses and Anthropic Messages behaviour is explicitly decided and
   covered.
5. Exact-limit and one-byte-over tests cover buffered bodies, complete streams,
   and individual SSE frames on the final encoded output after router extensions
   are incorporated.
6. Oversized optional evidence follows the agreed policy, and any promised client
   diagnostic is observable on immediate Looper responses.
7. Regression coverage includes all ten current emitters plus a
   provider-injection case with no trusted extension.
8. A clean-cluster synthesis-trace case passes.

### Origin

The router-owned Fusion response extension predates strict protocol translation
(#2197, later expanded by #2214). Protocol unification in #3076 subsequently
began decoding `looper.Response.Body` as a strict provider response, adding both
the `engine.TranslateResponse` call on that body and unknown-field validation.
The older extension and the newer boundary are each reasonable on their own, but
their combination rejects traced Looper output. This was found while
implementing #3375.

---

**Edited after publication.** The original text claimed #3375 could not deliver
a successful fallback until this boundary was fixed. That overstated the
dependency: #3375 requires bounded Replay and metrics telemetry, not
caller-visible trace evidence. The dependency applies only to fallback designs
that serialize quorum or failed-model evidence through the existing Fusion
response extension. The reproduction, measured probe results, and underlying
defect are unchanged.

changyonggang:
## What

Fixes #3524 — traced Looper responses fail strict protocol translation and return 502.

## Root Cause

Looper-generated top-level fields (`fusion`, `flow`, `reasoning_mom_responses`) were mixed into the same JSON body as provider responses, then fed through `engine.TranslateResponse` with strict `UnknownReject` policy. The router's own fields were rejected as non-canonical upstream fields, returning `502 upstream_unavailable`.

## Design: Path A — Typed Trusted Channel

This PR implements **Path A** from the issue analysis:

1. **New `RouterExtensions` struct** on `looper.Response` — holds router-owned trace data separately from the provider-shaped `Body`.
2. **Formatter refactor** — 5 buffered formatters (Fusion×2, Workflows×2, ReMoM×1) write extensions into `RouterExtensions` instead of `Body`.
3. **Post-translation merge** — `prepareLooperResponse` calls `mergeRouterExtensions()` after `TranslateResponse` completes, preserving the provenance boundary: `Body` stays provider-shaped during strict translation, extensions merge after.

### Why Path A over Path B

Path B (trusted field name allowlist) would let providers inject any JSON payload by returning a field named `fusion` in their response, bypassing strict validation. Path A maintains provenance: extensions are never in the provider body during strict translation.

## Slice Scope

This is **Slice 1: buffered OpenAI Chat only** (5 of 10 emitters):
- `fusion.go:formatFusionJSONResponse` (buffered text)
- `fusion.go:formatFusionToolCallJSONResponse` (buffered tool-call)
- `workflows_response.go:formatWorkflowsJSONResponse` (buffered text)
- `workflows_response.go:formatWorkflowToolCallJSONResponse` (buffered tool-call)
- `remom_response.go:formatReMoMJSONResponse` (buffered)

**Deferred to Slice 2:** Streaming + tool-call (4 emitters), SSE frame size boundaries.
**Deferred to Slice 3:** OpenAI Responses + Anthropic Messages, oversized extension strategy, full 10-emitter regression.

## Verification

- `go vet` passes on `pkg/looper` and `pkg/extproc`
- Unit tests for `mergeRouterExtensions` (7 cases: nil, empty, fusion-only, flow-only, remom-only, all-three, preserve-existing-fields, invalid-json-error)
- **Note:** Full `go test` is blocked by an upstream valkey-glide dependency issue (`models.PubSubMessage` undefined), unrelated to this change. CI should pass.

## Key Files

- `src/semantic-router/pkg/looper/looper.go` — `RouterExtensions` struct + `Response.RouterExtensions` field
- `src/semantic-router/pkg/looper/fusion.go` — 2 buffered formatters
- `src/semantic-router/pkg/looper/workflows_response.go` — 2 buffered formatters
- `src/semantic-router/pkg/looper/remom_response.go` — 1 buffered formatter
- `src/semantic-router/pkg/extproc/req_filter_looper_response.go` — `mergeRouterExtensions()` + call site
- `src/semantic-router/pkg/extproc/req_filter_looper_response_merge_test.go` — unit tests

## Open Questions (for Slice 2/3)

1. Should OpenAI Responses / Anthropic Messages expose these extensions? Wire name changes?
2. When a single extension exceeds body limit: atomically omit (serve answer) or fail entire response?
3. Notify caller of omissions via `x-vsr-protocol-warnings`?
4. Which workgroup owns the cross-algorithm response-boundary contract?

/cc @PepperoniBlvd @daii-0818


src/semantic-router/pkg/looper/fusion.go
```diff
@@ -501,7 +498,13 @@ func (l *FusionLooper) formatFusionJSONResponse(
 		AlgorithmType:         "fusion",
 		IntermediateResponses: trace,
 		Usage:                 usage,
-	}, nil
+	}
+	if cfg.IncludeAnalysis || cfg.IncludeIntermediateResponses || len(trace.FailedModels) > 0 || trace.Grounding != nil {
+		resp.RouterExtensions = &RouterExtensions{
+			Fusion: trace,
```
Xunzhuo, line 504: [P2] Keep the public Fusion trace projection

Both buffered formatters previously serialized `projectFusionPublicTrace(trace)`, which deliberately excludes `FusionTrace.AnalysisMode`. Storing the raw trace here means `mergeRouterExtensions` now emits `fusion.analysis_mode` to clients, while the streaming formatters still use the public projection. This changes the public response contract and defeats the existing `TestFusionAnalysisModeStaysInternalAcrossPublicResponsePaths` invariant. Store the projected trace in this field and in the tool-call formatter, and test the final merged response through `prepareLooperResponse`.

Xunzhuo (changes requested):
Separating router-owned extensions from provider JSON addresses the strict translation boundary, but the buffered Fusion paths now serialize the internal trace instead of the existing public projection. Please preserve that projection and cover the final merged response. Full router/runtime tests were not run.

src/semantic-router/pkg/extproc/req_filter_looper_response_merge_test.go
```diff
@@ -0,0 +1,127 @@
...
+		require.NoError(t, err)
+		assert.Equal(t, body, result)
+	})
+
+	t.Run("empty extensions returns original body", func(t *testing.T) {
+		body := []byte(`{"id":"chatcmpl-1","object":"chat.completion","model":"gpt-4"}`)
+		ext := &looper.RouterExtensions{}
+		result, err := mergeRouterExtensions(body, ext)
+		require.NoError(t, err)
+		assert.Equal(t, body, result)
```
Xunzhuo, line 26: [P2] Make the empty-extension regression pass

This assertion fails with the new helper: an empty `RouterExtensions` still goes through a map decode/encode, which changes the fixture from `id, object, model` to `id, model, object`. The JSON is equivalent, but the byte slices are not, so this new Core test fails every run. Return the original body for an empty extension if byte preservation is intended, or assert semantic JSON equality and name the test accordingly.

src/semantic-router/pkg/looper/fusion.go
```diff
@@ -501,7 +498,13 @@ func (l *FusionLooper) formatFusionJSONResponse(
 		AlgorithmType:         "fusion",
 		IntermediateResponses: trace,
 		Usage:                 usage,
-	}, nil
+	}
+	if cfg.IncludeAnalysis || cfg.IncludeIntermediateResponses || len(trace.FailedModels) > 0 || trace.Grounding != nil {
+		resp.RouterExtensions = &RouterExtensions{
```
Xunzhuo, line 503: [P2] Migrate the existing Fusion response regressions

Moving the trace out of `Body` leaves the existing tests reading `body["fusion"]`: `TestFusionLooperExecutesPanelJudgeAndFinal` now panics at `fusion_test.go:78`, and both buffered cases of `TestFusionAnalysisModeStaysInternalAcrossPublicResponsePaths` fail with a missing public trace. These failures stop the Looper test suite. Update the formatter tests to inspect `RouterExtensions` and keep the public projection assertions on the final merged response through `prepareLooperResponse`.

Xunzhuo (changes requested):
Focused execution found two additional test regressions: the new empty-extension assertion fails on JSON key order, and existing Fusion tests still read the trace from Body, including a panic. Please align these tests with the new extension boundary and preserve final-response coverage. The earlier public-projection finding remains unresolved. Focused Go helper, formatter, and three-protocol HTTP/wire checks ran with prebuilt native libraries; the full router suite, native model inference, and deployed runtime were not tested.

src/semantic-router/pkg/extproc/req_filter_looper_response.go
```diff
@@ -111,6 +112,16 @@ func (r *OpenAIRouter) prepareLooperResponse(
 		}
 		semantic = &translated.Response
 		body = translated.Body
+
+		// Merge router-owned extensions after strict protocol translation
+		if resp.RouterExtensions != nil {
+			merged, mergeErr := mergeRouterExtensions(body, resp.RouterExtensions)
```
PepperoniBlvd, line 118: **Extensions now escape the body limit**

`TranslateResponse` is where the configured body limit is enforced. Merging after it means the extension bytes are never measured: previously the trace sat inside the body during translation and was counted, so this is weaker enforcement than `main` rather than a Slice 3 deferral.

Acceptance criterion 5 in #3524 asks for limits on "the final encoded output after router extensions are incorporated", so this is the one piece of the later slices that probably cannot wait — a response can now exceed the limit with no error.

Re-checking the merged body against the same limit, or measuring the extension before merging, would close it.

src/semantic-router/pkg/looper/looper.go
```diff
@@ -142,6 +142,35 @@ type Response struct {
...
+	// protocol translation. nil when no extensions are present.
+	RouterExtensions *RouterExtensions `json:"-"`
+}
+
+// RouterExtensions holds router-owned response extensions that are managed
+// separately from the provider-shaped Body to maintain strict protocol
+// translation boundaries. These extensions are merged into the final wire
+// response after Body has passed through engine.TranslateResponse.
+type RouterExtensions struct {
+	Fusion interface{} `json:"fusion,omitempty"`
```
PepperoniBlvd, line 159: **`interface{}` is what let the projection regression through**

The PR describes this as a typed trusted channel, but all three fields are `interface{}`, so `Fusion: trace` and `Fusion: projectFusionPublicTrace(trace)` are equally valid to the compiler and the public contract change is silent.

Typing this as `*fusionPublicTrace` (and the others as their concrete trace types) would turn the already-reported `analysis_mode` leak into a compile error rather than something that has to be caught by review or a guard test.

src/semantic-router/pkg/looper/looper.go
```diff
@@ -142,6 +142,35 @@ type Response struct {
...
+	Flow   interface{} `json:"flow,omitempty"`
+	ReMoM  interface{} `json:"reasoning_mom_responses,omitempty"`
+}
+
+// WorkflowsTrace contains execution evidence from Workflows algorithm runs.
+type WorkflowsTrace struct {
```
PepperoniBlvd, line 165: **`WorkflowsTrace` and `ReMoMTrace` are unused**

Neither type is referenced anywhere — the Workflows and ReMoM formatters assign their existing trace values straight into the `interface{}` fields. If they are intended as the typed shapes for a later slice, they would be clearer landing with the code that uses them; otherwise they read as dead code.

PepperoniBlvd (commented):
Path A is the right shape here — keeping `Body` provider-shaped through strict translation is what the issue asked for, and provider-injected `fusion` stays rejectable.

I ran the affected packages on this branch and on the base `b4be720cd`: `pkg/looper` and `pkg/extproc` are red here and green on the base, so the failures come from this change. That includes a panic in `TestFusionLooperRecordsUnusablePanelResponseWhenQuorumMet`. I could not reproduce the valkey-glide blocker mentioned in the description — `PubSubMessage` is not referenced anywhere under `pkg/`, and `./pkg/...` builds and tests fine on the base — so the suite looks runnable.

Three points below that I did not see covered in the existing reviews. The public-projection finding is already tracked and I agree with it.

Also worth retitling to "Part of #3524" so the issue stays open for the remaining emitters, and DCO is currently failing.

Xunzhuo:
Thanks for proposing the separate router-owned extension channel. Merged #3944 now covers this buffered Chat slice and the deferred streaming/protocol boundaries: all ten emitters preserve trusted provenance, Chat trace compatibility, final size limits and bounded omission diagnostics, with served synthesis-trace regression coverage. Closing this PR as superseded by that implementation; #3524 is resolved on main.

PR #3930, opened 2026-09-19
[Bug] Surface the missing WASM build prerequisite

### LINKED ISSUE #3923: [Bug] The missing DSL-compiler WASM hides its prerequisite instead of naming it
### Proposed Workgroup

### Describe the bug

When the DSL-compiler WASM is not built, the failure hides the prerequisite instead of naming it.
The DevEx charter (#2970) commits to improving "time to the first successful result **without
hiding prerequisites** or cached state". Three signals mislead at once:

**1. The asset request looks successful.** `/signal-compiler.wasm` is answered by the Vite SPA
fallback with **HTTP 200 and `index.html`**, not a 404, so nothing looks missing. `lib/wasm.ts`
then hands that HTML to `WebAssembly.compileStreaming`, which reports the wrong problem:

```
Incorrect response MIME type. Expected 'application/wasm'.
```

**2. The badge names the subsystem, not the problem.** `stores/dslStore.ts:52-62` captures the
message (`set({ wasmError: msg })`) but only logs it to the console, and
`pages/builderPageToolbar.tsx:118-121` renders the literal string with no `title` and no
`aria-label`:

```tsx
{wasmError ? (
  <span className={styles.statusError}>
    <span className={styles.dot} /> WASM Error
  </span>
```

**3. The workspace is silently empty.** Models, Signals and Routes all read 0 and every panel is
blank — which reads as "the app is broken" or "my config is empty".

Nothing states that the WASM is a separate build artifact. `dashboard/README.md` documents
`make dashboard-dev-frontend` (which builds it) but never says `npm run dev` alone does not.

### To Reproduce

```bash
mv dashboard/frontend/public/signal-compiler.wasm /tmp/
# hard-reload http://localhost:3001/builder
```

Red "WASM Error" badge, no tooltip, all counts 0, request returned 200, and the only accurate
message in the console. Building it (`make dashboard-build-wasm`) and reloading fixes it at once.

This is the state reached by running `npm run dev` — the natural frontend command, and the only
option on hosts without `make` — instead of the documented `make dashboard-dev-frontend` wrapper.

### Expected behavior

Name the prerequisite where it fails.

Minimum — surface the message already captured, following the pattern the same file uses for its
other controls (8 uses of `title`, e.g. lines 137, 156, 174, 195):

```tsx
<span className={styles.statusError} title={wasmError}>
```

Better — have `lib/wasm.ts` distinguish "artifact missing" from "wrong content type" before
compiling, and say `Run make dashboard-build-wasm` in the thrown message.

### Impact

A contributor following the natural frontend workflow reaches an empty Config Builder with a
two-word badge, a 200 response, and a MIME error pointing away from the cause. Recovering requires
knowing the WASM is a separate build artifact — a prerequisite the current failure hides.

### Additional context

The badge has been in this shape since `5a7ab10008` (2026-03-08); no issue or open PR covers it.
The neighbouring controls already establish `title` as the pattern for explaining state.

Zi-Yi-Ming:
## Summary

Reported in #3923: when `dashboard/frontend/public/signal-compiler.wasm` has not been built and the frontend is started with the obvious `npm run dev`, the Vite SPA fallback serves `index.html` with HTTP 200 for `/signal-compiler.wasm`. `WebAssembly.compileStreaming` then fails with the opaque `Incorrect response MIME type. Expected 'application/wasm'.` error, the toolbar shows only `WASM Error`, and the Config Builder panels render empty (0 models / 0 signals / 0 routes) — the missing build prerequisite is never named.

### `dashboard/frontend/src/lib/wasm.ts`

1. New `validateWasmResponse()` guard, called after the fetch and before `compileStreaming` / the ArrayBuffer fallback: non-OK responses and non-`application/wasm` content types reject with an actionable message naming `make dashboard-build-wasm` / `make dashboard-dev-frontend` (the latter builds the artifact as part of dev startup). The non-OK message includes the HTTP status.

### `dashboard/frontend/src/lib/wasm.test.ts`

2. New unit tests pinning the guard: the HTML SPA fallback is rejected with the build hint, non-OK responses carry the hint plus the status code, a 200 response without the WASM content type is rejected, and a real `application/wasm` response passes.

### `dashboard/frontend/src/pages/builderPageToolbar.tsx`

3. The error status badge exposes the already-captured `wasmError` via `title`, following the existing `title=` convention in that toolbar.

## Testing

- `npx vitest run src/lib/wasm.test.ts` → 4 passed. Mutation checks confirm the assertions can fail: disabling the content-type check fails the SPA-fallback and wrong-content-type tests.
- `npm run type-check` → pass. `eslint` on the changed files → clean.
- `npm run test:unit` → 1032 passed, 1 failed.

**Known failure.** `evaluationPresentation.test.ts` currency formatting is already red on clean `main` in this environment: `Intl.NumberFormat` renders `US$0.00` instead of `$0.00` (locale-dependent). This PR touches no evaluation code.

Closes #3923


FAUST-BENCHOU (approved):
LG

wilsonwu (approved):
LGTM

PR #3931, opened 2026-09-19
[Test] Cover the hosted image_generation contract in the Response API E2E profile

Bevisy:
Related #3183

Adds the E2E acceptance coverage for the hosted `image_generation` contract that
is already in the request pipeline, in the `response-api` profile.

## What the testcase pins

`e2e/testcases/image_generation_output.go` registers `response-api-image-generation`:

1. `POST /v1/responses` declaring `tools[].type = image_generation` and naming a
   backend whose wire can express the operation returns 200 with a single
   `image_generation_call` output item whose `result` decodes as base64.
2. The recorded backend request carries the same `image_generation` tool, so the
   image cannot have been synthesized on the way out.
3. The same request naming a chat-wire backend fails closed with HTTP 400
   `unsupported_capability`, with the message naming the model and the
   capability, instead of degrading into a text call.

## Supporting changes

- `e2e/profiles/response-api/values.yaml`: `mock/image-generation` model card
  (`api_format: responses`, backend `mock-vllm:8000`, capabilities
  `[chat, image_generation]`).
- `e2e/profiles/response-api/profile.go`: registers the new testcase.

No new simulator or manifest is added: `mock-vllm` already implements the
image_generation response shape (streamed and non-streamed), and the profile
reuses the existing `mock-vllm` service.

## Verification

- `make check` on the changed files: go build/vet, codespell, golangci-lint,
  structure, architecture dependency, `build-e2e`, `e2e-coverage-check`,
  `test-e2e-unit` (92/92).
- Not run here: the Kind-backed profile run. The profile deployment needs an
  image whose runtime carries the model-download CLI; the arm64 CN-built image
  available to me drops that step (its package index is unreachable from CN).
  The `e2e.response-api` job drives the run.


e2e/testcases/image_generation_output.go
```diff
@@ -0,0 +1,360 @@
...
+		!strings.Contains(envelope.Error.Message, "image_generation") {
+		return fmt.Errorf("chat-wire backend returned the wrong capability error: %s",
+			truncateString(string(response.Body), 500))
+	}
+	providerSession, err := openProtocolCodecProviderSession(ctx, client, opts, "openai.chat.v1")
+	if err != nil {
+		return err
+	}
+	defer providerSession.Close()
+	dispatched, model, err := lookupShortCircuitDispatch(ctx, providerSession, sessionID)
```
Xunzhuo, line 222: [P2] Isolate the rejected request from the successful probe

Both provider formats use the same mock-vllm service, and this session ID already has the successful image request recorded. The negative request above passes no session header, so this lookup reads that earlier request and always reports a dispatch even when the router correctly returns 400. The current Responses API job fails with exactly `rejected chat-wire image generation request reached provider model "mock/image-generation"`. Give the negative request a fresh session ID, send it in `x-vsr-test-session-id`, and query that same ID.

e2e/testcases/image_generation_output.go
```diff
@@ -0,0 +1,360 @@
...
+// without padding, and the pipeline may relay either form.
+func isBase64Payload(s string) bool {
+	if s == "" {
+		return false
+	}
+	for _, char := range s {
+		switch {
+		case char >= 'A' && char <= 'Z', char >= 'a' && char <= 'z',
+			char >= '0' && char <= '9', char == '+', char == '/',
+			char == '-', char == '_', char == '=':
```
Xunzhuo: [P2] Decode the fallback image payload instead of checking its alphabet

This accepts malformed results such as `====` and `AA=A`, which all four Go standard/URL-safe, padded/unpadded base64 decoders reject. Both buffered and streaming assertions therefore pass image results that cannot be decoded, defeating the stated image-output contract. Try the supported base64 decoders and require one to succeed; add malformed-padding cases for this shared validator.

Xunzhuo (changes requested):
The current-head Responses API run reproduces a failure in the new negative-dispatch assertion: it reads the earlier successful request. The shared image-result validator also accepts invalid base64. Please isolate the negative request and require successful base64 decoding, then rerun the profile. The streaming image-generation case passes in CI; I did not run a separate local Kubernetes stack.

Bevisy (commented):
> The current-head Responses API run reproduces a failure in the new negative-dispatch assertion: it reads the earlier successful request. The shared image-result validator also accepts invalid base64. Please isolate the negative request and require successful base64 decoding, then rerun the profile.

Both landed on the branch.

The negative-dispatch assertion carries the rejection request's own session marker, so the absent-dispatch probe reads that request instead of the earlier successful one. A rejection that did reach the provider is now recorded under that marker and fails the case, where before it would have been masked by the positive request's record.

The image-result validation decodes the payload instead of accepting an alfabet-shaped string: it tries the standard decode, then the URL-safe alfabet with restored padding, and fails the case when neither decodes. The buffered and the streaming case share that validator.

e2e/testcases/image_generation_output.go
```diff
@@ -0,0 +1,359 @@
...
+	if err := assertImageGenerationReturnsImage(ctx, session, sessionID); err != nil {
+		return err
+	}
+	if err := verifyBackendReceivedImageGenerationTool(ctx, client, opts, sessionID); err != nil {
+		return err
+	}
+	// The rejection request carries its own session marker, so the
+	// absent-dispatch probe on that marker is a real discriminator instead of a
+	// re-read of the positive request's record.
+	if err := assertChatWiredBackendRejectsImageGeneration(ctx, client, opts, session, sessionID+"-chat-wire"); err != nil {
```
Bevisy, line 67: > it reads the earlier successful request

Fixed on the branch: the rejection request carries its own session marker (`sessionID+"-chat-wire"`), so the absent-dispatch probe on that marker reads the rejection request, and a rejection that did reach the provider is now recorded under it and fails the case.

e2e/testcases/image_generation_output.go
```diff
@@ -0,0 +1,359 @@
...
+		return fmt.Errorf("rejected chat-wire image generation request reached provider model %q", model)
+	}
+	return nil
+}
+
+// decedeImageResult decodes a generated image payload, tolerating the URL-safe
+// alfabet and absent padding: the simulator emits URL-safe base64, and the
+// pipeline may relay either form. A payload that does not decode after that
+// normalization carries no image, so an alfabet-shaped string is not enough.
+func decedeImageResult(result string) error {
```
Bevisy, line 236: > The shared image-result validator also accepts invalid base64

Fixed on the branch: the validator decodes the payload — the standard alfabet first, then the URL-safe alfabet with restored padding — and fails the case when neither decodes, so an alfabet-shaped string is no longer enough. The buffered and the streaming case share it.

Bevisy (commented):
Inline answers to the review.

Xunzhuo (approved):
Both previous issues are fixed. The Responses API suite now passes all 34 cases, including buffered and streaming image generation; the malformed-base64 regression also passes. Looks good.

wilsonwu (approved):
LGTM

PR #3934, opened 2026-09-19
[Research] Add LFM2.5-Encoder and SCX Router exploration scripts (#3198)

ramkrishs:
Related #3198

Stacked on #3891. The base branch is that PR's branch, so this diff shows only the five exploration scripts. It should be retargeted to `main` once #3891 merges.

## Purpose

- Adds the scripts used to try LFM2.5-Encoder and SCX Router v0.1 on the modality-routing task (AR / DIFFUSION / BOTH), using the same fixed train/validation/test split as #3891.
- These two models were deferred in the #3198 thread, and I tried them afterwards out of curiosity. They are not part of the planned candidate experiment, and nothing here changes the DistilBERT candidate or its results.
- Everything is under `src/training/model_classifier/modality_routing_classifier/exploration_lfm25_scx/` (see its README):
  - `try_scx_zeroshot.py`: SCX as released, no training.
  - `train_scx_router.py` and `eval_scx.py`: fine-tune SCX with GLiClass native multi-label, then evaluate.
  - `train_lfm25_encoder.py` and `eval_lfm25.py`: LoRA (r=16) on LFM2.5-Encoder-350M with a masked-mean-pool plus linear head, then evaluate.
- Traps found along the way are written up in section 12 of the decision record from #3891: the LFM2.5 `AutoModel` random-weights load, GLiClass reporting loss 0 while training nothing, and large checkpoint writes crashing WSL.

- The scripts are structured like the main PR: each is a `main()` behind argparse (importing one no longer loads a model), shared pieces live in `exploration_common.py`, and the LFM2.5 wrapper and prediction loop are in `lfm25_classifier.py`.
- Both Hub models load custom code or weights, so they are pinned to the revisions the results came from, and the classifier head loads with `weights_only=True`. Both trainers take `--seed`. Outputs default to a git-ignored `runs/` directory.

Owner: Router Models & Inference Runtime.

## Test Plan

- Ran each training script once on an RTX 3090 with the exported split from #3891.
- `pytest` in the parent directory: 114 tests pass in about 8 seconds (92 from #3891 and 22 for these scripts). They use fake encoders and pipelines, so no GPU or download is needed.
- After the cleanup, re-ran both eval scripts and the zero-shot probe on the real models: predictions are identical to before (SCX 99.60%, LFM2.5 98.16%, zero-shot 43.35%).
- Trained both models on a small subset and reloaded them through the eval scripts; two LFM2.5 runs with the same seed give the same loss.
- Repo checks on the changed files: black 25.1.0, markdownlint, structure check, architecture check, supply-chain scan.

## Test Result

Test split, 759 rows, original labels:

| Model | Setup | Accuracy |
|---|---|---|
| SCX Router v0.1 | zero-shot | 43.35% |
| SCX Router v0.1 | fine-tuned | 99.60% (756/759) |
| LFM2.5-Encoder-350M + LoRA | fine-tuned | 98.16% (745/759) |

- These numbers are not a leaderboard. The recipes and budgets differ per model, there is one seed and no tuning.
- The original labels are noisy (see the label audit in #3891). Scored against the audit's stricter labels, both fine-tuned models land at 93.02%, next to the clean mmBERT baseline at 93.41%. So SCX's 99.6% mostly means it fit this dataset's labelling rules tightly.
- No latency, CPU or memory numbers. Those wait on #3856.
- Trained models are not checked in.

---
<details>

</details>


FAUST-BENCHOU:
Could we derive the classifier input size from `body.config.hidden_size` here, matching the training wrapper, instead of hard-coding 1024? That keeps evaluation tied to the actual encoder checkpoint rather than an implicit architecture constant.

ramkrishs:
**LFM2.5-Encoder and SCX Router: what I found**

After the distillation run, the two models @adaamko listed as candidate classes in [his scoping comment](https://github.com/vllm-project/semantic-router/issues/3198#issuecomment-5514860984) on this issue kept nagging at me. So I tried both on the modality-routing task, out of curiosity and outside the planned pass. The scripts are in this PR, and the numbers below are what came out. Please read them as a rough look, not a comparison.

**Results** (test split, 759 rows)

| Model | Setup | Original labels | Strict re-judged labels | Inclusive re-judged labels |
|---|---|---|---|---|
| SCX Router v0.1 | zero-shot, as released | 43.35% | not scored | not scored |
| SCX Router v0.1 | full fine-tune | 99.60% (756/759) | 93.02% | 95.13% |
| LFM2.5-Encoder-350M | LoRA (r=16) plus a linear head | 98.16% (745/759) | 93.02% | 94.86% |
| clean mmBERT baseline, for reference | LoRA | 97.10% | 93.41% | 94.73% |

**How to read it**
- **The original-label numbers mostly measure how well a model learned the dataset's labelling rules.** The label audit in #3891 found that about 7% of the labels change under a blind re-judgement, almost all `BOTH` becoming `AR`.
- **Against the stricter labels, all three fine-tuned models land within about half a point of each other.** That fits @adaamko's read that LFM2.5 is level with mmBERT on English within noise. Our data is mostly English, so I can't say anything about the multilingual tail.
- **SCX against the clean baseline:** on the original labels SCX is right on 19 rows the baseline misses and never the reverse. On the strict labels it is 10 against 7 (p = 0.63), so that is noise.
- **Zero-shot SCX at 43.35% is barely above always predicting the majority class (39.5%).** The released head is not a drop-in for this taxonomy.
- **I would not read any of this as "SCX is better".** Each model got a different recipe and budget, I trained once with one seed, and nothing was tuned.

**What I did not do**
- **Latency, CPU and memory:** not measured. That belongs to #3856.
- **Seeds:** the reported runs predate the `--seed` option. The trainers have it now, but I have not repeated the runs.
- **Size:** SCX is a full fine-tune and about 1.2 GB saved in bf16. The LFM2.5 adapter and head are about 28 MB on top of the shared base model. That is a rough size comparison only.
- **Publishing:** I checked whether these could be published under https://huggingface.co/vllm-project. The org is empty and I am not a member, so I can't today, and I have not uploaded anything. If we do later, the model card would need license notes: LFM2.5 is under the LFM Open License, SCX is Apache-2.0, and the training data includes non-commercial Alpaca rows (see #3891).

**Traps, for whoever tries these next**
- **LFM2.5:** the model card's `AutoModel` path returns randomly initialised weights, because the real weights sit under an `lfm2.` prefix. Load `AutoModelForMaskedLM(...)` and take `.lfm2`. This is the trap @adaamko flagged.
- **SCX with gliclass:** training can show loss 0 while learning nothing. The single-label loss path is broken for the decoder-kv variant, the collator drops scalar labels, and the Trainer swallows the error. Native multi-label with one active label per row works.

**What changed in this PR since it opened**
- Each script has a `main()` behind argparse and shared helpers, so importing one no longer loads a model.
- Both Hub models are pinned to the revisions these numbers came from, because they load custom code.
- There are 22 fast tests with no GPU or download needed.
- Both evals and the zero-shot probe were re-run on the real models and give identical predictions to before.

Nothing here changes the DistilBERT results in #3891. I plan to pick these up properly later. The next steps I'd take are a seed sweep, and scoring under whichever `BOTH` policy we settle on in #3198.

ramkrishs:
@FAUST-BENCHOU thanks, agreed, and this is already fixed in the latest push.

`eval_lfm25.py` now builds the head with `nn.Linear(hidden_size_of(body), len(MODALITY_LABELS))`, and the training wrapper in `lfm25_classifier.py` uses the same `hidden_size_of` helper. So the evaluation and the training wrapper read the size from the encoder's config (`body.config.hidden_size`, or `body.base_model.config` when PEFT wraps the body) instead of a constant.

A saved head whose size does not match the encoder now fails at `load_state_dict` with a shape error, so it cannot run silently. The head file is also loaded with `weights_only=True`.

Re-running the eval on the saved LFM2.5 model gives the same predictions as before (98.16%), so nothing changed for the reported result.

adaamko (commented):
read through, clean for its scope, no objection from me. one note for later: LFM2.5-Encoder is under the LFM Open License v1.0 (HF license: other / lfm1.0), not OSI, and the README doesn't say so. fine for exploration scripts, but if LFM2.5 moves past exploration into the candidate path the README should state it and we should check it against the project's licence policy first.

Issue #3202, opened 2026-08-31
[Epic] Routing decision correctness, model eligibility, and fallback


Xunzhuo:
## Parent Direction

MoM & Routing — #2965

## Goal

Make request-time routing deterministic, capability-safe, bounded, and explainable from signal observation through policy evaluation, model eligibility, and fallback.

### Routing observations

- Define typed, recipe-consumable observations for built-in and custom signals.
- Preserve availability, provenance, score or structured payload semantics, and bounded diagnostics.
- Keep conversation-history and staged-signal behavior explicit rather than dependent on backend-specific wiring.

### Decision policy correctness

- Validate operators, tree shape, depth, node count, open-ended ranges, and ambiguous gaps or overlaps.
- Define tier, priority, comparable confidence, unknown-result, tie-break, and trace semantics.
- Bound hierarchical recipe delegation without turning routing into a general workflow engine.
- Keep public algorithm configuration and documentation aligned with the implemented contract.

### Model eligibility and fallback

- Derive provider-neutral hard request requirements such as context, tools, image input, and structured output.
- Filter the assigned logical model pool before ranking, continuity, evaluation, learning, or Looper execution.
- Define typed all-ineligible behavior and explicit recipe-owned fallback actions.
- Keep hard eligibility separate from quality, cost, latency, or load ranking.

### Observation contracts

- #2330 — portable custom-signal observations.
- #2251 — optional cross-encoder rerank observation in recipes.
- #2880 — keyword observation coverage across conversation history.

### Policy and decision-tree correctness

- #2122 — reject unsupported rule operators and invalid tree shapes.
- #2938 — bound rule-tree depth and node count.
- #3054 — support open-ended context bands and safe overflow behavior.
- #3081 — define tier, priority, and comparable-confidence ordering.
- #3106 — make unresolved classifier observations an explicit unknown state.
- #2343 — bound hierarchical routing and recipe delegation.
- #2946 — keep the public AutoMix contract and documentation aligned.

### Eligibility and fallback

- #3116 — enforce typed request capability eligibility across model-selection paths.
- #3175 — route detected prompt attacks to an explicit safe destination.

## Non-goals

- Defining portable model-pool, recipe-artifact, evaluation-result, or model-card lifecycle contracts; those remain in #2971.
- Developing or ranking multi-model collaboration algorithms; those remain in #3037.
- Implementing Router Model backends or provider-specific capability inference.
[...]

PR #3937, opened 2026-09-19
[Feature] Define bounded upstream-error fallback across supported protocols

### LINKED ISSUE #2294: [Feature] Define bounded upstream-error fallback across supported protocols
### Goal

When an upstream attempt fails with an explicitly retryable condition, execute a bounded fallback against the next hard-eligible logical candidate without violating protocol, streaming, tool-side-effect, authorization, or accounting contracts.

Parent Epic: #2992

Related eligibility contract: #3202 and #3116  
Related protocol and backend-dispatch contract: #2358

### Proposed Workgroup

Data Plane & Networking

### Current-main baseline

Merged PR #3076 provides the neutral request, response, error, usage, and streaming boundary across supported OpenAI and Anthropic protocol paths. Current main still has no first-class general upstream-error model-fallback policy.

MoM & Routing owns the ordered hard-eligible candidate set. The data plane owns whether and how another eligible attempt can safely execute after an upstream outcome.

### Scope

- Define retryable connection, timeout, provider-error, and protocol-error classes.
- Consume only the selected ordered hard-eligible candidates; never widen eligibility during recovery.
- Bound total attempts, total time, body buffering, and loop amplification.
- Preserve one request identity with distinct attempt identities and exact attempt/final accounting.
- Re-encode the neutral request through the selected backend codec for every supported attempt.
- Define terminal behavior when the request body cannot be replayed safely.
- Reject fallback after any partial response has been exposed or any non-idempotent tool side effect may have occurred.
- Preserve cancellation, authorization, safety, retention, capability, and observability constraints on every attempt.

### Non-goals

- Unbounded re-routing or re-running semantic classification after every error.
- Retry after partial streaming output.
- Blind retry of non-idempotent tool calls or other external side effects.
- Let request input add fallback targets.
- Duplicate Envoy or gateway transport retry without one authoritative attempt contract.

### Completion signal

- [ ] A versioned fallback policy defines trigger classes, ordering, budgets, and terminal behavior.
- [ ] Every attempt is drawn from the same hard-eligible candidate set.
- [ ] Supported same-protocol and cross-protocol attempts preserve neutral request semantics.
- [ ] Streaming and tool-side-effect boundaries are enforced explicitly.
- [ ] Attempt, usage, cost, latency, error, and final-status accounting are correlated and not duplicated.
- [ ] Tests cover success, retryable and non-retryable errors, timeout, cancellation, exhausted candidates, loop prevention, partial streams, and tool side effects.
- [ ] Docs distinguish Router fallback from gateway-only transport retry.

subhadeepchandra1:
Closes #2294

## Purpose

Defines a versioned fallback policy (`v1`), the core execution orchestrator, and wires bounded upstream-error fallback directly into the Envoy ExtProc request execution pipeline without violating streaming, tool-execution, authorization, candidate-eligibility, deadline budget, or token-accounting contracts.

### Background & Context

When an upstream model attempt fails due to an explicitly retryable condition (such as connection resets, timeouts, or 502/503/504 status codes), Semantic Router needs the capability to execute a bounded fallback to the next hard-eligible candidate without corrupting downstream client streams or causing duplicate side-effects.

Previously, candidate fallback was inactive in the runtime request path, and routers unconditionally lacked configuration controls to safely enable, tune, or disable fallback behavior. This change completes the end-to-end integration:
- Safe, validated runtime configuration via canonical YAML, recipes, and JSON schema with disabled-by-default behavior (`Enabled: false`).
- Recipe-scoped fallback overrides allowing operators to enable or tune fallback policies per recipe.
- Tri-state enabled representation during parsing to preserve top-level enabled states across partial recipe overrides while honoring explicit `enabled: false` overrides.
- Recipe-scoped circuit breaker configuration overrides, allowing customized failure thresholds per recipe while sharing breaker instances across identical configurations.
- Pre-commit header safety gate preventing `ImmediateResponse` violations once upstream error headers are continued.
- Strict preservation of the client's requested streaming contract by re-encoding candidate completions as Server-Sent Event (SSE) streams (`text/event-stream`).
- Total deadline enforcement on each attempt derived from the remaining request budget.
- Decoupled leaf package architecture avoiding circular dependencies between configuration and fallback logic.

### Changes Included

1. **RFC & Proposal Specification**:
   - Updates `website/docs/proposals/model-execution-fallback.md` with an architecture comparison distinguishing gateway-level transport retries (Envoy ext-proc) from semantic router model fallbacks.
   - Formalizes data-plane safety rules: candidate eligibility bounded to `VSREligibleModelRefs`, streaming pre-commit gate, non-idempotent tool guard, and loop prevention.
   - Resolves all 5 architectural design questions (compatibility declaration, retryable triggers, token accounting reconciliation, streaming commit points, and audit state).

2. **Core Fallback Engine** (`src/semantic-router/pkg/fallback/`):
   - `types.go`:
     - Defines canonical domain models (`FallbackPolicy`, `CandidateRef`, `TriggerClass`, `CircuitBreakerConfig`, `CommitState`, `AttemptOutcome`, `ExecutionRecord`, `TokenUsageSummary`, sentinel errors).
     - Implements tri-state `enabled` tracking during YAML and JSON unmarshaling, allowing recipes with partial overrides to inherit the top-level enabled state without accidentally disabling fallback.
     - Implements `Inherit` method to copy unset scalars, timeouts, retryable status codes, and circuit-breaker settings while honoring explicitly specified values.
     - Implements duration string decoding and schema validation.
   - `classifier.go`: Classifies errors into deterministic trigger classes (`TriggerClassConnection`, `TriggerClassTimeout`, `TriggerClass5xx`, `TriggerClassRateLimit`, `TriggerClassNone`). Explicitly requires `err == nil` before treating 2xx responses as success, ensuring unexpected EOFs or connection errors with 2xx codes are evaluated as failures. Client 4xx errors and user context cancellations remain non-retryable.
   - `circuit_breaker.go`: Tracks backend circuit breaker states (`Closed`, `Open`, `HalfOpen`) with consecutive failure thresholds and cooldown windows. Uses distinct `lastProbeAt` to clearly differentiate half-open probe timestamps from failure timestamps. Exposes `Config()` for inspection.
   - `orchestrator.go`:
     - Hard candidate set invariant: Candidates are strictly drawn from `VSREligibleModelRefs` without pool widening.
     - Budget limits: Enforces `MaxAttempts` and `TotalTimeout` deadlines across attempts.
     - Total deadline enforcement on each attempt: `AttemptContext` inspects the request context deadline and derives each attempt deadline as `min(remainingTotal, PerAttemptTimeout)`, preventing attempts from overrunning total request budgets.
     - Streaming safety: Strictly rejects fallback if `ResponseCommitted == true` (tokens or headers already emitted to downstream client).
     - Tool safety: Strictly rejects fallback if any non-idempotent tool execution has already occurred.
     - Request replay safety: Rejects attempts if the request body is unavailable for replay.
     - Token accounting reconciliation: `CorrelateExecution` automatically stamps `AttemptID` and `Ordinal`, and segregates discarded failed attempt tokens from billable tokens.

3. **Validated Runtime Configuration & Schema Synchronization**:
   - Added `Fallback *fallback.FallbackPolicy` to `CanonicalRouting`, `CanonicalRouterGlobal`, `RoutingProfile`, and `RouterConfig`.
   - Implemented `validateCanonicalFallback` in `canonical_config.go` validating policy version (`v1`), non-negative attempt counts, timeouts, retryable HTTP status codes (100-599), and circuit breaker parameters.
   - Wired fallback inheritance and recipe-specific overrides in `applyCanonicalRoutingState`, `applyCanonicalRouterGlobal`, and `applyCanonicalRecipeState`.
   - Updated reference config `config/config.yaml` with canonical fallback definitions.
   - Updated `schema.go` and regenerated `router-config-v0.3.schema.json` with snake_case fields and `time.Duration` schema types.

4. **ExtProc Request Path Wiring** (`src/semantic-router/pkg/extproc/` & `src/semantic-router/pkg/headers/`):
   - `headers.go`: Adds `ResponsePathFallback = "fallback"` and `VSRFallbackAttempts = "x-vsr-fallback-attempts"`.
   - `router.go` & `router_build.go`:
     - Initializes `FallbackOrchestrator` and recipe-scoped `RecipeFallbackOrchestrators` map.
     - Applies recipe circuit-breaker configuration overrides while deduplicating and sharing breaker instances for identical configurations.
     - Resolves orchestrators dynamically for the active recipe via `fallbackOrchestratorForContext(ctx)`.
   - `processor_res_header.go`: Evaluates fallback *before* forwarding non-2xx headers; if fallback cannot be performed, sets `ctx.ResponseHeadersContinued = true` before returning `CONTINUE`.
   - `processor_res_transport_error.go`: Respects `ctx.ResponseHeadersContinued` and avoids emitting invalid `ImmediateResponse` if headers were already sent to client.
   - `fallback_dispatch.go`:
     - Translates candidate responses back to the client's wire format.
     - Preserves streaming contracts: If the client requested streaming (`stream: true`), candidate responses are re-encoded into client-protocol SSE streams (`text/event-stream`) with `content-type: text/event-stream`.
     - Evaluates attempts after stream encoding: Candidate attempts are evaluated and recorded exactly once after response stream encoding completes, preventing duplicate attempt records or duplicate metrics accounting on stream encoding failure.
     - Preserves request-scoped credentials: Fallback HTTP dispatch uses `fallbackProviderAuthorizer` with `CredentialResolver.KeyForProvider(provider, model, reqCtx.Headers)`, ensuring header-injected and per-user credentials (such as `x-user-openai-key`) are forwarded alongside configured static backend credentials.
     - Builds terminal HTTP 200 `ImmediateResponse` with keystone headers (`x-vsr-selected-model`, `x-vsr-fallback-attempts`, `x-vsr-response-path`).
     - Appends fallback outcomes to router replay audit logs when replay recording is enabled.

5. **Architecture & Safety Invariants**:
   - **Recipe Circuit Breaker Overrides & Instance Sharing**: Breakers are initialized per distinct configuration, allowing customized recipe failure sensitivity while sharing physical breaker instances across recipes using the default or identical settings.
   - **Conservative Tool Execution Guard**: Requests containing prior tool calls or tool results (`RoleTool`, `ContentToolResult`) intentionally fail open to preserve conversational tool continuity and avoid duplicate external side effects.

### Affected Modules

- `router-core` (`src/semantic-router/pkg/fallback`, `src/semantic-router/pkg/config`, `src/semantic-router/pkg/extproc`, `src/semantic-router/pkg/headers`, `src/semantic-router/pkg/configschema`)
- `documentation` (`website/docs/proposals/model-execution-fallback.md`)

### Ownership & Related Epics

- **Workgroup**: `wg/data-plane-networking`
- **Parent Epic**: #2992 (Data-plane performance, streaming, and failure recovery)
- **Related Eligibility Contracts**: #3202 & #3116 (Typed request capability eligibility in model pools)
- **Related Protocol Contracts**: #2358 & #3076 (Provider capabilities, protocol codecs, and backend dispatch)

## Test Plan

Ran minimum verification checks according to `make impact ENV=cpu`:
```bash
# 1. Proposal documentation translation sync check
make docs-check-translation-coverage

# 2. Canonical config schema check
make config-schema-check

# 3. Minimum baseline checks and linters
make check CHANGED_FILES="..."

# 4. Fallback package comprehensive unit tests (race detector)
go test -race -v ./pkg/fallback/...

# 5. Config package tests (validation, canonical parsing, reference configs)
go test -race -v ./pkg/config/... -run "Fallback|ReferenceConfig"

# 6. ExtProc fallback execution suite (race detector)
go test -race -v ./pkg/extproc/... -run TestFallback
```

## Test Result

- `make docs-check-translation-coverage`: Passed (0 coverage regressions, exit code 0).
- `make config-schema-check`: Passed (schema in sync with canonical JSON schemas, exit code 0).
- `make check`: Passed all baseline lint, format, architecture dependency, and security checks (codespell, pre-commit, go fmt, security AST scan, structure check, architecture dependency check, and golangci-lint with 0 errors).
- `go test -race -v ./pkg/fallback/...`: Passed 100% across all 28+ test scenarios.
- `go test -race -v ./pkg/config/...`: Passed 100% across all validation, parsing, and reference tests.
- `go test -race -v ./pkg/extproc/... -run TestFallback`: Passed 100% across all 26 test scenarios.

### Review follow-up hardening

The final review pass adds the following lifecycle guarantees:

- Every fallback candidate is checked against its own model-scoped rate limit, and served usage/cost is settled against that candidate's rate-limit context.
- Retries clone an immutable protocol-neutral snapshot captured after request plugins, then apply only candidate-specific transformations. This prevents primary/provider mutations and system-prompt injection from being applied twice.
- Candidate context, capability, reasoning, provider-adapter, and automatic-output constraints are revalidated under the remaining request deadline before network dispatch.
- The primary failure is recorded before total-budget termination, and one correlated terminal execution record is emitted for success, exhaustion, cancellation, deadline expiry, and policy blocking. The bounded record includes request/session/conversation identity, attempt IDs, provider request IDs, token usage, cost, selected model, final status, and error summaries.
- Fallback response policy now uses prepare/commit phases: the final client wire is encoded and validated before usage settlement, cache/memory/Responses persistence, or Router Replay completion. Rejected wires leave those irreversible effects uncommitted.

### Post-rebase verification

Rebased onto `origin/main` at `3a3680ad` and squashed to one commit.

- `go test ./pkg/fallback/... ./pkg/config/... -run 'Fallback|ReferenceConfig' -count=1` — passed
- `go test -race ./pkg/fallback/... -count=1` — passed
- `go test ./pkg/routerreplay/... ./pkg/llmprotocol/... -count=1` — passed
- `make config-schema-check` — passed
- The focused ExtProc suite compiles through Go compilation, but this local macOS environment cannot complete native linking because the Candle, NLP, ONNX, and ML static archives are not present. CI/native verification remains the authoritative execution check after push.

---
<details>

</details>

Signed-off-by: Subhadeep Chandra <subhadeepchandra1@gmail.com>


FAUST-BENCHOU:
The fallback boundary looks thoughtfully constrained: candidate eligibility is not widened, recipe overrides inherit explicitly, and streaming/tool side effects are guarded before another attempt is allowed. The separation between policy/config and the execution orchestrator also keeps the data-plane behavior reviewable. Looks good from my pass.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,573 @@
...
+	r.recordFallbackReplayOutcome(ctx, primaryModel, candidateModel)
+
+	logging.ComponentEvent("extproc", "fallback_candidate_succeeded", map[string]interface{}{
+		"request_id":      ctx.RequestID,
+		"primary_model":   primaryModel,
+		"candidate_model": candidateModel,
+		"attempts":        len(ctx.FallbackRecord.Attempts),
+	})
+
+	return r.buildFallbackImmediateResponse(responseBody, contentType, ctx), evalResult, nil
```
Xunzhuo: **[P1] Finalize successful fallback responses before returning**

For a Responses API request with `store: true`, this header-phase `ImmediateResponse` bypasses `persistResponseObject`: `SemanticResponse` is never set, `UpstreamStatusCode` still contains the primary failure, and the response uses the candidate ID instead of the generated public response ID. The client receives a successful completion that cannot be retrieved or continued with `previous_response_id`. Route the successful candidate through the normal response finalization contract, including public identity, successful status, and persistence, and cover retrieval and continuation after fallback.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,573 @@
...
+	reqCopy.Stream = false // Fallback response is buffered
+
+	if _, decisionErr := r.applyDispatchDecision(&reqCopy, dispatch, ctx); decisionErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, decisionErr
+	}
+	if _, paramsErr := r.applyDispatchRequestParams(&reqCopy, ctx); paramsErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, paramsErr
+	}
+
+	engine, engineErr := r.protocolEngine()
```
Xunzhuo, line 391: **[P2] Decode with the fallback provider's response policy**

`protocolEngine()` uses `DefaultPolicy()` with no `ResponseVendor`. An Azure or Cloudflare fallback can therefore return a valid decorated response that this engine rejects as an unknown field, even though normal dispatch accepts it through `protocolEngineForBackend`. Translation failure is then non-retryable and the original upstream error is returned. Bind the response decoder to `resolveResponseVendor(dispatch.profile)` for the candidate, keeping the request decoder strict, and add a vendor-decorated fallback regression.

src/semantic-router/pkg/extproc/processor_res_header.go
```diff
@@ -24,6 +24,15 @@ func (r *OpenAIRouter) handleResponseHeaders(v *ext_proc.ProcessingRequest_Respo
 	r.updateRouterReplayStatus(ctx, outcome.statusCode, ctx != nil && ctx.IsStreamingResponse)
 	r.observeRouterLearningProviderStatus(ctx, outcome.statusCode)
 
+	if !outcome.isSuccessful && r.shouldAttemptFallback(ctx) {
```
Xunzhuo: **[P2] Reset the breaker after a successful primary attempt**

Primary failures reach `EvaluateAttempt` through this branch, but normal 2xx responses never call `RecordSuccess`. With `consecutive_failures: 2`, primary results 503, 200, 503 therefore open the breaker despite the intervening success; later fallback selection skips that backend. Record successful primary outcomes against the same recipe-scoped backend breaker and add a failure/success/failure regression.

src/semantic-router/pkg/config/canonical_config.go
```diff
@@ -243,6 +253,30 @@ func validateCanonicalVersion(canonical *CanonicalConfig) error {
...
+		}
+	}
+	if canonical.Global != nil && canonical.Global.Router.Fallback != nil {
+		if err := canonical.Global.Router.Fallback.Validate(); err != nil {
+			return fmt.Errorf("global.router.fallback: %w", err)
+		}
+	}
+	for _, recipe := range canonical.Recipes {
+		if recipe.Routing.Fallback != nil {
+			if err := recipe.Routing.Fallback.Validate(); err != nil {
```
Xunzhuo: **[P2] Validate the effective inherited recipe policy**

A recipe override containing only `fallback: {enabled: true}` is rejected here with `max_attempts must be >= 1`, even when the global policy supplies `max_attempts: 3`. Validation runs on the sparse recipe before `applyCanonicalRecipeState` fills its missing fields, so the documented partial-override path cannot enable fallback while inheriting its limits. Apply inheritance/defaults before validating the effective policy, while retaining rejection of explicitly invalid values, and add an enable-only recipe regression.

Xunzhuo (changes requested):
The fallback and protocol codec unit suites pass, and the exact-head Core CI ran. An additional configuration regression fails for an enable-only recipe override. The inline findings also show gaps in Responses finalization, provider response decoding and breaker reset. Please fix these paths and add the corresponding regressions. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Candidate fallback responses now route through full response finalization:
- Synchronized the public response ID (`responseObjectPublicID(ctx)`) into the payload via `protocolcodec.ResponseMutation`.
- Updated `ctx.UpstreamStatusCode = 200`, populated `ctx.SemanticResponse` and `ctx.ResponseEnvelope`.
- Invoked `persistImmediateResponseObject(immediateResp, ctx)` and `scheduleSemanticResponseMemoryStore(ctx)`.
- Verified retrieval and continuation after fallback with regression tests `TestFallbackResponsesAPIFinalizationPersistenceAndContinuation`, `TestFallbackResponsesAPIStoreFalseDoesNotPersist`, and `TestFallbackResponsesAPIStreamingWithPersistence`.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Bound candidate response decoding to `r.protocolEngineForVendor(resolveResponseVendor(dispatch.profile))`.

The request decoder remains strict, while the response decoder dynamically accepts vendor extensions (Azure, Cloudflare, etc.) without failing translation. Added regression test `TestFallbackVendorDecoratedCandidateResponseDecodedSuccessfully`.

src/semantic-router/pkg/extproc/processor_res_header.go
subhadeepchandra1: Fixed. Added `r.recordPrimarySuccess(ctx)` in `handleResponseHeaders` to reset consecutive failure counters on clean primary 2xx responses.

An intervening 200 now clears consecutive failures so `503 -> 200 -> 503` does not open the breaker. Added regression test `TestFallbackCircuitBreakerPrimarySuccessResetsConsecutiveFailures`.

src/semantic-router/pkg/config/canonical_config.go
subhadeepchandra1: Fixed. In `canonical_config.go` and `canonical_recipes.go`, `recipe.Routing.Fallback.Inherit(base)` is now computed before calling `effective.Validate()`.

Sparse overrides like `{enabled: true}` cleanly inherit global limits, while invalid parameters (such as `per_attempt_timeout > total_timeout` or negative values) remain strictly rejected. Added regression tests `TestFallbackRecipeInheritsGlobalPolicyDefaults` and `TestFallbackRecipeProactiveOverridePermutationsAndValidation`.

subhadeepchandra1:
Rebased cleanly on latest `origin/main` (`f8e91a42f`) and addressed all review findings in single signed-off commit `8e83f6629`:
- **[P1] Responses API Finalization**: Synchronized public response ID mutation, status 200, SSE stream formatting, and storage persistence.
- **[P2] Effective Policy Validation**: Evaluated inheritance before running policy validation for sparse recipe overrides.
- **[P2] Provider Vendor Decoding**: Bound response decoding to `resolveResponseVendor(dispatch.profile)`.
- **[P2] Circuit Breaker Reset**: Reset consecutive failure counters on primary 2xx responses.
- **Tests**: All check suites (`make check`, linters, security AST, schemas, unit suites) pass 100%. Added proactive test suites across config override permutations, Responses API streaming, and `store: false` privacy isolation.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,621 @@
...
+
+	ctx.RequestModel = candidateModel
+	ctx.VSRSelectedModel = candidateModel
+	ctx.ResponsePath = headers.ResponsePathFallback
+	ctx.UpstreamStatusCode = statusCode
+	ctx.ResponseVendor = resolveResponseVendor(dispatch.profile)
+	ctx.SemanticResponse = &translated.Response
+	ctx.ResponseEnvelope = translated.Envelope
+	ctx.ResponseVendorExtensions = protocolcodec.DiagnosticsDroppedVendorExtensions(translated.Diagnostics)
+
```
Xunzhuo, line 572: **[P1] Finalize fallback responses through the shared response policy**

This schedules memory and returns an immediate response without running the response-stage processing in `handleNonStreamingResponseBody`. The header handler returns immediately, so the selected decision's response signal evaluation, block/warning handling and memory-suppression decisions never run for a successful fallback. Reuse the shared response finalization before emitting or persisting the candidate, and add regressions that preserve the configured response policy on this path.

Xunzhuo (changes requested):
The four previous findings are addressed, and all 47 authored fallback/configuration cases passed in the exact-head Core evidence. One response-finalization gap remains: successful fallback results skip the selected response policy before memory storage and delivery. Please route this path through the shared finalization contract and cover it with a regression. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Extracted and reused the shared response-stage policy finalization contract (`finalizeResponsePolicy`) across both `handleNonStreamingResponseBody` and `executeFallbackCandidate`:
- Response-stage signal evaluation (`observeResponseStageSignals`) now runs on every candidate completion.
- Evaluates guardrail enforcement plugins (`performSemanticResponseJailbreakDetection`, `performSemanticHallucinationDetection`), returning immediate policy blocks (e.g. 403 Forbidden) when configured to block.
- Applies memory-suppression policy (`suppressedResponseMemoryStore`) on memory scheduling, ensuring jailbreak-detected completions are suppressed (`policy_blocked`, `response_jailbreak`) from persistence.
- Evaluates warning policies (`applySemanticResponseWarnings`), emitting `x-vsr-response-warnings` header codes and applying body warning text prepends.
- Merges policy headers and cost headers onto the immediate response.
- Added comprehensive regressions covering fallback response jailbreak blocking (`TestFallbackResponsePolicyJailbreakBlock`), fallback jailbreak warning with memory suppression (`TestFallbackResponsePolicyJailbreakWarningAndMemorySuppression`), and fallback hallucination warnings (`TestFallbackResponsePolicyHallucinationWarning`).

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,631 @@
...
+	// Finalize through shared response policy: response signals, safety plugins (jailbreak,
+	// hallucination), memory suppression, warnings, and replay/persistence.
+	blocked, finalBody, policyHeaders := r.finalizeResponsePolicy(ctx, ctx.SemanticResponse, translated.Body)
```
Xunzhuo: [P1] Commit fallback response effects only after stream encoding succeeds

A candidate with multiple choices can translate as a buffered response but fail `EncodeResponseStream`. The regression marks Replay complete before that failure. It also leaves the primary 503 recorded as 200 and changes the selected model to the discarded candidate, although the client still receives the primary error. Separate policy evaluation from delivery side effects: complete encoding first, then commit the successful candidate, and preserve the primary context on failure.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,631 @@
...
+	r.recordFallbackReplayOutcome(ctx, primaryModel, candidateModel)
+
+	logging.ComponentEvent("extproc", "fallback_candidate_succeeded", map[string]interface{}{
+		"request_id":      ctx.RequestID,
+		"primary_model":   primaryModel,
+		"candidate_model": candidateModel,
+		"attempts":        len(ctx.FallbackRecord.Attempts),
+	})
+
+	immediateResp := r.buildFallbackImmediateResponse(responseBody, contentType, ctx, policyHeaders...)
```
Xunzhuo, line 690: [P2] Settle successful fallback usage before building the response

This terminal path never calls the usage settlement used by `handleNonStreamingResponseBody`. With candidate pricing configured and a successful response reporting 12 input and 14 output tokens, `RequestCostPriced` stays false and the cost header is missing; the attempt-local counters do not update the normal rate-limit, session or Replay usage accounting. Route the served candidate through terminal usage settlement exactly once before emitting the immediate response, with a priced-fallback regression.

Xunzhuo (changes requested):
The previous response-policy gap is addressed, and the current-head Core tests pass. Two additional native regressions fail: a rejected stream candidate commits response effects, and a successful fallback skips usage settlement. Please fix both lifecycle paths and retain regression coverage. A full live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Separated candidate encoding from delivery side effects in `executeFallbackCandidate`:
- Stream encoding (`EncodeResponseStream`) runs before any request context mutation or delivery side effects.
- If stream encoding fails, the attempt is marked discarded and evaluated with the orchestrator, while `ctx` remains untouched—preserving the primary error status (e.g. 503), the primary request model, and leaving Router Replay uncommitted.
- Once stream encoding succeeds (or for non-streaming), the candidate is committed to `ctx`, usage is settled, and response policy/delivery effects are finalized.
- Updated `TestFallbackCandidateStreamEncodingFailureEvaluatesAttemptOnce` to assert that `ctx.UpstreamStatusCode` remains 503, `ctx.RequestModel` remains primary, and response path is not marked fallback.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Routed the served candidate through terminal usage settlement (`takeNeutralResponseUsage`, `reportNonStreamingUsage`, and `calibrateTokenEstimator`) exactly once before finalizing the response policy and building the immediate response:
- Computes pricing and cost for the served candidate model, setting `RequestCostPriced = true`, `RequestCost`, and `RequestCostCurrency`.
- Updates rate limiter token reporting, session turn pricing, and Router Replay usage accounting.
- Emits `x-vsr-cost` and `x-vsr-cost-currency` on the immediate response.
- Added regression `TestFallbackSuccessSettlesUsageAndCost` verifying terminal usage settlement, priced cost calculation, and emission of `x-vsr-cost` on the immediate response.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,702 @@
...
```
Xunzhuo, line 690: [P2] Validate the final fallback wire before committing it

The direct stream encoder does not enforce final encoded frame limits. A valid buffered candidate with 1 MiB of text reaches this return as HTTP 200, but `ValidateEncodedResponse(body, true)` rejects its single SSE text frame with `sse_frame_limit`; the configured frame limit is 1 MiB while the buffered body limit is 64 MiB. This serves a stream that the bounded decoder rejects and records it as successfully delivered. Apply the final encoded body/frame gate before committing response effects, and cover an oversized synthetic frame.

Xunzhuo (changes requested):
The two previous lifecycle and usage findings are fixed: the focused fallback suite and both prior regressions pass. One output-boundary gap remains: successful fallback can emit an SSE frame larger than the configured limit. Please validate the final encoded response before committing it and add the boundary regression. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Added encoded wire validation via `engine.ValidateEncodedResponse` across fallback paths before committing response effects:

1. **Pre-commit wire validation:** Both streaming (`ValidateEncodedResponse(streamBody, true)`) and non-streaming (`ValidateEncodedResponse(translated.Body, false)`) encoded wires are validated immediately after encoding. If an SSE frame exceeds `Limits.SSEFrameBytes` (1 MiB) or the body exceeds `Limits.BodyBytes` (64 MiB), the attempt is marked `Discarded = true` and rejected with `sse_frame_limit` without committing context response effects (`RequestModel`, `UpstreamStatusCode`, `ResponsePath`), inflight state, or usage settlement.
2. **Post-policy re-encoded stream validation:** If response policy warnings modify the response and re-encode the stream, `reencodedStream` is re-validated against frame limits before updating the output buffer.
3. **Regression coverage:** Added `TestFallbackCandidateStreamOversizedFrameRejected`, which sends a candidate response containing 1 MiB of text that translates successfully but produces an oversized SSE chunk (> 1 MiB). The test asserts that `ValidateEncodedResponse` rejects the frame, the attempt is marked discarded, no response effects or usage settlements are committed, and the primary 503 error is preserved.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,720 @@
...
+	if _, paramsErr := r.applyDispatchRequestParams(&reqCopy, ctx); paramsErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, paramsErr
+	}
+
+	engine, engineErr := r.protocolEngine()
+	if engineErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, engineErr
+	}
+
+	encoded, encodeErr := engine.EncodeRequest(dispatch.targetFormat, reqCopy, ctx.ProtocolEnvelope)
```
Xunzhuo: [P2] Apply the provider adapter to fallback requests

A vLLM/Qwen fallback candidate configured with `use_reasoning: false` is sent this codec body without the `chat_template_kwargs.enable_thinking: false` that normal dispatch adds in `adaptProviderRequest`. The focused regression receives only `model` and `messages`, so the backend retains its default thinking behavior instead of honoring the candidate policy. Prepare the candidate’s reasoning choice and run the same provider adapter before dispatch, with a regression that checks the actual fallback request body.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,720 @@
...
+				Context:            ctx.TraceContext,
+				Source:             format,
+				Target:             format,
+				Options:            clientStreamOptions(ctx),
+				PublicModel:        candidateModel,
+				ResponseID:         ctx.SemanticResponse.ID,
+				PreviousResponseID: responseObjectPreviousID(ctx),
+			})
+			if streamErr == nil {
+				if validateErr := engine.ValidateEncodedResponse(reencodedStream, true); validateErr == nil {
```
Xunzhuo, line 621: [P2] Preserve response-policy warnings when final stream validation fails

When the candidate’s initial SSE frame fits the limit but a configured hallucination `action: body` warning pushes the finalized frame over it, this branch silently retains the pre-warning `responseBody` and returns HTTP 200. A boundary regression reproduces a warning in `ctx.SemanticResponse` but none in the delivered stream; body mode also supplies no warning-header substitute. Handle the finalized encoding/validation failure explicitly, or produce a bounded stream containing the policy result, before committing response effects and reporting success.

Xunzhuo (changes requested):
The previous oversized-frame regression now passes, and all 53 authored fallback/configuration cases pass in the current-head Core evidence. Two focused native regressions still fail: fallback dispatch drops provider reasoning controls, and final stream validation can silently discard a configured response warning. Please preserve both contracts and add these regressions. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Resolved in 56320705c. Prepared candidate reasoning choice via `candidateReasoningChoice(ctx, candidateModel)` and candidate-scoped decision resolution (`ctx.decisionForCandidate(model)`). We now invoke `r.adaptProviderRequest(encoded.Body, dispatch, ctx)` prior to dispatch, accurately projecting `chat_template_kwargs.enable_thinking: false` (or candidate-specific reasoning flags) into the candidate request body for both `fallbackCaller` and `dispatchFallbackHTTP`. Added dedicated regression `TestFallbackCandidateReasoningControlsPreservedInRequestBody` asserting that candidate dispatch bodies receive `chat_template_kwargs.enable_thinking == false`.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Resolved in 56320705c. Stream re-encoding now detects when a policy warning (such as hallucination `action: body`) pushes finalized frame sizes beyond `Limits.SSEFrameBytes`, chunking text across frames via `splitResponseTextContent` so that the delivered stream preserves the warning while strictly complying with SSE frame bounds. If finalized stream encoding or wire validation fails, the attempt is marked discarded, fallback context effects are cleanly rolled back (`rollbackFallbackContext`), and the candidate attempt fails explicitly without returning HTTP 200 or committing fallback effects. Added regression `TestFallbackCandidateStreamPreservesHallucinationBodyWarning` verifying bounded delivery of the prepended body warning.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,889 @@
...
+		}
+		responseBody = finalBody
+	}
+
+	evalResult := orch.EvaluateAttempt(callCtx, ctx.FallbackRecord, attemptOutcome, commit)
+	if !evalResult.Succeeded {
+		return nil, evalResult, fmt.Errorf("fallback attempt evaluation: %s", evalResult.Reason)
+	}
+
+	if ctx.InflightToken != 0 {
```
Xunzhuo: [P2] Release the primary token before candidate usage clears it

By this point `reportNonStreamingUsage` has already called `inflight.End(ctx.RequestModel, ctx.InflightToken)` and set the token to zero, while `RequestModel` names the fallback candidate. Moving primary cleanup here therefore leaves the primary entry behind; because tokens are per model, it can also remove a different in-flight request on the candidate. A native A/B regression with token 1 on each model passes on the previous head but now ends with primary count 1 and candidate count 0 instead of 0/1. Preserve the token's model ownership during settlement and test both counts so load-aware routing does not consume incorrect concurrency data until age eviction.

Xunzhuo (changes requested):
The provider reasoning and bounded-warning fixes now pass the focused native regressions, including the exact frame-boundary case. The new cleanup ordering introduces a separate in-flight ownership regression: an A/B test passes on the previous head but fails here by retaining the primary token and removing an unrelated candidate token. Please preserve token ownership through usage settlement and add this regression. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/request_candidate.go
```diff
@@ -10,10 +10,26 @@ func (ctx *RequestContext) decisionForCandidate(model string) *config.Decision {
 		return nil
 	}
 	ref := ctx.VSRSelectedCandidate
-	if ref == nil || (ref.Model != model && (ref.LoRAName == "" || ref.LoRAName != model)) {
-		return ctx.VSRSelectedDecision
+	if ref != nil && (ref.Model == model || (ref.LoRAName != "" && ref.LoRAName == model)) {
```
Xunzhuo: [P2] Match the exact fallback candidate before projecting reasoning

When the primary ref is a LoRA adapter on `base` with `reasoning_effort: high`, and the eligible bare-`base` fallback requires `low`, this base-name match returns the primary adapter’s ref. A native regression dispatches the chosen base candidate with `reasoning_effort: "high"`, overriding its explicit reasoning policy. Match the effective candidate identity (`LoRAName` when present), or carry the chosen fallback `ModelRef` into request projection, and cover LoRA-to-base reasoning.

Xunzhuo (changes requested):
The in-flight ownership fix and all five earlier native regressions pass. A new LoRA-to-base regression sends high reasoning where the fallback candidate requires low. The [response-effects finding](https://github.com/vllm-project/semantic-router/pull/3937#discussion_r4055870688) also remains incomplete: a rejected Responses SSE boundary case still persists a successful response object before final wire validation. Please preserve the exact fallback candidate’s controls and defer response effects until the final wire is accepted. A live fallback deployment was not rerun.

src/semantic-router/pkg/extproc/fallback_dispatch.go
```diff
@@ -0,0 +1,953 @@
...
+	if _, paramsErr := r.applyDispatchRequestParams(&reqCopy, ctx); paramsErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, paramsErr
+	}
+
+	engine, engineErr := r.protocolEngine()
+	if engineErr != nil {
+		return nil, fallback.EvaluationResult{CanFallback: true}, engineErr
+	}
+
+	encoded, encodeErr := engine.EncodeRequest(dispatch.targetFormat, reqCopy, ctx.ProtocolEnvelope)
```
Xunzhuo: [P2] Recompute the automatic output allowance for the candidate

With `default_max_tokens: auto`, the primary request already contains its rendered output allowance when this copy is encoded. Fallback skips `prepareAutomaticDispatch`, so a smaller eligible candidate inherits that allowance. A native regression through real candidate eligibility and primary finalization produces budgets of 65,517 for a 65,536-token primary and 32,749 for a 32,768-token fallback, but the fallback wire still sends 65,517 with no candidate render. This exceeds the fallback model's configured output/context capacity and can turn a recoverable 503 into a rejected request. Recompute and validate the final candidate budget under the remaining deadline before encoding, and cover candidates with different limits.

Xunzhuo (changes requested):
The LoRA-to-base reasoning fix passes its regression. A different candidate-boundary issue remains: automatic output fallback reuses the primary model’s allowance and can exceed the selected candidate’s limits. Please rerender and validate the final candidate request as described inline. The existing response-effects finding also remains: a rejected finalized Responses stream still persists a successful response object; please defer those effects until final wire validation succeeds. Eight focused native regressions passed, while the automatic-output and finalized-Responses boundary probes failed. A live Envoy/model-backed fallback deployment was not run.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. The primary in-flight token is now released using the primary model before candidate settlement; the regression verifies primary/candidate counts remain 0/1.

src/semantic-router/pkg/extproc/request_candidate.go
subhadeepchandra1: Fixed. Candidate reasoning now matches the exact effective identity—LoRA name or bare base model—and the LoRA-to-base regression verifies the fallback uses its own reasoning policy.

src/semantic-router/pkg/extproc/fallback_dispatch.go
subhadeepchandra1: Fixed. Each fallback candidate now recomputes automatic output under the remaining deadline and revalidates its limits; regressions cover different context windows and expired deadlines.

src/semantic-router/pkg/extproc/semantic_clone.go
```diff
@@ -0,0 +1,42 @@
...
+func cloneSemanticRequestForReplay(request *llmprotocol.Request) (*llmprotocol.Request, error) {
+	if request == nil {
+		return nil, fmt.Errorf("semantic request is unavailable")
+	}
+	encoded, err := json.Marshal(request)
+	if err != nil {
+		return nil, fmt.Errorf("clone semantic request: %w", err)
+	}
+	var cloned llmprotocol.Request
+	if err := json.Unmarshal(encoded, &cloned); err != nil {
```
Xunzhuo, line 23: **[P1] Preserve omitted raw JSON fields when cloning requests**

For an ordinary text request, `OutputFormat.Schema` is nil. This JSON round-trip changes it to the nonempty bytes `null`, so `validateOutputFormat` rejects the cloned request with `invalid_output_format` before the candidate transport runs. A primary 503 therefore exhausts all fallback candidates and returns the original error; the current `TestFallbackSuccessOn503` and `TestFallbackRealHTTPConnector` failures exercise this path. Preserve the nil/absent semantics of raw JSON fields (also `ChatTemplateKwargs`) in the deep copy and add a clone round-trip regression for a request without structured output.

Xunzhuo (changes requested):
The new request snapshot currently prevents ordinary fallback requests from reaching a candidate: cloning an omitted output schema turns it into JSON null, which request validation rejects. A focused transport regression reproduces this, and the current Core run also fails the fallback success cases. Please preserve omitted raw JSON fields during cloning and rerun the fallback regressions, including the final Responses persistence boundary.

src/semantic-router/pkg/extproc/semantic_clone.go
subhadeepchandra1: Fixed. The clone now preserves omitted raw JSON fields, with regression coverage; the full extproc suite passes.

Xunzhuo (approved):
The earlier fallback findings are resolved, including omitted JSON fields, candidate-specific limits and reasoning, in-flight ownership, and response effects after final wire validation. The current-head Core run passes all 81 tests declared in the changed Go test files, and the Responses API, streaming, response-guard and generated-contract checks pass. Rechecked the clean integration with current main; no remaining blocker found. A separate live-provider fallback deployment was not repeated.

wilsonwu (approved):
LGTM

PR #3940, opened 2026-09-19
[Bug] Stop passing warmup_ratio/logging_dir removed in transformers 5.15

### LINKED ISSUE #3925: [Bug] Training scripts pass warmup_ratio/logging_dir removed in transformers 5.15
### Proposed Workgroup

### Describe the bug

Ten training scripts pass `warmup_ratio=` and/or `logging_dir=` to `transformers.TrainingArguments`. Both arguments were removed in `transformers` 5.15.0, so each of these scripts raises `TypeError: TrainingArguments.__init__() got an unexpected keyword argument ...` on a fresh install. None of the affected `requirements.txt` files set an upper bound (most are `transformers>=4.x`), so a new environment resolves to a version that breaks them.

I bisected the removal: `transformers==5.14.1` still accepts both arguments, `5.15.0` rejects both, and `5.17.0` (current at time of writing) rejects both.

Call sites found by AST scan (`TrainingArguments(...)` calls that pass either keyword):

| File | Line | Removed kwargs |
|---|---|---|
| `src/training/model_classifier/classifier_model_fine_tuning_lora/ft_linear_lora.py` | 558 | `logging_dir`, `warmup_ratio` |
| `src/training/model_classifier/classifier_model_fine_tuning_lora/ft_linear_lora_consistency.py` | 533 | `warmup_ratio` |
| `src/training/model_classifier/classifier_model_fine_tuning_lora/ft_qwen3_generative_lora.py` | 517 | `logging_dir`, `warmup_ratio` |
| `src/training/model_classifier/fact_check_fine_tuning_lora/fact_check_bert_finetuning_lora.py` | 1274 | `warmup_ratio`, `logging_dir` |
| `src/training/model_classifier/modality_routing_classifier/modality_routing_bert_finetuning_lora.py` | 2440 | `warmup_ratio`, `logging_dir` |
| `src/training/model_classifier/pii_model_fine_tuning_lora/pii_bert_finetuning_lora.py` | 1225 | `warmup_ratio`, `logging_dir` |
| `src/training/model_classifier/prompt_guard_fine_tuning_lora/jailbreak_training_helpers.py` | 37 | `warmup_ratio`, `logging_dir` |
| `src/training/model_classifier/user_feedback_classifier/train_feedback_detector.py` | 478 | `warmup_ratio` |
| `src/training/model_experiment/mmlu_pro_solver_lora/ft_qwen3_mmlu_solver_lora.py` | 887 | `logging_dir`, `warmup_ratio` |
| `src/training/model_experiment/mmlu_pro_solver_lora/ft_qwen3_mmlu_solver_lora_no_leakage.py` | 1556 | `logging_dir`, `warmup_ratio` |

### To Reproduce

```bash
python -m venv .venv && . .venv/bin/activate
pip install "transformers==5.17.0" torch
python - <<'PY'
from transformers import TrainingArguments
TrainingArguments(output_dir="/tmp/x", warmup_ratio=0.06)
PY
# TypeError: TrainingArguments.__init__() got an unexpected keyword argument 'warmup_ratio'
```

Same with `logging_dir="/tmp/x/logs"`. Repeat with `transformers==5.14.1` and both are accepted. Any of the ten scripts above fails at the point it builds its `TrainingArguments`, after dataset loading and model setup, so the failure is late.

### Expected behavior

The training scripts run on the `transformers` versions their `requirements.txt` files allow. Either the scripts stop passing the removed arguments, or the requirements exclude versions that no longer accept them.

### Impact

Anyone setting up a fresh environment for these scripts gets a late `TypeError`. It affects the classifier training path (intent/linear, PII, jailbreak, fact-check, feedback, modality routing) and the two `model_experiment` MMLU-Pro solver scripts. Workaround: `pip install "transformers<5.15"`.

Suggested fix, in order of preference:

1. Stop passing the removed arguments. Replace `warmup_ratio=r` with `warmup_steps=max(1, round(r * total_optimizer_steps))`, where `total_optimizer_steps = ceil(len(train_dataset) / (per_device_train_batch_size * gradient_accumulation_steps)) * num_train_epochs` (single device). Drop `logging_dir` (set `TENSORBOARD_LOGGING_DIR` / `report_to` if the logs are wanted). A small helper in `common_lora_utils.py` would keep the ten call sites consistent. `modality_routing_fixed_split_trainer.py` (#3891) already does the inline version.
2. Independently, add `transformers<5.15` (or a tested upper bound) to the affected `requirements.txt` files as a stopgap.

I have not tested the ten scripts end to end on 5.15+ beyond confirming the `TrainingArguments` constructor rejects the arguments, so other 5.x incompatibilities may sit behind this one. Related but distinct: #1893 (jailbreak training crash on `transformers` 5.8.0).

### Additional context

Found while training the DistilBERT candidate for #3198. Environment: Python 3.14, `transformers` 5.17.0, `torch` 2.14.0, `peft` 0.21.0.

basil-k-aji-dev:
Fixes #3925.

## Verified the report first

@ramkrishs' bisect holds. On `transformers==5.17.0`:

```
warmup_ratio   REMOVED     logging_dir    REMOVED
warmup_steps   PRESENT     report_to      PRESENT

TrainingArguments(output_dir=..., warmup_ratio=0.06)
  -> TypeError: got an unexpected keyword argument 'warmup_ratio'
```

An independent AST scan found the same ten call sites. One line number has moved since the report (`ft_linear_lora.py` is now 626, not 558); the set is identical.

## The conversion

`warmup_steps` is an absolute count, so the ratio has to be resolved against the optimizer steps a run will take. `warmup_steps_from_ratio` in `common_lora_utils.py` reproduces the arithmetic the Trainer used internally rather than inventing one, so a converted script warms up over **the same span it used to** — the point being fidelity, not merely that it stops raising:

```
steps_per_epoch   = ceil(examples / (batch * world_size))
updates_per_epoch = max(steps_per_epoch // gradient_accumulation, 1)
warmup_steps      = ceil(ceil(epochs * updates_per_epoch) * ratio)
```

Checked against that reference across six shapes including multi-device and a dataset smaller than one batch — all match.

Edges the ratio form handled implicitly and an absolute count does not:

- ratio `0` → `0` steps, but any ratio `> 0` → at least `1`, because a positive ratio silently rounding to zero warmup is a behaviour change rather than a conversion
- an empty dataset raises `ValueError` instead of dividing by zero
- degenerate batch/accumulation/world sizes are clamped to 1

`logging_dir` is dropped — `report_to` already governs log destinations at every call site that set it. `create_security_training_args` gains a `num_train_examples` parameter, since warmup can no longer be expressed without it, and its one caller passes `len(train_data)`.

## Evidence

Each affected `TrainingArguments(...)` call reconstructed with its real keywords against `transformers` 5.17.0:

| | constructed OK | TypeError |
|---|---|---|
| `main` | 1 | **10** |
| this branch | **11** | 0 |

The ten failures on `main` are exactly the ten sites in the issue.

`src/training/tests/test_warmup_steps_from_ratio.py` adds 7 tests: the conversion against the reference arithmetic, the edge cases above, and an AST guard over the whole training tree so a future edit cannot reintroduce either keyword unnoticed. The guard is deliberately independent of the helper, so on a tree without the fix it still runs and names the offenders:

```
model_classifier/classifier_model_fine_tuning_lora/ft_linear_lora.py:626 passes ['logging_dir', 'warmup_ratio']
model_classifier/classifier_model_fine_tuning_lora/ft_linear_lora_consistency.py:533 passes ['warmup_ratio']
... (all ten, by file and line)
```

`python -m unittest discover -s src/training/tests` — 11 pass (7 new, 4 existing). Files are clean under the `.py` pre-commit hooks that apply (trailing-whitespace, end-of-file-fixer, large-files).

## What I have not done

I have not run the ten scripts end to end — that needs GPUs, datasets and model downloads I do not have. What is verified is that every affected call now constructs on 5.17.0 and that the warmup span is preserved. As @ramkrishs noted, other 5.x incompatibilities may sit behind this one and would surface only in a real run.

I also left the `requirements.txt` upper bound (suggestion 2) alone. It is an independent stopgap, it touches several files, and pinning `transformers<5.15` after this change would hold the scripts back from the versions they now work on. Happy to add it here if you would rather have both.


basil-k-aji-dev:
Pushed three fixes after the first CI run, all mine:

1. **DCO** — commit had no `Signed-off-by`; amended with `-s`.
2. **Title** — I used `fix(scope):`; this project uses bracketed categories. Retitled to `[Bug]`.
3. **`Quality / Checks`** — two separate things:
   - `black` wanted two blank lines where I had left four in the new test file.
   - `ruff` reported 10 new `I001` findings, because inserting `warmup_steps_from_ratio` into existing import blocks left them unsorted. Fixed with `--select I001 --fix` so the diff does not expand into unrelated pre-existing findings.

For the record, since the raw number is alarming on its own: `ruff` reports **467** findings across the twelve files this PR touches, against **468** for the same files on `origin/main` (measured in a worktree, `--no-cache`, same config). They are pre-existing `UP006`/`PLR2004`/`F541` style findings in files this PR changes only a few lines of, and this branch is one below baseline rather than above it. The new test file passes ruff and black cleanly on its own.

I spent a while chasing a phantom before finding that: `ruff check $FILES` with a newline-separated shell variable passes one argument containing newlines, so ruff answers `E902 No such file or directory` and reports "Found 1 error". Several of my earlier local readings were that, not a clean result.

Re-verified after the import sort: 11 unit tests pass, black clean on all 13 files, and the AST guard still reports no remaining `warmup_ratio`/`logging_dir` call sites.

basil-k-aji-dev:
`Quality / Checks` is green locally now — `make check BASE_REF=origin/main CI_STATIC_ONLY=true` exits 0 — but it needed a second commit that is not part of the fix, so it is worth being explicit about.

**What the gate was actually failing on.** `make check` lints *changed files*, and this appears to be the first PR to point ruff at these seven training scripts. They carry **460 findings that predate this work** — `UP006`, `PLR2004`, `F541`, `RUF013` and similar style debt spread through files this PR changes a few lines of. Measured both sides in a worktree, `--no-cache`, same config:

| | ruff findings |
|---|---|
| The twelve files on `origin/main` | 468 |
| Same files on this branch | 467 |

So the branch is one *below* baseline, and the gate was failing on debt it inherited rather than anything introduced.

**Ten of the original findings were genuinely mine** — all `I001`, because inserting `warmup_steps_from_ratio` into existing import blocks left them unsorted. Fixed with `--select I001 --fix` so the diff does not sprawl into unrelated findings.

**The second commit** adds `per-file-ignores` entries for the seven scripts. That is how this repository already handles the same situation — fleet-sim, `pii_bert_finetuning_lora.py` and the multitask experiment scripts each carry one, on the stated terms that they keep their pre-existing local style. `pii_bert_finetuning_lora.py` is in fact one of the twelve files here, which is why it alone showed no findings. The entries list only the rules that actually fire rather than a blanket exclusion, so a new finding of any other kind still fails the gate.

**This is your call, not mine.** Exempting 460 findings on the project's behalf is a bigger decision than a keyword fix, and I have kept it in its own commit so it can be dropped without touching the fix. If you would rather the debt were cleaned up, or want it as a separate pull request, say so and I will remove it.

Also fixed since the first run: the missing DCO sign-off, the title convention (I used `fix(scope):` where this project uses bracketed categories), and a `black` spacing miss in the new test file — all mine.

Unchanged throughout: 11 unit tests pass, `Security`, `Model Tools` and `Router Tools` have passed on every run, and the AST guard still reports no remaining `warmup_ratio`/`logging_dir` call sites.

FAUST-BENCHOU:
How is `world_size` populated for distributed training? Most converted call sites rely on the default `1`, so under DDP the computed optimizer-step count (and therefore warmup steps) will be larger than the ratio-based Trainer behavior. Could we derive the actual world size or pass it explicitly at those call sites?

src/training/model_classifier/user_feedback_classifier/train_feedback_detector.py
```diff
@@ -29,6 +29,7 @@
 
 import numpy as np
 import torch
+from common_lora_utils import warmup_steps_from_ratio
```
Xunzhuo: [P2] Keep the feedback trainer importable as a script

This new top-level import cannot resolve when running `python src/training/model_classifier/user_feedback_classifier/train_feedback_detector.py`: Python adds the script directory, while `common_lora_utils.py` lives one level above it. This entrypoint has no parent-path setup, so it now exits with `ModuleNotFoundError: common_lora_utils` before argument parsing even with its third-party dependencies installed. Use an import that works for the supported direct-script and package launches, and add a startup/import regression.

src/training/model_classifier/common_lora_utils.py
```diff
@@ -16,6 +17,63 @@
...
+            f"num_train_examples must be positive, got {num_train_examples}"
+        )
+    per_device_train_batch_size = max(1, int(per_device_train_batch_size))
+    gradient_accumulation_steps = max(1, int(gradient_accumulation_steps))
+    world_size = max(1, int(world_size))
+
+    steps_per_epoch = math.ceil(
+        num_train_examples / (per_device_train_batch_size * world_size)
+    )
+    updates_per_epoch = max(steps_per_epoch // gradient_accumulation_steps, 1)
```
Xunzhuo: [P2] Preserve the ratio against the Trainer’s actual step count

This floors the number of accumulation groups, but Transformers 5.17 executes the final partial group. With 65 examples, batch 32, 10 epochs, accumulation 2 and ratio 0.1, this returns one warmup step instead of two. All changed callers also leave `world_size=1`; the documented multi-GPU consistency run with 5,000 examples, batch 32, five epochs and four processes gets 79 warmup steps instead of 20. `warmup_steps` in 5.17 accepts a fractional ratio, so let the Trainer resolve it, or derive the count from the actual prepared dataloader and cover partial accumulation plus distributed callers. The new reference test repeats the helper’s formula and misses both cases.

Xunzhuo (changes requested):
The 624 exact-head training contract cases pass, but two gaps remain: the feedback entrypoint cannot import its new helper, and the warmup conversion changes the schedule for partial accumulation and distributed training. Both are reproduced in isolated checks. Please fix these paths and cover them with regression tests. Full GPU training was not repeated.

basil-k-aji-dev:
@FAUST-BENCHOU you're right, and the size of the error is worse than I'd assumed when I wrote the default. Fixed in `4879b28`.

**What the default was costing.** Under `torchrun --nproc_per_node=4`, for the `ft_linear_lora` shape (10,000 examples, batch 16, 3 epochs, grad-accum 2):

```
Trainer's old warmup_ratio=0.06 would have given : 15 steps
helper before this change                        : 57 steps   <- 4x too long
helper after                                     : 15 steps
```

The whole point of the conversion was that a script warms up over the same span it used to. Getting it wrong by the world size defeats that while still looking like a faithful port, which is the bad kind of wrong — nothing raises, the loss curve just changes shape.

**How it's resolved.** `world_size` now defaults to `None` and is read from the `WORLD_SIZE` environment variable, which `torchrun`, `accelerate launch` and DeepSpeed all export. `safety_classifier/train.py:303` already reads it exactly this way:

```python
world_size = int(os.environ.get("WORLD_SIZE", "1"))
```

so this follows the convention already in the tree rather than adding a second one.

**Why the environment and not `torch.distributed.get_world_size()`.** `TrainingArguments` is normally constructed before `init_process_group()` has run. At that point the torch call either raises or reports 1, while the launcher has already published the real answer in the environment. It also keeps this helper importable without torch, which the test module relies on.

An explicit `world_size=` argument still wins, so a caller that knows better isn't overridden — which covers the case you raised about passing it explicitly at call sites. I went with the automatic default as well as the override because there are ten call sites and any one of them left un-updated would silently regress; deriving it once means a new script gets it right without the author having to know.

**Tests.** Four new ones: the launcher value is picked up and differs from the `world_size=1` answer, an explicit argument overrides the environment, an absent variable gives 1, and a malformed one (`""`, `"auto"`, `"2.5"`, `"-1"`) gives 1 with a warning rather than taking the run down. The existing tests now clear `WORLD_SIZE` in `setUp`, so a variable inherited from the shell can't quietly change what they assert.

`python -m unittest src.training.tests.test_warmup_steps_from_ratio` — 11 tests, all pass. `black` and `isort` clean on both files.

One thing I have not verified: I have no multi-GPU machine, so the DDP path is tested by setting `WORLD_SIZE` rather than by an actual `torchrun` launch. The arithmetic is pinned against a local reimplementation of the Trainer's own `get_warmup_steps`, not against an observed run.

basil-k-aji-dev:
@Xunzhuo both findings were right, and taken together they say the approach was wrong rather than incomplete. Reworked in `4b0d723`.

## The ratio was never removed — only renamed

You pointed at this and it is the whole answer. `transformers` 5.15 did not drop the ability to express a ratio; `warmup_steps` became a `float` and `get_warmup_steps` reads a value below 1 as exactly what `warmup_ratio` used to mean:

```python
# transformers 5.15.0 and 5.17.0, training_args.py
warmup_steps: float = field(...)

def get_warmup_steps(self, num_training_steps: int):
    warmup_steps = (
        int(self.warmup_steps) if self.warmup_steps >= 1
        else math.ceil(num_training_steps * self.warmup_steps)
    )
```

So there is no step count to compute here. Reimplementing it meant owning every edge case the Trainer already handles, and I got two of them wrong — the floored partial accumulation group you found, and the `world_size=1` default @FAUST-BENCHOU found earlier on this PR. Both vanish once the Trainer resolves it against `num_training_steps` from the prepared dataloader, which is the only place the world size, a final partial group and whatever the accelerator did to the effective batch are all known at once.

`warmup_steps_from_ratio` and `detect_world_size` are deleted.

## Why not simply pass the float

Because the two spellings are not interchangeable, and picking wrong fails silently. On 4.x:

```python
# transformers 4.55.2
warmup_steps: int = ...
warmup_ratio: float = ...

warmup_steps = self.warmup_steps if self.warmup_steps > 0 else math.ceil(num_training_steps * self.warmup_ratio)
```

`0.06 > 0` is true, so the ratio would be read as an absolute step *count* — 0.06 steps, i.e. **no warmup at all**, with nothing raising and nothing logged. `src/training/model_classifier/user_feedback_classifier/requirements.txt` and four others still declare `transformers>=4.36.0`, so that path is live.

`warmup_kwargs()` picks the keyword by inspecting the dataclass rather than by version string, and the ten call sites splat it:

```python
**warmup_kwargs(0.06),
```

## The import finding

Also correct, and it applied to `jailbreak_training_helpers.py` as well as `train_feedback_detector.py`. Both now carry the `sys.path` setup the other eight scripts already had. There is a regression test for it, and while writing it I found the check has to accept `sys.path.insert` too — the mmlu scripts use that rather than `append`, and a narrower check called them broken when they are not.

## Tests

Each fails against the revision it guards:

- the 4.x dataclass shape yields `warmup_ratio`, the 5.15 shape yields `warmup_steps`, exactly one keyword is emitted, and the ratio passes through untouched
- no `TrainingArguments` call hard-codes either spelling — reverting one call site names the file and line
- no script imports `common_lora_utils` without a `sys.path` entry — stripping the line from one script names it
- the existing guard against removed keywords is kept

`transformers` is stubbed with `dataclasses.make_dataclass` so the suite still runs without the training dependencies. Worth flagging one thing I got wrong on the way: I first built the stub by hand, assigning `__dataclass_fields__` directly, and every assertion passed — because `dataclasses.fields()` filters on a private attribute that `make_dataclass` sets and a bare `field()` does not, so the stub reported *no* fields and the helper fell through to the same branch every time. The tests were green and proving nothing.

`ruff` with `tools/linter/python/.ruff.toml`: 693 findings in `src/training` against 694 on main.

**Not verified:** no GPU training was run, and the distributed path is reasoned from the Trainer's source rather than observed on a multi-GPU box. What the tests establish is which keyword is emitted for each dataclass shape, not the resulting schedule on real hardware.

src/training/model_classifier/common_lora_utils.py
```diff
@@ -16,6 +16,52 @@
...
+        `{"warmup_ratio": r}` on transformers 4.x, `{"warmup_steps": r}` on
+        5.15+, to be splatted into the `TrainingArguments(...)` call.
+    """
+    from dataclasses import fields
+
+    from transformers import TrainingArguments
+
+    names = {f.name for f in fields(TrainingArguments)}
+    key = "warmup_ratio" if "warmup_ratio" in names else "warmup_steps"
+    return {key: warmup_ratio}
```
Xunzhuo, line 62: [P2] Preserve the full-run warmup ratio at the 1.0 boundary

The feedback trainer exposes `--warmup_ratio`, and `1.0` was a valid value in Transformers 4.x. Forwarding it as `warmup_steps=1.0` changes it into an absolute count on 5.15+: for 100 training steps, the previous API produces 100 warmup steps while this helper produces only one. The new test includes `1.0` but only checks that the value is forwarded, so it misses the schedule change. Handle this boundary explicitly without silently changing full-run warmup into one step, and test the resulting `get_warmup_steps` value.

Xunzhuo (changes requested):
The previous import and optimizer-step reconstruction issues are fixed. The 656 training contract cases and focused import checks pass, but the valid full-run warmup ratio still becomes a single step on Transformers 5.15+. Please handle that boundary and test the resulting schedule. Full GPU training was not repeated.

Xunzhuo:
Thanks for the compatibility work and the follow-up on Trainer-owned warmup calculation. This is now covered by merged #3944: all ten affected call sites use the shared TrainingArguments adapter, which preserves fractional warmup across the old/new APIs, keeps the logging destination, and handles standalone imports. The merged implementation has compatibility tests and CPU Trainer checks on 5.14.1, 5.15.0 and 5.17.0. Closing this PR as superseded; #3925 is resolved on main.

basil-k-aji-dev:
Superseded by #3944, which landed `training_args_compat.py` and closed #3925. Closing this.

Your review comment here and that merge arrived a few hours apart, and the merged version already answers the boundary you raised:

```python
# In the new API 1 means one step, not 100% of the run.
if ratio == 1:
    raise ValueError(
        "For full-run warmup, set warmup_steps to the total "
        "training steps with this Transformers version"
    )
```

You were right about the defect, and I had reproduced it before finding the merge:

```
 ratio | 4.x (warmup_ratio) | 5.15+ (warmup_steps=ratio)
   0.5 |                 50 |                         50
   1.0 |                100 |                          1   <-- breaks
```

`train_feedback_detector.py` exposes `--warmup_ratio` on the command line, so `1.0` was reachable, and my helper would have turned a full-run warmup into a single step with nothing to indicate it.

Two things the merged version does that mine did not, worth recording:

**`logging_dir` is preserved rather than dropped.** I removed the argument; #3944 forwards it through `TENSORBOARD_LOGGING_DIR`, so TensorBoard output keeps working. Mine would have silently moved people's logs.

**Signature inspection rather than a field-name check.** `inspect.signature(arguments_type).parameters` is the more general form of what I was doing with `dataclasses.fields`, and it works for any arguments type rather than only a dataclass.

Thank you for both reviews on this PR. The `world_size` default @FAUST-BENCHOU caught and the `1.0` boundary you caught were both real, and both were in code I had convinced myself was simple.

No action needed from anyone here. #3950 is unaffected and I have replied to your review there separately.

Issue #3846, opened 2026-09-16
[Bug] Router Memory Redis cache conflates distinct hybrid and adaptive retrieval policies


JiaoliangYu:
## Goal

Prevent the optional Router Memory Redis hot cache from reusing retrieval results produced under a different hybrid or adaptive retrieval policy.

Today, `memory.cacheKey` includes the query, user scope, project scope, limit, threshold, and memory types, but omits `RetrieveOptions.HybridSearch`, `HybridMode`, and `AdaptiveThreshold`. `CachingStore.Retrieve` returns a cache hit before invoking the backing store, while the Milvus and Valkey implementations use those omitted fields to change candidate depth, reranking, and threshold selection.

As a result, two requests with the same user and query can share one cache entry even though they require different result ordering or filtering. This can happen across decision-scoped memory policies, or across Router replicas and reloads that share the same Redis cache prefix.

## Proposed Workgroup

Agentic & Context

## Scope

- Make the Router Memory Redis cache key distinguish vector-only, hybrid, hybrid-mode, and adaptive-threshold retrieval behavior.
- Normalize the effective hybrid mode so semantically equivalent defaults do not create unnecessary duplicate entries.
- Version the cache-key encoding so entries written with the incomplete key cannot be interpreted under the corrected contract.
- Add deterministic key-level tests for each result-shaping retrieval option.
- Add a cache regression test showing that a result cached for one retrieval policy is not returned for a different policy with the same user, query, limit, and threshold.
- Preserve the existing per-user index, TTL, and invalidation behavior.

## Non-goals

- Change Milvus or Valkey vector search, hybrid reranking, or adaptive-threshold algorithms.
- Add or change public memory configuration fields.
- Redesign cache invalidation or introduce a new cache backend.
- Define cross-generation cache identity for embedding models, collections, or backend migrations.
- Change Router Memory retrieval behavior when the Redis hot cache is disabled.

## Completion signal

- Retrievals that differ only by `HybridSearch`, effective `HybridMode`, or `AdaptiveThreshold` use different Redis value keys when their results may differ.
- Equivalent retrieval policies continue to produce stable keys.
- A cached vector-only result cannot satisfy a hybrid or adaptive retrieval request without consulting the backing store.
- Existing per-user invalidation tests continue to pass.

Issue #3811, opened 2026-09-15
[Bug] Kubernetes CRDs can report Ready before runtime activation succeeds


Kritoooo:
## Proposed Workgroup

Enterprise & Environment

## Describe the bug

On main at `342d4523c`, the Kubernetes controller publishes a validated candidate through `config.Replace` and returns success. The reconciler then reports `Ready=True`, while the ExtProc watcher applies the candidate asynchronously.A later runtime reload failure is logged without reaching the reconciler. The reported CRD state can therefore disagree with the serving configuration.

The watcher also subscribes without consuming an already-published snapshot, so startup ordering can leave an update unapplied until another publication. These existing activation gaps were traced while reviewing #3759, whose original scope is static configuration validation (#3758).

## To Reproduce

Use a controlled runtime test in Kubernetes mode:

1. Submit a CRD candidate that passes validation and model preparation.
2. Block or fail the runtime reload after the controller publishes it.
3. Observe the controller result, CRD Ready condition and active runtime generation.

Current code acknowledges publication independently of the runtime result.
Also cover publishing a candidate before the watcher subscribes.
This report is based on the code path; a live-cluster reproduction is not claimed.

## Expected behavior / acceptance

- [ ] Acknowledgment waits for successful activation; first activation also requires a working listener.
- [ ] Invalid or failed candidates preserve the serving generation and report actionable failure status.
- [ ] Startup consumes the latest candidate, and retry/cancellation handling does not lose pending work.
- [ ] Status write failures do not block activation or rebuild unchanged active configurations.
- [ ] CRD replacement, conflicts and external status edits converge without acknowledging unactivated generations.
- [ ] Helm probes, namespaces, model progress and waiting-state consumers agree with the activation lifecycle.
- [ ] Initial health behavior, startup-probe budget, terminal watcher failures and upgrade requirements are explicit and tested.

## Additional context

Activation and deployment changes extracted from #3759 are tracked here. Related: #3189, which owns the broader versioned operational-status contract. This fix covers the current single-process activation path; versioned status schemas and multi-replica aggregation remain in that broader work.

PR #3944, opened 2026-09-19
[Bug] Resolve verified v0.4 release regressions

### LINKED ISSUE #2676: [Bug] Validate completeness for alias-form embedding model paths

### LINKED ISSUE #2691: [Bug] Prevent semantic-cache false hits on negated or antonym queries

### LINKED ISSUE #2727: [Bug] Guard the in-memory semantic cache against polarity mismatches

### LINKED ISSUE #2729: [Bug] Kubernetes reconciliation double-prefixes canonical backend_ref names
Current-main update: GET/PUT now preserves the submitted canonical document, so the original management API reproducer is fixed. The remaining defect is the Kubernetes reconciler's canonical export/import cycle: runtime endpoint names are exported unchanged and prefixed again on import. This adds one extra model prefix per reconciliation; it does not grow indefinitely because the static base remains fixed. Track that remaining round-trip behavior here and keep #2730 as the implementation link.

Original report (historical context):

### Proposed Workgroup

Enterprise & Environment

### Describe the bug

`PUT /config/router` is not idempotent when it receives the canonical document returned by `GET /config/router`.

The Dashboard configuration path does not show this behavior: it reads and writes `CanonicalConfig` directly through `/api/router/config/all` and `/api/router/config/update`. The management API instead parses the submitted canonical document and exports it again through `CanonicalConfigFromRouterConfig`, which prefixes canonical backend reference names a second time.

### To Reproduce

Start with a valid v0.3 provider model whose canonical backend reference is:

```yaml
providers:
  models:
    - name: test-model
      backend_refs:
        - name: test-model_primary
          endpoint: localhost:8000
          protocol: http
```

1. Read the active document with `GET /config/router`.
2. Submit that response unchanged with `PUT /config/router`.
3. Read the document again.

### Expected behavior

Submitting the canonical document returned by `GET /config/router` should not change it.

### Impact

- External control planes cannot safely perform read-modify-write updates.
- Configuration drift checks report false mismatches.
- Rollback code that republishes a previously read document can mutate it again.
- Repeated retries can keep extending backend reference names.

##### Actual result

The backend reference name becomes:

```yaml
name: test-model_test-model_primary
```

Repeating the same round trip adds another prefix.

##### Root cause

`canonicalEndpointName` always returns `modelName + "_" + backendRef.Name`, even when `backendRef.Name` already starts with the generated model prefix. The management update path invokes `CanonicalConfigFromRouterConfig` after parsing the submitted document.

The latent prefix behavior was introduced by #1553, and the management update normalization path was introduced by #1660. The issue is present in v0.3.0 and current main.

##### Proposed fix

Make canonical endpoint-name generation idempotent when the backend reference already contains the model prefix, and add regression coverage for:

- canonical export of an already-canonical backend reference;
- an unchanged `GET /config/router` → `PUT /config/router` round trip;
- repeated round trips remaining stable.

### LINKED ISSUE #3461: [Bug] Three local developer workflow issues in docker.mk and pre-commit config

### LINKED ISSUE #3470: [Bug] Dashboard Monitoring embeds a stale Grafana goto URL
### Proposed Workgroup

### Describe the bug

"dashboard/frontend/src/pages/MonitoringPage.tsx" embeds:

~~~
/embedded/grafana/goto/llm-router-metrics
~~~

Grafana "/goto" paths are short-link records, not stable dashboard identifiers. A fresh "vllm-sr serve" stack provisions the semantic-router dashboard with UID "vllm-semantic-router", but it does not provision a short link named "llm-router-metrics". The Dashboard availability probe can succeed while the Monitoring iframe opens a missing or stale short link.

### To Reproduce

1. Build and run "vllm-sr serve" from "main" at "cd250d0e" with the default Grafana service.
2. Open **Operate → Monitoring** in the Dashboard.
3. Inspect the iframe request to "/embedded/grafana/goto/llm-router-metrics".
4. Observe that the short-link target is unavailable in a fresh Grafana instance.
5. Open "/embedded/grafana/d/vllm-semantic-router/vllm-semantic-router-dashboard?orgId=1"; the provisioned dashboard loads successfully.

### Expected behavior

Monitoring should embed the stable UID and slug of the dashboard provisioned by "llm-router-dashboard.serve.json", or derive the URL from the same provisioned artifact. A frontend test should assert that the configured Monitoring target matches the provisioned Grafana dashboard contract.

### Impact

The top-level Grafana health check is green, but users can still receive an error inside the Monitoring iframe after a fresh install, container recreation, or production deployment. The workaround is to replace the "/goto" path with the stable "/d/vllm-semantic-router/vllm-semantic-router-dashboard" path.

### Additional context

The issue is independent of Router traffic and Prometheus health. Metrics can be collected correctly while the Dashboard entrypoint to those metrics remains broken.

### LINKED ISSUE #3500: [Bug] Streaming rejects a message_delta without stop_sequence that the buffered path accepts
### Proposed Workgroup

Data Plane & Networking

### Describe the bug

The router applies two different presence rules to the same Anthropic field depending on the
response path.

Anthropic's published OpenAPI schema lists `stop_sequence` under `required` for both `Message`
and `MessageDelta`, with a nullable type. Required governs presence, nullable governs value.
Source: the spec the official SDKs are generated from, linked as `openapi_spec_url` in
`anthropics/anthropic-sdk-typescript/.stats.yml`
(https://storage.googleapis.com/stainless-sdk-openapi-specs/anthropic/anthropic-e50cf35b74cc0471a2b5af7ea03765aa81c035f82588e9a1ba1b29aeaa17d064.yml,
read 2026-09-04):

```
MessageDelta.required: [container, stop_details, stop_reason, stop_sequence]
Message.required:      [id, type, role, content, model, stop_reason, stop_sequence, stop_details, usage, container]
stop_sequence:         {anyOf: [{type: string}, {type: null}], default: null}
```

The SDK type agrees (`stop_sequence: string | null`, not optional, in
https://github.com/anthropics/anthropic-sdk-typescript/blob/4140e0eaa597c0ad35218ffb20b66ef7fce7f639/src/resources/messages/messages.ts#L2802).

- **Streaming:** after the typed payload checks (a delta and usage must exist, and a matched value is
  required when the reason is `stop_sequence`), `provider_stream_validation.go:262` runs
  `requireProviderFields(delta, "stop_reason", "stop_sequence")` on every `message_delta`: both keys
  must be present, and null is accepted for the other stop reasons. A delta without the `stop_sequence` key fails the request with
  `stream_required_field` (category `upstream_unavailable`). That is the schema's reading, and it is
  what the repo's own fixtures do: the `providerRequiredFieldCases` `end_turn` row at `:304` carries
  `"stop_sequence": null` explicitly.
- **Buffered:** `provider_response_validation.go` has no presence check for `stop_sequence` on a
  `Message`. `official_schema_response_test.go:197` (`assertEncodedChatResource`, body at `:202`)
  feeds `TranslateResponse` a body carrying `stop_reason` and `usage` but no `stop_sequence` key,
  and expects success.

Same field, same schema requirement, opposite verdicts.

The stream also disagrees with itself: `message_start` carries a full `Message`, validated by
`validateAnthropicMessageStart` (`provider_stream_validation.go:148`) through the same presence-blind
`validateAnthropicResponseResource` (`provider_response_validation.go:218`), so the opening event may
omit `stop_sequence` while the closing `message_delta` may not.

Scope: `stop_sequence` only. The schema also lists `container` and `stop_details` as required and
nullable; the router requires neither on any path, and Anthropic's own streaming examples omit them,
so those two are already consistent and out of scope here.

The streaming presence branch is untested (the positive and mismatch cases are covered by
`019-anthropic-stop-sequence-stream-in.json` and the two `anthropic_stop_sequence_reason` rows).
`providerRequiredFieldCases` (`stream_contract_responses_test.go:288`) covers a `message_delta`
without a delta, without usage, and the two `anthropic_stop_sequence_reason` cases, but has no row
for a delta missing the `stop_sequence` key, or the `stop_reason` key: the presence branch at
`provider_stream_validation.go:260-262` has no rejection coverage at all. The golden rejection
fixtures that do produce `stream_required_field` on the Anthropic side (036, 045) are
`content_block_delta` cases. The neighbouring rule `anthropic_stop_sequence_required`
(`provider_stream_validation.go:191`) has no streaming test either, while its buffered counterpart
(`provider_response_validation.go:247`) is covered at `provider_response_validation_test.go:51`.

This is not theoretical. vLLM's Anthropic-compatible `/v1/messages` omits the key on both paths,
verified on 0.26.0 and on main as of 2026-09-04: streaming drops it via `exclude_unset` (fix proposed in
vllm-project/vllm#55325), and non-streaming drops it via `exclude_none` (separate, unfiled). Through
the router, the same request to vLLM succeeds with `"stream": false` and fails one event from the
end with `"stream": true`, after the complete answer has already been delivered. Claude Code
sends `"stream": true` on its requests (captured 2026-09-01), so a session against such a backend
fails at the terminal event with the answer already on screen. Verified by capturing vLLM's SSE on 0.26.0 and by the regression test
in vllm-project/vllm#55325, which fails on vLLM main without the fix.

### To Reproduce

Unit level, no backend needed, at main `87f446d` (2026-09-04):

1. `official_schema_response_test.go:202` already passes with `stop_sequence` omitted from the
   buffered response. That is the accepting path.
2. Push the same omission on a terminal `message_delta` into `newProviderStreamDecoder(anthropic)`,
   framed as the existing table rows are:
   ```
   "event: message_delta\ndata: {\"type\":\"message_delta\",\"delta\":{\"stop_reason\":\"end_turn\"},\"usage\":{\"input_tokens\":1,\"output_tokens\":1}}\n\n"
   ```
   The push itself is rejected with `stream_required_field`
   (`Upstream stream event is missing required field stop_sequence`).

Live, as observed. Backend: vLLM 0.26.0, `vllm serve Qwen/Qwen3.8-27B-FP8` on port 8000. Router:
one model with `api_format: anthropic` and a `backend_refs` entry pointing at that port. Send the
same request twice through the router:

```
curl -N -X POST http://<router>/v1/messages \
  -H 'content-type: application/json' -H 'x-api-key: x' -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"<routed-model>","max_tokens":64,"stream":false,"messages":[{"role":"user","content":"Reply with exactly: OK"}]}'
# 200; the router byte-replays the backend body (same-format passthrough), so the client
# receives it with no stop_sequence key (observed 2026-09-04):
# {"id":"chatcmpl-...","type":"message","role":"assistant","content":[...],"model":"Qwen/Qwen3.8-27B-FP8","stop_reason":"end_turn","usage":{"input_tokens":57,"output_tokens":27}}

curl -N -X POST http://<router>/v1/messages ... -d '{... "stream":true ...}'
# every content token arrives, then the terminal event fails validation (observed 2026-09-02 at 968f8ce7)
```

The backend's terminal event, captured directly from vLLM:

```
{"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"input_tokens":60,"output_tokens":37}}
```

Workaround until the producer fix lands: run vLLM with vllm-project/vllm#55325 applied, which makes
it emit `"stop_sequence": null`. A local relaxation of the check at `:262` also works but is the
wrong fix for the reason above.

### Expected behavior

One policy for both paths. Two consistent choices exist:

- Strict on all three surfaces (buffered `Message`, `message_start.message`, `message_delta.delta`) for
  `stop_sequence`, matching the existing `message_delta` check. The buffered
  path then also rejects a body without the key, which today means vLLM non-streaming responses
  fail until that producer bug is fixed upstream too.
- Lenient on both for this one field, treating an absent key as null with a warning. The four
  official SDKs we checked (TypeScript, Python, Java, Go) parse a response with the key absent
  without error.

My recommendation is strict on all three. This is the one required-nullable field the validator already
enforces on the stream, so the ask is consistency, not schema-wide presence enforcement: the router's own encoders already emit the key on both paths
(`stream_anthropic.go:80` and `codec_anthropic_response.go:17` serialize `stop_sequence` without
`omitempty`), and the producer bug is being fixed at its source. The cost is
that a buffered response from today's vLLM would start failing through the router too, until the
non-streaming producer fix lands; leniency has the smaller blast radius if that matters more. For
scale: a strict draft I have locally touches the decoder plus about a dozen test files whose fixtures
build minimal Anthropic bodies without the key, while a lenient draft touches the stream validator
and one fixture. Strict also means Anthropic bodies already sitting in the response cache without the
key would fail replay with a 502 until they age out. The maintainers own the call, and I will take the PR either way once the direction is set. If
strict is the direction, the least disruptive shape looks like the package's existing seam: gate the
presence rule on `Policy` and report an absent key through `appendProviderFieldOmission` as a
diagnostic for one release, then reject, so already-cached bodies and older producers get a window
rather than a 502 on upgrade.

Strict also has to say what happens to `message_start`. The consistent choice is the same presence
check on its `message`, which means the minimal fixtures at `stream_contract_anthropic_test.go:208`
and `:241` gain `"stop_reason": null, "stop_sequence": null`, as a real Anthropic `message_start`
carries. If you would rather leave `message_start` lenient, since its message is never the terminal
state, the scope narrows to buffered versus `message_delta` and I will write it that way.

Either way, one `providerRequiredFieldCases` row is unconditional: `stop_reason: "stop_sequence"`
with a null value, exercising the untested rule at `:191`. The missing-key case is a rejection row
in that table under the strict direction, or an accepting capability fixture (in the shape of
`019-anthropic-stop-sequence-stream-in.json`) under the lenient one, since that table only holds
rejections.

### Impact

Any Anthropic-format backend that omits `stop_sequence` on otherwise valid `end_turn`, `max_tokens`
or `tool_use` responses presents as a working server that fails only when streaming (an omission on a
`stop_sequence` reason already fails both paths through the matched-value rule). vLLM is one such backend today. Clients see the stream fail on its last
event with the answer already on screen, which reads as a router fault rather than a backend one. Observed error, router at `968f8ce7`:

```
category: upstream_unavailable
code:     stream_required_field
message:  Upstream stream event is missing required field stop_sequence
```

### Additional context

- Producer side: vllm-project/vllm#55324 (issue) and vllm-project/vllm#55325 (fix, open) make vLLM
  emit `"stop_sequence": null` on the streaming `message_delta`. The non-streaming omission is a
  separate vLLM defect, not yet filed.
- First observed through the router at `968f8ce7` (main, 2026-09-02); code references above are against main
  `87f446d` (2026-09-04).

### LINKED ISSUE #3524: [Bug] Traced Looper responses fail strict protocol translation and return 502

### LINKED ISSUE #3548: [Bug] Dashboard config backups are world-readable and can contain API keys

### LINKED ISSUE #3706: [Bug] golang.org/x/text v0.37.0 is vulnerable to GO-2026-5970 (infinite loop / DoS)
### Advisory

**[GO-2026-5970](https://pkg.go.dev/vuln/GO-2026-5970)** — Infinite loop on invalid input in `golang.org/x/text`.

Parsing certain malformed input with `golang.org/x/text` can enter an infinite loop, causing a denial of service. Affected range: all versions `< 0.39.0`. Fixed in **`0.39.0`**.

### Where it applies

`golang.org/x/text` is an indirect dependency pinned at `v0.37.0` in:

- `src/semantic-router/go.mod`
- `src/semantic-router/go.onnx.mod` (symlink to `go.mod`; the ONNX build lane shares the same declaration)
- `deploy/operator/go.mod`
- `e2e/go.mod`

The Router, Operator, and E2E modules still pin `v0.37.0`. `perf/go.mod` already uses `v0.40.0` on current `main` and no longer needs this bump.

### Context

This came out of the same dependency review as #3425. That PR originally also bumped `google.golang.org/grpc` to `1.83.1`, but `main` has since landed that independently, so `x/text` is the remaining item. #3425 has been rescoped to this bump.

### Fix

Bump every affected module to `golang.org/x/text v0.39.0`. PR #3425 does this for `src/semantic-router` (default module + ONNX modfile; `go.onnx.sum` is a symlink to `go.sum` so both build lanes are covered). `deploy/operator` and `e2e` remain in scope for separate changes. Keep this issue open until those modules and both Router build lanes are patched and validated.

Verified the target version is clean against `api.osv.dev`.

Xunzhuo:
Closes #3548
Closes #3524
Closes #3710
Closes #3846
Closes #3793
Closes #3500
Closes #3800
Closes #3925
Closes #3461
Closes #3718
Closes #2729
Closes #2727
Closes #3706
Closes #2676
Closes #3947
Closes #3949
Closes #3811
Closes #2691

Related #3470, #3688, #3515, #2467

## Purpose

Fix reproduced release regressions across routing, provider compatibility, configuration lifecycle, training entrypoints and Dashboard interactions.

- Carry Fusion/ReMoM/Workflow evidence through a private typed channel captured before decoration, preserving Chat compatibility while applying strict protocol validation and final output limits. Unsupported or oversized optional trace is omitted atomically with a bounded diagnostic. Honor the original Responses streaming request independently of internal aggregation, and use the Router-owned response ID consistently in JSON/SSE, stored objects and conversation lineage. Preserve required empty text and refusal fields when stored Responses are read back through the shared content serializer. Existing target capability checks remain in force. Accept documented Mistral usage tiers and report compatible Anthropic nullable-field omissions in buffered and streaming responses.
- Use neighboring Unicode runes for literal keyword boundaries, preserve explicit-regex, existing Han, underscore and punctuation behavior, and align Japanese/Korean continuous-text matching with the shared structure helper. Parse final classifier JSON after complete reasoning blocks. Partition Redis results with a versioned encoding of the effective retrieval policy, preserving equivalent defaults, TTL and per-user invalidation.
- Apply the unconditional English lexical polarity floor across in-memory, Redis, Valkey, Milvus, Qdrant and hybrid semantic caches. Read original queries by named fields, reject incomplete or opposite-polarity candidates, and retain valid later candidates within the bounded search. Preserve per-lookup scores, model partitions and selected-entry freshness; Hybrid by-ID reads now reject expired entries. The optional in-memory NLI verifier remains separate. Add pinned PAWS-derived negatives and unchanged human-labeled paraphrase controls with explicit provenance and scope.
- Acknowledge Kubernetes configuration only after serving-generation activation; preserve the previous generation on failure. Report immutable configuration writes explicitly and preserve backend references across export/import.
- Restrict configuration snapshots to their owner, open the provisioned Grafana dashboard and authenticate embedded data queries with the existing CSRF contract, preserve the original public Host and Origin for Grafana Live WebSocket validation while keeping the upstream URL fixed, and make expression actions keyboard operable. Preserve the actually advertised automatic entrypoint in Playground when specialized entrypoints are also available.
- Resolve Tools DB resources from the shared explicit resource root, refresh the path after config saves, and make bundled public Dashboard assets readable by the non-root runtime independently of build-checkout permissions.
- Surface Builder load/import/format/compiler failures without retaining stale deploy output, and clear the error state after a valid recovery.
- Revalidate cached dataset content and metadata even when filesystem timestamp precision hides consecutive edits; parsing/projection/proof caches remain bounded.
- Resolve registered local-model aliases before download discovery, preserve canonical required files and feature gates, and keep remote provider IDs outside local provisioning.
- Synchronize the affected Go text dependency and caller module graphs. Support Transformers 5.15+ argument construction, repair MMLU utility imports, preserve development-image identity and echo-service lifecycle, and assert actual request-header cardinality.
- Allow an idle managed benchmark worker to upgrade its image while retaining its store and credentials; add an absolute benchmark host-port override for occupied default ports.
- Bind the optional ML pipeline sidecar to loopback in the shipped local/sidecar entrypoints. Job/artifact ownership in #2467 remains separate work.

The patch incorporates and adapts contributor work from #2730, #2828, #3810, #3545, #3549, #3869 and #3950, with attribution retained in the commits. The Monitoring URL repair and its provisioned-artifact contract check are inherited from upstream #3938, which has already closed #3470; this PR adds embedded authentication and Grafana Live transport repairs. #3515 is also already closed upstream; this PR retains a scoped keyboard-interaction contribution. The follow-up includes reproduced Builder import/format error handling and Tools DB path/refresh failures, plus a dataset-cache consistency failure found by the rebased test suite. The follow-up cache audit found remote and hybrid callers bypassing the lexical floor; this revision closes those paths rather than treating the remaining executable work as model-quality calibration.

## Test Plan

Run the owning repository checks and before/after regressions at actual caller boundaries: strict codecs, Looper execution, Redis retrieval, runtime activation, configuration persistence, browser keyboard interaction and training argument construction. Build the final candidate images, qualify an isolated AMD local stack, and complete current-revision CI before release qualification.

## Test Result

**Candidate `bc9d8cb14d09f1af23301852928ec495ba72676b` is rebased onto main `cc80bdbda662bcd9ec1cb704d0c50680e760f45e`. [Current-revision product CI](https://github.com/vllm-project/semantic-router/actions/runs/35475759474) passes all 38 declared validation contracts and 9 build receipts selected by the actual changed-path plan, including the official aggregate gates. The tested merge tree equals the candidate tree. Normal AMD qualification, the maintained default-stack upgrade and post-upgrade readback are complete.**

- Current-revision Core 6682, Dashboard 1803, Storage 138, Candle CPU 25 required/57 named and production performance 51 checks pass. Source, inventory, platform and native/image/evidence digests were independently checked. Original benchmark budgets, corpus and comparator remain unchanged; overlapping test counts are not added together.
- Four normal AMD Make targets pass: three candidate images built and the official Envoy availability target passed. The matching Router image passes 51 top-level / 266 named component checks with no failures, skips or missing cases. Source proof covers 4743 blobs across six Router/bindings/CLI subtrees, four native libraries and six pinned model files; this is not a whole-repository source count.
- Seventeen actual routed requests pass across keyword positive/negative controls and Chat/Responses/Anthropic JSON/SSE. Three stored Responses GETs exactly match actual created output, public ID and status, and chained streaming preserves previous_response_id. The strict dynamic-config profile passes its five-stage lifecycle and 3+3+3 routed probes. Current Looper and Responses CI profiles pass their assertions; the combined chained-streaming-plus-GET wire proof comes from the AMD requests, while the CI persistence, standalone SSE and chain cases remain separately identified. The dynamic profile also retains older broad diagnostic sample matches of plugin 6/23, rules 5/6 and embedding 17/31. Their 32 mismatch mappings are unchanged from the preceding candidate, and an earlier candidate run has the same counts. Those cases use an existing nonzero-success criterion and include oracle/configuration mismatches and semantic expectations not established as a calibrated acceptance set for this profile/model; their framework PASS is not evidence of all-sample correctness or plugin blocking. The strict lifecycle result is separately 9/9.
- CI passes the existing acceptance contracts, not every diagnostic sample. EAG records domain 186/261, cache hits 10/15, guard detection 11/12, priority 3/4, plugin chain 22/23, rules 5/6 and plugin configuration 16/18; ML selection is 15/20, with the same 186/261 domain corpus result. Existing criteria are unchanged; some mismatches reflect old oracles/configuration and the separately disclosed model false positive. The 78 KiB streaming request returns an allowed 400 with no decision header. Without its raw response body, this does not independently prove an upstream context-limit rejection or successful complete routing. Anthropic failure-matrix counters describe expected negative controls. Hallucination 5/5 covers deterministic mock warning propagation; response-check streaming cases establish observation after delivery, not blocking. These boundaries remain separate from the strict repair assertions.
- Five actual browser pages pass. Grafana records 40 successful datasource POST observations including automatic refresh repetitions, not 40 distinct queries; Live upgrades to 101 and receives 4 frames. All 128 checked production resources match complete SHA256/MIME; HTML, authentication, API, bridge and WebSocket stay live while verified immutable static bytes are cached. Builder failure/recovery preserves active configuration. ToolsDB contents and non-root public asset access pass. Expected pre-login and nonfatal browser diagnostics remain in the evidence.
- The previous production source reproduces the ordinary opposite-polarity false hit in all five remote/hybrid backends using the same fixtures. The repair passes those actual database tests, the full AMD four-service storage inventory 134/134, and official storage 138/138 including PostgreSQL. Real pinned Candle CPU inference passes eight negation controls, three paraphrases and six PAWS-based cases: three unchanged human-labeled paraphrases and three explicitly local lexical-negative derivations, all above the 0.80 similarity threshold. This establishes the bounded L1 guard and positive-control behavior, not universal equivalence or full PAWS accuracy.
- The default AMD service was upgraded through the normal CLI with fresh online SQLite backups. Post-upgrade and post-cleanup entrypoint/SSE requests, database integrity, preserved configuration/users, Redis continuity and backend continuity pass independent readback. Final demo cleanup removes only its owned containers, networks and forwarding. Persistent state, models, deployment/rollback images and execution evidence are retained; earlier cleanup removed only verified unused task-specific database images and historical temporary inputs.
- Three actual Transformers versions (5.14.1,5.15.0,5.17.0) each completed four synthetic CPU Trainer steps on the earlier 53d revision. The complete tracked training tree/helper is identical at this candidate; this is source-equivalent evidence, not a new full GPU training run.

The AMD application stack uses three MIGraphX classifiers and ORT CPU embedding; the component semantic checks use Candle CPU. Logical backend aliases share one physical model, so no model-diversity or answer-quality improvement is claimed. This run upgrades an existing isolated demo; fresh installation/onboarding is retained as separately attributed prior evidence. Native macOS ARM, full Windows ACLs, complete GPU training, held-out model-quality evaluation and every Grafana panel/administrative operation were not run. The offline local umbrella previously stopped at an unavailable native prerequisite, and a broad offline ExtProc invocation lacked unrelated checkpoints; neither is counted as a full pass. Exact-current owning CI supplies the repository evidence.

[Community single-owner admission](https://github.com/vllm-project/semantic-router/actions/runs/35475767755/job/105984642191) is the maintainer's explicitly accepted exception for this unified cross-workgroup fix. Issue ownership and governance policy are unchanged; the maintainer will merge manually. This task does not merge or publish a release.

A further bounded model-quality risk remains unresolved: the fixed PII revision `6d3300c4bd7975f30a664503f6c725cf1fbbad48` predicts the ordinary fragment `electric` as TITLE in both the production Candle CPU path and a same-file Transformers 5.17 FP32 CPU reference using the [pinned model card's documented SIMPLE aggregation](https://huggingface.co/llm-semantic-router/Vela-1.0-Encoder-307M-PII/blob/6d3300c4bd7975f30a664503f6c725cf1fbbad48/README.md). The two other ordinary diagnostic blocks correspond to GPE/date entities under the existing deny-all policy, which conflicts with their old clean-case expectations. These three diagnostic inputs do not establish population error rates, a particular training cause or fresh full HTTP/backend parity. No threshold, aggregation, model or service was changed; held-out quality and retention evidence is required for a model repair. This is outside the 18 closing references.

Issue #3831 remains a confirmed release risk: benign long-input false positives occur through the real default model/window path and need a separate held-out evaluation before changing thresholds or aggregation. The completed #2691 scope is the unconditional English lexical floor across maintained semantic backends; cue-less, word-order-only and non-English equivalence and optional NLI remain separate enhancements. #3470 and #3515 are already closed upstream. #3688 and #2467 retain their separate persistence and ownership work; other audited issues retain specific implementation, validation or external-dependency conditions.

---
<details>

- [x] PR title begins with one bracketed category.
- [ ] Linked accepted work satisfies single-owner admission (explicit maintainer exception for this unified fix).
- [x] Commits are signed off.
- [x] Purpose, test plan and results describe actual scope and completed validation.

</details>


FAUST-BENCHOU:
The regression fixes I checked are scoped carefully, especially the config-snapshot hardening: owner-only files, symlink rejection, atomic replacement, and upgrade-time permission repair cover the credential exposure path rather than just masking it with chmod. No blocker from my pass; I’d wait for the stated final-revision qualification before approval.

Issue #3939, opened 2026-09-19
[Feature] Record which window decided a windowed guard score


subin9:
### Goal

Record which window produced a windowed guard score, so a score one window decided can be told apart from one the whole document supports.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- `riskiestWindow` in `src/semantic-router/pkg/classification/classifier_jailbreak_window.go` selects the window with the largest positive mass and returns its distribution alone. The binding already hands it one `LabelDistributionWindow` per window, carrying `Start`, `End` and the probabilities (`src/semantic-router/pkg/modelruntime/tasks/label_scores.go`), so the offsets exist and are dropped one layer above them.
- Carry the selected window's offsets and the number of windows scanned on the classification result, and surface them where the signal value is already surfaced: the routing preview trace and the replay record's signal values.
- `InputUsage` carries `OriginalTokens`, `ProcessedTokens` and `Truncated`, which answers whether the document was covered, not which part of it decided.

### Non-goals

- Changing the aggregation. Keeping the riskiest window is deliberate and fails closed, and this issue does not propose replacing it.
- Truncation counters, which are #3372. That issue covers a classifier silently scoring part of its input and leaves the result contract out of scope. A windowed scan is the opposite case: coverage is complete, `validateJailbreakInputCoverage` rejects the result when it is not, and nothing there reports which covered window carried the score.
- Chunk budget sizing, which is #3367.
- Per-window scores in the public classification API response.

### Evidence

On the four fixed inputs in #3831, scored against the checkpoint that issue names, two benign public descriptions read 1.742e-07 and 3.888e-07 on their first window and 0.946248 and 0.975538 on their second. Both are blocked at the configured threshold, and the signal value a reader sees is the tail window's score, indistinguishable in the trace from a score the whole document produced. That report spent its investigation on AMD against CPU, on transport shapes and on the label map before the scan explained the numbers.

### Completion signal

A preview trace and a replay record for a document longer than one window name the window that produced the score and how many windows were scanned, and a document that fits one window reads as one window.

Xunzhuo:
Completed by #3945. The selected guard window’s token offsets and scan count now travel with its score into preview and Replay signal values. Multi-window and single-window behavior are covered, and the riskiest-window aggregation is unchanged.

PR #3945, opened 2026-09-19
[Feature] Report which window a guard score came from

subin9:
Related #3939

## Purpose

Guard scans a prompt longer than one window in overlapping windows and keeps the riskiest one, and the result carries that distribution alone. The binding hands the backend one entry per window with its offsets, and they are dropped a layer above it, so a score a single window decided reads exactly like a score the whole prompt produced.

The difference is not academic. On the four fixed inputs in #3831, two benign public descriptions read 1.742e-07 and 3.888e-07 on their first window and 0.946248 and 0.975538 on their second. Both are blocked at the configured threshold, and nothing in the trace says the decision came from the tail of the document. That report worked through AMD against CPU, three transport shapes and the label map before the scan explained the numbers.

The window a scan kept and the number of windows it read now travel with the result and reach `signal_values` next to the score, which is the map the routing preview trace and the replay record already carry. The aggregation is unchanged: the riskiest window still wins, which is the fail closed behaviour a guard wants.

## Test Plan

- Pin the provenance at the backend: a two window scan whose second window carries the risk reports that window's offsets and a count of two.
- Pin what the signal publishes: a scanned result adds the three window values next to `jailbreak:<rule>`, and a result carrying no window adds none, so an unwindowed deployment reads as it does today.
- Drive the real boundary: a new E2E sends a prompt longer than one window and a prompt that fits in one to `/api/v1/routing/preview` on the deployed router and reads the window out of `signal_values`.

## Test Result

The E2E passes against the deployed baseline profile. A benign geology prompt of 1,536 tokens is scanned in six windows and the score the threshold reads comes from `[1275, 1536)`, the last one, while a one sentence prompt reports a single window starting at zero. The case is registered in the baseline contract list, so it runs in that lane rather than only being discoverable.

The unit coverage pins the same contract without a cluster: the backend reports the window it kept rather than the first one, and the signal publishes `window_start`, `window_end` and `windows` only when the result carries a window.


FAUST-BENCHOU:
Nice observability improvement. The provenance stays attached to the exact window whose distribution wins the existing risk aggregation, and the preview E2E verifies both multi-window and single-window behavior without changing the guard decision semantics. LGTM.

Xunzhuo (approved):
The selected window stays attached to the maximum-risk result through signal values, and results without a scan keep the existing behavior. Verified the three new regression tests, the 106-case E2E unit suite, Candle/ORT CPU model checks, and the deployed long/short-prompt assertion on this head. The deployed case reports six windows with the selected span [1275, 1536). No blocking findings.

wilsonwu (approved):
LGTM

PR #3948, opened 2026-09-19
[Docs] Describe how matching decisions are ranked

subin9:
Related #3081

## Purpose

The decision tutorial documents one ordering rule, that a higher `priority` wins when several decisions match. The engine ranks through three different branches, and which one runs depends on whether a matched decision carries a tier. `decision.tier` is not mentioned anywhere in the tutorial, yet 52 decisions across the six shipped configs set it, so all of them follow a rule the docs do not describe.

This is DR-01 from the ranking proposal merged for this issue. It records the ordering table as the engine runs it today and pins the table with unit tests, so the runtime slices that follow have a documented baseline to change rather than a rule nobody wrote down.

Two consequences are documented explicitly because they are easy to read past. The tiered branch never consults `routing.strategy`, and the `priority` branch is the one branch that does not push catch-alls last.

The compatibility section of the merged proposal counted only the tiers that hold more than one decision, and missed one shape that ships today: `config/config.yaml` sets `strategy: priority` and mixes 24 untiered decisions with a single tier 2 decision. Tiered ranking activates from a matched decision, so whenever that one decision matches, the untiered pool ranks by confidence instead of the configured priority. A test covers that case.

No runtime behavior changes here.

## Test Plan

`go test ./src/semantic-router/pkg/decision/`

The new `selection_order_test.go` pins one case per documented row: tier 0 ahead of a tiered decision, the ranking mode switching when an unrelated tiered decision matches, a catch-all outranking a real match under `strategy: priority`, an extra keyword match dropping a scored pool to priority ordering, an `AND` aggregate falling as a decision gains evidence, and a boolean `conversation` rule reporting 1.0 as evidence.

## Test Result

The six cases pass against the current engine, which is the point of the file: they describe today's behavior, not the contract that replaces it. The later slices in the proposal will change four of them, and that diff is the record of what each slice changed.

Not covered here: nothing runs the routing preview or a recipe probe, since this PR changes no selection behavior. The cross-route collision probes stay in DR-08.


Xunzhuo (approved):
The selection table matches the current ranking and confidence-pool implementation. All six new ranking cases passed in the exact-head Core run, and the documentation checks passed. The distinction between current behavior and the proposed semantics is clear.

PR #3950, opened 2026-09-19
[Bug] Resolve the tools database path against the project root

### LINKED ISSUE #3949: [Bug] Tools database paths fail to resolve in the dashboard backend
### Proposed Workgroup

### Describe the bug

`/api/tools-db` cannot resolve the database in a default checkout. With a session:

```
HTTP 404
Tools database not found: open D:\sr-3569\config\config\tools_db.json: The system cannot find the path specified.
```

`config` is doubled, and the file is not there either.

`resolveToolsDBPath` (`dashboard/backend/router/core_routes.go:160`) builds the fallback as
`filepath.Join(cfg.ConfigDir, "config", "tools_db.json")`. But `ConfigDir` is already the directory
holding the configuration file — `dashboard/backend/config/config.go:253-261` derives it as
`filepath.Dir(absConfigPath)` — so the join appends a second `config`.

The same value is read correctly a few lines away, which makes this a disagreement rather than a
platform quirk: `core_routes.go:233` uses `filepath.Dir(cfg.ConfigDir)` to reach the project root,
which only works if `ConfigDir` is the configuration directory.

Even with a single `config`, the target does not exist. The database ships at
`config/runtime/tools/tools_db.json`, and `config/config.yaml:2254` already names that location.
That value is not reaching the fallback: the error above prints the fallback string, which is
returned only when `toolSelection.ToolsDBPath` is empty (`core_routes.go:166-168`).

The tools integration is enabled by default (`global.integrations.tools.enabled: true`), so this is
the default state rather than an opt-in one.

### To Reproduce

With the backend running against the repository's own `config/config.yaml` — what
`make dashboard-dev-backend` does — request `/api/tools-db` with a session, or open the
Configuration page and watch the tools request.

### Expected behavior

The fallback should point where the database actually lives, and a configured `tools_db_path`
should win when it is set. Since `ConfigDir` is read as the configuration directory in one place
and as the project root in another, it would also help to state which one it is meant to be:
`core_routes.go:233` and the derivation in `config/config.go` agree with each other, while
`core_routes.go:160` and the comment in `dashboard/backend/router/recipe_store.go:15` are the
outliers.

### Impact

The tools database backs the Configuration page's tool-selection surface. In a default checkout it
never loads, and the 404 is indistinguishable from a genuinely missing database because the path in
the error is one the repository has never used.

### Additional context

The fallback was added by `8c912975d` (#1375, "fix error where tools_db.json couldn't be found"),
which changed `Join(configDir, "tools_db.json")` into
`Join(cfg.ConfigDir, "config", "tools_db.json")` — a fix that assumes `ConfigDir` is the repository
root. The file itself later moved under `runtime/tools/` (#1553), so the current path is wrong for
two reasons that arrived at different times. No open issue or PR mentions `resolveToolsDBPath`.

basil-k-aji-dev:
Fixes #3949

`/api/tools-db` answers 404 in a default checkout. `resolveToolsDBPath` returned a path that does not exist, by two separate routes.

## The relative path — what users actually hit

A configured `tools_db_path` is relative to the project root, which is how `config/config.yaml` spells it:

```yaml
global.integrations.tools.tools_db_path: config/runtime/tools/tools_db.json
```

That value was returned unchanged and handed to `os.ReadFile` (`handlers/tools.go:21`), so it resolved against the process working directory. For the documented dev launch that directory is `dashboard/backend` — `tools/make/dashboard.mk:40` runs `cd $(DASHBOARD_BACKEND_DIR) && go run main.go` — where no such file exists:

```
cwd              = .../semantic-router/dashboard/backend
configured value = "config/runtime/tools/tools_db.json"

BEFORE -> config/runtime/tools/tools_db.json                      exists: false
AFTER  -> .../semantic-router/config/runtime/tools/tools_db.json  exists: true
```

## The doubled fallback

The fallback joined `"config"` onto `cfg.ConfigDir`, which is already the directory holding the config file — `config.Load` derives it as `filepath.Dir` of the absolute config path — so the result carried `config` twice. `core_routes.go:233` already reads `ConfigDir` the other way, taking `filepath.Dir` of it to reach the project root, so the two were inconsistent within one file.

Relative paths are now anchored to the project root; absolute ones are returned as given, so a deployment pointing the database outside the checkout still works.

## Two corrections to the issue's diagnosis

**The doubling is not the default failure.** The issue says the configured value is not reaching the fallback. It is — the canonical default populates `ToolsDBPath`, so `ReadToolSelection` returns `config/runtime/tools/tools_db.json` and the doubled branch is only reached when the config will not parse. I checked with the repository's own parser before changing anything:

```
ReadToolSelection err = <nil>
ToolsDBPath (as read) = "config/runtime/tools/tools_db.json"
```

Both problems are fixed; only the relative path is what a user hits in a default checkout.

**The router's default is not stale.** The issue notes the database ships at `config/runtime/tools/tools_db.json` and suggests the fallback should point there. I left the router's `config/tools_db.json` default alone, because both are correct for different installations: every chart and manifest under `deploy/` mounts a `tools_db.json` at `config/tools_db.json` and names it explicitly, while this repository's own `config.yaml` names the runtime path. Moving the shared default would break those deployments. The dashboard now defers to the router's default and only makes it absolute.

## Tests

Four new tests — relative configured path, absolute configured path, absent key, unparseable config. All four fail against the previous revision:

```
--- FAIL: TestResolveToolsDBPathResolvesRelativeConfiguredPathAgainstProjectRoot
--- FAIL: TestResolveToolsDBPathDoesNotDoubleTheConfigDirectory
--- FAIL: TestResolveToolsDBPathFallbackIsUsedWhenTheConfigWillNotParse
--- FAIL: TestResolveToolsDBPathFallsBackWhenRouterContractCannotParse
```

**One existing test changes its expectation, and that deserves scrutiny.** `TestResolveToolsDBPathFallsBackWhenRouterContractCannotParse` asserted `filepath.Join(configDir, "config", "tools_db.json")` — the doubled path — so it pinned the bug rather than a requirement. It now asserts the same fallback anchored to the project root. `TestResolveToolsDBPathUsesRouterContractPath` is untouched and still passes, since absolute configured paths are unchanged.

`go test ./router/` passes. The package was green before this change as well, so the delta is only the four new cases.

## Not verified

No Windows run. The reporter's session was on Windows and the paths in their error are Windows paths; the fix is `filepath`-based throughout, but I could not exercise it there.


basil-k-aji-dev:
CI's `Check linked accepted issue` is failing on this PR:

```
At least one linked issue must carry the `accepted` label.
```

That is the lifecycle gate rather than anything in the change — #3949 is still `needs-acceptance`, and applying `accepted` is a Workgroup Lead action I cannot take. The issue proposes **Developer Experience & Ecosystem**, and `/api/tools-db` and the dashboard backend sit in that Workgroup.

Could a DX & Ecosystem lead take a look at #3949 and `/accept` it if it belongs there? I will rebase and the check should clear on its own. Happy to close this and reopen after acceptance if you would rather the PR did not sit against an unaccepted issue.

Everything else is green, and the change is scoped to `dashboard/backend/router`.

dashboard/backend/router/core_routes.go
```diff
@@ -156,17 +156,55 @@ func registerToolRoutes(mux *http.ServeMux, cfg *config.Config) {
...
+//
+// An absolute configured path is returned as given: it is already unambiguous,
+// and rewriting it would break a deployment that points outside the checkout.
+// defaultToolsDBPath mirrors the router's own default
+// (config.defaultCanonicalIntegrationGlobal), relative to the project root.
+const defaultToolsDBPath = "config/tools_db.json"
+
 func resolveToolsDBPath(cfg *config.Config) string {
-	toolsDBPath := filepath.Join(cfg.ConfigDir, "config", "tools_db.json")
+	projectRoot := filepath.Dir(cfg.ConfigDir)
```
Xunzhuo, line 193: [P2] Keep the tools database rooted correctly in the local stack

`ConfigDir` can be an explicit dashboard state directory, not the config file's parent: `container_start.py` sets `DASHBOARD_CONFIG_DIR=/app`, and `resolveConfigPaths` preserves it. This makes `projectRoot` `/` in the normal `vllm-sr serve` stack. The dashboard image installs the default database at `/app/config/tools_db.json`, but its canonical `config/tools_db.json` setting now resolves to `/config/tools_db.json`, so `/api/tools-db` changes from 200 to 404. Derive the resource root using the deployment contract and add a regression for the `/app` override alongside the checkout-shaped cases.

Xunzhuo (changes requested):
The checkout and absolute-path cases pass, but the new root calculation breaks the normal local stack, where DASHBOARD_CONFIG_DIR is /app. The isolated resolver/HTTP check reproduces a 200-to-404 change for the shipped default tools database. Please preserve that deployment layout and add a regression. Exact-head dashboard CI passes; a full stack launch and Windows run were not repeated.

Xunzhuo:
Thanks for diagnosing the path failures and adding regression cases. Closing this as superseded by merged #3944, which incorporates those cases, resolves Tools DB paths against the explicit resource root, preserves absolute paths, and refreshes the path after config saves. The fix for #3949 is now on main.

PR #3951, opened 2026-09-19
[Research] Add same-run evaluation harness for router-native candidates (#3856)

### LINKED ISSUE #3856: [Research] Same-run evaluation harness for router-native model candidates

karthikeyan1592:
Closes #3856
Related #3198 #3857 #3194

## Purpose

Land the same-run helper so BERT vs any candidate can be replayed on one host with one JSON schema. No router-runtime change. No candidate choice. Quality numbers stay with the #3194 contract; this PR emits per-row `label` / `output` plus latency, peak RSS, and CPU seconds.

## Expectations

- `row_id` hashes **text only** (`sha256(text)[:16]`), stable under gold relabelling
- Run JSON includes host fingerprint (`cpu_model`, `core_count`, `ram_gb`, torch/candle version, `binding`)
- `same_run_pair.py` **exits non-zero** on cross-host mismatch and writes no output file
- No pooled accuracy in this PR; pairing reports latency delta and routing agreement only
- `--binding hf|candle` so HF transformers vs `ClassifyMmBert32KModality` gap is measurable

## Files

| File | Change |
|---|---|
| `same_run_harness.py` | New — single-stream classify loop, host fingerprint, JSON output |
| `same_run_pair.py` | New — join on `row_id`, refuse cross-host, latency delta |
| `test_same_run.py` | New — 6 unit tests, no model download |
| `README.md` | Harness usage section added |

## QSL

Frozen 759-row `test.jsonl` from #3891 (`random_state=42`). Not reshuffled. `row_id = sha256(text)[:16]`.

## Test plan

- [ ] `python test_same_run.py` → 6 tests OK
- [ ] Confirm pair script exits non-zero when `cpu_model`/`core_count`/`ram_gb` differ and writes no output file
- [ ] Run BERT then candidate on **one** host using the README commands; pair the two JSONs on `row_id`
- [ ] Optional: `--binding candle` on mmBERT-32K if FFI is on `PYTHONPATH`

## Test results

`python test_same_run.py` → **6 tests OK (0.035s, no model download)**

BERT numbers from the issue comment are a smoke-test that the helper runs. **Do not reuse those times.** Run BERT and the candidate on your machine, pair them — that is the comparison.


karthikeyan1592:
@adaamko @ramkrishs @Roy214 — addresses all four items from your review on #3856. Tagging you for review since the PAT scope doesn't allow formal reviewer assignment from a fork.

src/training/model_classifier/modality_routing_classifier/same_run_harness.py
```diff
@@ -0,0 +1,374 @@
...
+        classify(row["text"])
+
+    cpu_before = time.process_time()
+    wall_before = time.perf_counter()
+    records = []
+    scored = 0
+    while True:
+        for row in qsl:
+            result = classify(row["text"])
+            if scored < len(qsl):
```
Xunzhuo, line 359: **[P2] Retain latency samples from every measured pass**

Keep the latency samples from every measured pass. When the QSL finishes before --min-duration-s, this guard discards every later request even though scored_queries, CPU time and RSS include those passes. A four-request run with latencies [1, 1, 100, 100] ms reports scored_queries=4 but p99=1 ms because only the first two records reach summarize(). This makes the requested measurement duration ineffective for tail latency and compares different measurement windows. Retain all timing samples (separately from a deduplicated quality row if needed) and add a multi-pass regression.

src/training/model_classifier/modality_routing_classifier/same_run_harness.py
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+    return classify
+
+
+def load_candle_adapter(model_id: str, max_length: int):
+    """Production path: ClassifyMmBert32KModality. Requires candle-binding."""
+    del max_length
+    try:
+        import candle_binding  # type: ignore
```
Xunzhuo: **[P2] Implement the Candle adapter before advertising it**

Provide a real adapter for the repository's Candle binding before exposing this option. candle-binding currently builds a Rust C ABI plus Go wrappers; it does not provide an importable Python candle_binding module or the Python methods used here. Adding its build output to PYTHONPATH therefore still cannot run the documented --binding candle path. Bridge the existing C ABI or invoke a supported Go helper, propagate initialization errors and the context limit, and exercise that adapter with a small fixture.

Xunzhuo (changes requested):
The same-run comparison still has two functional gaps: repeated measurement passes are omitted from latency statistics, and the Candle option has no implemented Python bridge to the repository binding. Please fix both and cover them with regressions. The six new unit tests pass; the training CI ran 649 existing contract cases. Model-backed HF/Candle inference was not run.

karthikeyan1592:
Thanks for the detailed review @Xunzhuo — both P2 items are addressed in commits bf048ed and 8d8cdb8.

**P2-1: Multi-pass latency**

Separated `quality_records` (first pass only, for label/output pairing) from `latency_samples` (every measured pass, for p50/p99). The original `scored < len(qsl)` guard was silently discarding all timing beyond the first pass while still counting those requests in `scored_queries`, CPU time, and RSS. Now every measured request contributes a timing sample regardless of pass.

Regression: `test_latency_samples_span_all_passes` — a 2-row QSL with a fast first pass (1 ms) and a slow second pass (100 ms). Before the fix p99 was ≤ 1 ms; after it is > 50 ms.

**P2-2: Candle adapter**

Replaced the phantom `import candle_binding` with a persistent Go subprocess using an NDJSON line protocol (one JSON request per line in, one JSON response per line out). The helper binary is resolved via `CANDLE_CLASSIFY_HELPER` env var or `shutil.which("candle-classify")`, with a clear build hint on failure. Stderr is propagated on pipe break or helper error. Cleanup is registered with `atexit` and `classify.close`.

Regressions: `test_missing_helper_raises_system_exit`, `test_helper_protocol_with_mock_binary` (real Python script as stand-in for the Go binary), `test_helper_error_response_raises`.

**End-to-end run — `--binding hf`, CPU, `llm-semantic-router/mmbert32k-modality-router-merged`**

Host: Intel i5-12450HX, 4 cores, 7.8 GB RAM, WSL2, Python 3.12.3, torch 2.14.0

| | e2e | forward | tokenize |
|---|---|---|---|
| p50 | 48.18 ms | 47.92 ms | 0.24 ms |
| p99 | 61.35 ms | 61.02 ms | 0.43 ms |
| mean | 48.94 ms | 48.66 ms | 0.26 ms |

1240 inferences over 60 s (31 passes × 40-row QSL). `n_records=40` (quality, first pass), `n_latency_samples=1240` (all passes). Peak RSS 1739 MB, CPU 121.6 s.

All 11 unit tests pass.

src/training/model_classifier/modality_routing_classifier/same_run_harness.py
```diff
@@ -0,0 +1,453 @@
...
+        raise SystemExit(
+            "--binding candle requires a compiled Go helper binary.\n"
+            "Build:  cd candle-binding && go build -o candle-classify ./cmd/classify-helper/\n"
+            "Export: CANDLE_CLASSIFY_HELPER=/path/to/candle-classify"
+        )
+
+    import atexit
+
+    try:
+        proc = subprocess.Popen(
```
Xunzhuo, line 222: **[P2] Include the helper process in resource measurements**

Inference now runs in this subprocess, but `run_single_stream` still measures only the parent with `time.process_time()` and `getrusage(RUSAGE_SELF)`. A protocol-compatible helper that uses 0.5 CPU seconds and allocates 100 MiB produces `cpu_s: 0.0` and `peak_rss_mb: 22.6` in the run report. These values feed the paired CPU/RSS deltas, so the Candle path appears artificially cheap against in-process HF inference. Collect resource usage from the persistent helper, or explicitly report these metrics as unavailable until they include the inference process.

Xunzhuo (changes requested):
The repeated-pass latency fix passes all 11 unit tests. The Candle helper is still absent from the documented build path, so the earlier adapter finding remains open. Moving inference into a child process also makes the reported CPU and peak RSS exclude the model process; the isolated subprocess probe reproduces that new accounting error. Please include the real helper and measure its resources. Full HF/Candle model inference was not run.

karthikeyan1592:
@Xunzhuo — both findings from this round are fixed in 217fcbe.

**The helper now exists at the documented path**

`candle-binding/cmd/classify-helper/main.go` is a real Go binary implementing the NDJSON protocol against `ClassifyMmBert32KModality`. Verified it launches and classifies correctly, standalone and driven by the harness:

```
classify-helper: initializing <model-path>
🎯 Initializing mmBERT-32K modality routing classifier: <model-path>
   mmBERT-32K modality router loaded (AR/DIFFUSION/BOTH, 32K context)
   mmBERT-32K modality router initialized (AR/DIFFUSION/BOTH)
classify-helper: ready
{"label":"AR", ...}
```

The two log lines in the middle are from the production init path in `semantic-router.go` itself, not anything new — confirming this is the real candle/Rust model loading, not a stub.

Also fixed: `--model` previously said "HF model ID or local path," which was wrong — candle only accepts a local directory, it doesn't auto-download. Passing a Hub ID now fails fast with the exact `huggingface-cli download` command to fix it, instead of a bare Rust FFI "file not found" error.

**Helper resource usage is now attributed**

The helper self-reports its own `getrusage(RUSAGE_SELF)` (CPU seconds, lifetime-peak RSS) on every response. `run_single_stream` folds the helper's measured-window CPU delta and peak RSS into `cpu_s`/`peak_rss_mb`, with a new `includes_helper_resources` flag so it's visible in the report whether a run's numbers include a child process or not.

Regression test `test_run_single_stream_attributes_helper_resources` reproduces your probe directly: a fake helper that deliberately burns ~0.05s CPU per request and reports 137 MB RSS. Before this fix `cpu_s`/`peak_rss_mb` would have shown ~0. Now they reflect it.

**Real end-to-end run — `--binding candle`, production model, CPU**

```json
{
  "scored_queries": 440,
  "cpu_s": 98.249,
  "wall_s": 60.097,
  "peak_rss_mb": 2492.9,
  "includes_helper_resources": true,
  "e2e": {"p50_ms": 127.394, "p99_ms": 204.162, "mean_ms": 136.513}
}
```

440 real inferences through the actual production candle path (confirmed via process tree — parent Python + child `candle-classify` at 128% CPU, 1.5 GB RSS mid-run) against `llm-semantic-router/mmbert32k-modality-router-merged`. 36/40 predictions matched gold on a held-out synthetic QSL, confirming this is genuine inference, not a stub returning fixed labels.

12/12 unit tests pass, black/codespell/gofmt/go vet all clean.

FAUST-BENCHOU:
The follow-up addresses the two important measurement gaps well: latency now covers every measured pass while quality rows remain deduplicated, and the Candle path is backed by an actual persistent helper instead of a nonexistent Python binding. The failure/cleanup handling around the helper also looks appropriately explicit. LGTM from my pass.

Issue #3392, opened 2026-09-03
[Feature] Define sticky tool-set state, storage, and trusted session identity


Zheng-Lu:
## Goal

Define the bounded state, storage, fingerprint, and trusted-session identity foundation required by #3347 without changing request-time tool selection yet.

## Proposed Workgroup

Agentic & Context

## Work item shape

This is a leaf implementation issue under #3347. It owns one independently reviewable foundation, not the complete sticky-selection capability. PR #3391 is the active implementation attempt; this issue is complete only after reviewed code is merged into `main`.

## Current-main baseline

Current `main` has request-time tool selection and session-related request context, but no canonical session tool-state schema, bounded store, tool-definition fingerprints, or trusted identity resolver dedicated to sticky selection.

## Scope and implementation approach

- Define content-minimized state containing tool identities, canonical fingerprints, and bounded lifecycle metadata; never persist raw schemas, prompts, arguments, or results.
- Implement deterministic definition, catalog, policy, and capability fingerprints.
- Provide a concurrency-safe bounded local store with TTL and per-identity quotas, plus a narrow storage interface for later shared implementations.
- Resolve only configured authenticated session provenance into opaque tenant-isolated keys. HMAC can isolate identifiers but does not make an untrusted identifier authoritative.
- Register the state in the canonical state inventory and keep default request behavior unchanged.
- Any additive configuration must remain disabled by default. It must not advertise usable sticky selection before the runtime reauthorization and invalidation path exists; unsupported enablement must fail validation clearly.

## Dependencies

- Parent delivery issue: #3347.
- Reuse existing configuration, session, tool-selection, and state-taxonomy contracts.
- This leaf is a prerequisite for the runtime reuse and invalidation work remaining under #3347.

## Configuration and API impact

No client-facing HTTP or inference protocol changes. This leaf may add bounded internal or disabled configuration, but it does not enable request-time tool reuse or require a production Redis backend.

## Non-goals

- Intercept or rewrite live request tool lists.
- Implement sticky merge, pinning, growth, reauthorization, or invalidation.
- Reuse authorization decisions.
- Implement distributed-store consistency or Redis CAS behavior.
- Store raw tool definitions or request content.
- Claim delivery before the implementation is merged.

## Completion signal

- [ ] State and fingerprint contracts are deterministic and documented.
- [ ] Bounds and disabled behavior validate across maintained producers.
- [ ] The local store passes concurrency, TTL, quota, eviction, and race tests.
[...]

Issue #2973, opened 2026-08-22
[Epic] Govern long-session state, continuity, and model switching


Xunzhuo:
## Parent Direction

Agentic & Context — #2987

## Goal

Govern bounded cross-request session state, conversation continuity, workflow resume, per-session inference budgets, and safe logical-model switching for long-running agentic workloads.

## Background

Current main already derives `AgenticSessionContext`, supports a session-aware selector, and records bounded local or shared session state for turns, models, usage, cache behavior, and switches. Remaining work should harden these supported surfaces. The legacy `model_switch_gate` configuration is not a public contract and must not be revived.

## Scope

- Define trusted session and conversation identity, state boundaries, TTL, concurrency, restart, reload, and cross-replica limitations.
- Preserve required provider state, tool-loop state, context, usage, provenance, and accounting across supported continuation paths.
- Keep model switches inside the configured logical-model candidate set and enforce portability, capability, safety, residency, and budget constraints.
- Add bounded recent-window evidence, cooldown, hysteresis, and deterministic de-escalation so one noisy turn cannot cause thrashing.
- Define fallback, shadow, canary, rollback, and workflow-resume behavior.
- Keep workflow execution at explicit Looper, gateway, or external agent-runtime boundaries.

## Current work

- #2471 — lifecycle-own Looper workflow state and make pause/resume semantics real.
- #2018 — reserve and reconcile bounded per-session inference tokens.
- #2999 — define Conversations API state and history semantics.
  - #2814 — remove Redis full-keyspace scans from conversation response lookup.
- #3347 — add bounded session-scoped sticky tool-set selection with re-authorization and safe invalidation.
- #3377 — gate model switches on calibrated recent-window progress evidence.

## Non-goals and boundaries

- Context transformation belongs to #2984; prompt-visible Router Memory belongs to #2339.
- Agent endpoint invocation and composition remain outside the Router; #2994 owns that boundary contract.
- MoM & Routing owns model pools and model-level selection algorithms.
- Per-session inference budgets are not API-key quota, billing, or token-rate-limit products.
- The Router does not become an unrestricted workflow engine or cross-request authorization state machine.

## Completion criteria

- [ ] Session, continuation, resume, and switch-state contracts are explicit and bounded.
- [ ] Supported continuations preserve required context, tool state, provider state, and accounting.
- [ ] Single-process and shared-store consistency limits are documented and tested.
- [ ] Static and session-aware baselines report quality, stability, latency, cost, and switch evidence.
[...]

PR #3956, opened 2026-09-20
[Feature] Add trusted-session sticky tool selection

Lcollection:
Related #2973

## Purpose

Add trusted-session sticky tool selection with bounded, deterministic state while keeping the existing stateless behavior as the default.

This change:

- reuses request-time selection and reauthorizes every tool on every request;
- keeps selected tools stable for a trusted session, pins previously called tools when policy allows, and applies bounded growth/replacement rules;
- invalidates stale state when catalog, schema, capability, policy, or retrieval fingerprints change;
- provides bounded in-memory and Redis-backed session state with TTL, quotas, revision/CAS protection, expiry, restart, and unavailable-store recovery;
- preserves deterministic ordering and provider-visible definition bytes for retained tools, including provider-prefix E2E coverage;
- emits content-minimized reuse, growth, invalidation, and fallback receipts and documents the configuration contract.

Sticky selection remains disabled by default. This PR does not cache authorization decisions, execute tools, persist raw schemas/arguments/prompts/results, or guarantee a provider prompt-cache hit.

The implementation uses the trusted-identity and bounded-state seam associated with #3392. #3392 remains a separate foundation dependency and is not represented as merged by this PR.

## Test Plan

Unit and package checks:

~~~text
go test ./pkg/sessiontools ./pkg/tools ./pkg/extproc ./pkg/protocolcodec ./pkg/config
go test -race ./pkg/sessiontools ./pkg/extproc
go test ./profiles/anthropic-shim ./profiles/all ./testcases
make build-e2e
git diff --check
~~~

Maintained live E2E checks cover sticky reuse/recovery, TTL expiry, Redis restart/unavailable recovery, invalidation, concurrent turns, and provider-prefix stability. The latest live run used the `anthropic-shim` profile and compared sticky behavior with a stateless baseline.

## Test Result

- All commands above passed locally.
- Live provider-prefix E2E passed (`sticky-tool-selection-provider-prefix`, 5.117s), including trusted first-turn selection, retained-prefix byte stability, ordered growth, provider cache read/write accounting, and stateless-baseline comparison.
- Kind/Redis temporary resources were removed after the run.
- Python `pytest` is not installed in the current environment, so the Python test suite was not run here; repository CI should provide that gate.
- Full repository-wide PR gates have not been run locally; please rely on CI for the remaining required checks.

## Review focus

Please review the session-state lifecycle and invalidation boundaries, per-request authorization behavior, deterministic merge/order guarantees, local/shared-store consistency, recovery behavior, and provider-prefix serialization tests.

All commits in this branch include `Signed-off-by:` lines.


FAUST-BENCHOU:
Nice separation between sticky selection and authorization. Retained tools are treated as session hints rather than cached authorization decisions, while the store remains bounded by TTL, quotas, revision checks, and state size. The recovery and provider-prefix coverage also exercise the important stateful boundaries. LGTM from my pass.

dashboard/frontend/src/lib/dslPluginSchemas.ts
```diff
@@ -0,0 +1,391 @@
+import { resolveCapabilityPluginFieldSchema } from './dslCapabilityPluginSchemas'
```
Xunzhuo: [P1] Use the exported capability-schema helper

`dslCapabilityPluginSchemas.ts` exports `getCapabilityPluginFieldSchema`, but this new module imports and calls `resolveCapabilityPluginFieldSchema`. The dashboard type check fails with TS2724, so the normal `tsc && vite build` cannot produce a bundle. Use the existing export, or update both modules consistently, and add the wrapper to the dashboard type-check coverage.

src/semantic-router/pkg/config/tool_selection_plugin_validation.go
```diff
@@ -54,6 +73,44 @@ func (c *ToolSelectionPluginConfig) validateModeConstraints(mode string) error {
...
+	if c.Sticky.MaxTools != nil {
+		if v := *c.Sticky.MaxTools; v <= 0 || v > StickyToolSelectionMaxToolsUpperBound {
+			return fmt.Errorf(
+				"tool_selection plugin: sticky.max_tools must be between 1 and %d",
+				StickyToolSelectionMaxToolsUpperBound,
+			)
+		}
+	}
+	if c.Sticky.MaxNewToolsPerTurn == nil {
+		return nil
```
Xunzhuo: [P2] Validate the effective growth limit when it is omitted

With sticky selection enabled and `max_tools: 1`, omitting `max_new_tools_per_turn` passes both the Go and CLI validators here, but its effective value is 2. `sessiontools.Manager.Select` rejects growth greater than `max_tools`, returning `Fallback: true, Reason: invalid_input` on every turn, so this accepted configuration never stores or reuses sticky state. Make the default compatible with the effective maximum, or reject the combination in both validators, and cover the omitted-growth boundary.

Xunzhuo (changes requested):
The dashboard type check and a config-to-manager regression reproduce two blockers: a missing export and a single-tool configuration that silently bypasses sticky state. Please fix both and rerun the affected checks after updating the branch.

The repository harness, focused Go/CLI tests, E2E unit tests, and 82 session-tool race tests (including 16 real Redis integration tests) passed. Native ExtProc validation is still pending an exact-head native build; deployed Envoy/Kind scenarios were not run.

Xunzhuo (commented):
Both previously reported defects are fixed: the Dashboard uses the exported capability-schema helper, and Go/Python validate the effective default growth against max_tools. Verified the TypeScript import graph, 15 focused Go tests, 86 CLI tests, and omitted/explicit growth boundaries. This completes the three-file follow-up review; deployed sticky-session behavior after integration with current main remains unverified.

Lcollection:
Thanks. I noticed that the GitHub Action didn't complete the deployment; do I need to test this local and submit a report?

AayushSaini101:
> Thanks. I noticed that the GitHub Action didn't complete the deployment; do I need to test this local and submit a report?

thanks for working @Lcollection there are some conflicts in your MR, can you please check than action will automatically run again after pushing the changes thanks : )

Lcollection:
@Xunzhuo @AayushSaini101

Latest head `871a122a7d527fc1bf13a0a9340d6c0d5047f11b` is synchronized with current `main` (`behind_by=0`) after the branch update. The complete PR validation is green, including CI Gate, PR Gate, Router, Dashboard, CPU/ORT/OpenVINO, QEMU, Kubernetes E2E, performance, and Router Image checks.

The remaining merge gate is review: the earlier `CHANGES_REQUESTED` has not been replaced by a formal approval. Per repository policy, please have at least two maintainers with write/maintain/admin permission submit `APPROVE` reviews. The PR author cannot re-request reviewers through the GitHub API due repository permissions.

PR #3959, opened 2026-09-20
[Feature] Rank decisions on declared evidence instead of every leaf

subin9:
Related #3081

## Purpose

DR-02 and DR-06 from the ranking proposal. Selection blended two different things into one number. A keyword rule, a `NOT` guard, a predicate, a conversation rule and a projection output all restate policy the operator already wrote, yet each contributed to the confidence that decided which decision won. A projection in particular restates a decision the operator made, so averaging it into the evidence diluted the real measurement.

Every leaf is now either policy or evidence. Policy leaves decide whether a decision matches and carry no weight in its confidence. Evidence is a signal that reports a measurement, and each type declares its kind: a classifier probability for `domain`, `classifier`, `complexity`, `jailbreak`, `safety`, `preference` and `kb`, a vector similarity for `embedding` and `reask`.

Two rules follow. A decision is comparable only when exactly one evidence leaf produced its score, because a mean over several leaves falls as a decision gains evidence: `mean(.90, .90)` beats `mean(.90, .90, .88)`. A pool ranks by confidence only when every member that is not a catch-all is comparable and all of those scores are the same kind, since a probability and a similarity are not the same quantity. Everything else ranks by priority, which is what the operator declared.

DR-06 is the validation half. The score kind table lives in `pkg/config` so the validator and the runtime read the same contract, and configuration load now warns for each pool whose members report different kinds, naming the profile, the tier and the decisions. That makes the fallback explicit instead of leaving an operator to wonder why `strategy: confidence` changed nothing.

## Compatibility

Every pool with more than one member in the six shipped configs ranks by priority under this contract.

| Pool | Members | Why it ranks by priority |
| --- | --- | --- |
| `config.yaml` tier 0 | 24 | members report both probabilities and similarities |
| `agent` tier 1 | 3 | every member rests on projections, a conversation rule or a guard |
| `agent` tier 3 | 5 | each member carries three or more evidence leaves of both kinds |
| `privacy` tier 2 | 2 | both members rest on policy leaves |
| `multi-objective` / `privacy-first` tier 2 | 2 | both members rest on policy leaves |

The last pool is the one observable change. It ranked by confidence before, where `omni`'s conversation predicate reported a structural 1.0 and outranked the projection score of `unified_privacy_sensitive_route` regardless of the priorities the operator wrote. It now ranks by priority.

This leaves confidence ordering for pools whose members each report one score of one kind, and no shipped pool is shaped that way today. If the project would rather keep multi-leaf decisions comparable, the alternative is an aggregate that does not move with leaf count, such as the minimum over same-kind evidence. I did not assume that, because the proposal asked for the stricter reading first.

## Test Plan

`go test ./src/semantic-router/pkg/decision/ ./src/semantic-router/pkg/config/`

## Test Result

Three existing tests encoded the old aggregate and had to change, which is the record of what this contract moves.

`TestTieredSelectionUnscoredPoolFallsBackToPriority` from #3080 keeps its winner, but its decision is no longer comparable, because it aggregates a domain and a complexity leaf. `TestTieredSelectionFullyScoredPoolStillRanksByConfidence` from the same PR gave a keyword rule a reported score to make a pool comparable, which the contract no longer allows. It is replaced by a pool of single evidence leaves of one kind, plus a new case where a probability and a similarity meet in one pool and priority decides. The `and/all-match` row of the contract matrix moves from 0.7 to 0.6, the embedding score alone, because the keyword leaf beside it is policy.

New cases cover the validator: mixed kinds in one tier, an untiered pool, a catch-all that does not count as a member, and policy leaves declaring no kind.

Not covered here: no recipe probe or end to end case asserts the new ordering at the routing boundary, and traces still do not report which mode ranked the pool. Those are DR-08 and DR-07.


src/semantic-router/pkg/config/signal_score_kind.go
```diff
@@ -0,0 +1,40 @@
...
+// model: a category probability, a classifier score, or a similarity. Every
+// other type is policy and is absent from this table.
+var signalScoreKinds = map[string]ScoreKind{
+	SignalTypeDomain:     ScoreKindProbability,
+	SignalTypeClassifier: ScoreKindProbability,
+	SignalTypeComplexity: ScoreKindProbability,
+	SignalTypeJailbreak:  ScoreKindProbability,
+	SignalTypeSafety:     ScoreKindProbability,
+	SignalTypePreference: ScoreKindProbability,
+	SignalTypeKB:         ScoreKindProbability,
```
Xunzhuo, line 31: **[P2] Classify the actual score quantity**

KB confidence is a similarity: `kbLabelMatchConfidence` returns `BestSimilarity`, while local complexity and contrastive jailbreak can report embedding margins rather than probabilities. This table tags them as probabilities, so a lower-priority KB match at 0.95 beats a domain probability of 0.80 instead of falling back to priority. The isolated decision regression reproduces that result. Carry the score kind from the producer/backend, or conservatively mark these variants incomparable; the signal type alone does not establish a probability.

src/semantic-router/pkg/decision/engine.go
```diff
@@ -255,6 +263,7 @@ func (e *DecisionEngine) evaluateDecisions(
 				MatchedRules:     resolved.evaluation.matchedRules,
 				ConfidenceScored: resolved.evaluation.scored,
 				CatchAll:         isCatchAllRules(decision.Rules),
+				ScoreKind:        resolved.evaluation.kind,
```
Xunzhuo, line 266: **[P2] Preserve mixed score kinds through OR evaluation**

The exported kind is only the child chosen by `evalOR`, whose `preferredMatch` still compares raw numbers without checking their kinds. With both branches matching, `OR(domain=0.80, embedding=0.95)` becomes a comparable similarity and beats a higher-priority embedding decision at 0.90. The mixed kinds have disappeared before `comparableConfidencePools` sees them. Apply the kind contract while aggregating OR matches so an incomparable numeric comparison cannot manufacture a single-kind winner.

src/semantic-router/pkg/decision/engine.go
```diff
@@ -671,19 +685,62 @@ func (e *DecisionEngine) evalAND(
...
+type andEvidence struct {
+	count      int
+	value      float64
+	kind       config.ScoreKind
+	mixedKinds bool
+	unreported bool
+}
+
+func (a *andEvidence) add(child nodeEvaluation) {
+	if child.evidence == 0 {
```
Xunzhuo, line 715: **[P2] Keep error-policy matches unscored in conjunctions**

A failed classifier with `on_error: match` returns `onError=true`, `scored=false`, and zero evidence. This early return drops that information; an AND containing it and one domain score is then marked scored by `apply`. In the isolated regression, a priority-10 decision with domain 0.95 and a failed guard beats the priority-100 decision at 0.90. Previously the error match kept this pool on priority, and the documented selected-confidence contract still excludes error-policy results. Track error-policy comparability separately from ordinary policy gates and cover this conjunction.

Xunzhuo (changes requested):
Three deterministic decision regressions still violate the new comparability contract: native similarity/margin scores are classified as probabilities, mixed-kind OR branches are compared before their kinds are checked, and an error-policy AND match becomes confidence-scored. Please fix these paths and add regressions for the route winners. The isolated decision probes reproduce all three; native model inference and a deployed routing stack were not run.

PR #3964, opened 2026-09-20
[Test] Verify the Snowflake Cortex failure envelope and raise its conformance

Bevisy:
# [Test] Verify the Snowflake Cortex failure envelope and raise its conformance

## What

Records the Snowflake Cortex AI transport failure envelope as an observed provider fixture, decodes it under a Snowflake response vendor, and raises the provider's conformance from `unverified` to `fixture_verified`.

Related: #3608

## Why the failure wire is what we have

Trial accounts are not allowed to reach the Cortex REST API: `POST {account}/api/v2/cortex/v1/chat/completions` answers `403` with the provider's own envelope

```json
{"code":"003001","message":"This account is not allowed to access this endpoint. Please contact Snowflake support.","request_id":"…","error_code":"003001"}
```

for a Tokyo-native model name, a nonexistent model name, a Cross-Region opt-in model name and an empty model name. The same account and the same PAT answer `200` on the Snowflake SQL API (`/api/v2/statements/`), so account, creds, hostname and network path are correct and the refusal is the account type.

## Changes

- `pkg/llmprotocol`: `ResponseVendorSnowflake` documents the flat failure envelope.
- `pkg/protocolcodec/snowflake_transport_error.go`: decodes that envelope at the transport edge, so the provider code, message and request ID reach the client instead of a malformed canonical decode. `003001` maps to the permission category; unobserved codes keep the neutral unavailability category.
- `pkg/protocolcodec/snowflake_cortex_conformance_test.go` + fixture: strict decoding rejects the envelope, and the Snowflake vendor preserves code, category, message and request ID. The fixture is the observed response with the request ID and the account identifier redacted, and its size is pinned so an edited fixture cannot pass as conformance.
- `pkg/protocolcodec/codec_anthropic_response.go`: the same envelope decodes on the Anthropic Messages path, which the provider declares along with chat completions.
- `pkg/extproc`: profiles declaring the `snowflake-cortex` provider select that vendor, with a regression covering both the provider profile and a Snowflake host without the provider id. The provider id, not the host, is the selector, because the account-scoped host also fronts the Snowflake SQL API, whose bodies do not share this envelope.
- Catalog resource and regenerated projections: `fixture_verified`, `verified_at: 2026-09-20`.

## Verified

- `go test ./pkg/protocolcodec/ ./pkg/catalog/ ./pkg/llmprotocol/` pass, including the new failure-envelope test.
- `go vet ./pkg/extproc/` and `go build ./pkg/extproc/` pass.
- `go test ./pkg/extproc/` passes, including the vendor resolution regression.
- `make model-catalog-test` (70 tests) and `make model-catalog-generated-check` pass.

## Not verified

The success envelope of both declared protocols has no observed fixture, so the provider does not claim `live_verified`. The remaining blocker is Snowflake-side.


src/semantic-router/pkg/protocolcodec/snowflake_transport_error.go
```diff
@@ -0,0 +1,56 @@
...
+// without a flat code/message pair falls back to the canonical OpenAI decoder,
+// so an unexpected shape still reports the same typed error instead of a silent
+// success.
+func decodeSnowflakeTransportError(
+	body []byte,
+	policy llmprotocol.Policy,
+	format llmprotocol.WireFormat,
+) (llmprotocol.TransportError, llmprotocol.Diagnostics, error) {
+	var wire snowflakeTransportErrorWire
+	if err := json.Unmarshal(bytes.TrimSpace(body), &wire); err != nil ||
```
Xunzhuo: [P2] Preserve the shared provider JSON validation

This successful return bypasses the provider JSON validation used by the other transport decoders. With the Snowflake vendor selected, `Engine.DecodeTransportError` accepts a body exceeding `Limits.BodyBytes`, duplicate `code` keys (silently taking the last value), and an unpaired Unicode surrogate. The later `ValidateTransportError` only checks the decoded fields, so it cannot enforce the original wire limits or detect malformed input. Decode the flat envelope through the shared provider validation and add negative boundary regressions before returning it.

Xunzhuo (changes requested):
The Snowflake envelope is decoded correctly, but the new path bypasses the shared provider JSON validation. Please preserve those checks before accepting the vendor envelope. The protocolcodec, catalog, and llmprotocol suites pass; an engine-level probe reproduces the validation bypass. Live Snowflake success responses were not tested.

src/semantic-router/pkg/protocolcodec/snowflake_transport_error.go
Bevisy: > [P2] Preserve the shared provider JSON validation
>
> This successful return bypasses the provider JSON validation used by the other transport decoders. With the Snowflake vendor selected, `Engine.DecodeTransportError` accepts a body exceeding `Limits.BodyBytes`, duplicate `code` keys (silently taking the last value), and an unpaired Unicode surrogate. The later `ValidateTransportError` only checks the decoded fields, so it cannot enforce the original wire limits or detect malformed input. Decode the flat envelope through the shared provider validation and add negative boundary regressions before returning it.

Confirmed at the engine level and fixed: the decoder now goes through `decodeProviderWireVendorAware`, the same shared provider JSON validation the Cloudflare and OpenAI transport decoders already use, and the fallback to the canonical decoder stays for unexpected shapes. Duplicate keys were the sharpest of the three — the last value won, and a repeated `code` reclassified `003001` from a permission refusal to upsteam unavailability.

Two negative boundary regressions pin the contract in `snowflake_cortex_conformance_test.go`: a body exactly at the policy limit decodes and one byte over is refused with `upstream_body_limit`, and a duplicated `code` key is refused with `upstream_duplicate_json_field`. Both are red on the previous decoder.

Bevisy: > [P2] Preserve the shared provider JSON validation

Right on both paths now, not only the chat one: the envelope also goes through the shared provider validation on the Anthropic Messages path the provider declares, so neither path can take an over-limit body, a duplicated field or a malformed string — the regressions above pin that contract.

The envelope's `request_id` reaches the transport error too, alongside its code and message.

FAUST-BENCHOU (approved):
The follow-up closes the validation gap cleanly: the Snowflake envelope now stays behind the shared wire-validation boundary on both declared protocol paths, with body-limit and malformed-field regressions covering the escape hatch. Keeping this at `fixture_verified` until an accepted response is observed also matches the evidence actually available. LGTM from my pass.

Xunzhuo (approved):
The Snowflake error decoder now preserves shared JSON validation, and the duplicate-field and size-limit regressions pass. The provider/vendor fixtures and native transport-error coverage pass, with current-head Streaming, Anthropic compatibility, and Responses API CI also passing. This approves the declared fixture-verified support; successful live Snowflake requests remain unverified.

wilsonwu (approved):
LGTM

Issue #3962, opened 2026-09-20
[Feature] Expose the model-download endpoint in the k8s values and operator guide


Bevisy:
## Goal

A k8s deployment on a network that cannot reach the public model hub directly
must be able to point the router's model download at an approved mirror through
the deployment values, and the operator and installation pages must show that
shape.

## Proposed Workgroup

Developer Experience & Ecosystem
([#2970](https://github.com/vllm-project/semantic-router/issues/2970)) — chart
values plus the k8s installation and operator pages: configuration, deployment,
diagnostics, and documentation surfaces.

## Current-main baseline

The runtime already honors an endpoint override, but only as a pod environment
variable, and the k8s surface does not carry it:

- `src/semantic-router/pkg/modeldownload/config_parser.go:238` — `GetDownloadConfig`
  reads `HF_ENDPOINT` with the default `https://huggingface.co`; the value is
  handed to the download CLI child process at
  `src/semantic-router/pkg/modeldownload/downloader.go:103-105` and reported at
  `src/semantic-router/pkg/modeldownload/ensure.go:86`.
- The local CLI already forwards it — `src/vllm-sr/cli/commands/runtime_support.py:69-84`
  (pass-through list; `append_passthrough_env_vars`, called from
  `src/vllm-sr/cli/commands/runtime.py:208`) — and that path is documented at
  `website/docs/troubleshooting/network-tips.md:56-66`.
- The chart's default env carries no endpoint: `deploy/helm/semantic-router/values.yaml:506-531`
  declares `LD_LIBRARY_PATH`, `HOME`, `TMPDIR`, `HF_HOME`, `HF_TOKEN`, and
  `HUGGINGFACE_HUB_TOKEN`. `extraEnv` (`values.yaml:534`) appends after `env`
  (`deploy/helm/semantic-router/templates/deployment.yaml:96-104`), so a
  deployment can add one by hand, but neither the values nor the k8s pages name
  the endpoint: `website/docs/installation/k8s/operator.md:297-311` documents
  `spec.env` with a token-only example.

Observed in a deployment without a direct route to the hub: the model step fails
at startup (`startup_failed: ensure models are downloaded: failed to download 1
out of 1 models`) and the release stays `not ready`. With the endpoint added
through the values' `extraEnv`, the same profile completes end-to-end.

## Scope

1. Declare the model-download endpoint in the chart's default env
   (`deploy/helm/semantic-router/values.yaml`), defaulting to the canonical
[...]

PR #3969, opened 2026-09-20
[Feature] Expose the model-download endpoint in the k8s values and operator guide

Bevisy:
Related #3962

## Purpose

A cluster without a direct route to the model hub cannot point model download
at an approved mirror through the deployment values. The runtime already honors
`HF_ENDPOINT` — read at
`src/semantic-router/pkg/modeldownload/config_parser.go:238`, handed to the
download CLI child process at
`src/semantic-router/pkg/modeldownload/downloader.go:103-105`, reported at
`src/semantic-router/pkg/modeldownload/ensure.go:86` — but the chart's default
env never carries the name (`deploy/helm/semantic-router/values.yaml:506-531`),
and the k8s operator page documents `spec.env` with a token-only example
(`website/docs/installation/k8s/operator.md:297-311`).

- `deploy/helm/semantic-router/values.yaml` — add `HF_ENDPOINT` to the default
  env, defaulting to the canonical endpoint, so the name is visible in the
  values and overridable per deployment.
- `deploy/helm/semantic-router/README.md` — resync the `env[...]` key rows with
  the values block; it was staler than the chart (`HOME`, `TMPDIR`, `HF_HOME`
  were missing, so index-based rows no longer named the right entries).
- `website/docs/installation/k8s/operator.md` — add a section for deployments
  without direct hub access, showing both the `spec.env` and `extraEnv` forms,
  plus the startup failure a missing endpoint produces.

The entry reaches the pod env through
`deploy/helm/semantic-router/templates/deployment.yaml:96-104`, which ranges
`env` and appends `extraEnv`.

Other deployment targets (`deploy/openshift/`, `deploy/kserve/`) likewise carry
no model-download endpoint; they are outside this fix, which scopes the chart.

Draft: #3962 is still `neads-acceptance`. This PR demonstrates the bounded
change ahead of acceptance and is meant to be rebased and undrafted once a
maintainer applies `/accept` and one Workgroup owner is confirmed.

## Test Plan

- `make helm-lint` on this chart.
- `make helm-template` and compare the rendered pod env with the values' `env`
  block.
- `make helm-ci-validate HELM_NAMESPACE=test-namespace`.
- Deploy on a cluster without direct hub access with the mirror endpoint set
  through `extraEnv` and confirm the model step passes startup.

## Test Result

Not run author-side: this host has no `helm` binary (`make helm-lint` reports
`helm: No such file or directory`), so the chart gates are left to CI. The
yaml and markdown edisted cleanly (`patch` linter: ok), and the render path is
traced above from the template.


FAUST-BENCHOU:
Could we avoid making the documented `extraEnv` override produce two `HF_ENDPOINT` entries in the rendered container env? Relying on the later duplicate to win works at runtime, but leaves an ambiguous Pod spec. A dedicated value or explicit deduplication would make the override contract cleaner.

PR #3974, opened 2026-09-20
[Feature] Apply routing.strategy inside the selected tier

subin9:
Related #3081

## Purpose

DR-04 and DR-05 from the ranking proposal, and the last slice that changes how decisions are ordered.

`routing.strategy` meant two different things. Where no matched decision carried a tier it chose between priority ordering and confidence ordering, as documented. Where one did, selection ranked by confidence ahead of priority and never read the setting at all, so a profile that declared `strategy: priority` ordered its decisions by evidence whenever a tiered decision happened to match. The `priority` branch was also the only one without a catch-all rule, so an unconditional fallback with a high priority could outrank a real match there.

One comparator now orders every case: tier ascending while any matched decision carries a tier, then catch-alls after real matches, then the configured strategy, then the remaining key, then the decision name. `strategy: priority` compares priority first and consults confidence only to break a tie, `strategy: confidence` does the reverse, and both behave the same way inside a tier as without one. That removes the three separate orderings the engine carried.

## Compatibility

No shipped configuration changes its winner.

Every catch-all in the six shipped configs already sits alone in the highest tier number with the lowest priority, so tier ordering placed it last before this change: `agent` tier 6, `balance` tier 15, and the four `multi-objective` profiles at tiers 2, 2, 3 and 6.

Confidence ordering inside a tier only differs from priority ordering in a pool that ranks by confidence at all, and no shipped pool does after the evidence and score-kind contracts. A profile that declares `strategy: confidence` and competes decisions on one evidence leaf of one kind is where the setting now takes effect, which is what the setting always claimed to do.

## Test Plan

`go test ./src/semantic-router/pkg/decision/ ./src/semantic-router/pkg/config/`

## Test Result

Two regressions carry the change. One matches two untiered decisions and an unrelated tiered decision, and asserts that the pool keeps the ordering the profile configured under each strategy, where it previously flipped to confidence as soon as the tiered decision matched. The other asserts a catch-all ranks after a real match under `strategy: priority` as well as under `strategy: confidence`.

Three existing tests assumed confidence ordering inside a tier without configuring it: two tier tests and the OR evidence case from the previous slice. Each now declares `strategy: confidence`, which is what they were exercising.

Not covered here: no probe or end to end case asserts this ordering at the routing boundary, and traces still do not report which key decided, which is DR-07.


Xunzhuo (approved):
The comparator now applies the configured strategy consistently within each tier and keeps catch-alls behind real matches in that tier. The changed decision regressions and six focused ExtProc routing assertions pass; the previous comparator fails the three intended behavior changes. Current-head Core and quality CI pass. No live Envoy deployment was rerun.

wilsonwu (approved):
LGTM

PR #3975, opened 2026-09-20
[Feature] Report how selection ranked the matched decisions

subin9:
Related #3081

## Purpose

DR-07 from the ranking proposal. The engine decided an order and reported nothing about it, so an operator who asked why a decision won could read the matched rules and the confidence but not which key actually separated it from the runner up, and not why confidence did not apply where it looked like it should.

Selection now records one trace per evaluation: the strategy that ran, whether tier ordering was in play and which tier the winner came from, whether that pool was comparable, the decision that made it incomparable and in which terms, the score kind behind the winner's confidence, the key that decided, and how many decisions matched. The fallback reason names the decision, so `keyword_route reported no comparable score` and `embedding_route reported a similarity against a probability` are the two shapes an operator sees.

It travels on the existing diagnostics that already carry the applied unknown policies, so both the traced and the untraced evaluation paths produce it, and the eval API reports it as `decision_ranking`. No new header and no new endpoint: a header would be a second contract to keep aligned, and the question this answers is one an operator asks while looking at a request, not on every response.

## Test Plan

`go test ./src/semantic-router/pkg/decision/ ./src/semantic-router/pkg/services/`

## Test Result

Five cases in `pkg/decision` cover a confidence winner in a comparable pool, a pool that mixed a similarity with a probability, a pool holding a member that reported no score, two decisions separated only by name, and a single matched decision. Each asserts the recorded key and, where one exists, the exact fallback sentence. One case in `pkg/services` asserts the eval response carries the trace rather than dropping it at the boundary.

Not covered here: no recipe probe asserts the trace at the routing boundary yet, which is DR-08, and the dashboard does not display the field.


src/semantic-router/pkg/services/classification_signal_types.go
```diff
@@ -136,6 +136,7 @@ type EvalResponse struct {
 	SignalValues           map[string]float64                      `json:"signal_values,omitempty"`      // Raw signal values per signal when exposed, e.g. "structure:many_questions" -> 4
 	SignalErrors           map[string]string                       `json:"signal_errors,omitempty"`
 	AppliedUnknownPolicies map[string]string                       `json:"applied_unknown_policies,omitempty"`
+	DecisionRanking        *decision.RankingTrace                  `json:"decision_ranking,omitempty"` // How selection ordered the matched decisions
```
Xunzhuo, line 139: [P2] Regenerate the published OpenAPI response contract

Adding `DecisionRanking` changes the reflected response schema for `/api/v1/routing/preview`, but the committed `website/static/openapi/apiserver/apiserver.openapi.json` still omits `decision_ranking`. The actual generator adds this object to both the 200 and 503 response schemas, while the committed artifact has neither property. Consumers of the published schema cannot discover or generate a typed client for the new diagnostic field. Please regenerate and commit the OpenAPI artifact with this response change.

Xunzhuo (changes requested):
The ranking trace reaches the preview response in both traced and untraced requests, and the focused ranking tests pass. One public contract update is missing: please regenerate the published OpenAPI artifact for `decision_ranking`. Running the OpenAPI generator and comparing its output confirms this drift. Validation covered the Core CI test evidence and a focused HTTP boundary regression; a model-backed recipe collision probe was not run.

src/semantic-router/pkg/services/classification_signal_types.go
subin9: Regenerated the artifact, and `make api-docs-check` now reports it up to date. `decision_ranking` appears in both the 200 and the 503 response schemas for `/api/v1/routing/preview`.

subin9:
Thanks : ) The collision probes are not in this PR. They land in the next one, which adds a probe per direction in the agent and privacy recipes. I ran the agent one against a router built from this branch: both tier 1 decisions matched and selection took `local_security_containment`, with a ranking trace reading `decided_by: priority` and `comparable: false`.

Xunzhuo (approved):
The published OpenAPI contract now includes decision_ranking in both response schemas, resolving the earlier finding. The ranking logic and preview wiring look good. Current-head Core and generated-contract checks pass, and a focused HTTP regression confirms the ranking survives both traced and untraced preview requests. A live model-backed recipe collision probe was not run.

Issue #3038, opened 2026-08-26
[Epic] Evaluate each Mixture-of-Models as a first-class model


Xunzhuo:
## Parent Direction

Evaluation & Quality — #2969

## Goal

Evaluate every published Mixture-of-Models as one user-visible model identity, compare it fairly with standalone models, and publish reproducible launch and regression scorecards.

Decision-level evaluation in #2333 asks whether an individual routing decision was correct. This Epic asks whether the complete MoM delivers a better model experience.

## Evaluation contract

Every run identifies the MoM version, model pool, backend revisions, recipe, Router Models, collaboration, fallback, cache and policy settings, datasets, evaluators, runtime, environment, and qualified standalone baselines.

Every MoM runs a mandatory versioned core suite. Objective-specific packs may extend it for cost, latency, security, domain, multimodal, or agentic goals, but cannot hide a core regression.

## Child issues

- #3238 — versioned manifest and result schemas; resolves proposal #3086.
- #3239 — manifest validation and recipe conformance.
- #3240 — mandatory core suite and standalone-model comparison runner.
- #3241 — objective-specific evaluation pack registry and runners.
- #3242 — regression reports and failure slices.
- #3243 — reproducible scorecards in MoM model cards.
- #3244 — reference MoM qualification and release-candidate gates.

Implementation PR #3083 remains paused until proposal #3086 and these bounded delivery slices are reviewed and assigned.

## Non-goals

- Replacing decision-level routing evaluation or isolated Router Model evaluation.
- Reducing every objective to one opaque aggregate score.
- Allowing extension packs to hide regressions in the common core suite.
- Turning benchmark publication into an SLA or automatic release decision.

## Completion criteria

- [ ] Every candidate MoM can run under one versioned, reproducible contract.
- [ ] The same run compares the MoM with qualified standalone baselines.
- [ ] At least one extension pack proves the registry contract.
- [ ] Scorecards include provenance, comparisons, limitations, and contract version.
- [ ] Two representative MoMs complete the contract-to-publication path.

## Interfaces

MoM & Routing defines MoM identity, objective, pool, recipe, and lifecycle. Other Workgroups supply implementation evidence. Evaluation & Quality owns the shared measurement, comparison, regression, and publication contract.

sujal111:
@Xunzhuo  @FAUST-BENCHOU  @AayushSaini101  can I start working on this ?

PR #3980, opened 2026-09-20
[Bug] Include SciCode plotting dependencies in the grading image

Xunzhuo:
Related #3038

## Purpose

SciCode tasks can import `mpl_toolkits`, but the grading image omitted Matplotlib. Include pinned Matplotlib and validate grading-library imports during image construction so missing dependencies fail setup before it records a new successful image build.

## Test Plan

Exercise sandbox setup and grading tests, including each dependency import failure, then run changed-path static and CLI documentation checks.

## Test Result

- 62 focused setup, grading, and harness tests passed.
- Changed-path static checks, CLI reference validation, and 10 documentation tests passed.
- Docker build behavior is covered with fixtures; an actual Linux image build has not been qualified by these tests.

<details>

- [x] The title uses a single bracketed category.
- [x] The title does not stack category prefixes.
- [x] The change references an accepted issue with one accountable Workgroup.
- [x] All commits include a DCO sign-off.
- [x] Purpose, Test Plan, and Test Result describe this change and its validation accurately.

</details>


FAUST-BENCHOU:
The dependency fix is scoped well, and validating the pinned scientific imports during image construction is a useful guard against recording a superficially successful but unusable grading image. The fixture coverage also verifies that a failed import does not publish the manifest. Looks good from the code side; the remaining confidence point is the real Linux image build.

PR #3981, opened 2026-09-20
[Bug] Reject unsupported dataset composition fields

Xunzhuo:
Related #3038

## Purpose

Dataset composition silently ignored unsupported request fields, so a caller could believe an exclusion or count constraint had been applied. Reject fields other than `dataset_ids` and `benchmarks` with HTTP 400 before publishing a dataset. Valid whole-benchmark composition keeps its existing behavior.

## Test Plan

- `PYTHONPATH=src/vllm-sr python -m pytest -q src/vllm-sr/tests/test_sr_bench_datasets.py src/vllm-sr/tests/test_sr_bench_dataset_validation.py src/vllm-sr/tests/test_sr_bench_dataset_fingerprints.py src/vllm-sr/tests/test_sr_bench_large_datasets.py`
- Synthetic HTTP regressions cover unsupported exclusion/count fields, unchanged dataset files after rejection, valid composition and no run submission.
- Changed-file static checks, CLI reference checks and documentation unit tests.

## Test Result

All three new HTTP cases reproduced the former HTTP 200 behavior before the fix. The four focused dataset suites pass 120 tests. Broader selected CLI tests pass 945 tests with six Linux-only skips on macOS; documentation unit tests pass 10 tests. Static and CLI reference checks pass. Final-head Linux CI passes all five selected verifications: 1,001 CLI cases, 1,851 Dashboard cases and 14 quality, security and packaging checks, with no selected skips or failures. DCO, Community and both final gates pass.

---
<details>

- [x] The PR links an accepted issue with exactly one Workgroup owner
- [x] Purpose, Test Plan and Test Result reflect the actual scope and remaining checks

</details>


FAUST-BENCHOU:
Good fail-closed behavior. Rejecting unsupported composition fields before publication is much safer than silently producing a dataset that does not honor the caller's constraints, and the regression verifies both store immutability on rejection and the unchanged valid path. LGTM.

PR #3984, opened 2026-09-21
[Test] Probe the routes that collide inside one tier

subin9:
Related #3081

## Purpose

DR-08 from the ranking proposal, and the last slice.

The maintained probes assert the decision a request selects, but every one of them describes a request that only one decision claims. Nothing would notice if the order between two competing decisions changed, which is exactly what the four preceding pull requests changed.

Two probes now cover a collision in each direction. In the agent recipe, a prompt-injection request that also carries an image matches `local_security_containment` through its security projection and `omni` through its image predicate, both in tier 1, and the recipe's own priorities select containment. In the privacy recipe, an on-prem request carrying an image matches `omni` and `local_privacy_policy` in tier 2, and the priorities select `omni`.

The first probe is the one with teeth. Before the evidence contract, `omni` reported a structural 1.0 as though it were measured evidence and outranked the projection score beside it, so that request selected `omni` whatever priorities the recipe declared.

## Test Plan

`make recipe-conformance-static`, and both probes against a router built from this branch.

## Test Result

Static conformance passes, including the coverage policy each recipe declares.

Both probes ran live against a router built from this branch.

Serving the agent recipe, the router answered the first probe with `local_security_containment` and reported `agent_has_images` and `policy_security_local_only` among the matched signals, so both tier 1 decisions did claim the request. Its ranking trace read `decided_by: priority`, `comparable: false`, naming the decision that reported no comparable score.

Serving the privacy recipe, the router answered the second probe with `omni`, reporting `privacy_has_images`, `local_handling_required` and `policy_privacy_local_only`, which is both tier 2 decisions matching. Its ranking trace read tier 2, `decided_by: priority`, two candidates.


FAUST-BENCHOU:
These probes add the missing kind of coverage: two decisions actually contend inside the same tier, so the test exercises ranking rather than merely confirming predicate matching. Covering the priority outcome in both directions makes the recipe contract much harder to regress accidentally. LGTM.

Xunzhuo (approved):
Both new probes exercise real same-tier collisions: the agent case matches containment and omni, while the privacy case matches omni and local handling. The exact-head live recipe reports confirm the declared priority selects the expected route in each case. No blocking findings.

wilsonwu (approved):
LGTM

PR #3985, opened 2026-09-21
[Bug] Fence sr-bench run admission during shutdown

Xunzhuo:
Related #3038

## Purpose

A run request can be accepted before shutdown but still be preparing its plan when cancellation events are collected. It can then create a worker that shutdown missed. Serialize final run creation and worker registration with closing admission so every accepted worker is cancelled and later attempts are rejected before persistence or dispatch.

Late new-run and recovery requests return HTTP 503 with `service_stopping`. Matching existing idempotency keys remain reconcilable, and unknown sent calls remain ineligible for automatic recovery. The signal handler starts a shutdown thread instead of acquiring engine locks directly.

The change fences new Engine attempts. It does not globally drain other store mutations or imply that already dispatched calls have completed.

## Test Plan

- `PYTHONPATH=src/vllm-sr python -m pytest src/vllm-sr/tests/test_sr_bench_shutdown.py src/vllm-sr/tests/test_sr_bench.py src/vllm-sr/tests/test_sr_bench_recovery.py`
- `make docs-cli-check docs-cli-test`
- Four-file `make check CI_STATIC_ONLY=true`, `make harness-check`, and `git diff --check`.

## Test Result

37 scoped tests pass, including seven new deterministic concurrency and real HTTP cases. CLI-reference tests, formatting, lint and architecture checks pass. The harness's 380 CI-tool tests and 74 agent-script tests pass across the original run and targeted reruns of six local fixture environment failures; the original aggregate command did not exit successfully. Registry, skill mirror, workflow contracts and pinned actionlint checks also pass. Synthetic workers make no model requests.

Final published-head CI is pending.

---
<details>

- [x] Affected modules are described without stacked title prefixes
- [x] Related issue is accepted and owned by `wg/evaluation-quality`
- [x] Purpose, Test Plan and Test Result reflect the scope and remaining CI qualification

</details>


FAUST-BENCHOU:
The admission fence closes the important race at the right boundary: durable creation and worker registration are now indivisible with respect to shutdown, while existing idempotency keys remain reconcilable after admission closes. The deterministic persisted-before-registration test is especially useful here. No code blocker from my pass; I’d wait for the pending final-head CI before approval.

PR #3986, opened 2026-09-21
[Feature] Record the decision ranking in the replay record

### LINKED ISSUE #3081: [Feature] Define explicit decision ranking semantics for tier, priority, and confidence

subin9:
Closes #3081

## Purpose

The issue asks for eval and replay traces that explain which ordering mode selected the winner. The eval API reports `decision_ranking` since #3975, but a replay record carried only the applied unknown policies, so the one surface an operator uses to reconstruct a past request could not answer why that decision won rather than the one beside it.

A record now carries the same ranking: the strategy selection ran under, the tier the winner came from, whether that pool ranked by confidence and which decision stopped it, the key that decided, and how many decisions matched.

The persistence type mirrors `decision.RankingTrace` rather than importing it, the way `HallucinationSpan` in the same file mirrors its extproc counterpart. A stored record keeps its shape when the engine's internal struct moves, and the store does not pull the decision engine in to hold nine scalars.

This is the last completion signal left on the issue. The ordering table and `decision.tier` are documented, evidence roles and score kinds have distinct runtime semantics, configuration load names the pools that fall back, the eval API and now replay both report the mode, unit tests cover mixed pools, priority ties, catch-alls, predicates, guards and heterogeneous provenance, and #3984 adds the maintained recipe probes for a cross-route collision.

## Test Plan

`go test ./src/semantic-router/pkg/extproc/ -run TestRespondDecisionUnresolvedFinalizesReplayAsFailed`, and `make api-docs-check`.

## Test Result

The replay regression now asserts the stored ranking, not just that a record exists: the winner, the key that decided, and the sentence naming the decision that reported no comparable score.

Adding the field changed the reflected replay response schema, so the published OpenAPI artifact is regenerated in the same commit and `make api-docs-check` reports it up to date.

Not covered here: the Dashboard does not display the field yet, and no end to end case reads a stored record back to assert it.


Xunzhuo (approved):

FAUST-BENCHOU:
Nice persistence boundary. Mirroring the ranking trace into a replay-owned DTO keeps stored records stable while still preserving the exact strategy, tier, comparability and deciding key needed to explain a historical route. The clone path, regression assertion and generated API contract are all updated together. LGTM.

PR #3521, opened 2026-09-05
[Refactor] Add an ordered context transformation plan to the existing IR

ethanjyx:
Closes #3472

## Purpose

Give reset, exact deduplication, turn selection, and the existing block compressor one request-owned transformation plan. Policies receive detached views and propose edits; shared validation preserves protected messages, complete selected turns, tool-exchange membership, and retained message order before committing a step.

Extend the existing `RequestIR` and `MessageIR` with trusted provenance, eligibility, protection reasons, and stable turn/exchange identity. Capture original semantic history before RAG and Memory enrichment so later topic evidence can read an immutable snapshot. Route configured compression through the same plan and record bounded receipts without request text or arbitrary callback errors.

No public configuration or new policy is enabled. RAG still precedes Memory; configured tool-result text compression remains supported while non-text blocks and tool structure stay intact. Reset triggers, deduplication equivalence, ranking, summarization, and token budgeting remain separate work.

Owner: `wg/agentic-context`.

## Test Plan

- Run the contextcompression and protocolcodec suites, including the race detector.
- Fuzz history-removal invariants and verify rejected edits never partially apply.
- Verify byte-compatible disabled plans and transformed round trips for OpenAI Chat, Responses, and Anthropic.
- Run request-pipeline tests with native libraries, repository lint, and the local image startup checks.

## Test Result

Passed on commit `7513f43d857d692e1e5b650ab60e9b7e23e36b04`:

- `make agent-validate`.
- `SKIP_MODEL_DEPENDENT_TESTS=true make agent-ci-gate AGENT_CHANGED_FILES_PATH=/tmp/context-plan-changed-files`: native Candle/ML/NLP builds, router build, baseline pre-commit, Go structural lint, structure and architecture checks, and 55 Go package test results, including the complete `pkg/extproc`, `pkg/contextcompression`, and `pkg/protocolcodec` suites.
- `go test -race ./pkg/contextcompression ./pkg/protocolcodec` from `src/semantic-router`.
- `go test ./pkg/contextcompression -run '^$' -fuzz '^FuzzTransformationRemovalInvariants$' -fuzztime=10s -parallel=2` (61,889 executions).
- Nine Linux request-pipeline tests in the final Dockerfile's `go-builder` image, with native Candle/ML/NLP libraries: original history capture, history selection without compression, tool-result compression, request controls, non-text preservation, RAG opt-in, neutral-request preparation, fail-closed handling, and recovery-tool name protection.
[...]

Issue #3472, opened 2026-09-04
[Refactor] Define an ordered context-transformation plan and protected-turn IR


Xunzhuo:
## Goal

Extend the existing `contextcompression.RequestIR` and `MessageIR` with one ordered transformation plan and protected-segment contract so reset, deduplication, turn selection, and compression cannot independently mutate the same request history.

## Proposed Workgroup

Agentic & Context

## Background

Current request preparation already applies RAG, then Router Memory, then the context-compression stage before provider encoding. The active children of #2984 add topic evidence, optional history reset, deduplication, prior-turn selection, and summarization around that shared path. A second IR or another raw-message mutation pipeline would make ordering and protected-content behavior diverge.

## Work item shape and dependency

This is the shared integration seam for behavior-changing context work. #3342 signal-contract work and #3345 research can proceed independently. #3343, #3344, and #3346 may develop bounded policy and fixtures in parallel, but their request-pipeline integration must use this seam after it lands.

## Scope

- Extend, rather than replace, `pkg/contextcompression.RequestIR` and `MessageIR` with typed source, eligibility, protection, and complete-turn/tool-exchange identity.
- Preserve an immutable view of original conversation history for #3342 topic evidence so injected RAG or Memory content cannot change that signal accidentally.
- Keep the current RAG-then-Memory preparation order and represent their injected content with explicit provenance before lossy context transforms run.
- Define the behavior-changing order as eligible-history reset (#3343), exact safe deduplication (#3346), complete-turn selection (#3344), existing configured block compression, then provider encoding.
- Protect system/developer instructions, the live user turn, multimodal blocks, authorization/safety context, and atomic assistant/tool exchanges throughout the plan.
- Give each step declared inputs, outputs, protected invariants, skip/failure behavior, and a content-minimized receipt.
- Add unit and property tests for baseline preservation, ordering, idempotency, protected content, RAG/Memory provenance, tool-exchange atomicity, and protocol-neutral round trips.

## Configuration and API impact

This is an internal refactor. It adds no public configuration or client protocol and preserves current RAG, Memory, and compression behavior until an existing child policy explicitly enables a new step.

## Non-goals

- Create a second context IR or compression framework.
- Implement topic detection, reset policy, deduplication policy, turn ranking, summarization, RAG, Memory, or token budgeting.
- Reorder current RAG and Memory behavior as an incidental refactor.
- Permit lossy mutation without an explicit plan entry and receipt.

## Completion signal

[...]

PR #3987, opened 2026-09-21
[Feature] Deduplicate adjacent repeated context turns without changing conversation semantics

### LINKED ISSUE #3346: [Feature] Deduplicate repeated context without changing conversation semantics
### Goal

Remove only context segments that are provably redundant for the current request while preserving temporal meaning, confirmations, retries, provenance, and tool state.

### Background

Repeated transport or history material can consume context more than once, but identical text is not automatically redundant: repetition may represent confirmation, retry, correction, elapsed time, or a second tool execution. Current `main` has no narrow identity-based deduplication stage. This task defines the deterministic subset that can be removed without changing conversation meaning.

Parent Epic: #2984

### Integration dependency

Identity rules, eligibility policy, and fixtures may proceed independently. Request-pipeline integration must use #3472 between eligible-history reset and complete-turn selection and must not add a second raw-history mutation path.

### Proposed Workgroup

Agentic & Context

### Scope

- Define a narrow eligibility contract for exact or explicitly normalized duplicate content.
- Preserve message role, turn position, timestamps or ordering facts, and source provenance when they affect meaning.
- Treat tool calls and tool results as atomic exchanges and retain repeated executions when their identity or outcome differs.
- Distinguish duplicated transport or history material from intentional repeated instructions, confirmations, corrections, and retries.
- Apply deterministic bounds and fail open when equivalence is uncertain.
- Record content-minimized deduplication plans, counts, reasons, and recovery metadata.

### Non-goals

- Semantic near-duplicate detection or paraphrase collapse.
- Router response-cache lookup or Router Memory consolidation.
- Removing repeated system, developer, authorization, or safety instructions merely because their text matches.
- Reordering retained content.

### Completion signal

- [ ] The eligibility and identity rules are explicit and testable.
- [ ] Tool exchanges, corrections, confirmations, retries, and temporal repetitions have regression coverage.
- [ ] Uncertain equivalence retains the original content.
- [ ] Token reduction is measured together with behavioral equivalence on representative workloads.
- [ ] Replay and recovery receipts remain bounded and content-minimized.

yaojiejia:
Adds `context_dedup`, a disabled-by-default route-local context action that removes the later copy of prior turns a request carries twice in a row, as the `TransformDeduplicate` step of the shared context-transformation plan.

Closes #3346

## Purpose

Clients that rebuild a request from a transcript, retry after a timeout, or prepend server-side history to input they already hold deliver the same turns twice. The model reads the same exchange more than once, which costs input tokens and can bias it toward repeated material. Removing text that merely looks similar is unsafe: a repeated "yes" is a confirmation, a repeated question is a retry, and a second identical tool execution may have happened at a different time. This change removes only the narrow, provable case and keeps everything else. 

The step runs in the plan from #3472 / #3521, so it never mutates raw provider payloads and adds no second raw-history mutator:

```
decode → RAG → Memory → history reset → deduplication → turn selection → compression → provider encoding
```

Rules, in the order the issue asked for them:

- **Eligibility.** A turn is a candidate only when every message is unprotected history, no message belongs to a tool exchange, the turn opens with a user message and contains an assistant reply, and every message is text the policy can see. Each failed rule is counted under a bounded retention reason (`ineligible`, `tool_exchange`, `incomplete_turn`, `opaque_content`).
- **Identity.** Stage one compares roles and normalized text through the detached view. Stage two proves every matched message pair against the neutral request through a resolver: same role, equal item IDs or an empty later ID, the same blocks, only text or reasoning blocks, equal citations, cache directives, signatures, and reasoning scope. A refusal keeps the turn (`refusal_retained`); any other difference keeps it (`identity_mismatch`). Text equality alone is never permission.
- **Segments.** Only adjacent repetition is removed: a block of consecutive candidate turns immediately followed by an identical block loses its later copy, and the position is rescanned so a triple send collapses to one copy. Identical turns separated by other content stay (`non_adjacent`). The earlier copy is always kept, so ordering and provenance never change.
- **Bounds and failure.** `max_history_turns`, `max_history_bytes`, `max_segment_turns`, and `timeout_ms` reject the whole step, never a prefix; only eligible history counts against them. Every failed evaluation removes nothing; `fail_open` keeps the request and `fail_closed` rejects it before dispatch.
- **Receipts.** Counts, bounded reason codes, per-reason retention counts, and a bounded list of removed segments by pre-transform message index. No text reaches Router Replay, metrics, or logs. No recovery store is needed: every removed message has a byte-identical retained twin, so `recovery_status` is `not_required`.

Modules affected: `pkg/contextdedup` (new), `pkg/config`, `pkg/extproc`, `pkg/dsl`, `pkg/routerreplay/store`, `pkg/observability/metrics`, the generated config schema, OpenAPI artifact and CRDs, Dashboard config surfaces, `config/config.yaml` and `config/fragments/plugin/context-dedup/`, `e2e/`, and `website/docs/tutorials/plugin/context-dedup.md`.

Integration notes:

- The response cache is not bypassed: its identity covers the request as received and the decision's plugin configuration, and deduplication is a deterministic, behavior-preserving function of both.
- Steps are inserted by kind, so the plan order holds whichever of reset (#3786) or dedup registers first.
- Enabling any history action turns on the shared live-turn compression protection, so enabling dedup can slightly reduce what a separately configured `context_compression` reports. Documented.
- Decision admission and the compression overflow path run before this step, so it lowers the tokens a dispatched request carries but does not make an oversized request admissible. Documented; moving admission after the lossless steps is a pipeline-level follow-up.
- Internal Looper hops do not run this step until the shared context stage is extended to internal requests (#3786 changes that path). Documented.

Out of scope, listed as follow-ups: deduplicating repeated tool exchanges with normalized IDs once the shared view exposes tool call name, arguments, and result outcome; running the step on Looper internal hops; evaluating lossless steps ahead of context admission.

## Test Plan

```bash
cd src/semantic-router
go test -race -cover ./pkg/contextdedup/
go test -run TestEvaluation -v ./pkg/contextdedup/        # token reduction with behavioral equivalence
go test -run xxx -bench BenchmarkPlan ./pkg/contextdedup/
go test ./pkg/config/ ./pkg/dsl/ ./pkg/routerreplay/... ./pkg/observability/metrics/
CGO_ENABLED=1 SR_TEST_MODE=true go test ./pkg/extproc/    # with the native libraries on LD_LIBRARY_PATH
cd ../..
make config-schema-generate && make config-schema-check
make api-docs-generate && make api-docs-check
make generated-contract-check
cd website && npm run config:check && cd ..
cd dashboard/frontend && npx vitest run src/lib/dslCapabilityPluginSchemas.test.ts src/pages/configPageRoutingSurfaceContract.test.ts && cd ../..
cd e2e && go test ./pkg/testmatrix/ ./pkg/verification/ ./profiles/ai-gateway/ && cd ..
make e2e-test E2E_USE_EXISTING_CLUSTER=true E2E_TESTS=context-dedup-adjacent-turns,context-dedup-disabled
```

Unit coverage in `pkg/contextdedup` includes: adjacent single-turn, multi-turn block, and triple-send removal; non-adjacent, correction, retry, confirmation, refusal, and tool-exchange retention with both repeated and distinct call IDs; Anthropic `tool_result`-in-user turns; Memory and RAG content alongside eligible history; multimodal, live-turn, authorization, safety, and instruction protection; Responses item IDs equal, absent on the later copy, and distinct; whitespace normalization; limits, timeout, cancellation inside stage-two proof, and a missing or failing resolver; idempotent re-application; receipt content leakage; round trips through the OpenAI Chat, Responses, and Anthropic codecs; and reconciliation against executor receipts including a late cancellation. `pkg/extproc` tests cover binding, single registration, step ordering with a reset step registered afterwards, fail-open and fail-closed rejection mapping, the raw-request block, and replay diagnostics. E2E asserts the mock backend echo: the later copy of an adjacent repeated turn is removed while a corrected reply and a repeated confirmation reach the provider in full, and a disabled policy forwards everything.

## Test Result

- `pkg/contextdedup`: pass with `-race`, 97.6% statement coverage. `pkg/config`, `pkg/dsl`, `pkg/routerreplay/...`, `pkg/observability/metrics`, and the full `pkg/extproc` suite pass.
- Evaluation (heuristic token counter, default policy; every negative workload byte-identical before and after):

  | Workload | Before | After | Saved |
  | --- | --- | --- | --- |
  | Clean eight-turn conversation | 1854 | 1854 | 0% |
  | One turn sent twice in a four-turn history | 1090 | 899 | 18% |
  | Six-turn history sent twice | 2618 | 1472 | 44% |
  | Six-turn history sent three times | 3764 | 1472 | 61% |
  | Responses history with item IDs re-sent without IDs | 1864 | 1100 | 41% |
  | Repeated tool execution, distinct or identical call IDs | 1082 | 1082 | 0% |
  | Confirmation, correction, refusal retry, temporal repetition | unchanged | unchanged | 0% |

  `BenchmarkPlan`: about 0.16 ms for a 63-turn history sent twice at the default bounds.
- Schema, OpenAPI, catalog, generated-contract, dashboard, and E2E drift gates pass. `make check BASE_REF=origin/main` exits 0.
- E2E on the `envoy-ai-gateway` profile in a kind cluster: `context-dedup-adjacent-turns` and `context-dedup-disabled` both pass (adjacent repeat: 5 messages sent, 3 received upstream; correction and confirmation: 5 and 5; disabled: 5 and 5).
- Remaining risk: the admission-ordering limitation above is documented rather than changed, and Looper internal hops are not covered until #3786 lands.

---
<details>

</details>


src/semantic-router/pkg/contextdedup/step.go
```diff
@@ -0,0 +1,163 @@
...
+	if a.blocked != "" {
+		a.record(failed(a.policy, view, a.blocked))
+		return contextcompression.TransformationEdits{}, errBlocked(a.blocked)
+	}
+	if a.policy.Timeout > 0 {
+		bounded, cancel := context.WithTimeout(ctx, a.policy.Timeout)
+		defer cancel()
+		ctx = bounded
+	}
+	edits, diagnostics := plan(ctx, a.policy, a.resolver, view)
```
Xunzhuo, line 64: [P2] Reject edits when the deduplication deadline has expired

Please check the derived context after `plan` returns before accepting its edits. `plan` only polls initially and every 64 comparisons, so a small number of large text turns can finish after `timeout_ms` without another check; the shared executor then checks the still-live parent context. With `timeout_ms: 1`, `fail_closed`, and two identical complete turns totaling 1 MiB (within the default byte bound), I reproduced the stage taking 1.6–2.1 ms, returning no error, and removing two messages in all three runs. The configured timeout should instead preserve the request and reject dispatch. Add a regression for a deadline expiring during a scan with fewer than 64 comparisons.

Xunzhuo (changes requested):
The adjacent-turn and protocol protections are covered by the new unit tests and deployed enabled/disabled cases, but the deduplication timeout can expire without preventing edits under fail_closed. Please check the derived deadline before returning the proposal and add a regression for that path. I reproduced the failure with a 1 MiB history within the configured bound; a post-plan deadline check makes the same cases reject without removing messages.

src/semantic-router/pkg/contextdedup/step.go
yaojiejia: Good catch, thanks. Fixed in b1fa39a83: the action now checks its own deadline after the plan returns, so anything produced after the timeout gets dropped as cancelled, meaning fail_closed rejects without removing anything and fail_open just leaves the request alone. I also added TestActionDeadlineExpiringDuringShortScanRejectsEdits, which expires the deadline inside a scan of under 64 comparisons and checks both modes.

src/semantic-router/pkg/extproc/req_filter_context_dedup.go
```diff
@@ -0,0 +1,175 @@
...
+	if ctx == nil || !ctx.ContextDedupPolicy.IsEnabled() || ctx.ContextDedupAction != nil {
+		return
+	}
+	blocked := ""
+	if request == nil || ctx.SemanticRequest == nil {
+		blocked = contextdedup.ReasonUnsupportedRepresentation
+	}
+	action := contextdedup.NewAction(contextDedupPolicy(ctx.ContextDedupPolicy), blocked)
+	action.WithResolver(func(id int) (llmprotocol.Message, bool) {
+		return contextdedup.RequestResolver(ctx.ContextRequestIR)(id)
```
Xunzhuo: [P2] Retain the resolver across message lookups

Calling `RequestResolver(ctx.ContextRequestIR)` inside this callback creates a fresh position cache for every lookup, rebuilding the entire message index for each comparison. With the default 128-turn / 1 MiB / 50 ms limits, a 17 KB history containing 2,049 messages exhausted the deadline in all three isolated runs using this wrapper; retaining one resolver completed in 1.0–1.2 ms. This maps to a 503 under fail-closed or skipped deduplication under fail-open. Please lazily create one resolver after the IR exists, reuse it for the action, and cover the production registration path. Its existing generation/count checks handle later removals.

Xunzhuo (changes requested):
The deadline fix and both failure-mode regressions pass. I found a separate integration issue: the registered resolver discards its index on every lookup, causing eligible histories to exhaust the default timeout. Please retain the request-local resolver and add coverage through the registration path. The deduplication package tests pass; the issue is reproduced with the actual planner and shared executor using the production resolver callback. I did not rerun a deployed Envoy stack.

src/semantic-router/pkg/extproc/req_filter_context_dedup.go
yaojiejia: Good catch. Fixed in 6abb43192: the resolver is now created once the IR exists and reused for the rest of the request instead of being rebuilt per lookup. Added TestContextDedupLargeHistoryCompletesWithinDefaultTimeout, which runs ~3,000 eligible messages through the registration path at the default limits; it failed with cancelled before and passes in about 10 ms now.

PR #3998, opened 2026-09-21
[Feature] Export a shadow comparison dataset from replay records

subin9:
Related #3280

## Purpose

Shadow dispatch records each candidate answer on the primary replay record, and the manifest builder that turns those observations into a versioned comparison dataset landed with the previous slice. Nothing called it. `shadowdataset.Build` had no caller anywhere in the tree, no documentation, and no way to reach it from a running router, so an operator holding a week of shadow traffic still had no dataset and the issue's offline-export requirement had nothing to assert against.

`GET /api/v1/observability/replays/dataset` is that caller. It reads the records the existing replay filters select, with their captured bodies, and answers with the manifest.

The split plan travels in the request rather than in configuration, because the manifest's digest is its identity and a citation has to name everything needed to rebuild it: `seed` fixes split assignment and each repeated `split` is written as `name:weight`. The same records under the same plan rebuild the same digest, so two people quoting one manifest are quoting the same rows in the same splits.

A selection larger than 5000 records is refused rather than paged. A page of a wider match is not the dataset its digest claims, and publishing an identity for something nobody can reproduce is worse than making the caller narrow the filters. Nothing else about what enters the dataset changed: an observation still enters whole or not at all, and each one left out is still counted under an explicit reason.

The manifest carries identity, output digests and lineage, never prompt or response text, so replay redaction has nothing to strip from it. A reader with `replay.read` and no detail rights reads the same bytes an operator does, which is the property that lets a manifest be published beside the numbers it supports.

Two things are deliberately not here. There is no configuration block, because every choice this export makes is already in the request and a config field with no behaviour behind it would only have to be migrated later. Storage destinations and blinded judging remain the next slices, and this endpoint hands the manifest back rather than writing it anywhere.

## Test Plan

At the API boundary, six unit cases hold the contract the handler owns rather than the one `Build` already owns: the export reads whole records instead of list summaries, repeats the same digest on a second call, applies the replay filters to the selection, refuses a selection past the cap, rejects an incomplete split plan, and rejects a non-GET. Two more cases hold the multi-arm contract in the manifest package itself: several candidates answering one request read in a fixed order regardless of the order the router appended their outcomes, and one candidate answering twice keeps both arms.

At the real boundary, a new case on the `router-replay` profile drives a live shadow dispatch to the candidate model through the gateway, then exports over the management API and reads the manifest back: one example, the primary arm naming the selected model with a digest, one shadow arm naming the candidate, lineage pointing at the replay record, no captured text anywhere in the body, the same digest on a repeat export, and the same digest again when a viewer token without replay-detail rights asks. The profile's shadow recipe now captures the request body, since an observation with no captured request is excluded from a comparison by design.

The published OpenAPI artifact and the endpoint index were regenerated from the route catalog.

## Test Result

The `router-replay` profile ran here against a local kind cluster and all eleven of its cases pass, including the new one. The exported manifest holds a single example whose primary arm is the selected model with its recorded digest and whose shadow arm is the candidate model, its lineage points at the replay record Postgres confirmed, and neither the prompt nor the answer appears anywhere in the response body. A second export and a read with a viewer token that has no replay-detail rights both return the same digest as the first.

The eight unit cases pass, and two were checked by deleting the code they cover. The export case fails if the handler asks for a list page instead of whole records, because a summary carries no request body and every example would then be excluded while the response still read as a success. The ordering cases fail if the arm sort is removed.

What is not covered by CI: the unit cases run against the in-memory store, which serves whole records from its list path, so they cannot catch a detail flag lost on a database query. Only the E2E case covers that, and the `router-replay` profile is `selection: manual`, so no PR lane runs it.


Xunzhuo (approved):
Reviewed the complete patch and the replay storage, dataset builder, authentication/redaction, and E2E call paths, including an independent pass. No blocking findings.

Validation: the exact-head core evidence includes all six new dataset export regressions and both manifest ordering/arm-retention tests passing; Router Tools ran 454 harness and 106 E2E unit cases. A fresh credential-free integration with real PostgreSQL exercised QueryPage through the extproc export handler and authenticated management mux, confirming whole filtered records, anonymous rejection, stable viewer/operator digests, and exclusion of captured prompt/response content.

The full deployed router-replay/Envoy shadow-dispatch profile and model inference were not run in this review.

wilsonwu (approved):
LGTM

Issue #3663, opened 2026-09-09
[Feature] Build selector data, objective, feature, evaluation, and parity foundations


Xunzhuo:
### Goal

Replace the current row-oriented selector training assumptions with one versioned foundation that KNN, KMeans, SVM, and MLP can consume consistently and that native runtimes can reproduce exactly.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Define a `QueryOutcomeSet` snapshot: one query and source identity with candidate outcomes containing model reference, success, quality, latency, cost, and optional approved signals.
- Pin candidate-set identity and split queries before feature, label, or utility derivation so duplicate `(query, model)` rows cannot leak across train, validation, and test sets.
- Define versioned selector objectives that actually apply configured quality, latency, cost, and failure weights instead of hard-coded `0.9/0.1` scoring.
- Define a content-addressed `FeatureSnapshot` that pins embedding model revision, tokenizer/preprocessing, pooling, normalization, dimension, extra-signal schema, and source snapshot digest.
- Remove ambiguous 768/1024 feature assumptions and prevent stale embedding-cache reuse across changed data or model revisions.
- Provide held-out evaluation against strongest, cheapest, global-best, random, current-router, and oracle baselines, reporting quality, latency, cost, regret, coverage, route share, and slices.
- Add a Python-to-native conformance harness for features, artifact parsing, per-candidate scores, selected model, and rejection behavior.
- Split shared dataset/objective/features/evaluation/export-conformance code from algorithm-specific trainers so four migrations can proceed independently.
- Do not run existing in-process `eval`/`exec` benchmark evaluation on user-provided content from the Console; require an isolated evaluator or disable that evaluator profile until one exists.

### Non-goals

- Implementing any one selector trainer or choosing a new routing algorithm.
- Storing prompts or responses without explicit data-retention and redaction policy.
- Replacing the routing-policy ownership of #2238 or the shared evaluation infrastructure of #2333.

### Completion signal

- [ ] One immutable snapshot feeds all four trainer fixtures without duplicated or conflicting query labels.
- [ ] Objective weights and feature provenance change artifact identity and are covered by tests.
- [ ] Split leakage, candidate-pool drift, stale cache reuse, and dimension mismatch fail deterministically.
- [ ] The conformance harness can compare a Python fixture with its native runtime result at score and decision level.

Depends on: #3661

Related: #2238, #2330, #2333, #3280

Parent Epic: #3660

[...]

PR #3999, opened 2026-09-21
feat(training): add query-outcome snapshots and query-level splitting

ankit373:
First increment of #3663. Opening as accepted-and-unowned per CONTRIBUTING; happy to close or re-scope if you'd rather it wait or land in a different order.

Selector training reads one row per `(query, model)` and has no train/val/test split today, so models are scored on what they learned from. A row-level split would be worse than none, since the same query would land on both sides. `QueryOutcomeSet` makes the query the unit and `split_by_query` partitions whole queries off a blake2b digest of query identity plus seed, so the holdout is identical across processes and input orderings. `candidate_set_id` pins the model roster so a changed one is a different snapshot.

Additive only, no trainer wired to it yet. 11 tests, black clean. Objectives, `FeatureSnapshot`, baseline evaluation and the native conformance harness follow separately.


Xunzhuo (approved):
The query-level split keeps every candidate outcome for a query on the same side of the holdout and remains stable across process hash seeds. Rechecked the complete patch and current-main integration; all 11 focused regressions, Ruff, Black, and the current CI gates pass. No GPU training or model-quality claim is included in this additive data-contract change.

wilsonwu (approved):
LGTM

PR #4001, opened 2026-09-21
[Test] Cover context compression effects at the provider boundary

P-Slark:
## Purpose

Related #3180. Continue after #3649 and #3795 by proving the route-local context_compression plugin changes the actual provider request.

## Changes

- add a deterministic e2e-plugins context-compression route
- assert large tool output is compressed before provider dispatch
- preserve tool-call correlation, recent messages, and the current user request
- verify an authorized bypass directive preserves the original tool output
- register the contract in the baseline E2E matrix

## Validation

- targeted ai-gateway, testcase, and testmatrix Go tests
- targeted Go vet
- E2E binary build and coverage drift gate
- deterministic compressor probe (2453 to 96 estimated tokens)
- git diff --check

The complete E2E unit runner reaches unrelated Windows-only command-fixture failures in pkg/docker and pkg/framework; all changed packages pass.


e2e/testcases/context_compression_e2e.go
```diff
@@ -0,0 +1,269 @@
...
+		"messages": []any{
+			map[string]any{"role": "user", "content": "Collect authentication diagnostics."},
+			map[string]any{
+				"role": "assistant",
+				"tool_calls": []any{map[string]any{
+					"id":   "call_context_compression",
+					"type": "function",
+					"function": map[string]any{
+						"name":      "diagnostics",
+						"arguments": `{"service":"auth"}`,
```
Xunzhuo: **[P2] Align the fixture with the tool-intent compression query**

`RequestIR.QueryFor` prefers the matching tool call's name and arguments, so this request is scored with `diagnostics {"service":"auth"}`, not the final user message about the authentication token validator. At the configured 96-token budget, the real compressor drops `authentication token validator failed`; the current Gateway run fails this exact new baseline case with "compressed output removed query-relevant evidence". The unit test supplies an already-compressed string and misses that mismatch. Include the intended query terms in the tool call (and update its envelope expectation), or use another fixture that exercises the real query-selection path, then verify both deployed cases.

Xunzhuo (changes requested):
The new required context-compression probe fails against the deployed router because its fixture selects a different compression query than its assertions assume. The exact-commit Gateway run reproduces this, and a focused compressor check confirms the tool-intent versus last-user difference. Please align the tool-call fixture with the evidence being retained and rerun the deployed extractive and bypass cases.

e2e/pkg/testmatrix/testcases_test.go
```diff
@@ -11,7 +11,7 @@ func TestBaselineScopesPreserveEveryFunctionalCase(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(full) != 38 || len(standard) != 36 {
+	if len(full) != 39 || len(standard) != 37 {
```
Xunzhuo: [P2] Keep the baseline inventory assertion in sync with the new case

Current `main` already includes `security-window-provenance` and expects 39 full / 37 standard cases. The clean merge of this head also retains `context-compression-provider-boundary`, producing 40 / 38 while `TestBaselineScopesPreserveEveryFunctionalCase` still expects 39 / 37. Running `go test -count=1 ./pkg/testmatrix` on that integration fails with `inventory changed: standard=38 full=40`; main alone passes. Please update the inventory assertion for the combined cases and rerun `make test-e2e-unit`.

Xunzhuo (changes requested):
The current-main integration exposes a deterministic baseline inventory regression: the added case is retained, but the test still expects one fewer case. The focused E2E matrix test fails on the clean merge and passes on main. Please synchronize the assertion and rerun the E2E unit gate. This follow-up was cluster-free; I did not rerun deployed runtime checks. The earlier compression-fixture finding also remains outstanding.

e2e/testcases/context_compression_e2e.go
P-Slark: Fixed in e2fd89cf1. The correlated tool call now carries `{"query":"authentication token validator failed"}`, and the provider-envelope assertion was updated accordingly. I also ran an exact real-compressor probe with the deployed policy values (`min_tokens: 200`, `target_tokens: 96`, BM25): the tool-output query contains the evidence phrase and the compressed result retains it. The focused E2E packages, vet, compressor package, and E2E binary build pass.

e2e/pkg/testmatrix/testcases_test.go
P-Slark: Fixed in e2fd89cf1 after merging current `main` in 21f8bdb03. The combined baseline assertion is now 40 full / 38 standard, and `go test -count=1 ./pkg/testmatrix` passes. The broader focused profile/testcase/matrix/verification suite and E2E binary build also pass. The complete Windows `go test ./...` still reaches unrelated host-dependent failures in cluster/docker/framework/soak/model-selection (missing Docker/kind and Windows fixture behavior).

P-Slark:
@Xunzhuo All requested changes are addressed: the compression fixture now uses the correlated tool intent, the baseline inventory is synchronized, and all review threads are resolved. I have also merged current main and rerun the focused E2E contracts successfully. Fresh hosted CI is starting; could you please re-review when it is green?

PR #3887, opened 2026-09-17
[Bug] Validate OpenClaw skill IDs before provisioning

zdy782:
Related #2468

## Purpose

Validate selected OpenClaw skill IDs against the server-owned skills catalog before provisioning starts. A request containing a path such as `../github`, an alias such as `GitHub` or ` github`, or an unknown ID now returns HTTP 400 before runtime discovery, workspace writes, or asynchronous job creation. The entire selection is checked, including mixed valid and invalid IDs.

IDs must be lowercase ASCII alphanumeric groups separated by single hyphens or underscores and must match a catalog entry exactly. Malformed catalog JSON returns a generic HTTP 500 without exposing its path. Empty or omitted selections retain the no-skills behavior. Add HTTP-handler regression coverage and document the input contract.

This is the first bounded slice of the accepted `wg/enterprise-environment` issue. It does not close #2468: symlink-aware workspace containment, server-owned skill content, image provenance restrictions, and container privilege reduction remain follow-up work.

## Test Plan

- In `dashboard/backend`: `go test -ldflags=-w ./handlers -run 'Test.*(OpenClaw|Provision|Skill)' -count=1 -timeout=120s`
- `git diff origin/main --check`
- Run codespell on the four changed files with the repository ignore-word list.
- `python3 tools/ci/sync_evaluation_catalogs.py --check`
- `make check CHANGED_FILES="dashboard/README.md dashboard/backend/handlers/openclaw_provision.go dashboard/backend/handlers/openclaw_skills.go dashboard/backend/handlers/openclaw_skills_test.go"`

## Test Result

The OpenClaw handler tests, changed-file spelling check, diff whitespace check, and evaluation catalog check pass on macOS ARM64 with Go 1.26.1. Tests cover synchronous and asynchronous rejection, unknown IDs, traversal and alias inputs, malformed catalogs, exact catalog membership, and omitted/empty selections. Positive selections advance to the deliberately disabled runtime check; these tests do not claim a successful container deployment.

`make check` stops at the repository-wide `codespell-tracked` baseline on existing multilingual entries in `config/recipes/built-in/latest/mom-v1/config.yaml` and `recipe.dsl`, before the dashboard checks. Full dashboard validation is not claimed. An earlier full backend test attempt on the previous base failed in evaluation deployment-registry tests on macOS; the focused OpenClaw tests were rerun successfully after rebasing onto current main. The `-w` linker option omits debug information to fit the local disk limit; it does not disable tests.

---
[...]

PR #4006, opened 2026-09-21
[Security] Constrain OpenClaw provisioning paths, assets, images, and runtime

### LINKED ISSUE #2468: [Security] Constrain OpenClaw image and skill provisioning inputs

zdy782:
Completes the remaining accepted scope of #2468. #3887 delivered the catalog-ID validation slice and explicitly left workspace containment, catalog-owned skill content, image provenance policy, and least-privilege deployment as follow-up; this PR delivers those.

## Workspace containment

- `validateProvisionPaths` proves the per-container data directory and workspace stay inside the handler dataDir after symlink-aware resolution. Container data dirs that resolve outside the data directory (e.g. a symlink planted under `containers/`) are rejected with a 400 before any workspace mutation.

## Catalog-owned skill content

- Skills are materialized exclusively from the server-owned skill pack directory (`skillpacks/`, overridable via `OPENCLAW_SKILL_PACKS_PATH`) with strict pack IDs matching the catalog ID pattern.
- `resolveOpenClawSkillPack` proves both the pack directory and its `SKILL.md` stay inside the skill-pack root after `EvalSymlinks`, rejecting unknown IDs, aliases, and escaping symlinks.
- `copyOpenClawSkillPack` copies known assets by server identity; destinations are always derived from the strict catalog ID and proven contained inside the workspace skills root. Directory entries resolve symlinks before copying and never leave the server-owned root.
- `fetchSkillContent` no longer runs `cat` inside the caller-selected image to extract content; it only emits the server-catalog metadata stub as a fallback for catalog entries without a server-side pack.

## Image provenance policy

- New administrator-controlled policy via `OPENCLAW_IMAGE_ALLOWLIST` (exact refs or `registry/repo/` prefixes), `OPENCLAW_IMAGE_ALLOWED_TAGS`, and `OPENCLAW_IMAGE_DIGESTS_ONLY` (requires `image@sha256:...`).
- Enforcement runs on the resolved image before any runtime work, workspace mutation, or async provisioning job starts.

## Least privilege

- Provisioned containers now run with a read-only root filesystem, all capabilities dropped, `no-new-privileges`, and a bounded `noexec,nosuid` `/tmp` tmpfs. Writable state stays in the existing named volume and bind mounts.
- The forced `--user 0:0` is removed so containers run as the image user; `docker run`, compose YAML, and the provisioning path all share the same flags.

## Tests

Negative tests without publishing host-specific paths:

- containment, separator, alias, symlink-escape, unknown-ID for skill pack resolution and copy
- image allowlist / digest / tag policy rejection and env parsing
- provision-path symlink escape and empty-name rejection
- least-privilege flags present in runtime args, docker run cmd, and compose YAML (both bridge and host variants)

Full `dashboard/backend` suite passes (`go test ./...`), `go vet` and `gofmt` clean.

Closes #2468


dashboard/backend/handlers/openclaw_skillpacks.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+			if err := copyDirWithinRoot(entryEval, targetPath, root); err != nil {
+				return err
+			}
+		case info.Mode().IsRegular():
+			data, err := os.ReadFile(entryEval)
+			if err != nil {
+				return fmt.Errorf("failed to read %s: %w", name, err)
+			}
+			if err := os.WriteFile(targetPath, data, 0o644); err != nil {
```
Xunzhuo: [P1] Reject symlinks in materialization destinations

`rootEval` only resolves the skills directory; the destination below it is never resolved or opened without following symlinks. If an existing workspace has `skills/github` or `skills/github/SKILL.md` linked to a writable path outside the workspace, this write follows that link and overwrites the outside file. Both cases reproduce with `copyOpenClawSkillPack` returning nil. Validate every destination against the trusted workspace and use symlink-safe creation; apply the same protection to the provisioning fallback instead of continuing with unchecked writes.

dashboard/backend/handlers/openclaw_image_policy.go
```diff
@@ -0,0 +1,186 @@
...
+		ref = ref[:idx]
+	}
+
+	if policy.AllowDigestsOnly && digest == "" {
+		return fmt.Errorf("image policy requires digest-pinned references (image@sha256:...); got %q", image)
+	}
+	if digest != "" && !strings.HasPrefix(digest, "sha256:") {
+		return fmt.Errorf("image policy only accepts sha256 digests; got %q", image)
+	}
+	if policy.AllowedTags != nil && tag != "" {
```
Xunzhuo: [P1] Validate the effective tag that the runtime will pull

With the repository allowlisted and `OPENCLAW_IMAGE_ALLOWED_TAGS=v1`, an untagged request such as `ghcr.io/openclaw/openclaw` leaves `tag` empty and bypasses this check, but Docker pulls `:latest`. Lowercasing the whole reference also lets a request for the distinct tag `:RELEASE` pass an exact `:release` allowlist, while provisioning passes the original reference to Docker. Parse the effective reference, apply the implicit `latest` tag, preserve tag case, and validate the same reference sent to the runtime.

dashboard/backend/handlers/openclaw_image_policy.go
```diff
@@ -0,0 +1,186 @@
...
+	cDir = filepath.Join(dataDirEval, "containers", containerName)
+	cDirEval, err := filepath.EvalSymlinks(cDir)
+	if err == nil {
+		// The directory already exists (reprovision): ensure it did not
+		// become a symlink pointing outside dataDir.
+		if !pathWithinDir(cDirEval, dataDirEval) {
+			return "", "", fmt.Errorf("container data dir resolves outside the data directory")
+		}
+		cDir = cDirEval
+	} else if !os.IsNotExist(err) {
```
Xunzhuo: [P2] Resolve existing parents before accepting a new path

When `dataDir/containers` is a symlink outside `dataDir` and the requested container directory does not exist yet, `EvalSymlinks(cDir)` returns ENOENT and this branch accepts the unresolved path. The workspace check does the same; the subsequent `MkdirAll` then creates the workspace outside the data directory. Resolve and validate the existing ancestor chain before appending missing components, and cover this new-container case in the containment tests.

Xunzhuo (changes requested):
Please close the destination-containment and image-policy bypasses before merging. Backend handler tests pass, while targeted regression probes reproduce the inline findings. Live OpenClaw provisioning was not run.

zdy782:
Addressed all three findings in 1f8844e:

- **P1 destination containment**: destination paths are now walked component-by-component via `containedMkdirAll` and `writeContainedFile`; pre-existing destination symlinks (`skills/<id>` or `skills/<id>/SKILL.md` pointing outside the workspace) are rejected instead of followed, and the provisioning fallback uses the same contained helpers. Covered by `TestCopyOpenClawSkillPackRejectsEscapingDestinationSymlink`, `TestCopyOpenClawSkillPackRejectsEscapingDestinationFileSymlink`, and `TestWriteContainedFileRejectsSymlink`, which verify the outside file is not modified.
- **P1 image policy bypass**: `splitImageReference` now applies the runtime's implicit `:latest`, so an untagged request no longer bypasses the tag allowlist (`TestValidateOpenClawImageUntaggedImpliesLatest`); tag comparison is case-sensitive so `:RELEASE` no longer satisfies an entry of `:release` (`TestValidateOpenClawImageTagCasePreserved`), while repository matching uses `strings.EqualFold` for case-insensitive normalization.
- **P2 ancestor resolution**: `resolveExistingWithin` resolves the deepest existing ancestor symlink-aware before appending missing components, so a symlinked `dataDir/containers` with a brand-new container name is rejected (`TestValidateProvisionPathsRejectsContainersSymlinkWithNewContainer`), and a fully-missing containers dir still provisions the first worker (`TestValidateProvisionPathsAcceptsDeepNewContainer`).

Also fixed the failing Quality/Dashboard lint checks (govet shadow findings, unused `openClawStateVolumeArgs`); the changed files are clean under `golangci-lint run ./handlers/`, and the full `dashboard/backend` suite passes on x86_64 with Go 1.25.

zdy782:
Follow-up hardening after re-checking the P1 destination-containment finding:

- I reproduced a filesystem TOCTOU in the pre-fix `writeContainedFile`: racing `SKILL.md` between a regular file and an escaping symlink could land the `os.WriteFile` after the `Lstat` check and modify the outside file.
- `fe5188b` moves contained directory creation and file writes onto Go 1.25 `os.Root` operations rooted at the trusted workspace. The existing deterministic symlink rejection remains, while the actual write/create operation can no longer follow a raced symlink outside the root.
- `e863546` also preserves administrator-provided tag case in `OPENCLAW_IMAGE_ALLOWED_TAGS`; repository entries remain normalized case-insensitively.
- `1a72c2c` adds regressions for the symlink-swap race and configured tag-case behavior, in addition to the existing three review-finding tests.

Independent sandbox checks against the pre-fix helper:
- static destination-dir/file symlink escape tests: pass
- implicit `:latest` and tag-case tests: pass
- new-container ancestor-symlink test: pass
- `go test -race` on the minimal static harness: pass
- symlink-swap stress regression: **fails on the pre-fix helper by modifying the outside file**, which is the gap addressed by `os.Root`.

The full PR workflow has been re-triggered on the new head. I could not run live OpenClaw provisioning in the independent sandbox because it has neither Docker nor Podman, so that E2E item remains explicitly unclaimed here.

zdy782:
Rebased onto current main and normalized the commit history in 4a84a6e:

- All 8 commits now carry the `Signed-off-by` trailer; the unsigned merge commit is gone (linear history), which should clear the DCO check.
- Rebased on top of latest main, which already carries the `pkg/fallback` / `pkg/llmprotocol` sources the dashboard image build needs; the Dockerfile `COPY` fix is retained on top.
- Fixed two errcheck lint findings on `rootFS.Close` in the `os.Root` containment helpers (amended into the same change).

Verified on x86_64 with Go 1.25: `go build ./...`, `go vet`, full `dashboard/backend` suite (handlers included, ~94s), and `golangci-lint run ./handlers/` is clean for the changed files.

PR #4007, opened 2026-09-21
[Feature] Add model_eval evaluation quality checks, cost-aware config generation, reasoning-mode eval

### LINKED ISSUE #4005: [Feature] model_eval evaluation quality checks, cost-aware config generation, reasoning-mode eval
### Goal

Add standalone CLI tools and additive options under `src/training/model_eval/` so that users can (1) check whether the two models have sufficient routing capability differences on a data set before spending GPU, (2) freeze train/test splits to prevent config overfitting, (3) evaluate multiple datasets concurrently, (4) predict signal labels without a running router, (5) calibrate cost-aware λ for routing configs, and (6) generate configs that reflect measured think-vs-nothink trade-offs.

The tools cover four stages of the config-generation lifecycle:

- Pre-evaluation: differentiation check + split freeze
- Evaluation: multi-dataset runner + MCQ judge
- Signal: direct classifier inference — load any HF sequence/token   classification model (intent, jailbreak, PII, etc.) and tag each question, without a running router
- Config generation: cost-aware λ-scan + cost-aware result_to_config

This is a follow-up to the `chat_template_kwargs` protocol pass-through (issue #3573, PR #3614) — the `--chat-template-kwargs` option here applies that protocol to the MMLU-Pro evaluator so thinking models can be evaluated in both think/no-think modes.

### Proposed Workgroup

Evaluation & Quality

### Scope

new standalone CLI tools + additive options to existing eval scripts, all under `src/training/model_eval/`. No changes to router Go code, CLI commands, or config schema.

### Non-goals

no router/CLI/config-schema changes; no CI wiring; no new datasets; no oracle framework.

### Completion signal

differentiation check emits go/no-go verdict; split freeze reproduces sha256; multi-dataset eval runs on 4 datasets;
MCQ judge produces analysis.json; signal prediction tags questions on CPU; cost calibration recommends λ; config generation produces full v0.3 config + optional patch mode. All modified files preserve upstream behavior when new flags are omitted.

### Additional context
I plan to open a PR with this feature.

PharosEast:
Closes #4005

## Purpose

Add standalone CLI tools and additive options under `src/training/model_eval/` so that users can check model differentiation, freeze splits, evaluate multiple datasets, predict signal labels, calibrate cost-aware λ, and generate configs reflecting measured think-vs-nothink trade-offs — all without running the router.

New files (6, all standalone):

- `differentiation_check.py` — go/no-go verdict on model differentiation (capability gap + saturation + mutual exclusivity rate)
- `split_evalset.py` — configurable-ratio train/test split (default 0.5) with sha256 manifest freeze
- `multi_benchmark_eval.py` — multi-dataset concurrent evaluation (openai lib, `--chat-template-kwargs` for reasoning-mode injection)
- `mcq_judge.py` — MCQ answer judging → judged.jsonl + analysis.json - `signal_predict.py` — direct HF sequence/token classification inference (no router, no config YAML; supports any classifier checkpoint)
- `cost_aware_calib.py` — λ-scan + recommendation + per-category token costs side-output

Modified files (3, additive only — defaults preserve upstream behavior):

- `result_to_config.py` — +`--mode patch` (merge-patch output),  +`--cost-lambda`, +`--token-costs-dir`, +`--nothink-results-dir`
- `signal_eval.py` — +`--custom-ground-truth` for external ground-truth  sets (`--dataset` relaxed to optional)
- `mmlu_pro_vllm_eval.py` — +`--chat-template-kwargs` for reasoning-mode  injection (follows up #3573 / #3614), +small-sample warning

All modifications are additive: when new flags are omitted, output is byte-identical to upstream.

## Test Plan

1. `py_compile` on all 9 files.
2. `differentiation_check.py` on 4B vs V4 (380 questions): verdict=GO,  gap=0.1316, n_common=380.
3. `split_evalset.py` reproduces frozen sha256 manifests; `--ratio 0.7`  produces 70/30 split.
4. `multi_benchmark_eval.py` runs MMLU/CMMLU/C-Eval/mmlu_pro.
5. `mcq_judge.py` produces SR-format analysis.json.
6. `signal_predict.py` tags 760 questions with domain labels on CPU.
7. `cost_aware_calib.py` recommends λ=0.1 (82.50%@136K vs 85.13%@835K).
8. `result_to_config.py --mode full` generates complete v0.3 config.
9. `result_to_config.py --mode patch` generates merge-patch YAML.
10. `result_to_config.py` (no new flags) produces identical output to upstream.
11. `mmlu_pro_vllm_eval.py --chat-template-kwargs` toggles think/no-think on a thinking model (acc 85.77% vs 82.48%, token ratio 33.2×).

## Test Result

- `py_compile` passes on all 9 files.
- All test cases verified locally (2026-09-21).
- `result_to_config.py` without new flags: byte-identical to upstream.


src/training/model_eval/result_to_config.py
```diff
@@ -357,12 +436,227 @@ def save_config(config, output_file):
...
+
+    norm_cost = cat_tokens / max_tokens_across_all_categories_for_this_model.
+    Returns 0.0 when no token cost data is available for the model/category.
+    """
+    if not token_costs or model_name not in token_costs:
+        return 0.0
+    cat_costs = token_costs[model_name]
+    if category_name not in cat_costs:
+        return 0.0
+    max_tok = max(cat_costs.values()) if cat_costs else 1.0
```
Xunzhuo: [P1] Use the calibration normalization when generating scores

The calibration scan normalizes cost across models within one category, but this helper normalizes across categories within one model. With one category, accuracies 0.90/0.85, token costs 1000/10, and lambda 0.1, calibration selects the cheap model (scores 0.80/0.849), while the generated config selects the expensive model (0.80/0.75). The recommended lambda therefore does not reproduce the measured policy or savings. Share the same per-category normalization between calibration and config generation and test their selected models end to end.

src/training/model_eval/result_to_config.py
```diff
@@ -357,12 +436,227 @@ def save_config(config, output_file):
...
+        for model_name, accuracy in ranked_models:
+            if cost_lambda > 0 and token_costs:
+                nc = _norm_cost(token_costs, model_name, category_name)
+                score = float(accuracy) - cost_lambda * nc
+            else:
+                score = float(accuracy)
+
+            if nothink_accuracies:
+                use_reasoning = _measured_use_reasoning(
+                    category_name, think_accs, nothink_accs,
```
Xunzhuo: [P2] Derive reasoning mode from the model being scored

This passes all model accuracies to the helper inside every model iteration, so the largest gap from any model becomes every model's use_reasoning flag. For A at think/nothink 0.90/0.40 and B at 0.80/0.90, both flags are true even though reasoning makes B worse. Compare the current model's matched think/nothink measurements, and derive the question-based threshold from actual sample counts rather than the number of models.

src/training/model_eval/split_evalset.py
```diff
@@ -0,0 +1,138 @@
...
+        train, eval_ = random_split(records, args.ratio, args.seed)
+        print(f"Random split: train={len(train)}, eval={len(eval_)}")
+
+    out_dir = Path(args.output_dir)
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    train_path = out_dir / f"{args.prefix}_train.jsonl"
+    eval_path = out_dir / f"{args.prefix}_eval.jsonl"
+
+    with open(train_path, "w") as f:
```
Xunzhuo, line 177: [P2] Check an existing freeze before replacing split artifacts

A second invocation with the same output paths and a different seed overwrites both splits and their manifest, exits successfully, and still prints that the eval set is FROZEN. The previous hashes are lost, so this workflow cannot detect that held-out examples have changed or moved into calibration. Verify an existing manifest before writing and reject changed inputs or split parameters unless the user explicitly requests a new split.

src/training/model_eval/signal_predict.py
```diff
@@ -0,0 +1,283 @@
...
+            batch, padding=True, truncation=True,
+            max_length=512, return_tensors="pt",
+            is_split_into_words=False,
+        ).to(device)
+        with torch.no_grad():
+            out = model(**inputs)
+            probs = torch.softmax(out.logits, dim=-1)
+            pred_ids = torch.argmax(probs, dim=-1).cpu().tolist()
+        for idx in range(len(batch)):
+            seq = pred_ids[idx]
```
Xunzhuo: [P2] Exclude padding from token-label aggregation

`seq` includes every padded position, although `inputs.attention_mask` marks those positions as invalid. A deterministic classifier returning O on the real token and B-PII on padding labels the same short text O when evaluated alone and B-PII when batched with a longer text. Its real-token predictions are unchanged. Filter padding (and special tokens) before aggregating labels and confidence so predictions do not depend on unrelated batch members.

src/training/model_eval/multi_benchmark_eval.py
```diff
@@ -0,0 +1,177 @@
...
+    # Resumable: skip already completed task_ids (unless --force)
+    done = set()
+    if args.force and os.path.exists(args.out):
+        os.remove(args.out)
+        print(f"[gen] --force: deleted existing {args.out}", flush=True)
+    if os.path.exists(args.out):
+        with open(args.out, encoding="utf-8") as f:
+            for line in f:
+                try:
+                    done.add(json.loads(line)["task_id"])
```
Xunzhuo: [P2] Keep failed generations eligible for resume

Every record is added to `done`, including records emitted after all generation retries fail with `error` set. With one successful and one failed record, restarting prints `done=2 todo=0` and never retries the failed task after the endpoint recovers. The transient failure is then retained as an unanswered benchmark task; `--force` discards all successful work. Resume only successful IDs and replace or deduplicate failed records when retrying them.

src/training/model_eval/differentiation_check.py
```diff
@@ -0,0 +1,141 @@
...
+    n = len(common_ids)
+    acc_a = (counts["both_correct"] + counts["a_only"]) / n
+    acc_b = (counts["both_correct"] + counts["b_only"]) / n
+    gap = abs(acc_a - acc_b)
+    # Mutual exclusivity rate: questions where exactly one model is right
+    excl = counts["a_only"] + counts["b_only"]
+    excl_rate = excl / n
+
+    # Saturation: both very high → no routing headroom
+    saturated = acc_a > 0.95 and acc_b > 0.95
```
Xunzhuo: [P2] Apply the configured saturation threshold

`--saturation-threshold` is only used in the output text; this comparison always uses 0.95. Running the CLI with both accuracies at 0.90, `--saturation-threshold 0.8` and `--gap-redline 0` reports GO and `saturated: false`, although both models exceed the requested threshold. Pass the configured value into `check_differentiation` and use it for this comparison.

src/training/model_eval/cost_aware_calib.py
```diff
@@ -0,0 +1,358 @@
...
+            cats[r["category"]].append(r.get("completion_tokens") or 0)
+        out[model_name] = {c: sum(v) / len(v) for c, v in cats.items()}
+    return out
+
+
+def write_token_costs(token_costs, out_dir):
+    """Write one JSON per model under out_dir (result_to_config.py format)."""
+    os.makedirs(out_dir, exist_ok=True)
+    for model, costs in token_costs.items():
+        path = os.path.join(out_dir, f"{model}.json")
```
Xunzhuo: [P2] Round-trip qualified model IDs in token-cost files

A model name such as `org/model` makes this write target `out_dir/org/model.json`, but only `out_dir` is created, so exporting token costs fails. Replacing the slash with an underscore alone does not solve the pipeline: the existing evaluator encodes model directories and `collect_model_accuracies` restores `org/model`, while `load_token_costs` reads flat filenames verbatim, leaving costs under a different key and silently disabling that model's penalty. Use one reversible model-ID representation across the writer and reader, and cover a qualified ID in the round-trip test.

Xunzhuo (changes requested):
Please fix the calibration-to-config mismatch and the inline evaluation/data-integrity issues before merging. The training contract suite and 61 focused regression tests pass, while targeted probes reproduce these gaps. Classifier and resume probes use deterministic stubs; real model inference was not run.

src/training/model_eval/result_to_config.py
PharosEast: Thank you! 
Now fixed:`_norm_cost` now normalizes per-category across models (`cost_m(b) / max_over_models(cost(b))` ), matching`cost_aware_calib.py scan()` . The recommended λ now reproduces the calibrated policy end-to-end. For the example case (0.90/0.85, 1000/10 tokens, λ=0.1), both the calibration scan and the generated config now pick the cheap model, so the λ recommended by`cost_aware_calib.py` reproduces the measured policy and savings in`result_to_config.py` .

src/training/model_eval/result_to_config.py
PharosEast: Thank you! 
Now fixed:`_measured_use_reasoning` now takes the current model's own think/nothink accuracy (not all models'). The threshold is a direct 2pp gap (`gap_min_pp=0.02` ), documented as ≈2 questions on a 100-question MMLU-Pro category, removing the`len(models)*4` estimate. A model where thinking hurts now gets`use_reasoning=false` .

src/training/model_eval/split_evalset.py
PharosEast: Thank you! 
Now fixed:`main()` now reads an existing manifest and verifies`sha256_input` ,`seed` ,`ratio` ,`stratify_key` match before writing. A mismatch is rejected with an error unless`--force` is passed. Added`sha256_input` to the manifest.

src/training/model_eval/signal_predict.py
PharosEast: Thank you! 
Now fixed:`predict_batch_token_cls` now filters by`inputs.attention_mask` before aggregating labels and confidence. A short text returns the same label whether evaluated alone or batched with a longer text.

src/training/model_eval/multi_benchmark_eval.py
PharosEast: Thank you! 
Now fixed:Resume now only skips records with`error is None` . Failed records remain in`todo` for retry. Existing failed records are pruned from the file on restart so retried tasks don't duplicate.

src/training/model_eval/differentiation_check.py
PharosEast: Thank you! 
Now fixed:`check_differentiation` takes`saturation_threshold` as a parameter;`main()` passes`args.saturation_threshold` . The configured value now drives the comparison, not just the output text.

src/training/model_eval/cost_aware_calib.py
PharosEast: Thank you! 
Now fixed:Added`_encode_model_id` /`_decode_model_id` (reversible`/` ↔`__slash__` ).`write_token_costs` encodes on write;`load_token_costs` (in`result_to_config.py` ) decodes on read.`org/model` now round-trips correctly.

src/training/model_eval/result_to_config.py
```diff
@@ -355,12 +436,231 @@ def save_config(config, output_file):
...
+                {
+                    "model": model_name,
+                    "score": round(score, 6),
+                    "use_reasoning": use_reasoning,
+                }
+            )
+
+        # When cost_lambda > 0, re-rank model_scores by the new score
+        if cost_lambda > 0 and token_costs:
+            model_scores.sort(key=lambda m: (-m["score"], m["model"]))
```
Xunzhuo: [P2] Preserve the calibrated tie-break policy in generated scores

The normalization example is fixed, but calibration breaks equal scores by lower token cost while this path breaks them by model name. With 64 questions, `a-expensive` at 56 correct/100 tokens and `z-cheap` at 55 correct/87.5 tokens, the scan recommends lambda 0.125 within the default 3pp tolerance and reports 12.5% savings. Both generated scores are 0.75, with the expensive model ranked first. The static selector keeps the first candidate at equal score, so an existing decision listing the expensive model first loses those savings. Preserve the calibrated tie-break through the generated routing contract and test the selected model end to end.

src/training/model_eval/split_evalset.py
```diff
@@ -0,0 +1,207 @@
...
+        prev_ratio = prev.get("ratio")
+        prev_stratify = prev.get("stratify_key")
+        mismatches = []
+        if prev_input is not None and prev_input != input_hash:
+            mismatches.append("input file changed")
+        if prev_seed is not None and prev_seed != args.seed:
+            mismatches.append(f"seed {prev_seed} -> {args.seed}")
+        if prev_ratio is not None and prev_ratio != args.ratio:
+            mismatches.append(f"ratio {prev_ratio} -> {args.ratio}")
+        if prev_stratify is not None and prev_stratify != args.stratify_key:
```
Xunzhuo: [P2] Treat an unstratified manifest as a frozen parameter value

A manifest produced by the default random split contains `stratify_key: null`. This guard skips that value, so rerunning the same 30 records (three source groups), seed and output directory with `--stratify-key source` succeeds without `--force`, changes the eval hash, and prints that the rewritten artifacts are content-identical and frozen. Compare the recorded null value with the requested key instead of treating it as missing, and reject this transition before writing either split.

src/training/model_eval/cost_aware_calib.py
```diff
@@ -0,0 +1,403 @@
...
+
+def _encode_model_id(model_name):
+    """Encode a (possibly slash-bearing) model id into a flat filename.
+
+    Reversible with ``_decode_model_id``. Used by write_token_costs and
+    load_token_costs (in result_to_config.py) so a qualified id like
+    ``org/model`` round-trips through the token-cost file and is looked
+    up under the same key at config-generation time.
+    """
+    return model_name.replace("/", "__slash__")
```
Xunzhuo: [P2] Escape the replacement marker in model IDs

The new filename encoding maps both `org/model` and the distinct served-model name `org__slash__model` to `org__slash__model.json`. Writing costs 12 and 99 for those models leaves one file; loading it returns only `org/model: 99`, silently overwriting one model and assigning the other model's cost to it. Use an injective encoding (or store and validate the original ID separately) and cover IDs containing the escape marker as well as slashes.

Xunzhuo (changes requested):
The original normalization, per-model reasoning, changed-seed, padding, resume, threshold and qualified-ID examples now pass. Three distinct cases still need fixes: calibration ties lose the measured savings, changing an unstratified frozen split overwrites held-out data, and the filename escape aliases distinct model IDs. Deterministic checks reproduce all three on the current source; live model inference and full router selection were not run. Please fix these cases and add regressions. The special-token portion of the earlier masking comment also remains open.

src/training/model_eval/result_to_config.py
PharosEast: Thank you!
Now Fixed: Tie-break changed from alphabetical (`m["model"]` ) to lower normalized token cost (`_norm_cost` ), matching the`scan()` policy so equal-score models keep the cheaper one first.

src/training/model_eval/split_evalset.py
PharosEast: Thank you!
Now Fixed: Removed the`is not None` guard; a recorded`stratify_key: null` is now compared directly with the requested key, so transitioning from random to stratified split without`--force` is rejected.

src/training/model_eval/cost_aware_calib.py
PharosEast: Thank you!
Now Fixed: Replaced`__slash__` string replacement with`urllib.parse.quote` /`unquote` (injective).`org/model` →`org%2Fmodel` and`org__slash__model` →`org__slash__model` now map to distinct filenames, eliminating the collision.

PharosEast:
@Xunzhuo 
Hello Xunzhuo, all checks have passed, please review, thanks !

Issue #3971, opened 2026-09-20
[Docs] FAQ, VSR vs gateways/llm-d, value metrics, and debugging


AayushSaini101:
## Summary
Add a short FAQ on the Semantic Router website for common customer/user questions about what VSR is, how it differs from AI gateways and llm-d, and how to measure and debug routing.

## Questions to cover

1. **What is vLLM Semantic Router?**  
   Is it an AI gateway, or something else? Clear positioning: content/policy-aware MoM control plane (“what” to run), not just L7 load balancing or provider fan-out.

2. **How do we measure routing accuracy and business value?**  
   Beyond classifier metrics — how to show a routing recipe keeps answer quality while cutting cost/latency vs a strong single-model baseline (evals, A/B or shadow traffic, cost + latency).

3. **If llm-d already does multi-model routing, why VSR?**  
   Boundary: VSR = intent/policy/MoM decisions (“what”); llm-d = pool/endpoint scheduling, load, prefix-cache (“which/how/where”).

4. **How do VSR and llm-d Endpoint Picker avoid conflicts?**  
   Priority when logical model choice and cache/load locality disagree; how the final decision is made visible (headers, traces, replay).

5. **How do we avoid hurting multi-turn / agentic workloads?**  
   When model switching is worth the cost (prefix-cache loss, tool drift, continuity); stickiness / escalate-only style policies.

6. **How do operators debug a bad outcome?**  
   Checklist: signal → policy → model selection → endpoint scheduling → fallback → final model; link to observability surfaces.

## Acceptance criteria
- [ ] FAQ page/section with clear answers to the above
- [ ] Simple table or diagram for VSR ↔ llm-d boundary
- [ ] Links to architecture, llm-d integration, and observability docs

PR #4008, opened 2026-09-21
[Docs] Add a short FAQ covering positioning, llm-d boundary, value metrics, and debugging

Zi-Yi-Ming:
Related #3971

## Purpose

Evaluating Semantic Router raises the same six questions repeatedly, and the answers are spread across five pages: what it is relative to an AI gateway, how to measure routing value, where the llm-d boundary sits, how the two systems avoid conflicting decisions, what model switching costs multi-turn workloads, and how to debug a bad route.

This adds a short FAQ page that answers each one in a few sentences and links to the page carrying the detail, rather than restating it. Two reasons for that shape: the linked pages change weekly, and a duplicated answer becomes wrong silently. The VSR versus llm-d table follows the split the llm-d integration page already states — Semantic Router picks the logical model or pool, llm-d picks a healthy replica inside it — so the FAQ does not introduce a second framing of the same boundary.

## Test Plan

Verified without a full Docusaurus build, which could not complete locally (the website dependency tree is large and the install did not finish in a bounded time):

- the frontmatter parses as YAML with the keys the sibling pages use (`sidebar_position`, `title`, `description`);
- every outbound link resolves to a page that exists on `main`;
- the quoted rule in the llm-d answer matches the integration page character for character;
- the table vocabulary matches the same page (`what` versus `which / how / where`);
- the file contains no raw HTML or MDX-only syntax, so it renders as plain Markdown.

## Test Result

The page is registered in `sidebars.ts` between the Overview category and Getting Started, following the explicit ordering that file uses.

Not verified here: the full `npm run build`, for the reason above — the four generated-asset checks (`generated:check`, `sync:install`, `contributors:rank`, `committers:activity`) and the Docusaurus render were not run to completion locally. Netlify preview covers the render on the pull request.

Not covered here: the zh-Hans translation, which is left to the i18n pass tracked by #3074 — the translation-staleness check reports untranslated pages without failing. No runtime behavior changes, and no configuration surface is added.


AayushSaini101:
@TwilightTechie @abhinav-m22 PTAL thanks

website/docs/faq.md
```diff
@@ -0,0 +1,119 @@
...
+
+## How do we avoid hurting multi-turn / agentic workloads?
+
+Switching models mid-conversation has a real cost — prefix-cache loss, tool drift and
+broken continuity. Treat it as a deliberate trade rather than an automatic
+optimization:
+
+- **Prefer stickiness where continuity matters.** Session-scoped learning protects an
+  established choice by default (built-in initialization defaults to protection on and
+  online adaptation off), so send a stable session or conversation identity for it to hold.
```
Xunzhuo: [P2] Require both identities for default conversation protection

The built-in initialization described here uses the default `scope: conversation`. `protectionIdentity()` requires both `x-session-id` and `x-conversation-id`; sending either one alone produces `missing_identity`, so requests continue without retaining the model. Please document both headers here, reserving session-only guidance for explicit `scope: session`, and link to the session-identification reference.

Xunzhuo (commented):
The FAQ has one setup detail to correct: default conversation protection requires both identity headers. Otherwise readers can follow this guidance and still switch models between turns. Translation coverage and Markdown/MDX checks pass; a full Docusaurus build and live routing were not run.

website/docs/faq.md
Zi-Yi-Ming: Fixed — the default `scope: conversation` now documents **both** headers, notes that sending either alone leaves the request without a retained model, and reserves session-only guidance for an explicit `scope: session`. Linked [Session identification](https://github.com/vllm-project/semantic-router/blob/main/website/docs/api/session-identification.md).

Verified against `protectionIdentity` in `src/semantic-router/pkg/extproc/router_learning_protection.go` (the `missing_identity` path) and the header table in `config/config.yaml`.

website/docs/faq.md
```diff
@@ -0,0 +1,122 @@
...
+`100 × (1 − candidate cost / baseline cost)`. Two disciplines keep the number honest:
+
+- the baseline is the best observed single **on the same cases**, not a per-question
+  oracle, and quality ties resolve to the lower total cost;
+- coverage is reported with the score — missing or ungraded results are not zeros, and
+  a small dev win or a zero observed difference does not establish equivalence.
+
+→ [sr-bench 1.0](benchmarking/sr-bench)
+
+## If llm-d already does multi-model routing, why VSR?
```
abhinav-m22: `VSR` isn't used in prose anywhere else in `website/docs`. The only other hits are the `x-vsr-*` header names. These are headings, so they become anchors and search results, and this is the page a first-time evaluator lands on. Worth spelling out:

- "If llm-d already does multi-model routing, why Semantic Router?"
- "How do Semantic Router and the llm-d Endpoint Picker avoid conflicts?"

website/docs/faq.md
```diff
@@ -0,0 +1,122 @@
...
+
+```
+signal → policy / decision → algorithm / model → plugin → endpoint scheduling → fallback → final model
+```
+
+1. **Preview before you generate.** `route preview` evaluates signals, projections and
+   the decision without calling a backend, and reports the trace evidence.
+2. **Read the provenance.** Learning-enabled preview exposes `selection_provenance`,
+   including the config and state identity and the sampling seed.
+3. **Then probe for real.** `route probe` sends an actual request through Envoy — but a
```
abhinav-m22: Every other page writes these as `vllm-sr route preview` and `vllm-sr route probe`. On a page for people who haven't installed anything yet, the bare form doesn't tell them which binary to run.

website/docs/faq.md
```diff
@@ -0,0 +1,122 @@
...
+   successful HTTP status is not sufficient: check `response.body.delivery` and the
+   final assistant output, because empty output, reasoning-only output and
+   `finish_reason: length` all fail delivery.
+4. **Check the UI path separately.** Exercise the same entrypoint in the Dashboard and
+   report API-only coverage apart from UI coverage.
+
+Delivery, route correctness and answer quality are three separate outcomes; qualify the
+ones you did not measure instead of inferring success.
+
+→ [API and Observability](tutorials/global/api-and-observability)
```
abhinav-m22: The steps above describe the CLI (`api/cli.md`) and the preview API (`api/apiserver.md`), but the only link is the global telemetry config page. `troubleshooting/vsr-headers` is also the natural place to send someone who wants to read the decision off the response. Suggestion:

`→ [CLI reference](api/cli) · [VSR headers](troubleshooting/vsr-headers) · [API and Observability](tutorials/global/api-and-observability)`

website/docs/faq.md
```diff
@@ -0,0 +1,122 @@
...
+resolves the pool first; llm-d then picks a replica inside it. Because the two
+decisions live at different layers, a disagreement between "the best model for this
+request" and "the best replica for cache locality" resolves itself — the first decision
+constrains the second.
+
+Two operational rules follow from that shape:
+
+- **Deploy and verify llm-d independently before adding Semantic Router**, and use one
+  supported llm-d release rather than mixing copied manifests.
+- **Make the final decision visible.** Record it through headers, traces and replay
```
abhinav-m22: This is the part of the original question that most needs a concrete answer, and it doesn't name a header. `x-selected-model` drives the `HTTPRoute` match on the llm-d page, and `x-vsr-selected-model` is the Router's receipt (`route probe --expect-selected-model` asserts it). Naming them here gives an operator something they can check.

AayushSaini101:
@Zi-Yi-Ming we need a comparision section like this in the website page as well like this:
image

website/docs/faq.md
Zi-Yi-Ming: Fixed — both headings now spell out Semantic Router. Confirmed the abbreviation appears nowhere else in `website/docs` prose.

website/docs/faq.md
Zi-Yi-Ming: Fixed — both invocations now read `vllm-sr route preview` and `vllm-sr route probe`.

website/docs/faq.md
Zi-Yi-Ming: Fixed — the closing line now links the CLI reference and the header reference alongside the observability page, as you suggested.

website/docs/faq.md
Zi-Yi-Ming: Fixed — the answer now names `x-vsr-selected-model` (the Router's receipt), `x-selected-model` (the value a gateway matches on), and `vllm-sr route probe --expect-selected-model`.

Zi-Yi-Ming:
@AayushSaini101 thanks — happy to add the comparison section. One question on the shape before I write it, because the repo has two different patterns here:

- **Docs pages use plain Markdown tables** throughout. The closest existing example is the "What it owns" table in `website/docs/installation/k8s/gateways.md`, which pairs an integration with the responsibility it carries.
- **The screenshot matches the homepage capability matrix** instead: `architectureMatrix` in `website/src/pages/index.tsx`, which is a React component with CSS Modules, a reveal animation and `<Translate>` keys.

A plain Markdown table on this page is a small diff and matches every other docs page. The styled version needs a new component and CSS under `website/src`, so it is a larger change than this page and probably belongs in its own PR.

Which would you like? If the intent is "make the boundary readable", the Markdown table should cover it; if it is "match the homepage visual language", I can scope the component separately.

AayushSaini101:
> @AayushSaini101 thanks — happy to add the comparison section. One question on the shape before I write it, because the repo has two different patterns here:
> 
> * **Docs pages use plain Markdown tables** throughout. The closest existing example is the "What it owns" table in `website/docs/installation/k8s/gateways.md`, which pairs an integration with the responsibility it carries.
> * **The screenshot matches the homepage capability matrix** instead: `architectureMatrix` in `website/src/pages/index.tsx`, which is a React component with CSS Modules, a reveal animation and `<Translate>` keys.
> 
> A plain Markdown table on this page is a small diff and matches every other docs page. The styled version needs a new component and CSS under `website/src`, so it is a larger change than this page and probably belongs in its own PR.
> 
> Which would you like? If the intent is "make the boundary readable", the Markdown table should cover it; if it is "match the homepage visual language", I can scope the component separately.

thanks for the explanation let's create another issue for considering this thanks : )

AayushSaini101 (approved):

wilsonwu (approved):
LGTM

Xunzhuo (approved):
The revised FAQ matches the session identity, routing header, and probe contracts. Documentation checks pass, and the previous identity guidance concern is resolved.

AayushSaini101:
@Zi-Yi-Ming can you open another issue for comparison diff on the main website thanks : )

Zi-Yi-Ming:
> @Zi-Yi-Ming can you open another issue for comparison diff on the main website thanks : )

Opened #4033 for the homepage comparison — reuses our own matrix styling and
compares Semantic Router / AI gateway / llm-d by responsibility. Let me know if
you want different rows or columns.

PR #3770, opened 2026-09-12
[Router] add cross-model KV transfer wire headers (#2976)

AayushSaini101:
Related: #2976 

## Summary

Adds step 1 of epic #2976: the shared wire contract for cross-model KV-cache reuse.

- Define request-side hint headers injected by the KVTransfer Coordinator on model switch:
  - `x-vsr-kv-source-pod` — source pod gRPC address
  - `x-vsr-kv-cache-id` — opaque session/cache identifier
  - `x-vsr-kv-mapper-id` — ridge-mapper artifact id (e.g. `qwen3-14b-32b-v1`)
- Define response status header from the target vLLM KVConnector plugin:
  - `x-vsr-kv-transfer-status` — `applied`, `fallback_reprefill`, or `unsupported`
- Add Go constants, status value constants, and unit tests in `pkg/headers/`.
- Document the headers in `website/docs/troubleshooting/vsr-headers.md`.

No runtime behavior changes yet — coordinator injection, registry write, and plugin consumption are follow-up steps (2–3).

Closes step I1 of the #2976 execution plan.

## Test plan

- [x] `go test ./pkg/headers/...`
- [ ] Manual: confirm no headers are emitted yet (constants only)
- [ ] Follow-up: wire coordinator to inject request headers (step 2–3)

PR #4009, opened 2026-09-21
[#2976] Add KV mapper artifact format under src/training/kv_mapper

ima-helikoptaaa:
Related #2976

## Purpose

Cross-model KV transfer (#2976, wg/mom-routing) already has wire headers
(`x-vsr-kv-mapper-id` and related constants from #3770). This PR defines
what that id names.

CPU-only Python package under `src/training/kv_mapper/`, same tree as the
other artifact families (classifiers, embeddings, eval):

- directory layout: `manifest.json`, `weights.safetensors`, `SHA256SUMS`
- tensor names `target.{layer}.{k|v}.{W|b}`
- `verify_compatibility()` against a deployment spec (model ids, HF
  revisions, variant, precision, TP, head order, n_kv_heads, head_dim)
- `make_mapper_id()` that pins HF weight revisions, dtype, and KV head
  count, not a routing alias (Adam's #2976 / #3792 note)

The Go router names the artifact in headers. Apply is a vLLM KVConnector.
Collect/fit/eval entrypoints and Hub publish follow in later PRs.

## Test Plan

```bash
cd src/training/kv_mapper
cd ../../..
```

Also covered by `make test-training-contracts` once numpy and safetensors
are available in that venv.

Covers write/read roundtrip, SHA256 mismatch, each compatibility field,
mapper id contents (revisions, dtype, n_kv_heads), and the training-tree
layout check.

## Test Result

```
Ran 4 tests in 0.052s
OK
```

Layout test `src.training.tests.test_kv_mapper_family` also OK.


src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+        if path.is_file():
+            lines.append(f"{_sha256_file(path)}  {name}")
+    (out_dir / CHECKSUMS_FILE).write_text("\n".join(lines) + "\n")
+
+
+def verify_checksums(out_dir: Path) -> None:
+    sums_path = out_dir / CHECKSUMS_FILE
+    if not sums_path.is_file():
+        raise FileNotFoundError(f"missing {sums_path}")
+    for line in sums_path.read_text().splitlines():
```

Xunzhuo (changes requested):
Please fix the artifact integrity check before merging.

Validation: all 4 KV mapper tests and the training-family layout test pass in a CPU environment; Training Contracts CI also reports all 660 cases passing, including the new tests. Additional small regressions reproduce the inline findings. Model-backed fitting and end-to-end KV reuse were not run.

src/training/kv_mapper/mapper_id.py
```diff
@@ -0,0 +1,47 @@
...
+    n_kv_heads: int,
+    bundle_version: int = 1,
+) -> str:
+    """Build a config-facing mapper id.
+
+    Pins the Hugging Face weight revisions the mapper was fitted on (not a routing
+    alias), plus dtype and KV head count. ``bundle_version`` bumps when the same
+    revisions are re-fitted with a new recipe; it is not a model revision.
+    """
+    prec = precision.lower().replace("float", "fp")
```
yaojiejia, line 36: `replace("float", "fp")` is a substring match, so `bfloat16` becomes `bfp16`, which is not a dtype name anyone uses (`bf16` is the short form). Since bf16 is the common production dtype, most real ids would carry the bad token. 

Suggest a small explicit alias table instead, and raise on anything unknown so typos fail at fit time rather than becoming a permanent id component:

```python
_PRECISION_ALIASES = {
    "fp16": "fp16", "float16": "fp16", "half": "fp16",
    "bf16": "bf16", "bfloat16": "bf16",
    "fp32": "fp32", "float32": "fp32",
}

```

src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+def verify_compatibility(manifest: Manifest, deployment: CompatibilitySpec) -> None:
+    """Refuse load when deployment layout does not match the fitted artifact."""
+    fitted = manifest.compatibility
+    checks: list[tuple[str, Any, Any]] = [
+        ("source_model", fitted.source_model, deployment.source_model),
+        ("source_revision", fitted.source_revision, deployment.source_revision),
+        ("target_model", fitted.target_model, deployment.target_model),
+        ("target_revision", fitted.target_revision, deployment.target_revision),
+        ("variant", fitted.variant, deployment.variant),
+        ("precision", fitted.precision, deployment.precision),
```
yaojiejia, line 118: This compares raw `precision` strings, but `make_mapper_id` normalizes them. A manifest with `float16` and a deployment spec with `fp16` get a matching id but fail here with `precision: artifact='float16' deployment='fp16'`.

Suggest normalizing in `CompatibilitySpec.__post_init__` with the same helper `make_mapper_id` uses, so both sides always hold the canonical token.

yaojiejia (changes requested):

PR #4011, opened 2026-09-21
[Bug] Stabilize vector store pagination for equal creation timestamps

### LINKED ISSUE #4003: [Bug] Vector store pagination is unstable when stores share the same creation timestamp
### Proposed Workgroup
Data Plane & Networking

### Describe the bug

Vector store cursor pagination can return duplicate or missing stores when multiple stores have the same `created_at` second.

`CreateStore` records `CreatedAt` with `time.Now().Unix()`. `Manager.ListStores` then collects stores by ranging over an in-memory Go map and sorts only by `CreatedAt`. When timestamps tie, the map-derived order is not deterministic.

Because `after` and `before` pagination use the cursor ID's position in that sorted slice, repeated list requests can use different page boundaries.

### To Reproduce

1. Start the router with the memory vector-store backend.
2. Create at least three vector stores within one second.
3. Request:

   ```
   GET /api/v1/storage/vector-stores?limit=2
   ```

4. Take the ID of the last store in the first page.
5. Request the next page:

   ```
   GET /api/v1/storage/vector-stores?limit=2&after=<last-store-id>
   ```

6. Repeat the two list requests several times and compare the page contents.
7. Observe that the relative order of stores with the same creation timestamp can change. The second page may contain a store already returned on the first page, or may omit a store that should be returned.

For example, the first page may return `[A, B]`, while the next request after `B` may return `[C, A]` or omit an expected store.

### Expected behavior

For an unchanged set of vector stores, list requests with the same `order` should use one deterministic ordering.

Traversing all pages with an `after` cursor should return every store exactly once. The corresponding `before` pagination should use the same deterministic ordering in the opposite direction.

### Impact

Clients that enumerate or reconcile vector stores cannot reliably use cursor pagination. They may process a store twice, skip a store, or incorrectly conclude that there are no more stores.

There is no reliable workaround for collections that require multiple pages.

### Additional context

Affected code:

- `src/semantic-router/pkg/vectorstore/manager.go`, `CreateStore` timestamp assignment
- `src/semantic-router/pkg/vectorstore/manager.go`, `ListStores` sorting and cursor pagination
- `src/semantic-router/pkg/apiserver/route_vectorstore.go`
- `src/semantic-router/pkg/apiserver/route_vectorstore_request.go`

The existing listing tests manually assign unique creation timestamps (`1` through `5`), so they do not cover the real same-second creation case.

A minimal fix appears to be sorting by `(CreatedAt, ID)`, reversing both comparisons for ascending order. Using `sort.SliceStable` alone would not fix the issue because the input order still comes from a Go map.

The fix should remain limited to deterministic ordering and focused regression tests. Changing timestamp precision, cursor format, persistence schema, or concurrent snapshot semantics is outside this issue.

Lcos-000:
Closes #4003

## Purpose

Vector store pagination used second-resolution `CreatedAt` as its only sort key. When multiple stores were created within the same second, `ListStores` copied entries from a Go map whose iteration order is nondeterministic. Cursor pagination using `after` or `before` could therefore repeat or omit stores.

Keep `CreatedAt` as the primary sort key and use `ID` as a deterministic tie-breaker for both ascending and descending order.

The change is limited to vector store ordering and regression tests. It does not change cursor formats, timestamp precision, schemas, or unrelated storage behavior.

Assisted-by: Codex GPT-5

## Test Plan

- `gofmt` and `git diff --check`
- `go vet ./pkg/vectorstore ./pkg/apiserver`
- Focused manager and API pagination tests
- Pure-Go package tests and race tests
- `make check` with the three changed files
- `make harness-check`
- RISC-V compile-only check for the vectorstore test binary

## Test Result

- `gofmt`, `git diff --check`, `go vet`, pre-commit, architecture checks, and golangci-lint passed.
- Pure-Go package tests and race tests passed.
- Harness registry, workflow validation, 380 harness tests, and 2/2 E2E framework unit tests passed.
- No unrelated files were modified.

---
<details>

- [x] Affected modules are described without stacked title prefixes
- [x] Linked issue is accepted and owned by `wg/data-plane-networking`
- [ ] Commits are signed off with `git commit -s`
- [x] Purpose, Test Plan, and Test Result reflect the scope and remaining blockers

</details>


Lcos-000:
it seems not the false on my changes

yaojiejia:
> it seems not the false on my changes

Yeah this is just a flaky CI failure. The Helm dependency download hit a 504 from GitHub Releases so your tests never actually ran. I've rerun the job and we'll see if it goes green.

Lcos-000:
it just like thtat all checks have passed

FAUST-BENCHOU (approved):
lg

wilsonwu (approved):
LGTM

PR #4012, opened 2026-09-21
[#2976] Fit full-head ridge mappers into the kv_mapper artifact

ima-helikoptaaa:
Related #2976

Depends on #4009 (A1 artifact format). Stacked: this branch is A1 plus the fit commit. After #4009 merges, this rebases to the fit commit only.

## Purpose

Cross-model KV transfer (#2976, wg/mom-routing) needs a fitter that writes
the mapper directory defined in #4009.

This PR adds the recipe under `src/training/kv_mapper/`:

- Pearson correlation to pick top-k source layers per target layer
- centered ridge with a bias term, keys and values separately
- `write_fitted_artifact()` so the solve lands as `manifest.json`,
  `weights.safetensors`, and `SHA256SUMS` with tensor names
  `target.{layer}.{k|v}.{W|b}`
- top-k, ridge alpha, and source-layer lists recorded in the manifest

GPU collection stays a later PR. This slice is the CPU fit + A1 write path.

## Test Plan

```bash
cd src/training/kv_mapper
cd ../../..
```

Covers recovering a known linear map, Pearson top-k, empty ridge error, and
a synthetic three-layer fit that round-trips through `read_artifact`.

## Test Result

```
Ran 8 tests in 0.148s
OK
```


src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+        if path.is_file():
+            lines.append(f"{_sha256_file(path)}  {name}")
+    (out_dir / CHECKSUMS_FILE).write_text("\n".join(lines) + "\n")
+
+
+def verify_checksums(out_dir: Path) -> None:
+    sums_path = out_dir / CHECKSUMS_FILE
+    if not sums_path.is_file():
+        raise FileNotFoundError(f"missing {sums_path}")
+    for line in sums_path.read_text().splitlines():
```

Xunzhuo (changes requested):
The fit tests pass, but this stacked branch still inherits the artifact integrity defect from #4009; please carry that fix into this branch before merging.

Validation: all 8 KV mapper tests and the training-family layout test pass in a CPU environment; Training Contracts CI also reports all 664 cases passing, including the new tests. Additional small regressions reproduce the inline findings. Model-backed fitting and end-to-end KV reuse were not run.

src/training/kv_mapper/mapper_id.py
```diff
@@ -0,0 +1,47 @@
...
+    n_kv_heads: int,
+    bundle_version: int = 1,
+) -> str:
+    """Build a config-facing mapper id.
+
+    Pins the Hugging Face weight revisions the mapper was fitted on (not a routing
+    alias), plus dtype and KV head count. ``bundle_version`` bumps when the same
+    revisions are re-fitted with a new recipe; it is not a model revision.
+    """
+    prec = precision.lower().replace("float", "fp")
```
yaojiejia, line 36: pls check this review for the same reason https://github.com/vllm-project/semantic-router/pull/4009#discussion_r4068040197

src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+def verify_compatibility(manifest: Manifest, deployment: CompatibilitySpec) -> None:
+    """Refuse load when deployment layout does not match the fitted artifact."""
+    fitted = manifest.compatibility
+    checks: list[tuple[str, Any, Any]] = [
+        ("source_model", fitted.source_model, deployment.source_model),
+        ("source_revision", fitted.source_revision, deployment.source_revision),
+        ("target_model", fitted.target_model, deployment.target_model),
+        ("target_revision", fitted.target_revision, deployment.target_revision),
+        ("variant", fitted.variant, deployment.variant),
+        ("precision", fitted.precision, deployment.precision),
```
yaojiejia, line 118: pls check this review for the same reason https://github.com/vllm-project/semantic-router/pull/4009#discussion_r4068052401

yaojiejia (changes requested):

PR #4013, opened 2026-09-21
[#2976] Collect pre-RoPE KV activations for mapper fitting

ima-helikoptaaa:
Related #2976

Stacked on #4012 (fit) and #4009 (artifact format). This branch is A1 + A3 + the collect commit. After those merge, this rebases to collect only.

## Purpose

Cross-model KV transfer (#2976, wg/mom-routing) needs a collect path that
records the same HF revisions the mapper id pins, and that captures K after
`k_norm` (pre-RoPE) and V after `v_proj`.

This PR adds that under `src/training/kv_mapper/`:

- `collect.py`: window stride, layer-subset (`0:8` / `1,3,5` / `all`),
  `run.json` metadata (`source_revision`, `target_revision`,
  `rope_stripped_on_keys`), numpy `as_bshd` layouts
- `hooks.py`: forward hooks on `k_norm`/`k_proj` and `v_proj`
- `collect_run.py`: load source then target, write `run.json` +
  `activations.pt`

CI stays numpy-only (`make test-training-contracts`). `collect_run.py` is
the GPU entrypoint (torch + transformers).

## Test Plan

```bash
cd src/training/kv_mapper
cd ../../..
PYTHONPATH=. python3 -m unittest src.training.tests.test_kv_mapper_family
```

Covers stride defaults, BSHD/BHSD/flat layouts, layer-subset parsing, and
`run.json` round-trip of HF revisions. Torch `as_bshd` runs when torch is
installed (`tests/test_hooks.py`).

## Test Result

```
Ran 13 tests in 2.819s
OK

Ran 1 test in 0.000s
OK
```


src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+        if path.is_file():
+            lines.append(f"{_sha256_file(path)}  {name}")
+    (out_dir / CHECKSUMS_FILE).write_text("\n".join(lines) + "\n")
+
+
+def verify_checksums(out_dir: Path) -> None:
+    sums_path = out_dir / CHECKSUMS_FILE
+    if not sums_path.is_file():
+        raise FileNotFoundError(f"missing {sums_path}")
+    for line in sums_path.read_text().splitlines():
```

src/training/kv_mapper/collect_run.py
```diff
@@ -0,0 +1,140 @@
...
+    n_src = len(src_model.model.layers)
+    src_layers = parse_layer_subset(args.source_layer_subset, n_src)
+    ids = tok(
+        args.prompt,
+        add_special_tokens=False,
+        truncation=True,
+        max_length=args.seq_len,
+        return_tensors="pt",
+    )["input_ids"].to(args.source_device)
+    src_kv = capture_kv(
```
Xunzhuo, line 82: [P2] Make the collection options drive the captured samples

This path tokenizes args.prompt once and captures one sequence per model; corpus, dataset_config, seed, stride and num_sequences only reach run.json. For example, --num-sequences 8 --stride 4 still writes one prompt capture while the metadata identifies eight FineWeb sequences. That makes the saved calibration recipe inaccurate and prevents this entrypoint from collecting the requested training data. Build and capture the requested corpus windows, or explicitly restrict this command to a single-prompt mode and record its actual source and sample count.

Xunzhuo (changes requested):
Please make the collection output match the requested calibration recipe and fix the inherited artifact integrity check.

Validation: all 13 KV mapper tests and the training-family layout test pass in a CPU environment; Training Contracts CI also reports all 669 cases passing, including the new tests. Additional small regressions reproduce the inline findings. GPU collection with real model weights, teacher-forced injection, and end-to-end KV reuse were not run.

PR #4014, opened 2026-09-21
[#2976] Add paired bootstrap CIs for mapper eval

ima-helikoptaaa:
Related #2976

Stacked on #4013 (collect), #4012 (fit), and #4009 (format). This branch
includes those commits until they merge.

## Purpose

Cross-model KV transfer (#2976, wg/mom-routing) needs paired uncertainty on
the same held-out items before a mapper is called qualified (Adam on #2976).

This PR adds the CI path under `src/training/kv_mapper/`:

- `eval.py`: KV-space rel_err / cosine / R², per-item deltas, bootstrap mean CI
- `eval_report.py`: read an `arms` JSON dump, write a report vs a reference arm
  (default `cold`)
- equal-length check so mixed item sets cannot silently average

Inject / HellaSwag / CoQA still run on the GPU box. Their per-item scores
feed this report. Publish of the weight files is A5.

## Test Plan

```bash
cd src/training/kv_mapper
cd ../../..
PYTHONPATH=. python3 -m unittest src.training.tests.test_kv_mapper_family
```

Covers a known constant offset, length mismatch, report shape for inject-style
arms, and the CLI writing `report.json`.

## Test Result

```
Ran 18 tests in 1.333s
OK

Ran 1 test in 0.000s
OK
```


src/training/kv_mapper/artifact.py
```diff
@@ -0,0 +1,164 @@
...
+        if path.is_file():
+            lines.append(f"{_sha256_file(path)}  {name}")
+    (out_dir / CHECKSUMS_FILE).write_text("\n".join(lines) + "\n")
+
+
+def verify_checksums(out_dir: Path) -> None:
+    sums_path = out_dir / CHECKSUMS_FILE
+    if not sums_path.is_file():
+        raise FileNotFoundError(f"missing {sums_path}")
+    for line in sums_path.read_text().splitlines():
```

src/training/kv_mapper/collect_run.py
```diff
@@ -0,0 +1,140 @@
...
+    n_src = len(src_model.model.layers)
+    src_layers = parse_layer_subset(args.source_layer_subset, n_src)
+    ids = tok(
+        args.prompt,
+        add_special_tokens=False,
+        truncation=True,
+        max_length=args.seq_len,
+        return_tensors="pt",
+    )["input_ids"].to(args.source_device)
+    src_kv = capture_kv(
```
Xunzhuo, line 82: [P2] Make the collection options drive the captured samples

This path tokenizes args.prompt once and captures one sequence per model; corpus, dataset_config, seed, stride and num_sequences only reach run.json. For example, --num-sequences 8 --stride 4 still writes one prompt capture while the metadata identifies eight FineWeb sequences. That makes the saved calibration recipe inaccurate and prevents this entrypoint from collecting the requested training data. Build and capture the requested corpus windows, or explicitly restrict this command to a single-prompt mode and record its actual source and sample count.

src/training/kv_mapper/eval.py
```diff
@@ -0,0 +1,115 @@
...
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+
+
+def kv_fit_metrics(pred: np.ndarray, true: np.ndarray) -> dict[str, float]:
+    pred = np.asarray(pred, dtype=np.float64).reshape(-1)
+    true = np.asarray(true, dtype=np.float64).reshape(-1)
+    n = min(pred.size, true.size)
```
Xunzhuo, line 15: [P2] Reject mismatched KV tensors instead of truncating them

Taking the smaller size silently discards missing predictions or unmatched targets before computing the fit metrics. For example, kv_fit_metrics([1, 2], [1, 2, 1000]) returns rel_err=0 and r2=1 even though the final target value has no prediction. These metrics are part of the held-out mapper qualification evidence, so incomplete captures must fail validation. Check that the original shapes match and are nonempty before flattening, with a regression for unequal shapes/lengths.

Xunzhuo (changes requested):
Please reject incomplete KV comparisons, make collection match its recorded recipe, and fix the inherited artifact integrity check.

Validation: all 18 KV mapper tests and the training-family layout test pass in a CPU environment; Training Contracts CI also reports all 674 cases passing, including the new tests. Additional small regressions reproduce the inline findings. GPU collection with real model weights, teacher-forced injection, and end-to-end KV reuse were not run.

Issue #3667, opened 2026-09-09
[Feature] Migrate MLP selector training with validation and runtime parity


Xunzhuo:
### Goal

Make MLP a reproducible, evaluated unified trainer with proper validation, artifact completeness, and exact native inference parity.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Derive the declared classification or utility target from the versioned selector objective rather than treating repeated candidate rows as contradictory labels.
- Use deterministic query-level train, validation, and test splits; early-stop on validation evidence and restore the best checkpoint.
- Record architecture, activations, preprocessing, feature identity, class/candidate order, objective, seed, optimizer, training snapshot, calibration, and numeric precision in the artifact.
- Add calibration or abstention behavior when required by the target contract and evaluate route share, regret, quality, latency, cost, and failure slices.
- Route device and precision selection through the executor/hardware plan instead of a frontend-only CPU/CUDA/MPS enum.
- Include MLP in artifact upload/download, validation, qualification, and runtime capability discovery.
- Add exact Python/native logits or utility scores, probabilities where applicable, and selected-model conformance tests.

### Non-goals

- Searching unbounded network architectures or promoting from training loss alone.
- Claiming accelerator support without a matching worker and qualification receipt.

### Completion signal

- [ ] MLP trains through the canonical run API with deterministic splits and restores the best validation checkpoint.
- [ ] The artifact is provenance-complete and accepted by the common validation and qualification flow.
- [ ] Python and native runtimes match declared scores and final selections within explicit numeric tolerances.
- [ ] Held-out evaluation prevents an artifact with regressed baselines from becoming binding-eligible.

Depends on: #3663, #3661, #3662

Parent Epic: #3660

Issue #3664, opened 2026-09-09
[Feature] Migrate KNN selector training to the unified contract


Xunzhuo:
### Goal

Make KNN a first-class unified trainer whose learned artifact, objective, and prediction behavior are identical between training and the native Router runtime.

### Proposed Workgroup

Router Models & Inference Runtime

### Scope

- Train from `QueryOutcomeSet` and the shared `FeatureSnapshot`, storing one neighbor feature per query with its candidate outcome vector rather than repeated conflicting class rows.
- Define and version the distance metric, normalization, neighbor weighting, missing-outcome behavior, and objective aggregation used to rank candidates.
- Remove the current Python/native mismatch between cosine-style training behavior, native L2 lookup, absolute latency factors, and global min-max latency scoring.
- Store candidate-set digest, `k`, metric, normalization, objective version, feature identity, and training snapshot identity in the artifact.
- Reject incompatible candidate pools and feature contracts before routing.
- Add held-out evaluation and exact Python/native score and selection conformance cases, including ties, missing outcomes, and degenerate neighborhoods.
- Register KNN through the common trainer/export/runtime capability interfaces rather than an algorithm-specific public API.

### Non-goals

- Adding approximate nearest-neighbor infrastructure or changing online routing-policy semantics.
- Maintaining compatibility with ambiguous artifacts that do not declare their feature and scoring contracts.

### Completion signal

- [ ] KNN trains and evaluates through the canonical run API from the shared selector snapshot.
- [ ] Python and native implementations produce matching neighbor order, candidate scores, and selected model for the conformance corpus.
- [ ] Candidate-set, feature, or objective drift is rejected with a typed diagnostic.
- [ ] The artifact passes the same qualification and binding-proposal flow used by other trainers.

Depends on: #3663, #3661, #3662

Parent Epic: #3660

PR #4022, opened 2026-09-22
feat(training): define one versioned selector objective

ankit373:
Second increment of #3663.

Scoring is written four times and the copies disagree. `data_loader` weighs a candidate against the slowest one in its own query group, KNN divides nanoseconds by 10s, KMeans/SVM/MLP divide milliseconds by 10s, and three of the four ignore the configured weight and hard-code `0.9 * quality + 0.1 * speed`. So a configured `quality_weight` cannot move a decision, and the group-relative rule can't be evaluated at inference at all, where one candidate arrives with no group.

`SelectorObjective` states it once: quality/latency/cost weights are configuration, failure is subtracted outright so a failed call can't win on recorded quality, latency and cost saturate against an absolute scale so a score means the same thing at labelling and inference, and `objective_id` stamps the weights onto the artifact. Additive; the four trainers migrate under #3664-#3667.

15 tests, mutation-checked (restoring the hard-coded 0.9/0.1 fails one). `make check CI_STATIC_ONLY=true` clean.


src/training/model_selection/ml_model_selection/objective.py
```diff
@@ -0,0 +1,110 @@
...
+            1.0 + max(0.0, float(outcome.latency_ms)) / self.latency_scale_ms
+        )
+        thrift = 1.0 / (1.0 + max(0.0, float(outcome.cost)) / self.cost_scale)
+        total = (
+            self.quality_weight * float(outcome.quality)
+            + self.latency_weight * speed
+            + self.cost_weight * thrift
+        )
+        if not outcome.success:
+            total -= self.failure_penalty
```
Xunzhuo: [P2] Keep failed candidates below successful candidates for every accepted objective

The fixed subtraction only disqualifies failures for the default weights. `SelectorObjective(quality_weight=1.0, latency_weight=1.0)` is accepted, but a failed candidate with quality 1.0 and latency 100 ms scores 0.9901, above a successful candidate with quality 0.2 and latency 30,000 ms (0.45). Both `best()` and `label_snapshots()` then choose the failed call, contradicting the failure-disqualification contract. Please enforce success eligibility independently of the weighted utility, or validate/normalize the objective so every accepted configuration preserves that ordering, and cover a non-default weighting in the regression test.

Xunzhuo (changes requested):
The new scoring rule is additive, and the objective tests pass. One correctness issue remains: an accepted non-default weighting can label a failed candidate as the winner despite an available successful candidate. Please make the failure eligibility invariant hold across supported configurations and add a regression for it.

ankit373:
Confirmed and fixed in 41e67693. Reproduced your case exactly: utilities were broke=1.9901 vs worked=0.4500, and `best()` returned `broke`.

You're right that no constant can work here, since the utility's bound is the weights themselves. Eligibility is now structural rather than numeric: `sort_key` ranks on success first, then utility, then `model_ref`, so a successful candidate outranks every failed one whatever the weights say. Failures still order among themselves for the query where all of them failed. `failure_penalty` is gone, since a tier can't be bought back.

My original test only covered the default weighting, which is the one configuration where the old arithmetic happened to hold. The regression now sweeps five weightings including yours, and reverting to the fixed subtraction fails four of them.

PR #4030, opened 2026-09-22
[Test] Derive the baseline-suite effective selection and gate its drift

gavinkvx:
Related #2379

#3501 derived the execution graph from `GetTestCases()` and left the baseline-suite selection layer outside the audit. This slice models it.

**What changes**
- `framework.EffectiveTestCases` extracts the runner's selection into one function (explicit `-tests` → baseline suite for `envoy-ai-gateway` → profile selection). `runTests` calls it; behavior is unchanged, the log line now names the deciding layer.
- `verification.DeriveEffectiveSelection` runs that same function for every profile × suite, so the audit reports what CI executes rather than a second selector.
- Gate E (`gates_selection_test.go`): for the baseline profile, every selected case runs under some suite; every suite omission is recorded by `testmatrix.BaselineStress`; no suite runs a case the profile does not select; no stale stress entry. Other profiles must be untouched by the suite. The workflow's `baseline-suite` default must be a suite `BaselineCases` accepts, carried by `E2E_BASELINE_SUITE`.
- `e2e-audit` gains an `effective_selection` block (per profile/suite selection, exclusions with reasons, baseline drift, CI dispatch).

**What the gate distinguishes**: registered-but-unreachable (Gate A) vs. omitted by one suite with a recorded reason (legitimate stress semantics) vs. declared by the profile but executed by no suite (fails, no allowlist).

**Verification**: `go test ./pkg/verification ./pkg/framework ./pkg/testmatrix` (19 verification tests), `make e2e-coverage-check`, `make build-e2e`; audit output byte-identical across runs (36 profiles / 194 testcases / 17 bounded unreachable / 0 CI overrides; standard 37, full 39, drift empty). Six live mutations (declared-but-never-run case, silent selector drop, stale stress entry, suite running outside the contract, invalid workflow default, runner skipping the suite layer) each turned the intended gate red and were reverted.

**Out of scope**: which CI plans dispatch `full` (`tools/ci/ci_plan.py`); the `-tests` explicit path (recorded, not gated); testcase metadata (next slice).


gavinkvx:
Opened #4030 for the effective-selection slice proposed above (production selector extracted and reused by the audit, Gate E on the baseline-suite layer). The minimal testcase-metadata proposal follows once it lands.

PR #4038, opened 2026-09-22
[Website] Add a capability comparison to the homepage

### LINKED ISSUE #4033: [Website] Add a capability comparison to the homepage
### Goal

The homepage already compares two things, and neither answers the question a
first-time evaluator arrives with:

- a "Fragmented today / With vLLM SR" dimension matrix at
  `website/src/pages/index.tsx:248` — a problem-to-solution framing with no third
  option in it;
- benchmark scores against one frontier model at
  `website/src/pages/index.tsx:182`.

### Proposed Workgroup

### Scope

Add one comparison section to the homepage, structured as:

| Capability | Semantic Router | AI gateway | llm-d |
| --- | --- | --- | --- |
| What it decides | which model, recipe and policy serve a request | how a request reaches a backend | which replica inside the chosen pool |
| What it reads | request content, policy, semantic evidence | protocol, credentials, rate limits | load, prefix-cache locality, replica health |
| What it owns | the model, recipe and policy decision | provider translation, credentials, rate limits, traffic policy | endpoint selection inside a pool |
| Where it runs | an Envoy ExtProc filter | the data plane | an endpoint picker |
| Decision receipt | `x-vsr-selected-model` | none | none |

Every cell is already stated by the repository, so the section introduces no new
claims: the Semantic Router and llm-d columns follow
`website/docs/installation/k8s/llm-d.md:10-15` and the boundary table at
`website/docs/faq.md:50-54`; the gateway column follows
`website/docs/installation/k8s/gateways.md:17` and `:35`; the receipt row follows
`website/docs/troubleshooting/vsr-headers.md:69`.

Compare **categories** rather than named products, so no row depends on an
unverified claim about a third party.

Render it in the site's own comparison visual language — the flat bordered grid,
uppercase monospace header and accent colour on the value column already defined
at `website/src/pages/index.module.css:76` — with the same `role="table"`,
`role="row"`, `role="columnheader"` and mobile-label semantics the existing
matrix uses.

### Non-goals

- No pricing, tier, licence or self-hosting comparison. That is the axis the
  example table is built on, and it would need claims about specific products
  that the repository cannot back. The shape is what carries over, not the rows.
- No claims about named competitor products; the columns are categories.
- No change to the docs site, the FAQ page, or to routing behaviour. The existing
  dimension matrix and the benchmark cards stay as they are.

### Completion signal

- The section renders on the homepage in the same visual language as the existing
  matrix, and each row states what Semantic Router, an AI gateway and llm-d own.
- `npm run build` succeeds from `website/`.
- The block carries the same table semantics and mobile labels as
  `website/src/pages/index.tsx:248`.

Zi-Yi-Ming:
Closes #4033

## Purpose

The homepage compares a fragmented status quo against vLLM Semantic Router, and benchmark scores against one frontier model. Neither answers the question a first-time evaluator arrives with — how Semantic Router relates to the AI gateway and llm-d they are also looking at.

This adds one comparison section to the homepage, structured as a feature-by-option table (Capability | Semantic Router | AI gateway | llm-d). Columns are categories rather than named products, so every cell is backed by the repository's own documentation instead of a claim about a third party. It renders in the site's existing comparison visual language — the same matrix grid, uppercase monospace header, accent colour on the value column, ARIA table semantics and mobile labels the capability matrix uses — with a four-column desktop template scoped to min-width 769px so the mobile stack is untouched.

While in the FAQ neighbourhood, two alignment fixes from the review thread:

- the sidebar: the FAQ was the only bare item sitting between two categories; it moves up next to the Introduction, so top-level entries stay at the top and the section flow is uninterrupted;
- the page: the arrow-prefixed footer links were the only pages on the site ending that way; they become inline sentences in the style the other pages use, with each link kept next to the answer it belongs to.

## Test Plan

- `npm run build` completes from `website/` on this branch: the node-based pre-steps (`sync:install`, `contributors:rank`, `committers:activity`) pass, and Docusaurus compiles both locales and emits static files successfully;
- `npx esbuild` parses the modified `index.tsx` with zero errors;
- every docs link in the modified `faq.md` resolves to a page that exists on `main`;
- the sidebar order after the change: Introduction, FAQ, Overview, Getting Started, then the category flow uninterrupted.

## Test Result

`npm run build` succeeds (`DOCUSAURUS_RC=0`, both locales emitted), with the new section rendering through the existing matrix classes, so the responsive behaviour (stacked cards with mobile labels under 768px) is inherited rather than reimplemented.

Not verified here: the generated-reference pre-step fails on Windows because `src/vllm-sr/cli/` imports `fcntl`, which does not exist there — that check is orthogonal to this change (no generated artifacts are modified) and runs on CI. Netlify preview covers the visual pass.


AayushSaini101:
Thanks for working @Zi-Yi-Ming can you add screenshots for the visualisation thanks : )

Zi-Yi-Ming:
@AayushSaini101  Added below.

**Top — the FAQ page.** Two changes: the FAQ entry now sits directly under the
Introduction in the sidebar instead of between the Overview and Getting Started
categories, and each answer closes with an inline link sentence instead of the
arrow-prefixed footer, which was the only page on the site ending that way.

**Bottom — the homepage comparison.** The new "Where it fits" section: one row per
capability, across Semantic Router, an AI gateway and llm-d. It renders through the
existing matrix classes, so it matches the Architecture section's visual language —
flat grid, uppercase monospace header, accent on the Semantic Router column.

The columns are categories rather than named products, so every cell is backed by the
documentation instead of a claim about a specific third party.


屏幕截图 2026-09-22 183721

屏幕截图 2026-09-22 183140

AayushSaini101:
Thanks for sharing, there are various line break presnet in the docs for FAQ can you make it clean, as well as in the difference table please use proper standard for mentioning like "An Endpoint Picker" instead of "a ..." thanks
image @Zi-Yi-Ming

Zi-Yi-Ming:
@AayushSaini101  Shortened the three headings so each fits on one line:

- "What is vLLM Semantic Router? Is it an AI gateway, or something else?" →
  "Is it an AI gateway, or something else?"
- "If llm-d already does multi-model routing, why Semantic Router?" →
  "Why Semantic Router and not just llm-d?"
- "How do Semantic Router and the llm-d Endpoint Picker avoid conflicts?" →
  "How do we avoid conflicts with llm-d?"

Screenshot below — the headings now sit on one line each; everything else on the
page is unchanged.


image

AayushSaini101 (approved):
Thanks for your hard work : )

AayushSaini101:
image
@Zi-Yi-Ming can you please update the Ui/UX here the heading of LLM-D is showing in blue color but according to the style it should be presnet over semantic router and same formatting of the text like "Captial words in the starting" after implementing that please share the screenshots thanks : )

Zi-Yi-Ming:
@AayushSaini101 Sure — since #4038 is already merged, this will go as a follow-up PR against main. Both issues are ours: the accent rule we inherited targets the last table column, which becomes llm-d once a fourth column is added, and the cell text was written as lowercase fragments while the capability matrix it mirrors opens its cells with capitalised words.

Before I open it: every PR here links an issue, and #4033 is closed now — would you like a small issue filed for this fix first, or is the follow-up PR on its own fine?

AayushSaini101:
> @AayushSaini101 Sure — since #4038 is already merged, this will go as a follow-up PR against main. Both issues are ours: the accent rule we inherited targets the last table column, which becomes llm-d once a fourth column is added, and the cell text was written as lowercase fragments while the capability matrix it mirrors opens its cells with capitalised words.
> 
> Before I open it: every PR here links an issue, and #4033 is closed now — would you like a small issue filed for this fix first, or is the follow-up PR on its own fine?

yes, sure please open a small follow PR : )

